PR #4, opened 2021-03-05
Fix TPU training in example

thomwolf:
We are using `accelerator.distributed_type` in our `collate_fn` but the `accelerator` is defined later.


sgugger:
Ah thanks a lot for fixing!

ngoquanghuy99 (commented):
thanks!

PR #8, opened 2021-03-09
More flexible RNG synchronization

sgugger:
Currently, the RNG synchronization at the beginning of each epoch is done very broadly, which means the potential data augmentation is done the same way on each process. While this is probably not hurting a big deal unless the underlying dataset isn't very big, there is a way to use local RNG with `torch.Generator` introduced in PyTorch 1.6.

This PR enables this by default and leaves some flexibility to the user to set the RNGs they want to synchronize (among torch, cuda, xla and local generator).

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import List, Optional, Union
+
 import torch
 
 from packaging import version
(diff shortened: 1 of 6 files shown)


src/accelerate/accelerator.py
```diff
@@ -42,6 +44,18 @@ class Accelerator:
         cpu (:obj:`bool`, `optional`):
             Whether or not to force the script to execute on CPU. Will ignore GPU available if set to :obj:`True` and
             force the execution on one process only.
+        rng_types (list of :obj:`str` or :class:`~accelerate.utils.RNGType`):
```
LysandreJik, line 47: I like this!

src/accelerate/utils.py
```diff
@@ -35,28 +44,47 @@ def set_seed(seed: int):
     np.random.seed(seed)
     torch.manual_seed(seed)
     torch.cuda.manual_seed_all(seed)
+    if is_tpu_available():
+        xm.set_rng_state(seed)
     # ^^ safe to call this function even if cuda is not available
```
LysandreJik, line 49: Maybe you can move this comment above if you want to keep it

LysandreJik (approved):
LGTM! Thanks for the in-person explanation :)

PR #13, opened 2021-03-17
Don't error on non-Tensors objects in move to device

### LINKED ISSUE #12: Can't send the values of int to device
My training data looks like:

src_image, target_image, src_camera, target_camera, src_camera_idx, target_camera_idx

Where src_camera_idx, target_camera_idx are integers

When I try to apply accelerate I get the following error:
TypeError: Can't send the values of type <class 'int'> to device cuda:0, only of nested list/tuple/dicts of tensors or objects having a `to` method.

We don't need to send the integers to the device. Perhaps instead of raising an error here, you can simply skip the items that cannot be moved to device? Or at least give me the option to skip them if I know my data has such objects.

sgugger:
As pointed out in #12, users may have some non-Tensor objects in their dataloaders, which is perfectly fine. `accelerate` should not error on those when trying to send a batch on the proper device but just do a noop.

Fixes #12

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -105,10 +105,7 @@ def send_to_device(tensor, device):
     elif isinstance(tensor, dict):
         return type(tensor)({k: send_to_device(v, device) for k, v in tensor.items()})
     elif not hasattr(tensor, "to"):
-        raise TypeError(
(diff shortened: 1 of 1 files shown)


LysandreJik (approved):
LGTM. I understand that this is applied automatically on each dataloader and output, hence why putting a warning isn't warranted here as it would just flood the user's logs unnecessarily.

PR #15, opened 2021-03-22
Add KwargsHandlers

### LINKED ISSUE #11: How to pass custom arguments to DDP / GradScaler
Thanks for creating this nice tool!

I am wondering what's the best way to use accelerate, but still keep the capability to tune the arguments of DDP / GradScaler / (perhaps other classes as well). They have some pretty useful arguments (e.g. https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html).

sgugger:
This PR introduces a new argument in the main `Accelerator` class called `kwargs_handlers` which can take a list of config-like objects to customize how the `DistributedDataParallel` or `GradScaler` are created.

For instance, to activate `find_unused_parameters = True`, one would create the `Accelerator` like this:

```
from accelerate import Accelerator, DistributedDataParallelKwargs

ddp_handler = DistributedDataParallelKwargs(find_unused_parameters=True)
accelerator = Accelerator(kwargs_handlers=[ddp_handler])
```

`kwargs_handlers` takes the list of all `KwargsHandler`s desired by the user, to make the API easily scalable to more objects we want to make customizable by the users.

Fixes #11

### DIFF
--- a/examples/cv_example.py
+++ b/examples/cv_example.py
@@ -22,7 +22,6 @@ from torch.optim.lr_scheduler import OneCycleLR
 from torch.utils.data import DataLoader, Dataset
 
 import PIL
-import torchvision
 from accelerate import Accelerator
 from timm import create_model
 from torchvision.transforms import Compose, RandomResizedCrop, Resize, ToTensor
--- a/examples/nlp_example.py
+++ b/examples/nlp_example.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import argparse
 
+import torch
 from torch.utils.data import DataLoader
 
 from accelerate import Accelerator, DistributedType
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -64,7 +64,7 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
         return self.optimizer.param_groups
 
     @param_groups.setter
-    def param_group(self, param_groups):
+    def param_groups(self, param_groups):
         self.optimizer.param_groups = param_groups
 
     @property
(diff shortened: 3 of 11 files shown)


examples/cv_example.py
```diff
@@ -22,7 +22,6 @@
 from torch.utils.data import DataLoader, Dataset
 
 import PIL
-import torchvision
```
sgugger, line 25: Unrelated to this PR, this is to make quality pass.

examples/nlp_example.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 import argparse
 
+import torch
```
sgugger, line 17: Unrelated to this PR, this is to make quality pass.

src/accelerate/optimizer.py
```diff
@@ -64,15 +64,15 @@ def param_groups(self):
         return self.optimizer.param_groups
 
     @param_groups.setter
-    def param_group(self, param_groups):
+    def param_groups(self, param_groups):
```
sgugger, line 67: Unrelated to this PR, this is to make quality pass.

src/accelerate/optimizer.py
```diff
@@ -64,15 +64,15 @@ def param_groups(self):
...
         self.optimizer.param_groups = param_groups
 
     @property
     def defaults(self):
         return self.optimizer.defaults
 
     @defaults.setter
-    def param_group(self, defaults):
+    def defaults(self, defaults):
```
sgugger, line 75: Unrelated to this PR, this is to make quality pass.

LysandreJik (approved):
I like the idea, and I love the implementation! It's great that we're keeping IntelliSense for the kwargs, while at the same time limiting to only those accepted by `Accelerate`.

Regarding tests, maybe they could be a bit more complete by trying to pass a `DistributedDataParallelKwargs` or a `GradScalerKwargs` and ensuring that the `self.scaler` and `model` are indeed affected by these? 

This way you wouldn't only test the kwargs handler main behavior, but also their implementation within the `Accelerate` class.

LysandreJik (approved):
Very nice, thanks for adding the tests!

PR #16, opened 2021-03-24
Add YAML config support

sgugger:
This PR adds support and defaults toward a YAML config while retaining backward compatibility: it's possible to pass along a json config file and if a default JSON config was created by the user, it will still be used.

However new configs created will default to the YAML format.

### DIFF
--- a/src/accelerate/commands/config.py
+++ b/src/accelerate/commands/config.py
@@ -18,8 +18,10 @@ import argparse
 import json
 import os
 from dataclasses import dataclass
+from enum import Enum
 from typing import Optional
 
+import yaml
 from accelerate.state import DistributedType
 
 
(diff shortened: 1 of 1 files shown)


src/accelerate/commands/config.py
```diff
@@ -41,17 +50,39 @@ class LaunchConfig:
...
 
+    @classmethod
+    def from_yaml_file(cls, yaml_file=None):
+        yaml_file = default_yaml_config_file if yaml_file is None else yaml_file
+        with open(yaml_file, "r", encoding="utf-8") as f:
+            return cls(**yaml.safe_load(f))
+
+    def to_yaml_file(self, yaml_file):
+        with open(yaml_file, "w", encoding="utf-8") as f:
+            yaml.dump(self.to_dict(), f)
```
philschmid, line 84:
```suggestion
            yaml.safe_dump(self.to_dict(), f)
```

> safe_dump(data, stream=None) serializes the given Python object into the stream. If stream is None, it returns the produced stream. safe_dump produces only standard YAML tags and cannot represent an arbitrary Python object.

philschmid (approved):
LGTM. The only thing, which we could change is the `dump` method to `safe_dump`

PR #19, opened 2021-03-26
Add utility to pad tensor across processes to max length

sgugger:
This PR solves the problem of having predictions of different sizes across processes (for instance when doing padding to the maximum length of a batch) that can't be gathered (the script will hang).
To do this it introduces a new method called `pad_across_processes` that will pad a tensor (or nested list/tuple/dict of tensors) to the maximum length across processes on a given dimension.


LysandreJik (approved):
LGTM! As said in person, let's take some time next week to set up some GPU/Multi-GPU CI. According to how long it takes, we can probably leverage Transformers' CI.

PR #20, opened 2021-03-26
Alternate diff

sgugger:
I recently discovered you can have a \`\`\`diff, so trying it for a spin hear.


LysandreJik (approved):
This looks great! Fun to see that when reviewing files changed, it also applies the color changes directly from the diff view

PR #21, opened 2021-03-29
Use proper size

sgugger:
The `size` used inside the `pad_across_processes` function was wrong, this PR fixes that.


LysandreJik (approved):
LGTM!

PR #22, opened 2021-03-30
Cleaner diffs in README and index

sgugger:
This PR adds the diff styling to the examples in the index of the documentation and indents all the code in the README/this index by two to make everything align properly.


LysandreJik (approved):
This looks good!

PR #23, opened 2021-03-30
Add defaults for compute_environmnent

sgugger:
#17 introduced a breaking change since it requires configs to now have a `"compute_environment"` field. This is fixed by re-running `accelerate config` but we can also be nice to the user and make sure it works for old configs as well :-)

This PR does that.


philschmid (approved):
LGTM. Sorry for the bug. I haven´t tested old configurations.

sgugger:
No problem @philschmid, I should have caught it during the review :-)

PR #26, opened 2021-04-09
Launch script on sagemaker

philschmid:
# This PR adds single-gpu support for Amazon SageMaker

I've been working on a SageMaker integration accelerate to launch your training script on SageMaker the same way as you do for TPU, GPU etc.
I finished the version of it and tested it with the `nlp_example.py`
To get started you run
```
~/Projects/huggingface/accelerate: accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 1
```
fill out all required information

and then launch with
```
accelerate launch ./examples/nlp_example.py
```
Output
```
Configuring Amazon SageMaker environment
Converting hyperparamters
Creating estimator
2021-04-08 11:56:50 Starting - Starting the training job...
2021-04-08 11:57:13 Starting - Launching requested ML instancesProfilerReport-1617883008: InProgress
.........
2021-04-08 11:58:54 Starting - Preparing the instances for training.........
2021-04-08 12:00:24 Downloading - Downloading input data
2021-04-08 12:00:24 Training - Downloading the training image..................
2021-04-08 12:03:39 Training - Training image download completed. Training in progress..

........

epoch 0: {'accuracy': 0.7598039215686274, 'f1': 0.8178438661710037}
epoch 1: {'accuracy': 0.8357843137254902, 'f1': 0.882249560632689}
epoch 2: {'accuracy': 0.8406862745098039, 'f1': 0.8869565217391304}

........

2021-04-08 12:05:40 Uploading - Uploading generated training model
2021-04-08 12:05:40 Completed - Training job completed
Training seconds: 331
Billable seconds: 331
You can find your model data at: s3://sagemaker-us-east-1-558105141721/accelerate-sagemaker-1-2021-04-08-11-56-47-108/output/model.tar.gz
```
I also created a new Documentation Section "Guides" and added SageMaker there. 

screencapture-127-0-0-1-5500-docs-build-html-sagemaker-html-2021-04-09-11_39_15

P.S. not user if should always add 🤗 in front of Accelerate.


examples/nlp_example.py
```diff
@@ -162,8 +162,8 @@ def collate_fn(examples):
 
 def main():
     parser = argparse.ArgumentParser(description="Simple example of training script.")
-    parser.add_argument("--fp16", action="store_true", help="If passed, will use FP16 training.")
-    parser.add_argument("--cpu", action="store_true", help="If passed, will train on the CPU.")
+    parser.add_argument("--fp16", type=bool, default=False, help="If passed, will use FP16 training.")
+    parser.add_argument("--cpu", type=bool, default=False, help="If passed, will train on the CPU.")
```
sgugger, line 166: I see you did something for those later on, is it still necessary? We should make sure we support those as they are very common, or if they are not, it should be made clear in the docs.

examples/requirements.txt
```diff
@@ -0,0 +1 @@
+accelerate # used to be installed in Amazon SageMaker environment
```
sgugger, line 1: Should we amke one folder for each example? The CV example would require more.

src/accelerate/commands/launch.py
```diff
@@ -168,11 +184,100 @@ def tpu_launcher(args):
...
+    parser = argparse.ArgumentParser()
+    _, unknown = parser.parse_known_args(nargs)
+    for index, argument in enumerate(unknown):
+        if argument.startswith(("-", "--")):
+            action = None
+            if index + 1 < len(unknown):  # checks if next index would be in list
+                if unknown[index + 1].startswith(("-", "--")):  # checks if next element is an key
+                    action = "store_true"
+            else:  # action_true parameter is last element
+                action = "store_true"
```
sgugger, line 211: There should be something for `store_false` too.

docs/source/sagemaker.md
```diff
@@ -0,0 +1,134 @@
...
+
+```bash
+accelerate launch path_to_script.py --args_to_the_script
+```
+
+This will launch your training script using your configuration. The only thing you have to do is provide all the arguments needed by your training script as named arguments.
+
+**Examples**
+
+_**Note**: If you run one of the example scripts, don't for get to add `accelerator.save('/opt/ml/model')`to it._
```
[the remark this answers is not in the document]
philschmid: The current example doesn't show how to save models at all. So that's why I added it here.

examples/nlp_example.py
philschmid: SageMaker cannot pass `store_true` or `store_false` to a script executed. 

> Note that SageMaker doesn’t support argparse actions. If you want to use, for example, boolean hyperparameters, you need to specify type as bool in your script and provide an explicit True or False value for this hyperparameter when instantiating PyTorch Estimator.

https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#prepare-a-pytorch-training-script

should we add it to the documentation then?

examples/requirements.txt
philschmid: for the CV example to work we would only need to add `timm`. `torchvision` should be inside the DLC since it uses the AWS PyTorch DLC as base

src/accelerate/commands/launch.py
philschmid: puh. I that even possible? like how can someone cloud tell the difference between `store_true` and `store_false` without knowing the script...

examples/nlp_example.py
sgugger: Yes, it should be properly documented.

src/accelerate/commands/launch.py
sgugger: I have no idea what you're doing it here then. Also I though the script could not have `--store_true`/`--store_false`

On top of the documentation, we should add a nice error message here if the parser has arguments with `--store_true`/`--store_false`.

philschmid:
screencapture-127-0-0-1-5500-docs-build-html-sagemaker-html-2021-04-13-09_51_47

examples/nlp_example.py
philschmid: I reworked the documentation to use `.rst` to highlight it as a warning. ([see PR comment](https://github.com/huggingface/accelerate/pull/26#issuecomment-818525901))

src/accelerate/commands/launch.py
philschmid: For SageMaker to work I needed to convert the `args.training_script_args` from `--foo bar --john 2 --store_true --another key` to
 ```python
{
"foo":"bar",
"john": 2,
"store_true": True,
"another": "key"
}
```
To achieve this I built this dynamic args parser, which takes the `args.training_script_args` and creates an `argsparser` out of it to apply afterward on `args.training_script_args` to create this `dict`. 

While creating this I haven't thought of `store_false`, since I never have seen it in real use, that's why I expected that there is only `store_true` and implemented it like that.

File: examples/requirements.txt (hunk 2 above)
Comment (philschmid) [reply examples/requirements.txt:1]: Should we add `timm` to the `requirements.txt` to make both examples runnable out of the box?

File: src/accelerate/commands/launch.py (hunk 3 above)
Comment (philschmid) [reply src/accelerate/commands/launch.py]: As discussed I adjusted the convert script to raise an error if someone uses `store` actions and added a case into the test.

File: docs/source/sagemaker.rst [hunk 5]
```diff
@@ -0,0 +1,169 @@
...
+
+Getting Started
+-----------------------------------------------------------------------------------------------------------------------
+
+Setup & Installation
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+Before you can run your 🤗 Accelerate scripts on Amazon SageMaker you need to sign up for an AWS account. If you do not
+have an AWS account yet learn more `here <https://docs.aws.amazon.com/sagemaker/latest/dg/gs-set-up.html>`_.
```
sgugger, line 33:
```suggestion
have an AWS account yet learn more `here <https://docs.aws.amazon.com/sagemaker/latest/dg/gs-set-up.html>`__.
```

docs/source/sagemaker.rst
```diff
@@ -0,0 +1,169 @@
...
+.. code-block:: diff
+
+    - torch.save()
+    + accelerator.save('/opt/ml/model')
+
+
+.. warning::
+    SageMaker doesn’t support argparse actions. If you want to use, for example, boolean hyperparameters, you need to
+    specify type as bool in your script and provide an explicit True or False value for this hyperparameter. `[REF]
+    <https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#prepare-a-pytorch-training-script>`_
```
sgugger, line 88:
```suggestion
    <https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#prepare-a-pytorch-training-script>`__.
```

src/accelerate/commands/launch.py
```diff
@@ -168,11 +184,105 @@ def tpu_launcher(args):
...
+    source_dir = os.path.dirname(args.training_script)
+    if not source_dir:  # checks if string is empty
+        source_dir = "."
+    entry_point = os.path.basename(args.training_script)
+    if not entry_point.endswith(".py"):
+        raise ValueError(f'Your training script should be a python script and not "{entry_point}"')
+
+    print("Converting Arguments to Hyperparameters")
+    converted_hyperparameters = _convert_nargs_to_dict(args.training_script_args)
+    hyperparameters = {"fp16": args.fp16, **converted_hyperparameters}
```
sgugger, line 261: fp16 might not be accepted by the script so it should not be added like this. It should be added in the env instead, in the variable `"USE_FP16"` (which is what accelerate will look for).

sgugger (approved):
Thanks for updating, three last tiny comments and we should be good to merge!

LysandreJik (approved):
Great, LGTM!

PR #35, opened 2021-04-20
fix cluster.py indent error

JTT94:
Remove indent from https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/config/cluster.py, led to undefined num_processes when using TPU.


sgugger (approved):
Thank you so much for fixing this!

PR #43, opened 2021-04-23
Add `set_to_none` to AcceleratedOptimizer.zero_grad

### LINKED ISSUE #42: AcceleratedOptimizer `zero_grad` argument not supported: `set_to_none`
Currently the `AcceleratedOptimizer` class doesn't support the argument `set_to_none`, is this an intentional exclusion?

sgugger:
This PR adds support for the keyword argument `set_to_none` in `AcceleratedOptimizer.zero_grad`, with a version check as PyTorch only introduced it in v1.7.0.

Fixes #42

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -14,6 +14,8 @@
 
 import torch
 
+from packaging import version
+
 from .state import AcceleratorState, DistributedType, is_tpu_available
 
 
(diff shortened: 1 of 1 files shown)


src/accelerate/optimizer.py
```diff
@@ -86,8 +88,18 @@ def load_state_dict(self, state_dict):
...
+    def zero_grad(self, set_to_none=None):
+        if version.parse(torch.__version__) < version.parse("1.7.0"):
+            if set_to_none is not None:
+                raise ValueError(
+                    "`set_to_none` for Optimizer.zero_grad` was introduced in PyTorch 1.7.0 and can't be used for "
+                    f"earlier versions (found version {torch.__version__})."
+                )
+            self.optimizer.zero_grad()
+        else:
+            if set_to_none is not None:
```
wangyuxinwhy, line 100: if set_to_none is None:

sgugger: Good catch!

PR #56, opened 2021-04-27
Support for multi-GPU in notebook_launcher

### LINKED ISSUE #36: Multi GPU Training Not Working
While using Accelerate, it is only utilizing 1 out of the 2 GPUs present. I am training using the general instructions in the repository. The architecture is AutoEncoder. 

```
dataloader = DataLoader(dataset, batch_size = 2048, shuffle=True, pin_memory=False, num_workers=20)
encoder = Encoder(bottleneck_size = 2, embedding_size = 40, vocab = dataset.vocab).to(device)
decoder = Decoder(bottleneck_size = 2, embedding_size = 40, vocab = dataset.vocab).to(device)
model = AutoEncoder(encoder, decoder).to(device)
loss = nn.CrossEntropyLoss()
optimizer = torch.optim.Adam(model.parameters())
model, optimizer, dataloader = accelerator.prepare(model, optimizer, dataloader)
```
I am transferring the samples in the batch to the device using the code below:
```
    for x in batch:
        batch[x] = batch[x].to(device)
```
The device is being determined by using:

```device = accelerator.device```

Both devices are visible which can be confirmed by using ```torch.cuda.device_count()``` which returns 2.

Devices are RTX 2080 with CUDA Version 11.2. Driver version is 460.67.
Distro is PopOS!.

### PR DESCRIPTION
**Pull request by sgugger**
This PR adds support for multi-GPU training in the `notebook_launcher` and cleans a bit the API.

The same constraints as for TPU are applicable: no defining `accelerator` outside of the training function **and** CUDA has to be uninitiliazed until inside the trainnig function or the forking process is not happy.

Fixes #36

### DIFF
--- a/src/accelerate/notebook_launcher.py
+++ b/src/accelerate/notebook_launcher.py
@@ -14,7 +14,6 @@
 
 import os
 import sys
-import warnings
 
 import torch
 from torch.multiprocessing import start_processes
@@ -68,9 +67,9 @@ def notebook_launcher(function, args=(), num_processes=None, use_fp16=False):
         else:
             # No need for a distributed launch otherwise as it's either CPU or one GPU.
             if torch.cuda.is_available():
-                print(f"Launching training on one GPU.")
+                print("Launching training on one GPU.")
             else:
-                print(f"Launching training on CPU.")
+                print("Launching training on CPU.")
             function(*args)
 
     else:
@@ -115,7 +114,7 @@ def notebook_launcher(function, args=(), num_processes=None, use_fp16=False):
         else:
             # No need for a distributed launch otherwise as it's either CPU or one GPU.
             if torch.cuda.is_available():
-                print(f"Launching training on one GPU.")
+                print("Launching training on one GPU.")
             else:
-                print(f"Launching training on CPU.")
+                print("Launching training on CPU.")
             function(*args)
--- a/src/accelerate/notebook_launcher.py
+++ b/src/accelerate/notebook_launcher.py
@@ -22,7 +22,7 @@ from .state import AcceleratorState
 from .utils import PrepareForLaunch
 
 
-def notebook_launcher(function, args=(), num_processes=None, use_fp16=False):
+def notebook_launcher(function, args=(), num_processes=None, use_fp16=False, use_port="29500"):
     """
     Launches a training function, using several processes if it's possible in the current environment (TPU with
     multiple cores for instance).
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/accelerate/notebook_launcher.py [hunk 1]
```diff
@@ -65,15 +61,60 @@ def notebook_launcher(function, args=(), num_processes=None, start_method="fork"
...
+                    "To launch a multi-GPU training from your notebook, you need to avoid running any instruction "
+                    "using `torch.cuda` in any cell. Restart your notebook and make sure no cells use any CUDA "
+                    "function."
+                )
+
+            # torch.distributed will expect a few environment variable to be here. We set the ones common to each
+            # process here (the other ones will be set be the launcher).
+            os.environ["WORLD_SIZE"] = str(num_processes)
+            os.environ["MASTER_ADDR"] = "127.0.0.1"
+            os.environ["MASTER_PORT"] = "29500"
```
philschmid, line 101: I am not that experienced with `torch.distributed`, but is it good to have them hardcoded and not overrideable? We could add parameters to adjust them.

philschmid (approved):
LGTM. Had one question regarding hardcoded values

src/accelerate/notebook_launcher.py
sgugger: Those are the default values in `torch.distributed.launch`. The master address should not change since this does not support multi-node training. I can add a kwarg for the master port.

philschmid: Yeah saw the default [here](https://github.com/pytorch/pytorch/blob/master/torch/distributed/elastic_launch.py#L394). But who knows what setup customers are using, e.g. IPV6, custom cnames etc.

src/accelerate/notebook_launcher.py
```diff
@@ -65,15 +63,60 @@ def notebook_launcher(function, args=(), num_processes=None, start_method="fork"
...
+                    "To launch a multi-GPU training from your notebook, you need to avoid running any instruction "
+                    "using `torch.cuda` in any cell. Restart your notebook and make sure no cells use any CUDA "
+                    "function."
+                )
+
+            # torch.distributed will expect a few environment variable to be here. We set the ones common to each
+            # process here (the other ones will be set be the launcher).
+            os.environ["WORLD_SIZE"] = str(num_processes)
+            os.environ["MASTER_ADDR"] = "127.0.0.1"
+            os.environ["MASTER_PORT"] = use_port
```
philschmid, line 103:
```suggestion
            os.environ["MASTER_PORT"] = str(use_port)
```
Would add the extra conversion, might be else error prone

PR #58, opened 2021-04-28
update launch.py

Jesse1eung:
update launch.py for supporting the case in which one wanna run multi accelerate instances in single node multigpu. avoiding the port conflict.


sgugger (approved):
Thanks for the fix! I hadn't thought of that case but it makes total sense.

Issue #61, opened 2021-05-03
accelerator.prepare fails with IterableDataset


ddkalamk:
I am trying to use accelerator library in my example script but it is failing when trying to use with IterableDataset. Here is the error message:

```
Traceback (most recent call last):
  File "run_pretrain.py", line 612, in <module>
    main()
  File "run_pretrain.py", line 466, in main
    model, optimizer, train_dataloader, eval_dataloader = accelerator.prepare(
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/accelerator.py", line 201, in prepare
    result = tuple(self._prepare_one(obj) for obj in args)
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/accelerator.py", line 201, in <genexpr>
    result = tuple(self._prepare_one(obj) for obj in args)
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/accelerator.py", line 159, in _prepare_one
    return self.prepare_data_loader(obj)
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/accelerator.py", line 231, in prepare_data_loader
    return prepare_data_loader(
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/data_loader.py", line 416, in prepare_data_loader
    return DataLoaderShard(
  File "/data/nfs_home/ddkalamk/bert/accelerate/src/accelerate/data_loader.py", line 280, in __init__
    super().__init__(dataset, **kwargs)
  File "/nfs_home/ddkalamk/anaconda3/envs/bert/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 194, in __init__
    raise ValueError(
ValueError: DataLoader with IterableDataset: expected unspecified batch_sampler option, but got batch_sampler=<torch.utils.data.sampler.BatchSampler object at 0x2ba0839164f0>
srun: error: pcl-skx35: task 0: Exited with exit code 1
```

I am using pytorch v1.6.0 but it seems it has same issue even with latest pytorch.
Is there anything I am missing?

sgugger:
Fixed by #62. Thanks for flagging and fixing :-)

PR #62, opened 2021-05-03
Fix batch_sampler error for IterableDataset

ddkalamk:
Possible fix for https://github.com/huggingface/accelerate/issues/61

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -418,7 +418,8 @@ def prepare_data_loader(
     }
 
     # Need to provide batch_size as batch_sampler is None for Iterable dataset
-    if new_batch_sampler is None: kwargs["batch_size"] = dataloader.batch_size // num_processes if split_batches else dataloader.batch_size
+    if new_batch_sampler is None:
+        kwargs["batch_size"] = dataloader.batch_size // num_processes if split_batches else dataloader.batch_size
 
     return DataLoaderShard(
         new_dataset,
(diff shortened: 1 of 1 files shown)


src/accelerate/data_loader.py
```diff
@@ -365,6 +365,9 @@ def prepare_data_loader(
 
     new_dataset = dataloader.dataset
     new_batch_sampler = dataloader.batch_sampler
+    if isinstance(new_dataset, IterableDataset):
+        # Iterable dataset doesn't like batch_sampler, data_loader creates a default one for it
+        new_batch_sampler = None
```
sgugger, line 370:
```suggestion
    # Iterable dataset doesn't like batch_sampler, but data_loader creates a default one for it
    new_batch_sampler = dataloader.batch_sampler if not isinstance(new_dataset, IterableDataset) else None
```
Let's group the two lines together.

src/accelerate/data_loader.py
```diff
@@ -413,6 +416,10 @@ def prepare_data_loader(
         for k in _PYTORCH_DATALOADER_KWARGS
         if k not in ignore_kwargs
     }
+
+    # Need to provide batch_size as batch_sampler is None for Iterable dataset
+    if new_batch_sampler is None: kwargs["batch_size"] = dataloader.batch_size // num_processes if split_batches else dataloader.batch_size
```
sgugger, line 421:
```suggestion
    # Need to provide batch_size as batch_sampler is None for Iterable dataset
    if new_batch_sampler is None:
        kwargs["batch_size"] = dataloader.batch_size // num_processes if split_batches else dataloader.batch_size
```

sgugger (approved):
Thanks a lot for fixing this!

Issue #32, opened 2021-04-19
dsitributed on mulit-CPU


arita37:
Just wondering if there is way to distribute over CPU  (single node, or multi nodes).
It would be very useful features for some sparse models.

sgugger:
I don't think this is supported for now. How do you use PyTorch for this use case? Accelerate is syntax sugar over PyTorch, so if the feature does not exist in PyTorch, we probably won't be able to add it.

arita37:
Horovod can handle on distributed CPU:
https://github.com/huggingface/accelerate/issues/32#issuecomment-822466053

sgugger:
Ok, so support for Horovod is what you need. It's on the roadmap :-)

ddkalamk:
> I don't think this is supported for now. How do you use PyTorch for this use case? Accelerate is syntax sugar over PyTorch, so if the feature does not exist in PyTorch, we probably won't be able to add it.

Could you please consider support for MPI backend in torch DDP? Ideally, I want to specify backend from command line argument so I can specify some custom backend. An example use for Intel oneCCL backend (using Intel torch-ccl) can be found at https://github.com/ddkalamk/transformers/blob/pcl-v4.0.0/examples/question-answering/run_squad.py#L746

Assumption is we launch the application using MPI (similar to horovod) and based on environment variable initialize distributed backend.

sgugger:
Will look into this one too!

sgugger:
Should be fixed by #63

PR #63, opened 2021-05-03
Add distributed multi-node cpu only support (MULTI_CPU)

ddkalamk:
Possible fix for https://github.com/huggingface/accelerate/issues/32

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,15 @@ For instance, here is how you would run the GLUE example on the MRPC task (from 
 accelerate launch examples/nlp_example.py
 ```
 
+## Launching multi-CPU run using MPI
+
+🤗 Here is another way to launch multi-CPU run using MPI. You can learn how to install Open MPI on [this page](https://www.open-mpi.org/faq/?category=building#easy-build). You can use Intel MPI or MVAPICH as well.
+Once you have MPI setup on your cluster, just run:
+
+```bash
+mpirun -np 2 python examples/nlp_example.py
+```
+
 ## Launching your training from a notebook
 
 🤗 Accelerate also provides a `notebook_launcher` function you can use in a notebook to launch a distributed training. This is especially useful for Colab or Kaggle notebooks with a TPU backend. Just define your training loop in a `training_function` then in your last cell, add:
@@ -188,6 +197,8 @@ pip install accelerate
 ## Supported integrations
 
 - CPU only
+- multi-CPU on one node (machine)
+- multi-CPU on several nodes (machines)
 - single GPU
 - multi-GPU on one node (machine)
 - multi-GPU on several nodes (machines)
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -223,6 +223,9 @@ class Accelerator:
                 output_device=self.local_process_index,
                 **kwargs,
             )
+        elif self.distributed_type == DistributedType.MULTI_CPU:
+            kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
+            model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
         if self.native_amp:
             model.forward = torch.cuda.amp.autocast()(model.forward)
         return model
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -22,16 +22,16 @@ from .config_utils import _ask_field, _convert_distributed_mode, _convert_yes_no
 
 def get_cluster_input():
     distributed_type = _ask_field(
-        "Which type of machine are you using? ([0] No distributed training, [1] multi-GPU, [2] TPU): ",
+        "Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): ",
         _convert_distributed_mode,
-        error_message="Please enter 0, 1 or 2.",
+        error_message="Please enter 0, 1, 2 or 3.",
     )
 
     machine_rank = 0
     num_machines = 1
     main_process_ip = None
     main_process_port = None
-    if distributed_type == DistributedType.MULTI_GPU:
+    if distributed_type == DistributedType.MULTI_GPU or distributed_type == DistributedType.MULTI_CPU:
         num_machines = _ask_field(
             "How many different machines will you use (use more than 1 for multi-node training)? [1]: ",
             lambda x: int(x),
--- a/src/accelerate/commands/config/config_utils.py
+++ b/src/accelerate/commands/config/config_utils.py
@@ -37,7 +37,7 @@ def _convert_compute_environment(value):
 
 def _convert_distributed_mode(value):
     value = int(value)
-    return DistributedType(["NO", "MULTI_GPU", "TPU"][value])
+    return DistributedType(["NO", "MULTI_CPU", "MULTI_GPU", "TPU"][value])
 
 
 def _convert_sagemaker_distributed_mode(value):
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -20,11 +20,32 @@ import torch
 
 
 try:
+    import torch_ccl  # noqa: F401
+
+    _ccl_available = True
+except ImportError:
+    _ccl_available = False
+
+
+try:
     import torch_xla.core.xla_model as xm
 
     _tpu_available = True
 except ImportError:
     _tpu_available = False
+
+
+def get_int_from_env(env_keys, default):
+    """Returns the first positive env value found in the `env_keys` list or the default."""
+    for e in env_keys:
+        val = int(os.environ.get(e, -1))
+        if val >= 0:
+            return val
+    return default
+
+
+def is_ccl_available():
+    return _ccl_available
 
 
 def is_tpu_available():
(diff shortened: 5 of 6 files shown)


src/accelerate/accelerator.py
```diff
@@ -223,6 +223,12 @@ def prepare_model(self, model):
                 output_device=self.local_process_index,
                 **kwargs,
             )
+        elif self.distributed_type == DistributedType.MULTI_CPU:
+            kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
+            model = torch.nn.parallel.DistributedDataParallel(
+                model,
+                **kwargs,
+            )
```
sgugger, line 231:
```suggestion
            model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
```
This can probably fit in one line.

src/accelerate/commands/config/cluster.py
```diff
@@ -22,16 +22,16 @@
 
 def get_cluster_input():
     distributed_type = _ask_field(
-        "Which type of machine are you using? ([0] No distributed training, [1] multi-GPU, [2] TPU): ",
+        "Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): ",
```
sgugger, line 25:
```suggestion
        "Which type of machine are you using? ([0] No distributed training, [1] multinode-CPU, [2] multi-GPU, [3] TPU): ",
```
This is not multi-CPU on one machine, it's multiple machines for CPU, so I think we should be more explicit than multi-CPU

src/accelerate/state.py
```diff
@@ -27,6 +35,18 @@
     _tpu_available = False
 
 
+def env2int(env_list, default=-1):
```
sgugger, line 38: The name is a bit too short, and we don't use 2 like this in this repo (for instance we have `send_to_device` not `send2device`). It also returns the first match What about `get_int_from_env`?

`env_list` is also not the right name for the argument as we don't pass a list of envs, but a lsit of names to query in the env, so I would use `env_keys` instead of `env_list`. `default` should not have a default value since there is not a canonical one (the function is used with -1 once, 0 twice and 1 twice below).

Lastly, a short docstring that explains what this function does (returning the first positive env found in the list or the default) would be awesome!

src/accelerate/commands/config/config_utils.py
```diff
@@ -37,7 +37,7 @@ def _convert_compute_environment(value):
 
 def _convert_distributed_mode(value):
     value = int(value)
-    return DistributedType(["NO", "MULTI_GPU", "TPU"][value])
+    return DistributedType(["NO", "MULTI_CPU", "MULTI_GPU", "TPU"][value])
```
sgugger, line 40: I think the value should also be `MULTI_NODE_CPU`.

src/accelerate/state.py
```diff
@@ -123,12 +146,43 @@ def __init__(self, fp16: bool = None, cpu: bool = False, _from_accelerator: bool
...
+                else:
+                    backend = "gloo"
+                rank = env2int(["RANK", "PMI_RANK", "OMPI_COMM_WORLD_RANK", "MV2_COMM_WORLD_RANK"], 0)
+                size = env2int(["WORLD_SIZE", "PMI_SIZE", "OMPI_COMM_WORLD_SIZE", "MV2_COMM_WORLD_SIZE"], 1)
+                local_rank = env2int(["LOCAL_RANK", "MPI_LOCALRANKID", "OMPI_COMM_WORLD_LOCAL_RANK", "MV2_COMM_WORLD_LOCAL_RANK"], 0)
+                local_size = env2int(["MPI_LOCALNRANKS", "OMPI_COMM_WORLD_LOCAL_SIZE", "MV2_COMM_WORLD_LOCAL_SIZE"], 1)
+                self.local_process_index = local_rank
+                os.environ["RANK"] = str(rank)
+                os.environ["WORLD_SIZE"] = str(size)
+                if not os.environ.get("MASTER_PORT", None): os.environ["MASTER_PORT"] = "29500"
```
sgugger, line 171:
```suggestion
                if not os.environ.get("MASTER_PORT", None):
                    os.environ["MASTER_PORT"] = "29500"
```

src/accelerate/state.py
```diff
@@ -123,12 +146,43 @@ def __init__(self, fp16: bool = None, cpu: bool = False, _from_accelerator: bool
...
+                if not os.environ.get("MASTER_ADDR", None):
+                    if local_size != size and backend != "mpi":
+                        print("Warning: Looks like distributed multinode run but MASTER_ADDR env not set, using '127.0.0.1' as default")
+                        print("If this run hangs, try exporting rank 0's hostname as MASTER_ADDR")
```
sgugger, line 175: I thin we should raise a ValueError here telling the user to pass a `MASTER_ADDR` instead of relying on a default.

src/accelerate/state.py
```diff
@@ -123,12 +146,43 @@ def __init__(self, fp16: bool = None, cpu: bool = False, _from_accelerator: bool
...
+                        os.environ["MASTER_ADDR"] = "127.0.0.1"
+                if not torch.distributed.is_initialized():
+                    torch.distributed.init_process_group(backend, rank=rank, world_size=size)
+                    self.backend = backend
+                self.num_processes = torch.distributed.get_world_size()
+                self.process_index = torch.distributed.get_rank()
+                self.local_process_index = local_rank
+                self.local_num_processes = local_size
```
sgugger, line 183: This one is not used anywhere, so I would remove it.

src/accelerate/state.py
```diff
@@ -139,7 +193,7 @@ def __init__(self, fp16: bool = None, cpu: bool = False, _from_accelerator: bool
 
     def __repr__(self):
         return (
-            f"Distributed environment: {self.distributed_type}\n"
+            f"Distributed environment: {self.distributed_type}  Backend: {self.backend}\n"
```
sgugger, line 196: We should only display the backend if it's not None.

sgugger (commented):
Thanks a lot for digging into this! It looks great and with a little bit of polishing (mostly naming), it should be ready to be merged soon!

You should add a line in the main README in the list at the end with `- multinode CPU`! Also it looks like your suggested integration supports different types of launchers (from the number of env variables looked at). Could you document this somewhere?

src/accelerate/commands/config/cluster.py
ddkalamk: Well, actually we can have multi-cpu on a single machine. The typical use case is one process (or rank as it is called in MPI world) per socket. So, on a dual socket machine, we can have two ranks per node. So, I suggest keeping it as multi-CPU.

src/accelerate/state.py
ddkalamk: Some MPI-like custom backends can initialize without MASTER_ADDR so I just put a warning. I can change to ValueError if you think that's a better choice.

ddkalamk:
I have made the suggested changes. Please take a look.
Thanks.

src/accelerate/state.py
sgugger: We are in a test where `backend != "mpi"`, so I would adapt the error message (to say the backend needs a MASTER_ADDR) and raise an error. For MPI, it will still work without the MASTER_ADDR.

ddkalamk: Made the change.

README.md
```diff
@@ -151,6 +151,15 @@ For instance, here is how you would run the GLUE example on the MRPC task (from
 accelerate launch examples/nlp_example.py
 ```
 
+## Launching multi-CPU run using MPI
+
+🤗 Here is another way to launch multi-CPU run using MPI
+On your cluster just run:
+
+```bash
+mpirun -np 2 python examples/nlp_example.py
```
sgugger, line 160: Does this require an install of some library? Let's add how to!

src/accelerate/state.py
```diff
@@ -27,6 +35,19 @@
     _tpu_available = False
 
 
+def get_int_from_env(env_keys, default):
+    """Returns the first positive env found in the `env_keys` list or the default."""
```
sgugger, line 39:
```suggestion
    """Returns the first positive env value found in the `env_keys` list or the default."""
```

sgugger (approved):
Thanks for all the adjustments, two more little things and we should be good to merge!

README.md
ddkalamk: There are multiple options e.g. [Intel MPI](https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/mpi-library.html#gs.0hn0hn), [OpenMPI](https://www.open-mpi.org/) or MVAPICH. I assume, typical HPC clusters would have some version available or users can install them in userspace easily. example instructions for Open-MPI are [here](https://www.open-mpi.org/faq/?category=building#easy-build).

ddkalamk: Ok, added a line on how to get OpenMPI...

sgugger:
Last step, could you run `make style` on your branch? This should take care of the failing test.
Let me know if run into any trouble and I can push it on the PR instead.

ddkalamk:
Sorry, my system doesn't have black. Make gave error.

```
black tests src examples
make: black: Command not found
make: *** [style] Error 127
```
Can you please go ahead and push it to PR?
Thanks.

ddkalamk:
Just installed black and fixed formatting... It is still giving an error about torch_ccl import but I don't know how to fix it...

src/accelerate/state.py
```diff
@@ -19,6 +19,14 @@
 import torch
 
 
+try:
+    import torch_ccl as ccl
```
sgugger, line 23:
```suggestion
    import torch_ccl
```
I think this should take care of the problem

ddkalamk:
Fixed issue with flake8 and squashed to single commit... Should be good to merge now.

sgugger:
Thanks again for all your work on this!

PR #67, opened 2021-05-05
Honor namedtuples in inputs/outputs

### LINKED ISSUE #65: Feature request: Add support for NamedTuples in dataloaders
In order to produce self-documenting code, our team has the habit of using NamedTuples instead of plain tuples as the return type of our datasets.

Since they are sub-classes of tuples, they are handled by every PyTorch mechanism we've encountered as if they are tuple, so it all works smoothly.

When using Accelerate, I get an error in `send_to_device` at line 114. `type(tensor)(send_to_device(t, device) for t in tensor)` raises a type error because `type(tensor)` returns my NamedTuple, which cannot use the generator `(send_to_device(t, device) for t in tensor)` as argument. We would need some way to transform the generator into a type that is accepted by the NamedTuple.

sgugger:
This PR adds support for namedtuples in inputs/outputs of models and dataloaders.

There is sadly no direct check to know if a Python object is a namedtuple, and nametuple can't be instantiated from a generator like regular tuples and lists, the generator needs to be unpacked. Sadly list and tuples don't want that kind of unpacking, so I had to add a utility function that does the distinction depending on the type of object received.

Fixes #65

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,53 @@
+# Copyright 2021 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+import torch
+
+from accelerate.utils import send_to_device
+
+
+TestNamedTuple = namedtuple("TestNamedTuple", "a b")
+
+
+class UtilsTester(unittest.TestCase):
```
LysandreJik, line 26: Great tests :)

LysandreJik (approved):
Great, LGTM! Sorry for the delayed review.

PR #81, opened 2021-05-13
Use `optimizer` for consistency

kumapo:
it would be nice if names of `torch.optim.Adam` instances in README.md are consistent.


sgugger (approved):
Yes, we had used `optim` when we displayed the diff in another way and needed shorter lines, but this is not the case anymore.

Thanks for the fix!

PR #82, opened 2021-05-13
Add DeepSpeed support

vasudevgupta7:
This PR will add DeepSpeed support to Accelerate.

User's code should look like this:

```diff
from accelerate import Accelerator

model: torch.nn.Module
optimizer: torch.optim.Optimizer
tr_data: torch.utils.data.DataLoader
eval_data: torch.utils.data.DataLoader

gradient_accumulation_steps: int
epochs: int

accelerator = Accelerator(fp16=True)
model, optimizer, tr_data, eval_data = accelerator.prepare(model, optimizer, tr_data, eval_data)

# training loop
for e in range(epochs):
    for step, batch in enumerate(tr_data):
        output = model(**batch)
        loss = output.loss / gradient_accumulation_steps
        accelerator.backward(loss)
        if (step + 1) % gradient_accumulation_steps == 0 or step == len(train_data) - 1:
            optimizer.step()
            optimizer.zero_grad()
-           lr_scheduler.step()
+           if not optimizer.is_overflow:
+               lr_scheduler.step()

# distributed evaluation
model.eval()
for batch in eval_data:
    output = model(**batch)
    predictions = accelerator.gather(output)
    labels = accelerator.gather(batch["labels"])

# for saving your model
accelerator.wait_for_everyone()
unwrapped_model = accelerator.unwrap_model(model)

# if its a HuggingFace Transformer
- unwrapped_model.save_pretrained(args.output_dir, save_function=accelerator.save)
+ unwrapped_model.save_pretrained(args.output_dir, state_dict=accelerator.get_state_dict(model), save_function=accelerator.save)
```

**Running the script**

```shell
accelerate config
accelerate launch <training_script.py> <training_args>
```

@sgugger


src/accelerate/accelerator.py
```diff
@@ -86,18 +85,10 @@ def __init__(
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
+        if deepspeed_plugin is not None:
+            assert isinstance(deepspeed_plugin, DeepSpeedPlugin), "deepspeed_plugin must be instance of DeepSpeedPlugin"
```
sgugger, line 89:
```suggestion
            assert isinstance(deepspeed_plugin, DeepSpeedPlugin), "`deepspeed_plugin` must be a DeepSpeedPlugin object."
```

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
             model.forward = torch.cuda.amp.autocast()(model.forward)
         return model
 
+    def _prepare_deepspeed(self, *args):
+
+        ds_plugin = self.state.deepspeed_plugin
+        self.ds_config = ds_plugin.ds_config
+
+        batch_size = [obj.batch_size for obj in args if hasattr(obj, "batch_size")]
+        assert len(batch_size) > 0, "You must specify training_dataloader in `accelerate.prepare()` when using DeepSpeed"
```
sgugger, line 254:
```suggestion
        assert len(batch_size) > 0, "You must specify a training or evaluation dataloader in `accelerate.prepare()` when using DeepSpeed."
```

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
             model.forward = torch.cuda.amp.autocast()(model.forward)
         return model
 
+    def _prepare_deepspeed(self, *args):
+
+        ds_plugin = self.state.deepspeed_plugin
+        self.ds_config = ds_plugin.ds_config
+
+        batch_size = [obj.batch_size for obj in args if hasattr(obj, "batch_size")]
```
sgugger, line 253: Rename to batch_sizes (since there are several of them)?

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
...
+        assert len(batch_size) > 0, "You must specify training_dataloader in `accelerate.prepare()` when using DeepSpeed"
+        logger.info("Since you passed both train & eval dataloader, `is_train_batch_min` will decide the `train_batch_size`")
```
sgugger, line 255:
```suggestion
        logger.info("Since you passed several dataloaders, `is_train_batch_min` will determine the `train_batch_size`".)
```

+ I think we should check the list has more than one element before issuing this info!

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
...
+        batch_size = min(batch_size) if ds_plugin.is_train_batch_min else max(batch_size)
+
+        self.ds_config["train_batch_size"] = batch_size * ds_plugin.gradient_accumulation_steps * self.num_processes
+
```
sgugger, line 259: Nothing needs to be done for the `eval_batch_size`?

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
...
+            if isinstance(obj, torch.nn.Module):
+                model = obj
+                model_parameters = filter(lambda p: p.requires_grad, model.parameters())
+            elif isinstance(obj, (torch.optim.Optimizer, dict)):
+                optimizer = obj
+
+        # TODO: how dict based optimizer should be passed (in optimizer or ds_plugin) ??
+        # if isinstance(optimizer, dict):
+        #     self.ds_config.update({"optimizer": optimizer})
+        #     optimizer = None
```
sgugger, line 275: I don't understand this part, could you explain?

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
...
+            assert isinstance(model, deepspeed.DeepSpeedEngine), "PipelineEngine not supported currently"
+            for i in range(len(result)):
+                if isinstance(result[i], torch.nn.Module):
+                    result[i] = DeepSpeedEngineWrapper(model)
+                elif isinstance(result[i], torch.optim.Optimizer):
+                    result[i] = DeepSpeedOptimizerWrapper(optimizer, model)
+            self.model = model # pointing for model.backward()
+        
+        if self.distributed_type == DistributedType.DEEPSPEED:
+            assert hasattr(self, "model"), "Accelerator instance must have model as its attribute"
```
sgugger, line 289: The error message should be more along the lines of "You need to pass the model along the optimizer when using Deepspeed."

src/accelerate/accelerator.py
```diff
@@ -230,6 +245,51 @@ def prepare_model(self, model):
...
+
+        # TODO: how dict based optimizer should be passed (in optimizer or ds_plugin) ??
+        # if isinstance(optimizer, dict):
+        #     self.ds_config.update({"optimizer": optimizer})
+        #     optimizer = None
+
+        # useful when only eval_dataloader is given into `accelerator.prepare()`
+        if model is not None:
+            model, optimizer, _, _ = deepspeed.initialize(model=model, optimizer=optimizer, config_params=self.ds_config, dist_init_required=False, model_parameters=model_parameters)
+            assert isinstance(model, deepspeed.DeepSpeedEngine), "PipelineEngine not supported currently"
```
sgugger, line 280: I'm not sure I understand this error message.

src/accelerate/accelerator.py
```diff
@@ -248,7 +308,9 @@ def backward(self, loss):
         """
         Use :obj:`accelerator.backward(loss)` in lieu of :obj:`loss.backward()`.
         """
-        if self.scaler is not None:
+        if self.distributed_type == DistributedType.DEEPSPEED:
+            self.model.backward(loss)
```
sgugger, line 312: Great to have this handled here :-)

src/accelerate/commands/config/cluster.py
```diff
@@ -16,22 +16,22 @@
...
 from .config_utils import _ask_field, _convert_distributed_mode, _convert_yes_no_to_bool
 
 
 def get_cluster_input():
     distributed_type = _ask_field(
-        "Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): ",
+        "Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] DeepSpeed [4] TPU): ",
         _convert_distributed_mode,
-        error_message="Please enter 0, 1, 2 or 3.",
+        error_message="Please enter 0, 1, 2, 3 or 4.",
```
sgugger, line 27: I think deepspeed is a subcase of multi-CPU/multi-GPU no? I would rather ask the question after.

src/accelerate/commands/config/cluster.py
```diff
@@ -50,6 +50,25 @@ def get_cluster_input():
...
+                lambda x: int(x),
+                default=2,
+            )
+            gradient_accumulation_steps = _ask_field(
+                "What should be your number of gradient accumulation steps? [1]: ",
+                lambda x: int(x),
+                default=1,
+            )
+            is_train_batch_min = _ask_field(
+                "Should accelerate choose minimum batch size incase both train & eval dataloaders are passed? [True]: ",
```
sgugger, line 67: I think this one should just be left at True: there are going to be tons of parameters for deepspeed so let's keep the essential ones in the question (and use the default for the others). Users can edit the config after!

src/accelerate/commands/config/config_args.py
```diff
@@ -111,13 +111,25 @@ def __post_init__(self):
...
+    num_processes: int = None
     machine_rank: int = 0
     num_machines: int = 1
     main_process_ip: Optional[str] = None
     main_process_port: Optional[int] = None
     main_training_function: str = "main"
 
+    def __post_init__(self):
+        super().__post_init__()
+        assert self.num_processes is not None, "num_processes can't be None"
```
sgugger, line 123: Why add this default?

src/accelerate/commands/launch.py
```diff
@@ -151,6 +151,44 @@ def multi_gpu_launcher(args):
...
+        cmd.extend(["--num_gpus", str(args.num_processes)])
+
+    cmd.append(args.training_script)
+    cmd.extend(args.training_script_args)
+
+    current_env = os.environ.copy()
+    current_env["USE_FP16"] = str(args.fp16)
+    current_env["ZERO_STAGE"] = int(args.zero_stage)
+    current_env["GRADIENT_ACCUMULATION_STEPS"] = int(args.gradient_accumulation_steps)
+    current_env["IS_TRAIN_BATCH_MIN"] = bool(args.is_train_batch_min)
```
sgugger, line 183: Are those all expected by deepspeed?

src/accelerate/commands/launch.py
```diff
@@ -276,14 +314,15 @@ def sagemaker_launcher(sagemaker_config: SageMakerConfig, args):
 
 def launch_command(args):
     # Sanity checks
-    if args.multi_gpu and args.tpu:
+    if args.multi_gpu and args.tpu: # TODO: fix this for deepspeed
```
sgugger, line 317: There is something done in Transformers for that, if you need help.

src/accelerate/deepspeed_utils.py
```diff
@@ -0,0 +1,177 @@
...
+from dataclasses import dataclass, field
+
+from .state import is_apex_available
+
+@dataclass
+class DeepSpeedPlugin:
+
+    gradient_accumulation_steps: int = field(default=1, metadata={"help": "Number of steps to accumulate gradients before updating optimizer states"})
+    zero_stage: int = field(default=0, metadata={"help": "Possible options are 0,1,2"})
+    is_train_batch_min: str = field(default=True, metadata={"help": "If both train & eval dataloaders are specified, this will decide the train_batch_size"})
```
sgugger, line 26: Make sure to use the `make style` command, this is not super readable here :-)

src/accelerate/deepspeed_utils.py
```diff
@@ -0,0 +1,177 @@
...
+        self.ds_config = {
+            "train_batch_size": None,
+            "fp16": {"enabled": False if self.fp16 is None else self.fp16},
+            "gradient_accumulation_steps": self.gradient_accumulation_steps,
+            "zero_optimization": {"stage": self.zero_stage},
+            "steps_per_print": float("inf"), # this will stop deepspeed from logging @ stdout
+        }
+
+
+class DeepSpeedEngineWrapper(nn.Module):
```
sgugger, line 40: This is copied from `deepspeed`? Could we subclass their class then and only change the method that requires it?

src/accelerate/deepspeed_utils.py
```diff
@@ -0,0 +1,177 @@
...
+        self.model.micro_steps += 1
+
+    def save_checkpoint(self, *args, **kwargs):
+        return self.model.save_checkpoint(*args, **kwargs)
+
+    def load_checkpoint(self, *args, **kwargs):
+        return self.model.load_checkpoint(*args, **kwargs)
+
+
+class DeepSpeedOptimizerWrapper(torch.optim.Optimizer):
```
sgugger, line 122: Should this subclass the `AcceleratedOptimizer` for most properties and only overrides the necessary ones?

src/accelerate/optimizer.py
```diff
@@ -113,3 +113,8 @@ def step(self):
     def _switch_parameters(self, parameters_map):
         for param_group in self.optimizer.param_groups:
             param_group["params"] = [parameters_map.get(p, p) for p in param_group["params"]]
+
+    @property
+    def is_overflow(self):
+        """ This needs to be implemented at the end """
+        return False # TODO: implement it
```
sgugger, line 120: I can do this next week!

src/accelerate/state.py
```diff
@@ -34,6 +35,20 @@
...
+except:
+    _deepspeed_available = False
+
+try:
+    from apex import amp
+
+    _apex_available = True
+except ImportError:
+    _apex_available = False
+
```
sgugger, line 51: Apex is not supported yet. Where do you need it?

src/accelerate/state.py
```diff
@@ -34,6 +35,20 @@
 except ImportError:
     _tpu_available = False
 
+try:
+    import deepspeed
+
+    _deepspeed_available = True
+except:
+    _deepspeed_available = False
```
sgugger, line 43: I would prefer not importing directly (see [here](https://github.com/huggingface/accelerate/blob/a96fbaaf19e650da101bfadb886ecc1d1b1cb205/src/accelerate/utils.py#L31) for an example on how to this).

sgugger (commented):
Thanks a lot for your work on this! This is great! I left a couple of comments to polish the implementation.

src/accelerate/deepspeed_utils.py
vasudevgupta7: Yes, we can do that I think. Then  we won't need `deepspeed.initialize()` since deepspeed.initialize() is doing just sending everything to `DeepSpeedEngine` (see [this](https://github.com/microsoft/DeepSpeed/blob/46f4573b1a8a9cda2b45f4de4e90b631cee3f80b/deepspeed/__init__.py#L119)).

src/accelerate/commands/launch.py
vasudevgupta7: yes in `deepspeed.initialize()` as `config_params`

src/accelerate/accelerator.py
vasudevgupta7: `deepspeed.initialize()` passes arguments to either of 2 class -  `DeepSpeedEngine` or  `PipelineEngine`. & I had to make changes to DeepSpeedEngine for preventing automatic gradient_accumulation inside deepspeed (in `DeepSpeedEngineWrapper`) & hence only `DeepSpeedEngine` is supported right now. But this line can be removed if we directly inherit `DeepSpeedEngineWrapper` from `DeepSpeedEngine` directly (as pointed out by you in a later comment).

src/accelerate/accelerator.py
vasudevgupta7: DeepSpeed can build optimizer automatically for us if optimizer key is included in `ds_config`, instead of giving `torch.optim.Optimizer` in `deepspeed.initialize()`. Passing optimizer dict is the recommended way by deepspeed since they have some fused implementation & it's more easy to switch their optimizers. So we need to include that somehow.

So one way is to let user pass optimizer as dict in accelerate.prepare() itself & later we can handle that; while other can be to include it in deepspeed_plugin. What way you would suggest??

src/accelerate/state.py
vasudevgupta7: its needed in DeepSpeedEngineWrapper.

src/accelerate/commands/config/config_args.py
vasudevgupta7: I need to fix it a bit. Will work on it.

src/accelerate/accelerator.py
vasudevgupta7: no

src/accelerate/accelerator.py
sgugger: Let's focus on this on another PR. We could have accelerate support this in general (since the goal is to have the same API for everything) but it would require more work.

src/accelerate/state.py
sgugger: Ok, then please use the same as below and not a direct import.

src/accelerate/accelerator.py
vasudevgupta7: okay. another PR after this then :)

sgugger:
Note for before merging this: the PR should be rebased on master and the reference to the model added in the `Accelerator` should be cleaned up in the method introduced in the PR above.

src/accelerate/accelerator.py
```diff
@@ -231,6 +255,70 @@ def prepare_model(self, model):
             model.forward = torch.cuda.amp.autocast()(model.forward)
         return model
 
+    def _prepare_deepspeed(self, *args):
+
+        ds_plugin = self.state.deepspeed_plugin
+        self.ds_config = ds_plugin.ds_config
+
+        bz = [obj.batch_size for obj in args if hasattr(obj, "batch_size")]
```
sgugger, line 263: Can we use `batch_sizes` here?

src/accelerate/accelerator.py
```diff
@@ -231,6 +255,70 @@ def prepare_model(self, model):
...
+        assert (
+            len(bz) > 0
+        ), "You must specify a training or evaluation dataloader in `accelerate.prepare()` when using DeepSpeed."
```
sgugger, line 266: It think this one would read better as (including the suggested rename form before):
```suggestion
        if len(batch_sizes) == 0:
            raise ValueError("You must specify a training or evaluation dataloader in `accelerate.prepare()` when using DeepSpeed.")
```

src/accelerate/accelerator.py
```diff
@@ -231,6 +255,70 @@ def prepare_model(self, model):
...
+        if len(bz) > 1:
+            logger.info(
+                "Since you passed both train & eval dataloader, `is_train_batch_min` will decide the `train_batch_size`"
+            )
+        batch_size_per_device = min(bz) if ds_plugin.is_train_batch_min else max(bz)
```
sgugger, line 271: Since we are computing it just after, can we log it? the log should be something like:
```
f"Since you passed both train and evaluation dataloader, `is_train_batch_min` (here {ds_plugin.is_train_batch_min} will decide the `train_batch_size` ({batch_size_per_device})."
```
This might need to be split in two lines as well to respect the 119 char limit.

src/accelerate/accelerator.py
```diff
@@ -231,6 +255,70 @@ def prepare_model(self, model):
...
+
+        result = [self._prepare_one(obj) if isinstance(obj, torch.utils.data.DataLoader) else obj for obj in args]
+
+        model = None
+        optimizer = None
+        for obj in result:
+            if isinstance(obj, torch.nn.Module):
+                model = obj
+            elif isinstance(obj, (torch.optim.Optimizer, dict)):
+                optimizer = obj
```
sgugger, line 285: There can be several models or several optimizers passed, so we should make a list of models and optimizers for that case here? If deepspeed only supports one, we should raise an error here.

src/accelerate/accelerator.py
```diff
@@ -357,6 +447,7 @@ def free_memory(self):
         method between two trainings with different models/optimizers.
         """
         self._optimizers = []
+        self.deepspeed_engine = None
```
sgugger, line 450: Is this the only reference we create? It seems so but I want to be certain.

src/accelerate/commands/config/cluster.py
```diff
@@ -50,6 +50,39 @@ def get_cluster_input():
...
+            default=False,
+            error_message="Please enter yes or no.",
+        )
+        if use_deepspeed:
+            distributed_type = DistributedType.DEEPSPEED
+
+        ds_config = {}
+        if distributed_type == DistributedType.DEEPSPEED:
+            ds_config["zero_stage"] = _ask_field(
+                "What should be your DeepSpeed's ZeRO optimization stage (0, 1, 2)? [0]: ",
```
sgugger, line 68:
```suggestion
                "What should be your DeepSpeed's ZeRO optimization stage (0, 1, 2)? [0]: ",
```
Should include 3?

src/accelerate/commands/launch.py
```diff
@@ -276,19 +338,26 @@ def sagemaker_launcher(sagemaker_config: SageMakerConfig, args):
...
         defaults = load_config_from_file(args.config_file)
-        if not args.multi_gpu and not args.tpu:
+        if not args.multi_gpu and not args.tpu and not args.use_deepspeed:
+            args.use_deepspeed = defaults.distributed_type == DistributedType.DEEPSPEED
             args.multi_gpu = defaults.distributed_type == DistributedType.MULTI_GPU
             args.tpu = defaults.distributed_type == DistributedType.TPU
         if defaults.compute_environment == ComputeEnvironment.LOCAL_MACHINE:
             # Update args with the defaults
             for name, attr in defaults.__dict__.items():
+                if name == "ds_config":
```
sgugger, line 355: Should this be a more general check of whether `attr` is a dict?

src/run_glue_no_trainer.py
```diff
@@ -0,0 +1,461 @@
+# coding=utf-8
```
sgugger, line 1: This file should be removed afterward :-)

src/accelerate/accelerator.py
```diff
@@ -231,6 +255,70 @@ def prepare_model(self, model):
             model.forward = torch.cuda.amp.autocast()(model.forward)
         return model
 
+    def _prepare_deepspeed(self, *args):
+
+        ds_plugin = self.state.deepspeed_plugin
+        self.ds_config = ds_plugin.ds_config
```
sgugger, line 261: Can we say `self.deepspeed_config` to go with `self.state.deepspeed_plugin`?

Also would love if this function was using `deepspeed_plugin` instead of `ds_pluging`.

sgugger (approved):
Thanks a lot for all your amazing work! I think we are in good shape to have this merge, I just left some last comments around naming.

src/run_glue_no_trainer.py
vasudevgupta7: will remove.

src/accelerate/accelerator.py
vasudevgupta7: I didn't understand the question. Could you please clarify??

I will rename `ds_plugin` to `deepspeed_plugin`.

sgugger: I would prefer we use `deepspeed_plugin` and `deepspeed_config` instead of `ds_plugin` and `ds_config`

src/accelerate/commands/launch.py
sgugger: Yes, that's why I'm suggesting we check `if isinstance(attr, dict)` instead of the name.

src/accelerate/accelerator.py
vasudevgupta7: yes

src/accelerate/accelerator.py
```diff
@@ -74,10 +84,18 @@ def __init__(
...
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
-        self.state = AcceleratorState(fp16=fp16, cpu=cpu, _from_accelerator=True)
+        if deepspeed_plugin is None:
+            deepspeed_plugin = DeepSpeedPlugin()  # init from env variables
+        else:
+            assert isinstance(
+                deepspeed_plugin, DeepSpeedPlugin
+            ), "`deepspeed_plugin` must be a DeepSpeedPlugin object."
```
LysandreJik, line 96: Do we really need a `DeepSpeedPlugin` to be initialized if deepspeed isn't available?

src/accelerate/commands/config/cluster.py
```diff
@@ -50,6 +50,39 @@ def get_cluster_input():
                 "What is the port you will use to communicate with the main process? ",
                 lambda x: int(x),
             )
+
+    deepspeed_config = None
+    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO]:
+        use_deepspeed = _ask_field(
+            "Do you want to use DeepSpeed (yes/no)? [NO]",
```
LysandreJik, line 57:
```suggestion
            "Do you want to use DeepSpeed (yes/no)? [no]",
```

LysandreJik (commented):
Cool, great job implementing that @vasudevgupta7!

Two items that should be handled in my opinion:
- It seems that `accelerate` will not run once this PR is merged if `deepspeed` isn't in the environment. Is this correct?
- There really should be some tests to ensure that everything is working correctly. I would advocate for a test suite that runs accelerate with and without deepspeed, but I know the infra to support that isn't there yet.

src/accelerate/deepspeed_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        default=None, metadata={"help": "Number of steps to accumulate gradients before updating optimizer states"}
+    )
+    zero_stage: int = field(
+        default=None,
+        metadata={"help": "Possible options are 0,1,2,3; Default will be taken from environment variable"},
+    )
+    is_train_batch_min: str = field(
+        default=True,
+        metadata={"help": "If both train & eval dataloaders are specified, this will decide the train_batch_size"},
+    )
```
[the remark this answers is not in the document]
vasudevgupta7: DeepSpeed needs to know the batch size of the data even if we build dataloader by ourself. & we are extracting batch size from dataloader internally. But whenever 2 dataloaders are passed in `accelerator.prepare()` (i.e. both train & eval), we can't directly take the batch size from `train_dataloader`, since internally we don't know which is train & which is  eval (because of the way accelerator.prepare() takes in args). So this argument will help users to decide what is train batch size & what is eval.

src/accelerate/deepspeed_utils.py
```diff
@@ -0,0 +1,150 @@
...
+
+    auto_opt_mapping: bool = field(
+        default=True,
+        metadata={"help": "whether to map torch.adam to deepspeed optimizer version of adam based on config"},
+    )
+
+    offload_optimizer_device: bool = field(default=None, metadata={"help": "Possible options are none|cpu|nvme"})
+
+    fp16: bool = field(repr=False, default=None, metadata={"help": "You need not define this here"})
```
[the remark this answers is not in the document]
vasudevgupta7: I will try to find other solution.

src/accelerate/commands/config/cluster.py
hamishdickson: I think you probably want a space at the end of this to be consistent with the other _ask_fields

sgugger: Should actually be 
```
          "Do you want to use DeepSpeed? [yes/NO]: ",
```
to be compatible with what is actually in the lib for now.

src/accelerate/commands/config/cluster.py
```diff
@@ -50,6 +50,39 @@ def get_cluster_input():
...
+        if distributed_type == DistributedType.DEEPSPEED:
+            deepspeed_config["zero_stage"] = _ask_field(
+                "What should be your DeepSpeed's ZeRO optimization stage (0, 1, 2, 3)? [2]: ",
+                lambda x: int(x),
+                default=2,
+            )
+
+            if deepspeed_config["zero_stage"] >= 2:
+                deepspeed_config["offload_optimizer_device"] = _ask_field(
+                    "Where to offload optimizer states (none|cpu|nvme)? [none]: ",
```
sgugger, line 75:
```suggestion
                    "Where to offload optimizer states? [NONE/cpu/nvme]: ",
```

README.md
```diff
@@ -160,6 +160,26 @@ Once you have MPI setup on your cluster, just run:
 mpirun -np 2 python examples/nlp_example.py
 ```
 
+## Launching training using DeepSpeed
+
+🤗 Accelerate supports training on single/multiple GPUs using DeepSpeed. For using deepspeed, you don't need to change anything. You can set everything using just `accelerate config`. However, if you desire to tweak your deepspeed related args from your python script, we provide you the `DeepSpeedPlugin`.
```
sgugger, line 165:
```suggestion
🤗 Accelerate supports training on single/multiple GPUs using DeepSpeed. to use it, you don't need to change anything in your training code; you can set everything using just `accelerate config`. However, if you desire to tweak your DeepSpeed related args from your python script, we provide you the `DeepSpeedPlugin`.
```

README.md
```diff
@@ -160,6 +160,26 @@ Once you have MPI setup on your cluster, just run:
...
+
+```python
+from accelerator import Accelerator, DeepSpeedPlugin
+
+# deepspeed needs to know your gradient accumulation steps before hand, so don't forget to pass it
+# Remember you need to do gradient accumulation by yourself, just like you would have done without deepspeed
```
sgugger, line 171:
```suggestion
# Remember you still need to do gradient accumulation by yourself, just like you would have done without deepspeed
```

README.md
```diff
@@ -160,6 +160,26 @@ Once you have MPI setup on your cluster, just run:
...
+deepspeed_plugin = DeepSpeedPlugin(zero_stage=2, gradient_accumulation_steps=2)
+accelerator = Accelerator(fp16=True, deepspeed_plugin=deepspeed_plugin)
+
+# How to save your 🤗 Transformer?
+accelerator.wait_for_everyone()
+unwrapped_model = accelerator.unwrap_model(model)
+unwrapped_model.save_pretrained(save_dir, save_function=accelerator.save, state_dict=accelerator.get_state_dict(model))
+```
+
+Note: DeepSpeed support is experimental for now. In case you get into some issue, please raise an issue.
```
sgugger, line 181:
```suggestion
Note: DeepSpeed support is experimental for now. In case you get into some problem, please open an issue.
```

sgugger (approved):
Just a few last tweaks to the doc and this should be good to merge!

PR #88, opened 2021-05-17
Add `unscale_gradients` method.

sgugger:
This PR adds a new method to the `Accelerator` class to unscale the gradients, which is necessary before doing some operations like gradient clipping (also supported in its own method currently, but some users might have more exotic strategies).


LysandreJik (approved):
Cool, LGTM!

PR #89, opened 2021-05-17
Add Accelerator.free_memory

### LINKED ISSUE #84: Add warning not to reuse `Accelerator`
I was training multiple models, and there seems to be some GPU memory leak.
I think it's because `Accelerator` appends reference to optimizer, which in turn reference some GPU memory.
https://github.com/huggingface/accelerate/blob/df260fa71af9c23587f6c0a86df5ad17f69822d2/src/accelerate/accelerator.py#L164

When reusing `Accelerator` to train multiple models, the GPU memory used by old optimizer is not released.

sgugger:
This PR adds a `free_memory` to the `Accelerator` class, to free memory between two different trainings. This cleans up the internal list to the optimizers, but we will also add other objects stored in the future (for instance #82 adds a reference to the model for deespeed).

Fixes #84

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
 from typing import List, Optional, Union
 
 import torch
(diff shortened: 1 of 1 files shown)


LysandreJik (approved):
Cool that's clean!

PR #104, opened 2021-06-14
Pass along kwargs to backward

### LINKED ISSUE #103: Accelerator backward with retain graph
I want to use `loss.backward(retain_graph=True)` instead of normal backward. How can I do this using accelerate?

sgugger:
This PR allows the use of all keyword arguments for the `backward` method.

Fixes #103


LysandreJik (approved):
Great, LGTM!

PR #105, opened 2021-06-15
added closure argument to optimizer.step()

pmelchior:
Optimizers generally support a `closure` argument. This PR adds the argument to `AcceleratedOptimizer` and passes it on. Tested on CPU and GPU, but not on TPU.

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -102,14 +102,14 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
                 set_to_none = False
             self.optimizer.zero_grad(set_to_none=set_to_none)
 
-    def step(self):
+    def step(self, closure=None):
         if self.state.distributed_type == DistributedType.TPU:
-            xm.optimizer_step(self.optimizer)
+            xm.optimizer_step(self.optimizer, dict(closure=closure))
         elif self.scaler is not None:
-            self.scaler.step(self.optimizer)
+            self.scaler.step(self.optimizer, closure)
             self.scaler.update()
         else:
-            self.optimizer.step()
+            self.optimizer.step(closure)
 
     def _switch_parameters(self, parameters_map):
         for param_group in self.optimizer.param_groups:


src/accelerate/optimizer.py
```diff
@@ -102,14 +102,14 @@ def zero_grad(self, set_to_none=None):
                 set_to_none = False
             self.optimizer.zero_grad(set_to_none=set_to_none)
 
-    def step(self):
+    def step(self, closure=None):
         if self.state.distributed_type == DistributedType.TPU:
-            xm.optimizer_step(self.optimizer)
+            xm.optimizer_step(self.optimizer, dict(closure=closure))
```
sgugger, line 107:
```suggestion
            xm.optimizer_step(self.optimizer, optimizer_args={"closure": closure})
```
The [PyTorch XLA doc](https://pytorch.org/xla/release/1.8/index.html#torch_xla.core.xla_model.optimizer_step) seems to suggest we have to use `optimizer_args` to pass this along and it's not the second argument.

sgugger (approved):
Thanks for adding this!

src/accelerate/optimizer.py
pmelchior: good catch!

src/accelerate/optimizer.py
```diff
@@ -102,14 +102,14 @@ def zero_grad(self, set_to_none=None):
                 set_to_none = False
             self.optimizer.zero_grad(set_to_none=set_to_none)
 
-    def step(self):
+    def step(self, closure=None):
         if self.state.distributed_type == DistributedType.TPU:
-            xm.optimizer_step(self.optimizer)
+            xm.optimizer_step(self.optimizer, optimizer_args={"closure": closure})
```
sgugger, line 107: Actually thinking a little bit more about it, let's not pass anything if there is no closure:
```suggestion
            optimizer_args = {"closure": closure} if closure is not None else {}
            xm.optimizer_step(self.optimizer, optimizer_args=optimizer_args)
```

pmelchior: I concur, this is a little more explicit, but it shouldn't matter (famous last words...) because the base `torch.optim.Optimizer` has a `closure` argument. For the other options, we pass that along anyway.

sgugger: If you don't mind committing the suggestion then, we're all good to merge :-)

PR #106, opened 2021-06-15
Fix nlp_example

jjiang2cal:
loss.backward does gradient all_reduce in pytorch DDP. In gradient
accumulation, we do not want all_reduce in every step, so use
no_sync in DDP to avoid unnecessary gradient sync.

### DIFF
--- a/examples/nlp_example.py
+++ b/examples/nlp_example.py
@@ -137,8 +137,15 @@ def training_function(config, args):
             outputs = model(**batch)
             loss = outputs.loss
             loss = loss / gradient_accumulation_steps
-            accelerator.backward(loss)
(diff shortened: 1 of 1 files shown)


sgugger:
That's completely true, but the goal of the example is to be as simple as possible and it does work with computing the gradients at each step, so I'm a bit torn on whether we should have this or not.

What do you think @LysandreJik @thomwolf ?

lvwerra:
If not in the example it would be great if there would a hint/warning in the documentation somewhere. We made that mistake in one of our training script because of just copy-pasting the example.

sgugger:
Accelerate will soon handle gradient accumulation directly, specifically for this problem :-)

PR #108, opened 2021-06-16
Fix import error for torch 1.4.0

sgugger:
The `start_processes` function introduced with the `notebook_launcher` requires torch 1.5.0, this PR moves the import to avoid having an import error.


LysandreJik (approved):
Great catch, LGTM

PR #115, opened 2021-06-28
Unwrap optimizer before unscaling

### LINKED ISSUE #113: Unscaling Wrapped optimizer making the gradient is unscaled twice
Hi, thank you for the useful tool. 

https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L416-L417

I'm referring to these lines of code where the wrapped optimizer is unscaled when gradient clipping. I think the inner optimizer  should be unscaled instead. Since the scaler will mark the optimizer state based on its id, giving wrapped optimizer and unscaling it would make the gradient get unscaled but the inner optimizer is not marked as unscaled. Then when calling `accelerator.backward()`, gradient get unscaled again.

sgugger:
As mentioned in #113, the `unscale_gradients` method of `Accelerator` is unscaling the wrapped optimizer and not the inner one, which results in gradients being unscaled twice.

This PR fixes that.

Fixes #113

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -413,8 +413,10 @@ class Accelerator:
                 optimizer = self._optimizers
             elif not isinstance(optimizer, (tuple, list)):
                 optimizer = [optimizer]
-            for optimizer in optimizer:
-                self.scaler.unscale_(optimizer)
+            for opt in optimizer:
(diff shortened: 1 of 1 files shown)


LysandreJik (approved):
LGTM! Weird failure in the tests

sgugger:
The tests are not properly setup :see_no_evil:

Issue #119, opened 2021-07-06
Wrong dataloader length when split_batches = True


topwasu:
I think when `split_batches=True`, the output of `len(dataloader)` is X times smaller than what it's supposed to actually be, where X is the number of processes. 

Config:
```
compute_environment: LOCAL_MACHINE
distributed_type: MULTI_GPU
fp16: false
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
num_machines: 1
num_processes: 2
```

Code to reproduce the bug:
```
import torch
from accelerate import Accelerator
accelerator = Accelerator(split_batches=True)
x = torch.rand(100, 1024)
y = torch.arange(100)

dataset = torch.utils.data.TensorDataset(x, y)
dataloader = accelerator.prepare(torch.utils.data.DataLoader(dataset, batch_size=10))

ct = 0
for batch in dataloader:
    ct += 1
accelerator.print(ct, len(dataloader))
```

Excepted output:
`10 10`

The output I got:
`10 5`

sgugger:
Indeed, can reproduce and this should be fixed by the PR linked above. Thanks a lot for reporting!

PR #121, opened 2021-07-07
Fix DataLoader length when split_batches=True

sgugger:
As pointed out by #119, the dataloader length is incorrect when `split_batches=True` (even if the length of the iterator is the right one). This PR fixes that by fixing the length of the batch sampler used.


LysandreJik (approved):
LGTM

PR #127, opened 2021-07-20
Fix `OptimWrapper` init

### LINKED ISSUE #126: Error when loading optimizer state
Thanks for this awesome product!!

When I try to load optimizer state on TPUs, I get an error, since self.device_placement is never initialized in AcceleratedOptimizer..!
https://github.com/huggingface/accelerate/blob/e0a420f7cb32124cadeeae690b56e463f8fc598f/src/accelerate/optimizer.py#L83-L86

sgugger:
The `device_placement` was not initialized in the `OptimWrapper` init, this PR fixes that.
Fixes #126


LysandreJik (approved):
LGTM!

Issue #101, opened 2021-06-07
Could there be a bug in mixed precision?


voldemortX:
When I use torch 1.6.0 & accelerate 0.3.0 and set mixed precision as `yes` in `accelerate config`, nothing happens (still full precision training). If I set in the code `Accelerator(fp16=True)` then amp is triggered, but the loss becomes inf right away.

But if I use the pytorch way (i.e. autocast in the code myself), the training is normal and amp is enabled.

So I wonder if there is a possible bug in accelerate.

My enviroment is single 2080 Ti, local machine.
The code with this problem is [here](https://github.com/voldemortX/DST-CBC/pull/9).

sgugger:
Hi there, how do you know the script is not launched in mixed precision but use full precision? Just tried on my side and it runs properly in mixed precision.

voldemortX:
Well, I know that because the training speed & memory usage is the same as full precision.

voldemortX:
I think maybe this is due to my code structure (training may be too complex?).

sgugger:
Memory usage won't be very different unless you are using a very large batch size. For the speed you have to make sure all the dimensions of the tensors you are using are multiple of 8.

voldemortX:
I did use a batch size of 8. If I use PyTorch's own autocast, speed-up is normal. (mixed precision performance is very obvious on a Turing card)

sgugger:
It's hard to say what's going wrong without seeing any code.

voldemortX:
Sorry I don't really have a simple reproduceable code sample...
It's a rather big project, if you're interested, the pytorch autocast version (which is working) is here:
https://github.com/voldemortX/DST-CBC/blob/master/segmentation/main.py

voldemortX:
On second thought, probably it's because I wrapped my code with `with autocast(False)`?

sgugger:
You are not letting Accelerate handle mixed precision here, you are doing it in your script yourself: when the `is_mixed_precision` flag is True, you are also scaling the loss which means it will be scaled twice.

voldemortX:
Thanks for your answers! I'll try remove my own mixed precision handling, and see if the loss explosion issue can be reproduced.

voldemortX:
@sgugger I removed my own mixed precision handling:
https://github.com/voldemortX/DST-CBC/blob/test-fp16/segmentation/main.py

[...]

PR #134, opened 2021-08-05
Fix fp16 by converting outputs back to FP32

sgugger:
As was pointed out in #101, there is currently a bug in mixed precision training: the outputs are properly computed in mixed precision but are returned in FP16, and the loss computation is not inside a `torch.cuda.amp.autocast` context manager so is executed in full FP16, which is generally unstable (especially for softmax).

This was not discovered with Transformers models as the loss is computed inside the model, which is generally a better idea if one wants to use mixed precision with Accelerate.

To fix the problem, this PR:
- automatically converts the outputs of the model to FP32 so that the loss is executed in full precision (generally a better idea)
- adds a context manager `accelerator.autocast` for more complex loss functions that should be executed in mixed precision (as with all things accelerate, this context manager always work, it just does nothing if FP16 is not activated).

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -556,7 +556,7 @@ class Accelerator:
     @contextmanager
     def autocast(self):
         """
-        Will apply automatic mixed precision inside the block inside thics context manager, if it is enabled. Nothing
+        Will apply automatic mixed-precision inside the block inside this context manager, if it is enabled. Nothing
         different will happen otherwise.
         """
         if self.native_amp:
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -550,3 +552,17 @@ def get_state_dict(self, model):
                 state_dict[k] = state_dict[k].float()
 
         return state_dict
+
+    @contextmanager
+    def autocast(self):
+        """
+        Will apply automatic mixed precision inside the block inside thics context manager, if it is enabled. Nothing
```
LysandreJik, line 559:
```suggestion
        Will apply automatic mixed-precision inside the block inside this context manager, if it is enabled. Nothing
```

LysandreJik (approved):
Looks great! Cool workaround on `autocast`. LGTM!

PR #138, opened 2021-08-09
Add caveat on weight-tying on TPUs

sgugger:
This PR adds documentation on weight-tying on TPUs and the fact it gets broken.


LysandreJik (approved):
LGTM, great tip

PR #139, opened 2021-08-09
Add optimizer not stepped property

sgugger:
This PR adds a property to check whether or not the optimizer step was skipped in Mixed Precision or DeepSpeed training. This is not necessary to use in most cases, but could be important if there are not many training steps.

### DIFF
--- a/docs/source/quicktour.rst
+++ b/docs/source/quicktour.rst
@@ -358,6 +358,20 @@ softmax). However you might want to put your loss computation inside the `accele
     with accelerator.autocast():
         loss = complex_loss_function(outputs, target):
 
+Another caveat with Mixed Precision training is that the gradient will skip a few updates at the beginning and
+sometimes during training: because of the dynamic loss scaling strategy, there are points at training where the
+gradients have overflown, and the loss scaling factor is reduced to avoid this happening again at the next step.
+
+This means that you may update your learning rate scheduler when there was no update, which is fine in general, but may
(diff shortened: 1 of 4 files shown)


docs/source/quicktour.rst
```diff
@@ -358,6 +358,20 @@ softmax). However you might want to put your loss computation inside the `accele
     with accelerator.autocast():
         loss = complex_loss_function(outputs, target):
 
+Another caveat with Mixed Precision training is that the gradient will skip a few updates at the beginning and
+sometimes during training: because of the dynamic loss scaling strategy, there are points at training where the
```
LysandreJik, line 362:
```suggestion
sometimes during training: because of the dynamic loss scaling strategy, there are points during training where the
```

docs/source/quicktour.rst
```diff
@@ -358,6 +358,20 @@ softmax). However you might want to put your loss computation inside the `accele
...
+gradients have overflown, and the loss scaling factor is reduced to avoid this happening again at the next step.
+
+This means that you may update your learning rate scheduler when there was no update, which is fine in general, but may
+have an impace when you have very little training data, or if the first learning rate values of your scheduler are very
```
LysandreJik, line 366:
```suggestion
have an impact when you have very little training data, or if the first learning rate values of your scheduler are very
```

LysandreJik (approved):
Great workaround!

PR #150, opened 2021-08-30
[DeepSpeed] allow untested optimizers deepspeed

patrickvonplaten:
Similar to how it's done in `transformers`, I think we should always allow untested optimizers, e.g. for AdamW.

See: https://github.com/huggingface/transformers/blob/b6ddb08a664565848c4741e0ff949a4d12989e83/src/transformers/deepspeed.py#L342

cc @stas00 @vasudevgupta7

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -388,4 +388,5 @@ class DeepSpeedPlugin:
                 },
             },
             "steps_per_print": float("inf"),  # this will stop deepspeed from logging @ stdout
+            "zero_allow_untested_optimizer": True,
         }


sgugger (approved):
Agreed, thanks for adding this!

stas00:
FYI, I'm not involved with accelerate's deepspeed integration. I suggested that it will be done in same way as transformers when it was first contributed, but it was still committed in a different way. Thus now the user has to use a completely different setup when switching from transformers and developers have to maintain something totally different. So I'm not participating in this particular endeavour.

PR #151, opened 2021-08-30
[Docs] Machine config is yaml not json

patrickvonplaten:
I think the default config is in `.yaml` format not in `.json`. Was a bit confused when reading the docs.


sgugger (approved):
Thanks a lot for fixing!

PR #152, opened 2021-08-30
Fix gather for 0d tensor

sgugger:
Right now, using `accelerator.gather` on 0d tensors will fail since the tensors can't be concatenated from each process. This PR fixes that by adding a new dimension in that case.


LysandreJik (approved):
Nice trick

PR #164, opened 2021-09-15
Central dataloader

sgugger:
This PR adds a new API called central dataloader. The main concept is that the dataloader passed along to the `accelerator.prepare` method will behave like this: they will only be iterated on the main process, which will then dispatch the batches on all the other processes.

This solves two different problems:
- it makes sure all processes have the same underlying data (a feat usually achieved by synchronizing random number generators) when there is some random events out of the script control that can change the data: for instance we are streaming a dataset and there could be timeouts on some of the processes but not all
- when using an IterableDataset, it avoids iterating through (and thus processing) the data number of processes times, which saves CPU resources and make sure the GPUs are maxed.

The cost is a new synchronization at each batch, to broadcast the batch gathered on process 0 to the other processes.

The feature is activated by passing `central_dataloader=True` to the `Accelerator` at its creation. It is compatible with `split_batches=True` (in which case one batch of the dataloader on process 0 is split on each process, instead of iterating through the central dataloader to get `num_processes` batches and sending each to their right process). It requires `put_on_device=True`.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -82,7 +82,7 @@ class Accelerator:
 
             Will default to :obj:`["torch"]` for PyTorch versions <=1.5.1 and :obj:`["generator"]` for PyTorch versions
             >= 1.6.
-        main_dataloader (:obj:`bool`, `optional`, defaults to :obj:`False`):
+        central_dataloader (:obj:`bool`, `optional`, defaults to :obj:`False`):
             If set to :obj:`True`, the datalaoder prepared by the Accelerator is only iterated through on the main
             process and then the batches are split and broadcast to each process.
         kwargs_handlers (list of kwargs handlers, `optional`)
@@ -103,7 +103,7 @@ class Accelerator:
         cpu: bool = False,
         deepspeed_plugin: DeepSpeedPlugin = None,
         rng_types: Optional[List[Union[str, RNGType]]] = None,
-        main_dataloader: bool = False,
+        central_dataloader: bool = False,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
         if deepspeed_plugin is None:  # init from env variables
@@ -117,7 +117,7 @@ class Accelerator:
 
         self.device_placement = device_placement
         self.split_batches = split_batches
-        self.main_dataloader = main_dataloader
+        self.central_dataloader = central_dataloader
 
         # Kwargs handlers
         self.ddp_handler = None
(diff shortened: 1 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -82,6 +82,9 @@ class Accelerator:
 
             Will default to :obj:`["torch"]` for PyTorch versions <=1.5.1 and :obj:`["generator"]` for PyTorch versions
             >= 1.6.
+        central_dataloader (:obj:`bool`, `optional`, defaults to :obj:`False`):
```
LysandreJik, line 85: I don't like `central` as I don't really understand what it means from a quick read, but I spent 10 minutes trying to find a better name without success, so fine to leave it like it is :)

src/accelerate/data_loader.py
```diff
@@ -421,6 +524,13 @@ def prepare_data_loader(
     if new_batch_sampler is None:
         kwargs["batch_size"] = dataloader.batch_size // num_processes if split_batches else dataloader.batch_size
 
+    if central_dataloader:
+        data_loader = DataLoaderDispatcher(
+            new_dataset, split_batches=split_batches, batch_sampler=new_batch_sampler, **kwargs
+        )
+        data_loader.my_batch_size = dataloader.batch_size
```
LysandreJik, line 531: What is this `my_batch_size` used for?

src/accelerate/test_utils/test_script.py
```diff
@@ -108,6 +108,71 @@ def dl_preparation_check():
...
+
+    dl = DataLoader(range(length), batch_size=8)
+    dl = prepare_data_loader(
+        dl, state.device, state.num_processes, state.process_index, put_on_device=True, central_dataloader=True
+    )
+    result = []
+    for batch in dl:
+        result.append(gather(batch))
+    result = torch.cat(result)
+    assert torch.equal(result.cpu(), torch.arange(0, length).long())
```
LysandreJik, line 123: Please add an error message if it fails :)

LysandreJik: Oh actually this applies to all `assert` in this file!

LysandreJik (approved):
This looks good @sgugger, great job on this

src/accelerate/data_loader.py
sgugger: Woopsie, this is some leftover to clean up! Thanks for flagging!

src/accelerate/test_utils/test_script.py
sgugger: This is just a test script ;-)

src/accelerate/accelerator.py
sgugger: Let's go for `dispath_batches` to look like `split_batches`

PR #168, opened 2021-09-20
Dynamic default for `dispatch_batches`

sgugger:
This PR changes the default of `dispatch_batches` to None, so that it becomes more dynamic:
- for `IterableDataset` it will then default to `True`
- for other datasets, it will keep the old default of `False`

This is because the option is really better and safer for `IterableDataset` (cc @lvwerra )

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -454,7 +454,7 @@ def prepare_data_loader(
         This does not support :obj:`BatchSampler` with varying batch size yet.
     """
     if dispatch_batches is None:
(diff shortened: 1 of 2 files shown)


src/accelerate/data_loader.py
```diff
@@ -452,6 +453,9 @@ def prepare_data_loader(
 
         This does not support :obj:`BatchSampler` with varying batch size yet.
     """
+    if dispatch_batches is None:
+        dispatch_batches = False if not put_on_device else isinstance(new_dataset, IterableDataset)
```
LysandreJik: Clean!

LysandreJik (approved):
LGTM

Issue #172, opened 2021-09-26
A bug in the function convert_to_fp32()


xyh97:
Hi, recently I found that there might be a bug in the function convert_to_fp32().
https://github.com/huggingface/accelerate/blob/1b1463fe2c79a48ed81da9b8ec2dbbfa08c05a7a/src/accelerate/utils.py#L258
I think the correct one should be
`return recursively_apply(_convert_to_fp32, tensor, test_type=_is_fp16_tensor) `

Thanks.

reppy4620:
I met the error because of this i think.

If set fp_16 = True in Accelerator, model return "True".

sgugger:
Thanks for the report and the hint for the fix. It's fixed by the PR mentioned above and I will do a patch release on Monday with this fix.

PR #173, opened 2021-09-26
fix bug #172

reppy4620:
Currently, when fp_16 = True is set in the Accelerator, the model returns a bool, so I fixed it based on #172.


sgugger:
Thanks for fixing this!

PR #175, opened 2021-09-27
Fix length of DispatchDataLoader

sgugger:
Currently, the `DispatchDataLoader` object as the same length as the underlying dataloader, even if it will return 8 times less batches.


LysandreJik (approved):
LGTM

PR #177, opened 2021-09-29
Fix send_to_device with non-tensor data

### LINKED ISSUE #176: TypeError: Can't apply _send_to_device on object of type <class 'int'>, only of nested list/tuple/dicts of objects that satisfy _has_to_method.
When I use 0.4.0 version , it is ok. But it is error when using 0.5.1 version. And the data is a dict, including tensor and List[List[int]]

sgugger:
By mistake, the refactor in #164 introduced a bug in `send_to_device` and made it error when non-tensor data is passed. This PR fixes that.

Fixes #176

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -201,7 +201,7 @@ def send_to_device(tensor, device):
     def _has_to_method(t):
         return hasattr(t, "to")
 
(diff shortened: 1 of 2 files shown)


LysandreJik (approved):
Looks good to me - could a test catch this issue?

PR #180, opened 2021-10-01
Use collections.abc.Mapping to handle both the dict and the UserDict types

mariosasko:
As discussed in #179, it's enough to use `collections.abc.Mapping` to handle both the `dict` and the `UserDict` types in the codebase. This PR addresses that.

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import importlib
 import os
 import random
-from collections.abc import Mapping
 from dataclasses import dataclass, field
 from enum import Enum
 from typing import List, Optional, Union
@@ -163,7 +163,16 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
                 for o in data
             ),
         )
-    elif isinstance(data, Mapping):
+    elif isinstance(data, collections.UserDict):
+        return type(data)(
+            {
+                k: recursively_apply(
+                    func, v, *args, test_type=test_type, error_on_other_type=error_on_other_type, **kwargs
+                )
+                for k, v in data.items()
+            }
+        )
+    elif isinstance(data, dict):
         return type(data)(
             **{
                 k: recursively_apply(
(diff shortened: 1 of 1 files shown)


mariosasko:
cc @sgugger

src/accelerate/utils.py
```diff
@@ -163,16 +163,7 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
...
-                k: recursively_apply(
-                    func, v, *args, test_type=test_type, error_on_other_type=error_on_other_type, **kwargs
-                )
-                for k, v in data.items()
-            }
-        )
-    elif isinstance(data, dict):
+    elif isinstance(data, Mapping):
         return type(data)(
             **{
```
sgugger, line 168: You would need to remove the `**` for this to work with UserDict I believe.
```suggestion
            {
```

src/accelerate/utils.py
```diff
@@ -471,9 +462,7 @@ def concatenate(data, dim=0):
     """
     if isinstance(data[0], (tuple, list)):
         return honor_type(data[0], (concatenate([d[i] for d in data], dim=dim) for i in range(len(data[0]))))
-    elif isinstance(data[0], UserDict):
-        return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
-    elif isinstance(data[0], dict):
+    elif isinstance(data[0], Mapping):
         return type(data[0])(**{k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
```
sgugger, line 466:
```suggestion
        return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
```
Same here

sgugger (approved):
Thanks for taking the time!

src/accelerate/utils.py
mariosasko: `UserDict` also works with the `**` syntax. You can test this with the following code or by looking at the source [here](https://github.com/python/cpython/blob/2c56c97f015a7ea81719615ddcf3c745fba5b4f3/Lib/collections/__init__.py#L964-L984) (warning you that the code is a bit messy in Python 3.6-3.8 compared to Python 3.9):
```python
from collections import UserDict
assert UserDict(**{"a": 1, "b": 2}) == UserDict({"a": 1, "b": 2})
```

sgugger: But not the `BatchEncoding` subclass we are doing all of this for :-)

You can try and see that:
```
from transformers import BatchEncoding
assert BatchEncoding(**{"a": 1, "b": 2}) == BatchEncoding({"a": 1, "b": 2})
```
fails

mariosasko: Oh, I see. Thanks.

mariosasko:
@sgugger I've reverted the changes in these two functions, so now the code should work with `BatchEncoding`. However, even though this would add some complexity, maybe a cleaner approach would be to use a specific check for `BatchEncoding`. But for now, I think the current solution is good enough.

src/accelerate/utils.py
sgugger: This one still needs to be changed.

sgugger:
Thanks again for your contribution!

PR #183, opened 2021-10-09
fix: use `store_true` on argparse in nlp example

monologg:
In example, it makes error when using the commands below.

```bash
python ./nlp_example.py --cpu
python ./nlp_example.py --fp16
```

So fixed the argparse option as `--store_true`, which is also same in `cv_example.py`

https://github.com/huggingface/accelerate/blob/c5c73e02389af9be0ff5e928ea0e94527b434846/examples/cv_example.py#L189-L190


sgugger (approved):
Thanks for fixing!

PR #199, opened 2021-11-12
fix typo in code snippet

MrZilinXiao:
Both `accelerator` or `Accelerator` would do the trick, but `accelerate` won't since we never import it and even we do, the `backward()` method doesn't exist in `accelerate`.


sgugger (approved):
Oh that was a bad typo indeed, thanks a lot for fixing!

PR #204, opened 2021-12-03
Add high-level API reference to README

Chris-hughes10:
* Updates the README with a reference to a high-level API built on top of Accelerate
* Fixes a small type in the notebook launcher's log statement

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,6 @@ Then, you will need to install PyTorch: refer to the [official installation page
 pip install accelerate
 ```
 
-
-
 ## Supported integrations
 
 - CPU only
--- a/README.md
+++ b/README.md
@@ -204,8 +204,14 @@ You should use 🤗 Accelerate when you want to easily run your training scripts
 
 ## Why shouldn't I use 🤗 Accelerate?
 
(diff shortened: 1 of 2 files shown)


sgugger (commented):
Thanks for your PR! I was more thinking of having a separate section named "Frameworks using Accelerate" where you could put the same content. There is also kornia using Accelerate internally in their trainer I would add there :-)

Chris-hughes10:
> Thanks for your PR! I was more thinking of having a separate section named "Frameworks using Accelerate" where you could put the same content. There is also kornia using Accelerate internally in their trainer I would add there :-)

Sure thing, I can create an additional section, migrate the content, and will also look into and add Kornia.

Chris-hughes10:
@sgugger I have updated the README, hopefully that is closer to what you had in mind. Would you also like to me add a corresponding section to the docs?

README.md
```diff
@@ -220,6 +227,8 @@ Then, you will need to install PyTorch: refer to the [official installation page
 pip install accelerate
 ```
 
+
+
```
sgugger, line 231: No need for those two new lines here :-)

sgugger (approved):
Thanks a lot for adding this new section!

README.md
Chris-hughes10: Removed.

PR #206, opened 2021-12-08
fix rng_types in accelerator

s-kumano:
The lines 152--154 in `accelerator.py` have the risk of causing `AttributeError: 'Accelerator' object has no attribute 'rng_types'` (ex. line 392 in `accelerator.py`)

So this PR:
- Define `self.rng_types` in advance.


sgugger (approved):
Good catch!

Issue #220, opened 2021-12-30
Attribute name conflicts in AcceleratedOptimizer


yuxinyuan:
I notice that `AcceleratedOptimizer` uses attribute name `state` to store a reference to `AcceleratorState`. However, pytorch's `Optimizer` class also has a `state` attribute. (Links are pasted below)

https://github.com/huggingface/accelerate/blob/18e5a56cbbca9c6a344853b5dce7675e2c5798ea/src/accelerate/optimizer.py#L56

https://github.com/pytorch/pytorch/blob/fb78a31916d471ebdfb0466dc53446f27f4f41b2/torch/optim/optimizer.py#L44

Since `AcceleratedOptimizer` should only serve as a thin wrapper for `Optimizer`, I think the `state` attribute of `AcceleratedOptimizer`  should be renamed.

sgugger:
Good point. I won't have time to work on this today and am leaving on vacation, but will do something about this the week I'm back!

Or if you want to open a PR, we could rename this attribute to `accelerator_state` and then have a property that returns the inner optimizer state when a user tries to access the state attribute.

PR #224, opened 2022-01-04
Rename state to avoid name conflicts with pytorch's Optimizer class.

yuxinyuan:
#220 
This pull request tries to solve the abovementioned issue. It renames the original `state` attribute to `accelerator_state` and then have a property that returns the inner optimizer state when a user tries to access the state attribute.


sgugger (approved):
Thanks a lot for tackling this!

PR #227, opened 2022-01-10
Fix lr scheduler num samples

### LINKED ISSUE #225: Should `get_linear_schedule_with_warmup` consider `gradient_accumulation_steps`?
I think `num_training_steps` should be `len(train_dataloader) * num_epochs // gradient_accumulation_steps`

sgugger:
This PR fixes the length of the LR scheduler in the NLP example, as pointed out in #225.

Fixes #225


LysandreJik:
:+1:

PR #238, opened 2022-01-26
add support of gather_object

### LINKED ISSUE #235: stronger gather implement
`gather` is crucial in evaluating model. current implementation assumes each tensor across devices has same size, which is very unlikely to happen in the real case. calling `pad_across_processes` results to additional 0s in the output, leading to incorrect results. 

```python
def _gpu_gather_one(tensor, same_size = True):
    if tensor.ndim == 0:
        tensor = tensor.clone()[None]

    if not same_size:
        size = torch.tensor(tensor.shape[0])[None]
        size_list = [torch.zeros(1) for _ in range(torch.distributed.get_world_size())]
        torch.distributed.all_gather(size_list, size)

    output_tensors = [torch.zeros_like(tensor) for _ in range(torch.distributed.get_world_size())]
    torch.distributed.all_gather(output_tensors, tensor)

    if not same_ssize:
        output_tensors = [i[:j] for i, j in zip(output_tensors, size_list)]
    return torch.cat(output_tensors, dim=0)
```

one another workaround is to convert results into `list` and use `gather_object` to gather. However, `gather_object` is also not provided in accelerate.

ZhiyuanChen:
fixes #235

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -18,7 +18,7 @@ import random
 from collections.abc import Mapping
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import List, Optional, Union
+from typing import Any, List, Optional, Union
 
 import numpy as np
 import torch
@@ -345,6 +345,40 @@ def gather(tensor):
         return tensor
 
 
+
+def _gpu_gather_object(object: Any):
+    def _gpu_gather_object_one(object: Any):
+        output_objects = [None for _ in range(torch.distributed.get_world_size())]
+        torch.distributed.all_gather_object(output_objects, object)
+        return [i for j in output_objects for i in j]
+
+    return recursively_apply(_gpu_gather_one, tensor, error_on_other_type=True)
+
+
+_cpu_gather_object = _gpu_gather_object
+
+
+def gather_object(object: Any):
+    """
+    Recursively gather object in a nested list/tuple/dictionary of objects from all devices.
+
+    Args:
+        object (nested list/tuple/dictionary of :obj:`Any`):
+            The data to gather.
+
+    Returns:
+        The same data structure as :obj:`Any` with all objects sent to the proper device.
+    """
+    if AcceleratorState().distributed_type == DistributedType.TPU:
+        raise NotImplementedError('gather objects in TPU is not supported')
+    elif AcceleratorState().distributed_type == DistributedType.MULTI_GPU:
+        return _gpu_gather_object(object)
+    elif AcceleratorState().distributed_type == DistributedType.MULTI_CPU:
+        return _cpu_gather_object(object)
+    else:
+        return object
+
+    
 def _gpu_broadcast(data, src=0):
     def _gpu_broadcast_one(tensor, src=0):
         torch.distributed.broadcast(tensor, src=src)
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -368,7 +368,7 @@ def gather_object(object: Any):
     Recursively gather object in a nested list/tuple/dictionary of objects from all devices.
 
     Args:
-        object (nested list/tuple/dictionary of any object):
+        object (nested list/tuple/dictionary of picklable object):
             The data to gather.
 
     Returns:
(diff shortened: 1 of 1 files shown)


src/accelerate/utils.py
```diff
@@ -345,6 +345,40 @@ def gather(tensor):
...
+
+_cpu_gather_object = _gpu_gather_object
+
+
+def gather_object(object: Any):
+    """
+    Recursively gather object in a nested list/tuple/dictionary of objects from all devices.
+
+    Args:
+        object (nested list/tuple/dictionary of :obj:`Any`):
```
sgugger, line 366:
```suggestion
        object (nested list/tuple/dictionary of any object):
```

src/accelerate/utils.py
```diff
@@ -345,6 +345,40 @@ def gather(tensor):
         return tensor
 
 
+
+def _gpu_gather_object(object: Any):
+    def _gpu_gather_object_one(object: Any):
+        output_objects = [None for _ in range(torch.distributed.get_world_size())]
+        torch.distributed.all_gather_object(output_objects, object)
+        return [i for j in output_objects for i in j]
```
sgugger, line 353: I don't understand this line. This presupposes that every element in `output_objects` is iterable, but `object` is supposed to be a primitive object at this point (tensor or other if we remove `error_on_type=True`).

sgugger (commented):
Thanks a lot for working on this! Could you explain a bit more the inner work of your implementation?

ZhiyuanChen:
> I don't understand this line. This presupposes that every element in output_objects is iterable, but object is supposed to be a primitive object at this point (tensor or other if we remove error_on_type=True).

Should we separate `gather_object` to two functions, `gather_iterable` perform `[i for j in output_objects for i in j]` so it works much like original `gather`, while `gather_object` doesn not perform concatenation so it has better support for every objects

sgugger:
It can't really concatenate if the tensors are not of the same size, so `gather_object` should not concatenate IMO.

ZhiyuanChen:
> It can't really concatenate if the tensors are not of the same size, so `gather_object` should not concatenate IMO.

Yes we can leave it to the user to concat manually.

ZhiyuanChen:
I have resolved all conversations, could you please review it again?

sgugger (approved):
Thanks for addressing all comments, this looks good to merge!

ZhiyuanChen:
> Thanks for addressing all comments, this looks good to merge! Could you just run `make style` on your branch to fix the quality failure?

Sorry for this late reply.... I believe I've resolved the quality issue.

src/accelerate/utils.py
```diff
@@ -350,6 +350,39 @@ def gather(tensor):
         return tensor
 
 
+def _gpu_gather_object(object: Any):
+    def _gpu_gather_object_one(object: Any):
+        output_objects = [None for _ in range(state.num_processes)]
```
sgugger, line 355:
```suggestion
        output_objects = [None for _ in range(AcceleratorState().num_processes)]
```
`state` is not defined.

src/accelerate/utils.py
```diff
@@ -350,6 +350,39 @@ def gather(tensor):
...
+        torch.distributed.all_gather_object(output_objects, object)
+        return output_objects
+
+    return recursively_apply(_gpu_gather_one, tensor)
```
sgugger, line 359:
```suggestion
    return recursively_apply(_gpu_gather_object_one, object)
```

sgugger (commented):
Here is how to solve the last quality issues (variables not defined).

ZhiyuanChen:
Hmmmm, is it possible for us to determine the type of input data and auto call corresponding gather function? I just noticed [torchvison](https://github.com/pytorch/vision/blob/main/references/detection/utils.py#L70-L83) uses `gather_object` at all time

sgugger:
That's too magical for Accelerate's philosophy, users should pick which function they want to use.

All green now so we can merge your PR :-)

ZhiyuanChen:
> That's too magical for Accelerate's philosophy, users should pick which function they want to use.

Haha, thats the reason why i love this repo so much. Users should be trained programmer and should know what they're doing/supposed to do. 

> All green now so we can merge your PR :-) Thanks again!

Thank you for all the review and comments~

PR #241, opened 2022-01-28
enhance compatibility of honor type

### LINKED ISSUE #237: typing (list/tuple/dict) in accelerate.utils
Hi, 

I can't help noticing the typing in accelerate.utils is a bit messy.

The type check for `dict` is actually `Mapping` (which I think is a better approach as it allows user defined `Mapping`, maybe there needs some doc work.
And the type check for list/tuple is `(list, tuple)` which I believe `Sequence` could works better. 

Also, https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils.py#L119-L122 has specific behavior for type `namedtuple` and may not work with some other type, maybe it's better to use EAFP here?

```python
    try:
        return type(obj)(generator)
    except TypeError:
        return type(obj)(*list(generator))
```

ZhiyuanChen:
fixes #237

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -115,12 +115,11 @@ def honor_type(obj, generator):
     """
     Cast a generator to the same type as obj (list, tuple or namedtuple)
     """
-    # There is no direct check whether an object if of type namedtuple sadly, this is a workaround.
-    if isinstance(obj, tuple) and hasattr(obj, "_fields"):
-        # Can instantiate a namedtuple from a generator directly, contrary to a tuple/list.
+    try:
+        return type(obj)(generator)
+    except TypeError:
+        # Some objects may not be able to instantiate from a generator directly
         return type(obj)(*list(generator))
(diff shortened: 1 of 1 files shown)


ZhiyuanChen (commented):
Maybe it would be better if we raise an exception here if cannot re-instantiation the original type

sgugger (approved):
What did you have in mind? The second try will raise an exception if it doesn't succeed so we will already have something.

Thanks a lot for your PR in any case!

ZhiyuanChen:
Sorry I was on my Chinese new year leave and just get back.
Yeah I do agree this should raise exact error with enough information for debugging.

sgugger:
Could you just run `male style` on your branch so we can merge this?

ZhiyuanChen:
hi, I just noticed the existence of this pr......
I've fixed the linting issue related to this pr, but there seems to be more on other code

sgugger:
Seems like the checks are all passing now, thanks again!

PR #246, opened 2022-02-04
make deepspeed optimizer match parameters of passed optimizer

jmhessel:
I am trying to do some prompt tuning with deepspeed+accelerate, and by default, when using deepspeed, all model parameters are updated. Instead, I'd like to only update the ones that I put in the original optimizer.


sgugger (approved):
Thanks for fixing!

sgugger:
Could you just run `make style` on your branch to fix the quality issue? Thanks!

jmhessel:
is this okay? it looks like there are a few other lines in there that also got style fixed

sgugger:
Looks good, thanks again!

Issue #243, opened 2022-02-01
Feature Request: bfloat16 training


ikergarcia1996:
Hi!
Thanks for the accelerate project, It is helping me a lot in my research :)

I am working with T5 models. Training T5 models (I found this issue with mT5 and T5-large) with FP16 is not possible (without tricks that may slowdown the training) the loss will overflow or underflow leading to NaN loss (https://github.com/huggingface/transformers/pull/10956). The best solution proposed up to date is training with BF16 instead of FP16. However I cannot find any option to use BF16 instead of FP16 in accelerate. 
![LossNan](https://user-images.githubusercontent.com/18737249/151950811-939f9bc1-6ce2-446b-b76f-8743c8f4a98e.png)

The adoption of the bf16 format is spreading and more and more models are being trained in this format. In addition, hardware that supports bf16 is also becoming more widespread. Would it be possible to add support for bf16 in accelerate? Or is there a way to use BF16 that I haven't figured out? I can't find anything in the documentation

sgugger:
Hi there! You're right it's not added yet. It came up pretty recently (and you will need a very recent GPU) or were you talking of bfloat16 support for TPU training?

ikergarcia1996:
I am talking about running bfloat16 in a GPU. I am currently experimenting with an Ampere GPU that supports this feature. 
Taking a look at the AMP documentation in PyTorch, using bf16 should be as simple as adding an additional parameter to autocast 
```
with autocast(dtype=torch.bfloat16):
    ...
```
Would it be possible to add an option to the configuration to specify the data type of the autocast function? 
I could help with a pull request if that's OK.

sgugger:
You just have to be a bit careful since that argument was only introduced in recent versions of PyTorch and Accelerate supports version 1.4.0 and above, but I think that's the gist of it yes. I'm happy to review a PR if you want to dive into this :-)

neel04:
>... bfloat16 support for TPU training?

Just to confirm - TPU training uses bfloat16 right? And does it work in multi-TPU settings (as provided by [TRC)](https://sites.research.google/trc/about/)

sgugger:
They use bfloat16 internally, but convert back outputs of the layers to FP32. There is a an environment variable you can set to have the training in full bfloat16.

Closing the issue since the PR adding support on GPU has been merged :-)

cyk1337:
Hi @ikergarcia1996, I also met the same `Nan` problem for training mt5 using fp16. How did you solve it?

[...]

PR #247, opened 2022-02-04
[WIP] Add bfloat16 support #243

ikergarcia1996:
Implement bfloat16 support (#243). I added a new flag named "mixed precision" that can be "no" for FP32 training. "fp16" for FP16 training and "bf16" for bfloat16 training. This version of the code is compatible with previous scripts/config files using accelerate, the fp16 flag has the same behavior as before (Should the users be advice to use the new flag when they use the fp16 flag with a deprecation warning?)

I tested the code using a T5-large model. FP16 mixed precision results NaN loss, thanks to bfloat16 training using mixed precision is possible. 

image

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -219,6 +219,13 @@ class Accelerator:
         return self.local_process_index == 0
 
     @property
+    def use_fp16(self):
+        if self.mixed_precision != "no":
+            return True
+        else:
+            return False
+
+    @property
     def mixed_precision(self):
         if self.distributed_type == DistributedType.DEEPSPEED:
             if self.state.deepspeed_plugin.deepspeed_config["fp16"]["enabled"]:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -18,6 +18,7 @@ from contextlib import contextmanager
 from typing import List, Optional, Union
 
 import torch
+import warnings
 
 from packaging import version
 
@@ -64,6 +65,10 @@ class Accelerator:
             Whether or not to use mixed precision training. Will default to the value in the environment variable
             :obj:`USE_FP16`, which will use the default value in the accelerate config of the current system or the
             flag passed with the :obj:`accelerate.launch` command.
+        mixed_precision (:obj:`str`, `optional`):
+            Whether or not to use mixed precision training (fp16 or bfloat16). Will default to the value in the
+            environment variable :obj:`MIXED_PRECISION`, which will use the default value in the accelerate config
+            of the current system or the flag passed with the :obj:`accelerate.launch` command.
         cpu (:obj:`bool`, `optional`):
             Whether or not to force the script to execute on CPU. Will ignore GPU available if set to :obj:`True` and
             force the execution on one process only.
@@ -101,12 +106,24 @@ class Accelerator:
         device_placement: bool = True,
         split_batches: bool = False,
         fp16: bool = None,
+        mixed_precision: str = "no",
         cpu: bool = False,
         deepspeed_plugin: DeepSpeedPlugin = None,
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         dispatch_batches: Optional[bool] = None,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
+        mixed_precision = mixed_precision.lower()
+        assert mixed_precision in [
+            "no",
+            "fp16",
+            "bf16",
+        ], f"Unknown mixed_precision: {mixed_precision}. Choose between 'no', 'fp16' and 'bf16'."
+
+        if fp16:
+            warnings.warn('--fp16 flag is deprecated. Use "--mixed_precision fp16" instead.', DeprecationWarning)
+            mixed_precision = "fp16"
+
         if deepspeed_plugin is None:  # init from env variables
             deepspeed_plugin = DeepSpeedPlugin() if os.environ.get("USE_DEEPSPEED", "false") == "true" else None
         else:
@@ -139,7 +156,11 @@ class Accelerator:
 
         kwargs = self.init_handler.to_kwargs() if self.init_handler is not None else {}
         self.state = AcceleratorState(
-            fp16=fp16, cpu=cpu, deepspeed_plugin=deepspeed_plugin, _from_accelerator=True, **kwargs
+            mixed_precision=mixed_precision,
+            cpu=cpu,
+            deepspeed_plugin=deepspeed_plugin,
+            _from_accelerator=True,
+            **kwargs,
         )
 
         self.device_placement = device_placement
@@ -149,8 +170,12 @@ class Accelerator:
         # Mixed precision attributes
         self.scaler = None
         self.native_amp = False
-        if self.state.use_fp16:
+        if self.state.mixed_precision == "fp16":
             self.native_amp = version.parse(torch.__version__) >= version.parse("1.6")
+            kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
+            self.scaler = torch.cuda.amp.GradScaler(**kwargs)
+        elif self.state.mixed_precision == "bf16":
+            self.native_amp = version.parse(torch.__version__) >= version.parse("1.10")
             kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
             self.scaler = torch.cuda.amp.GradScaler(**kwargs)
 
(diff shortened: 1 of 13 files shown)


src/accelerate/utils.py
```diff
@@ -240,23 +240,23 @@ def _initialize_tensor(tensor_info):
...
+        The same data structure as :obj:`tensor` with all tensors that were in FP16/BF16 precision converted to FP32.
     """
 
     def _convert_to_fp32(tensor):
         return tensor.float()
 
-    def _is_fp16_tensor(tensor):
-        return hasattr(tensor, "dtype") and tensor.dtype == torch.float16
+    def _is_fp16_bf16_tensor(tensor):
+        return hasattr(tensor, "dtype") and (tensor.dtype == torch.float16 or tensor.dtype == torch.bfloat16)
```
sgugger, line 257: This would also need a version check, as I'm not sure the `torch.bfloat16` was there in PyTorch 1.6.

sgugger (approved):
Thanks a lot for your PR, this is great work!

I'm okay with the deprecation cycle, which, coupled with an automatic update of the configs should be pretty smooth for the user.

A few important fixes to bring in the PR:
- make sure to only use the env variable when the `mixed_precision` attribute is not set (we shouldn't overrride a `"no"` passed by the user), and that this attributes default to `None`.
- make sure the bfloat16 functionality is only accessed when we have PyTorch 1.10 or later
I've left comments for those.

src/accelerate/accelerator.py
```diff
@@ -311,7 +348,12 @@ def prepare_model(self, model):
             kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
             model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
         if self.native_amp:
-            model.forward = torch.cuda.amp.autocast()(model.forward)
+            if self.mixed_precision == "fp16":
+                model.forward = torch.cuda.amp.autocast(dtype=torch.float16)(model.forward)
```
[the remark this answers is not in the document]
ikergarcia1996: Version test added

src/accelerate/commands/config/config_args.py
```diff
@@ -97,6 +98,12 @@ def from_yaml_file(cls, yaml_file=None):
             config_dict = yaml.safe_load(f)
         if "compute_environment" not in config_dict:
             config_dict["compute_environment"] = ComputeEnvironment.LOCAL_MACHINE
+        if "mixed_precision" not in config_dict:
```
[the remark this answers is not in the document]
ikergarcia1996: I forgot about the "from_json_file" method, i have updated it too and I added a line to delete the fp16 key from old config files

ikergarcia1996:
I have updated the PR with the fixes you mention

src/accelerate/accelerator.py
```diff
@@ -101,12 +106,29 @@ def __init__(
...
         fp16: bool = None,
+        mixed_precision: str = None,
         cpu: bool = False,
         deepspeed_plugin: DeepSpeedPlugin = None,
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         dispatch_batches: Optional[bool] = None,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
+
+        if mixed_precision:
```
sgugger, line 117:
```suggestion
        if mixed_precision is not None:
```
We usually prefer explicit tests instead of relying on Python's bool conversion.

src/accelerate/accelerator.py
```diff
@@ -101,12 +106,29 @@ def __init__(
...
+            mixed_precision = mixed_precision.lower()
+            if mixed_precision not in [
+                "no",
+                "fp16",
+                "bf16",
+            ]:
```
sgugger, line 123:
```suggestion
            if mixed_precision not in ["no", "fp16", "bf16"]:
```
Should fit in one line now.

src/accelerate/accelerator.py
```diff
@@ -149,10 +175,14 @@ def __init__(
...
         self.native_amp = False
-        if self.state.use_fp16:
+        if self.state.mixed_precision == "fp16":
             self.native_amp = version.parse(torch.__version__) >= version.parse("1.6")
             kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
             self.scaler = torch.cuda.amp.GradScaler(**kwargs)
+        elif self.state.mixed_precision == "bf16":
+            self.native_amp = version.parse(torch.__version__) >= version.parse("1.10")
+            kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
+            self.scaler = torch.cuda.amp.GradScaler(**kwargs)
```
sgugger, line 185: I realize we are silently not training in mixed precision when:
- `mixed_precision="fp16"` and torch version is < 1.6
- `mixed_precision="bf16"` and torch version is < 1.10

I think we should raise an error message in both cases, or at least a warning.

src/accelerate/commands/config/config_args.py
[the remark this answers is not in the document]
sgugger: Nice!

src/accelerate/commands/launch.py
```diff
@@ -53,9 +54,20 @@ def launch_command_parser(subparsers=None):
...
+        choices=["no", "fp16", "bf16"],
+        help="Whether or not to use mixed precision training. "
+        "Choose between FP16 and BF16 (bfloat16) training. "
+        "BF16 training is only supported on Nvidia Ampere GPUs and PyTorch 1.10 or later.",
+    )
+
     parser.add_argument(
         "--fp16", default=False, action="store_true", help="Whether or not to use mixed precision training."
     )
+
```
sgugger, line 70: No need for this new line I think.
```suggestion
```

src/accelerate/commands/launch.py
```diff
@@ -134,7 +146,20 @@ def simple_launcher(args):
...
     current_env = os.environ.copy()
     current_env["USE_CPU"] = str(args.cpu)
-    current_env["USE_FP16"] = str(args.fp16)
+
+    mixed_precision = args.mixed_precision.lower()
+    if mixed_precision not in [
+        "no",
+        "fp16",
+        "bf16",
+    ]:
```
sgugger, line 155:
```suggestion
    if mixed_precision not in ["no", "fp16", "bf16"]:
```

src/accelerate/commands/launch.py
```diff
@@ -168,7 +193,20 @@ def multi_gpu_launcher(args):
...
 
     current_env = os.environ.copy()
-    current_env["USE_FP16"] = str(args.fp16)
+    mixed_precision = args.mixed_precision.lower()
+
+    if mixed_precision not in [
+        "no",
+        "fp16",
+        "bf16",
+    ]:
```
sgugger, line 202:
```suggestion
    if mixed_precision not in ["no", "fp16", "bf16"]:
```

src/accelerate/commands/launch.py
```diff
@@ -201,7 +239,20 @@ def deepspeed_launcher(args):
...
```
sgugger, line 248:
```suggestion
    if mixed_precision not in ["no", "fp16", "bf16"]:
```

src/accelerate/commands/launch.py
```diff
@@ -309,8 +360,21 @@ def sagemaker_launcher(sagemaker_config: SageMakerConfig, args):
...
     hyperparameters = _convert_nargs_to_dict(args.training_script_args)
 
-    environment = {"USE_FP16": args.fp16}  # Environment variables to be set for use during training job
+    mixed_precision = args.mixed_precision.lower()
+
+    if mixed_precision not in [
+        "no",
+        "fp16",
+        "bf16",
+    ]:
```
sgugger, line 369:
```suggestion
    if mixed_precision not in ["no", "fp16", "bf16"]:
```

src/accelerate/notebook_launcher.py
```diff
@@ -39,6 +40,8 @@ def notebook_launcher(function, args=(), num_processes=None, use_fp16=False, use
             the number of GPUs available otherwise.
         use_fp16 (:obj:`bool`, `optional`, defaults to :obj:`False`):
             If :obj:`True`, will use mixed precision training on multi-GPU.
```
sgugger, line 42: You can remove this from the documentation (we try to avoid documenting deprecated args so users don't start using them).

src/accelerate/notebook_launcher.py
```diff
@@ -110,7 +113,22 @@ def notebook_launcher(function, args=(), num_processes=None, use_fp16=False, use
...
             os.environ["MASTER_ADDR"] = "127.0.0.1"
             os.environ["MASTER_PORT"] = str(use_port)
-            os.environ["USE_FP16"] = str(use_fp16)
+
+            mixed_precision = mixed_precision.lower()
+            if mixed_precision not in [
+                "no",
+                "fp16",
+                "bf16",
+            ]:
```
sgugger, line 122:
```suggestion
            if mixed_precision not in ["no", "fp16", "bf16"]:
```

sgugger (approved):
Thanks for iterating, I have a few more comments, mostly regarding styling.

ikergarcia1996:
I think they are solved now. Maybe we should add a warning in 'cluster.py' to warn if the user chooses the "fp16" or "bf16" option but doesn't have the correct version of Pytorch installed?

sgugger:
This [comment](https://github.com/huggingface/accelerate/pull/247/files#r800669214) and this [one](https://github.com/huggingface/accelerate/pull/247/files#r800671521) don't seem to have been addressed. Could you take care of them?
Quality check is also still failing.

We can also add warnings when the options are chosen, yes!

ikergarcia1996:
Sorry the previous commit to solve the problem of silently not training in mixed precision was incorrect. I have a question about implementation. If  the version is correct we set self.native_amp to True. 
```Python
self.native_amp = version.parse(torch.__version__) >= version.parse("1.6")
```
If the version is incorrect we disable mixed precision by setting that variable to False. Is it used for anything else? If we are going to raise an error, should we remove it? 
Should we raise a error always if the pytorch version is not correct? I'm not sure I understand this part of the code 100% and I don't know if I've implemented right the errors if the pytorch version is not compatible with the chosen mixed precision.

sgugger:
The way you did it is great! Thanks again for your awesome contribution :-)

Issue #215, opened 2021-12-23
Feature Request: sum/mean reduce, sequential_ddp_run


Scitator:
Hi!
Thanks for the accelerate project, looks very similar to [my own implementation](https://github.com/catalyst-team/catalyst/blob/master/catalyst/core/engine.py#L20), so... I have a few proposals ;)

-----
What do you think about:
```python
accelerator = ...
accelerator.all_gather(obj)  # already implemented
accelerator.sum_reduce(obj)  # not yet
accelerator.mean_reduce(obj)  # not yet 
```

so you could use accelerator as a backend for metrics computation.

As for examples,
- you always have to use `mean_reduce` for such simple batch-based metrics as `accuracy`
- you always have to use `all_gather` for distributed computation of dataset-based metrics like `AUC` or `Precision/Recall/F1` micro/macro/weighted aggregations.

-----
Another proposal - add something similar to `sequential_ddp_run` or [`ddp_sync_run`](https://github.com/catalyst-team/catalyst/blob/master/catalyst/core/engine.py#L155), so you could run some selected code on 0-rank first. Usecase is simple - dataset preparation - no one want to download MNIST on all process simultaneously ;)

sgugger:
Thanks for the issue, always appreciate having some feedback and ideas of new features! Having more primitives like `sum_reduce` and `mean_reduce` definitely sounds like a good idea. I'm probably not going to have time to dig into that before going on holiday, but I can look into that when I'm back the second week of January. If you want to work on some however, feel free to open a PR :-) 

It think your second proposition is already implemented as a context manager:

```
with accelerator.main_process_first():
```

or the `local_main_process_first` version if it's something that needs to happen on every node.

PR #249, opened 2022-02-08
Accelerate + Animus/Catalyst = 🚀 

Scitator:
Hi guys, nice project!

After recent experiments with accelerate, I found it very close to my own "torch-backend" ideas, so... I decided to move a couple of frameworks to accelerate.

tl;dr
- Catalyst is based on Accelerate starting from [v22 edition](https://github.com/catalyst-team/catalyst/releases/tag/v22.02rc0)
- Animus - from [v0.0.1](https://github.com/Scitator/animus)
- Animus & Catalyst added to readme/frameworks section
- frameworks sorted alphabetically (it's easier to navigate)

✌️

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,11 @@ You shouldn't use 🤗 Accelerate if you don't want to write a training loop you
 
 If you like the simplicity of 🤗 Accelerate but would prefer a higher-level abstraction around your training loop, some frameworks that are built on top of 🤗 Accelerate are listed below:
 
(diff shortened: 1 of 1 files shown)


Scitator:
btw, it would be great if you could comment on my "Accelerator mixin" - [Engine](https://github.com/Scitator/animus/blob/main/animus/torch/engine.py) in terms of "a bit more user-friendly Accelerator API". I found it useful for me to run pipelines within notebooks or python scripts without CLI.

sgugger:
Ah thanks! I am wrapping up some things in Transformers and will go back to Accelerate pretty soon (haven't forgotten #215 !)

You mixins look great! Having one per kind of distributed setup is a bit against the philosophy of Accelerate (where the same exact code runs on any setup) but I like the abstraction.

To run things in notebooks, we also have `notebook_launcher`, I don't know if you tried it or not.

Scitator:
yeah, I found `notebook_launcher` an interesting functional alternative... nevertheless, from previous experience I knew a few limitations of pure functional way, so... a small class-based API was created ;)

sgugger (approved):
Sorry I read this as an issue and not a PR 🤦‍♂️

PR #250, opened 2022-02-08
Upgrade black to version ~=22.0

LysandreJik:
Upgrades the `accelerate` library quality tool `black` to use the first stable release of `black`, version 22.0.


sgugger (approved):
Thanks!

Issue #226, opened 2022-01-10
About the use of gather to compute metrics


jbschiratti:
First of all, thank you for developing Accelerate! I'm new to it but I already love it, it's a great framework.

I used the code below to train a naive model on MNIST data using 3 GPUs (on a single node/machine). This code uses an `Accuracy` class to compute the epoch-wise accuracy from predictions and labels. It stores predictions/labels at each step to eventually compute accuracy (or another metric, such as ROC AUC) _once_ at the end of an epoch. At each step, `accelerator.gather()` is used to gather all predictions/labels from GPU devices. I added a `print` statement in the `Accuracy` class to check the number of samples used to compute the accuracy. The test set of MNIST is composed of 10 000 samples but the print statement in this class shows that 10 167 samples are actually used. 

Full code: 
```python
from __future__ import print_function

import argparse
import os
import os.path
import threading
from functools import partial

import torch
import torch.nn as nn
import torch.nn.functional as F
import torch.optim as optim

from accelerate import Accelerator
from sklearn.metrics import accuracy_score
from torch.optim.lr_scheduler import StepLR
from torch.utils.data import DataLoader
from torchvision import datasets, transforms
from tqdm.auto import tqdm as original_tqdm


class Accuracy:
    """Accuracy score."""
    def __init__(self):
        super().__init__()
        self.__build()

    def __build(self):
        self._lock = threading.Lock()
        self._predictions = []
        self._targets = []

    def reset(self):
        self._predictions.clear()
        self._targets.clear()

    def update(self, output):
        y_pred, y_true = output
        with self._lock:
            self._predictions.append(y_pred)
            self._targets.append(y_true)

    def compute(self):
        with self._lock:
            predictions = torch.cat(self._predictions, dim=0).numpy()
            targets = torch.cat(self._targets, dim=0).numpy()
            print(f'Shapes: predictions {predictions.shape}, targets {targets.shape}')
            return accuracy_score(y_true=targets, y_pred=predictions)


class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(1, 32, 3, 1)
        self.conv2 = nn.Conv2d(32, 64, 3, 1)
        self.dropout1 = nn.Dropout2d(0.25)
        self.dropout2 = nn.Dropout2d(0.5)
        self.fc1 = nn.Linear(9216, 128)
        self.fc2 = nn.Linear(128, 10)

    def forward(self, x):
[...]

PR #252, opened 2022-02-11
Computer vision example on MNIST

jbschiratti:
This PR aims at adding another computer vision example (training a FCN on MNIST data).
In #226, @sgugger agreed that such an example would be useful for Accelerate.
I would be happy to discuss this example (and finalize the PR if it still of interest).


examples/README.md
```diff
@@ -99,7 +99,7 @@ To run it in each of these various modes, use the following commands:
 
 ## Simple vision example
 
-The [cv_example.py](./cv_example.py) script is a simple example to fine-tune a ResNet-50 on a classification task ([Ofxord-IIT Pet Dataset](https://www.robots.ox.ac.uk/~vgg/data/pets/)).
+The [cv_example.py](./pet_datase_example.py) script is a simple example to fine-tune a ResNet-50 on a classification task ([Ofxord-IIT Pet Dataset](https://www.robots.ox.ac.uk/~vgg/data/pets/)).
```
sgugger, line 102: You should also mention you new example here :-)

examples/mnist_example.py
```diff
@@ -0,0 +1,259 @@
...
+# in any of the following settings (with the same script):
+#   - single CPU or single GPU
+#   - multi GPUS (using PyTorch distributed mode)
+#   - (multi) TPUs
+#   - fp16 (mixed-precision) or fp32 (normal precision)
+#
+########################################################################
+
+
+from __future__ import print_function
```
sgugger, line 30: We don't support Python 2, so there is no need for this.

examples/mnist_example.py
```diff
@@ -0,0 +1,259 @@
...
+                preds = output.argmax(dim=1, keepdim=True)
+
+                # Compute accuracy:
+                # We first gather predictions & labels from all processes
+                bs = _test_loader.batch_size
+                y_pred = accelerator.gather(preds).detach().cpu()[:bs]
+                y_true = accelerator.gather(target).detach().cpu()[:bs]
+                test_accuracy.update((y_pred, y_true))
+
+        test_acc = test_accuracy.compute()
```
sgugger, line 222: I'd prefer to see all the logic to compute the accuracy here. We don't need a generic class for this in this particular example, and it would make it easier to read I believe.

examples/mnist_example.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # SAVE TRAINED MODEL (OPTIONAL)
+    if accelerator.is_main_process and args.out_dir is not None:
+        accelerator.wait_for_everyone()
+        unwrapped_model = accelerator.unwrap_model(model)
+        unwrapped_model.save(args.out_dir, save_function=accelerator.save)
+
+
+if __name__ == '__main__':
+
```
sgugger, line 258:
```suggestion
```

sgugger (commented):
Thanks for adding this example! I think it's great to showcase a pure PyTorch example with Accelerate.

I just have two remarks:
- it would be easier to read if the `evaluate` and `train_one_epoch` functions where outside of the main function, so they variables can be named without _. We can pass the `accelerator` object to those :-)
- it would also be easier to read if the code computing the accuracy was directly all inside the `evaluate` function, instead of relying on a class.

Issue #171, opened 2021-09-24
Add utility function to save/load all models/optimizers stored in an accelerator


yuxinyuan:
Hi all,
I'm not from the nlp regime, but I love the concept of accelerate. However, I would like to propose some new features.

Since every accelerator.prepare call will essentially store a reference of the object in accelerator. It should be easy to implement a save_all()/load_all() utility that saves/loads all the pytorch module and optimizer state dicts to/from a checkpoint. This utility also has the benifit to save and load the state dict of grad scaler, and potentially any other internal states (which are all agnostic to accelerator users)

I guess the singature could look something like `def save_all(self, filename: str, dict_of_additional_usr_data: Dict[str, Any])`

What do the maintainers think of this utility?

sgugger:
Yes, this is on the roadmap and should come in the next few weeks! It will save everything you said plus all the RNG states, so that one can resume training in the exact same conditions. It will also provide an easy way to upload said checkpoint to the Hub so that you can easily resume training on another machine.

I am swamped until the end of next week but hoping to get into adding this the first week of October :-)

lvwerra:
Is this feature still on the roadmap?

sgugger:
Yes, I just need the time to dig into it. I have a few things to wrap up in Transformers but plan to work on Accelerate only afterward for a couple of months.

PR #255, opened 2022-02-17
Add in checkpointing capability

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/171

This PR adds in two functions, `Accelerator.save_state` and `Accelerator.load_state`, which will go through and checkpoint the current state of everything accelerator touches.

I've included a testing script, but this needs to be properly turned into a test, as currently it is just a script with a bunch of asserts. I have tested successfully on:

- Single GPU
- Single CPU
- Multi-GPU

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -620,6 +620,56 @@ class Accelerator:
         torch.save(states, output_states_file)
         logger.info(f"Random states saved in {output_states_file}")
 
+    def load_state(self, input_dir: str):
+        """
+        Loads the current states of the model, optimizer, scalar, and RNG generators from `folder_name`.
+
+        Args:
+            input_dir (:obj:`str` or :obj:`os.PathLike`):
+                The name of the folder all relevant weights and states were saved in.
+        """
+        # Check if folder exists
+        input_dir = os.path.expanduser(input_dir)
+        if not os.path.isdir(input_dir):
+            raise ValueError(f"Tried to find {input_dir} but folder does not exist")
+        logger.info(f"Loading states from {input_dir}")
+
+        # Model states
+        for i, model in enumerate(self._models):
+            weights_name = "pytorch_model"
+            if i == 0:
+                weights_name += f"_{i}"
+            weights_name += ".bin"
+            input_model_file = os.path.join(input_dir, weights_name)
+            self._models[i].load_state_dict(torch.load(input_model_file))
+        logger.info("All model weights loaded successfully")
+        # Optimizer states
+        for i, opt in enumerate(self._optimizers):
+            # state = opt.state_dict()
+            optimizer_name = "optimizer"
+            if i == 0:
+                optimizer_name += f"_{i}"
+            optimizer_name += ".pt"
+            input_optimizer_file = os.path.join(input_dir, optimizer_name)
+            self._optimizers.load_state_dict(torch.load(input_optimizer_file))
+        logger.info("All optimizer states loaded successfully")
+        # GradScalar state
+        if self.scalar is not None:
+            scalar_name = "scalar.bin"
+            input_scalar_file = os.path.join(input_dir, scalar_name)
+            self.scalar.load_state_dict(torch.load(input_scalar_file))
+            logger.info("GradScalar state loaded successfully")
+        # Random states
+        states = torch.load(os.path.join(input_dir, "random_states.pkl"))
+        random.setstate(states["random_state"])
+        np.random.set_state(states["numpy_random_seed"])
+        torch.set_rng_state(states["torch_manual_seed"])
+        torch.cuda.set_rng_state_all(states["torch_cuda_manual_seed"])
+        # ^^ safe to call this function even if cuda is not available
+        if is_tpu_available():
+            xm.set_rng_state(states["xm_seed"])
+        logger.info("All random states loaded successfully")
+
     def free_memory(self):
         """
         Will release all references to the internal objects stored and call the garbage collector. You should call this
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -29,6 +29,10 @@ from .kwargs_handlers import DistributedDataParallelKwargs, GradScalerKwargs, In
 from .optimizer import AcceleratedOptimizer
 from .state import AcceleratorState, DistributedType, is_deepspeed_available, is_tpu_available
 from .utils import (
+    MODEL_NAME,
+    OPTIMIZER_NAME,
+    RNG_STATE_NAME,
+    SCALAR_NAME,
     DeepSpeedPlugin,
     RNGType,
     convert_outputs_to_fp32,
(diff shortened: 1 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
         """
         save(obj, f)
 
+    def save_state(self, output_dir: str):
```
sgugger, line 569: As much as possible, I would prefer to have everything in another module and have `accelerator` be a thin wrapper. I think we could have a new `checkpointing.py` file where a function that does all of this (taking `models`, `optimizers`, `scaler`). Same for the loading.

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+        output_dir = os.path.expanduser(output_dir)
+        os.makedirs(output_dir, exist_ok=True)
+        logger.info(f"Saving current state to {output_dir}")
+
+        # Model states
+        for i, model in enumerate(self._models):
+            state = self.get_state_dict(model)
+            weights_name = "pytorch_model"
+            if i == 0:
+                weights_name += f"_{i}"
```
sgugger, line 587:
```suggestion
            if i != 0:
                weights_name += f"_{i}"
```
Is it a typo? I think it's for every model **but** the first one.

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+            weights_name += ".bin"
+            output_model_file = os.path.join(output_dir, weights_name)
+            torch.save(state, output_model_file)
```
sgugger, line 590: Should be `accelerator.save` here (otherwise won't work for TPUs).

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+            logger.info(f"Model weights saved in {output_model_file}")
+        # Optimizer states
+        for i, opt in enumerate(self._optimizers):
+            state = opt.state_dict()
+            optimizer_name = "optimizer"
+            if i == 0:
+                optimizer_name += f"_{i}"
+            optimizer_name += ".pt"
+            output_optimizer_file = os.path.join(output_dir, optimizer_name)
+            torch.save(state, output_optimizer_file)
```
sgugger, line 600: Same here:
```suggestion
            accelerator.save(state, output_optimizer_file)
```

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+            logger.info(f"Optimizer state saved in {output_optimizer_file}")
+        # GradScalar state
+        if self.scaler is not None:
+            state = self.scaler.state_dict()
+            scalar_name = "scaler.bin"
+            output_scalar_file = os.path.join(output_dir, scalar_name)
+            torch.save(state, output_scalar_file)
+            logger.info(f"GradScalar state saved in {output_scalar_file}")
```
sgugger, line 608:
```suggestion
            logger.info(f"Gradient scaler state saved in {output_scaler_file}")
```

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+            optimizer_name += ".pt"
+            output_optimizer_file = os.path.join(output_dir, optimizer_name)
+            torch.save(state, output_optimizer_file)
+            logger.info(f"Optimizer state saved in {output_optimizer_file}")
+        # GradScalar state
+        if self.scaler is not None:
+            state = self.scaler.state_dict()
+            scalar_name = "scaler.bin"
+            output_scalar_file = os.path.join(output_dir, scalar_name)
+            torch.save(state, output_scalar_file)
```
sgugger, line 607:
```suggestion
            output_scaler_file = os.path.join(output_dir, scaler_name)
            torch.save(state, output_scaler_file)
```

src/accelerate/accelerator.py
```diff
@@ -560,6 +566,109 @@ def save(self, obj, f):
...
+        for i, opt in enumerate(self._optimizers):
+            state = opt.state_dict()
+            optimizer_name = "optimizer"
+            if i == 0:
+                optimizer_name += f"_{i}"
+            optimizer_name += ".pt"
+            output_optimizer_file = os.path.join(output_dir, optimizer_name)
+            torch.save(state, output_optimizer_file)
+            logger.info(f"Optimizer state saved in {output_optimizer_file}")
+        # GradScalar state
```
sgugger, line 602: I realize we are missing the learning rate scheduler -> we should thus add it as an optional argument to the `save_state` function, or add a `register_for_checkpoints` method to the Accelerator object that would take objects not passed to the `prepare` method but the user still wants to save.

sgugger (approved):
Thanks a lot for working on this! I've left a few comments on the general API and some nits. The most important thing is to figure out how to save the LR scheduler with the rest :-)

tests/test_state_checkpointing.py
```diff
@@ -0,0 +1,148 @@
...
+        (a3, b3) = model_unwrapped.a.item(), model_unwrapped.b.item()
+        opt_state3 = optimizer.state_dict()
+        assert a1 == a3
+        assert b1 == b3
+        assert opt_state1 == opt_state3
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Simple example of training script.")
+    parser.add_argument("--fp16", action="store_true", help="If passed, will use FP16 training.")
```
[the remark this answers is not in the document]
muellerzr: I'd like some advice on how to convert my script -> a proper test. I tried looking in `test_multigpu` but still doesn't quite line up with how the testing workflow operates

src/accelerate/accelerator.py
```diff
@@ -560,6 +562,36 @@ def save(self, obj, f):
         """
         save(obj, f)
 
+    def save_state(self, output_dir: str):
+        """
+        Saves the current states of the model, optimizer, scaler, and RNG generators to `folder_name`.
```
sgugger, line 567:
```suggestion
        Saves the current states of the model, optimizer, scaler, and RNG generators.
```
It's actually output_dir but since we document it just after, no need to mention it up there I believe.

src/accelerate/accelerator.py
```diff
@@ -560,6 +562,36 @@ def save(self, obj, f):
...
+        # Check if folder exists
+        output_dir = os.path.expanduser(output_dir)
+        os.makedirs(output_dir, exist_ok=True)
+        logger.info(f"Saving current state to {output_dir}")
+        weights = [self.get_state_dict(m) for m in self._models]
+        return save_accelerator_state(output_dir, weights, self._optimizers, self.state.process_index, self.scaler)
+
+    def load_state(self, input_dir: str):
+        """
+        Loads the current states of the model, optimizer, scaler, and RNG generators from `folder_name`.
```
sgugger, line 582:
```suggestion
        Loads the current states of the model, optimizer, scaler, and RNG generators.
```

src/accelerate/checkpointing.py
```diff
@@ -0,0 +1,134 @@
+# Copyright 2021 The HuggingFace Team. All rights reserved.
```
sgugger, line 1:
```suggestion
# Copyright 2022 The HuggingFace Team. All rights reserved.
```

src/accelerate/checkpointing.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def save_accelerator_state(
+    output_dir: str, model_states: List[dict], optimizers: list, process_index: int, scaler: GradScaler = None
+):
+    """
+    Saves the current states of the `models`, `optimizers`, scaler, and RNG generators to `output_dir`.
```
sgugger, line 40:
```suggestion
    Saves the current states of the models, optimizers, scaler, and RNG generators to a given directory.
```
Let's leave the argument names for the arg descriptions and be moe "English" in the description (or if you want them, you need to put them in double backticks or add an `:obj:` to have them code-formatted)

src/accelerate/checkpointing.py
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        output_dir (:obj:`str` or :obj:`os.PathLike`):
+            The name of the folder to save all relevant weights and states.
+        model_states (:obj:`list`):
+            A list of model states
+        optimizers (:obj:`list`):
+            A list of optimizer instances
+        process_index (:obj:`int`):
+            The current process index in the Accelerator state
+        scaler (:obj:`GradScaler`), Optional:
```
sgugger, line 51:
```suggestion
        scaler (:obj:`torch.cuda.amp.GradScaler`, `optional`):
```

src/accelerate/checkpointing.py
```diff
@@ -0,0 +1,134 @@
...
+            An optional `GradScaler` instance to save
```
sgugger, line 52:
```suggestion
            An optional gradient scaler instance to save
```
(The type is already described)

sgugger (approved):
Thanks a lot for continuing your work on this. Just left some nits and one last comment on the test you added, then we can merge this!

tests/test_state_checkpointing.py
```diff
@@ -0,0 +1,144 @@
...
+        "between fp16 and bf16 (bfloat16). Bf16 requires PyTorch >= 1.10."
+        "and an Nvidia Ampere GPU.",
+    )
+    parser.add_argument("--cpu", action="store_true", help="If passed, will train on the CPU.")
+    args = parser.parse_args()
+    CheckpointTest().test_can_resume_training(args)
+
+
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
muellerzr: @sgugger how would we delegate the tests to be ran on cpu, mixed precision, etc then in that case?

Issue #256, opened 2022-02-21
Running launch command without python script


parameter-concern:
Currently, the main argument to the `accelerate launch` command is a python script. For some projects, it would be convenient for me to be able to pass either a python module or a non-python executable (e.g. a shell script) to launch instead. This is not too hard to do manually with the `--module` or `--no_python` flags in `torch.distributed.launch`, but then I lose the convenience of the `accelerate` config and launch commands.

One example use case for this is that sometimes I like to bundle up code and utility commands (like training/fine-tuning and evaluation) into python packages and give them CLI entry points in `setup.py`. It would be nice to be able to pass in the training entry point to `accelerate launch` instead of the underlying `.py` file for consistency and ease of packaging/distributing/running.

If this is something that the maintainers or others would like to see supported, I'm happy to submit a PR, but I would understand if they didn't want to bloat up the launch arguments list and preferred if people just did stuff like this manually.

parameter-concern:
Below are some notes I took while hacking this into my local environment:

In the case of the `simple_launcher`, `multi_gpu_launcher`, and `deepspeed_launcher` used in `commands/launch.py`, this support would be easy to add by mimicking the `--module` and `--no_python` [arguments](https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py#L452) and [command setup](https://github.com/pytorch/pytorch/blob/master/torch/distributed/run.py#L660) from `torch.distributed.launch`.  The deepspeed [args](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/launch.py#L56) and [command setup](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/launcher/launch.py#L155) are very similar and actually a bit simpler to read.

For the `tpu_launcher`, the argument is already interpreted as a module which is imported and run, so a `--no_python` flag could probably just be ignored and setting the `--module` flag could allow the argument to be used by `importlib.import_module` directly if given instead of parsing the path.

For the `sagemaker_launcher`, both arguments would probably just be ignored if set, since sagemaker's [`entry_point`](https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/sagemaker.pytorch.html) is required to be a python script, as far as I understand (I haven't used sagemaker before).

sgugger:
[...]

PR #258, opened 2022-02-23
Add launch flags --module and --no_python (#256)

parameter-concern:
A few notes:

- I have tested the various configurations of `simple_launcher()` locally, but not the others.
- I did not add any pytest tests.
- There is a bit of repetition in the spots that check for when both flags are set together. This could be done in the sanity check section before a specific launcher is called, or it could be omitted entirely since the error would hopefully be obvious, but I thought the best solution was to just do it where the flags are used.
- I ran `black` but it changed a lot of unrelated code, so I undid that.
- It's kind of a niche feature so I didn't bother adding anything to the .rst docs.

I'm happy to take feedback on these notes or any other comments.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -204,9 +204,9 @@ def multi_gpu_launcher(args):
 
     if args.module and args.no_python:
         raise ValueError("--module and --no_python cannot be used together")
-    if args.module:
+    elif args.module:
         cmd.append("--module")
-    if args.no_python:
+    elif args.no_python:
         cmd.append("--no_python")
     cmd.append(args.training_script)
     cmd.extend(args.training_script_args)
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -251,9 +251,9 @@ def deepspeed_launcher(args):
 
     if args.module and args.no_python:
         raise ValueError("--module and --no_python cannot be used together")
-    if args.module:
+    elif args.module:
         cmd.append("--module")
-    if args.no_python:
+    elif args.no_python:
         cmd.append("--no_python")
     cmd.append(args.training_script)
     cmd.extend(args.training_script_args)
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -93,6 +93,17 @@ def launch_command_parser(subparsers=None):
         help="The name of the main function to be executed in your script (only for TPU training).",
     )
     parser.add_argument(
+        "-m",
+        "--module",
+        action="store_true",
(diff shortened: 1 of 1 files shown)


src/accelerate/commands/launch.py
```diff
@@ -140,7 +151,14 @@ def launch_command_parser(subparsers=None):
...
 
 def simple_launcher(args):
-    cmd = [sys.executable, args.training_script]
+    cmd = []
+    if args.no_python and args.module:
+        raise ValueError("--module and --no_python cannot be used together")
+    if not args.no_python:
+        cmd.append(sys.executable)
+        if args.module:
+            cmd.append("-m")
```
sgugger, line 160: Since both can't be true at the same time, the test should not be nested.
```suggestion
    elif args.module:
        cmd.append("-m")
```

src/accelerate/commands/launch.py
```diff
@@ -184,6 +201,13 @@ def multi_gpu_launcher(args):
         cmd.extend(["--nproc_per_node", str(args.num_processes)])
         if args.main_process_port is not None:
             cmd.extend(["--master_port", str(args.main_process_port)])
+
+    if args.module and args.no_python:
+        raise ValueError("--module and --no_python cannot be used together")
+    if args.module:
+        cmd.append("--module")
+    if args.no_python:
```
sgugger, line 209:
```suggestion
    elif args.module:
        cmd.append("--module")
    elif args.no_python:
```

src/accelerate/commands/launch.py
```diff
@@ -226,6 +249,12 @@ def deepspeed_launcher(args):
     else:
         cmd.extend(["--num_gpus", str(args.num_processes)])
 
+    if args.module and args.no_python:
+        raise ValueError("--module and --no_python cannot be used together")
+    if args.module:
+        cmd.append("--module")
+    if args.no_python:
```
sgugger, line 256:
```suggestion
    if args.module and args.no_python:
        raise ValueError("--module and --no_python cannot be used together")
    elif args.module:
        cmd.append("--module")
    elif args.no_python:
```

sgugger (approved):
Thanks for working on this!
There is some formatting issue. Could you make sure you have the right version of black installed with `pip install -e .[quality]` before running `make style`?

src/accelerate/commands/launch.py
parameter-concern: Good catch, thanks. I think after the un-indent, it still has to stay an `if` and not an `elif`, otherwise it won't get evaluated because the `if not args.no_python` above would always be evaluated `True` first.

```suggestion
    if args.module:
        cmd.append("-m")
```

sgugger:
Thanks again for working on this!

PR #259, opened 2022-02-24
Add  `debug_launcher`

sgugger:
This PR adds support for a `debug_launcher` will allows the user to launch a function in a distributed fashion on CPU for quick debugging. It will also let us test more extensively the library in the CI.

More tests will follow in subsequent PRs :-)


LysandreJik (approved):
Very clean, love the `patch_environment` approach!

PR #263, opened 2022-03-01
Add a flag to use CPU only in the config

### LINKED ISSUE #261: How to force `accelerate launch` to use CPU instead of GPU?
Config `accelerate` to use CPU:

```sh
$ accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 1 
How many different machines will you use (use more than 1 for multi-node training)? [1]: 1
How many processes in total will you use? [1]: 1
Do you wish to use FP16 (mixed precision)? [yes/NO]:
```

Test script:

```python
accelerator = Accelerator()
print(accelerator.device)
```

Run:

```sh
$ accelerate launch main.py
cuda  # expected cpu
```

I know that there is an option `accelerator = Accelerator(cpu=True)`, but I would like to get this behavior using command line.

sgugger:
This PR adds a field in the config to only use CPU even if GPUs are available. It also simplifies the state initialization using that new flag.

Fixes #261

### DIFF
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -52,6 +52,17 @@ def get_cluster_input():
                 lambda x: int(x),
             )
 
+    if distributed_type == DistributedType.NO:
+        use_cpu = _ask_field(
+            "Do you want to run your training on CPU only (even if a GPU is available)? [no]:",
+            lambda x: bool(x),
+            default=False,
+        )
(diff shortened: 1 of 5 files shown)


LysandreJik (approved):
Very cool!

muellerzr (approved):
Agreed, very nice and clean!

PR #264, opened 2022-03-01
Basic fixes for DeepSpeed

sgugger:
This PR includes basic fixes to the DeepSpeed integration to make sure the base `accelerate test` passes.

The main issue is that the primitive such as gather/broadcast were behaving in DeepSpeed like in single-process mode.


LysandreJik (approved):
Nice :)

Issue #236, opened 2022-01-21
different seed for each device


ZhiyuanChen:
Current `set_seed` implementation set the same seed for every devices, which could results the same data agumentation applied for each data across the process and damage model performance. Consider set different random seed for each device, such as:
```python
    def init_seed(self):
        """
        Set up random seed
        """
        torch.manual_seed(self.rank + self.seed)
        torch.cuda.manual_seed(self.rank + self.seed)
        np.random.seed(self.rank + self.seed)
        random.seed(self.rank + self.seed)
```

sgugger:
This is just a small utility function, setting the same or a different seed for each process is the responsibility of the user. You can call it in your script with `set_seed(seed + accelerator.process_index)` for instance.

But I'll see if we can add a boolean argument to do what you suggest!

PR #266, opened 2022-03-02
Ability to set the seed with randomness from inside Accelerate

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/236

### DIFF
--- a/src/accelerate/utils.py
+++ b/src/accelerate/utils.py
@@ -69,7 +69,7 @@ def set_seed(seed: int, device_specific: bool = False):
 
     Args:
         seed (:obj:`int`): The seed to set.
-        device_specific (:obj:`bool`): Whether to differ the seed on each device slightly with `self.process_index`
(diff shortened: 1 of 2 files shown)


src/accelerate/utils.py
```diff
@@ -63,13 +63,16 @@ class TensorInformation:
...
 
 
-def set_seed(seed: int):
+def set_seed(seed: int, device_specific: bool = False):
     """
     Helper function for reproducible behavior to set the seed in ``random``, ``numpy``, ``torch``.
 
     Args:
         seed (:obj:`int`): The seed to set.
+        device_specific (:obj:`bool`): Whether to differ the seed on each device slightly with `self.process_index`
```
sgugger, line 72:
```suggestion
        device_specific (:obj:`bool`, `optional`, defaults to :obj:`False`): Whether to differ the seed on each device slightly with `self.process_index`
```

sgugger (approved):
Thanks a lot for adding this!

PR #269, opened 2022-03-07
Don't use dispatch_batches when torch is < 1.8.0

sgugger:
`torch.distributed` does not have `broadcast_object_list` until v1.8.0 and above, so we can't use the `DispatchDataLoader` for versions smaller than this. This PR adapts the defaults and raises proper ImportError.

This was reported on the [forums](https://discuss.huggingface.co/t/why-my-accelerate-just-doesnt-work/15409)


muellerzr (approved):
Very clean! TIL about both `is True` vs `== True` and `version.parse` and the magic it can pull. Nice!

PR #270, opened 2022-03-07
Implementation of saving and loading custom states

muellerzr:
Introduces a `register_for_checkpointing` function, as well as two save/loading functions for those registered objects. 

Would like an initial review for if I'm on the right track 😄

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -33,10 +33,10 @@ from .utils import (
     convert_outputs_to_fp32,
     extract_model_from_parallel,
     gather,
+    get_pretty_name,
     pad_across_processes,
     save,
     wait_for_everyone,
-    get_pretty_name
 )
 
 
@@ -664,19 +664,19 @@ class Accelerator:
         """
         Makes note of `objects` and will save or load them in during `save_state` or `load_state`.
 
-        These should be utilized when the state is being loaded or saved in the same script. It is not designed
-        to be used in different scripts
+        These should be utilized when the state is being loaded or saved in the same script. It is not designed to be
+        used in different scripts
 
         Note: Every `object` must have a `load_state_dict` and `state_dict` function to be stored.
         """
         invalid_objects = []
         for obj in objects:
-            if not hasattr(obj, 'state_dict') or not hasattr(obj, 'load_state_dict'):
+            if not hasattr(obj, "state_dict") or not hasattr(obj, "load_state_dict"):
                 invalid_objects.append(obj)
         if len(invalid_objects) > 0:
             err = "All `objects` must include a `state_dict` and `load_state_dict` function to be stored. The following inputs are invalid:"
             for index, obj in enumerate(invalid_objects):
-                err += f'\n\t- Item at index {index}, `{get_pretty_name(obj)}`'
+                err += f"\n\t- Item at index {index}, `{get_pretty_name(obj)}`"
             raise ValueError(err)
         self._custom_objects.extend(objects)
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -664,8 +664,8 @@ class Accelerator:
         """
         Makes note of `objects` and will save or load them in during `save_state` or `load_state`.
 
-        Note: These should be utilized when the state is being loaded or saved in the same script. It
-        is not designed to be used in different scripts
+        Note: These should be utilized when the state is being loaded or saved in the same script. It is not designed
+        to be used in different scripts
         """
         self._custom_objects = list(objects)
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -22,7 +22,7 @@ import torch
 
 from packaging import version
 
-from .checkpointing import load_accelerator_state, save_accelerator_state
+from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
 from .data_loader import prepare_data_loader
 from .kwargs_handlers import DistributedDataParallelKwargs, GradScalerKwargs, InitProcessGroupKwargs, KwargsHandler
 from .optimizer import AcceleratedOptimizer
@@ -188,6 +188,7 @@ class Accelerator:
         # Internal references to the training objects
         self._optimizers = []
         self._models = []
+        self._custom_objects = []
 
         # RNG Types
         self.rng_types = rng_types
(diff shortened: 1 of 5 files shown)


src/accelerate/accelerator.py
```diff
@@ -591,6 +597,14 @@ def load_state(self, input_dir: str):
             raise ValueError(f"Tried to find {input_dir} but folder does not exist")
         logger.info(f"Loading states from {input_dir}")
         load_accelerator_state(input_dir, self._models, self._optimizers, self.state.process_index, self.scaler)
+        custom_checkpoints = [f for f in os.listdir(input_dir) if "custom_checkpoint" in f]
+        if len(custom_checkpoints) != (len(self._custom_objects) + 1):
```
sgugger, line 601: Why the +1 here?

src/accelerate/accelerator.py
```diff
@@ -646,6 +660,15 @@ def get_state_dict(self, model):
...
         return state_dict
 
+    def register_for_checkpointing(self, *objects):
+        """
+        Makes note of `objects` and will save or load them in during `save_state` or `load_state`.
+
+        Note: These should be utilized when the state is being loaded or saved in the same script. It is not designed
+        to be used in different scripts
+        """
+        self._custom_objects = list(objects)
```
sgugger, line 670: I would append them to the existing `_custom_objects`.

src/accelerate/checkpointing.py
```diff
@@ -132,3 +132,27 @@ def load_accelerator_state(input_dir, models, optimizers, process_index, scaler=
...
         xm.set_rng_state(states["xm_seed"])
     logger.info("All random states loaded successfully")
+
+
+def save_custom_state(obj, path, index: int = 0):
+    """
+    Saves the state of `obj` to `{path}/custom_checkpoint_{index}.pkl`
+    """
+    # Should this be the right way to get a qual_name type value from `obj`?
+    save_location = f"{path}/custom_checkpoint_{index}.pkl"
```
sgugger, line 142: You should use a Path or `os.path.join` here otherwise it won't work in Windows.

src/accelerate/checkpointing.py
```diff
@@ -132,3 +132,27 @@ def load_accelerator_state(input_dir, models, optimizers, process_index, scaler=
...
+    torch.save(obj, save_location)
+
+
+def load_custom_state(obj, path, index: int = 0):
+    """
+    Loads the state of `obj` at `{path}/custom_checkpoint_{index}.pkl`
+    """
+    load_location = f"{path}/custom_checkpoint_{index}.pkl"
+    if hasattr(obj, "load_state_dict"):
+        obj = obj.load_state_dict(load_location)
```
sgugger, line 155: We don't need a torch.load inside here?

sgugger (commented):
Looking good for now!

src/accelerate/accelerator.py
muellerzr: For the custom objects + the saved objected was my thought at the time. But you're write its a bug

src/accelerate/accelerator.py
```diff
@@ -591,6 +598,13 @@ def load_state(self, input_dir: str):
             raise ValueError(f"Tried to find {input_dir} but folder does not exist")
         logger.info(f"Loading states from {input_dir}")
         load_accelerator_state(input_dir, self._models, self._optimizers, self.state.process_index, self.scaler)
+        custom_checkpoints = [f for f in os.listdir(input_dir) if "custom_checkpoint" in f]
+        if len(custom_checkpoints) != len(self._custom_objects):
+            logger.warn("Warning! You are loading a state that does not include registered objects.")
```
sgugger, line 603: The warning message does not match what the test is testing :-)

src/accelerate/accelerator.py
```diff
@@ -646,6 +660,26 @@ def get_state_dict(self, model):
...
         return state_dict
 
+    def register_for_checkpointing(self, *objects):
+        """
+        Makes note of `objects` and will save or load them in during `save_state` or `load_state`.
+
+        These should be utilized when the state is being loaded or saved in the same script. It is not designed to be
+        used in different scripts
+
+        Note: Every `object` must have a `load_state_dict` and `state_dict` function to be stored.
```
sgugger, line 670: This is not a proper docstring **but** it will be once #271 is merged, so giving you a free pass ;-)

src/accelerate/checkpointing.py
```diff
@@ -132,3 +133,23 @@ def load_accelerator_state(input_dir, models, optimizers, process_index, scaler=
...
+
+def save_custom_state(obj, path, index: int = 0):
+    """
+    Saves the state of `obj` to `{path}/custom_checkpoint_{index}.pkl`
+    """
+    # Should this be the right way to get a qual_name type value from `obj`?
+    save_location = Path(path) / f"custom_checkpoint_{index}.pkl"
+    logger.info(f"Saving the state of {get_pretty_name(obj)} to {save_location}")
+    obj = obj.state_dict()
+    torch.save(obj, save_location)
```
sgugger, line 146:
```suggestion
    torch.save(obj.state_dict(), save_location)
```
Can be combined now.

tests/test_state_checkpointing.py
```diff
@@ -123,3 +125,43 @@ def test_can_resume_training(self):
             self.assertEqual(b1, b3)
             self.assertEqual(opt_state1, opt_state3)
             self.assertEqual(ground_truth_rands, test_rands)
+
+
+class CustomItemsTest(unittest.TestCase):
```
sgugger, line 130: Not sure we need a new class for this. Can be included in the current one!

sgugger (approved):
Thanks for working on this!

src/accelerate/accelerator.py
```diff
@@ -600,7 +600,9 @@ def load_state(self, input_dir: str):
         load_accelerator_state(input_dir, self._models, self._optimizers, self.state.process_index, self.scaler)
         custom_checkpoints = [f for f in os.listdir(input_dir) if "custom_checkpoint" in f]
         if len(custom_checkpoints) != len(self._custom_objects):
-            logger.warn("Warning! You are loading a state that does not include registered objects.")
+            logger.warn(
+                "Warning! Number of found checkpoints does not match the number of registered objects. Skipping."
```
sgugger, line 604: Can you print the two numbers?

muellerzr: How does 
```python
            err = "Warning! Number of found checkpoints does not match the number of registered objects:"
            err += f"\n\tFound checkpoints: {len(custom_checkpoints)}"
            err += f"\n\tRegistered objects: {len(self._custom_objects)}\nSkipping."
```
Sound?

sgugger (approved):
Great warning!

Issue #277, opened 2022-03-11
Tool to get current installed versions


LysandreJik:
A tool to get current installed versions helps when opening issues to provide all software versions in the current environment.

This helps to debug issues easier or to catch versions mismatches straight away. A similar tool has been implemented for transformers [here](https://github.com/huggingface/transformers/blob/master/src/transformers/commands/env.py).

Accelerate would benefit from such a tool, which could be as simple as a runtime method such as `get_versions()` as proposed in https://github.com/huggingface/accelerate/pull/254.

johnnv1:
Hi, i  would like to work on this

LysandreJik:
Thanks for your contribution @johnnv1! Closing this issue.

PR #280, opened 2022-03-12
add env command

johnnv1:
I have added an `env` command as suggested at #277 

example of use:
```console
accelerate env
```
outputs

<details>
<summary>output</summary>

- `accelerate` version: 0.6.0.dev0
- Platform: Linux-5.13.0-35-generic-x86_64-with-Ubuntu-20.04-focal
- Python version: 3.6.15
- Numpy version: 1.19.5
- PyTorch version (GPU?): 1.10.2+cu102 (False)
</details>


src/accelerate/commands/env.py
```diff
@@ -0,0 +1,51 @@
+import argparse
+import platform
+
+import numpy as np
+
+from accelerate import __version__ as version
+from accelerate.file_utils import is_torch_available
```
sgugger, line 7: Note that the command submodules will import `accelerate.state` (via the lancuh command) so the code will fail if PyTorch is not installed anyway (which shouldn't be the case since it's a dependency).

I'd remove that import and the test below that uses it.

src/accelerate/commands/env.py
```diff
@@ -0,0 +1,51 @@
...
+        pt_version = torch.__version__
+        pt_cuda_available = torch.cuda.is_available()
+
+    info = {
+        "`accelerate` version": version,
+        "Platform": platform.platform(),
+        "Python version": platform.python_version(),
+        "Numpy version": np.__version__,
+        "PyTorch version (GPU?)": f"{pt_version} ({pt_cuda_available})",
+    }
```
sgugger, line 36: Maybe we could add the content of the default config to this info? It can be done in a followup PR if you don't have time to add it right now.

src/accelerate/file_utils.py
```diff
@@ -0,0 +1,19 @@
+import importlib.util
```
sgugger, line 1: As discussed above, not need to add that new file.

src/accelerate/utils.py
```diff
@@ -30,6 +31,14 @@
 from .state import AcceleratorState, DistributedType, is_deepspeed_available, is_tpu_available
 
 
+# The package importlib_metadata is in a different place, depending on the python version.
+if sys.version_info < (3, 8):
+    import importlib_metadata  # noqa: F401
+
+else:
+    import importlib.metadata as importlib_metadata  # noqa: F401
```
sgugger, line 39: This shouldn't be necessary either

sgugger (commented):
Thanks a lot for your PR! Contrarily to Transformers, `torch` is a core dependency of Accelerate, so we don't need to protect the torch import in a `is_torch_available` test.

Other than that, PR looks pretty good!

src/accelerate/commands/env.py
johnnv1: I don't understand, default config for what?

sgugger: The default config of Accelerate in `default_config.yaml`

johnnv1: Okay, got it, I'll add this

sgugger (approved):
Great work, thanks again!

PR #282, opened 2022-03-14
replace texts and link (master -> main)

johnnv1:
I don't change at docs files, but also have old references to `master` branch at `docs/source/installation.mdx`. I should also update the docs?


sgugger (approved):
Yes, any reference in the MDX files of the doc would need updating as well.
Thanks a lot for working on this!

PR #289, opened 2022-03-22
Added the code snippet for saving the model

MATRIX4284:


examples/nlp_example.py
```diff
@@ -142,6 +142,10 @@ def collate_fn(examples):
                 optimizer.step()
                 lr_scheduler.step()
                 optimizer.zero_grad()
+        
+        accelerator.wait_for_everyone()
+        unwrapped_model = accelerator.unwrap_model(model)
+        unwrapped_model.save_pretrained('accl_out', save_function=accelerator.save, state_dict=accelerator.get_state_dict(model))
```
sgugger, line 148: This should be an argument of the script, and we should just save if it is set.

sgugger (approved):
Thanks for adding this! Let's make it a bit more flexible by adding an argument to the script for the output folder (and if not set, we don't save).

examples/nlp_example.py
MATRIX4284: Sure tha

MATRIX4284: Do you want me to add the argument as a part of this pr or the next

sgugger: In this PR please :-)

PR #290, opened 2022-03-22
Document save/load state

muellerzr:
Adds documentation to the new checkpointing capabilities

### DIFF
--- a/docs/source/checkpoint.mdx
+++ b/docs/source/checkpoint.mdx
@@ -1,0 +1,25 @@
+<!--Copyright 2022 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+-->
+
+# Checkpointing
+
+When training a PyTorch model with Accelerate, you may often want to save and continue a state of training. Doing so requires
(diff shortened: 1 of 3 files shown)


docs/source/checkpoint.mdx
```diff
@@ -0,0 +1,25 @@
...
+-->
+
+# Checkpointing
+
+When training a PyTorch model with Accelerate, you may often want to save and continue a state of training. Doing so requires
+saving and loading the model, optimizer, RNG generators, and the GradScaler. Inside Accelerate are two convience functions to achieve this quickly:
+- Use [`~Accelerator.save_state`] for saving everything mentioned above to a folder location
+- Use [`~Accelerator.load_state`] for loading everything stored from an earlier `save_state`
+
+It should be noted that these states _must_ be from the same contiguous training script, they should not be from two seperate runs or scripts.
```
sgugger, line 20:
```suggestion
It should be noted that the expectation is that those states come from the same training script, they should not be from two separate scripts.
```

docs/source/checkpoint.mdx
```diff
@@ -0,0 +1,25 @@
...
+saving and loading the model, optimizer, RNG generators, and the GradScaler. Inside Accelerate are two convience functions to achieve this quickly:
+- Use [`~Accelerator.save_state`] for saving everything mentioned above to a folder location
+- Use [`~Accelerator.load_state`] for loading everything stored from an earlier `save_state`
+
+It should be noted that these states _must_ be from the same contiguous training script, they should not be from two seperate runs or scripts. 
+
+- By using [`~Accelerator.register_for_checkpointing`], you can register custom objects to be automatically stored or loaded from the two prior functions,
+so long as the object has a `state_dict` **and** a `load_state_dict` functionality. This could include objects such as a learning rate scheduler. 
+
+[[autodoc]] Checkpointing
```
sgugger, line 25: There is no object in the lib called `Checkpointing`, so this won't work. this is where you should show a full example.

sgugger (commented):
Thanks for working on this! The checkpoint page could be expanded a little ;-)

sgugger (approved):
Thanks for adding an example!

PR #292, opened 2022-03-24
Refactor precisions to its own enum

muellerzr:
Seperates out the precision types to its own enum for auto-complete and code simplicity

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -27,7 +27,7 @@ from typing import Dict, List
 from accelerate.commands.config import default_config_file, load_config_from_file
 from accelerate.commands.config.config_args import SageMakerConfig
 from accelerate.state import ComputeEnvironment, DistributedType
-from accelerate.utils import PrepareForLaunch, is_sagemaker_available
+from accelerate.utils import PrecisionType, PrepareForLaunch, is_sagemaker_available
 
 
 def launch_command_parser(subparsers=None):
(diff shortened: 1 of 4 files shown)


src/accelerate/commands/launch.py
```diff
@@ -58,7 +58,7 @@ def launch_command_parser(subparsers=None):
         "--mixed_precision",
         default="no",
         type=str,
-        choices=["no", "fp16", "bf16"],
+        choices=PrecisionType,
```
sgugger, line 61: Here I think you might need to convert this in a list for `choices` to work the proper way.

src/accelerate/utils.py
```diff
@@ -51,7 +51,26 @@ def is_sagemaker_available():
 OPTIMIZER_NAME = "optimizer"
 
 
-class RNGType(Enum):
+class MetaEnum(EnumMeta):
+    def __contains__(cls, item):
```
sgugger, line 55: A small docstring here on what this adds to a regular `Enum` would be great :-)

sgugger (approved):
Thanks for cleaning this!

src/accelerate/utils.py
```diff
@@ -51,7 +51,36 @@ def is_sagemaker_available():
 OPTIMIZER_NAME = "optimizer"
 
 
-class RNGType(Enum):
+class MetaEnum(EnumMeta):
```
sgugger, line 54:
```suggestion
class EnumWithContains(EnumMeta):
```
nit: the name is not very descriptive on what the metaclass does, would switch to this.

src/accelerate/utils.py
```diff
@@ -51,7 +51,36 @@ def is_sagemaker_available():
...
+
+    def __contains__(cls, item):
+        try:
+            cls(item)
+        except ValueError:
+            return False
+        return True
+
+
+class BaseEnum(Enum, metaclass=MetaEnum):
```
sgugger, line 65:
```suggestion
class BaseEnum(Enum, metaclass=EnumWithContains):
```
To go with the suggestion above.

sgugger (approved):

Issue #291, opened 2022-03-23
Add logging capability


muellerzr:
# Feature Request: Logging Capability

## Feature Summary
- Add a method called `Accelerator.log` that takes in dictionary-like values and saves to configured logger
- `Accelerator.__init__` should get a new param of `log_with` that takes a list of strings as potential options or `"all"`
- The default should be `None` and no loggers are configured. If `"all"` is selected, find all installed loggers (similar to `Trainer`)

## Timeline
Supported integrations:
- [x] TensorBoard/X
- [x] WandB
- [x] CometML
- [ ] NeptuneAI
﻿

ashutoshsaboo:
Is it possible to add support for custom loggers too? Like the open-source experiment tracker Aim (others above other than tboard are closed source): https://github.com/aimhubio/aim .

muellerzr:
@ashutoshsaboo essentially this will just be a small convenience wrapper to a few back-ends. For any and all loggers you can simply have them log below the following if statement and it should run without issue:
```python
if accelerator.is_main_process:
  # Do my logging of everything here
```

ashutoshsaboo:
That's fine, Would I still be able to use Accelerator.log? I was thinking more in the sense of having a callback interface (like Trainer implements) which Accelerate could also do, and can be extended easily? This helps various lib owners provide their own custom impls that can be used out of the box. @muellerzr

muellerzr:
Putting Neptune.AI off the roadmap for now, given that their API is much more user-specific and needs specific key namings when logging items, rather than being generalizable.

They should follow the `if accelerator.is_main_process:` and do the logging themselves.

congchan:
Hi, what parameters should I use to enable tensorboard and disable wandb, when using `accelerate launch --config_file`

PR #293, opened 2022-03-24
Add logging capabilities

muellerzr:
Start of an implementation for a convinence API when logging with Accelerate. I've opted to use abstract functions here in the base class so that if people wish to integrate with their own API's and have it be compatible, an error can be raised if they missed an implementation. Let me know what you think so far @sgugger

Closes https://github.com/huggingface/accelerate/issues/291


src/accelerate/accelerator.py
```diff
@@ -17,10 +17,12 @@
 import sys
 import warnings
 from contextlib import contextmanager
+from pathlib import Path
 from typing import List, Optional, Union
 
 import torch
 
+from accelerate.accelerate.src.accelerate.tracking import TensorBoardTracker
```
sgugger, line 25:
```suggestion
from .tracking import TensorBoardTracker
```
:scream:

src/accelerate/accelerator.py
```diff
@@ -111,9 +115,17 @@ def __init__(
         cpu: bool = False,
         deepspeed_plugin: DeepSpeedPlugin = None,
         rng_types: Optional[List[Union[str, RNGType]]] = None,
+        log_with: Optional[List[Union[LoggerType, str]]] = None,
         dispatch_batches: Optional[bool] = None,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
+        if log_with is not None:
+            if not isinstance(log_with, list):
```
sgugger, line 123: The `"all"` should be dealt with around here too.

src/accelerate/accelerator.py
```diff
@@ -556,6 +568,30 @@ def wait_for_everyone(self):
...
+            project_name (`str`):
+        """
+        project_location = Path(project_name)
+        self.trackers = []
+        for tracker in self.log_with:
+            if tracker.lower() == "tensorboard" and is_tensorboard_available():
+                self.trackers.append(TensorBoardTracker(project_location))
+        if config is not None:
+            for tracker in self.trackers:
+                tracker.store_init_configuration(config)
```
sgugger, line 586: Nice!

src/accelerate/tracking.py
```diff
@@ -0,0 +1,72 @@
...
+# limitations under the License.
+
+# Expectation:
+# Provide a project dir name, then each type of logger gets stored in project/{`logging_dir`}
+
+import logging
+from abc import ABCMeta, abstractmethod
+from pathlib import Path
+
+from torch.utils import tensorboard
```
sgugger, line 22: Should be inside an `is_tensorboard_available` flag.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,72 @@
...
+        logger.info(
+            "Make sure to log any initial configurations with `self.store_init_configuration` before training!"
+        )
+
+    def store_init_configuration(self, values):
+        self.writer.add_hparams(values, metric_dict={})
+        self.writer.flush()
+        logger.info("Stored initial configuration hyperparameters to TensorBoard")
+
+    # add_scalar has the option for `global_step`, should we include this?
```
sgugger, line 60: Accelerate has no idea about the global step though. We could add it in an optional argument to `log`? Or add something that keeps track of the number of time the `optimizer.step` is called internally.

sgugger (approved):
Nice API! Thanks a lot for working on this!

src/accelerate/tracking.py
muellerzr: That's more or less what my thinking is, having it be an optional argument. I don't think there's a need for us to keep track of `optimizer.step` calls internally for this yet, but we'll see what changes as this develops

src/accelerate/accelerator.py
muellerzr: *That is what we call VSCode not being friendly and me missing it!*

src/accelerate/tracking.py
muellerzr: Decided to make it an optional argument. If folks seem to find it annoying then we can go ahead and track `optimizer.step` next

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+
+    def store_init_configuration(self, values: dict):
+        """
+        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment.
+
+        Args:
+            values (`dict`):
+                Values to be stored as initial hyperparameters as key-value pairs. Value be of type `bool`, `str`,
+                `float`, `int`, or `None`.
```
sgugger, line 94:
```suggestion
            values (Dictionary `str` to `bool`, `str`, `float` or `int`):
                Values to be stored as initial hyperparameters as key-value pairs. The values need to have type `bool`, `str`,
                `float`, `int`, or `None`.
```

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+    def log(self, values, step: Optional[int]):
+        pass
+
+
+class TensorBoardTracker(GeneralTracker):
+    """
+    A `Tracker` class that supports `tensorboard`. Should be initialized at the start of your script.
+
+    Args:
+        run_name (`str`):
```
sgugger, line 73: Seems to be optional and has a default ;-)

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+        logger.info("Stored initial configuration hyperparameters to TensorBoard")
+
+    def log(self, values: dict, step: Optional[int] = None):
+        """
+        Logs `values` to the current run.
+
+        Args:
+            values (`dict`):
+                Values to be logged as key-value pairs. Value be of type `int`, `float`, or `str`.
+            step (`int`, Optional):
```
sgugger, line 107:
```suggestion
            step (`int`, *optional*):
```

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+        self.writer.flush()
+        logger.info("Successfully logged to TensorBoard")
+
+
+class WandBTracker(GeneralTracker):
+    """
+    A `Tracker` class that supports `wandb`. Should be initialized at the start of your script.
+
+    Args:
+        run_name (`str`):
```
sgugger, line 124: Same comment as above for the `run_name` of Tensorboard.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+        logger.info(
+            "Make sure to log any initial configurations with `self.store_init_configuration` before training!"
+        )
+
+    def store_init_configuration(self, values: dict):
+        """
+        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment.
+
+        Args:
+            values (`dict`):
```
sgugger, line 143: Same comments as above.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def log(self, values: dict, step: Optional[int] = None):
+        """
+        Logs `values` to the current run.
+
+        Args:
+            values (`dict`):
+                Values to be logged as key-value pairs.
+            step (`int`, Optional):
+                The run step. If included, the log will be affiliated with this step.
```
sgugger, line 158: And here too ;-)

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        wandb.log(values, step=step)
+        logger.info("Successfully logged to WandB")
+
+
+class CometMLTracker(GeneralTracker):
+    """
```
sgugger, line 165: Same comments on the docstrings.

sgugger (approved):
Nice! We just need a bit more polishing for the docstrings, as well as a nice new doc page about logging ;-)

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+    Args:
+        run_name (`str`):
+            The name of the experiment run. Logs are then stored in `tensorboard/{run_name}`
+    """
+
+    log_directory = Path("tensorboard")
+
+    def __init__(self, run_name: str):
+        self.run_name = run_name
+        self.writer = tensorboard.SummaryWriter(self.logging_dir / self.run_name)
```
sgugger, line 89: The logging_dir should also be something that could be customized and default to the usual PyTorch default.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+class TensorBoardTracker(GeneralTracker):
+    """
+    A `Tracker` class that supports `tensorboard`. Should be initialized at the start of your script.
+
+    Args:
+        run_name (`str`):
+            The name of the experiment run. Logs are then stored in `tensorboard/{run_name}`
+    """
+
+    log_directory = Path("tensorboard")
```
sgugger, line 85: This variable doesn't seem used.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+class WandBTracker(GeneralTracker):
+    """
+    A `Tracker` class that supports `wandb`. Should be initialized at the start of your script.
+
+    Args:
+        run_name (`str`):
+            The name of the experiment run.
+    """
+
+    log_directory = None
```
sgugger, line 136: This doesn't seem used either.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+    A `Tracker` class that supports `comet_ml`. Should be initialized at the start of your script.
+
+    API keys must be stored in a Comet config file.
+
+    Args:
+        run_name (`str`):
+            The name of the experiment run.
+    """
+
+    log_directory = None
```
sgugger, line 183: Same here.

src/accelerate/tracking.py
```diff
@@ -0,0 +1,210 @@
...
+    "Returns a list of all supported available trackers in the system"
+    return _available_trackers
+
+
+class GeneralTracker(object, metaclass=ABCMeta):
+    """
+    A base Tracker class to be used for all logging integration implementations.
+    """
+
+    log_directory: str
```
sgugger, line 57: We should remove this from the abtract class as it's not required for all of the trackers.

muellerzr:
There's a weird issue when doing `pip install tensorflow` where it will just install a higher `setuptools`, and that setuptools will break importing pytorch. pinning it in the dev requirements "fixes" it. Need to play with it more, but a test is there. Once that's solved we probably should mimic something like this for the TF tests as well, since this goes through and looks at the actual values stored

setup.py
```diff
@@ -21,6 +21,8 @@
 extras["test"] = [
     "pytest",
     "pytest-xdist",
+    "tensorflow",
```
sgugger, line 24: Not following why we have to add TensoFlow here. Can't it be just `tensorboard`?

tests/test_tracking.py
```diff
@@ -0,0 +1,89 @@
...
+
+class TensorBoardTrackingTest(unittest.TestCase):
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree("test_project_with_config")
+        shutil.rmtree("test_project_with_log")
+
+    def test_init_trackers(self):
+        hps = None
+        project_name = "test_project_with_config"
```
sgugger, line 39: It would probably be best to use a temporary directory here, to avoid users running the tests locally getting this folder.

tests/test_tracking.py
```diff
@@ -0,0 +1,89 @@
...
+        keys.sort()
+        self.assertEqual(keys, ["learning_rate", "num_iterations", "some_boolean", "some_string"])
+        self.assertEqual(hps["num_iterations"].number_value, 12)
+        self.assertEqual(hps["learning_rate"].number_value, 0.01)
+        self.assertEqual(hps["some_boolean"].bool_value, False)
+        self.assertEqual(hps["some_string"].string_value, "some_value")
+
+    def test_log(self):
+        step = None
+        project_name = "test_project_with_log"
```
sgugger, line 67: Same here.

tests/test_tracking.py
```diff
@@ -0,0 +1,89 @@
...
+        accelerator = Accelerator(log_with="tensorboard")
+        accelerator.init_trackers(project_name)
+        values = {"total_loss": 0.1, "iteration": 1, "my_text": "some_value"}
+        accelerator.log(values, step=0)
+        # Logged values are stored in the outermost-tfevents file and can be read in as a TFRecord
+        # Names are randomly generated each time
+        log = list(filter(lambda x: x.is_file(), Path(project_name).iterdir()))[0]
+        serialized_examples = tf.data.TFRecordDataset(log)
```
sgugger, line 75: Is it why the TF dep is necessary? There is really no other way to load a TensorBoard log?

sgugger (commented):
Thanks for adding the tests!

tests/test_tracking.py
muellerzr: Yup, sadly: https://github.com/tensorflow/tensorboard#how-can-i-export-data-from-tensorboard

We have to use `tf` to parse it back in.

src/accelerate/tracking.py
```diff
@@ -80,6 +80,16 @@ def log(self, values: dict, step: Optional[int]):
         """
         pass
 
+    @abstractmethod
+    def finish(self):
+        """
+        Should run any finalizing functions within the tracking API. If the API should not have one, just return:
+        ```python
+        super().finish()
+        ```
```
sgugger, line 89:
```suggestion
        Should run any finalizing functions within the tracking API. If the API should not have one, just don't overwrite that method.
```

src/accelerate/tracking.py
```diff
@@ -218,3 +242,7 @@ def log(self, values: dict, step: Optional[int] = None):
             self.writer.set_step(step)
         self.writer.log_others(values)
         logger.info("Successfully logged to CometML")
+
+    def finish(self):
+        """Do nothing"""
+        super().finish()
```
sgugger, line 248:
```suggestion
```
No need to write something as calling the super method is then the default ;-)

tests/test_tracking.py
```diff
@@ -29,29 +32,29 @@
...
-        # Pull HPS through careful parsing
-        for event in si:
-            for value in event.summary.value:
-                proto_bytes = value.metadata.plugin_data.content
-                plugin_data = plugin_data_pb2.HParamsPluginData.FromString(proto_bytes)
-                if plugin_data.HasField("session_start_info"):
-                    hps = dict(plugin_data.session_start_info.hparams)
+        with tempfile.TemporaryDirectory() as dirpath:
+            oldpwd = os.getcwd()
+            os.chdir(dirpath)
```
sgugger, line 40: I'd like to avoid this. The API should accept any logging dir for TensorBoard at least, no?

docs/source/tracking.mdx
```diff
@@ -0,0 +1,137 @@
...
+        my_optimizer.zero_grad()
+        inputs, targets = batch
+        inputs = inputs.to(device)
+        targets = targets.to(device)
+        outputs = my_model(inputs)
+        loss = my_loss_function(outputs, targets)
+        accelerator.backward(loss)
+        my_optimizer.step()
+        accelerator.log({"training_loss": loss}, step=step)
+```
```
sgugger, line 79: We should mention the train end method in this example, since it's relevant for trackers.

setup.py
```diff
@@ -20,9 +20,11 @@
 extras["docs"] = []
 extras["test"] = [
     "pytest",
-    "pytest-xdist",
+    "pytest-xdist"
 ]
-extras["dev"] = extras["quality"] + extras["test"]
+extras["trackers"] = ["wandb", "comet-ml"]
+extras["test_trackers"] = ["tensorflow"]
```
sgugger, line 26: I would group those together.

sgugger (approved):
Thanks a lot for all your work on this! Have a few tiny nits on the new setup extras, but otherwise good to go on my side!

PR #298, opened 2022-03-29
Fix example for datasets v2

sgugger:
Datasets v2 seems to have removed `rename_column_` so this PR adapts the example to run with it.


muellerzr (approved):
Thanks!

PR #300, opened 2022-03-30
Leave default as None in `mixed_precision` for launch command

sgugger:
The default in `mixed_precision` for the launch command should be `None`, otherwise the config value is not used. Thanks @nbroad1881 for reporting the bug!


muellerzr (approved):
Nice catch!

PR #301, opened 2022-03-30
Pass `lr_scheduler` to `Accelerator.prepare`

sgugger:
This PR adds support for passing the learning rate scheduler to `Accelerator.prepare`. This only concerns learning rate schedulers that are called at each optimizer step, not schedulers that are called at the end of each epoch.

**Why?**

There are three reasons motivating this change:
- users don't always remember they have to define the scheduler after they have prepared the training dataloader (since its size changes). This way is easier.
- the scheduler step should be synchronized with the optimizer step, so when the step is skipped for the optimizer (in mixed precision), it should also be in the scheduler. With this PR, this is now the case.
- this will unlock more cases where the optimizer step is skipped: having Accelerate automatically handle gradient accumulation


src/accelerate/accelerator.py
```diff
@@ -92,6 +93,9 @@ class Accelerator:
             If set to `True`, the dataloader prepared by the Accelerator is only iterated through on the main process
             and then the batches are split and broadcast to each process. Will default to `True` for `DataLoader` whose
             underlying dataset is an `IterableDataset`, `False` otherwise.
+        step_scheduler_with_optimizer (`bool`, *optional`, defaults to `True`):
+            Set `True` if you step your learning rate scheduler at the same time as the optimizer, `False` if you only
+            do it under certain circumstances (at the end of each epoch, for instance).
```
muellerzr, line 98:
```suggestion
            Set `True` if the learning rate scheduler is stepped at the same time as the optimizer, `False` if only
            done under certain circumstances (at the end of each epoch, for instance).
```

src/accelerate/accelerator.py
```diff
@@ -458,6 +472,21 @@ def prepare_data_loader(self, data_loader):
     def prepare_optimizer(self, optimizer):
         return AcceleratedOptimizer(optimizer, device_placement=self.device_placement, scaler=self.scaler)
 
+    def prepare_scheduler(self, scheduler):
+        # We try to find the optimizer associated with `scheduler`, the default is the full list.
+        optimizer = self._optimizers
+        for opt in self._optimizers:
+            if getattr(scheduler, "optimizer", None) == opt.optimizer:
+                optimizer = opt
+                break
```
muellerzr, line 481: Clever! I had wondered how you'd do this 😄

src/accelerate/checkpointing.py
```diff
@@ -65,6 +72,13 @@ def save_accelerator_state(
         output_optimizer_file = os.path.join(output_dir, optimizer_name)
         save(state, output_optimizer_file)
         logger.info(f"Optimizer state saved in {output_optimizer_file}")
+    # Scheduler states
+    for i, scheduler in enumerate(schedulers):
+        state = scheduler.state_dict()
+        scheduler_name = f"{SCHEDULER_NAME}.bin" if i == 0 else f"{SCHEDULER_NAME}_{i}.bin"
+        output_scheduler_file = os.path.join(output_dir, scheduler_name)
+        save(state, output_scheduler_file)
+        logger.info(f"Optimizer state saved in {output_scheduler_file}")
```
muellerzr, line 81:
```suggestion
        logger.info(f"Scheduler state saved in {output_scheduler_file}")
```

src/accelerate/scheduler.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    A wrapper around a learning rate scheduler that will only step when the optimizer(s) have a training step. Useful
+    to avoid making a scheduler step too fast when:
+
+    - gradients went overflow and there was no training step (in mixed precision training)
+    - step was skipped because of gradient accumulation
+
+    Args:
+        scheduler (`torch.optim.lr_scheduler._LRScheduler`): The scheduler to wrap.
+        optimizers (one or a list of `torch.optim.Optimizer`): The optimizers used.
```
muellerzr, line 28:
```suggestion
        scheduler (`torch.optim.lr_scheduler._LRScheduler`): 
            The scheduler to wrap.
        optimizers (one or a list of `torch.optim.Optimizer`): 
            The optimizers used.
```

muellerzr (approved):
Great work on this! Left a few suggestions for documentation consistency and a copy/paste victim 😉

LysandreJik (approved):
As said offline, LGTM!

PR #304, opened 2022-03-31
Create new TestCase classes and clean up W&B tests

muellerzr:
Cleans up the tests in trackers to better utilize mocking, tmpdir's, and make wandb like us.

(It didn't like if we removed the original logging dir, but doesn't complain when we just wipe everything inside it and keep writing to it!)

### DIFF
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -30,7 +30,7 @@ from ..utils import gather, is_tensorflow_available
 
 class TempDirTestCase(unittest.TestCase):
     """
-    A TestCase class that keeps a single tmpdir open for the duration of the class, wipes its data at the start of a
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Much cleaner!

src/accelerate/test_utils/testing.py
```diff
@@ -13,15 +13,84 @@
...
+        for path in Path(self.tmpdir).glob("**/*"):
+            if path.is_file():
+                path.unlink()
+            elif path.is_dir():
+                shutil.rmtree(path)
+
+
+class MockingTestCase(unittest.TestCase):
+    """
+    A TestCase class designed to dynamically add various mockers that should be used in every test, mimicing a behavior
```
muellerzr, line 63:
```suggestion
    A TestCase class designed to dynamically add various mockers that should be used in every test, mimicking the behavior
```

Issue #303, opened 2022-03-31
Docs state custom trackers are supported, but logic isn't there


muellerzr:
Realized this morning that the directions are a bit misleading, as we don't support custom trackers, but we do provide directions for creating one.

PR #305, opened 2022-03-31
Have custom trackers work with the API

muellerzr:
Implements the logic needed for custom trackers, so long as they subclass GeneralTracker and tests

closes https://github.com/huggingface/accelerate/issues/303


src/accelerate/accelerator.py
```diff
@@ -631,7 +637,10 @@ def init_trackers(self, project_name: str, config: Optional[dict] = None):
         """
         self.trackers = []
         for tracker in self.log_with:
-            if str(tracker).lower() == "tensorboard" and is_tensorboard_available():
+            if issubclass(type(tracker), GeneralTracker):
+                # Custom tracker that they already initalized
```
sgugger, line 641:
```suggestion
                # Custom trackers are already initialized
```

sgugger (approved):
Very nice!

PR #306, opened 2022-03-31
Write tests for comet_ml

muellerzr:
Found out comet also has an offline mode. Parsing their bits was trivial, we're just checking the messenger log here to pick up that the right values got stored.

Now we're fully tested 😄 

Key bits:
Slight behavioral differences between online vs offline for Comet:
- We use an `OfflineExperiment` rather than `Experiment`

Probably wouldn't be a bad idea to update the docs with this as well, basically stating if you are wanting to use comet offline to log, you should make those changes.


sgugger (approved):
Nice test!

PR #307, opened 2022-04-04
Update example scripts

muellerzr:
Start of updating the example scripts to use both the save/load_state and tracking

### DIFF
--- a/examples/complete_nlp_example.py
+++ b/examples/complete_nlp_example.py
@@ -170,28 +170,28 @@ def training_function(config, args):
         if args.with_tracking:
             total_loss = 0
         for step, batch in enumerate(train_dataloader):
+            # We need to skip steps until we reach the resumed step
             if args.resume_from_checkpoint and epoch == 0 and step < resume_step:
                 continue
-            else:
-                # We could avoid this line since we set the accelerator with `device_placement=True`.
-                batch.to(accelerator.device)
-                outputs = model(**batch)
-                loss = outputs.loss
-                loss = loss / gradient_accumulation_steps
-                # We keep track of the loss at each epoch
-                if args.with_tracking:
-                    total_loss += loss.detach().float()
-                accelerator.backward(loss)
-                if step % gradient_accumulation_steps == 0:
-                    optimizer.step()
-                    lr_scheduler.step()
-                    optimizer.zero_grad()
+            # We could avoid this line since we set the accelerator with `device_placement=True`.
+            batch.to(accelerator.device)
+            outputs = model(**batch)
+            loss = outputs.loss
+            loss = loss / gradient_accumulation_steps
+            # We keep track of the loss at each epoch
+            if args.with_tracking:
+                total_loss += loss.detach().float()
+            accelerator.backward(loss)
+            if step % gradient_accumulation_steps == 0:
+                optimizer.step()
+                lr_scheduler.step()
+                optimizer.zero_grad()
 
-                overall_step += 1
+            overall_step += 1
 
-                if isinstance(checkpointing_steps, int):
-                    if overall_step % checkpointing_steps == 0:
-                        accelerator.save_state(f"step_{overall_step}")
+            if isinstance(checkpointing_steps, int):
+                if overall_step % checkpointing_steps == 0:
+                    accelerator.save_state(f"step_{overall_step}")
         if state_restored:
             model.eval()
             for step, batch in enumerate(eval_dataloader):
--- a/examples/complete_nlp_example.py
+++ b/examples/complete_nlp_example.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+import os
 
 import torch
 from torch.utils.data import DataLoader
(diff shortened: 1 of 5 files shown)


examples/nlp_example.py
```diff
@@ -51,15 +51,28 @@
...
     # Sample hyper-parameters for learning rate, batch size, seed and a few other HPs
     lr = config["lr"]
     num_epochs = int(config["num_epochs"])
     correct_bias = config["correct_bias"]
     seed = int(config["seed"])
     batch_size = int(config["batch_size"])
 
+    # We need to initialize the trackers we use, and also store our configuration
+    if args.with_tracking:
+        accelerator.init_trackers("cv_example", config)
```
sgugger, line 74: This is the NLP example ;-)

examples/nlp_example.py
```diff
@@ -172,6 +214,19 @@ def main():
...
+        required=False,
+        help="Whether the various states should be saved at the end of every n steps, or 'epoch' for each epoch.",
+    )
+    parser.add_argument(
+        "--resume_from_checkpoint", required=False, help="If the training should continue from a checkpoint folder."
+    )
+    parser.add_argument(
+        "--with_tracking",
+        required=False,
+        help="Whether to load in all available experiment trackers from the enviornment and use them for logging.",
```
sgugger, line 228:
```suggestion
        help="Whether to load in all available experiment trackers from the environment and use them for logging.",
```

examples/nlp_example.py
```diff
@@ -172,6 +214,19 @@ def main():
         "and an Nvidia Ampere GPU.",
     )
     parser.add_argument("--cpu", action="store_true", help="If passed, will train on the CPU.")
+    parser.add_argument(
+        "--checkpointing_steps",
+        required=False,
```
sgugger, line 219: I'd put `type=str` and `default=None`here instead of `required=False`

examples/nlp_example.py
```diff
@@ -172,6 +214,19 @@ def main():
...
+        help="Whether the various states should be saved at the end of every n steps, or 'epoch' for each epoch.",
+    )
+    parser.add_argument(
+        "--resume_from_checkpoint", required=False, help="If the training should continue from a checkpoint folder."
```
sgugger, line 223: I'd put `type=str` and `default=None`here

examples/nlp_example.py
```diff
@@ -127,21 +140,38 @@ def collate_fn(examples):
         model, optimizer, train_dataloader, eval_dataloader, lr_scheduler
     )
 
+    # Potentially load in the weights and states from a previous save
+    if args.resume_from_checkpoint:
```
sgugger, line 144: There is a difference between this argument being None or "". In the second case we want to resume from the last checkpoint ideally.

examples/nlp_example.py
```diff
@@ -127,21 +140,38 @@ def collate_fn(examples):
...
+        accelerator.print(f"Resumed from checkpoint: {args.resume_from_checkpoint}")
+        accelerator.load_state(args.resume_from_checkpoint)
+
+    overall_step = 0
```
sgugger, line 148: When resuming a training, we need to skip some steps.

sgugger (commented):
Left some comments! I was thinking the base example should be left as is to be as simple as possible, and we could add a "complete_nlp_example" that contains the checkpoint and sharding, so that a reader only wanting a base training loop is not too lost. Wdyt?

examples/complete_nlp_example.py
```diff
@@ -0,0 +1,262 @@
...
+            accelerator.print(f"Resumed from checkpoint: {args.resume_from_checkpoint}")
+            accelerator.load_state(args.resume_from_checkpoint)
+            resume_step = None
+        else:
+            # Get the most recent checkpoint
+            dirs = [f.name for f in os.scandir(os.getcwd()) if f.is_dir()]
+            dirs.sort(key=os.path.getctime)
+            path = dirs[-1]  # Sorts folders by date modified, most recent checkpoint is the last
+            if "epoch" in path.name:
+                num_epochs = -int(path.name.replace("epoch_", ""))
```
sgugger, line 159: Did you mean `-=` here?

examples/complete_nlp_example.py
```diff
@@ -0,0 +1,262 @@
...
+            else:
+                resume_step = int(path.name.replace("step_", ""))
+                num_epochs = -resume_step // len(train_dataloader)
```
sgugger, line 162: Same here, this is weird.

examples/complete_nlp_example.py
```diff
@@ -0,0 +1,262 @@
...
+    # Now we train the model
+    for epoch in range(num_epochs):
+        model.train()
+        if args.with_tracking:
+            total_loss = 0
+        for step, batch in enumerate(train_dataloader):
+            if args.resume_from_checkpoint and epoch == 0:
+                for _ in range(resume_step - 1):
+                    pass
+            else:
```
sgugger, line 176: Then no need for an else branch

sgugger (commented):
A couple comments more on the step skipping!

examples/cv_example.py
```diff
@@ -149,10 +149,6 @@ def training_function(config, args):
         model, optimizer, train_dataloader, eval_dataloader, lr_scheduler
     )
 
-    # Instantiate learning rate scheduler after preparing the training dataloader as the prepare method
-    # may change its length.
-    lr_scheduler = OneCycleLR(optimizer=optimizer, max_lr=lr, epochs=num_epochs, steps_per_epoch=len(train_dataloader))
```
sgugger, line 154: Good catch!

sgugger (approved):
LGTM! I think we could add a mention in the examples README on why there are two different scripts for each modality.

PR #308, opened 2022-04-04
Fix training in DeepSpeed

### LINKED ISSUE #302: DeepSpeed integration doesn't seem to work with generated config
I have used `accelerate config` to generate config that uses deepspeed

Config:
```
deepspeed_config:
  offload_optimizer_device: NONE
  zero_stage: 3
distributed_type: DEEPSPEED
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 8
use_cpu: false
```

If I run this, I get a key error:

```
  File "/.../site-packages/accelerate/accelerator.py", line 240, in mixed_precision
    if self.state.deepspeed_plugin.deepspeed_config["fp16"]["enabled"]:
KeyError: 'fp16'
```

I am on python 3.7 and using accelerate  0.6.1 and deepspeed 0.6.1

I suspect the issue is caused by this line:
https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L274

which is querying the config that is written on this line:
https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils.py#L726
which does not have the keys in question

sgugger:
When mixed precision is set to no or bf16 with DeepSpeed, everything fails in Accelerate becasue the deepspeed config as no "fp16" key. This PR fixes that.

Fixes #302

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -285,9 +285,9 @@ class Accelerator:
     def mixed_precision(self):
         if self.distributed_type == DistributedType.DEEPSPEED:
             config = self.state.deepspeed_plugin.deepspeed_config
-            if "fp16" in config and config["fp16"]["enabled"]:
+            if "fp16" in config and config["fp16"].get("enabled", False):
                 mixed_precision = "fp16"
-            elif "bf16" in config and config["bf16"]["enabled"]:
+            elif "bf16" in config and config["bf16"].get("enabled", False):
                 mixed_precision = "bf16"
             else:
                 mixed_precision = "no"
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -285,9 +285,9 @@ class Accelerator:
     def mixed_precision(self):
         if self.distributed_type == DistributedType.DEEPSPEED:
             config = self.state.deepspeed_plugin.deepspeed_config
-            if "fp16" in config and config["fp16"].get("enabled", False):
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -284,9 +284,10 @@ def use_fp16(self):
     @property
     def mixed_precision(self):
         if self.distributed_type == DistributedType.DEEPSPEED:
-            if self.state.deepspeed_plugin.deepspeed_config["fp16"]["enabled"]:
+            config = self.state.deepspeed_plugin.deepspeed_config
+            if "fp16" in config and config["fp16"].get("enabled", False):
```
muellerzr, line 288:
```suggestion
            if config.get("fp16", {}).get("enabled", False):
```

src/accelerate/accelerator.py
```diff
@@ -284,9 +284,10 @@ def use_fp16(self):
...
                 mixed_precision = "fp16"
-            elif self.state.deepspeed_plugin.deepspeed_config["bf16"]["enabled"]:
+            elif "bf16" in config and config["bf16"].get("enabled", False):
```
muellerzr, line 290:
```suggestion
            elif config.get("bf16", {}).get("enabled", False):
```

muellerzr (approved):
LG2M! We could get a bit fancier with it if we want, but choose which is more readable to you 😄

PR #309, opened 2022-04-04
Use --no_local_rank for DeepSpeed launch

sgugger:
This way a user doesn't have to add a `local_rank` argument in their script.


muellerzr (approved):
Nice!

PR #311, opened 2022-04-08
Fix Accelerate CLI CPU option + small fix for W&B tests

muellerzr:
# Fix Accelerate CLI CPU option + Small W&B testfix

## What does this add?

- Change CPU param to use `convert_yes_no_to_bool` in the CLI Config
- Changes the logic for W&B `.log` tests to be `assertTrue` and "str" in `output`

## Why is this needed?

- Fixes an undocumented issue where the Accelerate Config CPU param expected a boolean argument, but it was phrased in a way where it seemed like yes/no could be passed in [forums post](https://discuss.huggingface.co/t/accelerate-on-1-gpu/16617)
- Noticed that the CI seemed to be slightly unstable with the W&B log parsing (unsurprising), so rather than a `self.assertEqual` I opted for a `self.assertTrue` and runs an `in` check, since we only care that the value itself is in the log, with potentially a little bit of noise

### DIFF
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -54,9 +54,10 @@ def get_cluster_input():
 
     if distributed_type == DistributedType.NO:
         use_cpu = _ask_field(
-            "Do you want to run your training on CPU only (even if a GPU is available)? [no]:",
-            lambda x: bool(x),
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks a lot for fixing!

PR #312, opened 2022-04-11
Refactor Examples by Feature

muellerzr:
# Seperate out example scripts by feature

## What does this add?

To jump ahead of a similar problem in Transformers currently, this PR adds additional example scripts for `Accelerate` features individually, and provides a framework for how to write a new script when some new feature is added.

Each new script is based on the original `nlp_example.py` script, and adds *one* single capability to each one. This is then documented in the `README.md` in `examples/by_feature`.

## Who is it for?

- Users of Accelerate and Transformers

## Why is it needed?

We received feedback recently that the current example scripts in `transformers` are becoming too bulky, and I especially noticed this in the `complete_nlp_example.py` script. As more features get added, the logic will get exponentially worse to navigate as time goes by

## Anticipated maintence burden? (What will happen in say, 3 months if something changes)

- Potentially we should test these scripts similar to how we do in `transformers` (this can also be done in this PR)

## Other thoughts

One more thought I had @sgugger, is perhaps we should leave this as the catch-all for examples inside of Accelerate, and keep the transformers scripts to their base (e.g. `nlp_example.py`), with no fancy bits. But to make sure folks know where to *find* more features, we can point them to our example scripts here.

This way we have far less spaghetti code, it's easier to maintain, and updates don't need to happen as frequently. WDYT?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -186,9 +186,19 @@ To run it in each of these various modes, use the following commands:
 
 ## Finer Examples
 
-While the first two scripts are extremely barebones when it comes to what you can do with accelerate, there are two complementary scripts for each: `complete_cv_example.py` and `complete_nlp_example.py`.
+While the first two scripts are extremely barebones when it comes to what you can do with accelerate, more advanced features are documented in two other locations.
 
-These two serve to document what else the `Accelerate` library is able to do, outside of just the basic training loop. They each have the following additional arguments:
+### `by_feature` examples
+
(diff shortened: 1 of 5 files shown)


examples/nlp_example.py
```diff
@@ -64,6 +63,8 @@ def training_function(config, args):
     datasets = load_dataset("glue", "mrpc")
     metric = load_metric("glue", "mrpc")
 
+    set_seed(seed)
+
```
muellerzr, line 67: Just as a note, I noticed we set_seed *before* building the dataloaders, hence why this change.

sgugger (approved):
Very nice! Thanks a lot for doing this!

PR #316, opened 2022-04-14
Refactor Tracker logic and write guards for logging_dir

muellerzr:
# Refactor `logging_dir` logic and make better guards for trackers that need one

## What does this add?

This PR refactors much of the logic when it comes to how Trackers are initialized and checked inside of `Accelerate`

## Who is it for?

- Users of Accelerate who want to use W&B, TensorBoard, etc and want clear errors as to what went wrong

## Why is it needed?

I noticed that users were finding it confusing as to why a `logging_dir` must always be passed in when using `"all"`, and this was because of `TensorBoard` requiring a logging directory.

On top of this, the current logic to check if a tracker can be added is a heap of spaghetti code that isn't maintainable. 

### User-facing:

When users forget to pass in a `logging_dir` to `Accelerate.__init__` and a particular tracker needs them, an error is now raised stating:
```
Logging with `TensorBoardTracker` requires a `logging_dir` to be passed in.
```

### Internal structure:

Each `GeneralTracker` implementation now requires that the `requires_logging_directory` be set, as this is a foundational method for checking whether `logging_dir` should be passed. As a result this is an abstract property, not a default (as the user should know explicitly).

There also now exists a `LOGGER_TYPE_TO_CLASS` dictionary to severely reduce the boilerplate code needed in `Accelerator.init_trackers` when it comes to checking if a logging_dir argument is needed

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -29,7 +29,7 @@ from .kwargs_handlers import DistributedDataParallelKwargs, GradScalerKwargs, In
 from .optimizer import AcceleratedOptimizer
 from .scheduler import AcceleratedScheduler
 from .state import AcceleratorState, DistributedType, is_deepspeed_available
-from .tracking import LOGGER_TYPE_TO_CLASS, GeneralTracker, get_available_trackers
+from .tracking import LOGGER_TYPE_TO_CLASS, GeneralTracker, filter_trackers
 from .utils import (
     DeepSpeedPlugin,
     LoggerType,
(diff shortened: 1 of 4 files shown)


src/accelerate/tracking.py
```diff
@@ -54,6 +54,14 @@ class GeneralTracker(object, metaclass=ABCMeta):
     A base Tracker class to be used for all logging integration implementations.
     """
 
+    @abstractproperty
+    def requires_logging_directory(self):
+        """
+        Whether the logger requires a directory to store their logs. Should either return `True` or `False`. The
+        default behavior is `False`
+        """
+        pass
```
sgugger, line 63: Why not use a class attribute directly? We could set a default then (I don't think this actually defaults to False).

src/accelerate/accelerator.py
```diff
@@ -644,12 +653,13 @@ def init_trackers(self, project_name: str, config: Optional[dict] = None):
             if issubclass(type(tracker), GeneralTracker):
                 # Custom trackers are already initialized
                 self.trackers.append(tracker)
-            elif str(tracker).lower() == "tensorboard" and is_tensorboard_available():
-                self.trackers.append(TensorBoardTracker(project_name, self.logging_dir))
-            elif str(tracker).lower() == "wandb" and is_wandb_available():
-                self.trackers.append(WandBTracker(project_name))
-            elif str(tracker).lower() == "comet_ml" and is_comet_ml_available():
-                self.trackers.append(CometMLTracker(project_name))
+            else:
```
sgugger, line 656: Very nice!

sgugger (approved):
Thanks for fixing! The logic for all the logging in the init of Accelerator is starting to take a lot of code, so I think we should refactor it in a separate method to keep the init cleam.

src/accelerate/tracking.py
muellerzr: This won't right. (This is also evidence of me going back and forth with it). Having it like this IMO forces implementations to think about how they want the behavior to go, and have it explicit in the source code for users interested in the code to see. 

For example, forcing it to be an abstract property/function, W&B could eventually make it a conditional where if it's in `offline_mode` (by checking the env var), then it would be True, otherwise False.

But, I'm not against just making it a normal class attr directly

sgugger: In that case, you should remove the default behavior mention in the docstring, as there is no default :-)

muellerzr:
@sgugger pinging for a rereview just to make sure the refactor seems sound to you 😄

src/accelerate/accelerator.py
```diff
@@ -132,32 +129,13 @@ def __init__(
...
-                        )
-                    if issubclass(type(log_type), GeneralTracker):
-                        loggers.append(log_type)
-                    else:
-                        log_type = LoggerType(log_type)
-                        if log_type not in loggers:
-                            loggers.append(log_type)
-        self.log_with = loggers
         self.logging_dir = logging_dir
+        self.log_with = filter_trackers(log_with, self.logging_dir)
```
sgugger, line 138: Much cleaner!

sgugger (approved):
LGTM, thanks for refactoring the bit that gets all loggers outside of the Accelerator!

PR #317, opened 2022-04-19
Create Cross-Validation example

muellerzr:
# Add Cross-Validation Example + Tests

## What does this add?

This adds an example to `examples/by_feature` for performing Cross Validation with `Accelerate`. 

Along with this, since we cannot run Cross-Validation well on 4 examples (our test sample dataset), I have elected to move the bits over from `transformers` so we can have `slow` tests. The CV examples run under slow, and take quite a bit (even on GPU, a minute). Ideally I should come up with a better solution for this, but I think for now if it is the only one it is not the end of the world.

This also implements the "skip_example" feature mentioned in my previous PR for making sure all aspects of a test are implemented in `complete_*_example.py`, since cross-validation is too niche for that.

## Who is it for?

Users of Accelerate, especially Kagglers, who want to perform Cross Validation with Accelerate and still have clean code

## Why is it needed?

Cross-Validation has a variety of setups that need to happen, and its not always straightforward how to do so with Accelerate in those instances (such as running `Accelerator.clean_memory()` at the end of a fold. This example shows how to do this

### User-facing:

A new `cross_validation.py` script, with a single added argument of `num_folds`

## Basic Usage Example(s):

python examples/cross_validation.py --num_folds 2

## Anticipated maintence burden? (What will happen in say, 3 months if something changes)

Most likely a better way of having "longer" example datasets to test will be used, and this test will need a refactor as a result

### DIFF
--- a/examples/by_feature/README.md
+++ b/examples/by_feature/README.md
@@ -54,3 +54,15 @@ These arguments should be added at the end of any method for starting the python
 ```bash
 accelerate launch ./tracking.py --with_tracking
 ```
+
+### Cross Validation (`cross_validation.py`)
+
+- Shows how to use `Accelerator.free_memory` and run cross validation efficiently with `datasets`.
+- Arguments available:
+  - `num_folds`, the number of folds the training dataset should be split into.
+
+These arguments should be added at the end of any method for starting the python script (such as `python`, `accelerate launch`, `python -m torch.distributed.launch`), such as:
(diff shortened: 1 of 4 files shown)


muellerzr:
Still todo is update the script docs, but code wise it's good

sgugger (approved):
Thanks a lot for adding this!

Issue #287, opened 2022-03-19
Stuck forever in accelerator.backward without any logs


ashutoshsaboo:
Hi there, 

I've a fairly standard training loop, see below. I've added a bunch of logs in between, just to see where is it stuck. To my surprise, the last log that I see is the log4 one below - just before `accelerator.backward`. Am I doing something wrong here? Or why is it stuck if you could please help? @sgugger 

As I inspect on nvidia-smi, the processes on the GPUs haven't been killed and have still memory occupied. I'm running this on multi-GPUs on the Sagemaker notebook (which is similar to a EC2) - i.e, not on the Sagemaker training job (which requires different configurations altogether in `accelerate config` - i.e, I haven't selected the AWS (Amazon SageMaker) sagemaker option while setting this up). I'm on accelerate==0.6.1 version. 

The training loop is fairly similar to the one here: https://github.com/huggingface/transformers/blob/master/examples/pytorch/text-classification/run_glue_no_trainer.py . 

```
        train_model.train()
        total_train_loss=0
        total_epoch_steps=0
        total_samples=0
        probs=[]
        refs=[]
        for step, batch in enumerate(dataloader):
            outputs = train_model(**batch)
            loss = outputs.loss
            loss = loss / args.gradient_accumulation_steps            
            probabilities = outputs.logits
            train_loss = accelerator.gather(loss).sum().item()
            print(train_loss, probabilities, "log1")
            predictions = accelerator.gather(probabilities.argmax(dim=-1)).detach().cpu()
            references = accelerator.gather(batch["labels"]).detach().cpu()
            print(predictions, references, "log2")
            batch_samples=len(accelerator.gather(batch['input_ids']))
            print(batch_samples, "log3")
            total_samples+=batch_samples
            total_train_loss+=(train_loss*batch_samples)
            print(total_train_loss, "log4")
            accelerator.backward(loss)
            print("backward done")
            if step % args.gradient_accumulation_steps == 0 or step == len(train_dataloader) - 1:
                print(step)
                print("optimizer step")
                optimizer.step()
                lr_scheduler.step()
                optimizer.zero_grad()
                progress_bar.update(1)
                completed_steps+=1
                total_epoch_steps+=1
            
            print("after loop")
            probs.extend(probabilities.softmax(dim=-1).detach().cpu().tolist())
            refs.extend(references.tolist())
            print(train_loss, step, epoch, refs)
            print(accelerator.is_main_process)
            metric.add_batch(
[...]

PR #319, opened 2022-04-19
Update examples to show how to deal with extra validation copies

muellerzr:
# Update examples to show how to truncate the validation set for metrics

## What does this add?

Based off [this issue](https://github.com/huggingface/accelerate/issues/287) this PR updates all examples to show how to get rid of extra samples that get added when performing distributed training on the validation set.

Testing on a multigpu system will happen tommorow, but @sgugger pretty sure the way I have it setup ensures that this only runs when we have distributed systems, and that's where this problem arises?

## Who is it for?

Should close https://github.com/huggingface/accelerate/issues/287

## Why is it needed?

It's unclear from the scripts how to alleviate this behavior, and it's not documented anywhere. So, with this PR it now is


muellerzr:
Note: This is just an initial to make sure the format and whatnot looks right and then all the other examples will follow suite :)

sgugger (commented):
Thanks for the PR! I'd just put this in a specific feature example instead of the base one.

examples/by_feature/multi_node_metrics.py
```diff
@@ -0,0 +1,215 @@
+# coding=utf-8
```
sgugger, line 1: Should be named `multi_process_metrics` as the problem arises as soon as you have several processes, not several nodes.

examples/by_feature/multi_node_metrics.py
```diff
@@ -0,0 +1,215 @@
...
+# Copyright 2021 The HuggingFace Inc. team. All rights reserved.
```
sgugger, line 2:
```suggestion
# Copyright 2022 The HuggingFace Inc. team. All rights reserved.
```

sgugger (approved):
Thanks! Not sure what your question is about tests. To test this we would need to know in advance the exact value of the metric and make sure we get that again but it's very finnicky since the metric computed in the base script is roughly the same.

PR #320, opened 2022-04-20
fix typo in docs of accelerate tracking

loubnabnl:
In some code examples `Accelerate` was imported from `accelerate` instead of `Accelerator` .


muellerzr (approved):
Very nice catch! LG2M

sgugger (approved):
Thanks for fixing!

PR #321, opened 2022-04-20
PyTorch FSDP Feature Incorporation

pacman100:
Changes to enable the PyTorch FSDP features.


src/accelerate/accelerator.py
```diff
@@ -348,6 +349,32 @@ def _prepare_one(self, obj, first_pass=False):
...
+                obj.optimizer = optimizers
+                for i, opt in enumerate(self._optimizers):
+                    if getattr(obj.scheduler, "optimizer", None) == opt.optimizer:
+                        obj.scheduler.optimizer = optimizers[i]
+                        obj.optimizers = [optimizers[i]]
+                        break
+                self._schedulers.append(obj)
+            result.append(obj)
+        self._optimizers = optimizers
+        return tuple(result)
```
sgugger, line 376: This is a first step but it assumes there is only one model and one optimizer with all the model parameters. We can call this method before the model is prepared and prepare it there, which would allow us to identify which parameters are use in which optimizers and then properly re-instantiate them with the right parameters/parameters groups.

Even better would be to replace the parameters in the existing optimizers without re-instantiating as we are losing parameter groups etc.

src/accelerate/accelerator.py
```diff
@@ -359,6 +386,19 @@ def prepare(self, *args):
...
+            optimizer_present = False
+            for obj in args:
+                if isinstance(obj, torch.nn.Module):
+                    model_count += 1
+                if isinstance(obj, torch.optim.Optimizer):
+                    optimizer_present = True
+            if model_count > 1 and optimizer_present:
+                raise ValueError(
+                    "For FSDP to work, prepare must be called for all the models before optimizers are created"
+                )
```
sgugger, line 400: Is this still necessary?

src/accelerate/accelerator.py
```diff
@@ -405,6 +448,21 @@ def prepare_model(self, model):
...
+
+            auto_wrap_policy = None
+            cpu_offload = None
+            if os.environ.get("FSDP_OFFLOAD_PARAMS", "false") == "true":
+                cpu_offload = CPUOffload(offload_params=True)
+
+            if int(os.environ.get("FSDP_MIN_NUM_PARAMS", "0")) > 0:
+                auto_wrap_policy = functools.partial(
+                    default_auto_wrap_policy, min_num_params=int(os.environ["FSDP_MIN_NUM_PARAMS"])
+                )
```
sgugger, line 464: Those two should also be added to the state, and we should get them from it here.

src/accelerate/commands/launch.py
```diff
@@ -472,6 +499,10 @@ def launch_command(args):
     # Use the proper launcher
     if args.use_deepspeed and not args.cpu:
         deepspeed_launcher(args)
+    elif args.use_fsdp and not args.cpu:
+        if version.parse(torch.__version__) < version.parse("1.12.0.dev20220418+cu113"):
+            raise ValueError("FSDP requires PyTorch >= 1.12.0.dev20220418+cu113")
```
sgugger, line 504: Was it not out in 1.11 ?. Let's use the `base_version` attribute to compare to 1.12.0 without anything added ;-)

src/accelerate/state.py
```diff
@@ -203,10 +204,13 @@ def __init__(
                 self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                 self.device = torch.device("cuda", self.local_process_index)
                 torch.cuda.set_device(self.device)
-                self.mixed_precision = (
-                    parse_choice_from_env("MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
-                )
-
+                if os.environ.get("USE_FSDP", "false") == "true":
+                    self.distributed_type = DistributedType.FSDP
+                    self.mixed_precision = "no"
```
sgugger, line 209: We should throw an error if we have mixed_precision different from "no" instead of silently ignoring what the user said in their config.

sgugger (commented):
Thanks a lot for your PR! Be careful, you committed some `.DS_store` files, you can add this entry to the `gitignore` to avoid this.

I understand the logic in `_prepare_fsdp` but we should support more than the simple use case (at least optimizers with parameter groups) otherwise users will get subtle bugs that hard to find.

I also think the `offload` and `min_num_params` (basically the FSDP config) should end up in the `AcceleratorState` for easy access.

src/accelerate/accelerator.py
pacman100: Hello, have updated the code as per yesterday's offline discussion.

src/accelerate/accelerator.py
pacman100: Yes, currently we will only be able to handle the case of single model and hence throwing error in case of multiple models.

src/accelerate/accelerator.py

src/accelerate/commands/launch.py
pacman100: As mentioned offline, at present the model saving/checkpointing is supported through nightly version only.

src/accelerate/state.py
pacman100: Done.

src/accelerate/accelerator.py
```diff
@@ -155,6 +160,16 @@ def __init__(
...
 
+        if os.environ.get("USE_FSDP", "false") == "true":
+            if version.parse(torch.__version__) < version.parse("1.12.0.dev20220418+cu113"):
+                raise ValueError("FSDP requires PyTorch >= 1.12.0.dev20220418+cu113")
+            if fsdp_plugin is None:  # init from env variables
+                fsdp_plugin = FullyShardedDataParallelPlugin()
+            else:
+                assert isinstance(
+                    fsdp_plugin, FullyShardedDataParallelPlugin
+                ), "`fsdp_plugin` must be a FullyShardedDataParallelPlugin object."
```
sgugger, line 171: For new code, we try to avoid assert, so please use a test and rasie a `TypeError` here.

src/accelerate/accelerator.py
```diff
@@ -326,6 +342,43 @@ def _prepare_one(self, obj, first_pass=False):
...
+        optimizers = []
+
+        self._schedulers = []
+        self._models = []
+        intermediate_result = []
+        for obj in args:
+            if isinstance(obj, torch.optim.Optimizer):
+                if len(obj.param_groups) > 1:
+                    logger.warn(
+                        "FSDP Warning: When using FSDP, several parameter groups will be conflated into a single one due to nested module wrapping and parameter flattening."
```
sgugger, line 360: Split to respect the 119 char limit please :-)

src/accelerate/accelerator.py
sgugger: The error message could be made clearer then.

src/accelerate/accelerator.py
```diff
@@ -373,16 +443,33 @@ def prepare(self, *args):
...
+            model = FSDP(
+                model,
+                sharding_strategy=fsdp_plugin.sharding_strategy,
+                cpu_offload=fsdp_plugin.cpu_offload,
+                auto_wrap_policy=fsdp_plugin.auto_wrap_policy,
+                backward_prefetch=fsdp_plugin.backward_prefetch,
+                ignored_modules=fsdp_plugin.ignored_modules,
+            ).to(self.device)
+            if fsdp_plugin.cpu_offload.offload_params:
+                model.to("cpu")
```
sgugger, line 472: There are two consecutive calls to `to` here in this case, would be better to only do one.

src/accelerate/state.py
```diff
@@ -206,7 +208,13 @@ def __init__(
                 self.mixed_precision = (
                     parse_choice_from_env("MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
                 )
-
+                if os.environ.get("USE_FSDP", "false") == "true":
+                    self.distributed_type = DistributedType.FSDP
+                    if self.mixed_precision != "no":
+                        raise ValueError(
+                            "Mixed precision is currently not supported for FSDPS. Please set `mixed_precision` to `no`."
```
sgugger, line 215:
```suggestion
                            "Mixed precision is currently not supported for FSDP. Please set `mixed_precision` to `no`."
```

src/accelerate/utils.py
```diff
@@ -738,6 +739,68 @@ def __post_init__(self):
...
+                self.auto_wrap_policy = functools.partial(default_auto_wrap_policy, min_num_params=self.min_num_params)
+
+        self.fsdp_config = {
+            "sharding_strategy": str(self.sharding_strategy),
+            "cpu_offload": self.cpu_offload.offload_params,
+            "min_num_params": self.min_num_params,
+            "is_default_auto_wrap_policy": self.auto_wrap_policy.__name__ == "default_auto_wrap_policy",
+            "ignored_modules_present": self.ignored_modules is not None,
+            "backward_prefetch_present": self.backward_prefetch is not None,
+        }
```
sgugger, line 801: Why add this here, is this used anywhere? The dataclass in itself should be enough to contain the full config?

src/accelerate/utils.py
```diff
@@ -738,6 +739,68 @@ def __post_init__(self):
         }
 
 
+@dataclass
+class FullyShardedDataParallelPlugin:
+    """
+    This plugin is used to enable fully sharded data parallelism.
+    """
+
+    from torch.distributed.fsdp.fully_sharded_data_parallel import BackwardPrefetch, CPUOffload, ShardingStrategy
```
sgugger, line 748: As said offline this is the line making the CI red.

sgugger (commented):
Looking good! There are still a few things to fix and it would be great to add some documentation as well :-)

src/accelerate/utils.py
pacman100: Got it. Thanks for pointing this out. Fixed this.

src/accelerate/utils.py
pacman100: It was used sort of "__repr__" to be printed when printing the accelerator state. removed this as the dataclass should be the full config.

src/accelerate/accelerator.py

src/accelerate/accelerator.py

src/accelerate/accelerator.py

src/accelerate/accelerator.py

docs/source/_toctree.yml
```diff
@@ -1,26 +1,28 @@
...
+    - local: kwargs
+      title: Kwargs Handlers
+    - local: internal
+      title: Internals
+    - local: checkpoint
+      title: Checkpointing
+    - local: tracking
+      title: Experiment Tracking
+    - local: fsdp
+      title: Fully Sharded Data Parallel
```
sgugger, line 27: Can we add the new guide without changing the indentation of the whole file? ;-)

docs/source/fsdp.mdx
```diff
@@ -0,0 +1,107 @@
...
+`Offload Params`: Decides Whether to offload parameters and gradients to CPU.
+```
+
+For more control, users can leverage the `FullyShardedDataParallelPlugin` wherein they can specify `auto_wrap_policy`, `backward_prefetch` and `ignored_modules`.
+After creating an instance of this class, users can pass it to the Accelerator class instantiation.
+For more information on these options, please refer to the PyTorch [FullyShardedDataParallel](https://github.com/pytorch/pytorch/blob/0df2e863fbd5993a7b9e652910792bd21a516ff3/torch/distributed/fsdp/fully_sharded_data_parallel.py#L236) code.
+
+[[autodoc]] utils.FullyShardedDataParallelPlugin
+
+## Few caveats to be aware of:
```
sgugger, line 74: I would put this before the API doc (so before the line [[autodoc]] above).

docs/source/fsdp.mdx
```diff
@@ -0,0 +1,107 @@
...
++         optimizer, train_dataloader, eval_dataloader, lr_scheduler
++        )
+
+```
+
+- In case of a single model, if you have created optimizer with multiple optimizer groups and called prepare with them together,
+  then the optimizer groups will be lost and the following warning is displayed:
+
+  > FSDP Warning: When using FSDP, several parameter groups will be conflated into
+  > a single one due to nested module wrapping and parameter flattening.
```
sgugger, line 104: You should highlight that FSDP throws away any parameter groups anyway.

sgugger (approved):
Thanks for adding documentation! We'll be just missing the test. I'll let you decide if you want to include it in this PR or in a follow-up one.

docs/source/fsdp.mdx
pacman100: Done.

docs/source/fsdp.mdx
pacman100: Done.

Issue #322, opened 2022-04-21
Should be continue instead of pass command at checkpoining iteration block


takiholadi:
https://github.com/huggingface/accelerate/blob/fa476d03ce39e67f6b5632f3308fa07f2fb37140/examples/by_feature/checkpointing.py#L196-L203

Some examples include such a checkpointing block.

There should be `continue` instead of `pass` at last line.

PR #323, opened 2022-04-21
Fixup all checkpointing examples

muellerzr:
# Fix logic in *all* checkpointing examples

## What does this add?

This PR fixes a number of bugs currently present in the save/load examples

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/322

## Why is it needed?

As I was exploring solving 322 and writing tests, I was noticing that some behaviors weren't *quite* behaving how I would have expected them to. 

It also didn't make logical sense to me that if we resume at epoch 1, the numbering starts at epoch 0 again (and thus, our checkpoint saves do as well!). So, that behavior had to change slightly.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -192,10 +192,11 @@ class FeatureExamplesTests(TempDirTestCase):
         with mock.patch("accelerate.Accelerator.print") as mocked_print:
             with mock.patch.object(sys, "argv", testargs):
                 checkpointing.main()
-            for call in mocked_print.mock_calls:
-                self.assertNotIn("epoch 0:", call.args)
-                self.assertNotIn("epoch 1:", call.args)
-            self.assertIn("epoch 2:", mocked_print.mock_calls[-1].args[0])
+            with self.assertRaises(AssertionError):
+                mocked_print.assert_any_call("epoch 0:", {"accuracy": 0.5, "f1": 0.0})
+            with self.assertRaises(AssertionError):
+                mocked_print.assert_any_call("epoch 1:", {"accuracy": 0.5, "f1": 0.0})
+            mocked_print.assert_any_call("epoch 2:", {"accuracy": 0.5, "f1": 0.0})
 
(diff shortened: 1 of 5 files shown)


muellerzr:
@sgugger 50/50 on whether to consider these "slow" tests or not. They add ~1.5 min to the CI

tests/test_examples.py
```diff
@@ -158,30 +158,59 @@ def test_cv_examples(self):
...
+        testargs = f"""
+        checkpointing.py
+        --resume_from_checkpoint {os.path.join(self.tmpdir, "epoch_1")}
+        """.split()
+        with mock.patch("accelerate.Accelerator.print") as mocked_print:
             with mock.patch.object(sys, "argv", testargs):
                 checkpointing.main()
-                self.assertTrue(os.path.exists(os.path.join(tmpdir, "epoch_0")))
+            with self.assertRaises(AssertionError):
+                mocked_print.assert_any_call("epoch 0:", {"accuracy": 0.5, "f1": 0.0})
```
sgugger, line 196: How are we sure we will get those values exactly?

sgugger (approved):
LGTM! Just one question on the tests added (nice addition btw :-) )

Thanks a lot for fixing those!

tests/test_examples.py
muellerzr: This stems from the scheduler we use in the examples, it makes it impossible for the model to train quickly so we always get an accuracy of .5 and an f1 of 0 for all of our epochs. Hence why none of these example tests check for if we get "good" accuracy, it's for independent behavior.

*But* I dug deeper and found `mock.ANY`. In this case we only care about matching the `epoch *` text, not the results. So instead we have something like this:

```python
dummy_results = {"accuracy":mock.ANY, "f1":mock.ANY}
with self.assertRaises(AssertionError):
    mocked_print.assert_any_call("epoch 0:", dummy_results)
```

Which helps me sleep much better at night

sgugger: You never know if your model might have learned a tiny something, so yes, much better :-)

PR #324, opened 2022-04-22
Add a memory-aware decorator for CUDA OOM avoidance

muellerzr:
# Introduce a `memory_aware` decorator to avoid CUDA Out-of-Memory

## What does this add?

This PR adds a decorator to help avoid CUDA OOM when running the training function. It is heavily inspired by [TOMA](https://github.com/BlackHC/toma/), with its methodology being a brute-force try/catch loop.

If a CUDA OOM, resource not available, or CPU OOM occurs at all when running the training function, the registered dataloader's batch size is cut in half and the function is run again.

It's used by adding **at most** *four* lines of code.

The important distinction between this and `toma` is we have a verbose error if the user accidentally passes the batch size into their decorated train function:
```
E           TypeError: Batch size was passed into `f` as the first argument when called. 
E                       Remove this as the decorator already does so:
E                           `f(value=64)`
```

## Who is it for?

Users of Accelerate who may be running big models, large datasets, etc and worry about hitting CUDA OOM issues

### User-facing:

Scripts using Accelerate that want this decorator will likely need a refactor. examples/by_feature/memory.py serves as an example script for what this looks like as a whole. A small snippet is shown below

## Basic Usage Example(s):

```diff
def training_function(config, args):
    accelerator = Accelerator(cpu=args.cpu, mixed_precision=args.mixed_precision)
    lr = config["lr"]
    num_epochs = int(config["num_epochs"])
    correct_bias = config["correct_bias"]
    seed = int(config["seed"])
    batch_size = int(config["batch_size"])
-   train_dataloader, eval_dataloader = get_dataloaders(accelerator, batch_size)
    metric = load_metric("glue", "mrpc")

    gradient_accumulation_steps = 1
    if batch_size > MAX_GPU_BATCH_SIZE:
        gradient_accumulation_steps = batch_size // MAX_GPU_BATCH_SIZE
        batch_size = MAX_GPU_BATCH_SIZE

    set_seed(seed)
    model = AutoModelForSequenceClassification.from_pretrained("bert-base-cased", return_dict=True)
    model = model.to(accelerator.device)
    optimizer = AdamW(params=model.parameters(), lr=lr, correct_bias=correct_bias)

+   @memory_aware(starting_batch_size=batch_size)
+    def inner_training_loop(batch_size):
+       nonlocal model, optimizer
        train_dataloader, eval_dataloader = get_dataloaders(accelerator, batch_size)

        lr_scheduler = get_linear_schedule_with_warmup(
            optimizer=optimizer,
            num_warmup_steps=100,
            num_training_steps=(len(train_dataloader) * num_epochs) // gradient_accumulation_steps,
        )
        model, optimizer, train_dataloader, eval_dataloader, lr_scheduler = accelerator.prepare(
            model, optimizer, train_dataloader, eval_dataloader, lr_scheduler
        )

        for epoch in range(num_epochs):
            model.train()
            for step, batch in enumerate(train_dataloader):
                batch.to(accelerator.device)
                outputs = model(**batch)
                loss = outputs.loss
                loss = loss / gradient_accumulation_steps
                accelerator.backward(loss)
                if step % gradient_accumulation_steps == 0:
                    optimizer.step()
                    lr_scheduler.step()
                    optimizer.zero_grad()

            model.eval()
            for step, batch in enumerate(eval_dataloader):
                batch.to(accelerator.device)
                with torch.no_grad():
                    outputs = model(**batch)
                predictions = outputs.logits.argmax(dim=-1)
                predictions, references = accelerator.gather((predictions, batch["labels"]))
                metric.add_batch(
                    predictions=predictions,
                    references=references,
                )

            eval_metric = metric.compute()
            accelerator.print(f"epoch {epoch}:", eval_metric)
+   inner_training_loop()


def main():
    parser = argparse.ArgumentParser(description="Simple example of training script.")
    parser.add_argument(
        "--mixed_precision",
        type=str,
        default="no",
        choices=["no", "fp16", "bf16"],
        help="Whether to use mixed precision. Choose"
        "between fp16 and bf16 (bfloat16). Bf16 requires PyTorch >= 1.10."
        "and an Nvidia Ampere GPU.",
    )
    parser.add_argument("--cpu", action="store_true", help="If passed, will train on the CPU.")
    args = parser.parse_args()
    config = {"lr": 2e-5, "num_epochs": 3, "correct_bias": True, "seed": 42, "batch_size": 16}
    training_function(config, args)


if __name__ == "__main__":
    main()
```

This would be used specifically when you *need* to ensure that your code will always run, without worrying about the best batch size that can fit. Just **make it run**


muellerzr:
Still TODO:

- Add documentation
- Write tests

src/accelerate/memory_utils.py
```diff
@@ -0,0 +1,82 @@
...
+            A list of prepared or unprepared `DataLoaders`.
+    """
+    if not isinstance(dataloaders, (tuple, list)):
+        dataloaders = [dataloaders]
+    if not all(
+        (isinstance(dataloader, (DataLoaderDispatcher, DataLoaderShard, DataLoader)) for dataloader in dataloaders)
+    ):
+        raise TypeError(
+            "Unsupported operation attempted. One or more dataloaders passed were not of type(s) `DataLoaderDispatcher`, `DataLoaderShard`, or `torch.utils.DataLoader`"
+        )
```
sgugger, line 66: We might hit some custom types of users here. The only thing we need to check is that there is a `batch_size` attribute and that it's an iterator, no?

src/accelerate/memory_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        return functools.partial(memory_aware, dataloaders=dataloaders)
+
+    def decorator(*args, **kwargs):
+        while True:
+            try:
+                return function(*args, **kwargs)
+            except Exception as e:
+                if should_reduce_batch_size(e):
+                    for dataloader in dataloaders:
+                        dataloader.batch_size /= 2
```
sgugger, line 78: Does this actually work? PyTorch fixes the `batch_sampler` at init, so I think this is ignore by the `dataloader`.

sgugger (commented):
Thanks for adding this! On top of the decorator, should we add a simple function `find_batch_size` that takes the dataloaders and the training function and ends by returning the batch size found?

muellerzr:
Usage example looks like the following now:
```python
# Assume we are in script.py
model = get_model()
optimizer = get_optimizer()
dl = get_train_dl()
lr_scheduler = get_lr_scheduler()

@memory_aware(starting_batch_size=128, function_arg_name="dataloaders", dataloader_arg_name="batch_size", dataloader_function=get_train_dl)
def training_function(dataloaders):
  accelerator = Accelerator(cpu=args.cpu, mixed_precision=args.mixed_precision)
  model, optimizer, dl, lr_scheduler = accelerator.prepare(
    model, optimizer, dl, lr_scheduler
  )
  for epoch in range(args.num_epochs):
    model.train()
    for step, batch in enumerate(dl):
      ...

def main():
  # Argparse
  training_function(args)

File: examples/by_feature/memory.py [hunk 3]
```diff
@@ -0,0 +1,218 @@
...
+    model = model.to(accelerator.device)
+
+    # Instantiate optimizer
+    optimizer = AdamW(params=model.parameters(), lr=lr, correct_bias=correct_bias)
+
+    # New Code #
+    # We now can define an inner training loop function. It should take a batch size as the only parameter,
+    # and build the dataloaders in there.
+    # It also gets our decorator
+    @memory_aware(starting_batch_size=batch_size)
```
sgugger, line 140: Still not fond of that name ;-) What about `@find_max_batch_size`?

tests/test_memory_utils.py
```diff
@@ -0,0 +1,62 @@
...
+                raise_fake_out_of_memory()
+
+        f()
+        self.assertListEqual(batch_sizes, [128, 64, 32, 16, 8])
+
+    def test_memory_explicit(self):
+        batch_sizes = []
+
+        @memory_aware(starting_batch_size=128)
+        def f(batch_size, arg1):
```
sgugger, line 42: Same here.

tests/test_memory_utils.py
```diff
@@ -0,0 +1,62 @@
...
+                raise_fake_out_of_memory()
+            return batch_size, arg1
+
+        bs, arg1 = f("hello")
+        self.assertListEqual(batch_sizes, [128, 64, 32, 16, 8])
+        self.assertListEqual([8, "hello"], [bs, arg1])
+
+    def test_verbose_guard(self):
+        @memory_aware(starting_batch_size=128)
+        def f(batch_size, arg1, arg2):
```
sgugger, line 55: Same there!

sgugger (approved):
Very nice API and I love the tests. Just put some suggestions on naming.

examples/by_feature/memory.py
muellerzr: Mostly since it's a start high -> go low approach if it broke, rather than find the maximum

sgugger: You do end up with the max that fits on the GPU, but your name works for me too.

docs/source/memory.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+One of the most frustrating errors when it comes to running training scripts is hitting "CUDA Out-of-Memory", 
+as the entire script needs to be restarted, progress is lost, and typically a developer would want to simply
+start their script and let it run.
+
+`Accelerate` provides a utility heavily based on [toma](https://github.com/BlackHC/toma) to give this capability.
+
+## find_executable_batch_size
+
+[[autodoc]] memory_utils.find_executable_batch_size
```
sgugger, line 23: Put the API doc at the end.

sgugger (approved):
Thanks for adding documentation!

PR #326, opened 2022-04-26
Introduce reduce operator

muellerzr:
# Introduce a `reduce` Op to `Accelerator`

## What does this add?

This PR introduces a `reduce` operator, based on catalyst's implementation + some nuonces for TPU's.

## Who is it for?

Users who want to have something akin to `.reduce()` for calculating their metrics

## Why is it needed?

Was requested in https://github.com/huggingface/accelerate/issues/215

### User-facing:

Introduces `Accelerator.reduce`

### Internal structure:

Adds a `reduce` function in `operations.py`

## Basic Usage Example(s):

```python
accelerator = Accelerator()
loss = F.binary_cross_entropy(input, target, reduction="none")
loss = accelerator.reduce(loss, "sum")
```


src/accelerate/utils.py
```diff
@@ -621,6 +622,35 @@ def _pad_across_processes(tensor, dim=0, pad_index=0, pad_first=False):
...
 
 
+def reduce(tensor: torch.Tensor, reduction="mean"):
+    """
+    Reduce `tensor` across all processes and perform a `reduction`.
+
+    Args:
+        tensor (`torch.Tensor`):
+            The data to reduce
+        reduction (`str`, *optional*, defaults to "mean"):
```
sgugger, line 632:
```suggestion
        reduction (`str`, *optional*, defaults to `"mean"`):
```

src/accelerate/utils.py
```diff
@@ -621,6 +622,35 @@ def _pad_across_processes(tensor, dim=0, pad_index=0, pad_first=False):
     )
 
 
+def reduce(tensor: torch.Tensor, reduction="mean"):
+    """
+    Reduce `tensor` across all processes and perform a `reduction`.
```
sgugger, line 627:
```suggestion
    Reduce a tensor across all processes by the mean of a given operation.
```
The arguments are described just below, so I like to keep the general description pure English :-)

sgugger (approved):
Thanks for adding this. Should we support nested structures of Tensors instead of just Tensors?

sgugger (approved):

PR #329, opened 2022-04-28
Fix default config dicts

sgugger:
If an existing config does not have anything on FSDP (as will be the case for most users since the future just went out) there is currently an error in the launcher that the config has a `None` `fsdp_config`. This PR takes care of this by making sure all dicts attributes of the config are initialized in the post_init.

cc @pacman100 for information.


muellerzr (approved):
LG2M bar quality!

PR #330, opened 2022-04-28
Clean up tests + fix import

muellerzr:
Refactors `TestCase` decorators to use `unittest.skipUnless` to simplify the logic greatly.

Solves import bug with `comet_ml` on `test_tracking.py`

Introduces requires decorators for `comet_ml` and `wandb`


sgugger (approved):
LGTM, thanks!

PR #334, opened 2022-04-28
Add guards for batch size finder

muellerzr:
Makes sure that if zero is reached when finding the batch size, we raise an error stating no batch size will fit


tests/test_memory_utils.py
```diff
@@ -50,6 +50,26 @@ def mock_training_loop_function(batch_size, arg1):
         self.assertListEqual(batch_sizes, [128, 64, 32, 16, 8])
         self.assertListEqual([bs, arg1], [8, "hello"])
 
+    def test_zero(self):
```
sgugger, line 53:
```suggestion
    def test_start_zero(self):
```

sgugger (approved):
Thanks for fixing!

PR #335, opened 2022-04-29
Patchfix infinite loop in bs finder

muellerzr:
Raising any error if it doesn't relate to CUDA OOM was removed in https://github.com/huggingface/accelerate/pull/334, this adds it back. 

Otherwise this would result in an infinite loop constantly reducing a bs of 0


sgugger (approved):
Thanks for fixing!

PR #337, opened 2022-04-29
Introduce multiprocess logger

muellerzr:
# Introduce Multi-Process aware Logger

## What does this add?

Adds a new `LoggerAdapter` and a `get_logger` helper to help aid with printing logs repeatedly when in multiprocess

## Why is it needed?

Users were reporting that the logs can confuse them with the actions actually occuring (such as logging multiple times a model was saved even though it only happened once).

Rather than an onslaught of if statements, this PR takes the simpler approach of creating an Adapter that will do the check for us. 

By default it will only log to the main processor, but if all processes should log just pass `main_process_only=False` (see examples below) 

### User-facing:

Loggers in Accelerate now accept a `main_process_only` kwarg

## Basic Usage Example(s):

```python
from accelerate.logging import get_logger
logger = get_logger("my_logger")

# If we only want it called in the main process
logger.info("My message!")
# or
logger.info("My message!", main_process_only=True)

# If it should be called across all processes
logger.info("My message!", main_process_only=False)
```

### DIFF
--- a/src/accelerate/logging.py
+++ b/src/accelerate/logging.py
@@ -13,26 +13,39 @@
 # limitations under the License.
 
 import logging
+from sys import exc_info
 from .state import AcceleratorState
+
+old_factory = logging.getLogRecordFactory()
+
+def multiprocess_record_factory(context_id):
+    def record_factory(*args, **kwargs):
+        record = old_factory(*args, **kwargs)
+
+        record.main_process_only = not main_process_only or (main_process_only and AcceleratorState().local_process_index == 0)
+
 
 class MultiProcessAdapter(logging.LoggerAdapter):
     """
     An adapter to assist with logging in multiprocess.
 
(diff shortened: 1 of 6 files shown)


src/accelerate/logging.py
```diff
@@ -0,0 +1,60 @@
...
+    @staticmethod
+    def _should_log(main_process_only):
+        "Check if log should be performed"
+        return not main_process_only or (main_process_only and AcceleratorState().local_process_index == 0)
+
+    def log(self, level, msg, *args, **kwargs):
+        """
+        Delegates logger call after checking if we should log
+        """
+        main_process_only = kwargs.pop("main_process_only", True)
```
sgugger, line 37: Nice!

sgugger (approved):
Looks really neat!

PR #340, opened 2022-05-04
Refactor utils into its own module

muellerzr:
Refactors all of `utils.py` into `utils.__init__.py` and all of the `*_utils.py` modules into the parent utils


examples/by_feature/memory.py
```diff
@@ -19,7 +19,7 @@
 from accelerate import Accelerator, DistributedType
 
 # New Code #
-from accelerate.memory_utils import find_executable_batch_size
+from accelerate.utils import find_executable_batch_size
```
sgugger, line 22: This one is rather important -> let's put it in the main init as well so that users import it directly from `accelerate` (and this way they'll be immune to the next cleanup)

src/accelerate/__init__.py
```diff
@@ -5,7 +5,12 @@
...
 from .launchers import debug_launcher, notebook_launcher
-from .state import DistributedType
-from .utils import DeepSpeedPlugin, synchronize_rng_states
+from .utils import (
+    DeepSpeedPlugin,
+    DistributedDataParallelKwargs,
+    DistributedType,
+    GradScalerKwargs,
+    InitProcessGroupKwargs,
+    synchronize_rng_states,
```
sgugger, line 15: So as said above, let's add the batch size finder here.

sgugger (approved):
Very nice! Thanks a lot for doing our spring cleanup.
The only breaking change I see for a public facing API is `from accelerate.memory_utils import find_executable_batch_size`, so let's maybe keep a `memory_utils` for a couple of release that raises a deprecation warning, reimport `find_executable_batch_size` there for backward compatibility, and also put `find_executable_batch_size` in the main init as it's a public API. WDYT?

src/accelerate/memory_utils.py
```diff
@@ -0,0 +1,28 @@
...
+# flake8: noqa
+# There's no way to ignore "F401 '...' imported but unused" warnings in this
+# module, but to preserve other warnings. So, don't check this module at all
+
+
+import warnings
+
+
+warnings.warn(
+    "memory_utils has been reorganized to utils.memory. Import `find_executable_batchsize` from the main `__init__`: `from accelerate import find_executable_batch_size` to avoid this",
```
sgugger, line 24:
```suggestion
    "memory_utils has been reorganized to utils.memory. Import `find_executable_batchsize` from the main `__init__`: "
    "`from accelerate import find_executable_batch_size` to avoid this warning.",
```

PR #342, opened 2022-05-05
Handle Manual Wrapping in FSDP. Minor fix of fsdp example.

pacman100:
### What does this PR do?
- Minor Fix of the FSDP example `fsdp_with_peak_mem_tracking.py`. Logging peak memory usage to trackers such as `wandb` only when it is enabled. 
- Handling `Manual wrapping` of the model when FSDP is enabled. If user has manually wrapped the model, then don't wrap it again.


sgugger (approved):
LGTM, don't forget to write a description of your PR before merging though :-)

PR #344, opened 2022-05-05
Better prompt for number of training devices

muellerzr:
cc @sgugger , let me know if I got this logic right/understanding this correctly. Should be a better solution than the previous PR


sgugger (approved):
Nice idea! I would just replace "TPUs" by "TPU cores".

PR #345, opened 2022-05-05
Big model inference

sgugger:
This PR introduces a set of tools to enable inference on large models that do not fit fully on one GPU, or even in RAM. Namely
- a context manager to initalize empty models
- a function to load a sharded checkpoint directly on the right devices
- a set of custom hooks that allow execution of a model split on different devices, as well as CPU or disk offload
- a magic method that auto-determines a device map for a given model, maximizing the GPU spaces, available RAM before using disk offload as a last resort.
- a function that wraps the last three blocks in one simple call (`load_checkpoint_and_dispatchl`)

Example: 
- download a repository with a sharded checkpoint, for instance [bigscience/T0pp](https://huggingface.co/bigscience/T0pp/tree/sharded) (sharded branch) or [sgugger/sharded-gpt-j-6B](https://huggingface.co/sgugger/sharded-gpt-j-6B).
- checkout this PR

```py
from accelerate import init_empty_weights, dispatch_model, infer_auto_device_map, load_sharded_checkpoint_in_model
from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, AutoModelForSeq2SeqLM

checkpoint = "bigscience/T0pp"
config = AutoConfig.from_pretrained(checkpoint)

# Initializes an empty shell with the model. This is instant and does not take any RAM.
with init_empty_weights():
    model = AutoModelForSeq2SeqLM.from_config(config)
# Initialize the model under the previous context manager breaks the tied weights.
model.tie_weights()
# Determine a device map that maximizes the available GPUs with `device_map="auto"`, you can also write your own.
# `no_split_module_classes` is only important for this automatic `device_map`, and avoids the corresponding layer being split across different devices.
model = load_checkpoint_and_dispatch(model, "T0pp", device_map="auto", no_split_module_classes=["T5Block"])

# The rest should feel familiar, the model can be used as usual even if it sits on multiple device!
tokenizer = AutoTokenizer.from_pretrained(checkpoint)
inputs = tokenizer("Task: copy but say the opposite. PSG won its match against Barca.", return_tensors="pt")
inputs = inputs.to(0)
output = model.generate(inputs["input_ids"])
tokenizer.decode(output[0].tolist())
```

TODO:
- [x] polish final API
- [x] write more tests of `infer_auto_device_map`
- [x] simplify code of `infer_auto_device_map`
- [ ] write doc


src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,203 @@
...
+
+import torch
+import torch.nn as nn
+
+from .hooks import AlignDevicesHook, add_hook_to_module, attach_align_device_hook, attach_align_device_hook_on_blocks
+from .utils import OffloadedWeightsLoader, extract_submodules_state_dict, offload_state_dict
+
+
+@contextmanager
+def init_empty_weights(include_buffers: bool = False):
```
patrickvonplaten, line 27: Nice! Think that's a good use of a context manager mechanism - it's quite intuitive that inside the context init uses no RAM.

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,203 @@
...
+    memory-mapped array in a given folder. During the forward pass, parameters will be accessed from that folder and
+    put on the execution device passed as they are needed, then offloaded again.
+
+    Args:
+        model (`torch.nn.Module`): The model to offload.
+        offload_dir (`str` or `os.PathLike`):
+            The folder in which to offload the model weights (or where the model weights are already offloaded).
+        execution_device (`torch.device`, *optional*):
+            The device on which the forward pass of the model will be executed (should be a GPU). Will default to the
+            model first parameter device.
```
patrickvonplaten, line 121:
```suggestion
            model's first parameter device.
```

tests/test_big_modeling.py
```diff
@@ -0,0 +1,204 @@
...
+
+        with TemporaryDirectory() as tmp_dir:
+            dispatch_model(model, device_map, offload_dir=tmp_dir)
+            output = model(x)
+            self.assertTrue(torch.allclose(expected, output.cpu(), atol=1e-5))
+
+    @require_multi_gpu
+    def test_dispatch_model_multi_gpu(self):
+        model = BiggerModelForTest()
+        device_map = {"linear1": "cpu", "linear2": "disk", "batchnorm": "cpu", "linear3": 0, "linear4": 1}
```
patrickvonplaten, line 156: Could it make sense to have models automatically create such a map for themselves? Think this would have to be model-specific and be in Transformers no?

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,203 @@
...
+        main_device (`str`, `int` or `torch.device`, *optional*):
+            The main execution device. Will default to the first device in the `device_map` different from `"cpu"` or
+            `"disk"`.
+        state_dict (`Dict[str, torch.Tensor]`, *optional*):
+            The state dict of the part of the model that will be kept on CPU.
+        offload_dir (`str` or `os.PathLike`):
+            The folder in which to offload the model weights (or where the model weights are already offloaded).
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            Whether or not to offload the buffers with the model parameters.
+    """
```
patrickvonplaten, line 167: Maybe worth adding a quick test here in the beginning that the parameter names of the model all have to be present in the `device_map()` so that people would get a quick / good error message?

src/accelerate/utils/modeling.py
```diff
@@ -0,0 +1,454 @@
...
+        max_memory["cpu"] = psutil.virtual_memory().available
+        return max_memory
+
+    for key in max_memory:
+        if isinstance(max_memory[key], str):
+            max_memory[key] = convert_file_size_to_int(max_memory[key])
+    return max_memory
+
+
+def infer_auto_device_map(model, max_memory=None, no_split_module_classes=None):
```
patrickvonplaten, line 279: Think a type hint for `max_memory` that it should be a dict could be useful here

patrickvonplaten:
Super cool PR! Two questions:

1. How fast is generate with parts of the weights offloaded compared to not offloaded? Guess it strongly depends on the hardware, but e.g. on a V100, for let's say `gpt2`?

2. More generally, do you think we could eventually do the following code:

```py
checkpoint = "bigscience/T0pp"
config = AutoConfig.from_pretrained(checkpoint)

# Initializes an empty shell with the model. This is instant and does not take any RAM.
with init_empty_weights():
    model = AutoModelForSeq2SeqLM.from_config(config)
# Initialize the model under the previous context manager breaks the tied weights.
model.tie_weights()
# Determine a device map that maximizes the available GPUs, you can also write your own.
# If using a different model, adjust `T5Block` to the proper class (for instance `"GPTJBlock"`)
device_map = infer_auto_device_map(model, no_split_module_classes=["T5Block"])
# Load the sharded checkpoint inside the model. This will load each part of the model on the device specified by `device_map`
load_sharded_checkpoint_in_model(model, path_to_local_t0pp, device_map=device_map)
# This will make that model that leaves on several different devices just work.
model = dispatch_model(model, device_map)
```

fully in Tranformers' `from_pretrained(...)` method with an optional flag? E.g. from a user's perspective wouldn't it be very cool to be able to just do something liike:

```py
from transformers import AutoModelForSeq2SeqLM

model = AutoModelForSeq2SeqLM.from_pretrained(checkpoint, device_map="auto")
```

would this be desirable in a future PR?

src/accelerate/big_modeling.py
sgugger: Yes, definitely needed!

philschmid:
> I'll post more benchmarks toward the end but basically:
> 
> * having the model on 2 GPUs instead of one for generate does not change speed. (I do have an NVlink though, so maybe it could slower, though there is only one change of GPU per forward pass, so it's not much)
> * offloading to CPU his 10-15x slower than GPU (for the part offloaded, which is why it's import to maximize GPU use before using CPU offload)
> * offloading to disk is 20-30x slower than GPU.

It would be interesting to compare to [deepseed inference](https://www.deepspeed.ai/tutorials/inference-tutorial/).

sgugger:
I doubt it will be faster @philschmid but it won't be a black box that replaces your model.

src/accelerate/utils/modeling.py
```diff
@@ -0,0 +1,489 @@
...
+    Find the tied parameters in a given model.
+
+    Args:
+        model (`torch.nn.Module`): The model to inspect.
+
+    <Tip warning={true}>
+
+    The signature has more arguments, but they are for the recursive part of this function and you should ignore them.
+
+    </Tip>
```
muellerzr, line 160: We should just keep those as kwargs then I think, since the user should never need to worry about them

tests/test_big_modeling.py
```diff
@@ -0,0 +1,204 @@
...
+        output = model(x)
+        self.assertTrue(torch.allclose(expected, output.cpu()))
+
+    @slow
+    @require_cuda
+    def test_cpu_offload_gpt2(self):
+        tokenizer = AutoTokenizer.from_pretrained("gpt2")
+        inputs = tokenizer("Hello world! My name is", return_tensors="pt").to(0)
+
+        gpt2 = AutoModelForCausalLM.from_pretrained("gpt2")
```
muellerzr, line 105: Just to make sure I understand correctly, we just check to make sure it *runs*, we don't worry about if the outputs make sense?

Should we just to make sure the params didn't get messed with at all, etc?

src/accelerate/hooks.py
```diff
@@ -0,0 +1,411 @@
...
+        module (`torch.nn.Module`): The module to attach a hook to.
+        hook (`ModelHook`): The hook to attach.
+
+    Returns:
+        `torch.nn.Module`: The same module, with the hook attached (the module is modified in place, so the result can
+        be discarded).
+    """
+    if hasattr(module, "_hf_hook") and hasattr(module, "_old_forward"):
+        # If we already put some hook on this module, we replace it with the new one.
+        old_forward = module._old_forward
```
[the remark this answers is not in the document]
sgugger: Too much feedback I believe. This is also very internal, I don't think the average user will use those hooks.

src/accelerate/hooks.py
```diff
@@ -0,0 +1,411 @@
...
+            Whether or not the weights should be offloaded after the forward pass. It can be one boolean for the whole
+            module, or a dictionary mapping module name to boolean.
+        weights_map (`Mapping[str, torch.Tensor]`, *optional*):
+            When the model weights are offloaded, a (potentially lazy) map from param names to the tensor values.
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            Whether or not to include the associated module's buffers when offloading.
+        module_name (`str`, *optional*, defaults to `""`):
+            The name of the module.
+    """
+    # If one device and one offload, we've go one hook.
```
pacman100, line 357:
```suggestion
    # If one device and one offload, we've got one hook.
```

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,260 @@
...
+        execution_device (`torch.device`, *optional*):
+            The device on which the forward pass of the model will be executed (should be a GPU). Will default to the
+            model first parameter device.
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            Whether or not to offload the buffers with the model parameters.
+        state_dict (`Dict[str, torch.Tensor]`, *optional*):
+            The state dict of the model that will be kept on CPU.
+    """
+    if execution_device is None:
+        execution_device = next(iter(model.parameter())).device
```
pacman100, line 102:
```suggestion
        execution_device = next(iter(model.parameters())).device
```

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,260 @@
...
+        execution_device (`torch.device`, *optional*):
+            The device on which the forward pass of the model will be executed (should be a GPU). Will default to the
+            model's first parameter device.
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            Whether or not to offload the buffers with the model parameters.
+    """
+    if not os.path.isdir(offload_dir) or not os.path.isfile(os.path.join(offload_dir, "index.json")):
+        offload_state_dict(offload_dir, model.state_dict())
+    if execution_device is None:
+        execution_device = next(iter(model.parameter())).device
```
pacman100, line 136:
```suggestion
        execution_device = next(iter(model.parameters())).device
```

pacman100 (commented):
Great functionality! Few comments and suggestions from my side.

src/accelerate/utils/modeling.py
```diff
@@ -0,0 +1,568 @@
...
+    # Get default / clean up max_memory
+    max_memory = get_max_memory(max_memory)
+    if no_split_module_classes is None:
+        no_split_module_classes = []
+    elif not isinstance(no_split_module_classes, (list, tuple)):
+        no_split_module_classes = [no_split_module_classes]
+
+    devices = list(max_memory.keys()) + ["disk"]
+    # Devices that need to keep space for a potential offloaded layer.
+    main_devices = [devices[0], "cpu"]
```
LysandreJik, line 342: `max_memory` is a dictionary, which is unordered in python <= 3.6.

Here, we're relying on the ordering of the dictionary in order to retrieve the `main_devices`. I'm open to hearing what you think; I would push to always consider dictionaries as unordered and to treat them as such, but as we're dropping python 3.6 I understand if you want to have the asumption that dictionaries are ordered.

These lists could likely be `set`s however: `devices` should theoretically not have the same device twice, and the same holds for `main_devices`.

If no GPU is available, then `devices` is a list containing `["cpu", "disk"]`. In that case, we end up with `main_devices` containing `['cpu', 'cpu']`.

src/accelerate/utils/modeling.py
```diff
@@ -0,0 +1,568 @@
...
+def clean_device_map(device_map: Dict[str, Union[int, str, torch.device]], module_name: str = ""):
+    """
+    Cleans a device_map by grouping all submodules that go on the same device together.
+    """
+    # Get the value of the current module and if there is only one split across several keys, regroup it.
+    values = [v for k, v in device_map.items() if k.startswith(module_name)]
+    if len(set(values)) == 1 and len(values) > 1:
+        for k in [k for k in device_map if k.startswith(module_name)]:
+            del device_map[k]
+        device_map[module_name] = values[0]
```
LysandreJik, line 291: If I understand this correctly, this will return an empty device map if everything fits on a single device. If it receives the following device map, for example:
```py
clean_device_map({'transformer': 0, 'lm_head': 0})
```
it will return the following:
```
{'': 0}
```
which, unfortunately, cannot be digested by the code later on as keys cannot find a common radix

src/accelerate/utils/modeling.py
```diff
@@ -0,0 +1,568 @@
...
+        )
+
+    checkpoint_files = None
+    index_filename = None
+    if os.path.isfile(checkpoint):
+        if str(checkpoint).endswith(".json"):
+            index_filename = checkpoint
+        else:
+            checkpoint_files = [checkpoint]
+    elif os.path.isdir(checkpoint):
```
LysandreJik, line 502: Nitpick: this fails right now if we specify a checkpoint like we would do in the transformers library. For example, this fails:

```py
with init_empty_weights():
    model = GPT2LMHeadModel.from_pretrained('./gpt2')

model = load_checkpoint_and_dispatch(
    model,
    "./gpt2",
    device_map='auto'
)
```

whereas intuitively I would expect it to work. I understand if that's wanted behavior and that passing the `pytorch_model.bin` is what you want instead.

src/accelerate/utils/modeling.py
LysandreJik: If you want a reproducible code example, I've tried with `gpt2` which I can load entirely in my model's memory. For example:

```py
from accelerate import init_empty_weights, load_checkpoint_and_dispatch
from transformers import GPT2LMHeadModel
import torch

GPT2LMHeadModel.from_pretrained('gpt2').save_pretrained('./gpt2')

with init_empty_weights():
    model = GPT2LMHeadModel.from_pretrained('gpt2')

model = load_checkpoint_and_dispatch(
    model,
    "./gpt2/pytorch_model.bin",
    device_map='auto'
)

```

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,260 @@
...
+        offload_dir (`str` or `os.PathLike`):
+            The folder in which to offload the model weights (or where the model weights are already offloaded).
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            Whether or not to offload the buffers with the model parameters.
+    """
+    # Error early if the device map is incomplete.
+    check_device_map(model, device_map)
+
+    if main_device is None:
+        main_device = [d for d in device_map.values() if d not in ["cpu", "disk"]][0]
```
LysandreJik, line 181: This and the code below tells me that it may not have been made so that `cpu` is a `main_device` :smile: 
I'll stop bugging you with it.

LysandreJik (approved):
Very impressive how you managed to distill such black magic in clear, understandable code. I've left some comments and nitpicks, but IMO most of my comments can be taken care of in a future PR or in a version 2. I think this PR is already very ready to be merged!

A bunch of feedback and nits:

---

<details>
<summary>Defining a device map</summary>

How would you go about crafting a `device_map` when not using `auto`? I think providing an example of how to do so, and the level of granularity offered by it would be interesting.

The way I did it personally was to infer the auto device map, to save that to a dict, and manually update weights inside it. Is this the recommended way to handle it?
</details>

---

<details>
<summary>Offload folder</summary>
The offload folder is super useful when we want to offload to disk. I'm wondering if it shouldn't default to a temporary folder in the case that it wasn't provided?

It could be accompanied by a warning mentioning that choosing a manual folder on the fastest storage device would be the best.
</details>

---

<details>
<summary>Lack of GPU</summary>
I may have been mistaken, but my understanding is that this can also be very powerful even if there is no GPU available: it would provide disk offload even when the only device is a CPU.

It seems that right now this isn't handled (see comments)
</details>

---
<details>
<summary>*nitpick* Better error messages</summary>
(nitpick, likely for future version) When trying out the `with init_empty_weights()` context manager, I initially forgot to remove the `.to('cuda')` I had previously used:

```py
with init_empty_weights():
    model = OPTModel(config).to('cuda')
```

This, of course, cannot work, but the error was a bit cryptic (even if understandable):
```
NotImplementedError: Cannot copy out of meta tensor; no data!
```
Maybe a distinct message saying that `to` should not be used when using `init_empty_weights` would help.

If that's a rough edge that we want to fix, doing inference with the model right after `init_empty_weights` results in a similarly cryptic error.
</details>

---

src/accelerate/big_modeling.py
sgugger: Good catch!

src/accelerate/utils/modeling.py
sgugger: Assuming the modeling file is called `pytorch_model.bin` is very Transformers-centric. For Accelerate, I don't want to make this assumption.

LysandreJik: Sounds good, will likely become very intuitive after implementing it in `transformers` so happy to keep it this way :+1:

src/accelerate/big_modeling.py
```diff
@@ -0,0 +1,260 @@
...
+        offload_folder (`str` or `os.PathLike`, *optional*):
+            If the `device_map` contains any value `"disk"`, the folder where we will offload weights.
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            In the layers that are offloaded on the CPU or the hard drive, whether or not to offload the buffers as
+            well as the parameters.
+    """
+    if device_map == "auto":
+        device_map = infer_auto_device_map(
+            model, max_memory=max_memory, no_split_module_classes=no_split_module_classes
+        )
```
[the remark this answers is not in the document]
sgugger: I will store it in `model.hf_device_map` to avoid conflicts with potential existing `device_map`.

src/accelerate/utils/modeling.py
sgugger: This shouldn't rely on the order (we don't care if the "main" GPU is 0, 1, etc.) we just need to make sure we don't end up with `"cpu"` twice :-)

src/accelerate/utils/modeling.py
sgugger: Fixed

LysandreJik (approved):
Thanks for working on the doc!

Issue #350, opened 2022-05-07
Evaluation could have duplicated validation data


allanj:
When we use the accelerator `gather` function, potentially, we would receive more validation data than we have.
For example, the batch size for each device is `8` and we have `3` devices. Our actually total data is 22 samples.
When we do `gather`, we receive `24` samples but not `22`, this might cause an inconsistent evaluation metric.

What I did after I realized this issue, I did the following two lines.
```python3
## here predictions is after we do gather operation and did certain post-processing
predictions = predictions[:len(valid_dataloader.dataset)]
targets = targets[:len(valid_dataloader.dataset)]
```

Shall we kind of specify this note in the README or the example (https://github.com/huggingface/accelerate/blob/main/examples/nlp_example.py#L160-L169)?

muellerzr:
Thanks for catching this! Will fix these examples next week. The transformers no_trainer examples have the right behavior, I just did not update these it seems.

https://github.com/huggingface/transformers/blob/main/examples/pytorch/text-classification/run_glue_no_trainer.py#L525

muellerzr:
This will be closed by https://github.com/huggingface/accelerate/pull/352, but we came to a conclusion that this is specified inside of the https://github.com/huggingface/accelerate/blob/main/examples/by_feature/multi_process_metrics.py example here. A note in the README couldn't hurt

PR #352, opened 2022-05-09
Fix sample calculation in all examples

muellerzr:
# Propogate validation calculation fix across all examples

## What does this add?

This PR makes sure that all validation metrics are calculated correctly when in multiprocess. There still lives the direct example in examples/multi_process_metrics.py, hence why there wasn't an explaination in these

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/350

### DIFF
--- a/examples/by_feature/checkpointing.py
+++ b/examples/by_feature/checkpointing.py
@@ -229,6 +229,7 @@ def training_function(config, args):
                     accelerator.save_state(output_dir)
 
         model.eval()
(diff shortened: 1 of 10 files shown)


sgugger (commented):
I'd prefer this to be in the complete examples only (not the base one) and have one specific by_feature example showcasing the changes to do for proper distributed evaluation.

sgugger (approved):

PR #355, opened 2022-05-10
Fixing metric eval in distributed setup

pacman100:
### What does this PR do?
1. Fixing off by 1 issue during metric evaluation in distributed setup.


sgugger (approved):
Thanks for fixing!

PR #356, opened 2022-05-11
DeepSpeed and FSDP plugin support through script

pacman100:
### What does this PR do?

1. Setting env variables when DeepSpeed /FSDP plugins are provided directly through script without using accelerate launch utility.


sgugger (approved):
LGTM, thanks a lot for fixing this!

PR #360, opened 2022-05-12
Handle deprication errors in launch

muellerzr:
- Checks for the right launcher based on the torch version
- Adds `num_cpu_threads_per_process` and sets it to env to get rid of torchrun warning. Default is torchrun default

### DIFF
--- a/src/accelerate/utils/versions.py
+++ b/src/accelerate/utils/versions.py
@@ -1,0 +1,25 @@
+# Copyright 2022 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/versions.py
```diff
@@ -0,0 +1,25 @@
...
+
+from packaging.version import parse
+
+
+if sys.version_info < (3, 8):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
+
+_torch_version = parse(importlib_metadata.version("torch"))
```
sgugger, line 25: No need to make it private since we reuse it.

sgugger (approved):
Thanks for dealing with this!

PR #361, opened 2022-05-12
fix tracking

pacman100:
### What does this PR do?

1. When dictionary of metrics is provided to the accelerate tracker, tensorboard wasn't able to log it. Example with dict of metrics is in this transformers script [run_glue_no_trainer.py](https://github.com/huggingface/transformers/blob/9f16a1cc13029c37fd3b7713d0edb743a8a86e39/examples/pytorch/text-classification/run_glue_no_trainer.py#L541). Fixed support for dictionary of int/float metrics 
2. Fixing examples so that run is created only for the main process else wandb will create num_processes runs with no data.
3. comet-ml logging wasn't helpful as everything was being logged in others `log_others`. No plots were being created. Fixing it to be able to get plots for metrics using `log_metric` and `log_metrics`.

Post these changes, below are the plots for the [run_glue_no_trainer.py](https://github.com/huggingface/transformers/blob/main/examples/pytorch/text-classification/run_glue_no_trainer.py) script on MRPC task using all trackers.
Screenshot 2022-05-12 at 8 11 16 PM
Screenshot 2022-05-12 at 8 10 51 PM
Screenshot 2022-05-12 at 8 10 17 PM

### DIFF
--- a/examples/by_feature/fsdp_with_peak_mem_tracking.py
+++ b/examples/by_feature/fsdp_with_peak_mem_tracking.py
@@ -294,7 +294,7 @@ def training_function(config, args):
                     {
                         "accuracy": eval_metric["accuracy"],
                         "f1": eval_metric["f1"],
-                        "train_loss": total_loss.item() if type(total_loss) == torch.Tensor else total_loss,
+                        "train_loss": total_loss.item(),
                     },
                     step=epoch,
                 )
--- a/examples/by_feature/fsdp_with_peak_mem_tracking.py
+++ b/examples/by_feature/fsdp_with_peak_mem_tracking.py
@@ -103,11 +103,12 @@ def training_function(config, args):
 
     # We need to initialize the trackers we use, and also store our configuration
     if args.with_tracking:
-        run = os.path.split(__file__)[-1].split(".")[0]
-        if args.logging_dir:
-            run = os.path.join(args.logging_dir, run)
-            accelerator.print(run)
-        accelerator.init_trackers(run, config)
+        if accelerator.is_main_process:
+            run = os.path.split(__file__)[-1].split(".")[0]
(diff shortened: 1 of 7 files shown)


examples/by_feature/fsdp_with_peak_mem_tracking.py
```diff
@@ -293,7 +294,7 @@ def collate_fn(examples):
                     {
                         "accuracy": eval_metric["accuracy"],
                         "f1": eval_metric["f1"],
-                        "train_loss": total_loss,
+                        "train_loss": total_loss.item() if type(total_loss) == torch.Tensor else total_loss,
```
sgugger, line 297: Do we really have either a Tensor or something else? Can't it always be `.item()` here?

examples/by_feature/tracking.py
```diff
@@ -208,15 +209,16 @@ def training_function(config, args):
...
         # To actually log, we call `Accelerator.log`
-        # The values passed can be of `str`, `int`, or `float`
+        # The values passed can be of `str`, `int`, `float` or `dict` of `str` to `float`/`int`
         if args.with_tracking:
             accelerator.log(
                 {
                     "accuracy": eval_metric["accuracy"],
                     "f1": eval_metric["f1"],
-                    "train_loss": total_loss,
+                    "train_loss": total_loss.item() if type(total_loss) == torch.Tensor else total_loss,
```
sgugger, line 218: Same comment

sgugger (commented):
Thanks for looking into this!

sgugger:
Oh, also make sure to edit the title of the PR ;-)

muellerzr (approved):
LG2M, will handle looking at the failing diff test tomorrow morning, as not sure quite what's going on there.

PR #363, opened 2022-05-13
Fix minor typo in `notebook_launcher` error message

tmabraham:
Fixing a minor typo, `num_process=...`  -->  `num_processes=...` in the error message when `num_processes` is `None` for `notebook_launcher`.


sgugger (approved):
Thanks for fixing!

PR #371, opened 2022-05-18
Add a utility for writing a barebones config file

muellerzr:
# Introduce barebones config file generator

## What does this add?

This PR adds a `write_basic_config` file that will write a barebones config file automatically setting the GPU's to be the maximum GPUs available (or CPU), and takes in one param for mixed precision

## Who is it for?

Folks who want to avoid running `accelerate config` and for tests that need to be ran with `accelerate launch`.

## Why is it needed?

Along with a request in the fastai integration https://github.com/fastai/fastai/pull/3646, I was working on fixing up the tests we currently run and was noticing some frustration when trying to setup and configure them to be launched on multi gpu. 

### User-facing:

Adds a `write_basic_config` config to `test_utils/testing.py`, that should be used in tests but users can optionally use it to setup a basic configuration if wanted. It **will not** override an already setup configuration.

## Basic Usage Example(s):

```python
from accelerate.test_utils import write_basic_config
write_basic_config("fp16")
```

You'd use it when you don't want to run `accelerate config` and don't care about TPU setup or a fine-grained configuration setup for multi-gpu

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -124,8 +124,8 @@ def write_basic_config(mixed_precision="no", save_location: str = default_json_c
             Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
         save_location (`str`, *optional*, defaults to `default_json_config_file`):
             Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`. Default
-            location is inside the huggingface cache folder as default_config.yaml, can be overriden by setting the
-            `HF_HOME` environmental variable.
+            location is inside the huggingface cache folder (`~/.cache/huggingface`) but can be overriden by setting the
+            `HF_HOME` environmental variable, followed by `accelerate/default_config.yaml`.
     """
     path = Path(save_location)
     path.parent.mkdir(parents=True, exist_ok=True)
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -124,7 +124,7 @@ def write_basic_config(mixed_precision="no", save_location: str = default_json_c
             Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
         save_location (`str`, *optional*, defaults to `default_json_config_file`):
             Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`. Default
-            location is inside the huggingface cache folder as default_config.yaml, can be overriden by setting the 
+            location is inside the huggingface cache folder as default_config.yaml, can be overriden by setting the
             `HF_HOME` environmental variable.
     """
     path = Path(save_location)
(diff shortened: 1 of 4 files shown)


src/accelerate/utils/other.py
```diff
@@ -109,3 +112,45 @@ def get_pretty_name(obj):
...
+    set CPU if it is a CPU-only machine.
+
+    Args:
+        mixed_precision (`str`, *optional*, defaults to "no"):
+            Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
+        save_location (`str`, *optional*):
+            Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`.
+    """
+    if save_location is None:
+        save_location = default_json_config_file
```
sgugger, line 129: This is the default already, so I don't think this part is useful.

src/accelerate/utils/other.py
```diff
@@ -109,3 +112,45 @@ def get_pretty_name(obj):
...
+def write_basic_config(mixed_precision="no", save_location: str = default_json_config_file):
+    """
+    Creates and saves a basic cluster config to be used on a local machine with potentially multiple GPUs. Will also
+    set CPU if it is a CPU-only machine.
+
+    Args:
+        mixed_precision (`str`, *optional*, defaults to "no"):
+            Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
+        save_location (`str`, *optional*):
+            Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`.
```
sgugger, line 126: The default should be indicated there.

sgugger (approved):
Thanks for adding this!

src/accelerate/utils/other.py
```diff
@@ -122,11 +122,9 @@ def write_basic_config(mixed_precision="no", save_location: str = default_json_c
     Args:
         mixed_precision (`str`, *optional*, defaults to "no"):
             Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
-        save_location (`str`, *optional*):
+        save_location (`str`, *optional*, defaults to "~/.cache/huggingface/accelerate/default_config.yaml"):
```
sgugger, line 125: This won't be the default location for everyone, so I'd just say defaults to the default config path and expand a bit in the docstring that it's inside the default cache of hugging face, and may add which env variable control it.

src/accelerate/utils/other.py
```diff
@@ -122,8 +122,10 @@ def write_basic_config(mixed_precision="no", save_location: str = default_json_c
     Args:
         mixed_precision (`str`, *optional*, defaults to "no"):
             Mixed Precision to use. Should be one of "no", "fp16", or "bf16"
-        save_location (`str`, *optional*, defaults to "~/.cache/huggingface/accelerate/default_config.yaml"):
-            Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`.
+        save_location (`str`, *optional*, defaults to `default_json_config_file`):
+            Optional custom save location. Should be passed to `--config_file` when using `accelerate launch`. Default
+            location is inside the huggingface cache folder as default_config.yaml, can be overriden by setting the 
+            `HF_HOME` environmental variable.
```
sgugger, line 128:
```suggestion
            location is inside the huggingface cache folder (`~/.cache/huggingface` but can be overriden by setting the 
            `HF_HOME` environmental variable) followed by `accelerate/default_config.yaml`.
```

PR #373, opened 2022-05-18
Refactor tests to use accelerate launch

muellerzr:
# Refactor tests to use accelerate launch

## What does this add?

This PR refactors all the example tests to use `accelerate launch` to be ran, adjusting any testing logic that needs to be done to compensate for this. Mostly changing the accelerate.print mocks to use string matching instead based on stdout from `subprocess.run`

## Why is it needed?

In order to have our tests be ran on more than one GPU, the tests needed to be refactored to allow for the `accelerate launch` command to be used via subprocess rather than calling it in python directly. 

### User-facing:

All of the example scripts now contain a small snippet denoting that a particular import is only used for testing, and should be ignored

### Internal structure:

`mocked_dataloaders` was migrated to the test utils.

### DIFF
--- a/examples/by_feature/checkpointing.py
+++ b/examples/by_feature/checkpointing.py
@@ -104,7 +104,7 @@ def get_dataloaders(accelerator: Accelerator, batch_size: int = 16):
 
 
 # For testing only
-if os.environ.get("USE_MOCKED_DATALOADERS", None) == "1":
+if os.environ.get("TESTING_MOCKED_DATALOADERS", None) == "1":
     from accelerate.test_utils.training import mocked_dataloaders
 
     get_dataloaders = mocked_dataloaders  # noqa: F811
--- a/examples/by_feature/checkpointing.py
+++ b/examples/by_feature/checkpointing.py
@@ -55,7 +55,6 @@ from transformers import (
 MAX_GPU_BATCH_SIZE = 16
 EVAL_BATCH_SIZE = 32
 
-
 def get_dataloaders(accelerator: Accelerator, batch_size: int = 16):
     """
     Creates a set of `DataLoader`s for the `glue` dataset,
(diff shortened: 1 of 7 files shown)


examples/by_feature/checkpointing.py
```diff
@@ -103,6 +103,13 @@ def collate_fn(examples):
     return train_dataloader, eval_dataloader
 
 
+# For testing only
+if os.environ.get("USE_MOCKED_DATALOADERS", None) == "1":
```
sgugger, line 107: Since it's for testing only, maybe add it in the name of the env variable? `TESTING_MOCKED_DATALOADERS` for instance.

examples/by_feature/fsdp_with_peak_mem_tracking.py
```diff
@@ -74,6 +74,13 @@ def __exit__(self, *exc):
         # print(f"delta used/peak {self.used:4d}/{self.peaked:4d}")
 
 
+# For testing only
+if os.environ.get("USE_MOCKED_DATALOADERS", None) == "1":
```
sgugger, line 78: Same here.

tests/test_examples.py
```diff
@@ -159,88 +112,89 @@ def test_cv_examples(self):
...
 
 
+@mock.patch.dict(os.environ, {"USE_MOCKED_DATALOADERS": "1"})
 class FeatureExamplesTests(TempDirTestCase):
     clear_on_setup = False
 
-    @mock.patch("checkpointing.get_dataloaders", mocked_dataloaders)
+    @classmethod
+    def setUpClass(cls):
+        super(FeatureExamplesTests, cls).setUpClass()
```
sgugger, line 121:
```suggestion
        super().setUpClass()
```
We're not in Python 2 ;-)

tests/test_examples.py
```diff
@@ -159,88 +112,89 @@ def test_cv_examples(self):
...
+        cls._tmpdir = tempfile.mkdtemp()
+        cls.configPath = os.path.join(cls._tmpdir, "default_config.yml")
+
+        write_basic_config(save_location=cls.configPath)
+        cls._launch_args = ["accelerate", "launch", "--config_file", cls.configPath]
+
+    @classmethod
+    def tearDownClass(cls):
+        super(FeatureExamplesTests, cls).tearDownClass()
```
sgugger, line 130:
```suggestion
        super().tearDownClass()
```

sgugger (approved):
Thanks for working on this! Good for me with some renames.

Issue #375, opened 2022-05-19
Shuffling of IterableDataset disabled when dataloader is passed to accelerator.prepare()


loubnabnl:
Hi, I have an issue with shuffling an `IterableDataset` that is passed to a dataloader in `accelerate`. I am trying to use `ShufflerIterDataPipe` from `torch.utils.data.datapipes.iter.combinatorics`.
As you can see below, it looks like the shuffling is disabled:

```python
from accelerate import Accelerator
from datasets import load_dataset
from torch.utils.data.dataloader import DataLoader
from torch.utils.data.datapipes.iter.combinatorics import ShufflerIterDataPipe

data = load_dataset("lvwerra/codeparrot-clean-valid", streaming=True, split="train")
shuffled_data = ShufflerIterDataPipe(data, buffer_size=100)
dataloader = DataLoader(shuffled_data, batch_size=4, shuffle=True)
print(f"Before accelerate prepare: {dataloader.dataset._shuffle_enabled}")

accelerator = Accelerator()
dataloader = accelerator.prepare(dataloader)
print(f"After accelerate prepare: {dataloader.dataset._shuffle_enabled}")
```

Output:

```
Before accelerate prepare: True
After accelerate prepare: False
```

I think it’s because `ShufflerIterDataPipe` returns a new Pytorch `IterableDataset` that is also an `IterDataPipe`. When such instance is passed to a Pytorch dataloader, for the shuffling to be enabled we must specify  `shuffle=True`  in the dataloader, the distinction is done [here](https://github.com/pytorch/pytorch/blob/d1bb420fb1087d680b7ad526536e14acbc2f18bf/torch/utils/data/dataloader.py#L260). I don't think this is handled during the dataloader preparation in `accelerate`.

```
Python 3.9.12
torch=1.11.0
accelerate=0.8.0

compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: MULTI_GPU
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
mixed_precision: fp16
num_machines: 1
num_processes: 2
use_cpu: false
```

lvwerra:
You mean `dispatch_batch`? That's `true` by default for iterable datasets, no?

sgugger:
[...]

PR #376, opened 2022-05-19
fix shuffling for ShufflerIterDataPipe instances

loubnabnl:
fixes https://github.com/huggingface/accelerate/issues/375

cc @sgugger

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -17,6 +17,7 @@ from typing import List, Optional, Union
 
 import torch
 from torch.utils.data import BatchSampler, DataLoader, IterableDataset
+from torch.utils.data.datapipes.iter.combinatorics import ShufflerIterDataPipe
 
 from packaging import version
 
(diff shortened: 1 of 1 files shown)


src/accelerate/data_loader.py
```diff
@@ -17,6 +17,7 @@
 
 import torch
 from torch.utils.data import BatchSampler, DataLoader, IterableDataset
+from torch.utils.data.datapipes.iter.combinatorics import ShufflerIterDataPipe
```
sgugger, line 20: This will break compatibility with older versions of PyTorch as I believe datapipes were just introduced in PyTorch 1.11.

src/accelerate/data_loader.py
```diff
@@ -326,12 +327,15 @@ class DataLoaderDispatcher(DataLoader):
     """
 
     def __init__(self, dataset, split_batches: bool = False, **kwargs):
+        shuffle = dataset._shuffle_enabled if isinstance(dataset, ShufflerIterDataPipe) else False
```
sgugger, line 330: There should be a versioning test of PyTorch first, then this instance check. :-)

sgugger (commented):
Looks good in principle! We just have to make a tiny bit more effort to keep supporting older PyTorch versions :-)

sgugger (approved):
Perfect, thanks a lot!

src/accelerate/data_loader.py
```diff
@@ -326,12 +326,21 @@ class DataLoaderDispatcher(DataLoader):
     """
 
     def __init__(self, dataset, split_batches: bool = False, **kwargs):
+        shuffle = False
+        if version.parse(torch.__version__) >= version.parse("1.11.0"):
```
sgugger, line 330:
```suggestion
        if version.parse(torch.__version__) >= version.parse("1.11.99"):
```
Just in case we get many micro releases ;-)

sgugger: Arf, review before coffee. I read > and thought for a minute it needed PyTorch 1.12, you can ignore :-)

PR #378, opened 2022-05-20
fix deepspeed zero stage-1

pacman100:
### What does this PR do?
1. In DeepSpeed Zero Stage-1, the offload_optimizer was not present in accelerate config and as such was being set to "None" whereas the deepspeed expects it to "none"


sgugger (approved):
Thanks for the fix!

PR #379, opened 2022-05-20
Better check for deepspeed availability

sgugger:
We are having issues in the tests of Transformers where there is a folder called `deepspeed` containing test files for DeepSpeed.
That folder gets interpreted as the "DeepSpeed" package by Accelerate, which then tries to import things.

This PR fixes that by making the test for `deepspeed` more robust (like we have for Datasets in Transformers).

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -57,8 +57,9 @@ def is_deepspeed_available():
     # AND checking it has an author field in the metadata that is HuggingFace.
     if package_exists:
         try:
-            _ = importlib_metadata.version("deepspeed")
             _deepspeed_metadata = importlib_metadata.metadata("deepspeed")
+            if _deepspeed_metadata.get("version", None) is None:
+                return False
             if _deepspeed_metadata.get("author", "") != "DeepSpeed Team":
                 return False
             return True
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/imports.py
```diff
@@ -45,7 +52,18 @@ def is_tpu_available():
 
 
 def is_deepspeed_available():
-    return importlib.util.find_spec("deepspeed") is not None
+    package_exists = importlib.util.find_spec("deepspeed") is not None
+    # Check we're not importing a "deepspeed" directory somewhere but the actual library by trying to grab the version
+    # AND checking it has an author field in the metadata that is HuggingFace.
+    if package_exists:
+        try:
+            _ = importlib_metadata.version("deepspeed")
```
muellerzr, line 60:
```suggestion
```
If I'm not mistaken this is unnecessary. You can do `_deepspeed_metadata.get("version")` if it's needed?

muellerzr (approved):
LG2M! Just one question

src/accelerate/utils/imports.py
muellerzr: `importlib_metadata.metadata` will also then raise the same error if it's not installed

PR #380, opened 2022-05-20
Refactor some parts of `load_checkpoint_in_model`  in utils

sgugger:
This PR refactors some parts of `load_checkpoint_in_model` as independent utils, so we can re-use them in Transformers.


muellerzr (approved):
LG2M!

PR #386, opened 2022-05-23
Host a notebook example

muellerzr:
# Notebook Example

## What does this add?

This PR adds a notebook example into the main accelerate codebase, rather than just in the transformers documentation. 

Also a small bugfix to the base cv example


examples/cv_example.py
```diff
@@ -41,7 +41,7 @@
 # in the readme for examples:
 # https://github.com/huggingface/accelerate/tree/main/examples
 #
-########################################################################
+# #######################################################################
```
sgugger, line 44:
```suggestion
########################################################################
```

sgugger (commented):
Would be better to put the notebook in the notebooks repo pretty please :-)

PR #391, opened 2022-05-24
Build Docker Images nightly

muellerzr:
# Build Docker Images Nightly
## What does this add?

This PR adds a workflow to publish two docker containers nightly: 

- [huggingface/accelerate-cpu](https://hub.docker.com/r/huggingface/accelerate-cpu)
- [huggingface/accelerate-gpu](https://hub.docker.com/r/huggingface/accelerate-gpu)

These will eventually be up-streamed and used in our nightly tests for testing multigpu, gpu, etc. 

I also modified the Dockerfile slightly to add a different end-user when launching the docker container for security reasons.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -2,7 +2,6 @@ name: Build Docker images (scheduled)
 
 on:
   workflow_dispatch:
-  pull_request:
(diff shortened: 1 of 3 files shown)


sgugger (approved):
Thanks for adding those!

PR #392, opened 2022-05-24
Better dispatch for submodules

sgugger:
While testing the `dispatch_model` features in [this PR](https://github.com/huggingface/transformers/pull/17401) I hit some edge case for model parallelism: when a submodule of a given model is never actually called in the forward pass (like the `h` of the GPT-2 model which is a module list which is iterated on in the forward but not directly called) the inputs aren't necessarily moved to the right device.

This PR adds a hook in all submodules to just move those inputs to the device, which fix this issue.

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -260,6 +260,24 @@ class AlignDevicesHook(ModelHook):
                     set_module_tensor_to_device(module, name, device, value=self.weights_map.get(name, None))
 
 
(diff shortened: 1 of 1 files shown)


LysandreJik (approved):
Perfect, should be very robust this way! Have you benchmarked the time difference between this implementation and the current one?

sgugger:
Not yet, will have a look before merging (though I don't expect much difference since there is not that many submodules).

Issue #393, opened 2022-05-25
ModuleNotFoundError: No module named 'packaging'


kakumarabhishek:
Hi,

I tried installing Accelerate both using `pip` and from source, and in both the cases, importing `accelerate` led to a `ModuleNotFoundError`.

### **Installing with pip**

```
$ pip install accelerate
Collecting accelerate
  Downloading accelerate-0.9.0-py3-none-any.whl (106 kB)
     |████████████████████████████████| 106 kB 5.4 MB/s 
Requirement already satisfied: numpy>=1.17 in ./miniconda3/envs/hfaccl/lib/python3.10/site-packages (from accelerate) (1.22.3)
Collecting pyyaml
  Downloading PyYAML-6.0-cp310-cp310-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_12_x86_64.manylinux2010_x86_64.whl (682 kB)
     |████████████████████████████████| 682 kB 17.9 MB/s 
Requirement already satisfied: torch>=1.4.0 in ./miniconda3/envs/hfaccl/lib/python3.10/site-packages (from accelerate) (1.11.0)
Requirement already satisfied: typing_extensions in ./miniconda3/envs/hfaccl/lib/python3.10/site-packages (from torch>=1.4.0->accelerate) (4.1.1)
Installing collected packages: pyyaml, accelerate
Successfully installed accelerate-0.9.0 pyyaml-6.0

$ python 
Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from accelerate import Accelerator
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/<PATH>/miniconda3/envs/hfaccl/lib/python3.10/site-packages/accelerate/__init__.py", line 7, in <module>
    from .accelerator import Accelerator
  File "/<PATH>/miniconda3/envs/hfaccl/lib/python3.10/site-packages/accelerate/accelerator.py", line 24, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```

### **Installing from source**

```
$ pip install git+https://github.com/huggingface/accelerate
Collecting git+https://github.com/huggingface/accelerate
  Cloning https://github.com/huggingface/accelerate to /tmp/pip-req-build-5_5o_ta0
[...]

PR #394, opened 2022-05-25
Add packaging to requirements

muellerzr:
# Add packaging to the project requirements

## What does this add?

This PR adds `packaging>=20.0` to the project requirements, which is the same pin as in [transformers](https://github.com/huggingface/transformers/blob/main/setup.py#L132)

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/393

## Why is it needed?

Current installations from source fail due to the `packaging` package needing to be installed in both accelerator.py and versions.py in huggingface/accelerate

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         ]
     },
     python_requires=">=3.6.0",
-    install_requires=["numpy>=1.17", "pyyaml", "torch>=1.4.0", "packaging>=20.0"],
+    install_requires=["numpy>=1.17", "packaging>=20.0", "pyyaml", "torch>=1.4.0"],
     extras_require=extras,
     classifiers=[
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -55,7 +55,7 @@
         ]
     },
     python_requires=">=3.6.0",
-    install_requires=["torch>=1.4.0", "pyyaml", "numpy>=1.17"],
+    install_requires=["numpy>=1.17", "pyyaml", "torch>=1.4.0", "packaging>=20.0"],
```
sgugger, line 58: If we're alphabetically sorting, let's really do it ;-)

muellerzr: Fixed, heh

sgugger (approved):

PR #395, opened 2022-05-25
Refactor version checking into a utility

muellerzr:
# Refactor version checking into a utility

## What does this add?

This PR adds two new functions:
- `compare_versions`
- `check_torch_version`

Which checks a particular version, querying with a string of the comparitor (such as "<" or ">=" or "!=")

## Why is it needed?

I was noticing we do a lot of checks with `version.parse` in the code, and we already had a versions submodule for getting the current torch version. As a result, it made sense to expand this utility further to have it check for all of our torch versions.

Also considering that most of them are comparing the torch version specifically, I added a `check_torch_version` which shorthands the query on the user.

### User-facing:

Nothing

### Internal structure:

Modifies all version checks to use `compare_versions` or `check_torch_version`

## Basic Usage Example(s):

For checking a non-torch library:
```python
from accelerate.utils import compare_versions
compare_versions("fastai", "<", "2.0.0")
```
For checking a torch version:
```python
from accelerate.utils import check_torch_version
check_torch_version("<=", "1.10")
```


sgugger (commented):
Nice refactor!

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
...
+def check_torch_version(operation: str, version: str):
+    """Compares the current PyTorch version to `version` with an `operation`
+
+    Args:
+        operation (`str`):
+            A string representation of an operator, such as ">" or "<="
+        version (`str`):
+            A string version of PyTorch
+    """
+    return check_torch_version(torch_version, operation, version)
```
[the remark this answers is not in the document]
muellerzr: Yup, nice catch!

src/accelerate/accelerator.py
```diff
@@ -173,7 +172,7 @@ def __init__(
             os.environ["USE_FSDP"] = "true"  # use FSDP if plugin is provided
 
         if os.environ.get("USE_FSDP", "false") == "true":
-            if version.parse(torch.__version__) < version.parse("1.12.0.dev20220418+cu113"):
+            if check_torch_version("<", "1.12.0.dev20220418+cu113"):
```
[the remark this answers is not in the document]
muellerzr: `is_torch_version` sounds better, will go with that.

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
     import importlib.metadata as importlib_metadata
 
 torch_version = parse(importlib_metadata.version("torch"))
+
+
+def compare_versions(library_or_version: Union[str, Version], operation: str, requirement_version: str):
+    """Compares `library_or_version` to `requirement_version` with an `operation`
+
+    Args:
+        library_or_version (`str`, `packaging.version.Version]`):
```
sgugger, line 35:
```suggestion
        library_or_version (`str` or `packaging.version.Version`):
```

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
     import importlib.metadata as importlib_metadata
 
 torch_version = parse(importlib_metadata.version("torch"))
+
+
+def compare_versions(library_or_version: Union[str, Version], operation: str, requirement_version: str):
+    """Compares `library_or_version` to `requirement_version` with an `operation`
```
sgugger, line 32:
```suggestion
    """
    Compares a library version to some requirement using a given operation.
```
I prefer to have complete English sentence describe the function :-)

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
...
+
+    Args:
+        library_or_version (`str`, `packaging.version.Version]`):
+            A library name or Version to check
+        operation (`str`):
+            A string representation of an operator, such as ">" or "<="
```
sgugger, line 38:
```suggestion
            A string representation of an operator, such as `">"` or `"<="`.
```

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
...
+    if operation not in STR_OPERATION_TO_FUNC.keys():
+        raise ValueError(f"`operation` must be one of {list(STR_OPERATION_TO_FUNC.keys())}, received {operation}")
+    operation = STR_OPERATION_TO_FUNC[operation]
+    if isinstance(library_or_version, str):
+        library_or_version = parse(importlib_metadata.version(library_or_version))
+    return operation(library_or_version, parse(requirement_version))
+
+
+def is_torch_version(operation: str, version: str):
+    """Compares the current PyTorch version to `version` with an `operation`
```
sgugger, line 51:
```suggestion
    """
    Compares the current PyTorch version to a given reference with an operation.
```

src/accelerate/utils/versions.py
```diff
@@ -23,3 +26,34 @@
...
+
+    Args:
+        operation (`str`):
+            A string representation of an operator, such as ">" or "<="
```
sgugger, line 55:
```suggestion
            A string representation of an operator, such as `">"` or `"<="`
```

sgugger (approved):
Left some nits to the doc, but otherwise looks great! Thanks for working on this!

PR #396, opened 2022-05-25
Include fastai in frameworks using Accelerate

muellerzr:
Adds fastai to the frameworks using accelerate section of the README


sgugger (approved):
Thanks for adding!

PR #398, opened 2022-05-25
Refactor offload_state_dict and fix in offload_weight

sgugger:
This PR refactors the `offload_state_dict` function using the two new utils introduced in #380 and fixes a small bug in one of them.


muellerzr (approved):
Very nice refactor!

PR #401, opened 2022-05-26
Hotfix all failing GPU tests

muellerzr:
# Hotfix all GPU tests and include new makefile commands

## What does this add?

This PR introduces a few hotfixes for the issues described in https://github.com/huggingface/accelerate/issues/402

Specifically it fixes the `Makefile` so that a user can run `make test_cpu` and `make test_cuda` rather than `make test` (and actually has them work)

## Why is it needed?

The solution proposed in #402 include separating out the `debug_launcher` related tests into their own group that they can then be ran separately when doing tests. This is a **hotfix** while we work towards figuring out how to solve the `debug_launcher` itself.


sgugger (approved):
LGTM as a hotfix!

Issue #403, opened 2022-05-26
mocked_dataloaders requires transformers and datasets


rfejgin:
I followed following the installation instructions on the README:
`pip install accelearate`

And then tried:
`accelerate config`
Result:
`ModuleNotFoundError: No module named 'datasets'`

It worked after I did a 
`pip install datasets`

Suggestion: either document the need to install `datasets` in the README, or add it as a dependency of the `accelerate` package.

muellerzr:
Can you please show the full stack trace? It shouldn't.

rfejgin:
Sure, here it is:
```
$ pip install accelerate
Collecting accelerate
  Downloading accelerate-0.9.0-py3-none-any.whl (106 kB)
     |################################| 106 kB 2.6 MB/s
Requirement already satisfied: numpy>=1.17 in /home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages (from accelerate) (1.20.3)
Requirement already satisfied: pyyaml in /home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages (from accelerate) (6.0)
Requirement already satisfied: torch>=1.4.0 in /home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages (from accelerate) (1.11.0)
Requirement already satisfied: typing-extensions in /home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages (from torch>=1.4.0->accelerate) (4.1.1)
Installing collected packages: accelerate
Successfully installed accelerate-0.9.0
$ accelerate config
Traceback (most recent call last):
  File "/home/rfejg/rfejg/miniconda3/envs/pt111/bin/accelerate", line 5, in <module>
    from accelerate.commands.accelerate_cli import main
  File "/home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages/accelerate/commands/accelerate_cli.py", line 22, in <module>
    from accelerate.commands.test import test_command_parser
  File "/home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages/accelerate/commands/test.py", line 20, in <module>
    from accelerate.test_utils import execute_subprocess_async
  File "/home/rfejg/rfejg/miniconda3/envs/pt111/lib/python3.8/site-packages/accelerate/test_utils/__init__.py", line 6, in <module>
    from .training import RegressionDataset, RegressionModel
[...]

PR #411, opened 2022-05-27
Move datasets and transformers to under func

muellerzr:
# Refactor test-specific imports for `mocked_dataloaders` under the function

## What does this add?

This PR changes when imports are called for importing datasets and transformers. 

## Who is it for?

closes https://github.com/huggingface/accelerate/issues/403

## Why is it needed?

Due to how imports are checked in python, currently the released Accelerate version on pypi technically requires datasets and transformers to be required. This PR remedy's this. I'd recommend a patch release afterwards. (So this wouldn't increase our minimum python from 3.6 *yet*)


sgugger (approved):
Thanks for fixing!

Issue #402, opened 2022-05-26
Issues with debug_launcher and CUDA


muellerzr:
I'm noticing CUDA re-initalization issues when running the tests on a GPU or Multi-GPU. These are not present when ran on a CPU. It seems to stem from the `debug_launcher` and as a result all tests that use it fail. 

Specifically, these seem to stem around the `AcceleratorState` and its persistence. 

## Problematic test: 
`test_cpu.py`
### Options I have explored:
- Make the CPU tests CPU-only. While this gets rid of one problem where CUDA was being initialized when it shouldn't be, it does not solve every problem (hotfix performed in https://github.com/huggingface/accelerate/pull/401)

## Problematic test:
`test_scheduler.py`
### Options I have explored:
- If this is the first test ran, then the tests will be able to be ran without issue. **However** this then causes headaches later as the `big_modeling_test.py` will fail. Specifically `test_cpu_offload`. 

## Potential hotfix
The hotfix for now is having `make tests` refactored to call everything *but* those two files first, then continue on to call the tests again with those two specific test files. If done this way, then the tests will all pass without issue. 

## To reproduce:

On a machine with at least one GPU, run:
```python
git clone https://github.com/huggingface/accelerate
cd accelerate
pip install -e .[test, test_trackers]
CUDA_VISIBLE_DEVICES="0" pytest -svx tests/ --ignore=tests/test_examples.py
```

@sgugger might potentially try and get to this later next week, I will also be exploring this more early next week. @pacman100 if you happen to have some availability to look at this as well it'd be much appreciated!

PR #413, opened 2022-05-31
Fix debug_launcher issues

muellerzr:
# Fix Debug Launcher + CUDA breakage with tests

## What does this add?

This PR introduces fixes for tests that run `debug_launcher`. They must be performed only on the CPU, as otherwise at some point if any previous tests were ran on the GPU, CUDA has been initialized for this instance of python and as a result torch will complain and break saying that a multiprocessing error has occured.

## Who is it for?

closes https://github.com/huggingface/accelerate/issues/402

## Why is it needed?

Currently running all of the tests will crash the CUDA process, due to it trying to reinitialize CUDA. From there we had the Makefile split into 2 separate calls to pytest, but given that this breakage (and since they really only run on the CPU), these tests are now under a `require_cpu` tag that will only run if CUDA is not available. 

> Note: If these tests are run *first*, it will not break. But for keeping it simple for the user we are going with this methodology

### User-facing:

Nothing

### Internal structure:

If any test should be ran on the GPU, decorate it with:

```python
from accelerate.test_utils import require_cpu

@require_cpu
def some_test_for_cpu_only(self):
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,8 @@ style:
 	python utils/style_doc.py src/accelerate docs/source --max_len 119
 	
 # Run tests for the library
-test_cpu:
+test:
 	python -m pytest -s -v ./tests/ --ignore=./tests/test_examples.py
-
-test_cuda:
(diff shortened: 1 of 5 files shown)


sgugger (approved):
Thanks a lot!

Issue #276, opened 2022-03-11
Issue template


LysandreJik:
A contributing guide was just merged, from which a lot may be taken to create an issue template: https://github.com/huggingface/accelerate/pull/254

Issue templates help contributors specify what maintainers want, helping issues get resolved quicker and better.

LysandreJik:
And now there's those cool YAML issues :smiley:

PR #415, opened 2022-05-31
Create an issue template for Accelerate

muellerzr:
# Introduce a bug template for Accelerate

## What does this add?

This PR adds a similar bug report template to that of [transformers](https://raw.githubusercontent.com/huggingface/transformers/main/.github/ISSUE_TEMPLATE/bug-report.yml) for users who want to open up a new bug, so that maintainers can have all the information they might need to help

## Why is it needed?

Providing a template ensures that the users are prompted to think about all the information the maintainers may need in order to help their problem as quickly and efficiently as possible.

## Issues this closes:

Closes https://github.com/huggingface/accelerate/issues/276


LysandreJik (approved):
Looks good to me!

PR #416, opened 2022-05-31
Introduce post-merge runners

muellerzr:
# Introduce post-merge commit CUDA workflow

## What does this add?

This PR adds a similar workflow to the nightly tests, that will run tests for Accelerate post a commit to main. The main difference is we do not run the slow tests so that if multiple PR's are accepted in rapid succession (or pushes to main in general), the queue is not hanging for ages to let everyone through. But, it is rapid enough to where if a push to main did break something in CUDA, we are made aware of it.


sgugger (approved):
Thanks for working on this!

PR #417, opened 2022-05-31
Switch to evaluate for metrics

sgugger:
This PR switches all example scripts to use `evaluate` instead of `datasets` for the metric computation. cc @lvwerra


muellerzr (approved):
Great!

PR #419, opened 2022-06-01
Speedup main CI

muellerzr:
# Speed up main GitHub CI

## What does this add?

This PR speeds up the time for the github CI tests to pass by ~2.5x through a few basic optimizations:

### Adjustments to the Makefile
Four new test conditionals were added to the `Makefile` based on groupings of how each example test needs to be ran, and how their topics relate to each other:  
- `test_example_differences`: The quick method that checks if the .py files are missing any parts in the examples/ dir
- `test_checkpoint_step`: tests relating to checkpoint by step and load checkpoint by step
- `test_checkpoint_epoch`: tests relating to checkpoint by epoch and loading checkpoint by epoch
- `test_rest`: Everything that did not fall under the previous three

### Use setup-python@v3 and cache@v3
By using these two specific actions, we can speed up the time it takes to setup the python environment, cache the pip installation, and also cache the model downloads. 

### Total Speedup
* Main tests: 2m42s -> 1m16s (~2x speedup)
* Example tests: 9m5s -> ~(3m47s, 4m20s) (~2.1-2.4x speedup)
(This also counts as the overall time taken)

Without the cache, it's ~8.5 minutes on that first commit/run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,34 +1,44 @@
 name: Run Tests
 
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
+
+env:
+  HF_HOME: ~/hf_cache
 
 jobs:
-  test:
+  run-tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        test-kind: [
+          test,
+          test_example_differences,
+          test_checkpoint_step,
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Very nice speed-up, thanks a lot!

PR #421, opened 2022-06-01
Enable slow tests during nightly

muellerzr:
# What does this PR do?

This PR enables the slow tests to be ran as part of the nightly builds. These seem to have gotten fixed during https://github.com/huggingface/accelerate/pull/420 as shown in https://github.com/huggingface/accelerate/runs/6697809890?check_suite_focus=true

(Failure in that runner is unrelated to the slow tests, and the multi-gpu one ran before the new docker image was out).

Finally stabilized our full CI test suite outside TPUs ❤️

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,12 +5,14 @@ on:
   schedule:
     - cron: "0 2 * * *"
 
+env:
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for fixing the typo!

PR #424, opened 2022-06-03
Introduce a Dependency Checker to trigger new Docker Builds on main

muellerzr:
# Introduce a Dependency Checker to trigger new Docker Builds on main

## What does this add?

This PR introduces a new workflow that will do a few things:

1. Introduces a workflow that will check if the `setup.py` was adjusted in a pull request
2. Adjusts `build-docker-images` to be triggered by a `workflow_call`, allowing `check_dependencies` to trigger it
3. Adjusts the `on-merge` workflow to be triggered on main by the completion of `check_setup`, rather than a push directly. This will ensure that the tests will be passing on main if a new version of anything is introduced with the other runners + nightlies

## Why is it needed?

A very annoying frustration I've seen over the last few months in the transformers repo is people freaking out when the runners are a failure, due to a new dependency introduced. This solves this problem, and I'll be building a similar workflow for transformers on Monday. (This was just the "test")

cc @ydshieh as this'll be coming to transformers afterwards 😉


utils/check_setup.py
```diff
@@ -0,0 +1,97 @@
...
+    for c in comments:
+        if c.user.login == "github-actions[bot]" and 'This PR modifies `setup.py`.' in c.body:
+            comment = c
+            break
+    original_diff = re.search(r'```([^`]*)```', comment.body).group(0)
+    new_body = comment.body.replace(original_diff, diff)
+    comment.edit(new_body)
+
+def main():
+    g = Github(os.environ["GITHUB_TOKEN"])
```
sgugger, line 86: This is not set anywhere, so I don't think this will work (also pashing the github token on a PR is not a trivial thing as you don't get secrets on pull requests. I'd just remove this whole writing comments thingy.

.github/workflows/check_dependencies.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Was setup changed 
+        id: was_changed
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            if [ `basename "${file}"` = "setup.py" ]; then
+              echo ::set-output name=changed::"1"
+            fi
+          done
+
+  write-message:
```
sgugger, line 38: I'd remove this job personally.

sgugger (approved):
Very nice idea, maybe a bit over-engineered? I don't think we need a bot comment to say the docker image is being update, it's just going to happen on the push and that's enough.

.github/workflows/check_dependencies.yml
```diff
@@ -0,0 +1,41 @@
+name: Check for dependency modification
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
ydshieh, line 5: Is there any reason that we should run this workflow `on pull_request`?

.github/workflows/check_dependencies.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            if [ `basename "${file}"` = "setup.py" ]; then
+              echo ::set-output name=changed::"1"
+            fi
+          done
+          
+  build-docker-containers:
+    needs: check-for-setup
+    if: (github.event_name == 'push') && (needs.check-for-setup.outputs.changed == '1')
```
ydshieh, line 40: Very nice - we should build the image only for the push event (to `main`).

.github/workflows/on-merge.yml
```diff
@@ -1,10 +1,11 @@
 name: Self-hosted runner (push to "main")
 
 on:
+  workflow_run:
+    workflows: ["Check for dependency modification"]
+    branches: ["main"]
+    types: [completed]
```
ydshieh, line 7: I was thinking that `completed` means the CI will be run only if the workflow `Check for dependency modification` is successful.

After reading this [doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow), it is not the case. And this is good, as we want to run the push CI anyway (even if something wrong in the setup check).

ydshieh (approved):
Great check! Left a comment about why we check setup on `pull_request`.
Thank you, @muellerzr !

Would be nice if you can test this PR works expected (change setup.py, for example). You might need to create a branch like `test_ci_XXX`, and modify the workflow files to be triggered on the experimentation branch(es). This is just a suggestion though. 

(sorry, I guess I am the one who freaks out, haha)

muellerzr:
@ydshieh you can see my almost 1,000 debug logs here: https://github.com/muellerzr/accelerate/actions

**Very** *very* comfortable with this working

.github/workflows/on-merge.yml
muellerzr: Yup exactly

.github/workflows/check_dependencies.yml
muellerzr: The docker yaml already checks main, do we trigger it here, it checks that, and it's skipped/uncalled if it's not.

.github/workflows/check_dependencies.yml
muellerzr: This was due to the overengineering part. We definitely could just trigger it on a push to main.

.github/workflows/check_dependencies.yml
ydshieh: Yeah, I meanly mean `push` in this remark, and you are doing it as what I would expect. (`to main` is to make it more clear only 😄 )

LysandreJik (approved):
LGTM!

PR #428, opened 2022-06-07
Introduce `no_sync` context wrapper + clean up some more warnings for DDP

muellerzr:
# Introduce `no_sync` context wrapper

## What does this add?

- This PR adds a `no_sync` wrapper that shadows `torch.nn.parallel.DistributedDataParallel.no_sync` but will do nothing if we're not in a DDP process
- It also refactors launch args for `torchrun` into a separate util to make reusing it in tests + launch easier

## Why is it needed?

This lets us use Accelerator eventually for gradient accumulation

### User-facing:

* Introduces `Accelerator.no_sync(model)` to perform the equivalent of `model.no_sync()` (if in ddp, if not it does nothing)

### Non-user facing:
- Introduced tests that mimic the official tests for `no_sync` in PyTorch to ensure those work, and CPU tests to make sure that when not in DDP it's a `noop` context manager


src/accelerate/accelerator.py
```diff
@@ -355,6 +356,22 @@ def _goes_first(self, is_main):
...
+        A context manager to disable gradient synchronizations across DDP processes by calling
+        `torch.nn.parallel.DistributedDataParallel.no_sync`.
+
+        If `model` is not in DDP, this context manager does nothing
+
+        Args:
+            model (`torch.nn.Module`):
+                PyTorch Module that was prepared with `Accelerator.prepare`
+        """
+        context = getattr(model, "no_sync", contextlib.nullcontext)
```
sgugger, line 371: I would double check we are in a distributed state on top of this.

sgugger (commented):
Thanks for working on this. I don't think this feature needs to be added to the base test `accelerate test`, it should just be tested independently in `tests.

muellerzr:
@sgugger good for a rereview, slow test logs can be seen here passing: https://github.com/huggingface/accelerate/runs/6794400970?check_suite_focus=true

PR #429, opened 2022-06-08
Add init to test_utils/scripts

muellerzr:
Fixes missing `__init__.py`. Confused as to why the tests didn't raise the same error as transformers. 

@sgugger


sgugger (approved):
Thanks for the quick fix!

PR #431, opened 2022-06-08
Create Gradient Accumulation Example

muellerzr:
# New Gradient Accumulation example

## What does this add?

This PR introduces an example showcasing the new `Accelerator.no_sync` capability through using it on gradient accumulation


sgugger (approved):
Perfect, thanks for adding this!

PR #405, opened 2022-05-27
DeepSpeed Revamp

pacman100:
### What does this PR do?
1. Support for DeepSpeed Config File
2. Support for gradient clipping and offloading params w/o config file
3. Fixing scheduler w/o config file
4. Simplifying model, optimizer and scheduler wrappers by removing stale code of current wrappers and directly relying on DeepSpeed.
5. Using HFDeepSpeedConfig when in Zero Stage-3 as per user consent to handle Deepspeed ZeRO-3 param gathering and automatically splitting the model onto multiple gpus during `from_pretrained` call

ToDo:
- [x] Write Tests
- [x] Redo experiments from [Testing Trainer and Accelerate Integration of DeepSpeed](https://wandb.ai/smangrul/Accelerate-MRPC-DeepSpeed/reports/Testing-Trainer-and-Accelerate-Integrations-of-DeepSpeed--VmlldzoyMDMwNzE1?accessToken=xz2scw676qasp8xta96ol11kcodxygg4dmploganrnyj2167t1vratbp12wlt8v2) to see if gaps between Accelerate integration and Trainer integrations are fixed.
- [x] ZeRo Stage-3 inference
- [x] Adding example leveraging config-file and related saving and loading of model when using deepspeed.
- [x] Documentation

PR #432, opened 2022-06-09
Migrate HFDeepSpeedConfig from trfrs to accelerate

pacman100:
### What does this PR do?
1. Migrates HFDeepSpeedConfig from transformers repo to accelerate repo as it is generic enough and specific bits remain in transformers.
2. Fix affected code and test; refactoring and improving few things.


src/accelerate/utils/dataclasses.py
```diff
@@ -215,7 +213,12 @@ class DeepSpeedPlugin:
     This plugin is used to integrate DeepSpeed.
     """
 
-    config_file: str = field(default=None, metadata={"help": "Path to the DeepSpeed config file."})
+    hf_ds_config: Any = field(
```
sgugger, line 216: This is breaking. While the API was marked as experimental and we can do such a change, I'd like to understand why we can't keep the old argument name here.

sgugger (commented):
LGTM except the change of argument name. Can we find a way to keep backward compatibility on this?
Also would it affect old accelerate configs of users (in which case we should add a conversion old config to new config to make their lives easier)

src/accelerate/utils/dataclasses.py
pacman100: Hello, this isn't breaking change as this feature was added in PR from couple days ago #405 . It isn't part of the latest release we have and as such won't be a breaking change. The renaming was done to clearly explain that it isn't just limited to file and can accept either file, dict or `HFDeepSpeedConfig` Object.

sgugger: Oh okay then, I thought it had been there for longer.

src/accelerate/state.py
```diff
@@ -193,3 +194,47 @@ def __repr__(self):
...
+        if isinstance(hf_ds_config, HfDeepSpeedConfig):
+            AcceleratorState._shared_state["deepspeed_plugin"] = DeepSpeedPlugin(
+                hf_ds_config=hf_ds_config, zero3_init_flag=True
+            )
+        else:
+            raise ValueError("`hf_ds_config` must be an instance of `accelerate.utils.HfDeepSpeedConfig`")
+
+    @staticmethod
+    def unset_deepspeed_config():
+        AcceleratorState._shared_state["deepspeed_plugin"] = None
```
sgugger, line 240: Let's check if the key exists before setting it a None?

sgugger (approved):
Nice new utilities! Could you add docstrings on those four new static methods?

src/accelerate/state.py
```diff
@@ -193,3 +194,47 @@ def __repr__(self):
...
+                or ds_config["train_micro_batch_size_per_gpu"] == "auto"
+            ):
+                ds_config["train_micro_batch_size_per_gpu"] = 1
+            if ds_config["train_batch_size"] == "auto":
+                del ds_config["train_batch_size"]
+        return ds_config
+
+    @staticmethod
+    def set_deepspeed_config(hf_ds_config):
+        from .utils import DeepSpeedPlugin, HfDeepSpeedConfig
```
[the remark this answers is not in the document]
pacman100: setting deepspeed_plugin doesn't use any deepspeed functionality; Accelerate and Trainer check DeepSpeed availability at different places. Should we still check deepspeed availability here?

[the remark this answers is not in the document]
sgugger: Oh I thought there was some DeepSpeed import in that file. Okay to leave as is.

src/accelerate/state.py
sgugger: As long as you're sure it won't make any deepspeed related stuff break, fine by me.

sgugger (approved):

Issue #436, opened 2022-06-10
ImportError: cannot import name 'ShardingStrategy' from 'torch.distributed.fsdp.fully_sharded_data_parallel


h2bit:
### System Info

```Shell
- `Accelerate` version: 0.9.0
- Platform: Linux-5.15.0-35-generic-x86_64-with-glibc2.17
- Python version: 3.8.13
- Numpy version: 1.22.3
- PyTorch version (GPU?): 1.11.0 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: FSDP
        - mixed_precision: fp16
        - use_cpu: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {'min_num_params': 100000000.0, 'offload_params': True, 'sharding_strategy': 1}
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

My code is in `accelerator.py` [accelerator.zip](https://github.com/huggingface/accelerate/files/8875389/accelerator.zip).

I followed the example in ["Fully Sharded Data Parallel" ](https://huggingface.co/docs/accelerate/fsdp), 
1. I configured accelerate environment by  `accelerate config`
> In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
> Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 2
> How many different machines will you use (use more than 1 for multi-node training)? [1]: 1
> Do you want to use DeepSpeed? [yes/NO]: NO
> Do you want to use FullyShardedDataParallel? [yes/NO]: yes
> What should be your sharding strategy ([1] FULL_SHARD, [2] SHARD_GRAD_OP)? [1]: 1
> Do you want to offload parameters and gradients to CPU? [yes/NO]: yes
> What should be your FSDP's minimum number of parameters for Default Auto Wrapping Policy? [1e8]: 
> How many GPU(s) should be used for distributed training? [1]:2
[...]

PR #437, opened 2022-06-10
fix fsdp PyTorch version dependency

pacman100:
### What does this PR do?

1. Check was happening after the point of using FSDP functionality. Fixed it t be before using any FSDP functionality
2.  Issue  #436  would not happen because a value error would be raised specifying that FSDP requires PyTorch nightlies.


muellerzr (approved):
Thanks!

sgugger (approved):
Thanks for fixing!

PR #441, opened 2022-06-11
Revamp TPU internals to be more efficient

muellerzr:
# Revamp TPU internals

## What does this add?

- Make `prepare_model` use [`MpModelWrapper`](https://pytorch.org/xla/master/index.html#torch_xla.distributed.xla_multiprocessing.MpModelWrapper) to distribute the model across all devices efficiently when used
- Changes `prepare_dataloader` to create an [`MpDeviceLoader`](https://github.com/pytorch/xla/blob/master/torch_xla/distributed/parallel_loader.py#L171) allowing for the dataloaders to be more efficient. 
- Changes `DataLoaderShard` to no longer do `xm.mark_step`, `MpDeviceLoader` will handle this for us. Instead if on TPU we set the device as `None` in `prepare_dataloader` letting `MpDeviceLoader` take over with the `device` when needed. 
- Allow for FP16 and BF16 precision types on the TPU via AcceleratorState since they are now supported

## Who is it for?

- Users of TPUs

## Why is it needed?

We currently have a number of "bad practices" in TPU handling due to improvements in the xla API. Here are some benchmarks I ran on a high-memory colab TPU instance with the nlp example script:

**Baseline**:
- Post warmup: 56 seconds

**W/ MpModelWrapper and MpDeviceLoader**:
- Post warmup: 48 seconds

**W/ previous and `default_tensor_type` change**:
- Post warmup: 34 to 36 seconds

Roughly a 40% boost to speed once this is all done. I also saw some speed increases on the initial launch as well anecdotally, but I did not time them.

I also saw a 2x speedup when using the new DataLoader class vs the old one

## About `default_tensor_type`:
Though we do set the device to `bf16` which helps with automatically converting the tensors to the right types, if we add `torch.set_default_tensor_type('torch.FloatTensor')` there is a considerable speedup when it comes to training TPUs. 

I'm a bit unsure as to where it would be best to put this, as either something hidden when we initialize the Accelerator or as a util that should get called when you are training on TPUs, open to ideas

This is pretty stable and considered as "good practices" when running on TPUs w/ XLA, so it's unlikely these will change. However after this PR a subsequent PR will be opened to change the `nlp_example` and `cv_example` notebook, as another best practice is to declare the model outside of `xm.spawn`. This includes the internals to make the model as memory efficient as we can, so that will be the last stage needed.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -712,17 +712,13 @@ class Accelerator:
         return tuple(result)
 
     def prepare_data_loader(self, data_loader):
-        if self.distributed_type == DistributedType.TPU and self.num_processes > 1:
-            device_placement = False
-        else:
-            device_placement = self.device_placement
         return prepare_data_loader(
             data_loader,
             self.device,
             num_processes=self.num_processes,
             process_index=self.process_index,
             split_batches=self.split_batches,
-            put_on_device=device_placement,
+            put_on_device=self.device_placement if self.distributedType != DistributedType.TPU else False,
             rng_types=self.rng_types.copy(),
             dispatch_batches=self.dispatch_batches,
         )
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -21,9 +21,10 @@ import warnings
 from contextlib import contextmanager
 from copy import deepcopy
 from typing import List, Optional, Union
-from accelerate.accelerate.src.accelerate.utils.imports import is_tpu_available
 
 import torch
+
+from accelerate.accelerate.src.accelerate.utils.imports import is_tpu_available
 
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
 from .data_loader import prepare_data_loader
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -70,7 +70,6 @@ if is_deepspeed_available():
     )
 
 if is_tpu_available():
-    import torch_xla.core.xla_model as xm
     import torch_xla.distributed.xla_multiprocessing as xmp
 
 logger = get_logger(__name__)
@@ -384,9 +383,7 @@ class Accelerator:
         """
         Use in replacement of `print()` to only print once per server.
         """
-        if self.distributed_type == DistributedType.TPU:
-            xm.master_print(*args, **kwargs)
-        elif self.is_local_main_process:
+        if self.is_local_main_process:
             print(*args, **kwargs)
 
     def _prepare_one(self, obj, first_pass=False):
@@ -553,7 +550,7 @@ class Accelerator:
             else:
                 model.forward = torch.cuda.amp.autocast()(model.forward)
             model.forward = convert_outputs_to_fp32(model.forward)
-        if self.distributed_type == DistributedType.TPU:
+        if self.distributed_type == DistributedType.TPU and self.num_processes > 1:
             model = xmp.MpModelWrapper(model).to(self.device)
         return model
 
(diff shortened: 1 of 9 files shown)


tmabraham:
Thanks for fixing up the TPU support with best practices! Looks great to me! 🔥

src/accelerate/accelerator.py
```diff
@@ -546,6 +553,8 @@ def prepare_model(self, model):
             else:
                 model.forward = torch.cuda.amp.autocast()(model.forward)
             model.forward = convert_outputs_to_fp32(model.forward)
+        if self.distributed_type == DistributedType.TPU:
+            model = xmp.MpModelWrapper(model).to(self.device)
```
sgugger, line 557: No, this shouldn't always be done IMO. This is an optimization for Colab/notebooks so we should only do it when we arrive at this code after a `notebook_launcher`.

Or be ready to benchmark the scripts on TPU to show that this does not affect performance in any way, but usually optimizations for memory come at a cost for speed.

src/accelerate/data_loader.py
```diff
@@ -565,15 +563,19 @@ def prepare_data_loader(
...
-    return DataLoaderShard(
-        new_dataset,
-        device=device if put_on_device else None,
-        batch_sampler=new_batch_sampler,
-        rng_types=rng_types,
-        generator=generator,
-        **kwargs,
-    )
+    if state.distributed_type == DistributedType.TPU:
+        return xpl.MpDeviceLoader(dataloader, device)
```
sgugger, line 580: Same, would like to see some benchmarks on launching a script on a TPU VM to make sure this does not affect speed.

sgugger (commented):
Thanks for your PR and running the benchmarks in Colab. Could you link to the documentation that shows those are best practices recommended by the torch XLA team? I haven't seen anything personally.

I'd like to make sure those changes are not speeding up the experience in Colab at the cost of the experience on TPU machines, so I'd like the same benchmarks to be run on a TPU VM before fully approving :-)

src/accelerate/accelerator.py
muellerzr: Here's the docs on [MpModelWrapper](https://pytorch.org/xla/release/1.11/index.html#torch_xla.distributed.xla_multiprocessing.MpModelWrapper), but here's their comment on it:

> This class should be used together with the spawn(…, start_method=’fork’) API to minimize the use of host memory. Instead of creating models on each multiprocessing process, hence replicating the model’s initial host memory, the model is created once at global scope, and then moved into each device inside the spawn() target function. This method has two advantages. First it uses only one copy of the memory pages to host the original model weights, and second it serializes the move of the wrapped model into each device, by lowering the load onto the system memory during the process.

Will perform benchmarks today and tomorrow

sgugger: Yes, and we don't use that spawn method when launching a script on TPU, hence why I don't think we should use it if we don't come from a colab/notebook here :-)

muellerzr: Makes sense! Will test and report back if there's a loss at all :)

muellerzr:
@sgugger this PR also modifies the TPU selection in `AcceleratorState` to actually use mixed precision types.

- [FP16 Support](https://github.com/pytorch/xla/issues/1936)
- [BF16 Support](https://pytorch.org/xla/release/1.11/index.html#xla-tensors-and-bfloat16)

muellerzr:
We're passing 😄 
image

sgugger (approved):
Thanks for revamping this!

src/accelerate/state.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         self.__dict__ = self._shared_state
         if parse_flag_from_env("USE_CPU"):
             cpu = True
+        self.use_fork = parse_flag_from_env("USE_LAUNCHER", 0)
```
sgugger, line 78:
```suggestion
        self.fork_launched = parse_flag_from_env("FORK_LAUNCHED", 0)
```
Might be clearer this way.

muellerzr: Makes sense!

PR #442, opened 2022-06-13
Offload and modules with unused submodules

sgugger:
This PR deals with models with unused submodules combined with offload (on CPU and disk). By unused submodules, we mean a something like the `ModuleWithUnusedSubModules` introduced in the test file in this PR: a module that defines a submodule, but does not call it during its forward, just does some operations using its weights.

Currently, the offload will fail for such modules because the weights are fetched in a pre-forward hook and the forward method is never called. This PR fixes that by introducing a new argument named `load_all_weights_classes` (if you have any better suggestion, please let me know as I'm not super happy with this name) which contains the name of such submodules used (like `no_split_module_classes` contains the names of the modules that shouldn't be split across devices).

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import functools
-from typing import Dict, Mapping, Optional, Union
+from typing import Dict, List, Mapping, Optional, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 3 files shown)


src/accelerate/hooks.py
```diff
@@ -371,6 +411,11 @@ def attach_align_device_hook_on_blocks(
             Whether or not to include the associated module's buffers when offloading.
         module_name (`str`, *optional*, defaults to `""`):
             The name of the module.
+        load_all_weights_classes (`List[str]`, *optional*):
```
muellerzr, line 414:
```suggestion
        preloaded_modules (`List[str]`, *optional*):
```

What about something such as `preloaded_modules`?

muellerzr (approved):
Thanks! The fix makes sense to me, left one naming suggestion/alternative

src/accelerate/hooks.py
sgugger: Maybe `preloaded_module_classes` to be similar to `no_split_module_classes`? I like `preloaded` a lot, thanks for the suggestion!

muellerzr: Works with me!

pacman100 (commented):
In `attach_align_device_hook_on_block`, I think below change is needed

```diff
if not isinstance(execution_device, Mapping):
-        execution_device = {key: offload for key in offload.keys()}
+        execution_device = {key: execution_device for key in offload.keys()}
``` 

Left a minor suggestion too.
Apart from that everything LGTM!

PR #456, opened 2022-06-21
Better checks for if a TPU device exists

muellerzr:
This PR fixes a situation where `torch_xla` is installed but a TPU device isn't present. 

This originally would cause the import to pass, even though a TPU device isn't found. Instead we combine this with a quick `xm.xla_device()`, that raises a `RuntimeError` if not on a TPU.

I've tested this on a CPU and TPU version of Colab. 

cc @sgugger


sgugger (approved):
Thanks a lot!

Issue #451, opened 2022-06-18
Cannot switch back to CPU training aftering doing TPU training


nalzok:
### System Info

```Shell
- `Accelerate` version: 0.10.0
- Platform: Linux-5.13.0-1023-gcp-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.22.4
- PyTorch version (GPU?): 1.11.0a0+gitbc2c6ed (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: TPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

First, configure Accelerate to do TPU training

```bash
$ pipenv run accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 3
What is the name of the function in your script that should be launched in all parallel scripts? [main]:
How many TPU cores should be used for distributed training? [1]:
```

Next, run the example script

```bash
$ pipenv run accelerate launch accelerate/examples/nlp_example.py
... output omitted ...
```

Then, configure Accelerate to do CPU training

```bash
$ pipenv run accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 1
How many different machines will you use (use more than 1 for multi-node training)? [1]: 1
How many CPU(s) should be used for distributed training? [1]:96
Do you wish to use FP16 or BF16 (mixed precision)? [NO/fp16/bf16]: NO
```

Finally, run the example script again

```bash
$ pipenv run accelerate launch accelerate/examples/nlp_example.py
... output omitted ...
```

### Expected behavior

```Shell
[...]

PR #457, opened 2022-06-21
Make TPU usage actually togglable

muellerzr:
# Make TPU usage toggle-able even if TPUs are available

## What does this add?

This PR introduces `AcceleratorState.use_tpu` and `Accelerator.use_tpu`.

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/451

## Why is it needed?

This new property combines `is_tpu_available()` and if `DistributedType.TPU` is being used. Otherwise we get situations like https://github.com/huggingface/accelerate/issues/451 where you can't turn off the TPU usage if wanted. Also any usage of checks for TPU's from `if distributed_type == DistributedType.TPU` have now been replaced with this new check when applicable. 

### User-facing:

Adds a new `Accelerator.use_tpu`

## Basic Usage Example(s):

```python
# I want some code run on the CPU in a TPU environment
accelerator = Accelerator(use_cpu=True)
if accelerator.use_tpu:
  print("Hello there!")
else:
  print("Goodbye there") # Actually gets called
```

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -269,6 +269,10 @@ class Accelerator:
             self.rng_types = ["torch"] if is_torch_version("<=", "1.5.1") else ["generator"]
 
     @property
+    def use_tpu(self):
+        return self.state.use_tpu
+
+    @property
     def distributed_type(self):
         return self.state.distributed_type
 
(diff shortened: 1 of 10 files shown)


sgugger (commented):
I don't understand what this PR changes in practice, as it just changes all the tests of `self.distributed_type == DistributedType.TPU` by `is_tpu_available() and self.distributed_type == DistributedType.TPU`.

But in the postinit of the `AcceleratorState`,  `self.distributed_type` is set to `DistributedType.TPU` only when `is_tpu_available()` is `True` 🤔

PR #458, opened 2022-06-21
Raise a clear warning if a user tries to modify the AcceleratorState

muellerzr:
# Raise warnings if attempting a modification of the `AcceleratorState`

## What does this add?

This PR adds a guard if the user tries to pass a different `mixed_precision` or `cpu` flag to `Accelerator()` after the `AcceleratorState` has been initialized once.

## Who is it for?

Users of Accelerate

## Why is it needed?

Currently if a modification is attempted to be made, it gets ignored leaving the user confused as to why they passed in a new mixed precision type, or told the Accelerator to run on the CPU, when in actually it has not due to the AcceleratorState being built once. 

### User-facing:
Nothing

### Internal structure:

A new `_check_initialized` function has been included in the `AcceleratorState` that fills a clear string template of what they tried to do, why they cannot do it, and how to achieve their desired behavior.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -205,7 +205,7 @@ class AcceleratorState:
             template = Template(
                 "AcceleratorState has already been initialized and cannot be changed, restart your runtime completely and pass `$flag` to `Accelerate()`."
             )
-            if cpu and not self.device == "cpu":
-                raise ValueError(template.substitute(flag="cpu=True"))
+            if cpu and self.device.type != "cpu":
+                raise ValueError(template.substitute(flag=self.device))
             if mixed_precision is not None and mixed_precision != self.mixed_precision:
                 raise ValueError(template.substitute(flag=f"mixed_precision='{mixed_precision}'"))
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -198,9 +198,10 @@ class AcceleratorState:
     def use_fp16(self):
         return self.mixed_precision != "no"
 
-    def _reset(self):
-        "Resets `_shared_state`, used internally"
-        self._shared_state = {}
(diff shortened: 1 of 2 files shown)


src/accelerate/state.py
```diff
@@ -196,3 +198,14 @@ def __repr__(self):
     @property
     def use_fp16(self):
         return self.mixed_precision != "no"
+
+    def _check_initialized(self, mixed_precision=None, cpu=None):
+        "Checks if a modification is trying to be made and the `AcceleratorState` has already been initialized"
+        if getattr(self, "initialized", False):
+            template = Template(
+                "AcceleratorState has already been initialized and cannot be changed, restart your runtime completely and pass `$flag` to `Accelerate()`."
+            )
```
sgugger, line 207: I don't understand why we are using this class, when you can just do a regular string with flag between curly brackets and then use the format method.

muellerzr: Definitely not because I forgot strings could do this... 👀

src/accelerate/state.py
```diff
@@ -196,3 +197,17 @@ def __repr__(self):
     @property
     def use_fp16(self):
         return self.mixed_precision != "no"
+
+    @classmethod
```
sgugger, line 201: Should be a `staticmethod` no?

src/accelerate/state.py
```diff
@@ -196,3 +197,17 @@ def __repr__(self):
...
+    def _reset_state(cls):
+        "Resets `_shared_state`, is used internally and should not be called"
+        cls._shared_state = {}
```
sgugger, line 204: And here you should change the value of `AcceleratorState._shared_state`

Issue #454, opened 2022-06-19
Disk offload fails with bfloat16 weights


noamwies:
### System Info

```Shell
- `Accelerate` version: 0.10.0
- Platform: Linux-5.10.104-aufs-3-x86_64-with-debian-10.10
- Python version: 3.7.3
- Numpy version: 1.21.6
- PyTorch version (GPU?): 1.11.0+cu113 (True)
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```
from transformers import AutoModelForCausalLM, AutoTokenizer

model = AutoModelForCausalLM.from_pretrained("bigscience/bloom", device_map="auto",  cache_dir="/mnt/local-ssd/hugginface_cache", offload_folder="/mnt/local-ssd/offload_folder", max_memory={"cpu": 17179869184})
```

![image](https://user-images.githubusercontent.com/3121971/174493108-b363a018-725b-4dd4-8bd7-dd9936443761.png)

### Expected behavior

```Shell
This script should Load the Bloom language model, but instead raises exception.
```

noamwies:
Thanks!
To be honest, I have already fixed the issue locally. My main motivation was to run Bloom with the least amount of resources possible, so I did not want to convert into `float32`. As an alternative, I used PyTorch's `tensor.view(torch.int16)` to reinterpret the underlined data as `int16` before converting to numpy, and then after loading the `int16` tensor once again, I reinterpret the underlined data with `tensor.view(torch.bfloat16)`.
@sgugger what did you think on my solution?

sgugger:
Oh, that's better that what I did! Do you want to make a PR with it?

noamwies:
Sound good, I opened the linked PR.

PR #460, opened 2022-06-21
Handle bfloat16 weights in disk offload

sgugger:
This PR handles a bug reported in #454 concerning bfloat16 and disk offload.
NumPy does not support bfloat16, so while storing those tensors on disk, we need to convert them back as float32 (can't convert them in float16 without losing precision). This not super efficient but it's the best we can offer until NumPy offers support for this dtype.

A new test is added to check the offload on disk with bfloat16 works. It doesn't need any specialized hardware since we don't do anything on the bfloat16 tensor.

### DIFF
--- a/tests/test_offload.py
+++ b/tests/test_offload.py
@@ -59,7 +59,7 @@ class OffloadTester(unittest.TestCase):
             dtypes.append(torch.bfloat16)
 
         for dtype in dtypes:
-            weight = torch.tensor([[1, 2, 3], [4, 5, 6]], dtype=dtype)
+            weight = torch.randn(2, 3, dtype=dtype)
             with TemporaryDirectory() as tmp_dir:
                 index = offload_weight(weight, "weight", tmp_dir, {})
                 weight_file = os.path.join(tmp_dir, "weight.dat")


tests/test_offload.py
```diff
@@ -49,6 +53,22 @@ def test_offload_state_dict(self):
                 self.assertTrue(os.path.isfile(weight_file))
                 # TODO: add tests on the fact weights are properly loaded
 
+    def test_offload_weight(self):
+        dtypes = [torch.float16, torch.float32]
+        if is_torch_version(">=", "1.10"):
+            dtypes.append(torch.bfloat16)
+
+        for dtype in dtypes:
+            weight = torch.tensor([[1, 2, 3], [4, 5, 6]], dtype=dtype)
```
muellerzr, line 62: Since we care about the prevision here, would it be a good idea to test these with floats that marginally get close to the float16 reach to ensure they're lossless? 

E.g.:
```python
weight = torch.rand((2,3), dtype=dtype)
```

muellerzr (commented):
One small comment/question there about the tests

tests/test_offload.py
sgugger: Probably yes 😅

PR #461, opened 2022-06-22
Handle bfloat16 weights in disk offload without adding memory overhead

noamwies:


noamwies:
This PR handles a bug reported in https://github.com/huggingface/accelerate/issues/454 concerning bfloat16 and disk offload..

sgugger (approved):
Perfect! This will avoid storage to blow up with disk offload, thanks a lot!

PR #462, opened 2022-06-22
Include bf16 support for TPUs and CPUs, and a better check for if a CUDA device supports BF16

muellerzr:
# Add BF16 support for TPUs and CPUs

## What does this add?

This PR includes better support on TPUs for BF16, introduces BF16 support for CPUs, and a helper to check if your GPU also happens to support bf16 if set. 

## Who is it for?

Users of Accelerate who want to train on BF16

## Why is it needed?

Post fixing the conditional for testing fp16 in the test script, I realized that CPU's can support bfloat16 (though is it advised is debatable), and that to enable TPU bf16 in Accelerate you should set an environment variable beforehand. This PR fixes these two.

### User-facing:

The user can now pass `mixed_precision="bf16"` and train on bf16 in modern CPUs and GPUs. 

### Internal structure:

Adds a `is_bf16_available` function that will check if we're on the CPU and a torch version > 1.10, runs `torch.cuda.is_bf16_available()`  if on the GPU, and will return whether the TPU should use BF16 or not as a param (useful for testing or opting to not run on bf16).

Internally sets the `XLA_USE_BF16` env variable in `AcceleratorState` based on if we're using BF16 or not. 

## Basic Usage Example(s):

```python
# When training on the CPU, TPU, or GPU
accelerator = Accelerate(mixed_precision="bf16")
```

If your GPU is not supported, it will raise an error stating so. 

When wanting to train on bf16 on CPU, GPU, and TPU.

## Does a similar feature exist? If so, why is this better?

For TPU, to use `bf16` you set `XLA_USE_BF16=1` to do so. We do this automatically for you.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -46,6 +46,7 @@ from .utils import (
     extract_model_from_parallel,
     gather,
     get_pretty_name,
+    is_bf16_available,
     is_deepspeed_available,
     is_torch_version,
     is_tpu_available,
@@ -244,18 +245,22 @@ class Accelerator:
         self.native_amp = False
         if self.state.mixed_precision == "fp16":
             self.native_amp = is_torch_version(">=", "1.6")
+            err = "fp16 mixed precision requires {requirement}"
             if not self.native_amp:
-                raise ValueError("fp16 mixed precision requires PyTorch >= 1.6")
-
+                raise ValueError(err.format(err="PyTorch >= 1.6"))
+            if not torch.cuda.is_available():
+                raise ValueError(err.format(err="a GPU"))
             kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
             self.scaler = torch.cuda.amp.GradScaler(**kwargs)
         elif self.state.mixed_precision == "bf16":
-            self.native_amp = is_torch_version(">=", "1.10")
+            self.native_amp = is_bf16_available(True)
             if mixed_precision == "bf16" and not self.native_amp:
-                raise ValueError("bf16 mixed precision requires PyTorch >= 1.10")
+                raise ValueError("bf16 mixed precision requires PyTorch >= 1.10 and a supported device.")
 
-            kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
-            self.scaler = torch.cuda.amp.GradScaler(**kwargs)
+            # Only on the GPU do we care about scaling the gradients
+            if torch.cuda.is_available():
+                kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
+                self.scaler = torch.cuda.amp.GradScaler(**kwargs)
 
         # Internal references to the training objects
         self._optimizers = []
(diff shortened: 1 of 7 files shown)


src/accelerate/accelerator.py
```diff
@@ -1054,8 +1060,10 @@ def autocast(self):
         if self.native_amp:
             if self.mixed_precision == "fp16" and is_torch_version(">=", "1.10"):
                 autocast_context = torch.cuda.amp.autocast(dtype=torch.float16)
-            elif self.mixed_precision == "bf16":
-                autocast_context = torch.cuda.amp.autocast(dtype=torch.bfloat16)
+            elif self.mixed_precision == "bf16" and is_bf16_available():
+                if self.distributed_type in [DistributedType.NO, DistributedType.MULTI_CPU, DistributedType.MULTI_GPU]:
+                    device_type = "cpu" if not torch.cuda.is_available() else "cuda"
+                    autocast_context = torch.autocast(dtype=torch.bfloat16, device_type=device_type)
```
sgugger, line 1066: Need to be extra sure that this always exists for PyTorch version for which `is_bf16_available()`

src/accelerate/launchers.py
```diff
@@ -65,14 +72,14 @@ def notebook_launcher(function, args=(), num_processes=None, use_fp16=False, mix
                 num_processes = 8
 
             launcher = PrepareForLaunch(function, distributed_type="TPU")
-            print(f"Launching a training on {num_processes} TPU cores.")
+            print(f"Launching a training on {num_processes} TPU cores")
```
sgugger, line 75: Not sure why the period is removed here, but I like them at the end of sentences :-p

src/accelerate/state.py
```diff
@@ -92,9 +92,11 @@ def __init__(
                 self.process_index = xm.get_ordinal()
                 self.local_process_index = xm.get_local_ordinal()
                 self.device = xm.xla_device()
-                self.mixed_precision = (
-                    parse_choice_from_env("MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
-                )
+                if mixed_precision == "bf16":
+                    os.environ["XLA_USE_BF16"] = 1
+                else:
+                    os.environ["XLA_USE_BF16"] = 0
```
sgugger, line 98: Do we want to set anything if `mixed_precision != "bf16"`? I'd remove the else statement. But if you really want it, then use one line for such a small test.

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -245,74 +252,77 @@ def training_check():
...
+                loss = torch.nn.functional.mse_loss(output, batch["y"])
+                accelerator.backward(loss)
+                optimizer.step()
+
+        model = accelerator.unwrap_model(model).cpu()
+        assert torch.allclose(old_model.a, model.a), "Did not obtain the same model on CPU or distributed training."
+        assert torch.allclose(old_model.b, model.b), "Did not obtain the same model on CPU or distributed training."
+
+        # TEST that previous fp16 flag still works
+        print("Legacy FP16 training check.")
```
sgugger, line 280: Not sure we need to keep this. We have done a couple of releases since we deprecated it, so it's okay if we stop testing it IMO.

sgugger (approved):
Thanks for all the fixes in this PR, plus the new BF16 support!

src/accelerate/launchers.py
muellerzr: That's from me messing with some things, put it back!

src/accelerate/accelerator.py
muellerzr: Fixed by adding a torch check for >= 1.10

sgugger (approved):

PR #463, opened 2022-06-23
Fix subtle bug in BF16

muellerzr:
This PR fixes two subtle bugs I noticed post my previous PR:
- We currently can no longer get the mixed precision from the environment variable (oops)
- If on a TPU and you select bf16 an error will be raised in the bf16 flag, this has the flag get ignored if a tpu is available in the check as all of the bf16 logic for TPU's lives in `state`


sgugger (approved):
Thanks for the fix!

PR #464, opened 2022-06-23
Link CI slow runners to the commit

muellerzr:
Currently slow runner tests don't show up on the commit to main as they're triggered separately, making it hard to realize when they fail. This PR makes the on-merge ci (aka the slow runner) reusable and has `check_dependencies` trigger it at the very end. This lets the commit SHA/message be linked and it will show up in the merge history with a checkmark or x if there was a failure. 

Some git specifics:

If you combine an `if: always()` with a `needs: {workflow}`, it will wait until `{workflow}` finishes running and then will always run.

### DIFF
--- a/.github/workflows/check_dependencies.yml
+++ b/.github/workflows/check_dependencies.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - actions
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


ydshieh (approved):
Thank you @muellerzr !

sgugger (approved):
Thanks for the fix!

PR #465, opened 2022-06-23
Cleanup CI Warnings

muellerzr:
Cleans most of the warnings in the CI, the two that are not fixed are DeprecationWarnings stemming from cometML and TensorBoard

### DIFF
--- a/examples/by_feature/checkpointing.py
+++ b/examples/by_feature/checkpointing.py
@@ -16,18 +16,13 @@ import argparse
 import os
 
 import torch
+from torch.optim import AdamW
 from torch.utils.data import DataLoader
 
 import evaluate
 from accelerate import Accelerator, DistributedType
 from datasets import load_dataset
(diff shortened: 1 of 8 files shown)


examples/by_feature/checkpointing.py
```diff
@@ -155,7 +155,7 @@ def training_function(config, args):
     model = model.to(accelerator.device)
 
     # Instantiate optimizer
-    optimizer = AdamW(params=model.parameters(), lr=lr, correct_bias=correct_bias)
+    optimizer = AdamW(params=model.parameters(), lr=lr, correct_bias=correct_bias, no_deprecation_warning=True)
```
sgugger, line 158: Switch to the torch `AdamW`? The HF ones as small bugs that can hurt larger trainings.

sgugger (approved):
Thanks for diving into those!

PR #468, opened 2022-06-23
Add docbuilder to setup.py

muellerzr:
Adds docbuilder to the setup.py similar to transformers. 

I could put it also in the `docs` subsection as well, but as we're mimicking it twice I'd rather we wait until doc-specific dependencies are needed, then it may be a good idea to have a `deps_list` util similar to that of `transformers`. For now though I think that's unnecessary.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ from setuptools import setup
 from setuptools import find_packages
 
 extras = {}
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Makes me think I need to migrate the doc styling to use the one in `hf-doc-builder` ;-)

PR #469, opened 2022-06-24
Fix when TPU device check is ran

muellerzr:
This PR fixes an issue where `xm.xla_device()` can't be called outside of `xm.spawn`. As a result the current behavior for `is_tpu_available` breaks the notebook launcher. 

The proposed fix is to check this in state directly outside the if chain so that checking if on a TPU device can be checked properly still.

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -51,7 +51,7 @@ def is_apex_available():
     return importlib.util.find_spec("apex") is not None
 
 
-def is_tpu_available(check_device=False):
+def is_tpu_available(check_device=True):
     "Checks if `torch_xla` is installed and potentially if a TPU is in the environment"
     if _tpu_available and check_device:
         try:
@@ -77,7 +77,7 @@ def is_deepspeed_available():
 
 def is_bf16_available(ignore_tpu=False):
     "Checks if bf16 is supported, optionally ignoring the TPU"
-    if is_tpu_available():
+    if is_tpu_available(check_device=False):
         return not ignore_tpu
     if is_torch_version(">=", "1.10"):
         if torch.cuda.is_available():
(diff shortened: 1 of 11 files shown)


src/accelerate/utils/imports.py
```diff
@@ -38,12 +38,7 @@
 try:
     import torch_xla.core.xla_model as xm  # noqa: F401
 
-    try:
-        # Will raise a RuntimeError if no XLA configuration is found
-        _ = xm.xla_device()
-        _tpu_available = True
-    except RuntimeError:
-        _tpu_available = False
```
sgugger, line 46: Mmm, removing this will break the other places we use `is_tpu_available`

sgugger (changes requested):
We can maybe add an argument to the `is_tpu_available` function to no check for a tpu device sometimes, but we can't remove the device test entirely as we added it for a reason.

src/accelerate/utils/imports.py
muellerzr: The added reason was specifically inside `AcceleratorState`, at the proposed location. But having it as an argument instead works as well. Will refactor

src/accelerate/utils/imports.py
```diff
@@ -56,8 +51,15 @@ def is_apex_available():
     return importlib.util.find_spec("apex") is not None
 
 
-def is_tpu_available():
-    "Checks if `torch_xla` is installed and if a TPU is in the environment"
+def is_tpu_available(check_device=False):
```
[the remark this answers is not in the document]
muellerzr: Or made into a separate function if we don't want `False` behavior. (I know we're not fans of that, but this is one case where it should be `False`)

sgugger (approved):
Thanks for iterating!

PR #470, opened 2022-06-24
Fix documentation not rendering on main

muellerzr:
This PR should solve some of why the docs on main don't render. The reason is we need deepspeed installed in our system to render the docs, otherwise we get a "cannot find ... in accelerate" for deepspeed utils such as this failed run:

https://github.com/huggingface/doc-builder/runs/6988543241?check_suite_focus=true

What's scary is this was a very hidden bug that didn't show on our runs to main. I *think* this solves the problem, but not sure how to 100% test it on deployed docs

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -13,5 +13,6 @@ jobs:
     with:
       commit_sha: ${{ github.sha }}
       package: accelerate
+      additional_args: yes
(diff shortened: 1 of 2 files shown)


muellerzr:
To fix the tests over there, `deepspeed` also needs to be installed. Should `deepspeed` be added to the `docs` extras and used instead @sgugger

PR #475, opened 2022-06-28
fixing fsdp autowrap functionality

pacman100:
### What does this PR do?
1. PyTorch has renamed `default_auto_wrap_policy` to `size_based_auto_wrap_policy`. This PR updates the same.


muellerzr:
@pacman100 I'd recommend perhaps doing a version check to see what version of torch we're on to import one vs the other?

pacman100:
> @pacman100 I'd recommend perhaps doing a version check to see what version of torch we're on to import one vs the other?

Done. As the FSDP support needs torch nightlies, just bumping up the version is viable.

muellerzr (approved):

sgugger (approved):

Issue #452, opened 2022-06-18
TPU training gives different result than CPU training


nalzok:
### System Info

```Shell
- `Accelerate` version: 0.10.0
- Platform: Linux-5.13.0-1023-gcp-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.22.4
- PyTorch version (GPU?): 1.11.0a0+gitbc2c6ed (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: TPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

We first do TPU training

```bash
$ pipenv run accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 3
What is the name of the function in your script that should be launched in all parallel scripts? [main]:
How many TPU cores should be used for distributed training? [1]:1

$ time pipenv run accelerate launch /tmp/accelerate/examples/nlp_example.py
...a lot of irrelevant warnings...
epoch 0: {'accuracy': 0.7720588235294118, 'f1': 0.8558139534883721}
epoch 1: {'accuracy': 0.8235294117647058, 'f1': 0.8615384615384616}
epoch 2: {'accuracy': 0.8676470588235294, 'f1': 0.9059233449477352}

________________________________________________________
Executed in  150.50 secs   fish           external
   usr time  305.93 secs    0.00 micros  305.93 secs
   sys time   36.76 secs  835.00 micros   36.76 secs
```

We then compare the result with CPU training by passing `--cpu`. Note that I am still getting the `OpKernel` warnings, which suggests the TPU is being used (or maybe the library is loaded but the hardware is not used? not sure). Anyway, that's the topic for another issue: https://github.com/huggingface/accelerate/issues/451.

```bash
[...]

Issue #450, opened 2022-06-18
Training with multiple TPU cores is incorrect and slower


nalzok:
### System Info

```Shell
- `Accelerate` version: 0.10.0
- Platform: Linux-5.13.0-1023-gcp-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.22.4
- PyTorch version (GPU?): 1.11.0a0+gitbc2c6ed (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: TPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

First, train with 8 TPU cores (run the script twice and count the second one to exclude the time used to download the datasets)

```python3
$ pipenv run accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
Which type of machine are you using? ([0] No distributed training, [1] multi-CPU, [2] multi-GPU, [3] TPU): 3
What is the name of the function in your script that should be launched in all parallel scripts? [main]:
How many TPU cores should be used for distributed training? [1]:8

$ time pipenv run accelerate launch accelerate/examples/nlp_example.py
...a lot of irrelevant warnings...
epoch 0: {'accuracy': 0.3125, 'f1': 0.0}
epoch 1: {'accuracy': 0.671875, 'f1': 0.8004750593824228}
epoch 2: {'accuracy': 0.6796875, 'f1': 0.8079625292740047}

________________________________________________________
Executed in  157.17 secs   fish           external
   usr time  474.51 secs  592.00 micros  474.51 secs
   sys time   99.76 secs    0.00 micros   99.76 secs
```

Next, train with 1 TPU core

```python
$ pipenv run accelerate config
In which compute environment are you running? ([0] This machine, [1] AWS (Amazon SageMaker)): 0
[...]

PR #479, opened 2022-06-28
Rm gradient accumulation on TPU

muellerzr:
Removes gradient accumulation from all examples if on a TPU as there's a major bug within it apparently on the `xla` side. Will investigate further eventually.

Solves #452  and #450

### DIFF
--- a/examples/by_feature/gradient_accumulation.py
+++ b/examples/by_feature/gradient_accumulation.py
@@ -110,7 +110,10 @@ def training_function(config, args):
     seed = int(config["seed"])
     batch_size = int(config["batch_size"])
     # New Code #
-    gradient_accumulation_steps = int(args.gradient_accumulation_steps)
+    if accelerator.distributed_type != DistributedType.TPU:
+        gradient_accumulation_steps = int(args.gradient_accumulation_steps)
+    else:
+        gradient_accumulation_steps = 1
 
     metric = evaluate.load("glue", "mrpc")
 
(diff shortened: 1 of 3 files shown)


examples/by_feature/gradient_accumulation.py
```diff
@@ -110,7 +110,10 @@ def training_function(config, args):
     seed = int(config["seed"])
     batch_size = int(config["batch_size"])
     # New Code #
-    gradient_accumulation_steps = int(args.gradient_accumulation_steps)
+    if accelerator.distributed_type != DistributedType.TPU:
+        gradient_accumulation_steps = int(args.gradient_accumulation_steps)
+    else:
+        gradient_accumulation_steps = 1
```
sgugger, line 116: No, you should raise an error at the beginning if on TPU with a gradient accumulation steps argument passed > 1.

examples/complete_nlp_example.py
```diff
@@ -196,7 +196,8 @@ def collate_fn(examples):
             if args.with_tracking:
                 total_loss += loss.detach().float()
             accelerator.backward(loss)
-            if step % gradient_accumulation_steps == 0:
+            # We currently do not support nor recommend gradient accumulation on the TPU as bugs were found on the XLA side when running our tests
+            if step % gradient_accumulation_steps == 0 or accelerator.distributed_type == DistributedType.TPU:
```
sgugger, line 200: Don't need to change here since `gradient_accumulation_steps` will always be 1.

examples/nlp_example.py
```diff
@@ -146,7 +146,8 @@ def training_function(config, args):
             loss = outputs.loss
             loss = loss / gradient_accumulation_steps
             accelerator.backward(loss)
-            if step % gradient_accumulation_steps == 0:
+            # We currently do not support nor recommend gradient accumulation on the TPU as bugs were found on the XLA side when running our tests
+            if step % gradient_accumulation_steps == 0 or accelerator.distributed_type == DistributedType.TPU:
```
sgugger, line 150: Same here.

sgugger (approved):
Thanks for addressing this!

PR #484, opened 2022-06-30
Introduce automatic gradient accumulation wrapper + fix a few test issues

muellerzr:
# Introduce Gradient Accumulation wrapper

## What does this add?

This PR adds:
- An `accelerator.accumulate` context manager that will allow for gradient accumulation automatically without the user having to change **any** code. Simply throw your code under it and call everything as normal. Accelerate will handle properly stepping the scheduler, optimizer, and zeroing.
- A new `_set_state` function to `AcceleratorState` to give the state knowledge of if gradients should be utilized and the optimizer/schedulers should be stepped. This is done internally and the user should never ever need to perform this. As a precautionary only state keys that need to be set in *this* PR have been added. A follow up will likely include the mixed precision.
- Lets `backward` handle the gradient accumulation division since it's tied to the `Accelerator` now
- Accelerate's prepared `DataLoaders` now mark in the `AcceleratorStep` if we are one away from the end. We can safely assume this since all dataloaders have a `__len__` or a way for us to check if we're on the second to last iteration.

## Who is it for?

Users of Accelerate that want a simple API for gradient accumulation

## Why is it needed?

Some feedback on Twitter included some notes that it looks very ugly to do so in its current state and a wrapper would be a very good idea, even in this library where we don't want *too* much wrapping

Also noticed that on a single GPU the script tests were having issues, it's because CUDA will yell at us if we use the `debug_launcher` on `n=1`. Instead we should just call `main` like normal. 

### User-facing:

* A new `Accelerator.accumulate` function
* Accelerator takes in a `gradient_accumulation_steps` param

### Internal structure:

* `AcceleratorState` can be set/changed via a private function
* `Accelerator` now keeps track of the step we're on
* `step` and `zero_grad` in both Scheduler and Optimizer are now deterministic if they're called by checking `AcceleratorState.sync_gradients`. By default this is set to `True`.

## Basic Usage Example(s):

```python
for batch in dataloader:
        output = model(**batch)
        loss = output.loss
        accelerator.backward(loss)
        optimizer.step()
        lr_scheduler.step()
        optimizer.zero_grad()
```

Technically since the default is `1` one could just keep this always on and never know the difference.

## Does a similar feature exist? If so, why is this better?

You can use `no_sync`, but the complexity on the user side is why this PR exists.

### DIFF
--- a/docs/source/accelerator.mdx
+++ b/docs/source/accelerator.mdx
@@ -46,13 +46,13 @@ actually be performed, and auto-scale the loss:
 accelerator = Accelerator(gradient_accumulation_steps=2)
 
 for (input, label) in enumerate(training_dataloader):
-  with accelerator.accumulate(model):
-    predictions = model(input)
-    loss = loss_function(predictions, labels)
-    accelerator.backward(loss)
-    optimizer.step()
-    scheduler.step()
-    optimizer.zero_grad()
+    with accelerator.accumulate(model):
+        predictions = model(input)
+        loss = loss_function(predictions, labels)
+        accelerator.backward(loss)
+        optimizer.step()
+        scheduler.step()
+        optimizer.zero_grad()
 ```
 
 **Note**: Using this with `dispatch_batches=True` is currently not supported.
--- a/docs/source/accelerator.mdx
+++ b/docs/source/accelerator.mdx
@@ -38,4 +38,23 @@ should search for and replace by the corresponding methods of your `accelerator`
 - Use [`~Accelerator.clip_grad_norm_`] instead of `torch.nn.utils.clip_grad_norm_` and
   [`~Accelerator.clip_grad_value_`] instead of `torch.nn.utils.clip_grad_value_`.
 
+To perform gradient accumulation use [`~Accelerator.accumulate`] and specify a `gradient_accumulation_steps`. 
+This will also automatically ensure the gradients are synced or unsynced when on multi-node, check if the step should
+actually be performed, and auto-scale the loss:
+
+```python
+accelerator = Accelerator(gradient_accumulation_steps=2)
+
+for (input, label) in enumerate(training_dataloader):
+  with accelerator.accumulate(model):
+    predictions = model(input)
+    loss = loss_function(predictions, labels)
+    accelerator.backward(loss)
+    optimizer.step()
+    scheduler.step()
+    optimizer.zero_grad()
+```
+
+**Note**: Using this with `dispatch_batches=True` is currently not supported.
+
 [[autodoc]] Accelerator
--- a/docs/source/accelerator.mdx
+++ b/docs/source/accelerator.mdx
@@ -39,7 +39,7 @@ should search for and replace by the corresponding methods of your `accelerator`
   [`~Accelerator.clip_grad_value_`] instead of `torch.nn.utils.clip_grad_value_`.
 
 To perform gradient accumulation use [`~Accelerator.accumulate`] and specify a `gradient_accumulation_steps`. 
-This will also automatically ensure the gradients are synced or unsynced when on multi-node, check if the step should
+This will also automatically ensure the gradients are synced or unsynced when on multi-device training, check if the step should
 actually be performed, and auto-scale the loss:
 
 ```python
@@ -55,6 +55,10 @@ for (input, label) in enumerate(training_dataloader):
         optimizer.zero_grad()
 ```
 
-**Note**: Using this with `dispatch_batches=True` is currently not supported.
+<Tip warning={true}>
+
+Using this with `dispatch_batches=True`  (which is the default for iterable datasets) is currently not supported.
+
+</Tip>
 
 [[autodoc]] Accelerator
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -235,12 +235,17 @@ class Accelerator:
             **kwargs,
         )
 
-        if gradient_accumulation_steps > 1 and self.state.distributed_type == DistributedType.TPU:
-            raise NotImplementedError(
-                "Gradient accumulation on TPU is not supported. Pass in `gradient_accumulation_steps=1`"
-            )
-        self.gradient_accumulation_steps = gradient_accumulation_steps
+        if gradient_accumulation_steps > 1:
+            if self.state.distributed_type == DistributedType.TPU:
+                raise NotImplementedError(
+                    "Gradient accumulation on TPU is not supported. Pass in `gradient_accumulation_steps=1`"
+                )
+            if dispatch_batches:
+                raise NotImplementedError(
+                    "Gradient accumulation with dispatched dataloaders is not supported. Pass in `gradient_accumulation_steps=1` or `dispatch_batches=False`"
+                )
 
+        self.gradient_accumulation_steps = gradient_accumulation_steps
         self.device_placement = device_placement
         self.split_batches = split_batches
         self.dispatch_batches = dispatch_batches
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -92,6 +92,9 @@ class Accelerator:
             default to the value in the environment variable `MIXED_PRECISION`, which will use the default value in the
             accelerate config of the current system or the flag passed with the `accelerate.launch` command. 'fp16'
             requires pytorch 1.6 or higher. 'bf16' requires pytorch 1.10 or higher.
+        gradient_accumulation_steps (`int`, *optional*):
+            The number of steps that should pass before gradients are accumulated. Should be combined with
+            `Accelerator.accumulate`
         cpu (`bool`, *optional*):
             Whether or not to force the script to execute on CPU. Will ignore GPU available if set to `True` and force
             the execution on one process only.
@@ -146,6 +149,7 @@ class Accelerator:
         split_batches: bool = False,
         fp16: bool = None,
         mixed_precision: Union[PrecisionType, str] = None,
+        gradient_accumulation_steps: int = 1,
         cpu: bool = False,
         deepspeed_plugin: DeepSpeedPlugin = None,
         fsdp_plugin: FullyShardedDataParallelPlugin = None,
@@ -230,6 +234,12 @@ class Accelerator:
             _from_accelerator=True,
             **kwargs,
         )
+
+        if gradient_accumulation_steps > 1 and self.state.distributed_type == DistributedType.TPU:
+            raise NotImplementedError(
+                "Gradient accumulation on TPU is not supported. Pass in `gradient_accumulation_steps=1`"
+            )
+        self.gradient_accumulation_steps = gradient_accumulation_steps
 
         self.device_placement = device_placement
         self.split_batches = split_batches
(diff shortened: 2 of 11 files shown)


src/accelerate/accelerator.py
```diff
@@ -367,6 +380,39 @@ def no_sync(self, model):
...
             yield
 
+    @property
+    def _do_sync(self) -> bool:
+        "Checks if gradients should be synchronized and the optimizers + schedulers should be stepped"
+        if self.state.end_of_dataloader:
+            self.step = 0
+            return True
+        self.step += 1
+        if (self.gradient_accumulation_steps == 1) or ((self.step % self.gradient_accumulation_steps) == 0):
```
sgugger, line 390: Note that the first part of the test is implied by the second ;-)

src/accelerate/accelerator.py
```diff
@@ -367,6 +380,39 @@ def no_sync(self, model):
         with context():
             yield
 
+    @property
+    def _do_sync(self) -> bool:
+        "Checks if gradients should be synchronized and the optimizers + schedulers should be stepped"
+        if self.state.end_of_dataloader:
```
sgugger, line 386: This shouldn't be in the state, the state only should contain the environment info, not some things that are subject to change. So store it directly in self.

src/accelerate/accelerator.py
```diff
@@ -367,6 +380,39 @@ def no_sync(self, model):
...
+            dataloader (`torch.utils.DataLoader`)
+                PyTorch DataLoader that was prepared with `Accelerator.prepare`
+        """
+
+        if self._do_sync:
+            context = contextlib.nullcontext
+            AcceleratorState._set_state("sync_gradients", True)
+        else:
+            context = self.no_sync
+            AcceleratorState._set_state("sync_gradients", False)
```
sgugger, line 411: Same here, this should be stored somewhere else.

sgugger (commented):
You should create a new object for shared variable state as the `AcceleratorState` is only meant to contain the environment info.

Also I don't see the code that triggers `end_of_dataloader`.

src/accelerate/scheduler.py
```diff
@@ -52,18 +53,18 @@ def step(self, *args, **kwargs):
...
+                # Split batches -> the training dataloader batch size is not changed so one step per training step
+                self.scheduler.step(*args, **kwargs)
+            else:
+                # Otherwise the training dataloader batch size was multiplied by `num_processes`, so we need to do
+                # num_processes steps per training step
+                num_processes = AcceleratorState().num_processes
+                for _ in range(num_processes):
+                    # Special case when using OneCycle and `drop_last` was not used
+                    if getattr(self.scheduler, "total_steps", 0) <= self.scheduler.last_epoch:
+                        self.scheduler.step(*args, **kwargs)
```
sgugger, line 67: For the scheduler, I'm not entirely sure how to proceed: we step for each process to account for Distributed training, not sure if we should step all the time or just when synchronizing gradients (since the length set for the scheduler probably does not account for gradient accumulation steps).

sgugger (approved):
Nice work! Left some comments.

src/accelerate/scheduler.py
muellerzr: Resolved: we update every time.

src/accelerate/data_loader.py
```diff
@@ -294,12 +296,26 @@ def __init__(self, dataset, device=None, rng_types=None, generator=None, **kwarg
...
+                self.gradient_state._set_end_of_dataloader(_current_batch == _dataloader_length)
+                _current_batch += 1
+            else:
+                self.gradient_state._set_end_of_dataloader(False)
             yield batch if self.device is None else send_to_device(batch, self.device)
+        if not self.gradient_state.end_of_dataloader:
+            self.gradient_state._set_end_of_dataloader(True)
+            logger.warn(
+                "Warning! DataLoader had no length and finished iterating. Backwards pass and stepping must be manually performed one last time."
+            )
```
sgugger, line 318: This is unnecessary complex. Why are you not doing `self.gradient_state._set_end_of_dataloader(True)` at the end of `iter`?

src/accelerate/scheduler.py
```diff
@@ -40,10 +47,11 @@ def __init__(self, scheduler, optimizers, step_with_optimizer: bool = True, spli
         self.scheduler = scheduler
         self.optimizers = optimizers if isinstance(optimizers, (list, tuple)) else [optimizers]
         self.split_batches = split_batches
+        self.gradient_state = GradientState()
         self.step_with_optimizer = step_with_optimizer
 
     def step(self, *args, **kwargs):
-        if not self.step_with_optimizer:
+        if not self.step_with_optimizer or not self.gradient_state.sync_gradients:
```
sgugger, line 54: Not sure about this second test. If `step_with_optimizer=True` we then only step for sync gradients, which is not what we want.

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -12,41 +12,62 @@
...
+from torch.optim.lr_scheduler import LambdaLR
 from torch.utils.data import DataLoader
 
 from accelerate import Accelerator
 from accelerate.test_utils import RegressionDataset, RegressionModel
 from accelerate.utils import DistributedType, set_seed
 
 
-def step_model(model, input, target, accelerator):
+warnings.filterwarnings("ignore", category=UserWarning, module="torch.optim.lr_scheduler")
```
sgugger, line 29: The filter should be done in the scheduler, not the test scripts.

src/accelerate/data_loader.py
muellerzr: This check has to happen as the last batch is being pushed out, otherwise it doesn't occur until outside the iteration loop. 

Example in python of what doesn't happen:

```python
a = [0,1,2,3]
class A:
    def __init__(self,a):
        self.a,self.is_end = a,False
    def __iter__(self):
        for item in a: yield item
        self.is_end = True
aa = A(a)
for i,item in enumerate(aa):
    print(f'At end at {i}?: {aa.is_end}')
print(aa.is_end)
```
Result:
```
At end at 0?: False
At end at 1?: False
At end at 2?: False
At end at 3?: False
True
```
What we want to have happen:
```python
class A:
    def __init__(self,a):
        self.a,self.is_end = a,False
    def __iter__(self):
        for i,item in enumerate(a):
            if i == len(a)-1:
                self.is_end = True 
            yield item
        self.is_end = True
```
```
At end at 0?: False
At end at 1?: False
At end at 2?: False
At end at 3?: True
True
```
Since we cannot check the length, this extra step + warning needs to be done.

muellerzr: I think there's a way to do this easier, will toy with it today.

src/accelerate/data_loader.py
```diff
@@ -294,12 +296,27 @@ def __init__(self, dataset, device=None, rng_types=None, generator=None, **kwarg
...
+                # But we still move it to the device so it is done before `StopIteration` is reached
+                if self.device is not None:
+                    current_batch = send_to_device(current_batch, self.device)
+                next_batch = next(dataloader_iter)
+                yield current_batch
+                current_batch = next_batch
+            except StopIteration:
+                self.gradient_state._set_end_of_dataloader(True)
+                yield current_batch
+                break
```
muellerzr, line 319: Much simpler way for us to iterate over the dataloader, and given we only move to device right before we yield there shouldn't be too much trouble with multiple batches in memory at one time. If so let me know.

sgugger: Very clever!

src/accelerate/state.py
```diff
@@ -209,3 +210,34 @@ def _check_initialized(self, mixed_precision=None, cpu=None):
...
+        self.initialized = True
+
+    def __repr__(self):
+        return f"Sync Gradients: {self.sync_gradients}\n" f"At end of current dataloader: {self.end_of_dataloader}\n"
+
+    def _set_sync_gradients(self, sync_gradients):
+        self.sync_gradients = sync_gradients
+
+    def _set_end_of_dataloader(self, end_of_dataloader):
+        self.end_of_dataloader = end_of_dataloader
```
sgugger, line 243: Small docstrings here would be great :-) (with a mention that it's private for a good reason and shouldn't be touched by the user directly!)

sgugger (approved):
Very elegant solution for the dataloaders!

src/accelerate/data_loader.py
```diff
@@ -304,7 +304,10 @@ def __iter__(self):
         self.gradient_state._set_end_of_dataloader(False)
         dataloader_iter = super().__iter__()
         # We iterate one batch ahead to check when we are at the end
-        current_batch = next(dataloader_iter)
+        try:
+            current_batch = next(dataloader_iter)
+        except StopIteration:
+            raise StopIteration("Tried iterating over an empty dataloader")
```
sgugger, line 310: No we shouldn't raise here, just yield nothing. A training loop over an empty dataloader works, just doesn't do anything.
We can put a warning however, just in case the user didn't intend that.

docs/source/accelerator.mdx
```diff
@@ -38,4 +38,23 @@ should search for and replace by the corresponding methods of your `accelerator`
 - Use [`~Accelerator.clip_grad_norm_`] instead of `torch.nn.utils.clip_grad_norm_` and
   [`~Accelerator.clip_grad_value_`] instead of `torch.nn.utils.clip_grad_value_`.
 
+To perform gradient accumulation use [`~Accelerator.accumulate`] and specify a `gradient_accumulation_steps`. 
+This will also automatically ensure the gradients are synced or unsynced when on multi-node, check if the step should
```
sgugger, line 42:
```suggestion
This will also automatically ensure the gradients are synced or unsynced when on multi-device training, check if the step should
```

docs/source/accelerator.mdx
```diff
@@ -38,4 +38,23 @@ should search for and replace by the corresponding methods of your `accelerator`
...
+    with accelerator.accumulate(model):
+        predictions = model(input)
+        loss = loss_function(predictions, labels)
+        accelerator.backward(loss)
+        optimizer.step()
+        scheduler.step()
+        optimizer.zero_grad()
+```
+
+**Note**: Using this with `dispatch_batches=True` is currently not supported.
```
sgugger, line 58: Use the proper syntax for a warning here
```suggestion
<Tip warning={true}>

Using this with `dispatch_batches=True`  (which is the default for iterable datasets) is currently not supported.

</Tip>
```

src/accelerate/accelerator.py
```diff
@@ -92,6 +92,9 @@ class Accelerator:
             default to the value in the environment variable `MIXED_PRECISION`, which will use the default value in the
             accelerate config of the current system or the flag passed with the `accelerate.launch` command. 'fp16'
             requires pytorch 1.6 or higher. 'bf16' requires pytorch 1.10 or higher.
+        gradient_accumulation_steps (`int`, *optional*):
```
sgugger, line 95:
```suggestion
        gradient_accumulation_steps (`int`, *optional*, default to 1):
```

src/accelerate/accelerator.py
```diff
@@ -92,6 +92,9 @@ class Accelerator:
...
+            The number of steps that should pass before gradients are accumulated. Should be combined with
```
sgugger, line 96:
```suggestion
            The number of steps that should pass before gradients are accumulated. A number > 1 should be combined with
```

sgugger (approved):

Issue #366, opened 2022-05-14
some accelerator enhancements


ZhiyuanChen:
Hi,

My [main code base](https://github.com/ZhiyuanChen/DanLing/tree/develop) have been switched to be based on accelerate for a while, it works grate! Thank you guys so much!

There are a few things I think it's a common need.

The first is `.distsibuted` attributes, which is basically `.num_processes > 1`. I mean, it's not a lot of works but .... I'm an extremely lazy person.
```
def distsibuted(self):
    """
    Runs in distributed mode
    """
    return self.num_processes > 1
```

Besides, these two decorators are often used in my code when saving models and downloading datasets, I think it could be useful.

```
def on_main_process(func):
    """
    Run func on main process
    """
    @wraps(func)
    def wrapper(self, *args, **kwargs):
        if self.is_main_process or not self.distributed:
            return func(self, *args, **kwargs)
    return wrapper


def on_local_main_process(func):
    """
    Run func on local main process
    """
    @wraps(func)
    def wrapper(self, *args, **kwargs):
        if self.is_local_main_process or not self.distributed:
            return func(self, *args, **kwargs)
    return wrapper
```

muellerzr:
Nice idea! @sgugger WDYT of instead having this wrap around AcceleratorState directly, rather than having this be attached to Accelerator?

sgugger:
I don't mind having wrappers on the Accelerator for those. As long as the naming is clear, it's not a bad thing to have multiple properties/methods/context managers for different use cases :-)

ZhiyuanChen:
I do think `distsibuted` should be in `AcceleratorState`, but for the decorators, I think it's better to put it in the `Accelerator`

sgugger:
There is no way to set it and make it work (you can't transform the training in distributed mode suddenly), so it should be a property.

ZhiyuanChen:
> There is no way to set it and make it work (you can't transform the training in distributed mode suddenly), so it should be a property.

Ohh, by property I meant the `@property` decorated function, and by variable I meant a value. So I assume variable is better.

Though as PyTorch is supporting elastic training, maybe other values (e.g. num_processes) should be set to a property?

sgugger:
I understood properly, and I still stand by the fact it should be a property, decorated by `@property` and not a variable.

ZhiyuanChen:
Should we makee other variables (e.g. `num_processes`) property too?

muellerzr:
[...]

PR #487, opened 2022-07-05
add use_distributed property

ZhiyuanChen:
https://github.com/huggingface/accelerate/issues/366

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -275,7 +275,7 @@ class Accelerator:
 
     @property
     def use_distributed(self):
-        return self.distributed_type != DistributedType.NO
+        return self.distributed_type != DistributedType.NO and self.num_processes > 1
 
     @property
     def distributed_type(self):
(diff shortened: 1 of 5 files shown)


ZhiyuanChen:
I noticed the rest modification proposed in the original issues have been resolved, so just add this last property

src/accelerate/accelerator.py
```diff
@@ -273,6 +273,10 @@ def __init__(
         if self.rng_types is None:
             self.rng_types = ["torch"] if is_torch_version("<=", "1.5.1") else ["generator"]
 
+    @property
+    def use_distributed(self):
+        return self.distributed_type != DistributedType.NO
```
sgugger, line 278: Maybe also check for the number of processes to be `> 1` here, since you are replacing all those tests?

sgugger (approved):
Thanks for your PR! I left a small comment.

ZhiyuanChen:
> Thanks for your PR! I left a small comment.

Thank you for the comments, I have updated the pr

PR #488, opened 2022-07-05
add on_main_process decorators

ZhiyuanChen:
https://github.com/huggingface/accelerate/issues/366


sgugger (commented):
Thanks for your PR! I think it would be great to have those as methods of the `Accelerator` object (your code refers to a `self`` that does not exist otherwise).

ZhiyuanChen:
Hi @sgugger , sorry for this delay, help this is better now

muellerzr:
@ZhiyuanChen could you run "make style; make quality" to solve the Quality Check issue? 😃

ZhiyuanChen:
> @ZhiyuanChen could you run "make style; make quality" to solve the Quality Check issue? 😃
> 
> Thanks!

Sorry, I thought I have fixed it but somehow didn't pushed...

src/accelerate/accelerator.py
```diff
@@ -356,23 +357,59 @@ def mixed_precision(self):
             mixed_precision = self.state.mixed_precision
         return mixed_precision
 
-    @contextmanager
-    def local_main_process_first(self):
+    def on_main_process(func):
+        """
+        Run func on main process only
```
sgugger, line 362:
```suggestion
        A decorator that will run the decorated function on the main process only.
```

src/accelerate/accelerator.py
```diff
@@ -356,23 +357,59 @@ def mixed_precision(self):
...
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            if self.is_main_process or not self.use_distributed:
+                return func(self, *args, **kwargs)
+
+        return wrapper
+
+    def on_local_main_process(func):
+        """
+        Run func on local main process only
```
sgugger, line 374:
```suggestion
        A decorator that will run the decorated function on the local main process only.
```

src/accelerate/accelerator.py
```diff
@@ -356,23 +357,59 @@ def mixed_precision(self):
...
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            if self.is_local_main_process or not self.use_distributed:
+                return func(self, *args, **kwargs)
+
+        return wrapper
+
+    def on_process(process_idx):
+        """
+        Run func on certain process only
```
sgugger, line 386:
```suggestion
        A decorator that will run the decorated function on a given process index only.
```

src/accelerate/accelerator.py
```diff
@@ -356,23 +357,59 @@ def mixed_precision(self):
...
-    @contextmanager
-    def main_process_first(self):
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            if self.is_local_main_process or not self.use_distributed:
+                return func(self, *args, **kwargs)
+
+        return wrapper
+
+    def on_process(process_idx):
```
sgugger, line 384:
```suggestion
    def on_process(process_idx, local=False):
```
Maybe we could group this one and the text in one decorator since it's one that takes arguments?

sgugger (commented):
Thanks a lot for iterating! Almost ready to merge, I just left some comments on the docstrings and a suggestion to group the decorators `on_process` and `on_local_process` together. Let me know your thoughts!

ZhiyuanChen:
> Thanks a lot for iterating! Almost ready to merge, I just left some comments on the docstrings and a suggestion to group the decorators `on_process` and `on_local_process` together. Let me know your thoughts!

Thank you very much tor your comments and suggestion. 

For the docstrings, i have accepted rll suggestions. 

For the group... I think its rather important to ensure api in a similar organisation, and grouping in this way break the consistency with `is_local_main_process` and `on_local_main_process`. 
Though we could use alias/partial/impl to make the implementation more compact

sgugger:
Let's roll with your choice and we'll see what users think then. We can always add aliases in the future :-)

Thanks again for your contribution!

ZhiyuanChen:
> Let's roll with your choice and we'll see what users think then. We can always add aliases in the future :-)

Haha, sure thing~

> Thanks again for your contribution!

No worries~

Issue #285, opened 2022-03-18
Issues with saving model/optimizer and loading them back


ashutoshsaboo:
Hello @sgugger ,

Came across multiple related issues regarding this - https://github.com/huggingface/accelerate/issues/242, https://github.com/huggingface/accelerate/issues/154 . They were all closed with this PR - https://github.com/huggingface/accelerate/pull/255, but unfortunately the PR doesn't seem to have much documentation. 

I was looking for specifically: saving a model, it's optimizer state, LR scheduler state, it's random seeds/states, epoch/step count, and other related similar states for reproducible training runs and resuming them correctly. 

I know there's this very brief doc here: [here](https://huggingface.co/docs/accelerate/main/en/accelerator#accelerate.Accelerator.save_state) and [here](https://huggingface.co/docs/accelerate/main/en/accelerator#accelerate.Accelerator.load_state) , but it looks like there are still few grey areas not documented currently regarding it's usage. 
a) My question is specifically that, like in the official example here: [link](https://github.com/huggingface/transformers/blob/master/examples/pytorch/text-classification/run_glue_no_trainer.py#L455) that saves using `save_pretrained` only in the main process, should I be using these only in the main process (both save/load) too, and in case of load_state I will have to call prepare() after load_state is done to prepare them for multi-gpu training/inference after that is done (or does load_state do all of that internally itself?)? 
b) Does the save_state method call `save_pretrained` methods for the model internally or do I have to do both? FWIW, I'm using HF's BERT and other pretrained models from the transformers lib, so if there are any other specialized methods specifically for those then please advise on the same. If there's any simple toy example that already uses these new checkpointing methods, and if you can help share that'd be pretty helpful! 

The last release seems to be way back in Sept 2021 - https://github.com/huggingface/accelerate/releases/tag/v0.5.1 - and the PR is just about a month old. Any plans for a soonish version-bump release of accelerate?

Request: If some more detailed examples can be added to the docs that'd be really awesome and help clarify about some of these specifics to users more easily! 

Thanks so much in advance! :)

sgugger:
Hi there, thanks for reaching out! Saving should only be done in the main process (no need to save several times!) but the loading should be done in all processes so each process has the right state. We'll work on adding more documentation on #255 next week (Zach is on vacation right now :-) )

[...]

PR #489, opened 2022-07-06
Don't unwrap in save_state()

cccntu:
I think this is a simple fix for #285 

I haven't test this. Just looking for feedback first.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -977,7 +977,7 @@ class Accelerator:
         output_dir = os.path.expanduser(output_dir)
         os.makedirs(output_dir, exist_ok=True)
         logger.info(f"Saving current state to {output_dir}")
-        weights = [self.get_state_dict(m) for m in self._models]
+        weights = [self.get_state_dict(m, unwrap=False) for m in self._models]
         save_location = save_accelerator_state(
             output_dir, weights, self._optimizers, self._schedulers, self.state.process_index, self.scaler
         )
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks a lot for your PR!

I think this is sensible since the `save_state` definitely happens after `prepare`, as does `load_state` but double-checking with @muellerzr

muellerzr:
Since this went unnoticed, might be good to add in a deepspeed test to the checkpointing test available here: https://github.com/huggingface/accelerate/blob/main/tests/test_state_checkpointing.py

If you think you can manage that feel free to add it to this PR, worst case I can add it in as a follow up

cccntu:
Hi, although this solves the model weights issue, but now it fails when loading optimizer states.

Because deepspeed `optimizer.state_dict()` returns a dictionary.
https://github.com/microsoft/DeepSpeed/blob/9305916d6b858b20fca3ebe03940ceaaa13cdfbd/deepspeed/runtime/zero/stage_1_and_2.py#L1998

But deepspeed `optimizer.load_state_dict()`  expects a list of dictionary.
https://github.com/microsoft/DeepSpeed/blob/9305916d6b858b20fca3ebe03940ceaaa13cdfbd/deepspeed/runtime/zero/stage_1_and_2.py#L2135-L2167
because deepspeed `load_checkpoint` actually gets multiple state_dict (one for each partition or something) https://github.com/microsoft/DeepSpeed/blob/9305916d6b858b20fca3ebe03940ceaaa13cdfbd/deepspeed/runtime/engine.py#L2650

So I think I will just use this
```
if accelerator.distributed_type == DistributedType.DEEPSPEED:
    model.load_checkpoint(path)
    model.save_checkpoint(path)
else:
    accelerator.load_state(path)
    accelerator.save_state(path)
```

cccntu:
Since the test I tried to add fails when loading optimizer, I won't push it here. Please feel free to merge or close this PR.

sgugger:
Let's merge this first fix (thanks a lot!). Cc @pacman100 for the second test.

PR #493, opened 2022-07-07
update dataloader wrappers to have `total_batch_size` attribute

pacman100:
### What does this PR do?
1. A requirement was to get resulting batch size post `accelerator.prepare` call. Adding `total_batch_size` to provide the resulting batch_size across devices.

```python
my_dataloader = accelerator.prepare(my_dataloader)
accelerator.print(f"The resulting batch_size across devices is {my_dataloader.total_batch_size}")
```

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -39,7 +39,6 @@ if is_tpu_available(check_device=False):
     import torch_xla.distributed.parallel_loader as xpl
 
     class MpDeviceLoaderWrapper(xpl.MpDeviceLoader):
-
         @property
         def total_batch_size(self):
             return self._loader.total_batch_size
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -40,10 +40,10 @@ if is_tpu_available(check_device=False):
 
     class MpDeviceLoaderWrapper(xpl.MpDeviceLoader):
         """
-        Wrapper for the xpl.MpDeviceLoader class. This class is used to add `total_batch_size` property to the
-        xpl.MpDeviceLoader class.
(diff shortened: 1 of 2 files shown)


sgugger (commented):
Sounds like a useful addition, but I'd prefer have this as a property and no new arguments in the init that aren't really used to influence the behavior of the class.

src/accelerate/data_loader.py
```diff
@@ -38,6 +38,15 @@
 if is_tpu_available(check_device=False):
     import torch_xla.distributed.parallel_loader as xpl
 
+    class MpDeviceLoaderWrapper(xpl.MpDeviceLoader):
+        def __init__(self, loader, device, **kwargs):
+            super().__init__(loader, device, **kwargs)
```
sgugger, line 43:
```suggestion
```
Can be removed entirely ;-)

src/accelerate/data_loader.py
```diff
@@ -38,6 +38,12 @@
 if is_tpu_available(check_device=False):
     import torch_xla.distributed.parallel_loader as xpl
 
+    class MpDeviceLoaderWrapper(xpl.MpDeviceLoader):
+        @property
+        def total_batch_size(self):
+            return self._loader.total_batch_size
```
muellerzr, line 44: Let's add a docstring to the class here stating briefly what it is and the new property, mimicking how [Accelerator](https://github.com/huggingface/accelerate/blob/docs-revamp/src/accelerate/accelerator.py#L140-L152) has it, since this will be useful for folks and it should show in the documentation as it's different from the base torch dataloader.

Similarly add this property doc to the other two classes as well.

muellerzr (approved):
Great work!

sgugger (approved):

PR #500, opened 2022-07-08
Fix scheduler in gradient accumulation example

muellerzr:
Doesn't divide the number of steps in the scheduler since Accelerate handles it

### DIFF
--- a/src/accelerate/scheduler.py
+++ b/src/accelerate/scheduler.py
@@ -16,7 +16,7 @@
 
 import warnings
 
-from .state import AcceleratorState, GradientState
+from .state import AcceleratorState
 
 
 warnings.filterwarnings("ignore", category=UserWarning, module="torch.optim.lr_scheduler")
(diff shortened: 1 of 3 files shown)


sgugger (approved):
Thanks for fixing!

src/accelerate/scheduler.py
```diff
@@ -25,10 +25,11 @@
...
     """
     A wrapper around a learning rate scheduler that will only step when the optimizer(s) have a training step. Useful
-    to avoid making a scheduler step too fast when:
+    to avoid making a scheduler step too fast when gradients went overflow and there was no training step (in mixed
+    precision training)
 
-    - gradients went overflow and there was no training step (in mixed precision training)
-    - step was skipped because of gradient accumulation
+    When performing gradient accumulation scheduler lengths should not be changed accordingly, accelerate will handle
+    the right times to step.
```
sgugger, line 32:
```suggestion
    When performing gradient accumulation scheduler lengths should not be changed accordingly, accelerate will
    always step the scheduler to account for it.
```

sgugger (approved):

PR #511, opened 2022-07-12
Add gradient accumulation doc

muellerzr:
Adds a doc for gradient accumulation for the next release. It's a carbon copy of the one in #505.

### DIFF
--- a/docs/source/gradient_accumulation.mdx
+++ b/docs/source/gradient_accumulation.mdx
@@ -25,7 +25,7 @@ which can total to adding just one new line of code!
 This example will use a very simplistic PyTorch training loop that performs gradient accumulation every two batches:
 
 ```python
-device = 'cuda'
+device = "cuda"
 model.to(device)
 
 gradient_accumulation_steps = 2
@@ -39,7 +39,7 @@ for index, batch in enumerate(training_dataloader):
     loss = loss_function(outputs, targets)
     loss = loss / gradient_accumulation_steps
     loss.backward()
-    if (index+1) % gradient_accumulation_steps == 0:
+    if (index + 1) % gradient_accumulation_steps == 0:
         optimizer.step()
         scheduler.step()
 ```
--- a/docs/source/gradient_accumulation.mdx
+++ b/docs/source/gradient_accumulation.mdx
@@ -49,25 +49,25 @@ for index, batch in enumerate(training_dataloader):
 First the code shown earlier will be converted to utilize 🤗 Accelerate without the special gradient accumulation helper:
 
 ```diff
-+from accelerate import Accelerator
-+accelerator = Accelerator()
++ from accelerate import Accelerator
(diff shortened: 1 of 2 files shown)


docs/source/gradient_accumulation.mdx
```diff
@@ -0,0 +1,126 @@
...
+    inputs, targets = batch
+-    inputs = inputs.to(device)
+-    targets = targets.to(device)
+    outputs = model(inputs)
+    loss = loss_function(outputs, targets)
+    loss = loss / gradient_accumulation_steps
++    accelerator.backward(loss)
+    if (index+1) % gradient_accumulation_steps == 0:
+        optimizer.step()
+        scheduler.step()
```
sgugger, line 70: Make sure to add a consistent indent level of 2 for the base, so that you can have `+ ` or `- ` without changing indentation on any line.

docs/source/gradient_accumulation.mdx
```diff
@@ -0,0 +1,126 @@
...
+You just wrap it around the entire training part of our code: 
+
+```diff
+-for index, batch in enumerate(training_dataloader):
++for batch in training_dataloader:
++   with accelerator.accumulate(model):
+        optimizer.zero_grad()
+        inputs, targets = batch
+        outputs = model(inputs)
+```
```
sgugger, line 98: Same here.

docs/source/gradient_accumulation.mdx
```diff
@@ -0,0 +1,126 @@
...
+        scheduler.step()
+```
+
+<Tip warning={true}>
+In its current state, this code is not going to perform gradient accumulation efficiently due to a process called gradient synchronization. Read more about that in the Concepts tutorial!
+</Tip>
+
+## Letting 🤗 Accelerate handle gradient accumulation
+
+All that is left now is to let 🤗 Accelerate handle the gradient accumulation for us. To do so you should pass in a `gradient_accumulation_steps` parameter to `Accelerator`, dictating the number
```
sgugger, line 79:
```suggestion
All that is left now is to let 🤗 Accelerate handle the gradient accumulation for us. To do so you should pass in a `gradient_accumulation_steps` parameter to [`Accelerator`], dictating the number 
```

docs/source/gradient_accumulation.mdx
```diff
@@ -0,0 +1,126 @@
...
+
+and you can remove all the special checks for the step number and the loss adjustment:
+
+```diff
+-loss = loss / gradient_accumulation_steps
+accelerator.backward(loss)
+-if (index+1) % gradient_accumulation_steps == 0:
+optimizer.step()
+scheduler.step()
+```
```
sgugger, line 108: Same there + the step lines change since they lose and indent level.

sgugger (approved):
Thanks for adding this!

PR #513, opened 2022-07-13
Fix a bug when reduce a tensor.

wwhio:
return reduced result of a tensor

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -860,8 +860,7 @@ class Accelerator:
         Reduce the values in *tensor* across all processes based on *reduction*.
 
         Note:
-            The reduced value is returned only in the main process. In other processes, the returned value is the
-            same as input.
+            All processes get the reduced value.
 
         Args:
             tensor (`torch.Tensor`, or a nested tuple/list/dictionary of `torch.Tensor`):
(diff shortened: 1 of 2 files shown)


wwhio:
Fix a bug when using reduce at a tensor.

Before:
```
a = torch.tensor(accelerator.process_index, device=accelerator.device)

accelerator.reduce(a, reduction='sum') # don't change a
# or
a = accelerator.reduce(a, reduction='sum') # a is None
```

After:
```
a = torch.tensor(accelerator.process_index, device=accelerator.device)
a = accelerator.reduce(a, reduction='sum') # a is correctly reduced
```

pacman100:
Thanks for the fix! 😄

wwhio:
When I update the doc for this pr, I find that the `accelerate.utils.reduce` seems have inconsistent behavior on different type of devices.

In detail, it will reduce value on all processes, i.e. `all_reduce` when device is TPU. But only reduce to the main process, i.e. `reduce`, when device is GPU.

Could I change the behavior to `all_reduce` when device is GPU ?


For more infomation, see the doc stuff [here](https://github.com/wwhio/accelerate/blob/96a17aae1164da5511a50d7138bd6f8ae5303267/src/accelerate/accelerator.py#L862-L864) and the `accelerate.utils.reduce` code [here](https://github.com/wwhio/accelerate/blob/96a17aae1164da5511a50d7138bd6f8ae5303267/src/accelerate/utils/operations.py#L411-L441).

sgugger:
Yes, I think it should also be `all_reduce` on GPU to avoid bugs when not all processes have the same tensor later in the code.

PR #517, opened 2022-07-13
Fixup all example CI tests and properly fail

muellerzr:
# Fix all example CI

## What does this add?

This pr makes a variety of fixes to the example tests to solve all failures and properly ensure that failures occur when they need to

## Why is it needed?

The entire list of issues I noticed are below:

- We cannot mock `CometMLTracker` to run offline tests with the tracking example script, so the CI now explicitly uninstalls them
- Adds a new `requires_tracking` decorator for the example tests to make sure at least one tracking API is installed and `comet_ml` is not installed
- Creates a new exception class and a new `run_command` function which will properly call a subprocess test function and check its outputs properly as well as fully fail if the call did not work.
- Fixes two bugs in the cross validation example in regards to noting the final prediction score.

### DIFF
--- a/examples/by_feature/cross_validation.py
+++ b/examples/by_feature/cross_validation.py
@@ -229,20 +229,14 @@ def training_function(config, args):
             with torch.no_grad():
                 outputs = model(**batch)
             predictions = outputs.logits
-            predictions, references = accelerator.gather((predictions, batch["labels"]))
             fold_predictions.append(predictions.cpu())
-            metric.add_batch(
-                predictions=predictions.argmax(dim=-1),
-                references=references,
-            )
-        test_metric = metric.compute()
         # Use accelerator.print to print only on the main process.
         test_predictions.append(torch.cat(fold_predictions, dim=0))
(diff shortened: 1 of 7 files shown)


examples/by_feature/cross_validation.py
```diff
@@ -229,20 +229,14 @@ def training_function(config, args):
             with torch.no_grad():
                 outputs = model(**batch)
             predictions = outputs.logits
-            predictions, references = accelerator.gather((predictions, batch["labels"]))
```
sgugger, line 232: This will stop working on distributed setup. The predictions will need to be gathered (maybe not the labels).

sgugger (approved):
Thanks for all the fixes!

examples/by_feature/cross_validation.py
muellerzr: Should be fixed now

muellerzr:
All slow ci is now passing here: https://github.com/huggingface/accelerate/runs/7358153325

sgugger (approved):

PR #519, opened 2022-07-13
Add special Parameters modules support

younesbelkada:
This PR supports adding special `Parameter` modules when calling the context manager `init_empty_weights` that force assign the weights of each module into `nn.Parameter` module. For some specific usecases we might want to replce it with other modules such as `bnb.Int8Params` from `bitsandbytes`

Related PR: https://github.com/huggingface/transformers/pull/17901
Related discussion:  https://github.com/huggingface/transformers/pull/17901#discussion_r920230935

cc @sgugger @TimDettmers

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -35,27 +35,20 @@ def init_empty_weights(include_buffers: bool = False):
     """
     A context manager under which models are initialized with all parameters on the meta device, therefore creating an
     empty model. Useful when just initializing the model would blow the available RAM.
-
     Args:
         include_buffers (`bool`, *optional*, defaults to `False`):
             Whether or not to also put all buffers on the meta device while initializing.
-
     Example:
-
     ```pyton
     import torch.nn as nn
     from accelerate import init_empty_weights
-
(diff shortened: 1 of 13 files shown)


src/accelerate/big_modeling.py
```diff
@@ -39,9 +39,7 @@ def init_empty_weights(include_buffers: bool = False):
     Args:
         include_buffers (`bool`, *optional*, defaults to `False`):
             Whether or not to also put all buffers on the meta device while initializing.
-
     Example:
-
```
sgugger, line 44: Leaves those poor lines alone :-)

sgugger (approved):
Good for me as it doesn't change anything for regular parameters! Thanks!

src/accelerate/big_modeling.py
younesbelkada: 🤣 yeah don't know what happened there

younesbelkada: Fixed in aafea4a1e18ac6562bc678f3e607e0d2c5f89b38 ;)

PR #523, opened 2022-07-15
Add support for downcasting bf16 on TPUs

muellerzr:
This PR adds downcasting support for bf16 on TPUs. Per the XLA docs:

* By default both torch.float and torch.double are torch.float on TPUs.
* If `XLA_USE_BF16` is set, then `torch.float` and `torch.double` are both `bfloat16` on TPUs.
* If `XLA_DOWNCAST_BF16` is set, then `torch.float` is `bfloat16` on TPUs and `torch.double` is `float32` on TPUs.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -126,9 +126,6 @@ class Accelerator:
             accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
         logging_dir (`str`, `os.PathLike`, *optional*):
             A path to a directory for storing logs of locally-compatible loggers.
-        downcast_bf16 (`bool`, *optional*):
-            If set to `True` and using `mixed_precision="bf16"` then `torch.float` will become `bfloat16` and
-            `torch.double` will remain `float32` on TPUs.
         dispatch_batches (`bool`, *optional*):
             If set to `True`, the dataloader prepared by the Accelerator is only iterated through on the main process
             and then the batches are split and broadcast to each process. Will default to `True` for `DataLoader` whose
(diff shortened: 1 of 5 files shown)


src/accelerate/accelerator.py
```diff
@@ -126,6 +126,9 @@ class Accelerator:
             accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
         logging_dir (`str`, `os.PathLike`, *optional*):
             A path to a directory for storing logs of locally-compatible loggers.
+        downcast_bf16 (`bool`, *optional*):
+            If set to `True` and using `mixed_precision="bf16"` then `torch.float` will become `bfloat16` and
+            `torch.double` will remain `float32` on TPUs.
```
sgugger, line 131: Very much against adding a new Accelerator parameter for such a niche feature. It should be in the config only.

src/accelerate/accelerator.py
```diff
@@ -156,8 +159,9 @@ def __init__(
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         log_with: Optional[List[Union[str, LoggerType, GeneralTracker]]] = None,
         logging_dir: Optional[Union[str, os.PathLike]] = None,
+        downcast_bf16: Optional[bool] = False,
         dispatch_batches: Optional[bool] = None,
-        step_scheduler_with_optimizer: bool = True,
+        step_scheduler_with_optimizer: Optional[bool] = True,
```
sgugger, line 164: Doesn't seem to be able to be None, so I'm confused about this change in typing.

src/accelerate/accelerator.py
muellerzr: Okay, will adjust on Monday!

(This should be in the release after this next one anyways)

sgugger (approved):
Much cleaner, thanks!

Issue #520, opened 2022-07-13
Warn user when attempting to track when logger(s) are not installed


nateraw:
When you run `Accelerator(..., log_with="all", ...)` I think it should warn you if you don't have any available trackers installed. Right now, it just won't track/log, which is a bit confusing.

PR #524, opened 2022-07-15
Warn user if no trackers are installed

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/520


sgugger (approved):
Thanks for fixing!

PR #527, opened 2022-07-18
fix some parameter setting does not work for CPU DDP and bf16 fail in…

sywangyi:
… DDP path

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -257,6 +257,8 @@ def simple_launcher(args):
 
     current_env = os.environ.copy()
     current_env["USE_CPU"] = str(args.cpu)
+    current_env["MASTER_ADDR"] = args.main_process_ip
+    current_env["MASTER_PORT"] = str(args.main_process_port)
     try:
         mixed_precision = PrecisionType(args.mixed_precision.lower())
     except ValueError:
@@ -269,6 +271,7 @@ def simple_launcher(args):
         mixed_precision = "fp16"
 
     current_env["MIXED_PRECISION"] = str(mixed_precision)
+    current_env["OMP_NUM_THREADS"] = str(args.num_cpu_threads_per_process)
 
     process = subprocess.Popen(cmd, env=current_env)
     process.wait()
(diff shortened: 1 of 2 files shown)


sywangyi:
@sgugger @yao-matrix please help review

src/accelerate/commands/launch.py
```diff
@@ -257,6 +257,8 @@ def simple_launcher(args):
 
     current_env = os.environ.copy()
     current_env["USE_CPU"] = str(args.cpu)
+    current_env["MASTER_ADDR"] = args.main_process_ip
+    current_env["MASTER_PORT"] = str(args.main_process_port)
```
sgugger, line 261: Should this be under and `if args.num_machines > 1`? It seems related to multinode training.

sgugger (commented):
Thanks for the fixes!

src/accelerate/commands/launch.py
sywangyi: thanks for the suggestion, a patch is uploaded to fix it

src/accelerate/commands/launch.py
```diff
@@ -257,6 +257,12 @@ def simple_launcher(args):
 
     current_env = os.environ.copy()
     current_env["USE_CPU"] = str(args.cpu)
+    if args.num_machines > 1:
+        current_env["MASTER_ADDR"] = args.main_process_ip
+        current_env["MASTER_PORT"] = str(args.main_process_port)
+    elif args.num_processes > 1:
+        current_env["MASTER_ADDR"] = "127.0.0.1"
+        current_env["MASTER_PORT"] = "29500"
```
sgugger, line 265: If they are needed, I would rather use the ones set par the user than hardcoding a default, which may create hard-to-debug issues for users.

sywangyi: hi, sgugger, now in the accelerate config, if num_machines == 1, the master ip and port is not set. and if num_process > 1 and DDP is used in one machine, we still need the master ip and port

sywangyi: done, thanks for the suggestion

sgugger:
Thansk again for your contribution!

PR #529, opened 2022-07-18
Pick a better default for offload_state_dict

sgugger:
After seeing many issues of users struggling with the `offload_state_dict` argument, I've decided to let it default to `None` and pick a default depending on the computed device map (`True` if there are weights offloaded to disk, `False` otherwise).

The same is being implemented in Transformers [here](https://github.com/huggingface/transformers/pull/18183).


muellerzr (approved):
Thanks for this!

PR #539, opened 2022-07-20
Add 8 bit support - chapter II

younesbelkada:
Hi !

Here is a (dirty) attempt of enabling other modules support (other than `nn.Parameter`) on accelerate
This time I had to slightly modify the script `set_tensor_to_device` as following:

- If `is_buffer` : assigns the buffer to `new_value` as usual
- assign `module._parameters[tensor_name]` to the correct `new_value` ONLY when the device is different to a GPU. The latest case means that the new_value has been already set to the correct device so there is probably no need to put it again on the same device. 

I can't see any breaking changes but I highly suspect this statement to be wrong. Would love to hear your thoughts @sgugger @muellerzr 

Related to the PR: https://github.com/huggingface/transformers/pull/17901

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -126,8 +126,10 @@ def set_module_tensor_to_device(
             new_value = torch.tensor(value, device=device)
     if is_buffer:
         module._buffers[tensor_name] = new_value
-    else:
-        new_value = nn.Parameter(new_value, requires_grad=old_value.requires_grad)
+    else:  
+        param_cls = type(module._parameters[tensor_name])
+        kwargs = module._parameters[tensor_name].__dict__
+        new_value = param_cls(data=new_value, requires_grad=old_value.requires_grad, **kwargs)
         module._parameters[tensor_name] = new_value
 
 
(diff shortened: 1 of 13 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -124,11 +124,14 @@ def set_module_tensor_to_device(
...
-        new_value = nn.Parameter(new_value, requires_grad=old_value.requires_grad)
-        module._parameters[tensor_name] = new_value
+
+        if is_buffer:
+            module._buffers[tensor_name] = new_value
+        elif module._parameters[tensor_name].device.type != "cuda":
+            param_cls = type(module._parameters[tensor_name])
+            kwargs = module._parameters[tensor_name].__dict__
+            new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
+            module._parameters[tensor_name] = new_value
```
sgugger, line 134: This should stay at the previous level of indentation. To make sure there is absolutely zero breaking change, how about doing an `else`  with just `module._parameters[tensor_name].to(device)` ?

sgugger (commented):
Thanks for your PR! Left some comment to avoid breaking anything.

src/accelerate/utils/modeling.py
younesbelkada: Just added it and seems to work fine! 
Can we wait a bit before merging the PR?

younesbelkada: Great thanks a lot!!

PR #540, opened 2022-07-20
Add a gather_for_metrics capability

muellerzr:
# Introduce a `gather_for_metrics` function

## What does this add?

This PR adds a new function to `Accelerator` called `gather_for_metrics`, which assists with calculating the right metric in distributed setups

## Who is it for?

Users of accelerate that want to ensure that their reported metrics are fully accurate

## Why is it needed?

To assist with making sure all the batches have the right batch size, Accelerate will pad the length on the last batch to be duplicates of the last sample. These need to be dropped when calculating the final metrics on the last batch, and currently it looks something like:
```python
if accelerator.use_distributed:
    # Then see if we're on the last batch of our eval dataloader
    if step == len(eval_dataloader) - 1:
        # Last batch needs to be truncated on distributed systems as it contains additional samples
        predictions = predictions[: len(eval_dataloader.dataset) - samples_seen]
        references = references[: len(eval_dataloader.dataset) - samples_seen]
    else:
        # Otherwise we add the number of samples seen
        samples_seen += references.shape[0]
```
This PR adds a new utility called `accelerate.gather_for_metrics` which will handle this check for us, entirely thanks to the `GradientState` capability. 

> Note: this method currently doesn't work for TPU's, as it needs to be `eval_dataloader._loader.dataset`. This PR fixes this as well

### User-facing:

- A new `Accelerator.gather_for_metrics` function was added 

### Internal structure:

- Preprocessed dataloaders now have a new `total_dataset_length` attribute
- `GradientState` now keeps track of the number of samples seen

## Basic Usage Example(s):

When calculating metrics, users can now do the following to properly calculate their metrics:

```python
input, target = next(iter(dataloader))
with torch.no_grad():
    logits = ddp_model(ddp_input)
    logits, target = accelerator.gather_for_metrics((logits, ddp_target), dataloader)
    accuracy_multi = accuracy(logits.argmax(dim=-1), target)
```

Since this works on distributed and non-distributed systems, always if the evaluation dataset has been prepared by Accelerator. Users should just add this to any script that calculates metrics.

TODO:
- [x] Update the other examples to use this new API. multiprocess_metrics will serve as a lower-level example

### DIFF
--- a/docs/source/quicktour.mdx
+++ b/docs/source/quicktour.mdx
@@ -120,13 +120,13 @@ validation_dataloader = accelerator.prepare(validation_dataloader)
 
 As for your training dataloader, it will mean that (should you run your script on multiple devices) each device will
 only see part of the evaluation data. This means you will need to group your predictions together. This is very easy to
-do with the [`~Accelerator.gather_metrics`] method.
+do with the [`~Accelerator.gather_for_metrics`] method.
 
 ```python
 for inputs, targets in validation_dataloader:
     predictions = model(inputs)
     # Gather all predictions and targets
-    all_predictions, all_targets = accelerator.gather_metrics((predictions, targets))
+    all_predictions, all_targets = accelerator.gather_for_metrics((predictions, targets))
     # Example of use with a *Datasets.Metric*
     metric.add_batch(all_predictions, all_targets)
 ```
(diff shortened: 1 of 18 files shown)


sgugger (approved):
Very clever! It's really great, just make sure to update the [docs](https://huggingface.co/docs/accelerate/quicktour#distributed-evaluation) and the examples :-)

docs/source/quicktour.mdx
```diff
@@ -141,11 +140,14 @@ As for the training dataloader, passing your validation dataloader through
 Any instruction using your training dataloader length (for instance if you need the number of total training steps
 to create a learning rate scheduler) should go after the call to [`~Accelerator.prepare`].
 
+As some data at the end of the dataset may be duplicated so the batch can divide equally to all workers, metrics should be 
+calculated through the [`~Accelerator.gather_for_metrics`] method to automatically remove the duplicated data.
```
sgugger, line 144: Maybe mention it that if the user doesn't want that behavior for some reason, they can just use `gather`?

sgugger (approved):
Nice doc!

plamb-viso:
is this functionality in 0.11.0?

muellerzr:
It is not, we suggest not using it for now and doing the check manually as shown in the metric example script as some bugs were discovered: https://github.com/huggingface/accelerate/issues/575

plamb-viso:
Cool, thank you, excited for this change when it happens

PR #544, opened 2022-07-21
deepspeed version hotfix

pacman100:
### What does this PR do?

1. DeepSpeed integration is broken when using the latest deepspeed 0.6.7 version. Using version 0.6.5 for time being till there is resolution for https://github.com/microsoft/DeepSpeed/issues/2117 in order to not have failing test

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ extras["test"] = [
     "scipy",
     "sklearn",
     "parameterized",
-    "deepspeed",
+    "deepspeed==0.6.5",
 ]
 
 extras["test_trackers"] = ["wandb", "comet-ml", "tensorboard"]


muellerzr (approved):
Thanks! Might be worth keeping an issue open about this until its resolved to send people to as they get issues/they can find it easily

setup.py
```diff
@@ -28,7 +28,7 @@
     "scipy",
     "sklearn",
     "parameterized",
-    "deepspeed",
+    "deepspeed==0.6.5",
```
sgugger, line 31:
```suggestion
    "deepspeed<=0.6.5",
```
I don't think we required 0.6.5 as a min version before, so let's not pin exactly.

sgugger (approved):
Thanks!

src/accelerate/state.py
```diff
@@ -130,8 +130,16 @@ def __init__(
                 ), "DeepSpeed is not available => install it using `pip3 install deepspeed` or build it from source"
                 self.distributed_type = DistributedType.DEEPSPEED
                 if not torch.distributed.is_initialized():
-                    torch.distributed.init_process_group(backend="nccl", **kwargs)
+                    from .utils import compare_versions
```
sgugger, line 133: This can be imported at the top.

sgugger (approved):
Nice to actually fix the root of the issue!

muellerzr (approved):
Nice fix!

src/accelerate/state.py
pacman100: that was leading to circular depency error

muellerzr: I think it becomes a circular dep import which is why it's down there

sgugger: Ah yes, I can see that! Good for me then!

PR #545, opened 2022-07-21
Rename test extras to testing

muellerzr:
Super tiny nit, but keeps it inline with how transformers has their extras setup for tests. Mostly so I don't have to keep digging to see if I need to do `pip install -e .[tests]` `pip install -e .[testing]` `pip install -e .[test]` when bouncing back and forth between testing transformers and accelerate


sgugger (approved):
LGTM!

PR #547, opened 2022-07-21
Add production testing + fix failing CI

muellerzr:
Adds a production stage to the CI so that barebones testing without any added dependencies can be checked on the git ci first to ensure that trickle-down breakages cannot occur. 

Specifically breaks down the `big_modeling` into its own ci matrix and adds a `core` to the matrix

Also removes unused evaluate test since it adds a dep to the main library. Not wondering if it'd be good for us soon to leave `test_script` in the main lib but migrate out all the other test scripts that might want or need external deps to a seperate folder at some point

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         test-kind: [
           test_prod,
-          test,
+          test_core,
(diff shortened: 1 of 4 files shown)


sgugger (approved):
LGTM, thanks for adding those tests!

PR #548, opened 2022-07-21
Speed up CI more

muellerzr:
Noticed the CI was taking 6.5 minutes or so even with the cached process. This reduces the time it takes for tests to pass by limiting the number of epochs the example scripts run for to two epochs, following a similar setup to how we check the mocked dataloaders

Old time: 6.5min
New time: 4.5min


Makefile
```diff
@@ -25,33 +25,33 @@ style:
 	
 # Run tests for the library
 test:
-	python -m pytest -s -v ./tests/ --ignore=./tests/test_examples.py
+	python -m pytest -s -v ./tests/ --ignore=./tests/test_examples.py --durations=0
```
sgugger, line 28: I wouldn't put the `--durations=0` in the MakeFile as it's a bit spammy if a user executes the tests locally. I'd just add it to the command running the tests (which would be different from the make command).

sgugger (approved):
Thanks for diving into this. The durations should be in the github actions only. And further work could save them as artifacts (like in Transformers) instead of just displaying them.

Makefile
muellerzr: Will do so in a follow-up PR for the CI

PR #551, opened 2022-07-22
Fix a few minor issues with example code in docs

BenjaminBossan:
- `enumerate` is not actually used
- variable name "labels" does not match
- `prepare` method should be called


sgugger (approved):
Thanks for fixing!

BenjaminBossan:
@sgugger Thanks, I applied the change.

sgugger:
Thanks a lot!

PR #553, opened 2022-07-22
Create good defaults in `accelerate launch`

muellerzr:
Creates good defaults for unconfigured `accelerate launch` usage so that the user doesn't have to think much on it. (and actually support not needing to do `accelerate config` again)

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -686,7 +686,6 @@ def launch_command(args):
                     and getattr(args, name, None) is None
                 ):
                     setattr(args, name, attr)
-
         if not args.mixed_precision:
             if args.fp16:
                 args.mixed_precision = "fp16"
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks! Can you also log warnings here, to make sure the user is not surprised by the magical defaults?

src/accelerate/commands/launch.py
```diff
@@ -692,16 +696,26 @@ def launch_command(args):
...
             args.num_machines = 1
         if args.mixed_precision is None:
+            warned = True
+            logger.warn("`--mixed_precision` was not set, using a value of `'no'`.")
             args.mixed_precision = "no"
         if not hasattr(args, "use_cpu"):
-            args.use_cpu = torch.cuda.is_available()
+            args.use_cpu = args.cpu
+        if warned:
+            logger.warn("To avoid these warnings pass in values for each of the problematic parameters")
```
sgugger, line 715:
```suggestion
            logger.warn("To avoid these warnings pass in values for each of the problematic parameters or run `accelerate config`.")
```

src/accelerate/commands/launch.py
```diff
@@ -692,16 +696,26 @@ def launch_command(args):
...
 
     if args.multi_gpu and args.num_processes == 1:
+        logger.warn("`--multi_gpu` was passed but `num_processes` was not set. Automatically using all available GPUs")
```
sgugger, line 718: This one should have the text above to avoid warning (which you can put in a constant).

sgugger (approved):

Issue #555, opened 2022-07-22
checkpointing not saving model weights if calling `accelerator.prepare_model` instead of `accelerator.prepare`


csarron:
### System Info

```Shell
- `Accelerate` version: 0.11.0
- Platform: Linux-4.15.0-161-generic-x86_64-with-glibc2.10
- Python version: 3.8.12
- Numpy version: 1.23.1
- PyTorch version (GPU?): 1.12.0+cu113 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

  if we separately prepare the model (i.e. `model = accelerator.prepare_model(model)`) instead of preparing all at once (`model, optimizer, train_dataloader, eval_dataloader, lr_scheduler = accelerator.prepare(model, optimizer, train_dataloader, eval_dataloader, lr_scheduler)`), accelerator won't save model weights using the `accelerator.save_states`

possible problem: accelerator internally use [`_prepare_one`](https://github.com/huggingface/accelerate/blob/6c4edc362f85e79fac143c41a189bc90cb42e29e/src/accelerate/accelerator.py#L449) in the `prepare(self, *args)` method to append the model to `self._models`, but the `prepare_model(self, model)` does not append the model, see [here](https://github.com/huggingface/accelerate/blob/6c4edc362f85e79fac143c41a189bc90cb42e29e/src/accelerate/accelerator.py#L572). However, `accelerator.save_states` [depends on ](https://github.com/huggingface/accelerate/blob/6c4edc362f85e79fac143c41a189bc90cb42e29e/src/accelerate/accelerator.py#L1045) `self._models` to save model weights

### Expected behavior

```Shell
accelerator.save_states should also save model weights even if one separately calls `prepare_model` method, this is possibly a bug in `accelerator.prepare_model(model)`
```

PR #556, opened 2022-07-22
fix: saving model weights

csarron:
checkpointing not saving model weights if calling `accelerator.prepare_model` instead of `accelerator.prepare`
resolves issue: https://github.com/huggingface/accelerate/issues/555


muellerzr:
If we want to go this route, it would be good to update the rest of them as well. I.e. optimizer and scheduler

csarron:
@muellerzr thanks for the comment, I updated the optimizer and scheduler as well. If there are other possible changes, please let me know.

sgugger (approved):
This makes the whole API less surprising to the user, thanks a lot for fixing all of those!

PR #564, opened 2022-07-25
Add a tqdm helper

muellerzr:
Creates a `tqdm` helper that will automatically only show on the main process, avoiding the need to use a `if accelerator.is_local_main_process` each time


setup.py
```diff
@@ -50,7 +50,7 @@
         ]
     },
     python_requires=">=3.7.0",
-    install_requires=["numpy>=1.17", "packaging>=20.0", "psutil", "pyyaml", "torch>=1.4.0"],
+    install_requires=["numpy>=1.17", "packaging>=20.0", "psutil", "pyyaml", "torch>=1.4.0", "tqdm"],
```
sgugger, line 53: Let's make tqdm a soft dependency.

sgugger (approved):
Good for me, let's just wrap the dep to not be a core dep, just a soft dep (and the utils function `tqdm` then raises an `ImportError` if tqdm is not installed). Thanks!

PR #565, opened 2022-07-25
Update examples with grad accum

muellerzr:
Updates the NLP examples with gradient accumulation. (CV didn't get updated since it didn't have it to begin with)


sgugger (approved):
Thanks for updating those!

PR #567, opened 2022-07-26
ccl version check and import different module according to version

sywangyi:

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -17,7 +17,7 @@ from distutils.util import strtobool
 
 import torch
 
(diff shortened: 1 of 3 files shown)


sywangyi:
@liangan1 @yao-matrix @sgugger please have a review

muellerzr:
Thanks for this! Looks great to me 😄 could you run `make style; make quality` to get past that failure please? And then we're good!

sywangyi:
@muellerzr thanks for the suggestion, fix the "make quality" issue.

sgugger (approved):
Thanks for your PR!
Looks like this conflicts with an earlier PR #554 but yours is more complete. Could you just solve the merge conflict so we can merge?

PR #568, opened 2022-07-26
Rename actions to be a bit more accurate

muellerzr:
Renames the stages in the github actions to be a bit more accurate on what is being ran


sgugger (approved):
LGTM!

PR #573, opened 2022-07-26
Fix tests, add wandb to gitignore

muellerzr:
Actually fixes the multi-gpu examples, also added the `wandb` folder to the gitignore since it may be run when testing the tracking and create a `wandb` folder in the repo.


sgugger (approved):
I'll trust you on this, thanks for fixing!

PR #576, opened 2022-07-27
Fix `set_module_tensor_to_device`

sgugger:
It turns out that `to` converts a parameter to a tensor, so the else branch that was added in #539 breaks a lot of things 😬 
This PR fixes that.

cc @younesbelkada

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -127,7 +127,7 @@ def set_module_tensor_to_device(
 
         if is_buffer:
             module._buffers[tensor_name] = new_value
-        elif device != module._parameters[tensor_name].device:
+        elif torch.device(device) != module._parameters[tensor_name].device:
             param_cls = type(module._parameters[tensor_name])
             kwargs = module._parameters[tensor_name].__dict__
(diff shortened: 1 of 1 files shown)


muellerzr (approved):

src/accelerate/utils/modeling.py
```diff
@@ -127,13 +127,11 @@ def set_module_tensor_to_device(
 
         if is_buffer:
             module._buffers[tensor_name] = new_value
-        elif module._parameters[tensor_name].device.type != "cuda":
+        elif device != module._parameters[tensor_name].device:
```
younesbelkada, line 130:
```suggestion
        elif device != module._parameters[tensor_name].device.index:
```
The `device` param is an integer so to properly compare the devices we need to access the index of the `device` attribute of the parameter ;)

sgugger: Ah should be `torch.device(device)` on the left side to cover all cases!

sgugger: Fixed [here](https://github.com/huggingface/accelerate/pull/576/commits/10334600357bdd014acd67bb255884ee10b54fe6) normally.

younesbelkada: Can confirm this fixes the issue too !

Issue #575, opened 2022-07-27
`gather_for_metrics` wrongly computes metrics in distributed setup


pacman100:
### System Info

```Shell
- `Accelerate` version: 0.12.0.dev0
- Platform: Linux-5.4.0-122-generic-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.23.0
- PyTorch version (GPU?): 1.12.0+cu102 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: fp16
        - use_cpu: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - downcast_bf16: no
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

1. add the below code to `complete_nlp_example.py`:
```diff
  for step, batch in enumerate(eval_dataloader):
              # We could avoid this line since we set the accelerator with `device_placement=True`.
              batch.to(accelerator.device)
              with torch.no_grad():
                  outputs = model(**batch)
              predictions = outputs.logits.argmax(dim=-1)
              # It is slightly faster to call this once, than multiple times
              predictions, references = accelerator.gather_for_metrics((predictions, batch["labels"]), eval_dataloader)
              metric.add_batch(
                  predictions=predictions,
                  references=references,
              )
+ accelerator.print(accelerator.gradient_state.samples_seen)
```

2. MRPC dataset has the following samples along diff splits:
```
DatasetDict({                                                                                                                  
    train: Dataset({                                                                                                           
        features: ['sentence1', 'sentence2', 'label', 'idx'],                                                                  
        num_rows: 3668                                                                                                         
    })                                                                                                                         
    validation: Dataset({                                                                                                      
        features: ['sentence1', 'sentence2', 'label', 'idx'],                                                                  
        num_rows: 408                                                                                                          
    })                                                                                                                         
    test: Dataset({                                                                                                            
        features: ['sentence1', 'sentence2', 'label', 'idx'],                                                                  
[...]

PR #578, opened 2022-07-27
Fix gather_for_metrics

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/575 by completely changing how we check for the number of samples to drop. Instead uses modulo to learn how many extra samples there are initially, allowing us to simplify the logic.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -952,8 +952,7 @@ class Accelerator:
                 if self.gradient_state.end_of_dataloader:
                     # Last batch needs to be truncated on distributed systems as it contains additional samples
                     def _adjust_samples(tensor):
-                        return tensor[: dataloader.total_dataset_length - self.gradient_state.samples_seen]
-
+                        return tensor[: (self.gradient_state.samples_seen * 2) - (dataloader.total_dataset_length - 4)]
                     return recursively_apply(_adjust_samples, tensor)
                 else:
                     # Not at the end of the dataloader, no need to adjust the tensors
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -934,7 +934,7 @@ class Accelerator:
         """
         return gather(tensor)
 
-    def gather_for_metrics(self, tensor, dataloader):
+    def gather_for_metrics(self, tensor):
         """
         Gathers `tensor` and potentially drops duplicates in the last batch if on a distributed system. Should be used
         for gathering the inputs and targets for metric calculation.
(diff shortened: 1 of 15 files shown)


src/accelerate/accelerator.py
```diff
@@ -934,25 +934,28 @@ def gather(self, tensor):
...
                 The tensors for calculating metrics across all processes.
-            dataloader (`torch.utils.data.DataLoader`):
-                A dataloader prepared with `Accelerator.prepare`
         """
         tensor = self.gather(tensor)
         if self.use_distributed:
+            if self.gradient_state.remainder == -1:
+                logger.warn(
+                    "The used dataset had no length, returning gathered tensors. You should drop the remainder yourself."
+                )
```
sgugger, line 951: I think this can be an info, no need for a warning here.

src/accelerate/data_loader.py
```diff
@@ -332,7 +336,11 @@ def __iter__(self):
         if self.rng_types is not None:
             synchronize_rng_states(self.rng_types, self.generator)
         self.gradient_state._set_end_of_dataloader(False)
-        self.gradient_state._set_samples_seen(0)
+        try:
+            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
+            self.gradient_state._set_remainder(length % self.batch_size)
+        except:
```
sgugger, line 342: There should be an exception ehre :-)

src/accelerate/test_utils/scripts/test_metrics.py
```diff
@@ -22,50 +22,63 @@
...
+    inps, targs = torch.cat(inps), torch.cat(targs)
+    return inps, targs
+
+
+def test_torch_metrics(accelerator: Accelerator, num_samples=82):
+    model, ddp_model, dataloader = get_setup(accelerator, num_samples)
+    inps, targs = generate_predictions(ddp_model, dataloader, accelerator)
+    assert (
+        len(inps) == num_samples
+    ), f"Unexpected number of inputs:\n    Expected: {num_samples}\n    Actual: {len(inps)}"
```
sgugger, line 58: Nice new test!

sgugger (approved):
Very nice fix!

sgugger (approved):

src/accelerate/data_loader.py
```diff
@@ -453,7 +463,11 @@ def _fetch_batches(self, iterator):
 
     def __iter__(self):
         self.gradient_state._set_end_of_dataloader(False)
-        self.gradient_state._set_samples_seen(0)
+        try:
+            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
+            self.gradient_state._set_remainder(length % self.batch_size)
```
pacman100, line 468:
```suggestion
self.gradient_state._set_remainder(length % self.total_batch_size)
```

pacman100 (commented):
Great fix, simple and intuitive 😄. I think the below change is required in `DataLoaderDispatcher`.

src/accelerate/data_loader.py
pacman100: After the above change or without it, `test_metrics.py` fails when `dispatch_batches=True` when the `num_samples % num_processes != 0` , e.g., num_samples=97 and num_processes=2. Below is the modified test with print statements. I think, `IterableDatasetShard` should be used in conjunction with  `dispatch_batches=True` for expected behaviour or `DataLoaderDispatcher` needs to be fixed to handle normal dataloaders. 

Screenshot 2022-07-28 at 10 54 03 AM

muellerzr: Thanks, will change the test case slightly for this and resolve today!

PR #579, opened 2022-07-28
minor FSDP launcher fix

pacman100:
### What does this PR do?
1. FSDP launcher fix when not using a config file.


sgugger (approved):
Thanks for the fix!

PR #580, opened 2022-07-28
Skip and raise NotImplementedError for gather_for_metrics for now

muellerzr:
Raises an err since there's still a number of bugs we're discovering with this implementation.

Related discussions:

https://github.com/huggingface/accelerate/issues/575
https://github.com/huggingface/accelerate/pull/578


sgugger (approved):
Good for now while we fix the bug.

PR #581, opened 2022-07-28
Use main_process_first in examples

muellerzr:
Adds usage of `main_process_first` when calling `map()` in the examples


sgugger (approved):
Thanks for adding this!

PR #584, opened 2022-07-29
fix metric computation in examples

pacman100:
### What does this PR do?
1. This PR fixes metric computation for the `nlp_example.py`. Once everything seems fine, we can do this for all the other scripts.


muellerzr (approved):
Great! Thanks for this. This will also help me with understanding how to retry the gather_for_metrics as well. Thanks

pacman100:
This has a bug and the original indeed was correct 😅. Closing this.

sgugger:
Distributed evaluation is hard 😅

PR #588, opened 2022-08-01
Fix DispatchDataloader

sgugger:
This PR fixes the dispatch dataloader which was a bit broken by the recent refactors. It's untested (since it requires a multiGPU setup) but will be in the new evaluation testing @muellerzr will add soon.

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -400,7 +400,7 @@ class DataLoaderDispatcher(DataLoader):
         - **total_dataset_length** (`int`) -- Total length of the inner dataset across all processes.
     """
 
-    def __init__(self, dataset, split_batches: bool = False, **kwargs):
+    def __init__(self, dataset, split_batches: bool = False, _drop_last: bool = False, **kwargs):
         shuffle = False
         if is_torch_version(">=", "1.11.0"):
             from torch.utils.data.datapipes.iter.combinatorics import ShufflerIterDataPipe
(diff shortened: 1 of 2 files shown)


src/accelerate/data_loader.py
```diff
@@ -400,7 +400,7 @@ class DataLoaderDispatcher(DataLoader):
         - **total_dataset_length** (`int`) -- Total length of the inner dataset across all processes.
     """
 
-    def __init__(self, dataset, split_batches: bool = False, **kwargs):
+    def __init__(self, dataset, split_batches: bool = False, _drop_last: bool = False, **kwargs):
```
sgugger, line 403: PyTorch does not let us pass `drop_last`.

src/accelerate/data_loader.py
```diff
@@ -485,34 +492,31 @@ def __iter__(self):
...
+                # Batch size computation above is wrong, it's off by 1 so we fix it.
                 batch_size += 1
 
             data_slice = slice(self.state.process_index * batch_size, (self.state.process_index + 1) * batch_size)
-            next_batch, next_batch_info, next_skip = self._fetch_batches(main_iterator)
             batch = slice_tensors(batch, data_slice)
-            if not self._stop_iteration:
-                yield batch
-                batch, batch_info, skip = next_batch, next_batch_info, next_skip
-            else:
```
sgugger, line 506: Removed this logic as we don't need to take the next batch to have an informed decision on whether we are the last batch or not. This was actually causing the dataloader to stop one batch too early.

muellerzr (approved):
Thanks for iterating and working with me on this 🤗

Issue #592, opened 2022-08-02
How to use data visualization API with accelerate.tracking?


zh-plus:
The accelerate.tracking API only supports `accelerator.log`, which can be used to log key-value pairs. 

However, we can't use `accelerator` to visualize the data and predictions. For example, in [wandb docs](https://wandb.ai/stacey/mnist-viz/reports/Visualize-Predictions-over-Time--Vmlldzo1OTQxMTk), we need `wandb.run.log_artifact(test_data_at)` to sync the visualization table. 

In https://github.com/huggingface/accelerate/blob/29a222a2619a56636cc80fe62ba2e925f3f5d064/src/accelerate/tracking.py#L193 the `self.run` is initialized and used in log function like this: https://github.com/huggingface/accelerate/blob/29a222a2619a56636cc80fe62ba2e925f3f5d064/src/accelerate/tracking.py#L224

Right now if I want to access `self.run` (suppose I just use wandb), we need to iterate all the trackers in `Accelerator` like this:
```
wandb_run = accelerator.trackers[0].run
wandb_run.log_artifact(test_data_at)
```
It is kind of weird to access the internal list. Can we add a function in `Accelerator` such that `accelerator.wandb_run()` return the `accelerator.trackers[0].run`? The same as `accelerator.tensorboard_writer()` and `accelerator.cometml_writer()`.

muellerzr:
Thanks! Should be good now with https://github.com/huggingface/accelerate/pull/594 now 😄

PR #594, opened 2022-08-02
Shorthand way to grab a tracker

muellerzr:
# Shorthand way to grab a tracker

## What does this add?

This PR introduces a method to the Accelerator that lets you grab a tracker's internal run if some custom code with it is wanted to be ran

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/592

## Why is it needed?

There are needs when using `wandb` for example to access and toy with the `run`, such as stated in #592. Currently you need to index into the trackers, hope you find the right one, and then in the case of wandb call `run.run` which is a bit convoluted. 

### User-facing:

Adds a new `Accelerator.get_tracker` method. To get a tracker you should pass in the corresponding string to the `.name` attribute

### Internal structure:

Trackers now have a new abstract property of `tracker` which should be used to return the internal tracking mechanism normal users would expect to interact with.

## Basic Usage Example(s):

```python
accelerator = Accelerator(log_with="wandb")

accelerator.init_trackers("my_project_name")

    wandb_run = accelerator.get_tracker('wandb')
    wandb_run.log_artifact(test_data_at_xyz)
```

When custom interactions with a tracker is wanted.

### DIFF
--- a/docs/source/usage_guides/tracking.mdx
+++ b/docs/source/usage_guides/tracking.mdx
@@ -33,7 +33,7 @@ accelerator = Accelerator(log_with="wandb")
 accelerator = Accelerator(log_with=["wandb", LoggerType.TENSORBOARD])
 ```
 
-At the start of your experiment [`~Accelerator.init_trackers`] should be used to setup your project, and potentially add any experiment hyperparameters to be logged:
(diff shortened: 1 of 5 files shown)


sgugger (approved):
Thanks for adding this!

PR #595, opened 2022-08-02
Pin deepspeed

muellerzr:
Pins deepspeed for now as their latest pypi release has broken the CI installation. 

Deepspeed issue: https://github.com/microsoft/DeepSpeed/issues/2166


sgugger (approved):
Thanks for fixing this!

PR #598, opened 2022-08-02
Improve warning

muellerzr:
Noticed that these didn't quite get fixed properly before the merge happened


sgugger (approved):
Thanks for fixing!

Issue #597, opened 2022-08-02
allow_val_change parameter is not passed to wandb.config.update in the WandBTracker


plamb-viso:
As far as I know, setting `allow_val_change=True` when calling `wandb.init(...)` allows you to have hyperparameter values that change over time (like a scheduled learning rate). However, for this to work, `allow_val_change` also needs to be passed to `wandb.config.update(...)`. The stub can be viewed here:

https://github.com/wandb/wandb/blob/3966b7740896127c88bf2c143a8dd5114b5de20d/wandb/sdk/wandb_config.py#L181

And it is not passed to update in accelerate's code:

https://github.com/huggingface/accelerate/blob/b52b793ea8bac108ba61192eead3cf11ca02433d/src/accelerate/tracking.py#L223

This results in exceptions like:

```
[1,mpirank:0,algo-1]<stderr>:  File "/opt/ml/code/viso_ml_train/training/train.py", line 321, in <module>#015
--
[1,mpirank:0,algo-1]<stderr>:    accelerator.init_trackers(#015
[1,mpirank:0,algo-1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/accelerate/accelerator.py", line 1070, in init_trackers#015
[1,mpirank:0,algo-1]<stderr>:    tracker.store_init_configuration(config)#015
[1,mpirank:0,algo-1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/accelerate/tracking.py", line 223, in store_init_configuration#015
[1,mpirank:0,algo-1]<stderr>:    wandb.config.update(values)#015
[1,mpirank:0,algo-1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/wandb/sdk/wandb_config.py", line 178, in update#015
[1,mpirank:0,algo-1]<stderr>:    sanitized = self._update(d, allow_val_change)#015
[1,mpirank:0,algo-1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/wandb/sdk/wandb_config.py", line 171, in _update#015
[1,mpirank:0,algo-1]<stderr>:    sanitized = self._sanitize_dict(#015
[1,mpirank:0,algo-1]<stderr>:  File "/opt/conda/lib/python3.8/site-packages/wandb/sdk/wandb_config.py", line 231, in _sanitize_dict#015
[1,mpirank:0,algo-1]<stderr>:    k, v = self._sanitize(k, v, allow_val_change)#015
[...]

PR #599, opened 2022-08-02
Pass in config_kwargs to init

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/597

Lets you pass in a new `config_kwargs` to `init_trackers` that lets you set any config kwargs if they are supported by that tracker kind. (Currently just wandb)


sgugger (commented):
This is becoming way too confusing. Let's let the users register and initialize their trackers themselves for those cases as having two different dictionaries of kwargs in a method is not a very sensible API.

PR #604, opened 2022-08-04
fixing deepspeed slow tests issue

pacman100:
### What does. this PR do?
1. Disabling DeepSpeed and FSDP tests that are overwhelming the RAM of the CI instance running slow tests (Mostly CPU offload related tests). Now, `make test` passes end to end. Below is the report of the tests run on an instance with specifications matching the self hosted runners. 

[report_multi_gpu_tests.txt](https://github.com/huggingface/accelerate/files/9259282/report_multi_gpu_tests.txt)


sgugger (approved):
Thanks for fixing!

muellerzr (approved):
Thanks! Makes sense for right now

PR #605, opened 2022-08-04
Remove redundant `.run` in `WandBTracker`.

zh-plus:
Fix the "AttributeError: 'Run' object has no attribute 'run'" error caused by the redundant `.run` in `WandBTracker`.


muellerzr:
Thanks!

muellerzr (approved):

PR #606, opened 2022-08-04
Correct dtype casting

younesbelkada:
# What does this PR do?

This PR fixes a small "pre-bug" where in some very niche cases model buffers are saved in `int`, `uint` or `bool`. Therefore sometimes there is no need to cast their `dtype` to the desired `dtype` and we want to keep them in their native `dtype`.

I don't know if this is relevant but I ran `tests/test_big_modeling.py` slow tests to make sure this does not break anything on a multi-GPU setup (2x Tesla T4 16GB)

cc @sgugger @muellerzr


sgugger (approved):
Thanks a lot!

PR #607, opened 2022-08-05
Add `non_blocking` kwarg to `send_to_device()`

NouamaneTazi:
The .to() seems to throttle the forward pass, by creating a `cudaStreamSynchronize` everytime the `AlignDevicesHook` is called.
For example, I managed to reduce Walltime for a single forward pass of `bloom-175b` by 15% (with profiling on) by applying this fix.

cc @sgugger


sgugger (commented):
Thanks for your PR!

Since this method is used everywhere (and not just in the big model inference) can we put the `non_blocking` argument (with default to `True`) as an arg to `send_to_device`? This way if we see it causes some problem in another part of the lib, we can set it off without touching big model inference.

Chris-hughes10:
Hi @sgugger, was there a reason that this was closed and not merged? I came across this when I was about to request the same thing.

sgugger:
@Chris-hughes10 it was automatically closed by the stale bot due to the absence of activity. @NouamaneTazi is still investigating whether this could have some negative impact sometimes.

src/accelerate/utils/operations.py
```diff
@@ -117,13 +117,13 @@ def send_to_device(tensor, device):
         The same data structure as `tensor` with all tensors sent to the proper device.
     """
 
-    def _send_to_device(t, device):
-        return t.to(device)
+    def _send_to_device(t, device, non_blocking):
+        return t.to(device, non_blocking=non_blocking)
 
     def _has_to_method(t):
         return hasattr(t, "to")
```
NouamaneTazi, line 124: So far all objects in Pytorch which accept `.to()` method accept the `non_blocking` kwarg or at least kwargs in general, but are we sure this is going to be always the case?

- [torch.jit.ScriptModule.to](https://pytorch.org/docs/stable/generated/torch.jit.ScriptModule.html?highlight=#torch.jit.ScriptModule.to) (Python method, in ScriptModule)
- [torch.nn.Module.to](https://pytorch.org/docs/stable/generated/torch.nn.Module.html?highlight=#torch.nn.Module.to) (Python method, in Module)
- [torch.nn.utils.rnn.PackedSequence.to](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.PackedSequence.html?highlight=#torch.nn.utils.rnn.PackedSequence.to) (Python method, in PackedSequence)
- [torch.Tensor.to](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html?highlight=#torch.Tensor.to) (Python method, in torch.Tensor.to)

sgugger: This is for our own objects FYI, `BatchEncoding` in Transformers, which does not accept the `non_blocking` argument :-/
Maybe we can inspect the signature and only pass it when we see it?

Chris-hughes10:
@NouamaneTazi These are really interesting insights! If you are looking at inspecting signatures, do you think it would be too much complexity to set non-blocking automatically based on whether it is CPU -> GPU transfer or GPU -> CPU transfer by inspecting the tensor's current device?

src/accelerate/utils/operations.py
sgugger: Mmm, `inspect.signature(t.to)` should give you access to the keyword arguments. You can find them in the `.parameters` attribute, which will be a dict-like object.

NouamaneTazi:
CPU -> GPU and GPU -> CPU both lead to the same issues as mentioned above. Only GPU -> GPU is the safe operation but as I said above, it seems that it requires the two GPUs synchronization whether we set `non_blocking=True` or not

src/accelerate/utils/operations.py
NouamaneTazi: I get this error when trying `inspect.signature(t.to)` where `t` is a tensor
```
  File "<string>", line 1, in <module>
  File "/home/nouamane/miniconda3/envs/hf/lib/python3.9/inspect.py", line 3111, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/home/nouamane/miniconda3/envs/hf/lib/python3.9/inspect.py", line 2860, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/home/nouamane/miniconda3/envs/hf/lib/python3.9/inspect.py", line 2327, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
  File "/home/nouamane/miniconda3/envs/hf/lib/python3.9/inspect.py", line 2145, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <built-in method to of Tensor object at 0x7f7e98c57720>
```

Chris-hughes10:
> @Chris-hughes10 CPU -> GPU and GPU -> CPU both lead to the same issues as mentioned above. Only GPU -> GPU is the safe operation but as I said above, it seems that it requires the two GPUs synchronization whether we set `non_blocking=True` or not

Apologies, I misread the post above. Please ignore my previous suggestion!

src/accelerate/utils/operations.py
NouamaneTazi: I think this is related to the warning in [inspect's docs](https://docs.python.org/3/library/inspect.html#inspect.signature)

> Note: Some callables may not be introspectable in certain implementations of Python. For example, in CPython, some built-in functions defined in C provide no metadata about their arguments.

sgugger: Maybe we can isolate tensors in the function then? Something like:
```
if isinstance(t, torch.Tensor):
    return t.to(device, non_blocking=non_blocking)
else:
    # inspect signature here in a try/except block
```

NouamaneTazi: Wouldn't this be enough, or should I inspect signature?
  ```python
  def _send_to_device(t, device, non_blocking):
        if isinstance(t, torch.Tensor):
            return t.to(device, non_blocking=non_blocking)
        else:
            try:
                return t.to(device, non_blocking=non_blocking)
            except TypeError:
                return t.to(device)
```

Comment (sgugger) [reply src/accelerate/utils/operations.py:124]: Oh in this case, you don't even need the first test. Are you sure about the type of `TypeError`?

Comment (NouamaneTazi) [reply src/accelerate/utils/operations.py:124]: Seems to work for C defined methods as well:
```python
>> t.__abs__(a=5)
  File "<string>", line 1, in <module>
TypeError: Tensor.abs() takes no keyword arguments
```


```python
>> t.to(a=5)
  File "<string>", line 1, in <module>
TypeError: to() received an invalid combination of arguments - got (a=int, ), but expected one of:
 * (torch.device device, torch.dtype dtype, bool non_blocking, bool copy, *, torch.memory_format memory_format)
 * (torch.dtype dtype, bool non_blocking, bool copy, *, torch.memory_format memory_format)
 * (Tensor tensor, bool non_blocking, bool copy, *, torch.memory_format memory_format)
```

sgugger: Nice!

src/accelerate/utils/operations.py
```diff
@@ -103,7 +103,7 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
     return data
 
 
-def send_to_device(tensor, device):
+def send_to_device(tensor, device, non_blocking=False):
     """
     Recursively sends the elements in a nested list/tuple/dictionary of tensors to a given device.
```
NouamaneTazi, line 108: @sgugger the docs say that `tensor` is a torch.Tensor here, which isn't necessary the case iiuc. We can probably update the docs as well

sgugger (approved):

src/accelerate/utils/operations.py
sgugger: True

NouamaneTazi: Should I replace `tensors` with `tensor-like objects`?

PR #611, opened 2022-08-08
Make running tests more efficient

muellerzr:
Makes our tests only be ran when they really should be (aka when python source code, tests, or dependencies have been changed). This removes the lockup on merging PRs and having to wait for all the tests to pass on our single CI runner


sgugger (approved):
Thanks for working on this!
I think you should just include the github action files (test should be run if we change the files implementing the tests)

ydshieh:
So the purpose of `check-for-source` in `build_and_run_tests.yml` and `test.yml` are different?

- on PR request: check source file, and run the tests only if any changed found.
- on merge (to main): check `setup.py` to build the image if necessary. Then always run the tests?

ydshieh:
I am probably wrong above, as I see you have `needs: build-docker-containers`. 
But I am confused a bit: currently (without this PR), what happens if `setup.py` is not changed, but some `.py` files changed?

I will review in more detail tomorrow.

muellerzr:
@ydshieh if some `.py` was changed but not the setup then the tests are ran but Docker images won't be rebuilt

ydshieh:
Thanks, @muellerzr . So in this PR, what happens if no `.py` file change (`setup.py` not change + no other `.py` change)? If I understand, the tests still run. Is this the expected behavior?

ydshieh:
My only remaining question is as above: on merge, should we skip the tests if no `.py` file changes at all. Otherwise, all LGTM.

muellerzr:
@ydshieh (sorry this got buried in my priorities):

Re: on merge, yes it is because we never know if some other dependency has a regression issue or what. So we're still made aware of when this happens post merge and can immediately know it's unrelated to something directly in that PR without it silently not failing until the nightlies

sgugger (approved):
LGTM!

.github/workflows/build_and_run_tests.yml
```diff
@@ -10,9 +10,9 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
-  check-for-setup:
+  check-for-source:
     runs-on: ubuntu-latest
-    name: Check if setup was changed
+    name: Checks if source code or tests were changed
```
ydshieh, line 15: I feel this name should be `Checks if setup.py were changed`.

ydshieh (approved):
Thank you @muellerzr . LGTM, but maybe the job name should be changed. Leave you to make the final decision.

PR #612, opened 2022-08-08
Fix DeepSpeed CI

muellerzr:
Fixes the DeepSpeed CI on the runner by just separating out the integration tests to be ran in a later python process, avoiding the lockup and issues.

Passed CI can be seen here on multi-gpu: https://github.com/huggingface/accelerate/actions/runs/2819545133

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -60,7 +60,9 @@ jobs:
       - name: Run test on GPUs
         run: |
           source activate accelerate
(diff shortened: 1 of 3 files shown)


sgugger (approved):
Works for me! Thanks for fixing!

pacman100:
Thank you 😄. This is neat and finally all the tests would run as expected 🤗.

pacman100 (approved):

Issue #616, opened 2022-08-09
Gradient Accumulation Documentation has optimizer.zero_grad() at wrong step


JeremyAlain:
### System Info

```Shell
python=3.7
accelerate==0.12.0
torch==1.10.0
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks
- [ ] My own task or dataset (give details below)

### Reproduction

The official accelerate [documentation](https://huggingface.co/docs/accelerate/usage_guides/gradient_accumulation) for the new accumulate gradients wrapper is wrong! 

Here is the Code that is depicted on the website: 
```for batch in training_dataloader:
    with accelerator.accumulate(model):
        optimizer.zero_grad()
        inputs, targets = batch
        outputs = model(inputs)
        loss = loss_function(outputs, targets)
        accelerator.backward(loss)
        optimizer.step()
        scheduler.step()
```

On the official [Pull Request ](https://github.com/huggingface/accelerate/pull/484)of the feature the code is correct and looks like this:
```for batch in dataloader:
    with accelerator.accumulate(model):
        output = model(**batch)
        loss = output.loss
        accelerator.backward(loss)
        optimizer.step()
        lr_scheduler.step()
        optimizer.zero_grad()
```

The difference lies in the position of `optimizer.zero_grad()`. The version in the Pull Request is correct, while the version of the official website is incorrect. Let me explain: 

Let's look at the first code version (as on the documentation) and let's assume we have 4 accumulation steps. The way that the wrapper `accelerator.accumulate(model)`works is that it sets a field `synchronize` to true every 4 accumulation steps. This leads to the gradient being backpropagated with `optimizer.step()`every 4 steps. It also has the effect that `optimizer.zero_grad()` is internally only called when we reach the 4th accumulation step (before it is just skipped internally). 
What happens with this code is that for the first 3 steps, we accumulate gradients. We then reach the 4th step and `synchronize` is set to True. Since we then FIRST call `optimizer.zero_grad()` we clear all previous gradients (from the 3 previous steps). We then calculate the gradient for the current step and backpropagate that. This essentially means that instead of accumulation 4 gradients, we deleted 3 gradients and only used the last gradient to backpropagate. 

### Expected behavior
The position of `optimizer.zero_grad()`is correct at the end of the training loop i.e. it should come after `optimizer.step()` (as in the Code example of the Pull Request). This way the gradient accumulation works as expected and calculates the gradient of all `accumulate_gradient` steps.

[...]

PR #620, opened 2022-08-09
Fix grad accum example in docs

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/616


sgugger (approved):
Thanks for fixing!

PR #625, opened 2022-08-11
M1 mps fixes

pacman100:
### What does this PR do?
1. setting env variable "PYTORCH_ENABLE_MPS_FALLBACK" = "1" when using `mps` 
2. Warning which strongly recommends users to use PyTorch nightly (1.13.dev...) as per the issue https://github.com/pytorch/pytorch/issues/82707

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -226,7 +226,7 @@ class AcceleratorState:
 
                         if not is_torch_version(">", "1.12.0"):
                             warnings.warn(
-                                "We strongly recommend to install the latest PyTorch nightly version (1.13.0.dev...) on your MacOS machine. "
+                                "We strongly recommend to install PyTorch >= 1.13 (nightly version at the time of writing) on your MacOS machine. "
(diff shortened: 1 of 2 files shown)


src/accelerate/state.py
```diff
@@ -221,6 +222,14 @@ def __init__(
                                 "and/or you do not have an MPS-enabled device on this machine."
                             )
                     else:
+                        from .utils import is_torch_version
+
+                        if not is_torch_version(">", "1.12.0"):
+                            warnings.warn(
+                                "We strongly recommend to install the latest PyTorch nightly version (1.13.0.dev...) on your MacOS machine. "
```
sgugger, line 229: That warning may not age well once PyTorch 1.13 is released. Maybe:
```suggestion
                                "We strongly recommend to install PyTorch >= 1.13 (nightly version at the time of writing) on your MacOS machine. "
```

sgugger (approved):
Thanks for the fixes!

PR #626, opened 2022-08-11
DeepSpeed launcher related changes

pacman100:
### What does this PR do?
1. Removing 1 sub-process call for DeepSpeed for `Single Node Multi-GPU setup` and `Multi Node Multi-GPU setup using Standard launcher`.
2. Also resolves issue #663 
- [x] As discussed offline, the `distrib_run.run(distrib_args)` is stuck indefinitely for `Multi Node Multi-GPU setup` even for standard DDP. This needs to be fixed as that will solve `Multi Node Multi-GPU setup using Standard launcher` of DeepSpeed integration (Other DeepSpeed launchers like PDSH are working fine for Multi Node Multi-GPU) - Resolved

PR #627, opened 2022-08-11
Fix multi-node issues and simplify param logic

muellerzr:
Fixes the multi-gpu logic raised [here](https://github.com/huggingface/accelerate/pull/626/files) and also simplifies how args are setup for the namespace object


sgugger (approved):
Thanks for the fixes!

PR #630, opened 2022-08-12
minor tracker fixes for complete* examples

pacman100:
# What does this PR do?
1. `with_tracking` was failing for complete example scripts with error `/ in name of the run` (logs/complete_... was being used as run). This fixes it.
2. A minor fix in printing accelerate state's mixed precision


sgugger (approved):
Thanks for the fixes, LGTM but would also like to have @muellerzr look at it.

muellerzr (approved):
Thanks!

PR #631, opened 2022-08-12
Use torchrun for multinode

muellerzr:
Sets `torchrun` to be the only way to run multi-node, and adds a `rdvz_endpoint` param during config setup.


sgugger (approved):
LGTM, thanks!

PR #633, opened 2022-08-12
accelerate requires torch 1.9+

jeffra:
We noticed our DeepSpeed + Accelerate unit tests are failing on torch 1.8. `torch.distributed.run` requires torch 1.9+ so bumping your min torch version to 1.9. If you'd rather guard the torch run import depending on torch version feel free to close this PR :)

https://github.com/microsoft/DeepSpeed/runs/7810439796?check_suite_focus=true

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
         ]
     },
     python_requires=">=3.7.0",
-    install_requires=["numpy>=1.17", "packaging>=20.0", "psutil", "pyyaml", "torch>=1.4.0", "rich"],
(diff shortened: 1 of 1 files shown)


muellerzr:
Thanks! That's an issue I caused 😓 It should be only on multinode we require that, so will put back in the import check!

muellerzr:
@jeffra should be fixed now on main, can you try rerunning your tests? 😄

jeffra:
Thank you @muellerzr, I think we're good now :)

PR #637, opened 2022-08-15
Add static_graph arg to DistributedDataParallelKwargs.

rom1504:
supported by https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html

This is particularly useful when using gradient checkpointing

See https://discuss.pytorch.org/t/ddp-and-gradient-checkpointing/132244/3 for more details

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -60,6 +60,8 @@ class DistributedDataParallelKwargs(KwargsHandler):
 
     `gradient_as_bucket_view` is only available in PyTorch 1.7.0 and later versions.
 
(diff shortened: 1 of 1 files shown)


muellerzr:
Thanks! Could we add a warning to this please similar to the other one that exists for these kwargs mentioning its only available on pytorch>=1.11?

rom1504:
Sure, done

muellerzr:
@rom1504 looks like the last thing we need is to run `make style; make quality` :)

rom1504:
@muellerzr fixed

muellerzr:
Great! Thanks!!

PR #638, opened 2022-08-15
Test for min torch version + fix import issue

muellerzr:
Adds a workflow for testing the minimal torch version to make sure we always support it.

### DIFF
--- a/src/accelerate/__init__.py
+++ b/src/accelerate/__init__.py
@@ -17,7 +17,11 @@ from .utils import (
     find_executable_batch_size,
     infer_auto_device_map,
     is_torch_version,
+    is_rich_available,
     load_checkpoint_in_model,
-    rich,
     synchronize_rng_states,
 )
+
+
+if is_rich_available():
+    from .utils import rich
--- a/src/accelerate/__init__.py
+++ b/src/accelerate/__init__.py
@@ -5,6 +5,7 @@
 __version__ = "0.13.0.dev0"
 
 from .accelerator import Accelerator
+from .big_modeling import cpu_offload, disk_offload, dispatch_model, init_empty_weights, load_checkpoint_and_dispatch
 from .launchers import debug_launcher, notebook_launcher
 from .utils import (
     DeepSpeedPlugin,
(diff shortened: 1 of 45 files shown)


src/accelerate/__init__.py
```diff
@@ -16,7 +15,18 @@
...
+
+
+if is_torch_version(">=", "1.9.0"):
+    from .big_modeling import (
+        cpu_offload,
+        disk_offload,
+        dispatch_model,
+        init_empty_weights,
+        load_checkpoint_and_dispatch,
+    )
```
sgugger, line 32: It should be possible to import those regardless of the version (and then the actual use of it would trigger the error).

src/accelerate/checkpointing.py
```diff
@@ -48,7 +49,7 @@ def save_accelerator_state(
     optimizers: list,
     schedulers: list,
     process_index: int,
-    scaler: GradScaler = None,
+    scaler=None,
```
sgugger, line 52: Let's leave the type hint but as a forward reference?

src/accelerate/test_utils/testing.py
```diff
@@ -132,6 +133,16 @@ def require_fsdp(test_case):
     return unittest.skipUnless(is_torch_version(">=", "1.12.0"), "test requires torch version >= 1.12.0")(test_case)
 
 
+def require_torch_version(test_case=None, version=None):
```
sgugger, line 136:
```suggestion
def require_torch_min_version(test_case=None, version=None):
```
Naming isn't completely right IMO.

sgugger (approved):
Nice idea in principle (and then we should have scheduled jobs for tests of each intermediate version!). We can also move to pin 1.6.0 as a min dependency as part of this PR (which should simplify things). As discussed internally, I think two years of previous PyTorch versions is enough.

src/accelerate/checkpointing.py
muellerzr: Not needed since we're raising the min torch

src/accelerate/__init__.py
muellerzr: Fixed

sgugger (approved):
Thanks for the fixes!

PR #642, opened 2022-08-19
make `init_trackers` to launch on main process

Gladiator07:
While on distributed setup the `accelerator.init_trackers()` method initializes multiple runs of the tracker on all processes. It should be initialized on the main process. Added a `@on_main_process` decorator to achieve the expected behavior.

@muellerzr @sgugger

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1032,6 +1032,7 @@ class Accelerator:
         """
         wait_for_everyone()
 
+    @on_main_process
     def init_trackers(self, project_name: str, config: Optional[dict] = None, init_kwargs: Optional[dict] = {}):
         """
         Initializes a run for all trackers stored in `self.log_with`, potentially with starting configurations


muellerzr (approved):
Thanks! For right now I think this is fine, and if folks don't like it we can revert back or make it optional to do so. There may be situations when spinning up a tracker on all the processes could be useful (and then manually logging them instead), so we'll keep an eye out for this on feedback

muellerzr:
@Gladiator07 could you adjust the following code in the examples as well? Since this removes the need for the check:

https://github.com/huggingface/accelerate/blob/main/examples/by_feature/tracking.py#L168-L172

Gladiator07:
@muellerzr adjust as in? Should I remove the `accelerator.is_main_process` check?

muellerzr:
@Gladiator07 correct. We're putting this in place so that is no longer needed, so the examples should show this :)

Gladiator07:
Cool, doing this in a separate PR.

PR #643, opened 2022-08-19
remove check for main process for trackers initialization

Gladiator07:
Made the changes as discussed in [this PR](https://github.com/huggingface/accelerate/pull/642#issuecomment-1220651885).

@muellerzr

### DIFF
--- a/examples/by_feature/deepspeed_with_config_support.py
+++ b/examples/by_feature/deepspeed_with_config_support.py
@@ -588,14 +588,12 @@ def main():
         checkpointing_steps = None
 
     # We need to initialize the trackers we use, and also store our configuration.
-    # We initialize the trackers only on main process because `accelerator.log`
-    # only logs on main process and we don't want empty logs/runs on other processes.
+    # The trackers initializes automatically on the main process.
     if args.with_tracking:
-        if accelerator.is_main_process:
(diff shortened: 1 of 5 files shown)


muellerzr:
Thanks! Do you want to try to update the complete examples in this too? (The last spot would be in the docs, but I can update that :) )

Gladiator07:
Sure, I will update all the example scripts in the `examples` folder in this PR :)

muellerzr:
@Gladiator07 great! It should just be the complete_nlp_example and complete_cv_example

Gladiator07:
@muellerzr but many of the example scripts have the `is_main_process` check (for instance [here](https://github.com/huggingface/accelerate/blob/a3d94916a87d07471865261d015128870f513c38/examples/by_feature/deepspeed_with_config_support.py#L594)). Should I update them as well or just `complete_nlp_example` and `complete_cv_example` ?

muellerzr:
Ah, great point! Yes in that case those would be great to have done as well! :)

Gladiator07:
Hi @muellerzr, I have removed the `is_main_process` check from all example scripts. Please check it once.

muellerzr (approved):
Great work!

PR #649, opened 2022-08-22
Add aim tracker for accelerate

muellerzr:
Adds support for [Aim](https://github.com/aimhubio/aim)

cc @gorarakelyan to make sure it all seems alright with you! I thought about an easy `set_context` func but since our trackers all work off an abstract base class, seemed easier to just leave it as a kwarg. Let me know if there's something I seem to have missed! (Based this off my no-accelerate training script I wrote)

`Accelerator` will automatically only log on the main process and initialize the trackers on the main process, hence why there isn't any special "Are we on the main process" things in this PR.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -20,7 +20,7 @@ from abc import ABCMeta, abstractmethod, abstractproperty
 from typing import List, Optional, Union
 
 from .logging import get_logger
-from .utils import LoggerType, is_comet_ml_available, is_tensorboard_available, is_wandb_available
+from .utils import LoggerType, is_aim_available, is_comet_ml_available, is_tensorboard_available, is_wandb_available
 
 
 _available_trackers = []
@@ -39,6 +39,11 @@ if is_comet_ml_available():
     from comet_ml import Experiment
 
     _available_trackers.append(LoggerType.COMETML)
+
+if is_aim_available():
+    from aim import Run
+
+    _available_trackers.append(LoggerType.AIM)
 
 
 logger = get_logger(__name__)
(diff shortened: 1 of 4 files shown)


src/accelerate/tracking.py
```diff
@@ -320,7 +325,74 @@ def finish(self):
...
+        context = kwargs.pop("context", {})
+        epoch = kwargs.pop("epoch", None)
+        for key, value in values.items():
+            self.writer.track(value, key, step=step, epoch=epoch, context=context)
+
+    def finish(self):
+        """
+        Closes `aim` writer
+        """
+        self.writer.finalize()
```
gorarakelyan, line 387: @muellerzr the `finalize` method is more for an internal use (it handles storage indexing), which will become a private method in the upcoming releases.
The resources cleanup is done in `close` method, which should be used to properly finalize the run.

src/accelerate/tracking.py
```diff
@@ -320,7 +325,74 @@ def finish(self):
...
+                The run step. If included, the log will be affiliated with this step.
+            kwargs (`dict`, *optional*):
+                Additional key word arguments passed along to the `Run.track` method. Valid keys include `context` and
+                `epoch`.
+        """
+        # Note: replace this with the dictionary support when merged
+        context = kwargs.pop("context", {})
+        epoch = kwargs.pop("epoch", None)
+        for key, value in values.items():
+            self.writer.track(value, key, step=step, epoch=epoch, context=context)
```
gorarakelyan, line 381: What about passing the name argument as a keyword argument? I guess it will be more readable and will minimize the risk of forward incompatibilities.

gorarakelyan (commented):
hey @muellerzr! This looks awesome! 
Left a few minor recommendations to make the usage experience even smoother. Hope it makes sense to you.

src/accelerate/tracking.py
muellerzr: If you're worried about that, we can absolutely do so. Easy to achieve!

src/accelerate/tracking.py
```diff
@@ -320,7 +325,74 @@ def finish(self):
...
+        run_name (`str`):
+            The name of the experiment run.
+        kwargs:
+            Additional key word arguments passed along to the `Run.__init__` method.
+    """
+
+    name = "aim"
+    requires_logging_directory = True
+
+    def __init__(self, run_name: str, logging_dir: Optional[Union[str, os.PathLike]] = ".", **kwargs):
```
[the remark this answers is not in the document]
muellerzr: Interesting. I've been doing it the opposite way in my experiments. So e.g. if I have 3 trials of X hyperparameter, I'd just name it `run_hparam_x` *not* `run_hyparam_x_1`?

src/accelerate/tracking.py
muellerzr: @gorarakelyan I'd recommend updating this example, as this is where I got it from :) Will use close

https://colab.research.google.com/drive/14rIAjpEyklf5fSMiRbyZs6iYG7IVibcI?usp=sharing#scrollTo=qOu7icths6n5

src/accelerate/tracking.py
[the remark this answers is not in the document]
gorarakelyan: Yup, makes sense. What about assigning trial hashes (or ids) to run names and the sweep name to experiment? Like this:

| Run name  | Experiment |
| --- | --- |
| trial_1  | run_hparam_x  |
| trial_2  | run_hparam_x  |
| trial_3  | run_hparam_x  |
| --- | --- |
| trial_1  | run_hparam_y  |
| trial_2  | run_hparam_y  |
| trial_3  | run_hparam_y  |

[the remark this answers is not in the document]
muellerzr: Sounds good! I'll switch it to this as that makes perfect sense.

src/accelerate/tracking.py
```diff
@@ -320,7 +325,71 @@ def finish(self):
...
+        self.run_name = run_name
+        self.writer = Run(run_hash=run_name, repo=logging_dir, **kwargs)
```
gorarakelyan, line 344: @muellerzr The last minor thing - the `run_hash` is a unique auto-generated id of a run. It would be better to store the name as a `run.name`. Like this:
```py
self.writer = Run(repo=logging_dir, **kwargs)
self.writer.name = run_name
```

gorarakelyan:
@muellerzr the tracker looks awesome! Is there any example training script, which I can run to test out the integration? Thanks a million!

src/accelerate/tracking.py
muellerzr: Fixed, thanks!

muellerzr:
@gorarakelyan you can follow the directions here for installing the git version of accelerate + configuring it for your system:
https://huggingface.co/docs/accelerate/basic_tutorials/install

And then feel free to run the `nlp_example.py` script found in the examples directory here: https://github.com/huggingface/accelerate/blob/main/examples/nlp_example.py

Just make sure to also install `transformers` and `datasets` first :)

gorarakelyan:
@muellerzr thanks, will test it out soon! 🤗

src/accelerate/tracking.py
```diff
@@ -320,7 +325,72 @@ def finish(self):
...
+
+    def store_init_configuration(self, values: dict):
+        """
+        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment.
+
+        Args:
+            values (`dict`):
+                Values to be stored as initial hyperparameters as key-value pairs.
+        """
+        self.writer["hparams"].update(values)
```
gorarakelyan, line 363: Found one issue during testing. As Aim Run doesn't have an `update` method, let's just use a simple assignment here:
```py
self.writer["hparams"] = values
```

gorarakelyan (approved):
The rest seemed to work as expected!

src/accelerate/tracking.py
muellerzr: Makes sense, will update with this. Thanks for checking!

src/accelerate/tracking.py
```diff
@@ -320,7 +325,72 @@ def finish(self):
...
+
+    name = "aim"
+    requires_logging_directory = True
+
+    def __init__(self, run_name: str, logging_dir: Optional[Union[str, os.PathLike]] = ".", **kwargs):
+        self.run_name = run_name
+        self.writer = Run(repo=logging_dir, **kwargs)
+        self.writer.name = self.run_name
+        logger.info(f"Initialized Aim project {self.run_name}")
+        logger.info(
```
sgugger, line 347: Now that #655 was merged, this should be adapted.

sgugger (approved):
Thanks for adding support of this!

Issue #650, opened 2022-08-22
Set tracker log level 


nbroad1881:
# Background

I have set up my tracker as follows:

```python
from accelerate.logging import get_logger

logger = get_logger(__name__)

logging.basicConfig(
        format="%(asctime)s - %(levelname)s - %(name)s - %(message)s",
        datefmt="%m/%d/%Y %H:%M:%S",
        level=logging.WARNING,
    )

logger.setLevel(logging.WARNING)
```

This is followed by the `accelerator.init_trackers` step later in the script.

Even when setting the level to warning, I still get these messages logged:
`[2022-08-22 23:21:11,960][accelerate.tracking][INFO] - Successfully logged to WandB`

# Desired capability

Be able to set log level for the accelerate.tracking messages.

muellerzr:
Should be fixed now with https://github.com/huggingface/accelerate/pull/655

I don't believe it's easily possible to set the log level dynamically for those funcs, but instead we made them `debug` so it has to be opted in :)

PR #655, opened 2022-08-24
Use debug for loggers

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/650 by making the loggers use debug rather than info, to keep them quiet.


sgugger (approved):
LGTM, thanks!

PR #656, opened 2022-08-24
Small nits to grad accum docs

muellerzr:
Emphases more on how you should use either of the two API's we provide, and links to the grad synchronization tutorial at the end of the grad accum to advertise learning about the magic it wraps around

### DIFF
--- a/docs/source/concept_guides/gradient_synchronization.mdx
+++ b/docs/source/concept_guides/gradient_synchronization.mdx
@@ -116,4 +116,4 @@ for batch in dataloader:
         accelerator.backward(loss)
 ```
 
-As a result, you should either use *one or the other* when it comes to API choice. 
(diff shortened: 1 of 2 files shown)


docs/source/concept_guides/gradient_synchronization.mdx
```diff
@@ -114,4 +114,6 @@ for batch in dataloader:
         outputs = model(inputs)
         loss = loss_function(outputs, targets)
         accelerator.backward(loss)
-```
\ No newline at end of file
+```
+
+As a result, you should either use *one or the other* when it comes to API choice.
```
pacman100, line 119:
```suggestion
As a result, you should either use *`accelerator.accumulate` or `accelerator.no_sync`* when it comes to API choice. 
```

pacman100 (approved):
Thanks for this 😄! Left a nit for more clarity.

Issue #521, opened 2022-07-13
Save config to hparams.yaml for Tensorboard tracker


nateraw:
Right now, it seems `config` is ignored when you `accelerator.init_trackers(run, config)` for Tensorboard tracker. Would be nice if the config is saved as `hparams.yaml` so it shows up in Tensorboard.

sgugger:
Wdyt @muellerzr ?

muellerzr:
I think it makes sense to me, and still keeps in line with it being consistent throughout all the APIs for their respective trackers

Shreyz-max:
@muellerzr can I work on this?

muellerzr:
@Shreyz-max sure! 😄

Shreyz-max:
So, I did some research and found out that according to this [issue](https://github.com/tensorflow/tensorboard/issues/5476), the HParams dashboard currently requires at least one hparam and at least one metric defined in the hparams configuration but according to this line [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/tracking.py#L152) we are leaving the metric as empty. So, should I save the configs as a yaml file in the `tracking.py` file itself?

muellerzr:
@Shreyz-max that does seem like that'd be the way to do it then, so long as it can be shown in tensorboard correctly

Shreyz-max:
@muellerzr cool, thanks for the clarification. I'll complete it then.

Kyeongpil:
@nateraw 
I put config into init_trackers, but tensorboard couldn't show any configs.
Is there any reason why store_init_configuration uses os.path.join(self.logging_dir, str(project_run_name)) (where project_run_name is initialized using time.time()), not self.logging_dir?

PR #657, opened 2022-08-24
Saving hyperparams in yaml file for Tensorboard for #521

Shreyz-max:
This PR saves the hyperparameters in a YAML file for Tensorboard tracker. By default, the HParams dashboard currently requires at least one hparam and at least one metric defined in the hparams configuration but since we have no metric, the hyperparameters will be stored in a YAML file . The changes have been implemented in `tracking.py`.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -19,6 +19,7 @@ import os
 import time
 from abc import ABCMeta, abstractmethod, abstractproperty
 from typing import List, Optional, Union
+
 import yaml
 
 from .logging import get_logger
@@ -144,8 +145,8 @@ class TensorBoardTracker(GeneralTracker):
 
     def store_init_configuration(self, values: dict):
         """
-        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment.
-        Stores the hyperparameters in a yaml file for future use.
+        Logs `values` as hyperparameters for the run. Should be run at the beginning of your experiment. Stores the
+        hyperparameters in a yaml file for future use.
 
         Args:
             values (Dictionary `str` to `bool`, `str`, `float` or `int`):
@@ -155,11 +156,12 @@ class TensorBoardTracker(GeneralTracker):
         self.writer.add_hparams(values, metric_dict={})
         self.writer.flush()
         project_run_name = time.time()
-        with open(os.path.join(self.logging_dir, project_run_name,  "hparams.yml"), "w") as outfile:
+        with open(os.path.join(self.logging_dir, project_run_name, "hparams.yml"), "w") as outfile:
             try:
                 yaml.dump(values, outfile)
             except yaml.representer.RepresenterError:
-                logger.info("Serialization to store hyperparmeters failed")
+                logger.error("Serialization to store hyperparameters failed")
+                raise
         logger.info("Stored initial configuration hyperparameters to TensorBoard and hparams yaml file")
 
     def log(self, values: dict, step: Optional[int] = None, **kwargs):
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -153,7 +153,7 @@ class TensorBoardTracker(GeneralTracker):
         """
         self.writer.add_hparams(values, metric_dict={})
         self.writer.flush()
-        with open("hparams.yaml", "w") as outfile:
+        with open("self.logging_dir/hparams.yaml", "w") as outfile:
             yaml.dump(values, outfile)
         logger.info("Stored initial configuration hyperparameters to TensorBoard and hparams yaml file")
 
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -19,6 +19,8 @@ import os
 from abc import ABCMeta, abstractmethod, abstractproperty
 from typing import List, Optional, Union
 
+import yaml
+
 from .logging import get_logger
 from .utils import LoggerType, is_comet_ml_available, is_tensorboard_available, is_wandb_available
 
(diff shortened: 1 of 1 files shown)


src/accelerate/tracking.py
```diff
@@ -151,7 +153,9 @@ def store_init_configuration(self, values: dict):
         """
         self.writer.add_hparams(values, metric_dict={})
         self.writer.flush()
-        logger.info("Stored initial configuration hyperparameters to TensorBoard")
+        with open("hparams.yaml", "w") as outfile:
```
muellerzr, line 156: Perhaps it may be good to make use of the `logging_dir` to send the yaml to? Or should these be stored separately normally.

(This part):
```python
writer =  tensorboard.SummaryWriter(self.logging_dir, **kwargs)
```

muellerzr (approved):
Thank you for this! I left one comment just to make sure if we want to store it to the current dir or the already existing `logging_dir`

src/accelerate/tracking.py
Shreyz-max: So, should I save the yaml file in the `logging_dir`?

muellerzr: Yep, as this is where users would expect it to wind up landing. E.g

logging_dir/hparams.yaml

muellerzr:
cc @nateraw if you have any comments 🤗

src/accelerate/tracking.py
Shreyz-max: Yeah makes sense. I have made the changes now.

src/accelerate/tracking.py
```diff
@@ -151,7 +153,9 @@ def store_init_configuration(self, values: dict):
         """
         self.writer.add_hparams(values, metric_dict={})
         self.writer.flush()
-        logger.info("Stored initial configuration hyperparameters to TensorBoard")
+        with open("self.logging_dir/hparams.yaml", "w") as outfile:
```
nateraw, line 156: Would we want to use pathlib `Path(self.logging_dir) / 'hparams.yaml'` or `os.path.join(self.logging_dir, 'hparams.yaml')` instead here for sanity purposes?

src/accelerate/tracking.py
```diff
@@ -151,7 +153,9 @@ def store_init_configuration(self, values: dict):
...
+            yaml.dump(values, outfile)
```
nateraw, line 157: Do we need to check if all keys/vals are serializable here, or are we going to let the error yaml throws be the one we present to users in this case? (not sure if this is done elsewhere in the codebase)

nateraw (commented):
Couple comments here - will try to give it a try in an hour to give more detailed feedback if I have any.

src/accelerate/tracking.py
muellerzr: We do that somewhat here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/tracking.py#L170-L176

Maybe a try/catch that alerts the user that non-serializable bits were tried to be stored? Not sure what the try/catch error would be for that, otherwise I'm open to it just throwing a yaml error if we cannot :)

nateraw: could catch a `yaml.representer.RepresenterError` and return the message maybe?

nateraw: Along with more descriptive explanation of what happened under the hood accelerate...otherwise its the same as letting yaml handle it. 

I'm fine with either solution, just wanted to bring it up!

src/accelerate/tracking.py
nateraw: We should be careful here that `hparams.yaml` is saved separately for separate experiments. Should make sure that when you run 2 separate runs, you get 2 separate entries in the hparams dashboard of tensorboard.

src/accelerate/tracking.py
Shreyz-max: I have changed the hardcoded path here.

src/accelerate/tracking.py
Shreyz-max: Here, I think we can use`safe_dump` instead of  `dump` as it serialises only the basic tags and then we can add a try/catch with yaml error.

Shreyz-max: I have added the try/catch error.

Shreyz-max: I am not sure where to give a descriptive explanation. Should I add it as a comment or should I define it in the PR?

src/accelerate/tracking.py
Shreyz-max: The default tensorboard summary writer uses time.time() to create a folder for every run. So, I added the same for hparams.yml file. Hope this is okay.

src/accelerate/tracking.py
```diff
@@ -151,7 +154,13 @@ def store_init_configuration(self, values: dict):
         """
         self.writer.add_hparams(values, metric_dict={})
         self.writer.flush()
-        logger.info("Stored initial configuration hyperparameters to TensorBoard")
+        project_run_name = time.time()
+        with open(os.path.join(self.logging_dir, project_run_name,  "hparams.yml"), "w") as outfile:
+            try:
+                yaml.dump(values, outfile)
+            except yaml.representer.RepresenterError:
+                logger.info("Serialization to store hyperparmeters failed")
```
muellerzr, line 162: We should change this to actually raise the error, otherwise this will silently fail.

```suggestion
            except yaml.representer.RepresenterError:
                logger.error("Serialization to store hyperparameters failed")
                raise
```
I think just doing `logger.error` here is fine enough

muellerzr (changes requested):

src/accelerate/tracking.py
muellerzr: Doing so in the `logger.error` (see my suggestions) is good enough :)

src/accelerate/tracking.py
Shreyz-max: Yes, I have made the changes.

muellerzr:
@Shreyz-max cast `run_name` to a string on the yaml part and it should solve the error. E.g. `str(run_name)`

Great work on this, were almost there!

Shreyz-max:
Yeah, sorry about the mistake. I did not notice this at all.

Shreyz-max:
I suppose I need to create the folder.

muellerzr:
Great job! CC @nateraw for one last look, but lg2m!

muellerzr (approved):

Issue #526, opened 2022-07-17
Results on TPU worse than on GPU (using colab)


koba35:
### System Info

```Shell
- `Accelerate` version: 0.11.0.dev0
- Platform: Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic
- Python version: 3.7.13
- Numpy version: 1.21.6
- PyTorch version (GPU?): 1.11.0+cu102 (False)
- `Accelerate` default config:
	- compute_environment: LOCAL_MACHINE
	- distributed_type: TPU
	- mixed_precision: no
	- use_cpu: False
	- num_processes: 8
	- machine_rank: 0
	- num_machines: 1
	- main_process_ip: None
	- main_process_port: None
	- main_training_function: main
	- deepspeed_config: {}
	- fsdp_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

I created [notebook](https://colab.research.google.com/drive/1hOV5ebJOBpcYuh1-Ny0bahwK0YbZ0_0h?usp=sharing) for reproducing, but the steps are very easy:
1) Install all libs (I tried different versions, all share same behaivor)
2) Run accelerate config and choose TPU (or GPU with fp16)
3) Run accelerate launch accelerate/examples/nlp_example.py

When I choose training with TPU, I could get 0.848 f1 score. When I train with GPU, I got more than 0.9. I also tried different scripts, and always get much worse results with TPU. Maybe it something colab-specific, because as I can see in another TPU related issues [(for example)](https://github.com/huggingface/accelerate/pull/479) people getting results similar to my GPU results

### Expected behavior

```Shell
When I run example scripts in colab, I should get similar results with TPU and GPU.
```

koba35:
Yep, I tried adjust batch size and learning rates (multiply/divide by number) - but still no results, TPU in multi-processing perform much worse. I cannot achieve same performance (0.9+ f1), but I could achieve with single TPU now. 
BTW, this bug exist not only in colab - in TPU VM too

muellerzr:
For us to reproduce your results, please include the exact configurations you used for testing each one, including each batch size as you adjusted it as well as learning rates. These are critically important

koba35:
[...]

PR #665, opened 2022-08-29
Fix scheduler bug on TPUs

muellerzr:
TPUs should not be stepped n times like we do on GPUs. This will result in a *severe* performance degradation, and was the source of this issue https://github.com/huggingface/accelerate/issues/526

As a result, this PR adds a hard check if we `step_with_optimizer` to have it also perform the step if we are on the TPU.

Fixes https://github.com/huggingface/accelerate/issues/526

### DIFF
--- a/src/accelerate/scheduler.py
+++ b/src/accelerate/scheduler.py
@@ -17,6 +17,7 @@
 import warnings
 
 from .state import AcceleratorState
+from .utils import DistributedType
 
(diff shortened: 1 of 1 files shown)


sgugger:
As discussed offline, this is not the right solution IMO. There should be no differences for schedulers steps in multi-GPUs and TPUs.

PR #666, opened 2022-08-29
Improve docstrings and add examples where it makes sense

muellerzr:
This PR fixes a few nits in the docstrings (such as the Attributes issue discussed awhile ago in more places), as well as adds a few examples to places where users could get confused on the usage, or it is not extremely straightforward to use from the documentation. 

These specific cases are ones where numerous issues have been opened, asking how to use the specific item. 

This helps us document it in two places: the official documentation/tutorials for folks who need more information, and the docstring for folks who do not and just need the boilerplate.


sgugger (approved):
Great, thanks a lot for adding all of this!

PR #670, opened 2022-09-01
Cache torch_tpu check

muellerzr:
Mimics https://github.com/huggingface/transformers/pull/18777/ and caches the result of `is_tpu_available()`


sgugger (approved):
Nice!

PR #671, opened 2022-09-02
Manim animation of big model inference

muellerzr:
Adds an animation of big model inference directly to the docs, along with the code to create it. 

Animation is live [here](https://www.youtube.com/watch?v=MWCSGj9jEAo)


sgugger (approved):
Thanks for adding this!

PR #672, opened 2022-09-02
Fix multi-node issues from launch

muellerzr:
Solves a few lasting issues with the `torchrun` migration. Noteably:

- When doing multinode on the same network, it would just hang forever. 
- Defaults were not being set properly and overriden for `same_network`
- `node_rank` needs to be an `int`, not a string


sgugger (approved):
LGTM, thanks for the fixes!

PR #673, opened 2022-09-02
Make rich an optional dep

muellerzr:
Moves rich out from a mandatory dep to an optional dep, as discussed offline


sgugger (approved):
Thanks a lot!

PR #674, opened 2022-09-02
Specify local network on multinode

muellerzr:
Specifies that answering "no" means that the computers are on the cloud and/or on different network hosts

### DIFF
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -59,7 +59,7 @@ def get_cluster_input():
                 lambda x: int(x),
             )
             same_network = _ask_field(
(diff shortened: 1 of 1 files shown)


sgugger (approved):

src/accelerate/commands/config/cluster.py
```diff
@@ -59,7 +59,7 @@ def get_cluster_input():
                 lambda x: int(x),
             )
             same_network = _ask_field(
-                "Are all the machines on the same network? [YES/no]: ",
+                "Are all the machines on the same local network? [YES/no]: ",
```
pacman100, line 62:
```suggestion
                "Are all the machines on the same local network? Answer `no` if nodes are on the cloud and/or on the different network hosts [YES/no]: ",
```

pacman100 (approved):
Thanks 😄. A minor suggestion for further clarity.

PR #675, opened 2022-09-02
Mlflow tracker

nbroad1881:
I still have to do some tests, but I'm adding mlflow as a tracker.

Feedback welcome!


sgugger (commented):
LGTM! As you say, having some tests as well would be nice :-)

src/accelerate/tracking.py
```diff
@@ -398,9 +403,147 @@ def finish(self):
...
+            if len(str(value)) > self._MAX_PARAM_VAL_LENGTH:
+                logger.warning(
+                    f'Trainer is attempting to log a value of "{value}" for key "{name}" as a parameter. MLflow\'s'
+                    f" log_param() only accepts values no longer than {self._MAX_PARAM_VAL_LENGTH} characters so we dropped this attribute."
+                )
+                del values[name]
+
+        values_list = list(values.items())
+
+        # MLflow cannot log more than 100 values in one go, so we have to split it
```
nbroad1881, line 507: @sgugger @muellerzr , mlflow in AzureML cannot log more than 100 params (it's fine outside AzureML). Should I add anything to notify the user if it detects being in AzureML and the number of params > 100?

dks198: Hi @muellerzr,

How can we get through the error of having more than 100 params? Can we restrict the number of params we would like to log? I am stuck in this issue right now. Any help is greatly appreciated

Regards

muellerzr: Right below that you can see the for loop that was implemented

dks198: @muellerzr I am sorry if I am asking very question. I have seen the for loop however, how do I implement this into my azure notebook? Do I need to import the entire .py as a library and call the function? or just use the for loop within my notebook thats in Azure ML? Please help.

nbroad1881: @dks198 

I made this modified version that will save the training arguments to a json file. At the moment, there doesn't seem to be a way around the 100 param limit, so the params might need to be saved as artifacts instead.

https://gist.github.com/nbroad1881/dbd3083804bf3007de9b83a90dcf6428

dks198: @nbroad1881, thanks for sharing this. Can you please help me know how to use this in Azure ML Studio? Do I need to clone this entire notebook of yours to Azure ML?

nbroad1881: You can download the gist with 
```
wget https://gist.github.com/nbroad1881/dbd3083804bf3007de9b83a90dcf6428/raw/85a4ffc9c80b11fb3e9e8c86942b5b9e0101d98c/mlflow_tracker.py
```

Then you can import it by using 

`from mlflow_tracker import MLFlowTracker`

Then you create the tracker (filling in with your details)
```python
mlflow_tracker = MLflowTracker(
            experiment_name=,
            logging_dir=,
            run_id=,
            tags=,
            nested_run=,
            run_name=,
            description=,
        )
```

and pass it to the Accelerator
```python
Accelerator(log_with=[mlflow_tracker])
accelerator.init_trackers(project_name, param_dict)
```

src/accelerate/tracking.py
```diff
@@ -398,9 +410,148 @@ def finish(self):
...
+                    f" log_param() only accepts values no longer than {self._MAX_PARAM_VAL_LENGTH} characters so we dropped this attribute."
+                )
+                del values[name]
+
+        values_list = list(values.items())
+
+        if os.getenv("AML_CloudName") == "AzureCloud":
+            logger.info(
+                "It looks like you are using AzureML. Please note that there is a log limit of 100 total parameters when using AzureML."
+            )
```
muellerzr, line 513: It'd be nice if we could expand this conditional further to only log if the env and if the user is logging over 100 total parameters, and mention that they won't get dropped just split up upon multiple calls. Otherwise they'll get this every time they log something :) 

e.g.:

```suggestion
        if (os.getenv("AML_CloudName") == "AzureCloud") and (len(values_list) > self._MAX_PARAMS_TAGS_PER_BATCH):
            logger.info(
                f"More than {self._MAX_PARAMS_TAGS_PER_BATCH} passed when trying to log to AzureML, breaking up into multiple calls"
            )
```

muellerzr (commented):
Great work! I think just one more change and then we should be set (if we want to do tests later and you've seen it work on your system)

src/accelerate/tracking.py
nbroad1881: I took it out because it can now log 200 params so it isn't likely to be an issue anymore.

muellerzr: Awesome, thanks for your work! 🤗

muellerzr:
Thanks @nbroad1881, last bit is could we fix up the conflicts? Then we're good to go!

nbroad1881:
@muellerzr, it's ready for final review. It only took me 2 months 
# 😬

sgugger:
@nbroad1881 You might need to close this PR and open a fresh one as GitHub shows a diff of 68 files (probably due to a rebase).

nbroad1881:
$#@#! ok I didn't realize rebasing would do that

PR #682, opened 2022-09-06
Fix skip in dispatch dataloaders

sgugger:
This PR fixes the dead loop reported in #678. The `skip` part is some old code that wasn't updated in the last refactor of this dataloader, so should be removed altogether as it will create infinite loops as is.

PR #686, opened 2022-09-07
Clean up DispatchDataloader a bit more

sgugger:
As pointed out in #682 after its merge, there is a bit more code that can be cleaned up.


pacman100 (approved):
Thank you! 🤗

PR #688, opened 2022-09-09
rng state sync for FSDP

pacman100:
### What does this PR do?
1. `synchronize_rng_state` now is cognizant of FSDP.


sgugger (approved):
Good catch! Thanks for fixing!

PR #698, opened 2022-09-14
override DeepSpeed `grad_acc_steps` from `accelerator` obj

pacman100:
### What does this PR do?
1. Override DeepSpeed `grad_acc_steps` from `accelerator` obj when they aren't same.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -747,6 +747,12 @@ class Accelerator:
             batch_size_per_device = deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"]
             result = [obj for obj in args]
 
+        if self.gradient_accumulation_steps != deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"]:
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -747,6 +747,12 @@ def _prepare_deepspeed(self, *args):
             batch_size_per_device = deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"]
             result = [obj for obj in args]
 
+        if self.gradient_accumulation_steps != deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"]:
+            warnings.warn(
```
sgugger, line 751: No need for a warning here, I think a logging info is sufficient (the fact that args to the `Accelerator` supercede the config is a common behavior, you can do it to force CPU or a given mixed precision already).

sgugger (approved):
Thanks for adding this!

src/accelerate/accelerator.py

PR #703, opened 2022-09-19
Add in report generation for test failures

muellerzr:
Generates a report of failures when they occur on the job summary page. 

See an example here: https://github.com/huggingface/accelerate/actions/runs/3085436367/attempts/1#summary-8436903677


sgugger (approved):
Looking, good! Thanks for adding this!

PR #704, opened 2022-09-20
Update runners with report structure, adjust env variable

muellerzr:
Noticed that the env variable was automatically being set (which I didn't want), so it's now set to `IS_GITHUB_CI`. Also trickles in the reportlog to the other runners


sgugger (approved):
Thanks!

PR #705, opened 2022-09-21
torch 1.12.1 does not work with intel ccl 1.12.0

sywangyi:


sywangyi:
@sgugger @yao-matrix same with the PR with transformer

sgugger:
Same comment as in https://github.com/huggingface/transformers/pull/19132
Those kinds of checks belong in your libraries directly, not Accelerate or Transformers as we can't constantly add/remove checks each time there is a new release (also PyTorch 1.12.1 was out a month and a half ago...)

sywangyi:
@sgugger we will release a new oneCCL 1.12.1 to work with torch 1.12.1

PR #713, opened 2022-09-21
refactor(accelerate): readability improvements

ryanrussell:
Signed-off-by: Ryan Russell <git@ryanrussell.org>

### Description
One-off readability fix-ups in `src/accelerate/*.py`

Grew to a few one-off's when grep'ing to check for more occurrences... 

Seemed reasonable to put these together in a PR. 

Fixes:
- `devicement_placement` -> `device_placement`
- `intialize` -> `initialize`
- `pyton` -> `python`
- `avoig` -> `avoid`

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -125,8 +125,8 @@ class Accelerator:
             - `"tensorboard"`
             - `"wandb"`
             - `"comet_ml"`
-            If `"all"` is selected, will pick up all available trackers in the environment and initialize them. Can also
-            accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
+            If `"all"` is selected, will pick up all available trackers in the environment and initialize them. Can
+            also accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
         logging_dir (`str`, `os.PathLike`, *optional*):
             A path to a directory for storing logs of locally-compatible loggers.
         dispatch_batches (`bool`, *optional*):
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -125,7 +125,7 @@ class Accelerator:
             - `"tensorboard"`
             - `"wandb"`
             - `"comet_ml"`
-            If `"all`" is selected, will pick up all available trackers in the environment and intialize them. Can also
+            If `"all`" is selected, will pick up all available trackers in the environment and initialize them. Can also
             accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
         logging_dir (`str`, `os.PathLike`, *optional*):
             A path to a directory for storing logs of locally-compatible loggers.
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Thanks again! Just make sure to run `make style` on your branch to fix the quality error.

ryanrussell:
> Thanks again! Just make sure to run `make style` on your branch to fix the quality error.

Will do... interestingly enough the `Makefile` won't run for me! On `Ubuntu 20.04` .. `python` isn't a command... it's `python3`...

Let me open up a container or symlink and touch it up.

ryanrussell:
Hey @sgugger ,

Trying to run your Makefile has proven really tough for me:
1. `python` doesn't exist on ubuntu. Setup a symlink for `python` ...
2. `python3.9` running `make style` tries to change  30 files
`30 files reformatted, 55 files left unchanged.`
3. I did a `docker run ... 3.7.... -v .:/accelerate` ...
This still gave different results than the CI runner:
`5 files reformatted, 80 files left unchanged.`

So, after all this, I'm unable to actually get my black/isort to match what the test unhappy about.

It appears the test is failing for a line length style issue from a file that wasn't actually touched in this commit...

```
python utils/style_doc.py src/accelerate docs/source --max_len 119 --check_only
  File "utils/style_doc.py", line 556, in <module>
    main(*args.files, max_len=args.max_len, check_only=args.check_only)
  File "utils/style_doc.py", line 544, in main
    raise ValueError(f"{len(changed)} files should be restyled!")
```

Can we just merge this in, and fix that style next time you lint?

Is there a docker image or something setup to run your `Makefile` in a reproducible environment? I'm probably missing something...

ryanrussell:
Hey @sgugger  --

This is jumping the shark, but I do have a (arguably) nice pre-commit template that handles autoflake, isort, black and works well cross-platform.

Here's a screenshot. If you want to re-do the makefile, I can start a new ticket and switch it to pre-commit.

image

This also checks for conventional commits format(optional).

This would be a bit bigger in scope and more opinionated than what I thought I was starting with some readability PR's :)

Let me know if this is something that's of interest to swapping out for the huggingface team... if so, we can throw it in a new issue.

muellerzr:
Hi @ryanrussell, I think the issue here is the wrong black/flake8 may be installed? Can you try running `make style` again after doing `pip install -e .[quality]` (or `pip install accelerate[quality]` depending on how you have it installed)?

sgugger:
Hi @ryanrussell. 90% on the team is on Ubuntu and have no problem with running `make style` without doing any aliases. Not sure why you don't have the `python` command available after installing Python.

Then as @muellerzr said, you need to have the pinned versions we use for the libraries formatting the code, which is why we recommend using a virtual environment. I've pushed a style commit to move forward with this PR.

I don't want to add pre-commit hooks as I prefer having the ability to do the styling in a separate commit when I'm afraid it might break something.

ryanrussell:
Hey @sgugger -- thanks for helping fix this. Sounds like it's a user error on my side.

I'll wait to send any more PR's until I setup a venv with correct version pins and can replicate the expected linting from the CI...

Just found the discord, so joining there... if I run into any issues, much easier to iterate there than through comments :+1:

sgugger:
Failures are unrelated (something in the dev of evaluate messed up) so merging!

PR #715, opened 2022-09-23
More docstring nits

muellerzr:
Adds some more type hints and docs to functions I've noticed the one-off user has used before so it's documented (but also documents the best practice). 

Also felt like having the type annotations on `unwrap_model` eases a bit more into what exactly it *can* unwrap, let me know if we should use the fully qualified types (I imagine so) and how that would be annotated to fit in the 150ish char limit we have (e.g. newline for each type?)

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -674,7 +674,15 @@ class Accelerator:
 
         return result if len(result) > 1 else result[0]
 
-    def prepare_model(self, model):
+    def prepare_model(self, model: torch.nn.Module):
+        """
+        Prepares a PyTorch model for training in any distributed setup. It is recommended to use
+        [`Accelerator.prepare`] instead.
+
+        Args:
+            model (`torch.nn.Module`):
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1383,6 +1416,15 @@ def _get_devices(self, *args):
         return (model_device, optimizer_device)
 
     def get_state_dict(self, model, unwrap=True):
+        """
+        Returns the state dictionary of a model sent through [`Accelerator.prepare`] in full precision
+
+        Args:
+            model (`Module`, `DistributedDataParallel`, `FullyShardedDataParallel`, `MpModelWrapper`):
```
sgugger, line 1423: I think the other types are all subclasses of `Module`
```suggestion
            model (`torch.nn.Module`):
```

sgugger (approved):
Thanks for the nits! I don't think we need to go further than `torch.nn.Module` for the types of models being unwrapped as all the wrappers are themselves modules.

PR #716, opened 2022-09-23
Allow custom device placements for different objects

### LINKED ISSUE #624: Let `device_placement` parameter can be overridden for different type object
I'd like that model(with `Optimizer`) and `DataLoader` use different `device_placement` value to be possible. i.e., I want to let accelerate deal with device placement for model, but not for dataloader.

Use case: I have a complicated dataloading process(e.g. use DALI), so I'd like to let my `DataLoader` act like a `Sampler`, i.e. only provide indices. If I set `device_placement=True`, then indices provided by DataLoader will be send to device and I need to send it back to cpu for my dataloading process. If I set `device_placement=False`, I need to deal with device placement for model and optimizer by myself.

It is true that I can use identity function instead of `default_collate` as collate_fn of my `DataLoader`, so that indices as list of int without `to` method will be ignored by  `send_to_device`. But I think it is more explicit and elegant to make DataLoader don't deal with device placement, than to make data "incompatible" with `send_to_device`.
In addition, this behavior is not well documented: only the doc string of `prepare_data_loader` says "`put_on_device` only works if the batches are nested list, tuples or dictionaries of tensors"(and actually works for objects with `to` method instead of only tensors), which is vague too - what will happen if it doesn't work: Raising an error, ignoring whole container or ignoring one element? The Args Spec of `send_to_device` says data should be nested list/tuple/dictionary of `torch.Tensor`, users might expect an error will be raise if input doesn't satisfy the Spec.

It is also true that I can manually set `DataLoaderShard.device = None` to achieve what I want. But that depends on the implementation, not the interface.

sgugger:
This PR lets the user customize whether device_placement should be `True` or `False` for the different objects passed along to prepare. This feature was requested in #624

Fixes #624

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -616,8 +616,8 @@ class Accelerator:
                 - `torch.optim.Optimizer`: PyTorch Optimizer
                 - `torch.optim.lr_scheduler._LRScheduler`: PyTorch LR Scheduler
             device_placement (`List[bool]`, *optional*):
-                Used to customize whether automatic device placement should be performed for each object passed. Needs to be a list of the same length as
-                `args`.
+                Used to customize whether automatic device placement should be performed for each object passed. Needs
+                to be a list of the same length as `args`.
         """
         if device_placement is None:
             device_placement = [None for _ in args]
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -547,15 +547,15 @@ class Accelerator:
         if self.is_local_main_process:
             print(*args, **kwargs)
 
-    def _prepare_one(self, obj, first_pass=False):
+    def _prepare_one(self, obj, first_pass=False, device_placement=None):
         # First pass of preparation: DataLoader, model, optimizer
         if first_pass:
             if isinstance(obj, torch.utils.data.DataLoader):
-                return self.prepare_data_loader(obj)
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -602,17 +602,32 @@ def _prepare_fsdp(self, *args):
...
+            *args:
+                Any of the following type of objects:
+
+                - `torch.utils.data.DataLoader`: PyTorch Dataloader
+                - `torch.nn.Module`: PyTorch Module
+                - `torch.optim.Optimizer`: PyTorch Optimizer
+                - `torch.optim.lr_scheduler._LRScheduler`: PyTorch LR Scheduler
+            device_placement (`List[bool]`, *optional*):
+                Used to customize the device placement for each object passed. Needs to be a list of the same length as
+                `args`.
```
muellerzr, line 620: Perhaps we can be a bit more specific here:

```suggestion
            device_placement (`List[bool]`, *optional*):
                Used to customize whether automatic device placement should be performed for each object passed. Needs to be a list of the same length as
                `args`.
```

muellerzr (approved):
Looks great! Thanks for this. One suggestion on the docstring for clarity but otherwise awesome!

pacman100 (approved):
Thank you, LGTM! 🤗

PR #722, opened 2022-09-26
Specify gradients in model preparation

muellerzr:
Adjusts the docstring to specify that only models whose gradients will be utilized should be passed through accelerator.prepare


sgugger (commented):
That's actually not accurate. The model is also prepared to handle mixed precision training in the `prepare_model` method, so any model should be sent to the `prepare` call ideally.

sgugger (approved):
Thanks!

Issue #723, opened 2022-09-27
'GradientState' object has no attribute '_iterate_samples_seen'


pokameng:
'GradientState' object has no attribute '_iterate_samples_seen'
The class  GradientState  does not have this function  '_iterate_samples_seen'

muellerzr:
@pokameng please include a stack trace along with what you did to get there, everything else the initial prompt asked for. We can do absolutely nothing without that to help :)

pokameng:
@muellerzr  hello sir,
I intend to train a model using the diffusion model, but when I run the code as instructed, I am prompted with the following error.
AttributeError: 'GradientState' object has no attribute '_iterate_samples_seen'

The error can be found in accelerate/data_loader.py on line 350

My code is as follows:
from denoising_diffusion_pytorch import Unet, GaussianDiffusion, Trainer



img_path='/opt/data/share/106004/dataset/CelebA/celebA_hq/train_512'
model = Unet(
    dim = 64,
    dim_mults = (1, 2, 4, 8)
).cuda()

diffusion = GaussianDiffusion(
    model,
    image_size = 128,
    timesteps = 1000,           # number of steps
    sampling_timesteps = 250,   # number of sampling timesteps (using ddim for faster inference [see citation for ddim paper])
    loss_type = 'l1'            # L1 or L2
).cuda()

trainer = Trainer(
    diffusion,
    img_path,
    train_batch_size = 32,
    train_lr = 8e-5,
    train_num_steps = 700000,         # total training steps
    gradient_accumulate_every = 2,    # gradient accumulation steps
    ema_decay = 0.995,                # exponential moving average decay
    amp = True                        # turn on mixed precision
)

trainer.train()

muellerzr:
@pokameng can you try again, installing Accelerate with:

```bash
pip install git+https://github.com/huggingface/accelerate@iterate-samples-seen
```

pokameng:
ok  thanks
this problem is fixed

muellerzr:
Great! Thanks! For the time being please install accelerate via dev until the next release with the bugfix :) 

`pip install git+https://github.com/huggingface/accelerate`

PR #724, opened 2022-09-27
Fix regression issue on iterate_samples_seen

muellerzr:
There was a lasting `_iterate_samples_seen` that was unnoticed until https://github.com/huggingface/accelerate/issues/723


sgugger (approved):
Thanks for fixing!

PR #725, opened 2022-09-27
Build and Release docker images on a release

muellerzr:
Introduces a workflow that releases a docker image based on when a Github release has been made. It will pull the version id from the setup.py and that state

Relevant gh docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

Tested this on a separate branch using the current dev version. 

CPU tag can be seen here: https://hub.docker.com/r/huggingface/accelerate-cpu/tags
GPU tag can be seen here: https://hub.docker.com/r/huggingface/accelerate-gpu/tags

FFT, should we just merge these two release workflows together and check if 'dev' is in the version to automatically tag the `latest` tag? Or maybe just publish the prerelease tags? WDYT?

### DIFF
--- a/.github/workflows/build-docker-images-release.yml
+++ b/.github/workflows/build-docker-images-release.yml
@@ -1,4 +1,4 @@
-name: Build Docker images (scheduled)
+name: Build Docker images (releases)
 
 on:
   workflow_dispatch:
--- a/.github/workflows/build-docker-images-release.yml
+++ b/.github/workflows/build-docker-images-release.yml
@@ -1,0 +1,62 @@
+name: Build Docker images (scheduled)
+
+on:
+  workflow_dispatch:
+  release:
+
+concurrency:
+  group: docker-image-builds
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for adding this!

ydshieh (approved):
Thanks, @muellerzr . Are these images to be used by other people, or somewhere in accelerate testing?

ydshieh:
> FFT, should we just merge these two release workflows together and check if 'dev' is in the version to automatically tag the `latest` tag? Or maybe just publish the prerelease tags? WDYT?

I guess you are talking about the existing `build_docker_images.yml` here. I don't have particular opinion about if merging these 2 workflow files or not.

muellerzr:
@ydshieh by other folks, not for testing. We've had people use the latest images already, so made sense to also do releases 

Okay, will think on this and do a follow up most likely with the merge

Issue #717, opened 2022-09-24
Getting num_processes = 8 even though num_processes = 1 in accelerate env when the default port is busy


Hritikbansal:
### System Info

```Shell
- `Accelerate` version: 0.12.0
- Platform: Linux-5.4.0-107-generic-x86_64-with-glibc2.31
- Python version: 3.9.0
- Numpy version: 1.23.1
- PyTorch version (GPU?): 1.12.1 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: fp16
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - downcast_bf16: no
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

I decided to change my master port as the default 26500 port was already occupied. 

Running `CUDA_VISIBLE_DEVICE="6" accelerate launch --main_process_port 5432  <my_script>.py` returned:
```
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to `8`
        `--num_cpu_threads_per_process` was set to `32` to improve out-of-box performance
```

 Running `CUDA_VISIBLE_DEVICE="6" accelerate launch <my_script>.py` also returns the same thing:
```
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to `8`
        `--num_cpu_threads_per_process` was set to `32` to improve out-of-box performance
```
in addition to:

```
RuntimeError: The server socket has failed to listen on any local network address. The server socket has failed to bind to [::]:29500 (errno: 98 - Address already in use). The server socket has failed to bind to ?UNKNOWN? (errno: 98 - Address already in use).
```

This should not have happened 

### Expected behavior

```Shell
When my default 26500 port is not busy, then I code works fine as expected. I.e., it runs on just one process. 

The expected behavior is that this code works seamlessly on one process only, when the default master port is busy and a new master port has been used.
```

sgugger:
Thanks for the report!
[...]

PR #726, opened 2022-09-27
Fix default for num processes

sgugger:
As pointed out in #717, the current default for `num_processes` erases the value set by the config, which we don't want. This PR fixes that.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -845,7 +845,7 @@ def launch_command(args):
     else:
         if args.num_processes is None:
             warned.append("\t`--num_processes` was set to a value of `1`")
-            args.num_processes = 1
+            args.num_processes = torch.cuda.device_count() if args.multi_gpu else 1
         if args.num_machines is None:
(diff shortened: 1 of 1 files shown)


src/accelerate/commands/launch.py
```diff
@@ -845,7 +845,7 @@ def launch_command(args):
     else:
         if args.num_processes is None:
             warned.append("\t`--num_processes` was set to a value of `1`")
```
muellerzr, line 847:
```suggestion
            args.num_processes = torch.cuda.device_count() if args.multi_gpu else 1
            warned.append("\t`--num_processes` was set to a value of `{args.num_processes}`")
```

src/accelerate/commands/launch.py
```diff
@@ -845,7 +845,7 @@ def launch_command(args):
...
-            args.num_processes = 1
+            args.num_processes = torch.cuda.device_count() if args.multi_gpu else 1
```
muellerzr, line 848:
```suggestion
```

Issue #690, opened 2022-09-09
underlying lr_scheduler.step() never called with OneCycleLR in single GPU


miccio-dk:
### System Info

```Shell
Accelerate 0.10.0
Ubuntu 20.04.4 LTS
Python 3.9.12
Numpy 1.19.5
Pytorch 1.10.0
Hydra 1.2.0

compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: MULTI_GPU
fsdp_config: {}
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
mixed_precision: fp16
num_machines: 1
num_processes: 1
use_cpu: false
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

In my tests, the step function of the scheduler is never called inside the [accelerated scheduler wrapper](https://github.com/huggingface/accelerate/blob/v0.12.0/src/accelerate/scheduler.py#L53).
This appears to have been introduced by this [fix](https://github.com/huggingface/accelerate/pull/385).

Example code:
```py
data_loader = ...
model = ...
optimizer = ...
lr_scheduler = torch.optim.lr_scheduler.OneCycleLR(optimizer, max_lr=0.01, steps_per_epoch=len(data_loader), epochs=n_epochs)

data_loader, model, optimizer, lr_scheduler = accelerator.prepare(data_loader, model, optimizer, lr_scheduler)

for epoch in range(n_epochs):
    for batch in data_loader:
        optimizer.zero_grad()
        inputs, targets = batch
        outputs = model(inputs)
        loss = loss_function(outputs, targets)
        accelerator.backward(loss)
        optimizer.step()
        lr_scheduler.step()
```

### Expected behavior

```Shell
I expect `OneCycleLR.step()` to be called at some point inside `AcceleratedScheduler.step()`, supposedly on the [last line](https://github.com/huggingface/accelerate/blob/v0.12.0/src/accelerate/scheduler.py#L73).
```

sgugger:
You'll have to wait for @muellerzr to be back from holiday for this :-)

miccio-dk:
Hey, did you get the chance to look into this? :)

muellerzr:
[...]

PR #728, opened 2022-09-28
Fix issue with one-cycle logic

muellerzr:
Turns out the logic for one-cycle on the scheduler was backwards so `step` was never called.

Fixes https://github.com/huggingface/accelerate/issues/690 and adds some tests


sgugger (approved):
Thanks for adding the tests!

Issue #543, opened 2022-07-21
Feature request: Specifying GPU IDs


tmabraham:
It would be ideal to specify the GPU IDs that a script can use. i typically like to have a GPU ID argument in my script so that I can set the device, but if Accelerate is handling the devices I can't specify. The current alternative is to use `CUDA_VISIBLE_DEVICES` but a dedicated argument in `accelerate config` or in the `Accelerator` object would be ideal.

tanmoyio:
@muellerzr I want to work on it

skfirebox:
how i can change GPU 0 to other gpu? @muellerzr 
i am using Accelerate,
error is below given, please give the solution in details
[/home/suresh/myenv/lib/python3.8/site-packages/accelerate/accelerator.py:391](https://vscode-remote+ssh-002dremote-002b10-002e64-002e31-002e80.vscode-resource.vscode-cdn.net/home/suresh/myenv/lib/python3.8/site-packages/accelerate/accelerator.py:391): UserWarning: `log_with=tensorboard` was passed but no supported trackers are currently installed.
  warnings.warn(f"`log_with={log_with}` was passed but no supported trackers are currently installed.")

{'clip_sample_range', 'timestep_spacing', 'trained_betas', 'sample_max_value', 'variance_type', 'dynamic_thresholding_ratio'} was not found in config. Values will be initialized to default values.
{'num_attention_heads', 'attention_type', 'addition_time_embed_dim', 'dropout', 'reverse_transformer_layers_per_block', 'transformer_layers_per_block'} was not found in config. Values will be initialized to default values.
accelerator.device =  cuda
accelerator.device =  cuda
Traceback (most recent call last):
  File "/home/suresh/lora/diffusers/examples/kandinsky2_2_train/tune_decoder_lora.py", line 589, in <module>
    main()
  File "/home/suresh/lora/diffusers/examples/kandinsky2_2_train/tune_decoder_lora.py", line 362, in main
    image_encoder.to(accelerator.device)
  File "/home/suresh/myenv/lib/python3.8/site-packages/transformers/modeling_utils.py", line 2576, in to
    return super().to(*args, **kwargs)
  File "/home/suresh/myenv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1160, in to
    return self._apply(convert)
  File "/home/suresh/myenv/lib/python3.8/site-packages/torch/nn/modules/module.py", line 810, in _apply
    module._apply(fn)
[...]

PR #732, opened 2022-09-30
Allow for GPU-ID specification on CLI

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/543 by adding a `--gpu_ids` param to both the CLI as well as config, and has the CLI launcher utilize it as a result. This essentially just sets `CUDA_VISIBLE_DEVICES` dynamically before launching code

If need for it to occur in `notebook_launcher` is desired, I'll expand this further

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -262,7 +262,7 @@ def launch_command_parser(subparsers=None):
     parser.add_argument(
         "--gpu_ids",
         default="all",
-        help="What GPUs (by id) should be used for training as a comma-seperated list",
+        help="What GPUs (by id) should be used for training on this machine as a comma-seperated list",
     )
     parser.add_argument(
         "--machine_rank", type=int, default=None, help="The rank of the machine on which this script is launched."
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Looks nice! Thanks for adding this!

pacman100 (commented):
Hello, nice addition. However, I don't think `CUDA_VISIBLE_DEVICES` would do anything for Apple Silicon GPUs.

muellerzr:
@pacman100 correct, this is solely for NVIDIA GPUs, I'll specify that in the info description for it

src/accelerate/commands/launch.py
```diff
@@ -817,6 +830,13 @@ def launch_command(args):
             args.tpu = defaults.distributed_type == DistributedType.TPU
             args.use_fsdp = defaults.distributed_type == DistributedType.FSDP
             args.use_mps_device = defaults.distributed_type == DistributedType.MPS
+        if args.gpu_ids is None:
+            if defaults.gpu_ids is not None:
+                args.gpu_ids = defaults.gpu_ids
+            else:
+                args.gpu_ids = "all"
+        if len(args.gpu_ids.split(",")) < 2 and args.multi_gpu and (args.gpu_ids != "all"):
+            args.multi_gpu = False
```
pacman100, line 839: I meant with respect to these changes

muellerzr: Let me know what you think about my solution here: https://github.com/huggingface/accelerate/pull/732/commits/0467d4ff0bacda7ffb0d6087a7ef636cfbbb4960

pacman100: Nice! This should have expected behaviour. 🤗

PR #735, opened 2022-10-03
Remove auto-bug label in issue template

sgugger:
Same as in Transformers, it irks em to see the bug label auto-added whenever a user files an issue, since lots of them are just questions or issues within the user's code and not a bug in the library.


muellerzr (approved):
Thanks! Makes sense

PR #737, opened 2022-10-03
Add an example zoo to the documentation

muellerzr:
Adds a non-exhaustive example zoo to the docs, which also includes a few examples from popular integrations


sgugger (approved):
Super, thanks a lot for doing this!

PR #739, opened 2022-10-04
Fix memory leak

muellerzr:
Fixes a memory leak in the memory example. The `accelerator` should be the only one defined out of the loop, and `accelerator.free_memory()` should be the first thing called in the loop. This way all memory is actually freed and lingering references can be wiped.


sgugger (approved):
Makes sense, thanks for fixing!

PR #514, opened 2022-07-13
fixing deepspeed multi-node launcher

pacman100:
### What does this PR do?
Fixes issue #498 

1. Fixes deepspeed multinode launcher support. It uses `hostfile` as specified in the Official DeepSpeed documentation [Resource Configuration (multi-node)](https://www.deepspeed.ai/getting-started/#resource-configuration-multi-node)
```
DeepSpeed configures multi-node compute resources with hostfiles that
are compatible with [OpenMPI](https://www.open-mpi.org/) and [Horovod](https://github.com/horovod/horovod). 
A hostfile is a list of hostnames (or SSH aliases), which are machines accessible via passwordless SSH,
and slot counts, which specify the number of GPUs available on the system
```
An example of a `hostfile` is given below:
```
localhost slots=2
brutasse slots=2
```
2. Example of the accelerate env after answering `accelerate config` questionnaire:
```
- `Accelerate` version: 0.11.0.dev0
- Platform: Linux-5.4.0-121-generic-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.23.0
- PyTorch version (GPU?): 1.12.0+cu102 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - mixed_precision: fp16
        - use_cpu: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 2
        - main_process_ip: 
        - main_process_port: 8888
        - main_training_function: main
        - deepspeed_config: {
                 'deepspeed_hostfile': '/home/sourab/hostfile', 
                 'deepspeed_multinode_launcher': 'pdsh', 
                 'gradient_accumulation_steps': 1, 
                 'gradient_clipping': 1.0, 
                 'offload_optimizer_device': 'none', 
                 'offload_param_device': 'none', 
                 'zero3_init_flag': False, 
                 'zero_stage': 2
        }
        - fsdp_config: {}
``` 

3. Sample output when running modified `nlp_complete_example.py` through command `accelerate launch nlp_complete_example.py`.
```
[2022-07-13 09:11:42,881] [INFO] [runner.py:378:main] Using IP address of 192.XXX.X.XX for node localhost                      
[2022-07-13 09:11:42,882] [INFO] [multinode_runner.py:65:get_cmd] Running on the following workers: localhost,brutasse         
[...]

PR #741, opened 2022-10-04
[ds launcher] un-hijack PYTHONPATH

stas00:
I'm not sure why `PYTHONPATH` is set to python executable here as it expects dirs and not files:
https://github.com/huggingface/accelerate/pull/514/files#r987305929

but regardless accelerate shouldn't hijack `PYTHONPATH` 

This PR extends the `PYTHONPATH` instead if it was preset.

Thank you

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -574,7 +574,7 @@ def deepspeed_launcher(args):
         warnings.warn('--fp16 flag is deprecated. Use "--mixed_precision fp16" instead.', DeprecationWarning)
         mixed_precision = "fp16"
 
-    current_env["PYTHONPATH"] = env_var_path_add("PYTHONPATH", sys.executable)
+    current_env["PYTHONPATH"] = env_var_path_add("PYTHONPATH", os.path.abspath('.'))
     current_env["MIXED_PRECISION"] = str(mixed_precision)
     current_env["USE_DEEPSPEED"] = "true"
     current_env["DEEPSPEED_ZERO_STAGE"] = str(args.zero_stage)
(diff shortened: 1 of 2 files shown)


src/accelerate/commands/launch.py
```diff
@@ -573,7 +573,13 @@ def deepspeed_launcher(args):
...
         mixed_precision = "fp16"
 
-    current_env["PYTHONPATH"] = sys.executable
+    def env_path_add(env_var_name, new_path):
+        """extend path-based env variable with a new path"""
+        paths = [p for p in os.environ.get(env_var_name, "").split(":") if len(p) > 0]
+        paths.append(str(new_path))
+        return ":".join(paths)
+
+    current_env["PYTHONPATH"] = env_path_add("PYTHONPATH", sys.executable)
```
sgugger, line 582: Let's just write the three lines of code instead of adding a nested function? It will be shorter and as easy to read.

sgugger (approved):
Thanks for your PR! LGTM but let's wait to see what @pacman100 thinks.

src/accelerate/commands/launch.py
stas00: as it's likely that at some point you might want to extend other path-based env vars like `PATH` or `LD_LIBRARY_PATH` - one more option is to move it to utils. 

but if you prefer inlined and @pacman100 is an agreement I will rewrite.

sgugger: I'm fine having it as a util too, just don't like the nested function approach where it can't be reused.

stas00: pushed the change to move to utils, renamed the helper for better clarity now that it's far away from definition.

stas00:
@pacman100, please feel free to merge when you feel it's ready. Thanks.

PR #742, opened 2022-10-05
Auto grad accum example

muellerzr:
Showcases how to perform automatic gradient accumulation, similar to a popular feature that was toted by MosaicML awhile back.

The initial batch size starts at 256 to guarantee a CUDA OOM and the user can observe the print statement showing the new number of steps and batch size

### DIFF
--- a/examples/by_feature/automatic_gradient_accumulation.py
+++ b/examples/by_feature/automatic_gradient_accumulation.py
@@ -1,0 +1,232 @@
+# Copyright 2022 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


examples/by_feature/automatic_gradient_accumulation.py
```diff
@@ -0,0 +1,232 @@
...
+    num_epochs = int(config["num_epochs"])
+    seed = int(config["seed"])
+    observed_batch_size = int(config["batch_size"])
+
+    metric = evaluate.load("glue", "mrpc")
+
+    # New Code #
+    # We use the `find_executable_batch_size` decorator, passing in the desired observed batch size
+    # to train on. If a CUDA OOM error occurs, it will retry this loop cutting the batch size in
+    # half each time. From this, we can calcualte the number of gradient accumulation steps needed
```
sgugger, line 127:
```suggestion
    # half each time. From this, we can calculate the number of gradient accumulation steps needed
```

sgugger (approved):
LGTM, thanks a lot!

PR #747, opened 2022-10-07
[Device map] nn.Parameter don't have children

patrickvonplaten:
This is needed to allow to use the "device_map" for models that have `nn.Parameter` at the top level, such as the safety checker of `diffusers`: https://github.com/huggingface/diffusers/pull/772/files#r990269773

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -258,7 +258,7 @@ def get_max_layer_size(
     modules_to_treat = modules.copy()
     while len(modules_to_treat) > 0:
         module_name, module = modules_to_treat.pop(0)
-        modules_children = list(module.named_children())
+        modules_children = list(module.named_children()) if not isinstance(module, torch.nn.Parameter) else []
         if len(modules_children) == 0 or module.__class__.__name__ in no_split_module_classes:
             # No splitting this one so we compare to the max_size
             size = module_sizes[module_name]
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -258,7 +258,7 @@ def get_max_layer_size(
     modules_to_treat = modules.copy()
     while len(modules_to_treat) > 0:
         module_name, module = modules_to_treat.pop(0)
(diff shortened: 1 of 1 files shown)


sgugger (commented):
What's a reproducer? I'm not sure how you get to a parameter here since we only apply the `children` method.

patrickvonplaten:
> What's a reproducer? I'm not sure how you get to a parameter here since we only apply the `children` method.

Haha that was fast. Here a repro:

```
from diffusers.pipelines.stable_diffusion import StableDiffusionSafetyChecker

pipe = StableDiffusionSafetyChecker.from_pretrained('CompVis/stable-diffusion-safety-checker', device_map='auto')
```

sgugger:
Which branch should I check? On main I just get an error telling me device_map auto is not supported yet

patrickvonplaten:
Ah sorry just updated it above, can you try:

```python
from diffusers.pipelines.stable_diffusion import StableDiffusionSafetyChecker

StableDiffusionSafetyChecker._no_split_modules = ["CLIPEncoderLayer"]
pipe = StableDiffusionSafetyChecker.from_pretrained('CompVis/stable-diffusion-safety-checker', device_map='auto')
```

src/accelerate/utils/modeling.py
```diff
@@ -258,7 +258,7 @@ def get_max_layer_size(
     modules_to_treat = modules.copy()
     while len(modules_to_treat) > 0:
         module_name, module = modules_to_treat.pop(0)
-        modules_children = list(module.named_children())
+        modules_children = list(module.named_children()) if not isinstance(module, torch.nn.Parameter) else []
```
sgugger, line 261:
```suggestion
        modules_children = list(module.named_children()) if isinstance(module, torch.nn.Module) else []
```

sgugger (approved):
Thanks for the reproducer! The function is indeed used with some parameters inside accelerate, so this is needed.

Just suggested another test.

patrickvonplaten:
Awesome! Can I merge after one :heavy_check_mark: here or wait for another reviewer?

sgugger:
Nope, you can go ahead and merge!

Issue #745, opened 2022-10-06
Reduction on all distributed system only return the sum, irrespective of the provided parameter. Only works as intended on multi-cpu.


AbdurRafae:
https://github.com/huggingface/accelerate/blob/a54cd0abd8f1f1f5e0f64ac030427c30760309d8/src/accelerate/utils/operations.py#L442-L462

muellerzr:
@AbdurRafae when opening this issue there should have been a number of questions it prompted you for. In this case the most important way for us to quickly help and fix this would be for you to provide a reproducible example that we can then add to a test once the problem has been solved and we can look into. :)

sgugger:
Thanks a lot for flagging! Do you want to open a PR to fix this?

PR #748, opened 2022-10-08
Reduce function fix [WIP]

AbdurRafae:
The PR fixes the functionality of reduce operation in utils.operations as mentioned in #745. I have only tested it on multi-GPU since I don't have access to run tests on other distributed_type settings at the moment.


sgugger:
There is still one last styling error in the test. Make sure your version of black is 22.3.0 as maybe it comes from a different version?

AbdurRafae:
Apologies for not doing it right the first time. It's my first time contributing

sgugger:
No worries at all! But it now looks like you pushed a diff in way too many files. Are you sure you are using version 22.3?

PR #751, opened 2022-10-11
Add gpu_ids to SageMakerConfig though it should never be set

muellerzr:
Fixes issue reported here: https://discuss.huggingface.co/t/sagemakerconfig-object-has-no-attribute-gpu-ids/24174


sgugger (approved):
Thanks for fixing!

PR #753, opened 2022-10-11
Change num_cpu_threads_per_process default

muellerzr:
Modifies the default cpu_threads_per_process to be 1, and then if training on CPU then we adjust the number of threads based on an optimization to use them all. 

The need for this fix is there was not much of a want to have a high utilization on the CPU when training on GPUs, and when there was a performance gain wasn't really found. As a result we're limiting the tuning in this way to CPU-bound only training.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -319,7 +319,7 @@ def launch_command_parser(subparsers=None):
     parser.add_argument(
         "--num_cpu_threads_per_process",
         type=int,
-        default=None,
+        default=1,
         help="The number of CPU threads per process. Can be tuned for optimal performance.",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


src/accelerate/commands/launch.py
```diff
@@ -877,15 +877,15 @@ def launch_command(args):
         if not hasattr(args, "use_cpu"):
             args.use_cpu = args.cpu
 
-    if args.num_cpu_threads_per_process is None:
+    if args.num_cpu_threads_per_process == 1 and args.use_cpu:
```
sgugger, line 880: This means a chosen value of 1 is erased. Let's leave the default at `None` and change it at 1 when not in distributed CPU, or the other default when in distributed CPU.

sgugger (commented):
Thanks for working on this, left a comment.

src/accelerate/commands/launch.py
muellerzr: Adjusted :)

Issue #641, opened 2022-08-18
Gradient accumulation doesn't work with Accelerate's `clip_grad_norm_`


Gladiator07:
### System Info

```Shell
- `Accelerate` version: 0.13.0.dev0
- Platform: Linux-5.10.133+-x86_64-with-debian-bullseye-sid
- Python version: 3.7.12
- Numpy version: 1.21.6
- PyTorch version (GPU?): 1.11.0 (True)
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Steps to reproduce the behaviour:
You can directly run [this colab notebook](https://colab.research.google.com/drive/1ToR9b560xRfimpfRvGWk-aUBZO881-5M?usp=sharing) to get the error.

The main training method in the `Trainer` class is `train_one_epoch`
```python
for step, batch in enumerate(dataloader):
    with self._accelerator.accumulate(self.model):
        self.optimizer.zero_grad()
        _, loss = self.model(**batch)
        self._accelerator.backward(loss)
        self._accelerator.clip_grad_norm_(self.model.parameters(), self.args.max_grad_norm)
        self.optimizer.step()
        self.lr_scheduler.step()
        # assuming dataset has label as key
        self._trn_loss_meter.update(
            loss.item() * self.args.gradient_accumulation_steps, batch["label"].size(0)
        )
        if self._accelerator.sync_gradients:
            self.global_prog_bar.set_postfix(loss=self._trn_loss_meter.avg)
            self.global_prog_bar.update(1)
```
This will result in the following error:

```
─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ <ipython-input-21-5a5fa8902df5>:2 in <module>                                                    │
│                                                                                                  │
│ /usr/local/lib/python3.7/dist-packages/accelerate/launchers.py:83 in notebook_launcher           │
│                                                                                                  │
│    80 │   │   │   │   print("Launching training on one GPU.")                                    │
│    81 │   │   │   else:                                                                          │
│    82 │   │   │   │   print("Launching training on one CPU.")                                    │
│ ❱  83 │   │   │   function(*args)                                                                │
│    84 │                                                                                          │
[...]

PR #756, opened 2022-10-14
Return unclipped gradient from `clip_grad_norm_`

samuelstevens:
Returns the unclipped gradient from `clip_grad_norm_` like [`torch.nn.utils.clip_grad_norm_`](https://pytorch.org/docs/stable/_modules/torch/nn/utils/clip_grad.html#clip_grad_norm_) does.

Initially brought up in #641.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1194,6 +1194,9 @@ class Accelerator:
         """
         Should be used in place of `torch.nn.utils.clip_grad_norm_`.
 
+        Returns:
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for working on this!

lzy37ld:
@sgugger @samuelstevens 
Hi Thanks for this PR. I want to use clip_grad_norm_ in deepspeed intergration. But I notice that we only have a parameter for gradient clipping when use `accelerate launch`. How could I set gradient norm?

PR #761, opened 2022-10-14
Only wrap modules in DDP if they require grad

### LINKED ISSUE #760: RuntimeError: DistributedDataParallel is not needed when a module doesn't have any parameter that requires a gradient.
I normally just throw everything into `accelerator.prepare` that has any tensor buffers or parameters. However, if an object subclasses `torch.nn.Module`, has no parameters and we are in a multi-GPU context, then `accelerator.prepare` tries to wrap it in DistributedDataParallel, which throws an error:

```
RuntimeError: DistributedDataParallel is not needed when a module doesn't have any parameter that requires a gradient.
```

The quick fix is to simply move the object from `accelerator.prepare` to `obj.to(accelerator.device)` but it's nice to be able to leave everything in `accelerator.prepare` so I can add parameters, turn `requires_grad` on/off and not have to move code around.

I can open a new PR to fix this; I imagine it's just checking if any of the `model.parameters()` have `require_grad` set to true. If they don't then it's a no-op because it's already moved to `accelerator.device`.

https://github.com/huggingface/accelerate/blob/693d46826e32507376d44f99967df4710886c984/src/accelerate/accelerator.py#L781-L783

samuelstevens:
Like the title says. Fixes #760. 

There is [just such a check](https://github.com/pytorch/pytorch/blob/13cff2ee8ea1d7aea2ad201cbd77ebe2b9a29d25/torch/nn/parallel/distributed.py#L550-L555) in DistributedDataParallel, which I do the opposite of.

I would like to add a test so that this feature doesn't regress, but it's not clear to me how to add a test.

I could add a module with only buffers and add it to the `prepare()` call on line 78, then check that the device is the accelerator's device.

 https://github.com/huggingface/accelerate/blob/693d46826e32507376d44f99967df4710886c984/tests/test_kwargs_handlers.py#L73-L79

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -777,10 +777,11 @@ class Accelerator:
         if device_placement:
             model = model.to(self.device)
         if self.distributed_type == DistributedType.MULTI_GPU:
-            kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
-            model = torch.nn.parallel.DistributedDataParallel(
-                model, device_ids=[self.local_process_index], output_device=self.local_process_index, **kwargs
-            )
+            if any(p.requires_grad for p in model.parameters()):
+                kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
+                model = torch.nn.parallel.DistributedDataParallel(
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Very nice, thanks a lot!

Issue #767, opened 2022-10-17
Accelerate not moving anything to execution device after cpu_offload


piEsposito:
### System Info

```Shell
- `Accelerate` version: 0.13.2
- Platform: Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.31
- Python version: 3.9.12
- Numpy version: 1.23.3
- PyTorch version (GPU?): 1.12.1+cu116 (True)
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Run 

```python
import torch
from torch import nn
import accelerate

class Net(nn.Module):
    def __init__(self):
        super().__init__()
        self.nn_parameter = nn.Parameter(torch.ones(1), requires_grad=False)
        
    def forward(self, x):
        print(x.device, self.nn_parameter.device)
        return self.nn_parameter + x
    
net = Net()

assert torch.cuda.is_available()
net = accelerate.cpu_offload(net, execution_device=torch.device("cuda"))

x = torch.zeros(1, 1)
net(x)
```

Returns

`cpu meta`

and 

`RuntimeError: Expected all tensors to be on the same device, but found at least two devices, 
meta and cpu!`

### Expected behavior

```Shell
Should output


cuda:0 cuda:0
tensor([[1.]], device='cuda:0')
```
```

piEsposito:
I think I found the problem:

On https://github.com/huggingface/accelerate/blob/a63511107b099505bd499c6931bd28bf3dce73fe/src/accelerate/big_modeling.py#L146-L155

When we add the hook on top of the previous hook, the first hook is deleted and we lose all the devices placement hook.

sgugger:
Thanks for the very clear bug report. You have found the cause, although the fix is not quite right ;-)
Will comment on the PR!

PR #768, opened 2022-10-17
Move io_same_device hook to before attach_align_device hook on cpu_offload and disk_offload.

piEsposito:
Solves https://github.com/huggingface/accelerate/issues/767.

Moves the AlignDevicesHook with `io_same_device` from `accelerate.cpu_offload` and `accelerate.disk_offload` to before the `attach_align_device_hook`.

That way we can keep the changes on forward method for the whole module without deleting the hook we want to keep: the one with execution device and configurations on how to move the tensors between devices.

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -123,6 +123,7 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
     Args:
         module (`torch.nn.Module`): The module to attach a hook to.
         hook (`ModelHook`): The hook to attach.
+        append (`bool`): Whether, if module already contains a hook, should chain the new one with SequentialHook
 
     Returns:
         `torch.nn.Module`: The same module, with the hook attached (the module is modified in place, so the result can
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -108,7 +108,7 @@ class SequentialHook(ModelHook):
         return module
 
 
-def add_hook_to_module(module: nn.Module, hook: ModelHook):
+def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False):
     """
     Adds a hook to a given module. This will rewrite the `forward` method of the module to include the hook, to remove
     this behavior and restore the original `forward` method, use `remove_hook_from_module`.
@@ -128,6 +128,15 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook):
         `torch.nn.Module`: The same module, with the hook attached (the module is modified in place, so the result can
         be discarded).
     """
+
+    if append:
+        if hasattr(module, "_hf_hook") and (module._hf_hook is not None):
+            old_hook = module._hf_hook
+            remove_hook_from_module(module)
+
+            hooks = [old_hook, hook]
+            hook = SequentialHook(*hooks)
+
     if hasattr(module, "_hf_hook") and hasattr(module, "_old_forward"):
         # If we already put some hook on this module, we replace it with the new one.
         old_forward = module._old_forward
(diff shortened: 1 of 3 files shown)


sgugger:
The fix is not exactly right: by doing so, the hook that ensures the input and output of the model are on the same device is now erased. In your code sample in #767, since `x` is on the CPU, `net(x)` should also be on the CPU. This is not the case with your PR. The solution would be to write a util function that will:
- just add the hook if none is present
- extract the current hook if one is present and chain it with this hook using a `SequentialHook`.

This is slightly more advanced than the current PR, so let me know if you'd prefer for me to do it :-)

sgugger:
That works for me, though the name of the argument could simply be `append` :-)
Thanks for diving into this!

piEsposito:
@sgugger `append` it is.

src/accelerate/hooks.py
```diff
@@ -123,11 +123,21 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook):
     Args:
         module (`torch.nn.Module`): The module to attach a hook to.
         hook (`ModelHook`): The hook to attach.
+        append (`bool`): Whether, if module already contains a hook, should chain the new one with SequentialHook
```
sgugger, line 126:
```suggestion
        append (`bool`, *optional*, defaults to `False`):
            Whether the hook should be chained with an existing one (if module already contains a hook) or not.
```

Also can't comment on line 119 but this should also be mentioned in the Note.

src/accelerate/hooks.py
```diff
@@ -123,11 +123,21 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook):
...
 
     Returns:
         `torch.nn.Module`: The same module, with the hook attached (the module is modified in place, so the result can
         be discarded).
     """
+
+    if append:
+        if hasattr(module, "_hf_hook") and (module._hf_hook is not None):
```
sgugger, line 134: Those two ifs can be grouped.
Also the second test can be simplified in `getattr(module, "_hf_hook", None) is not None`

src/accelerate/hooks.py
```diff
@@ -123,11 +123,21 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook):
...
+            old_hook = module._hf_hook
+            remove_hook_from_module(module)
+
+            hooks = [old_hook, hook]
+            hook = SequentialHook(*hooks)
```
sgugger, line 139:
```suggestion
            hook = SequentialHook(old_hook, hook)
```

sgugger (commented):
Very nice, thanks! Left a couple of nits, and I think you should still put the hook with io first: just tested locally and we still have the same issue of `net(x)` being on the wrong device since it runs second and the input was already moved.

src/accelerate/hooks.py
piEsposito: Done.

src/accelerate/hooks.py
piEsposito: Done, thanks for the great idea.

src/accelerate/hooks.py
piEsposito: Done. Looks much better now.

piEsposito:
> Very nice, thanks! Left a couple of nits, and I think you should still put the hook with io first: just tested locally and we still have the same issue of `net(x)` being on the wrong device since it runs second and the input was already moved.

@sgugger I've just addressed your nits and moved the io hook to the top. Thanks for the review!

sgugger (approved):
Perfect, thanks! Re-tested locally and got the expected results for the code sample you shared in the issue.

piEsposito:
@sgugger, there is a test step that failed due to an http error when installing a lib. Can we try running it again or should I create a new commit to trigger the CI again?

PR #772, opened 2022-10-18
Regression cli tests

muellerzr:
# Add `accelerate launch` tests

## What does this add?

This PR introduces a subset of tests for checking `accelerate launch` and in turn `accelerate config`, and solves a bug I found during testing

## Why is it needed?

Users have pointed out that we're not quite doing a good enough job with backwards compatibility checks on config yamls. To ensure these can be caught new tests are added to run a simple script using `accelerate launch` with no configuration file or specific configuration files

### User-facing:

Nothing

### Internal structure:

`tests/test_configs` should contain (at *least*) the last 2 major versions of Accelerate plus the latest version, which are direct outputs from `accelerate config`. Since these tests take so fast to execute, I think it's fine if this just contains the last 5 or 6 configs until we are comfortable. 

As part of PR review, maintainers should be aware of any changes to the config file or CLI arguments and ensure that a new config is added to the folder for proper testing.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -432,7 +432,7 @@ def simple_launcher(args):
     current_env["USE_MPS_DEVICE"] = str(args.use_mps_device)
     if args.use_mps_device:
         current_env["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"
-    elif args.gpu_ids != "all":
+    elif args.gpu_ids != "all" and args.gpu_ids is not None:
         current_env["CUDA_VISIBLE_DEVICES"] = args.gpu_ids
     if args.num_machines > 1:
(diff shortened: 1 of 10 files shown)


src/accelerate/commands/launch.py
```diff
@@ -489,7 +489,7 @@ def multi_gpu_launcher(args):
 
     current_env = os.environ.copy()
     gpu_ids = getattr(args, "gpu_ids")
-    if gpu_ids != "all":
+    if gpu_ids != "all" and args.gpu_ids is not None:
```
muellerzr, line 492: This addition is needed for the test case of `CUDA_VISIBLE_DEVICES=""`

sgugger (approved):
Thanks a lot for working on this!

PR #773, opened 2022-10-18
Fix all github actions issues + depreciations

muellerzr:
Fixes a duplicate of writing to the same cache location that was causing a partial slowdown. Also goes ahead and changes `set-output` to `$GITHUB_ENV` as this will be githubs new way of modifying bits like that safer.

cc @ydshieh so you're aware of it, docs are here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files


ydshieh:
Thank you @muellerzr for updating my knowledge!

sgugger (approved):
Thanks!

PR #775, opened 2022-10-18
Fix flakey wandb test

muellerzr:
Completely redo's the wandb tests to run a single experiment that then checks that init and the logs worked fine. Uses a new processing system the wandb team shared with me that is **infinitely** more reliable and we can successfully parse from repeatedly without trouble


sgugger (approved):
Thanks for fixing!

PR #777, opened 2022-10-19
fix transformers tests

ArthurZucker:
# Fix Transformer `test_model_parallelism` tests
Will add accelerate tests for that asap. 
For the mean time, 
The last device is not left with `max_memory`. This cause the `test_model_parallelism` to fail in transformers.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -409,7 +409,7 @@ def get_balanced_memory(
 
     max_memory = get_max_memory(max_memory)
     # The last device is left with max_memory just in case the buffer is not enough.
-    for i in range(len(max_memory) - 1):
+    for i in range(num_devices - 1):
         max_memory[i] = min(0 if low_zero and i == 0 else per_gpu, max_memory[i])
 
     if low_zero:
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -408,8 +408,9 @@ def get_balanced_memory(
     per_gpu += buffer
 
     max_memory = get_max_memory(max_memory)
+    last_gpu = max(i for i in max_memory if isinstance(i, int) and max_memory[i] > 0)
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -409,7 +409,7 @@ def get_balanced_memory(
 
     max_memory = get_max_memory(max_memory)
     # The last device is left with max_memory just in case the buffer is not enough.
-    for i in range(len(max_memory) - 1):
+    for i in range(num_devices - 1):
```
younesbelkada, line 412:
```suggestion
    offset = 2 if "cpu" in (list(max_memory.keys())) else 1
    for i in range(len(max_memory) - offset):
```
I think this way it seems that the slow tests of `accelerate` are not broken and we also keep the `transformers` tests pass! Can you confirm this pass the `transformers` tests that were broken @ArthurZucker ? 🙏

younesbelkada: Ahhh great! I was missing this test! Thanks for clarifying 💪

sgugger: This won't work when the last device is GPU 3 but the user forced 0 at GPU 0, since `num_devices` will be 3 (instead of 4) and thus the max memory will be at GPU 2 instead of 3. (you can try running the multi-GPU tests and they will fail).

Agreed that `len(max_memory)` is not good either, so we should just compute the biggest integer in the keys of `max_memory`:
```py
last_gpu = max(i for i in max_memory if isintance(i, int))
```
then use this.

ArthurZucker: Okay will add this suggestion 💯

sgugger: Oh and we should avoid a GPU that the user set at 0, so actually
```
last_gpu = max(i for i in max_memory if isintance(i, int) and max_memory[i] > 0)
```

sgugger (approved):
Thanks for the fix!

Issue #765, opened 2022-10-17
Accelerate 0.13.0 breaks tracebacks in Jupyter-Lab


trygvebw:
### System Info

```Shell
Jupyter-Lab 3.4.3. Accelerate 0.13.0 or 0.13.1.
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. Open Jupyter-Lab and create a new notebook.
2. Run `import accelerate`
3. Run `abc`

Result: Sorry for the screenshot, but given that this is an issue with Jupyter, a screenshot shows the problem the best. The traceback (expected, as `abc` is a randomly chosen nondefined variable) is oddly formatted and, more importantly, doesn't show the contents of the cell that caused the error as it seems to read *that* from some file in `/tmp` that doesn't exist.

![image](https://user-images.githubusercontent.com/1441500/196119712-4a88c8c7-c30e-47ae-aac7-5d5f4d57c827.png)

Compare with the traceback I get if I *don't* import accelerate before step 3:

![image](https://user-images.githubusercontent.com/1441500/196120573-bd68c2df-83f0-420c-bdde-6fe5cc100b02.png)

### Expected behavior

```Shell
Accelerate should not impact the formatting of the traceback, and the traceback should show the line that caused the error and the context (the surrounding lines) if any.
```

muellerzr:
@trygvebw you have the optional dependency `rich` installed which can have a few issues with jupyter. Please do `pip uninstall rich` in your environment

sgugger:
@muellerzr I think we should also provide an env variable to turn off the use of `rich` if folks don't want to use it without uninstalling it.

muellerzr:
Hi @trygvebw, with https://github.com/huggingface/accelerate/pull/779 disabling rich can now be set in a jupyter notebook by running `%env DISABLE_RICH=1`. 

Please try it out if you have a moment by installing Accelerate with `pip install git+https://github.com/huggingface/accelerate@toggle-rich`. Thank you for your feedback! 🤗

fractaldna22:
thank you for this. If you see my posts on various repos, my posts are identical.  "the tracebacks look like this  * and i dont know why, they're unusable, and they used to look like this *
[...]

PR #779, opened 2022-10-19
Make rich toggleable and seperate out a new environment utility file

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/765 by making rich be a toggle-able setting. To disable rich for logging and stack trace, set the environment variable `DISABLE_RICH` to `1`.

This can be done a number of ways:

In Jupyter:
```python
%env DISABLE_RICH=1
# OR
%env DISABLE_RICH=yes
# OR
%env DISABLE_RICH=YES
```
In Bash:
```bash
export DISABLE_RICH=1
```
(with similar flags)

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import importlib
+import os
 import sys
 from functools import lru_cache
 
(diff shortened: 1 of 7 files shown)


sgugger (commented):
No need to create a new function IMO, the test can be done inside the availability test function (that's how we do it in Transformers for PyTorch/TF/Flax for isntance).

src/accelerate/utils/imports.py
```diff
@@ -126,7 +127,7 @@ def is_boto3_available():
 
 
 def is_rich_available():
-    return importlib.util.find_spec("rich") is not None
+    return (importlib.util.find_spec("rich") is not None) and (not os.environ.get("DISABLE_RICH", "0") == "1")
```
sgugger, line 130: Let's accept anything that's truthy here? Like yes/no, true/false, 0/1

sgugger (approved):
Thanks for working on this!

Issue #770, opened 2022-10-18
accelerate launch breaks in 0.13+ due to missing same_network


rossbucky:
### System Info

`accelerate launch` would fail by using the args directly without a config file.

`same_network` arg is missing in command/launch.py while it is used here.

```python
if getattr(args, "same_network"):
AttributeError: 'Namespace' object has no attribute 'same_network'
```
https://github.com/huggingface/accelerate/blob/a63511107b099505bd499c6931bd28bf3dce73fe/src/accelerate/commands/launch.py#L473

Is it possible to set the default value for launch.py args for better compatibility? Since auto jobs may not use dynamic consoles to choose setting but directly rely on args in accelerate launch. For example, `gpu_ids` is also missing default value `all` in launch.py, which also breaks the old working script.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

`accelerate launch --args` instead of setting it in the interactive console.

### Expected behavior

```Shell
1. compatibility
2. better tests for accelerate launch + args
```

rossbucky:
cc @sgugger @pacman100

muellerzr:
Thanks! This PR added some support for the regression tests: https://github.com/huggingface/accelerate/pull/772/files

I wasn't able to reproduce your behavior you found with `gpu_ids`, could you please post an example of the config you're working with?

muellerzr:
@rossbucky https://github.com/huggingface/accelerate/pull/778 this should also have helped some. Can you try installing from main to see if so? (the gpu_ids fix also just went in and I don't *think* it was sent out on the latest patch)

Try installing via `pip install git+https://github.com/huggingface/accelerate`. We'll certainly work on doing much better at checking for these types of things, thank you for bringing it to our attention 😄 🙏

rossbucky:
@muellerzr thanks for the fix. This would prevent it from crashing. But it would make the multi node not working. 

Can we add this?
```
parser.add_argument(
        "--same_network",
        default=False,
        help="blabla",
    )
```

muellerzr:
[...]

PR #780, opened 2022-10-19
Add same_network to CLI args and update docs

muellerzr:
Fulfills missing param not found in `launch` brought up in https://github.com/huggingface/accelerate/issues/770#issuecomment-1284295259, also working on a PR in with tests for checking that all possible args can be passed to the CLI manually soon


sgugger (approved):
LGTM!

Issue #684, opened 2022-09-07
Could `join` replace `gather_for_metrics`?


Chris-hughes10:
Hi,

Great job with accelerate! 

One persistent headache that I keep experiencing is the repeated samples during distributed evaluation. Although there is the `gather_for_metrics` functionality this doesn't always work depending on the output of the model; for example, Faster-RCNN in Torchvision, which outputs a list of dictionaries containing tensors.

I was wondering if it would be possible to remove the repeated behaviour completely, using something like the [join](https://pytorch.org/tutorials/advanced/generic_join.html) context manager, which enables training on uneven outputs.

If there is appetite for this, I would be happy to help you explore options.

Thanks

muellerzr:
Hey @Chris-hughes10, sorry for the delayed response on this we all had our vacations + getting *back* from vacations. Looking at this now and I do believe it'd be something we'd want to do!

Were you hoping for something more than just a wrapper around the context manager? 

E.g. this API would likely look something such as:

```python
with accelerator.uneven_outputs(model, optim, **kwargs):
   loss = model(input).sum()
   ...
```

Chris-hughes10:
Hey @muellerzr, no worries at all, glad that you are interested.

For me, it would be better if you didn't have to use the context manager at all, and it was all taken care of behind the scenes by the accelerator, but I understand that this is not as explicit and haven't looked into the details of how this would work.

I think that this would also require changes to the BatchSampler shard, so that it supports uneven batch sizes, as well as making sure that any wrappers that are added to the model or the optimizer play nicely with `joinable`.

muellerzr:
I also understand the benefit of that, however that does indeed consider a large amount of rewrite to get there. For now I think we'll move forward with just the small wrapper shown there, and I'll rename this github issue to reflect what would be done/proposed.

If you would like this feature please react with a 👍 to this comment :)

Chris-hughes10:
I agree that the small wrapper would be easier to implement to begin with, and I would be happy with this as a starting point, but I also think that the BatchSamplerShard would have to be modified at the same time. Without modifying the BatchSamplerShard, I don't think you will ever be in a position to have uneven batch sizes - as the additional samples will either be dropped or padded - so the wrapper wouldn't add any value.

Am I correct here or am I missing something?

muellerzr:
@Chris-hughes10 correct, it would not support dispatched batches, only sharded/iterable datasets.

Chris-hughes10:
[...]

PR #781, opened 2022-10-20
Add `even_batches` keyword to Accelerator

Chris-hughes10:
This PR enables the option of configuring the Accelerator to disable prepared DataLoaders ensuring that an even number of batches are observed across all processes. 

This continues the work described in: https://github.com/huggingface/accelerate/issues/684

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -153,6 +153,10 @@ class Accelerator:
             If set to `True`, the dataloader prepared by the Accelerator is only iterated through on the main process
             and then the batches are split and broadcast to each process. Will default to `True` for `DataLoader` whose
             underlying dataset is an `IterableDataset`, `False` otherwise.
+        even_batches ('bool', defaults to `True`):
+            If set to `True`, in cases where the total batch size across all processes does not exactly divide the
+            dataset, samples at the start of the dataset will be duplicated so the batch can be divided equally among
+            all workers.
         step_scheduler_with_optimizer (`bool`, *optional`, defaults to `True`):
             Set `True` if the learning rate scheduler is stepped at the same time as the optimizer, `False` if only
             done under certain circumstances (at the end of each epoch, for instance).
@@ -191,6 +195,7 @@ class Accelerator:
         log_with: Optional[List[Union[str, LoggerType, GeneralTracker]]] = None,
         logging_dir: Optional[Union[str, os.PathLike]] = None,
         dispatch_batches: Optional[bool] = None,
+        even_batches: bool = True,
         step_scheduler_with_optimizer: bool = True,
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
     ):
@@ -305,6 +310,7 @@ class Accelerator:
             raise ImportError(
                 "Using `DataLoaderDispatcher` requires PyTorch 1.8.0 minimum. You have {torch.__version__}."
             )
+        self.even_batches = even_batches
         self.step_scheduler_with_optimizer = step_scheduler_with_optimizer
 
         # Mixed precision attributes
@@ -1109,6 +1115,7 @@ class Accelerator:
             put_on_device=device_placement,
             rng_types=self.rng_types.copy(),
             dispatch_batches=self.dispatch_batches,
+            even_batches=self.even_batches,
         )
 
     def prepare_optimizer(self, optimizer: torch.optim.Optimizer, device_placement=None):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -119,7 +119,7 @@ class BatchSamplerShard(BatchSampler):
     <Tip warning={true}>
 
     `BatchSampler`s with varying batch sizes are not enabled by default. To enable this behaviour, set `even_batches`
-    equal to `True`
+    equal to `False`
 
     </Tip>"""
 
@@ -628,7 +628,7 @@ def prepare_data_loader(
     <Tip warning={true}>
 
     `BatchSampler`s with varying batch sizes are not enabled by default. To enable this behaviour, set `even_batches`
-    equal to `True`
+    equal to `False`
 
     </Tip>
     """
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -118,7 +118,8 @@ class BatchSamplerShard(BatchSampler):
 
     <Tip warning={true}>
 
-    This does not support `BatchSampler` with varying batch size yet.
+    `BatchSampler`s with varying batch sizes are not enabled by default. To enable this behaviour, set `even_batches`
+    equal to `True`
 
     </Tip>"""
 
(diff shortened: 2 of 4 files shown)


Chris-hughes10:
Hi, I have made a start on the work discussed in https://github.com/huggingface/accelerate/issues/684, and added an argument to the accelerator. 

I am slightly unsure about your testing strategy though, and which tests you require for this; I couldn't seem to find any integration tests for accelerator arguments. So far, I have tried to follow similar ideas to those seen in the end-to-end multi-GPU tests. 

I am slightly confused by exactly how the Multi-GPU tests are triggered though; whilst I see that there is a decorator to skip certain tests when multiple GPUs are not available, when I try to run this locally (on a machine with 2 GPUs), those tests only seem to execute on a single process.

Any guidance here would be appreciated! cc: @muellerzr @sgugger

muellerzr:
Hey @Chris-hughes10! Thanks!

You're absolutely right, I should probably write this all up in a contribution doc and will add it to my list of tasks!

Accelerates tests are setup in a few ways, but the key idea to remember is:

If it's something to be tested across multiple devices/workers, such as CPU, TPU, and mutli-GPU (such as your tests you want to add), these all get thrown into a script located in `accelerate/test_utils/scripts`

For example, the major integration script that checks the dataloaders operate properly is located at `accelerate/test_utils/scripts/test_script.py` (I wonder if we should section this out a bit more @sgugger since we now have a whole folder dedicated to it and we can clean it up some),  which you can find here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/scripts/test_script.py

These tests are setup like how a `pytest` test would be, where you have `def some_test` and `assert` statements to check if things are working properly. 

From here your script should be added as a test case to `tests/test.py`, `tests/test_cpu.py` (if applicable for multi-cpu),  `tests/test_multigpu.py`, and `tests/test_tpu.py`.

It has to be done this way because each platform has different launching configurations :) 

Finally, when running pytest you need to set the cuda visible devices for it to be able to pick up on the multiple GPUs, such as:

```bash
CUDA_VISIBLE_DEVICES="0,1" pytest -sv tests/test_multigpu.py
```
Or:
```bash
CUDA_VISIBLE_DEVICES="" pytest -sv tests/test_multigpu.py
```
(the latter will have all their tests skipped). 

All skipping decorator conditionals can be found here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/testing.py

Let me know if there are any other questions you have on this @Chris-hughes10, afterwards I can throw this all up into some documentation :)

Chris-hughes10:
Hey @muellerzr, Thanks for the info!

So, it looks like I was on the right track regarding the structure, but I was running with unittests and not Pytest. I have refactored it to pytest, and it seems to work alright using `CUDA_VISIBLE_DEVICES="0,1" pytest -sv tests/test_multigpu.py, so it is probably ready for an initial review. 

As an aside, how do you guys set up your development environment? I have set up a dev container, using the accelerate-gpu image, if it is useful, I could push that in another PR.

muellerzr:
@Chris-hughes10 I use conda for mine, and the CI's all run based off the conda images as there's hosted computes we use

Chris-hughes10:
> @Chris-hughes10 I use conda for mine, and the CI's all run based off the conda images as there's hosted computes we use

Ok, if you are using conda envs, I think that this could be useful. I'll create a small PR and you guys can see what you think.

src/accelerate/accelerator.py
```diff
@@ -153,6 +153,10 @@ class Accelerator:
             If set to `True`, the dataloader prepared by the Accelerator is only iterated through on the main process
             and then the batches are split and broadcast to each process. Will default to `True` for `DataLoader` whose
             underlying dataset is an `IterableDataset`, `False` otherwise.
+        even_batches ('bool', defaults to `True`):
```
sgugger, line 156:
```suggestion
        even_batches (`bool`, *optional*, defaults to `True`):
```

src/accelerate/data_loader.py
```diff
@@ -615,15 +616,15 @@ def prepare_data_loader(
             If set to `True`, the datalaoder prepared is only iterated through on the main process and then the batches
             are split and broadcast to each process. Will default to `True` when the underlying dataset is an
             `IterableDataset`, `False` otherwise.
+        even_batches ('bool', defaults to `True`):
```
sgugger, line 619:
```suggestion
        even_batches (`bool`, *optional*, defaults to `True`):
```

src/accelerate/data_loader.py
```diff
@@ -615,15 +616,15 @@ def prepare_data_loader(
...
+            If set to `True`, in cases where the total batch size across all processes does not exactly divide the
+            dataset, samples at the start of the dataset will be duplicated so the batch can be divided equally among
+            all workers.
 
     Returns:
         `torch.utils.data.dataloader.DataLoader`: A new data loader that will yield the portion of the batches
 
-    <Tip warning={true}>
-
-    This does not support `BatchSampler` with varying batch size yet.
```
sgugger, line 624: You should leave the warning but rephrase to state a `BatchSampler` with varying batch size will require `even_batches=False`.

sgugger (approved):
Looking nice! Left some comments on the docs but the rest looks good to me!

src/accelerate/data_loader.py
Chris-hughes10: Sure, I can update that

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -0,0 +1,103 @@
...
+    verify_dataloader_batch_sizes(
+        accelerator,
+        dataset_size=7,
+        batch_size=2,
+        process_0_expected_batch_sizes=[2, 2],
+        process_1_expected_batch_sizes=[2, 2],
+    )
+
+
+def test_can_disable_even_batches():
```
muellerzr, line 86: You're currently not testing anything at all, it's never being ran. Please see here to learn how: https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/scripts/test_sync.py#L229-L274

I also highly recommend using print statements so we can see just what is being tested at a given moment, and we know if they've run 😄

muellerzr (changes requested):

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
Chris-hughes10: Gah! I think I misinterpreted what you meant when you said PyTest format. Looking at the other example, it looks like it still needs a main method to drive running the tests; just what I removed a few commits back :satisfied:! I'll reinsert the main method and some print statements.

Chris-hughes10: I've just pushed some changes that hopefully resolve this. Just FYI, it is getting pretty late here in the UK, and I am heading off on vacation tomorrow until the 31st. If there are other changes required, and you would like this merged before I'm back, please feel free to update the branch with any further changes!

Chris-hughes10:
Hey @muellerzr, are there any further changes required on this?

muellerzr:
@Chris-hughes10 close, your tests still weren't passing, and this is because `nproc_per_node` needed to be specified in the test function. (Discussing with Sylvain if this should be *everywhere* or not) but basically without that `Accelerator.num_processes` was only ever equal to one. 

Went ahead and pushed the fix :)

muellerzr (approved):

sgugger:
Thanks for all your work on this!

PR #782, opened 2022-10-20
Add Dev Container configuration

Chris-hughes10:
This PR aims to make the contribution process smoother by providing a dev container configuration, which enables VSCode users to develop inside one of the defined Docker containers.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,25 @@
+// File only needed for VSCode users to have proper Docker based interpreters
+{
+    "name": "accelerate_dev_environment",
+    "build": {
+        // ACTION NEEDED: comment/uncomment the relevant line depending on whether you are in a CPU/GPU environment
+        // "dockerfile": "../docker/accelerate-cpu/Dockerfile"
+        "dockerfile": "../docker/accelerate-gpu/Dockerfile"
+    },
+    "runArgs": [
+        // ACTION NEEDED: comment the next line if your local machine has no GPUs available
+        "--gpus", "all",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+// File only needed for VSCode users to have proper Docker based interpreters
+{
+    "name": "accelerate_dev_environment",
+    "build": {
+        // ACTION NEEDED: comment/uncomment the relevant line depending on whether you are in a CPU/GPU environment
+        // "dockerfile": "../docker/accelerate-cpu/Dockerfile"
```
sgugger, line 6: I think we should enable CPU by default and have the user work on the container to activate GPU.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+        "dockerfile": "../docker/accelerate-gpu/Dockerfile"
+    },
+    "runArgs": [
+        // ACTION NEEDED: comment the next line if your local machine has no GPUs available
```
sgugger, line 10: Same here.

sgugger (commented):
Thanks for adding this! Left a few comments regarding the default of the container.

Also, if the users change the devcontainer.json for their local use, won't those changes be pushed to GitHub automatically?

.devcontainer/devcontainer.json
Chris-hughes10: Sure, I was torn on this, so am happy to go the other way.

sgugger (approved):

Issue #786, opened 2022-10-23
[Feature Request] Split `sampler.generator` and `loader.generator` to leave loader's generator unsynchronized by default.


YouJiacheng:
`DataLoaderShard` will synchronize `self.generator`(i.e. `loader.generator`) of all processes at each new iteration.
The [documentation](https://huggingface.co/docs/accelerate/quicktour#internal-mechanism) says "the generator attribute of a given sampler" will be synchronized.
However current implementation also synchronize loader's generator, because `prepare_data_loader` will pass sampler's generator to `DataLoaderShard.__init__`, which executes `self.generator = generator`
Theoretically, only `sampler.generator` is required to be synchronized so that each process sees the same random permutation of samples.
Is this intended, or the expected behavior should be `self.sampler_generator = generator`?

PyTorch `DataLoader` will use `loader.generator` to generate `base_seed` for workers.
Without `accelerate`, one can simply call `torch.manual_seed` with different seed, or pass different generator to `DataLoader` in different process, to get different `base_seed`.

`base_seed` will be used to seed main random number generator of worker process:
https://github.com/pytorch/pytorch/blob/eafc910d16a99200af089099e24468c7f8926a05/torch/utils/data/_utils/worker.py#L208-L229

However, loader prepared by `accelerate` will synchronize `loader.generator`, which is used to generate `base_seed`.

It took me 1 hour to figure out why `torch.manual_seed(device.index)` doesn't work -- all GPU's DataLoader get the same `base_seed` for workers. (Initially I expect `loader.generator is None`, however it doesn't. It turns out that `accelerate` will create a generator for prepared loader.)

The warning in the [documentation](https://huggingface.co/docs/accelerate/quicktour#internal-mechanism) clearly show that "Synchronization of the main torch (or CUDA or XLA) random number generator" can cause duplicated random data augmentation.
> Synchronization of the main torch (or CUDA or XLA) random number generator will affect any other potential random artifacts you could have in your dataset (like random data augmentation) in the sense that all processes will get the same random numbers from the torch random modules (so will apply the same random data augmentation if it’s controlled by torch).

However, with the mechanism described above, loader prepared by `accelerate` can cause duplicated random data augmentation without "Synchronization of the main torch (or CUDA or XLA) random number generator".

Even worse, this problem only happens if `num_workers > 1`! I think this is undesired.

sgugger:
[...]

PR #789, opened 2022-10-25
separate dataloader generator from sampler generator

pacman100:
### What does this PR do?
1. Fixes #786 

Example script for testing the updated code:
```python 
import torch
from torch.utils.data import default_collate,default_convert, Dataset, DataLoader, BatchSampler, RandomSampler, SequentialSampler
import numpy
import accelerate
from accelerate import Accelerator
from accelerate.utils import set_seed
from accelerate.utils.random import synchronize_rng_state
from accelerate.utils.dataclasses import DistributedType, RNGType
import os
import sys
import platform

class MyDataset(Dataset):
    def __len__(self):
        return 22
    
    def __getitem__(self, index):
#         print("MyDataset __getitem__", index)

        squeeze = False

        if isinstance(index, int):
            index = [index]
            squeeze = True
        elif isinstance(index, slice):
            index = list(range(*index.indices(self.size)))
        else:
            index = list(index)

        batch = [{"index": i, "label": i % 2, "random_augmentation": torch.rand(1).item()} for i in index]
#         print(batch)

        if squeeze:
            batch = batch[0]

        return batch
    

if __name__ == "__main__":
    
    num_workers = 4
    dataset = MyDataset()
    accelerator = Accelerator()
    torch.manual_seed(accelerator.process_index)
    
    accelerator.print("Starting conventional Dataloader with shuffle=False. Eval mode in general")
    loader = DataLoader(dataset, shuffle=False, batch_size=4, num_workers=num_workers)
    loader = accelerator.prepare(loader)
    all_examples = []
    for i, batch in enumerate(loader):
        print(f"{accelerator.process_index} | batch #{i} = {batch}")
        index, label = accelerator.gather_for_metrics((batch["index"], batch["label"]))
        all_examples.extend(index.detach().cpu().numpy().tolist())
        accelerator.print(f"{accelerator.process_index} | gathered batch #{i} | index = {index}, label = {label}")
    accelerator.print(f"{sorted(all_examples)=}")
    accelerator.print("Ending conventional Dataloader with shuffle=False. Eval mode in general")
    accelerator.print()
    accelerator.print()
    
    accelerator.print("Starting conventional Dataloader with shuffle=True")
    loader = DataLoader(dataset, shuffle=True, batch_size=4, num_workers=num_workers)
    loader = accelerator.prepare(loader)
    all_examples = []
    for i, batch in enumerate(loader):
        print(f"{accelerator.process_index} | batch #{i} = {batch}")
[...]

PR #790, opened 2022-10-25
separate dataloader generator from sync generator

YouJiacheng:
@pacman100 @sgugger 
What do you think about this way to fix the issue #786 ? 
Comparing to #789 .


pacman100:
Closing this. Thank you for your insights and suggestions 🤗

PR #792, opened 2022-10-26
Consider top-level buffers when computing `infer_auto_device_map` 

younesbelkada:
# What does this PR do?

This PR adds `list(model._buffers)` inside `modules_to_treat` when computing the `auto_device_map`. This scenario occured when I tried to add `accelerate` support for `BART`-like models when the [`final_logits_bias` is registered as a buffer and is different than a `uint` type](https://github.com/huggingface/transformers/blob/76296569266fbc238defbe68d06ad720b3a544e3/src/transformers/models/bart/modeling_bart.py#L1286). It seems that we need to assign a device to this buffer. 

The other solution is to "force-ignore" the buffer in `check_device_map` [here](https://github.com/huggingface/accelerate/blob/a5525406fcba79523fc5edf1051360e15447ef6a/src/accelerate/utils/modeling.py#L577) since the tensors that are in `model._buffers` are stored in the state_dict. 

cc @sgugger @muellerzr 

slow tests from `tests/test_bigmodeling.py` pass!

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -485,6 +485,7 @@ def infer_auto_device_map(
     # Ready ? This is going to be a bit messy.
     while len(modules_to_treat) > 0:
         name, module = modules_to_treat.pop(0)
+
         # Max size in the remaining layers may have changed since we took one, so we maybe update it.
         max_layer_names = [n for n in max_layer_names if not n.startswith(name)]
         if len(max_layer_names) == 0:
@@ -504,6 +505,13 @@ def infer_auto_device_map(
         # Reduce max size available by the largest layer.
         if devices[current_device] in main_devices:
             current_max_size = current_max_size - max_layer_size
+
+        # Case 0: Put directly the buffer on the device_map - usually buffers are small
+        if name in model._buffers:
+            current_memory_used += module_size
+            device_map[name] = devices[current_device]
+            continue
+
         # Case 1 -> We're too big!
         if current_max_size is not None and current_memory_used + module_size > current_max_size:
             # Split or not split?
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -475,10 +475,11 @@ def infer_auto_device_map(
     current_device = 0
     current_memory_used = 0
 
-    # Direct submodules and parameters
-    modules_to_treat = (
-        list(model.named_parameters(recurse=False)) + list(model.named_children()) + list(model.named_buffers())
-    )
+    # Direct submodules and parameters, including tensors that are registered as buffers
+    # but excluding buffers such as `running_mean` for batch_norm
+    filtered_buffers = list((m[0], m[1]) for m in model.named_buffers() if m[0] in model._buffers)
+    modules_to_treat = list(model.named_parameters(recurse=False)) + list(model.named_children()) + filtered_buffers
+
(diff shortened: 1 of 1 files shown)


sgugger (commented):
Can you explain how some buffers don't end up in `model._buffers` because I don't fully understand that part.

younesbelkada:
So if I understood it correctly, if you have some modules such as `nn.BatchNorm` in your model (as it is done in the `accelerate` CI test), the buffers `running_mean` and `running_var` will not be stored inside `model._buffers` but in `model.named_buffers()`. That is why I had to "filter out" the buffers by considering only the ones that are inside `model._buffers` and `model.named_buffers()` 

Here is an example that I have quickly tried: 
```
model = nn.Sequential(nn.Linear(1, 1), nn.BatchNorm1d(1), nn.Embedding(1, 1), nn.LayerNorm(1)) print(list(model.named_buffers()))
>>>[('1.running_mean', tensor([0.])), ('1.running_var', tensor([1.])), ('1.num_batches_tracked', tensor(0))]
print(list(model._buffers))
>>> []
model.register_buffer("position_bias", torch.ones(1))
print(list(model._buffers))
>>> ['position_bias']
print(list(model.named_buffers()))
>>> [('position_bias', tensor([1.])), ('1.running_mean', tensor([0.])), ('1.running_var', tensor([1.])), ('1.num_batches_tracked', tensor(0))]
```

sgugger:
I think in this case, it's just the difference between `named_buffers(recurse=True)` and `named_buffers(recurse=False)`. I'm not convinced this fix is the right fix, so would like to learn more what is failing.

younesbelkada:
Ah yes I see, you're probably right here! Let me dig a bit more and get back to you here

younesbelkada:
@sgugger I might have more clue on what is failing 
I think that the problem comes from the fact that the `infer_auto_device_map` does take into account only modules and submodules, I have made a script below to better illustrate the problem

```
import torch.nn as nn
import torch
from accelerate.utils import infer_auto_device_map
from accelerate.big_modeling import dispatch_model

class SubModule(nn.Module):
    def __init__(self):
        super().__init__()

        self.register_buffer("position_bias", torch.ones(1, 1000))

class Model(nn.Module):
    def __init__(self, wrap_module=True):
        super().__init__()
        self.l1 = nn.Linear(1000, 1000)
        self.l2 = nn.Linear(1000, 1000)
        self.l3 = nn.Linear(1000, 1000)

        self.bn1 = nn.BatchNorm1d(1000)
        self.bn2 = nn.BatchNorm1d(1000)

        if wrap_module:
            self.position_bias = SubModule()
        else:
            self.register_buffer("position_bias", torch.ones(1, 1000))

# Test 1: wrapping with a module
model = Model()
device_map = infer_auto_device_map(model, {0:"10MB", "cpu":"100MB"})
model = dispatch_model(model, device_map)

# Test 2: below will fail
model = Model(wrap_module=False)
device_map = infer_auto_device_map(model, {0:"10MB", "cpu":"100MB"})
model = dispatch_model(model, device_map)
```
Let me know what do you think!

sgugger:
Ah, in this case it looks very much like the problem #747 fixed for top-level parameters, so the fix should be pretty similar here too!

sgugger (approved):
Perfect, thanks!

younesbelkada:
The whole testing suite (including slow tests) is green! 🟢 Merging !

PR #795, opened 2022-10-27
Fix extraction of state dict in offload

sgugger:
When we end up with weights offloaded on CPU and disk, there is a little bug in the function that extracts the CPU state dict. In the added tests, it would catch the submodules beginning like the one passed (so `a.10` because `a.1` is in the list).


pacman100 (approved):
Thanks for the fix! LGTM 🤗

muellerzr (approved):
Thanks for the fix and test!

PR #797, opened 2022-10-28
Fix device_map="auto" on CPU-only envs

### LINKED ISSUE #796: Question regarding `device_map="auto"`
In `diffusers` we essentially want to offload all the maintenance of efficient model loading (meaning using torch's meta tensors to not allocate any memory at init). 

We'd like to do so via just one logic/kwarg being `device_map="auto"`. This means we would like to **always** advocate to pass `device_map="auto"` stating that this is always in every environment the recommend way of loading a modedl.
This works great on GPU, but when loading the model on CPU with `from_pretrained(...., device_map="auto")` we're getting: 
```
tests/models/test_models_unet_2d.py::UNetLDMModelTests::test_from_pretrained_hub - IndexError: list index out of range
```

=> Is this expected?

sgugger:
This PR makes it possible to run `dispatch_model` with `device_map="auto"` on CPU-only envs. It has been tested for models that fit on the CPU, but not with offload to disk (which should work in theory, but there might still be bugs).
Making sure the offload to disk works on a CPU-only env is going to be further work however and go in another PR.

Fixes #796 
cc @patrickvonplaten

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -245,11 +245,15 @@ def dispatch_model(
     check_device_map(model, device_map)
 
     if main_device is None:
-        main_device = [d for d in device_map.values() if d not in ["cpu", "disk"]][0]
+        if set(device_map.values()) == {"cpu"} or set(device_map.values()) == {"cpu", "disk"}:
+            main_device = "cpu"
+        else:
+            main_device = [d for d in device_map.values() if d not in ["cpu", "disk"]][0]
 
-    cpu_modules = [name for name, device in device_map.items() if device == "cpu"]
-    if state_dict is None and len(cpu_modules) > 0:
-        state_dict = extract_submodules_state_dict(model.state_dict(), cpu_modules)
+    if main_device != "cpu":
+        cpu_modules = [name for name, device in device_map.items() if device == "cpu"]
+        if state_dict is None and len(cpu_modules) > 0:
+            state_dict = extract_submodules_state_dict(model.state_dict(), cpu_modules)
 
     disk_modules = [name for name, device in device_map.items() if device == "disk"]
     if offload_dir is None and len(disk_modules) > 0:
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
LGTM! 🤗

pcuenca:
Is this expected to work on the `mps` device too? I've checked #596, but the only device I see here is `cpu`:
https://github.com/huggingface/accelerate/blob/0de16441266ba9a799aa338aa75c73bf078429e9/src/accelerate/utils/modeling.py#L464

Furthermore, `get_max_memory` doesn't seem to take it into account: https://github.com/huggingface/accelerate/blob/b816e258a95904736c8e3a5a2b14a18d5f407f4a/src/accelerate/utils/modeling.py#L275-L284

Alternatively, is it possible to override using a custom `device_map`?

Reference: https://github.com/huggingface/diffusers/pull/1042

sgugger:
No we don't support the "mps" device at all in big model inference. `device_map="auto"` will load the model on the CPU and then the user can't move them to the MPS device if they want.

pcuenca:
Ok understood, thanks! However, we get errors after moving to the MPS device, if we use `device_map="auto"`. I'm not sure what's causing it, but some tensors are being moved to the CPU when performing inference. I'll investigate further.

sgugger:
That's the side effect of using `device_map="auto"` @Narsil was warning you about. It does more than just load the model with empty weights but adds hooks to the model.

Narsil:
@pcuenca Loading weights efficiently is like 2 functions from `accelerate`.

https://github.com/huggingface/accelerate/blob/main/src/accelerate/big_modeling.py#L35-L110

And loading after the fact https://github.com/huggingface/transformers_bloom_parallel/blob/main/generate.py#L211-L270
(Couldn't fetch rapidly from `accelerate` but it has some more complete and robust function).

To be completely honest, I feel like all this should be in Pytorch proper. the `meta` device is really a bad workaround.

In the meantime maybe instead of pushing towards `device_map="auto"` you could instead use the 2 accelerate functions and wrapping them up nicely in `from_pretrained` ?

pcuenca:
> In the meantime maybe instead of pushing towards `device_map="auto"` you could instead use the 2 accelerate functions and wrapping them up nicely in `from_pretrained` ?

That's what I was thinking, yes! What I'm not sure about is whether to apply those only for the `mps` device (and keep using `device_map` in the others), or something else. I'll do some experiments to see what it looks like in practice. Thanks a lot!

Issue #501, opened 2022-07-09
Questions wrt training on TPU Pod


sumanthd17:
Hi Accelerate Team,

I'm looking to use `run_mlm_no_trainer.py` on TPU v3-128 pod. I have few questions before I want to get started with the process.

1. Can I stream the data directly GCP bucket, or should I have to download the data to the VM where I'm training?
2. Does accelerate library support training on TPU pod or is it limited to 8 cores? based on this #471 
3. Should I be using TPU node or TPU VM for better performance with accelerate library?
4. Is there a notebook or blog to help setup environment and run small tests on GCP VM for TPU training?
5. I want to train with dataset of order 1TB, will hf datasets be able to handle this data on a machine wth 256GB RAM? (possibly a Q for datasets repo, but just trying here as well)

Thanks in Advance

cc : @sgugger @muellerzr

sgugger:
Hi there!

1. You can do whatever you want since Accelerate will adapt to your training loop :-)
2. This is completely untested, so I can't guarantee it will work. Let us know if you run into any issues.
3. TPU VMs will probably have better performance (for the data pipeline)
4. Not that I know off
5. That's a question for the Datasets folks, you should go ask on the [forums](https://discuss.huggingface.co/) ;-)

sumanthd17:
Hi @sgugger 

I started trying this out an the first thing that popped up is that 
`ValueError: The number of devices must be either 1 or 8, got 128 instead`

How to get past this? This is probably be a lengthy thread with all the errors. I would be happy to document all the info once it's fixed

muellerzr:
@sumanthd17 can you provide the full error it gives for you there? (And nothing wrong with this being lengthy, usually these make great educational material eventually as well 😄 )

sumanthd17:
```
Traceback (most recent call last):
  File "/home/sumanth/bert/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/home/sumanth/bert/lib/python3.8/site-packages/accelerate/commands/accelerate_cli.py", line 43, in main
    args.func(args)
  File "/home/sumanth/bert/lib/python3.8/site-packages/accelerate/commands/launch.py", line 530, in launch_command
    tpu_launcher(args)
  File "/home/sumanth/bert/lib/python3.8/site-packages/accelerate/commands/launch.py", line 360, in tpu_launcher
    xmp.spawn(PrepareForLaunch(main_function), args=(), nprocs=args.num_processes)
[...]

PR #802, opened 2022-10-31
Introduce a pod-config command

muellerzr:
# Introduce a pod-config command to setup TPU pods

## What does this add?

This PR introduces a new command `accelerate pod-config` which will first ask the user for commands they wish to run across all TPU nodes in a GCP TPU cluster, before then running all of them in along with installing `accelerate` if specified

## Who is it for?

Part of https://github.com/huggingface/accelerate/issues/501

## Why is it needed?

When dealing with TPU pods, they need to have initial configurations across all of their processes with all the main data being available there and initial setups. This provides an easy tool to do so without having the user need to enter their own startup script, and can be run at any point in time.

### User-facing:

Adds a new `accelerate pod-config` command which can accept both a bash script of commands to run at once, or an unlimited amount of regular commands passed in

### Internal structure:

Configs now store both types of commands and users are prompted for these during `accelerate config`

## Basic Usage Example(s):

With an old `accelerate` config file:

```bash
accelerate pod-config --command "echo 'Hello World'" --tpu_zone "us-central1-a" --tpu_name "test-tpu"
```
With a new one fully configured:
```bash
accelerate pod-config
```
To install Accelerate in the new environment on all processes:
```bash
accelerate pod-config --install_accelerate
```

### DIFF
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -41,6 +41,8 @@ def get_cluster_input():
     main_process_port = None
     rdzv_backend = "static"
     same_network = True
+    tpu_name = None
+    tpu_zone = None
     if distributed_type in [DistributedType.MULTI_GPU, DistributedType.MULTI_CPU]:
         num_machines = _ask_field(
             "How many different machines will you use (use more than 1 for multi-node training)? [1]: ",
@@ -341,6 +343,23 @@ def get_cluster_input():
             "What is the name of the function in your script that should be launched in all parallel scripts? [main]: ",
             default="main",
         )
+        use_cluster = _ask_field(
+            "Are you using a TPU cluster? [yes/NO]: ",
+            _convert_yes_no_to_bool,
+            default=False,
+            error_message="Please enter yes or no.",
+        )
+        if use_cluster:
+            tpu_name = _ask_field(
+                "What is the name of your TPU cluster? ",
+                default=None,
+                error_message="Please enter the name of your TPU cluster.",
+            )
+            tpu_zone = _ask_field(
+                "What is the zone of your TPU cluster? ",
+                default=None,
+                error_message="Please enter the zone of your TPU cluster.",
+            )
     else:
         main_training_function = "main"
 
@@ -408,4 +427,6 @@ def get_cluster_input():
         use_cpu=use_cpu,
         rdzv_backend=rdzv_backend,
         same_network=same_network,
+        tpu_name=tpu_name,
+        tpu_zone=tpu_zone,
     )
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -365,7 +365,7 @@ def get_cluster_input():
                 error_message="Please enter the zone of your TPU cluster.",
             )
             run_commands = _ask_field(
-                "Do you have code you wish to be ran on startup in each pod? [yes/NO]: ",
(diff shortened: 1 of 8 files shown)


src/accelerate/commands/config/cluster.py
```diff
@@ -341,6 +347,50 @@ def get_cluster_input():
...
+                default=None,
+                error_message="Please enter the name of your TPU cluster.",
+            )
+            tpu_zone = _ask_field(
+                "What is the zone of your TPU cluster? ",
+                default=None,
+                error_message="Please enter the zone of your TPU cluster.",
+            )
+            run_commands = _ask_field(
+                "Do you have code you wish to be ran on startup in each pod? [yes/NO]: ",
```
pacman100, line 368:
```suggestion
                "Do you have code you wish to run on startup in each pod? [yes/NO]: ",
```

src/accelerate/commands/pod.py
```diff
@@ -0,0 +1,152 @@
...
+            args.tpu_zone = defaults.tpu_zone
+    if args.accelerate_version == "dev":
+        args.accelerate_version = "git+https://github.com/huggingface/accelerate.git"
+    elif args.accelerate_version == "latest":
+        args.accelerate_version = "accelerate -U"
+    elif isinstance(parse(args.accelerate_version), Version):
+        args.accelerate_version = f"accelerate=={args.accelerate_version}"
+
+    if not args.command_file and not args.command:
+        raise ValueError("You must specify either a command file or a command to run on the pod.")
```
pacman100, line 110: We can allow for mix and match by appending commands from the `args.command_file` to the `args.command`? Would it be better than raising error?

pacman100 (approved):
Thank you for making it easier to configure TPU pods 🤗. Just a nit and a comment, LGTM!

src/accelerate/commands/pod.py
```diff
@@ -0,0 +1,152 @@
...
+        args.tpu_name,
+        "--zone",
+        args.tpu_zone,
+        "--command",
+        args.command,
+        "--worker",
+        "all",
+    ]
+    if args.debug:
+        print(cmd)
```
sgugger, line 142:
```suggestion
        print(f"Running {' '.join(cmd)}")
```

tests/test_cli.py
```diff
@@ -63,3 +65,155 @@ def test_config_compatibility(self):
...
+    """
+    Test case for verifying the `accelerate pod-config` CLI passes the right `gcloud` command.
+    """
+
+    _tpu_name = "test-tpu"
+    _tpu_zone = "us-central1-a"
+    _command = "ls"
+    cmd = ["accelerate", "pod-config"]
+    _base_output = "cd /usr/share"
+    _command_file = "tests/test_samples/test_command_file.sh"
```
sgugger, line 80: That's a test class which is inherently private, thus we don't need to add all those `_` that hurt readability :-)

sgugger (approved):
Nice work, thanks a lot for spearheading this!

src/accelerate/commands/pod.py
muellerzr: With only allowing one vs the other it keeps the API simplistic, as otherwise we then have to worry about when in the order should the command come from in the CLI vs the bash script and that can be confusing to users. However in this particular case the check is just to ensure that you've passed *some* command to run in :)

muellerzr: We can keep this idea open though and if users want that we can enable it

Issue #801, opened 2022-10-31
TypeError: __init__() got an unexpected keyword argument 'differentiable'


comaniac:
### System Info

```Shell
- `Accelerate` version: 0.13.0.dev0 (commit 6dc429f6f748600c4319e438f6bdaa22261c4f6a)
- Platform: Linux-5.4.0-1075-aws-x86_64-with-glibc2.17
- Python version: 3.8.13
- Numpy version: 1.23.4
- PyTorch version (GPU?): 1.13.0+cu117 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - deepspeed_config: {}
        - distributed_type: FSDP
        - downcast_bf16: no
        - fsdp_config:
          - fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
          - fsdp_backward_prefetch_policy: NO_PREFETCH
          - fsdp_offload_params: false
          - fsdp_sharding_strategy: 1
          - fsdp_state_dict_type: FULL_STATE_DICT
          - fsdp_transformer_layer_cls_to_wrap: GPT2Block
        - machine_rank: 0
        - main_process_ip: null
        - main_process_port: null
        - main_training_function: main
        - mixed_precision: fp16
        - num_machines: 1
        - num_processes: 8
        - rdzv_backend: static
        - same_network: true
        - use_cpu: false
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

```
cd transformers; accelerate launch transformers/examples/pytorch/language-modeling/run_clm_no_trainer.py \
--dataset_name wikitext --dataset_config_name wikitext-2-raw-v1 --block_size 1024 --overwrite_cache \
--output_dir /tmp/gpt --model_name_or_path gpt2-medium --tokenizer_name gpt2-medium \
--per_device_train_batch_size 1 --max_train_steps 20 --learning_rate 2e-5 --gradient_accumulation_steps 1
```

### Expected behavior

```Shell
The example should run without issues. However, this is what I got:


  File "transformers/examples/pytorch/language-modeling/run_clm_no_trainer.py", line 676, in <module>
    main()
[...]

PR #803, opened 2022-10-31
Deal with optimizer.differentiable in PyTorch 1.13.0

comaniac:
Details are in #801.

cc @sgugger

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -652,8 +652,9 @@ class Accelerator:
                 except TypeError:
                     if "differentiable" in obj.optimizer.defaults:
                         # https://github.com/huggingface/accelerate/issues/801
-                        del obj.optimizer.defaults["differentiable"]
-                        optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                        defaults = obj.optimizer.defaults.copy()
+                        del defaults["differentiable"]
+                        optimizer = obj.optimizer.__class__(model.parameters(), **defaults)
                     else:
                         raise
                 obj = self.prepare_optimizer(optimizer)
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -647,7 +647,15 @@ class Accelerator:
                         "FSDP Warning: When using FSDP, several parameter groups will be conflated into "
                         "a single one due to nested module wrapping and parameter flattening."
                     )
-                optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                try:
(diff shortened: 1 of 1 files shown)


comaniac:
closing #801

src/accelerate/accelerator.py
```diff
@@ -647,7 +647,15 @@ def _prepare_fsdp(self, *args):
                         "FSDP Warning: When using FSDP, several parameter groups will be conflated into "
                         "a single one due to nested module wrapping and parameter flattening."
                     )
-                optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                try:
+                    optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                except TypeError:
+                    if "differentiable" in obj.optimizer.defaults:
+                        # https://github.com/huggingface/accelerate/issues/801
+                        del obj.optimizer.defaults["differentiable"]
```
sgugger, line 655: This will delete inplace, we should probably create a new dict here

sgugger (commented):
Thanks for working on this!

src/accelerate/accelerator.py
```diff
@@ -647,7 +647,15 @@ def _prepare_fsdp(self, *args):
...
+                        optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
```
comaniac, line 656:
```suggestion
                        defaults = obj.optimizer.defaults.copy()
                        del defaults["differentiable"]
                        optimizer = obj.optimizer.__class__(model.parameters(), **defaults)
```

src/accelerate/accelerator.py
comaniac: Good point. Done. PTAL.

src/accelerate/accelerator.py
```diff
@@ -647,7 +647,16 @@ def _prepare_fsdp(self, *args):
...
                         "a single one due to nested module wrapping and parameter flattening."
                     )
-                optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                try:
+                    optimizer = obj.optimizer.__class__(model.parameters(), **obj.optimizer.defaults)
+                except TypeError:
+                    if "differentiable" in obj.optimizer.defaults:
+                        # https://github.com/huggingface/accelerate/issues/801
+                        defaults = obj.optimizer.defaults.copy()
+                        del defaults["differentiable"]
```
sgugger, line 656:
```suggestion
                        defaults = {k: v for k, v in obj.optimizer.defaults.items() if k != "differentiable"}
```

sgugger (approved):
One last nit :-)

pacman100:
Thank you for the quick fix! 🤗

PR #808, opened 2022-11-01
Add compat checker

muellerzr:
This PR adds a compatibility checker that will run nightly looking to see if a new version of a core dependency was released and if so runs tests against it, while also reporting what that new library was to help us narrow down issues faster :) 

A report looks like [so](https://github.com/muellerzr/accelerate/issues/6) 
image

With a "Bug" added to the label. 

While there were not any issues in this release, this will help us in the future in case it or any other deps happen to release a breaking change and we can find out about it first and better trust we know exactly if/when there's an incompatibility issue :) (and if we know a release happened not at midnight UTC this action can be ran manually)

LMK what we think, it's okay if this is an "overengineering" thing and we don't worry about it :)

### DIFF
--- a/.github/workflows/run_upgrade_tests.yml
+++ b/.github/workflows/run_upgrade_tests.yml
@@ -2,6 +2,8 @@ name: New Dependency Versions and Compatibility
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
 
 env:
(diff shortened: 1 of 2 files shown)


muellerzr:
Closing as technically this isn't needed based on how tests are currently, but can be used for ideas later down the road on reporting if needed

PR #811, opened 2022-11-02
adding support to pickle and unpickle `AcceleratedOptimizer`

### LINKED ISSUE #805: Error when pickling accelerated optimizers with PyTorch 1.13
### System Info

```Shell
$ pip list
Package                  Version
------------------------ ----------
accelerate               0.13.2
numpy                    1.23.4
nvidia-cublas-cu11       11.10.3.66
nvidia-cuda-nvrtc-cu11   11.7.99
nvidia-cuda-runtime-cu11 11.7.99
nvidia-cudnn-cu11        8.5.0.96
packaging                21.3
pip                      22.3
psutil                   5.9.3
pyparsing                3.0.9
PyYAML                   6.0
setuptools               65.5.0
torch                    1.13.0
typing_extensions        4.4.0
wheel                    0.37.1
```

Same issue with torch-cpu.

OS: Ubuntu 20.04

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
import pickle
import accelerate
import torch

model = torch.nn.Linear(10, 10)
optimizer = torch.optim.SGD(model.parameters(), 0.1)
pickle.loads(pickle.dumps(optimizer))  # works

accelerator = accelerate.Accelerator(mixed_precision='fp16')
optimizer2 = accelerator.prepare(optimizer)
pickle.loads(pickle.dumps(optimizer2))  # fails
```

### Expected behavior

I ran into a problem with accelerate and PyTorch 1.13. It appears that optimizers cannot be pickled anymore after being accelerated. When running the attached script, I get:

```
  File ".../skorch/foo.py", line 12, in <module>
    pickle.loads(pickle.dumps(optimizer2))
  File ".../torch/optim/optimizer.py", line 84, in __setstate__
    self.defaults.setdefault('differentiable', False)
  File ".../accelerate/optimizer.py", line 90, in defaults
    return self.optimizer.defaults
AttributeError: 'AcceleratedOptimizer' object has no attribute 'optimizer'
```

The offending addition on PyTorch seems to be this line:

https://github.com/pytorch/pytorch/blob/23fe6c8ca15ec2cf6ea74f93aa91cae343ea534f/torch/optim/optimizer.py#L84

which was not present in PyTorch 1.12. At object creation time, PyTorch now tries to access the `defaults` attribute, which in turn calls the `defaults` property in accelerate, which requires the `optimizer` attribute, which doesn't exist and thus errors. At first glance, it looks like `AcceleratedOptimizer` might need its own `__getstate__` and `__setstate__` to solve this but I'm not sure.

pacman100:
### What does this PR do?
1. Fixes #805

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -157,3 +157,9 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
     def step_was_skipped(self):
         """Whether or not the optimizer step was skipped."""
         return self._is_overflow
+
+    def __getstate__(self):
+        return self.__dict__
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -159,7 +159,7 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
         return self._is_overflow
 
     def __getstate__(self):
-        return self.__dict__
+        return self.__dict__.copy()
 
     def __setstate__(self, state):
         self.__dict__.update(state)
--- a/tests/test_optimizer.py
+++ b/tests/test_optimizer.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import imp
 import pickle
 import unittest
 
 import torch
 
 from accelerate import Accelerator
+from accelerate.state import AcceleratorState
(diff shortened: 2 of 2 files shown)


sgugger (approved):
Nice! Can we also add a small regression test to make sure this doesn't get broken in the future?

sgugger:
Perfect, thanks!

PR #812, opened 2022-11-02
add `recurse` argument in `remove_hook_from_module`

younesbelkada:
# What does this PR do?

This PR adds `recurse` argument in `remove_hook_from_module` to be able to remove the hooks of a module recursively. Useful in the context of https://github.com/huggingface/optimum/pull/423 

cc @sgugger 
Thanks!

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -181,7 +181,7 @@ def remove_hook_from_module(module: nn.Module, recurse=False):
         module.forward = module._old_forward
         delattr(module, "_old_forward")
 
-    if recurse and len(list(module.children())) > 1:
+    if recurse:
         for child in module.children():
             remove_hook_from_module(child, recurse)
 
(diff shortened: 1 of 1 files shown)


src/accelerate/hooks.py
```diff
@@ -179,6 +181,10 @@ def remove_hook_from_module(module: nn.Module):
         module.forward = module._old_forward
         delattr(module, "_old_forward")
 
+    if recurse and len(list(module.children())) > 1:
```
sgugger, line 184: Not sure the check is useful (should be >0 but the loop will just be empty)
```suggestion
    if recurse:
```

src/accelerate/hooks.py
```diff
@@ -179,6 +181,10 @@ def remove_hook_from_module(module: nn.Module):
...
+        for _, child in module.named_children():
```
sgugger, line 185:
```suggestion
        for child in module.children():
```
No need to ask for the names if you don't want them ;-)

sgugger (commented):
Thanks for working on this!

src/accelerate/hooks.py
younesbelkada: Nice catch! Yes indeed ;)

younesbelkada:
Thanks for the suggestions! Can confirm the `BetterTransformers` tests pass on my side ;)

sgugger (approved):

PR #814, opened 2022-11-02
Update CLI docs and use mps rather than mps_device

muellerzr:
We have the following naming convention for device vs training schema:

Device: 
  - Use device name, such as `--tpu`, `--multi_gpu`, `--cpu`

Training Schema
  - Use schema name with "use" in the front, such as `--use_megatron_lm` and `--use_deepspeed`

Since MPS is a torch device rather than a schema, this PR changes the flag to be `--mps` and adds in a future warning for `--use_mps` so we can be consistent 🤗 

Also updates the CLI docs to reflect the latest refactorization.


pacman100 (approved):
Thanks! LGTM 🤗

Issue #471, opened 2022-06-24
Using Accelerate with TPU Pod VM like v3-32


huunguyen10:
Hi, thank you for great library.

I have just install accelerate on a TPU VM V3-32 but when I set number of TPU cores to 32 with `accelerate config` and run `accelerate test`, it throw an error:

 `ValueError: The number of devices must be either 1 or 8, got 32 instead`

So that mean accelerate haven't supported training on a TPU pod VM. Can you please add this feature to Accelerate?

By the way, I meet another problem, too. If I use `accelerate=0.9` with TPU VM `v2-alpha`, `accelerate test` run successfully. But if I use `accelerate=0.10` with `v2-alpha` or `tpu-vm-pt-1.11` or `tpu-vm-pt-1.10`, `accelerate test` can not finish runing, it just run forever. 

And when I run 
```
accelerate launch run_clm_no_trainer.py \
    --dataset_name wikitext \
    --dataset_config_name wikitext-2-raw-v1 \
    --model_name_or_path gpt2 \
    --output_dir /tmp/test-clm
```

it throw some errors (even `accelerate=0.9` with TPU VM `v2-alpha`).

```
06/24/2022 18:10:16 - INFO - run_clm_no_trainer - ***** Running training *****
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Num examples = 2318
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Num Epochs = 3
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Instantaneous batch size per device = 8
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Total train batch size (w. parallel, distributed & accumulation) = 64
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Gradient Accumulation steps = 1
06/24/2022 18:10:16 - INFO - run_clm_no_trainer -   Total optimization steps = 111
Grouping texts in chunks of 1024: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 37/37 [00:02<00:00, 16.44ba/s]
Grouping texts in chunks of 1024: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 37/37 [00:02<00:00, 16.31ba/s]
Grouping texts in chunks of 1024: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 37/37 [00:02<00:00, 16.66ba/s]
[...]

PR #815, opened 2022-11-02
Introduce TPU pod launcher

muellerzr:
This is a **heavy** POC actively in development and currently is awaiting on https://github.com/pytorch/xla/pull/4149 to see if we can push forward, however this PR is out here so that the community can know that it's being worked on and almost there :) 

## Proposed API:

`accelerate launch` now allows for a configured pod setup through three new params/config items:
- `use_cluster`, whether to use a TPU cluster
- `vm`, this mimics `xla_dist`'s `vm` argument and is a list of single compute VM names if you are not using an instance group. (generally not needed)
- `env`, this is a list of environment variables to set on each of the compute VM instances

Currently we only support non-Docker, as GCP doesn't support docker yet on the larger pods. 

To launch a script on a TPU pod, the API will look like such:

### Fully configured:
```bash
accelerate launch myscript.py --arg1 --arg2 ...
```
### No configuration:
```bash
accelerate launch --tpu --use_cluster myscript.py --arg1 --arg2 ...
```

## TODO:

Write tests

Closes https://github.com/huggingface/accelerate/issues/501 and closes https://github.com/huggingface/accelerate/issues/471

### DIFF
--- a/src/accelerate/commands/config/config.py
+++ b/src/accelerate/commands/config/config.py
@@ -21,13 +21,7 @@ from accelerate.utils import ComputeEnvironment
 
 from .cluster import get_cluster_input
 from .config_args import cache_dir, default_config_file, default_yaml_config_file, load_config_from_file  # noqa: F401
-from .config_utils import (  # noqa: F401
-    GroupedAction,
-    SubcommandHelpFormatter,
-    _ask_field,
-    _ask_options,
-    _convert_compute_environment,
-)
+from .config_utils import _ask_field, _ask_options, _convert_compute_environment  # noqa: F401
 from .sagemaker import get_sagemaker_input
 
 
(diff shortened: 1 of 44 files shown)


src/accelerate/commands/config/config.py
```diff
@@ -76,7 +76,7 @@ def config_command_parser(subparsers=None):
 
 def config_command(args):
     config = get_user_input()
-    if args.config_file is not None:
+    if args.__dict__["config_args.config_file"] is not None:
```
muellerzr, line 79: Note that this looks very ugly right now, it will be fixed by https://github.com/huggingface/accelerate/pull/853/files

sgugger (approved):
Very nice work, thanks a lot for working on this!

pacman100 (commented):
Great Work 🤗! Left a comment.

PR #818, opened 2022-11-03
Rename pod-config to tpu-config + docs

muellerzr:
Since it can be used on more than just TPU pods, renames the `pod-config` command to `tpu-config` based on a slack discussion. Also adds it to the CLI docs


sgugger (approved):
LGTM!

src/accelerate/commands/tpu.py
```diff
@@ -22,70 +22,66 @@
...
+    config_args.add_argument(
         "--config_file",
         type=str,
         default=None,
         help="Path to the config file to use for accelerate.",
     )
-
-    parser.add_argument(
-        "--pod_config_file",
-        type=str,
```
muellerzr, line 43: This arg was never used in the end, but not removed.

PR #820, opened 2022-11-04
Add `join_uneven_inputs` context manager to Accelerator

Chris-hughes10:
This PR adds a context manager which acts as a simple wrapper around [Join](https://pytorch.org/docs/stable/distributed.algorithms.join.html) to enable training with uneven inputs when using DDP.

This continues the work described in: https://github.com/huggingface/accelerate/issues/684

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -615,7 +615,7 @@ class Accelerator:
     @contextmanager
     def join_uneven_inputs(self, joinables, even_batches=None):
         """
-        TODO: docstring
+        A context manager that facilitates distributed training on uneven inputs, which acts as a wrapper around `torch.distributed.algorithms.join`.
         """
         if is_torch_version("<", "1.10.0"):
             raise ValueError(f"Joining uneven inputs requires PyTorch >= 1.10.0, You have {torch.__version__}.")
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -356,6 +356,7 @@ class Accelerator:
         self._optimizers = []
         self._models = []
         self._schedulers = []
+        self._dataloaders = []
         self._custom_objects = []
 
         # RNG Types
@@ -1136,7 +1137,7 @@ class Accelerator:
         """
         if device_placement is None:
             device_placement = self.device_placement if self.distributed_type != DistributedType.TPU else False
-        return prepare_data_loader(
+        prepared_data_loader = prepare_data_loader(
             data_loader,
             self.device,
             num_processes=self.num_processes,
@@ -1147,6 +1148,8 @@ class Accelerator:
             dispatch_batches=self.dispatch_batches,
             even_batches=self.even_batches,
         )
+        self._dataloaders.append(prepared_data_loader)
+        return prepared_data_loader
 
     def prepare_optimizer(self, optimizer: torch.optim.Optimizer, device_placement=None):
         """
@@ -1630,6 +1633,7 @@ class Accelerator:
         self._schedulers = []
         self._optimizers = []
         self._models = []
+        self._dataloaders = []
         self.deepspeed_engine_wrapped = None
         gc.collect()
         torch.cuda.empty_cache()
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -25,7 +25,7 @@ from typing import List, Optional, Union
 import torch
 
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
-from .data_loader import prepare_data_loader
+from .data_loader import DataLoaderDispatcher, prepare_data_loader
 from .logging import get_logger
 from .optimizer import AcceleratedOptimizer
 from .scheduler import AcceleratedScheduler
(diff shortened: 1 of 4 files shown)


Chris-hughes10:
@sgugger @muellerzr Is this the sort of thing that you had in mind? If so, I can write some docs and perhaps a small example.

One thing that I was thinking about whilst working on this though is that, although uneven batches is likely only going to be needed during evaluation, the approach that we have taken with setting a variable at the accelerator level is that it will apply to both training and validation. So an example would look something like:

```
accelerator = Accelerator(even_batches=True)

model, optimizer, train_dataloader, val_dataloader = create_components()

ddp_model, optimizer, train_dataloader = accelerator.prepare(model, optimizer, train_dataloader)

# training loop as normal
...


accelerator.even_batches = False

val_dataloader = accelerator.prepare(val_dataloader)

# run eval loop
with accelerator.join_uneven_inputs([ddp_model]):
    ...  
```
It's a bit clunky, but as it is an advanced use case, it may be acceptable.

I'm interested to hear any thoughts that you have!

muellerzr:
@Chris-hughes10 @sgugger what do you think about something like this. Here's an MVP example showing the basic behavior:

```python

class A:
    "A mock version of what we would call the dataloader"
    def __init__(self, a:int):
        self.a = a

class B:
    "A mock version of what we would call the Accelerator"
    @contextmanager
    def do_the_thing(self, a):
        old_a = a.a
        a.a = 2
        try:
            yield
        finally:
            a.a = old_a

a = A(1)
b = B()

with b.do_the_thing(a):
    print(a.a)
print(a.a)
```
You'd find that `a.a` prints 2 in the first and 1 in the second. We can apply that similarly here since we expect if the user is calling this that they will be training on uneven inputs, leading to an API such as:

```python
with accelerator.join_uneven_inputs([model], dataloader=eval_dataloader):
   ...
```
Inside this ctx manager `a.a` would be `eval_dataloader.even_batches`, and if it wasn't disabled originally we can print for the user that using this context manager will temporarily change this value or something. This let's us get rid of the explicit of setting in the accelerator and having to prepare later, so the only code change is the context manager. 

Lmk what you both think

sgugger (commented):
Mmm automatically change the value of `even_batches` inside the contextmanager might be a bit too magical so my first reaction would go against it. At the same time, if users wanted to activate it for evaluation and not training, this would make a nice API, so maybe it's a good idea still.

Would be interested to have others opinion on this proposed change!

muellerzr:
Agreed, maybe remove some of the magic with an `override` arg as well (that defaults to `True`) and a very explicit docstring as middle ground. (Also would love others opinions 😄 )

Chris-hughes10:
Hi both, I have given this some thought and have a couple of ideas. 

My first reaction was to agree with @sgugger, changing the `even_batches` value in the dataloader seems a little too magical, but it definitely makes the particular use case smoother. I think the main irk that I have with this is that it is changing a value on an unrelated object; if it was changing the value on the accelerator itself, I think it would be quite clean. Additionally, it seems a bit like a side effect, as it isn't really part of the same operation as the join.

One thought that I had, is that we could expose a method on the dataloader to make it easy to toggle this behaviour. Perhaps something like:
```
accelerator = Accelerator()

model, optimizer, train_dataloader, val_dataloader = create_components()

ddp_model, optimizer, train_dataloader, val_dataloader = accelerator.prepare(model, optimizer, train_dataloader, val_dataloader)

# training loop as normal
...


val_dataloader.ensure_even_batches(enabled=False)

# run eval loop
with accelerator.join_uneven_inputs([ddp_model]):
    ...  
```
There is still an extra step, but then it is very explicit what is going on. Conversely, this does mean that the flag in the accelerator is a little redundant. Also, it probably means that, in most cases, setting this flag and the context manager will have to be used together.

As an alternative, if we would like to keep everything accelerator-centric, we could potentially have the accelerator maintain a reference to any prepared dataloaders. This way, if we change the setting on the accelerator, we could change the value on all dataloaders. This could look something like:

```
accelerator = Accelerator(even_batches=True)

model, optimizer, train_dataloader, val_dataloader = create_components()

What do you guys think?

ddp_model, optimizer, train_dataloader, val_dataloader = accelerator.prepare(model, optimizer, train_dataloader, val_dataloader)

# training loop as normal
...

# run eval loop
with accelerator.join_uneven_inputs([ddp_model], even_batches=False):
    ...  
```
Personally, I think that this is quite clean, but it does mean adding additional logic to the accelerator, and perhaps the actual difference between this solution and the one proposed by @muellerzr is a bit academic.

pacman100:
Yes, I like the second approach proposed above ✨

muellerzr:
@Chris-hughes10 I too like that solution, as it keeps in line with how we design bits like this such as the gradient accumulation wrapper 😄 It also keeps the dataloader preparation the same as optimizer and model in the sense of pointers. This just means that there would also need to be a change to `accelerator.free_memory()` to also free the dataloader references. 

I also think it should just be an `Accelerator` only param, there wouldn't be a need to put this in `AcceleratorState` since we deem that too magical already 😃

Chris-hughes10:
It looks like there are failing tests that are unrelated to this PR. Perhaps https://github.com/huggingface/accelerate/pull/825 will solve the problem.

muellerzr:
@Chris-hughes10 correct, you can ignore the "checkpoint" and "rest" failures here.

src/accelerate/accelerator.py
```diff
@@ -608,6 +612,41 @@ def accumulate(self, model):
...
+            with contextlib.nullcontext(joinables):
+                yield
+
+        elif self.distributed_type == DistributedType.MULTI_GPU:
+            dl_even_batches_values = []
+
+            if even_batches is not None:
+                for dl in self._dataloaders:
+                    if not hasattr(dl, "even_batches"):
+                        raise ValueError("Overridding even_batches is not supported for iterable-style datasets")
```
sgugger, line 634: Maybe only throw an error if there are no dataloader with that attribute? I can imagine having an iterable dataset for training but a standard dataset for evaluation.

src/accelerate/accelerator.py
```diff
@@ -608,6 +612,41 @@ def accumulate(self, model):
...
+
+            enable_join = False if even_batches else True
+            try:
+                with Join(joinables, enable=enable_join, throw_on_early_termination=False):
+                    yield
+            finally:
+                for dl, even_batches_value in zip(self._dataloaders, dl_even_batches_values):
+                    dl.even_batches = even_batches_value
+        else:
+            raise ValueError("Joining uneven inputs is only supported for DistributedDataParallel training")
```
sgugger, line 648: I'd just issue a warning and exit.

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -103,11 +117,78 @@ def test_can_disable_even_batches():
...
+    with accelerator.join_uneven_inputs([ddp_model]):
+        for batch_idx, batch in enumerate(dl):
+            output = ddp_model(batch[0].float())
+            loss = output.sum()
+            loss.backward()
+            batch_idxs.append(batch_idx)
+
+    accelerator.wait_for_everyone()
+
+    if accelerator.process_index == 0:
```
sgugger, line 138: This test will need two processes, so probably needs some decorator/special launcher.

sgugger (commented):
Thanks for iterating on this!

src/accelerate/accelerator.py
Chris-hughes10: Good spot. Based on the other feedback, I have just raised a warning for this rather than an exception

src/accelerate/accelerator.py
Chris-hughes10: Done

Chris-hughes10:
I have made some more updates, so I think it should be feature complete now.

Unfortunately, I am getting a couple of failing tests on the CI, which look like they occur in the bf16 section of the test script, comparing training on CPU and GPU. I am unable to reproduce locally as I don't have a device with bf16 available, and the rest of the tests seem to pass. 

It is quite strange though, because looking at what it is failing, it doesn't really interact with anything in this PR other than the accelerator maintaining references to the dataloader; which doesn't seem to be a problem for the other training configurations in the script. After noticing that the accelerator state is cleared multiple times in the test - which does not include the dataloaders, as discussed above - I also tried adding resetting the accelerator manually before the bf16 section, using `accelerator.free_memory`, but this did not make any difference. 

Honesty, I am stumped at this point. Do you guys have any ideas on what could be the issue?

pacman100:
Hello @Chris-hughes10 , I went over the PR and tests. I found the cause of the tests failing. In the below screenshot please find the changes required in `test_accelerator.py` to resolve failing tests.

Screenshot 2022-11-10 at 7 28 32 PM

Chris-hughes10:
> Hello @Chris-hughes10 , I went over the PR and tests. I found the cause of the tests failing. In the below screenshot please find the changes required in `test_accelerator.py` to resolve failing tests.
> 
> Screenshot 2022-11-10 at 7 28 32 PM

Awesome, thanks for this! I didn't anticipate that the state would carry over outside of each test, but I suppose it makes sense for a singleton. I will make the change.

src/accelerate/accelerator.py
```diff
@@ -608,6 +612,91 @@ def accumulate(self, model):
...
+                    yield
+            finally:
+                # reset any batch samplers that have been modified
+                for dl_idx, even_batches_value in dl_even_batches_values:
+                    self._dataloaders[dl_idx].batch_sampler.even_batches = even_batches_value
+        else:
+            # Even when disabled, Join expects models to subclass Joinable, so skip entirely for single process runs
+            if self.distributed_type != DistributedType.NO:
+                warnings.warn(
+                    "Joining uneven inputs is only supported for DistributedDataParallel training, join_unenven_inputs has no effect."
```
muellerzr, line 694:
```suggestion
                    "Joining uneven inputs is only supported for multi-GPU training, as a result `join_uneven_inputs` will have no effect."
```

muellerzr (approved):
Great work! Left one small typofix inn the docs and then good to merge from my side! 😄

Chris-hughes10:
There seems to be a problem with building the PR documentation, is this something that I can fix?

muellerzr:
Thanks a bunch @Chris-hughes10, excellent work! I'll be keeping an eye on our CI that runs multi-gpu tests post this merge to see if anything is up and I'll let you know if there is and I can't seem to figure out what may be failing quickly :)

Chris-hughes10:
Thanks for helping to get this merged @muellerzr @sgugger @pacman100!

PR #821, opened 2022-11-06
Deepspeed example should use gather_for_metrics

HammadB:
I believe this example should be using gather_for_metrics here instead of gather to work correctly on distributed setups.

### DIFF
--- a/examples/by_feature/deepspeed_with_config_support.py
+++ b/examples/by_feature/deepspeed_with_config_support.py
@@ -288,7 +288,6 @@ def evaluate(args, model, eval_dataloader, accelerator, eval_dataset):
         losses.append(accelerator.gather_for_metrics(loss.repeat(args.per_device_eval_batch_size)))
 
     losses = torch.cat(losses)
-    losses = losses[: len(eval_dataset)]
     try:
         eval_loss = torch.mean(losses)
         perplexity = math.exp(eval_loss)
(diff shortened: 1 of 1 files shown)


examples/by_feature/deepspeed_with_config_support.py
```diff
@@ -285,7 +285,7 @@ def evaluate(args, model, eval_dataloader, accelerator, eval_dataset):
             outputs = model(**batch)
 
         loss = outputs.loss
-        losses.append(accelerator.gather(loss.repeat(args.per_device_eval_batch_size)))
+        losses.append(accelerator.gather_for_metrics(loss.repeat(args.per_device_eval_batch_size)))
 
     losses = torch.cat(losses)
     losses = losses[: len(eval_dataset)]
```
pacman100, line 291: This line becomes redundant when using `gather_for_metrics`

pacman100 (commented):
Hello, Thank you for the update! 🤗
Left a comment

examples/by_feature/deepspeed_with_config_support.py
HammadB: Ah yeah, makes sense thank you!

pacman100 (approved):
Thank you for updating the DeepSpeed example. LGTM! 🤗

sgugger (approved):
Thanks for fixing!

PR #823, opened 2022-11-07
Some final doc nits before release

muellerzr:
A few missing doc portions that needed updating prior to 0.14.0


sgugger (approved):
Thanks a lot!

PR #824, opened 2022-11-07
Rename command to be commands and allow for multiple commands to be entered during setup

muellerzr:
Small change to let multiple commands be passed during config


muellerzr:
@sgugger fails are unrelated, see https://github.com/huggingface/accelerate/pull/825

sgugger (approved):
LGTM, thanks for working on this!

PR #829, opened 2022-11-08
Add support for torch dynamo

sgugger:
This PR adds support for torch dynamo, which for now consists in wrapping the model in `dynamo.optimize`. So most of the PR is boilerplate to have everything setable with the config or args in the launcher.

Not adding any documentation yet as it's still very experimental and the underlying API might move a bit, this is mainly for internal use to be able to quickly benchmark things.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -165,7 +165,7 @@ class Accelerator:
             A list of `KwargHandler` to customize how the objects related to distributed training or mixed precision
             are created. See [kwargs](kwargs) for more information.
         dynamo_backend (`str` or `DynamoBackend`, *optional*, defaults to `"no"`):
-            Set to one of the possible dynamo backends to optimizer your training with torch dynamo.
+            Set to one of the possible dynamo backends to optimize your training with torch dynamo.
 
     **Available attributes:**
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -36,6 +36,7 @@ from .utils import (
     DeepSpeedPlugin,
     DistributedDataParallelKwargs,
     DistributedType,
+    DynamoBackend,
     FullyShardedDataParallelPlugin,
     GradScalerKwargs,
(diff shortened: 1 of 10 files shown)


muellerzr (approved):
Overall looks great to me! Nice work!

src/accelerate/accelerator.py
```diff
@@ -163,6 +164,8 @@ class Accelerator:
         kwargs_handlers (`List[KwargHandler]`, *optional*)
             A list of `KwargHandler` to customize how the objects related to distributed training or mixed precision
             are created. See [kwargs](kwargs) for more information.
+        dynamo_backend (`str` or `DynamoBackend`, *optional*, defaults to `"no"`):
+            Set to one of the possible dynamo backends to optimizer your training with torch dynamo.
```
pacman100, line 168:
```suggestion
            Set to one of the possible dynamo backends to optimize your training with torch dynamo.
```

src/accelerate/state.py
```diff
@@ -230,6 +236,9 @@ def __init__(
                 else:
                     self.device = torch.device("cuda")
                 self.mixed_precision = mixed_precision
+
+            if self.dynamo_backend != DynamoBackend.NO and self.mixed_precision != "no" and self.device.type == "cuda":
```
pacman100, line 240:
```suggestion
            if self.dynamo_backend != DynamoBackend.NO and self.mixed_precision == "no" and self.device.type == "cuda":
```

pacman100 (approved):
Hello! Thank you for all the work. Great feature addition which makes running efficient and faster training widely accessible and simple to use 🤗. Left 2 suggestions.

yuvalkirstain:
Hey, couldn't find any reference for the docs on Dynamo, will appreciate a pointer. Specifically, I wonder if it can be used alongside Deepspeed (I basically have a single-node setup, and would like to make the best out of my hardware).

sgugger:
I very much doubt it can be used with DeepSpeed yet. This is still experimental in PyTorch.

PR #830, opened 2022-11-09
Small questionairre CLI

muellerzr:
Built a small CLI questionnaire module to make choice selecting easier on the user. Based on the [bullet](https://github.com/bchao1/bullet/tree/master) library but only what's needed to get it working 😄 

Didn't manage to stay under 100 lines of code, but it is only ~250 lines so not bad 😄 

The API looks almost the exact same as `_ask_field`, except the second parameter is the options. 

I can also update our `y/n` to use this multiple choice schema if wanted


src/accelerate/commands/utils/__init__.py
```diff
@@ -0,0 +1,5 @@
+# flake8: noqa
```
sgugger, line 1: Let's name this submodule `menu` since it's about that?

src/accelerate/commands/utils/cursor.py
```diff
@@ -0,0 +1,64 @@
...
+from contextlib import contextmanager
+
+
+# Windows only
+if os.name == "nt":
+    import ctypes
+    import msvcrt  # noqa
+
+    class _CursorInfo(ctypes.Structure):
+        _fields_ = [("size", ctypes.c_int), ("visible", ctypes.c_byte)]
```
sgugger, line 30: You know how I feel about _ prefixing names.

src/accelerate/commands/utils/cursor.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2022 The HuggingFace Team. All rights reserved.
```
sgugger, line 1: If this is heavily inspired by another lib, it needs to be reflected in the copyright probably.

sgugger (commented):
Looks easy enough! Left a couple of comments.

src/accelerate/commands/utils/cursor.py
muellerzr: `_fields_` is very specific to `ctypes.Structure`, but will rename `CursorInfo`

sgugger: This and every function in this module pretty please!

src/accelerate/commands/config/cluster.py
```diff
@@ -227,12 +223,10 @@ def get_cluster_input():
...
-            sharding_strategy_query = "What should be your sharding strategy ("
-            for i, strategy in enumerate(FSDP_SHARDING_STRATEGY):
-                sharding_strategy_query += f"[{i+1}] {strategy}, "
-            sharding_strategy_query = sharding_strategy_query[:-2] + ")? [1]: "
-            fsdp_config["fsdp_sharding_strategy"] = _ask_field(
+            sharding_strategy_query = "What should be your sharding strategy?"
+            fsdp_config["fsdp_sharding_strategy"] = _ask_options(
                 sharding_strategy_query,
+                FSDP_SHARDING_STRATEGY,
                 lambda x: int(x),
```
pacman100, line 230:
```suggestion
                lambda x: int(x)+1,
```

pacman100: Because it's 1 based numbering

pacman100 (commented):
Thanks for working on this 🤗! This greatly improves the UX.  This currently doesn't support SageMaker config. Also left couple comments.

sgugger (approved):
Just tested it, feels very nice!

src/accelerate/commands/config/cluster.py
```diff
@@ -460,10 +463,9 @@ def get_cluster_input():
         if distributed_type == DistributedType.DEEPSPEED and use_deepspeed_config:
             mixed_precision = "no"
         else:
-            mixed_precision = _ask_field(
-                "Do you wish to use FP16 or BF16 (mixed precision)? [NO/fp16/bf16]: ",
-                lambda x: str(x).lower(),
-                default="no",
+            mixed_precision = _ask_options(
+                "Do you wish to use FP16 or BF16 (mixed precision)?",
+                ["no", "fp16", "bf16"],
```
sgugger, line 468: Missing the conversion function here.

muellerzr: Thanks! Fixed here and for sagemaker

src/accelerate/commands/config/cluster.py
```diff
@@ -284,23 +293,18 @@ def get_cluster_input():
                     lambda x: int(x),
                     default=1e8,
                 )
-            fsdp_backward_prefetch_query = "What should be your FSDP's backward prefetch policy ("
-            for i, backward_prefetch_policy in enumerate(FSDP_BACKWARD_PREFETCH):
-                fsdp_backward_prefetch_query += f"[{i}] {backward_prefetch_policy}, "
+            fsdp_backward_prefetch_query = "What should be your FSDP's backward prefetch policy?"
             fsdp_backward_prefetch_query = fsdp_backward_prefetch_query[:-2] + ")? [0]: "
```
pacman100, line 297: This line should have been removed

PR #839, opened 2022-11-10
Highlight selection with pretty colors

muellerzr:
Per request, highlights the selected choice in green like so:

image
image


sgugger (approved):
Pretty pretty!

Issue #835, opened 2022-11-10
"ValueError: Please make sure to properly initialize your"... when setting LOGLEVEL=DEBUG


0xdevalias:
### System Info

```Shell
- `Accelerate` version: 0.14.0
- Platform: Linux-5.4.0-124-generic-x86_64-with-debian-buster-sid
- Python version: 3.7.13
- Numpy version: 1.21.5
- PyTorch version (GPU?): 1.13.0+cu117 (True)
- `Accelerate` default config:
	Not found
```

```
Traceback (most recent call last):
  File "/workspace/diffusers/examples/dreambooth/train_dreambooth.py", line 689, in <module>
    main(args)
  File "/workspace/diffusers/examples/dreambooth/train_dreambooth.py", line 346, in main
    accelerator = Accelerator(
  File "/opt/conda/envs/dreambooth/lib/python3.10/site-packages/accelerate/accelerator.py", line 203, in __init__
    trackers = filter_trackers(log_with, self.logging_dir)
  File "/opt/conda/envs/dreambooth/lib/python3.10/site-packages/accelerate/tracking.py", line 580, in filter_trackers
    logger.debug(f"{log_with}")
  File "/opt/conda/envs/dreambooth/lib/python3.10/logging/__init__.py", line 1835, in debug
    self.log(DEBUG, msg, *args, **kwargs)
  File "/opt/conda/envs/dreambooth/lib/python3.10/site-packages/accelerate/logging.py", line 47, in log
    if self.isEnabledFor(level) and self._should_log(main_process_only):
  File "/opt/conda/envs/dreambooth/lib/python3.10/site-packages/accelerate/logging.py", line 32, in _should_log
    state = AcceleratorState()
  File "/opt/conda/envs/dreambooth/lib/python3.10/site-packages/accelerate/state.py", line 78, in __init__
    raise ValueError(
ValueError: Please make sure to properly initialize your accelerator via `accelerator = Accelerator()` before using any functionality from the `accelerate` library.
```
```


### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

In the `train_dreambooth.py` example from the diffusers repo:

[...]

Issue #834, opened 2022-11-10
Allow setting logger's LOGLEVEL via env var / CLI argument / etc


0xdevalias:
Currently (as best I can tell) there is no way to control accelerate's `logger` without modifying code. It would be nice if we were able to pass an ENV var / CLI argument to `accelerate` to change the log level without having to modify any code run by it.

- https://huggingface.co/docs/accelerate/main/en/package_reference/logging
- https://github.com/huggingface/accelerate/blob/main/src/accelerate/logging.py#L52

```
import logging
import os

LOGLEVEL = os.environ.get('LOGLEVEL', 'WARNING').upper()
logging.basicConfig(level=LOGLEVEL)
```

- https://docs.python.org/3/library/logging.html#logging.basicConfig

My  use case is wanting to crank up the logging on `train_dreambooth.py ([ref](https://github.com/huggingface/diffusers/tree/main/examples/dreambooth#training-with-prior-preservation-loss)) without having to modify the code itself:

```
from accelerate.logging import get_logger

LOGLEVEL = os.environ.get('LOGLEVEL', 'WARNING').upper()

logger = get_logger(__name__)
logger.basicConfig(level=LOGLEVEL)
```

```
AttributeError: 'MultiProcessAdapter' object has no attribute 'basicConfig'
```

0xdevalias:
Seemingly I can hack it in the code like this, but it would be nice if the `accelerate` logger was configurable directly, without code modification:

```
import logging
LOGLEVEL = os.environ.get('LOGLEVEL', 'WARNING').upper()
logging.basicConfig(level=LOGLEVEL)
```

sgugger:
cc @muellerzr

muellerzr:
@0xdevalias this is more of a user error but we can still improve it. In order for `logging.basicConfig` to work you **first need to define a logger**, and you specifically need to define *accelerate*'s logger. So e.g. the following would work:

```python
from accelerate.logging import get_logger
import logging

logger = get_logger(__name__)
logging.basicConfig("DEBUG")
```

That being said I'll introduce a PR shortly that let's you do:

```python
from accelerate.logging import get_logger

logger = get_logger(__name__, log_level="DEBUG")
```

muellerzr:
Thanks @0xdevalias, sorry I was a bit unclear in my last note so I removed it. This is a good point and the issue with `filter_trackers` and this should be solved in a PR shortly!

PR #842, opened 2022-11-10
Fix log error and add log level to get_logger

muellerzr:
Fixes https://github.com/huggingface/accelerate/issues/835
Fixes https://github.com/huggingface/accelerate/issues/834

The first comes from the fact that there is a `logger.debug` that could get called in `filter_trackers` before `AccelerateState()` was created, leading to the error mentioned. This PR moves it down in `__init__` to where it should be to allow for this to work. 

The second was solved by introducing a convenience `log_level` parameter to `accelerator.logging.get_logger()`

### DIFF
--- a/docs/source/package_reference/logging.mdx
+++ b/docs/source/package_reference/logging.mdx
@@ -23,7 +23,7 @@ To utilize this replace cases of `logging` with `accelerate.logging`:
 
 ## Setting the log level
 
-The log level can be set with the `LOG_LEVEL` environment variable or by passing 
+The log level can be set with the `ACCELERATE_LOG_LEVEL` environment variable or by passing 
 `log_level` to `get_logger`:
 ```python
 from accelerate.logging import get_logger
--- a/docs/source/package_reference/logging.mdx
+++ b/docs/source/package_reference/logging.mdx
@@ -21,4 +21,14 @@ To utilize this replace cases of `logging` with `accelerate.logging`:
 + logger = get_logger(__name__)
 ```
 
+## Setting the log level
+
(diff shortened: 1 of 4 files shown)


docs/source/package_reference/logging.mdx
```diff
@@ -21,4 +21,14 @@ To utilize this replace cases of `logging` with `accelerate.logging`:
 + logger = get_logger(__name__)
 ```
 
+## Setting the log level
+
+The log level can be set with the `LOG_LEVEL` environment variable or by passing
```
sgugger, line 26: Is this a standard name that comes from `logging`? Should we also look for an `ACCELERATE_LOG_LEVEL` if users want to use a different one from the standard one? (So priority would be to look for the accelerate log level, then the general one).

src/accelerate/logging.py
```diff
@@ -58,6 +59,8 @@ def get_logger(name: str):
     Args:
         name (`str`):
             The name for the logger, such as `__file__`
+        log_level (`str`, `optional`, defaults to `None`):
```
sgugger, line 62:
```suggestion
        log_level (`str`, *optional*):
```
Make sure to re-read the doc-writing guide ;-)

sgugger (approved):
Thanks for working on this! Just left a couple of comments.

docs/source/package_reference/logging.mdx
muellerzr: There is not, generally it's seen as `LOGLEVEL` for the recommended usage. However you're right having something Accelerate specific would be much better so I will opt for `ACCELERATE_LOG_LEVEL` instead

Issue #844, opened 2022-11-11
AttributeError: 'GradientState' object has no attribute 'samples_seen'


ernestchu:
https://github.com/huggingface/accelerate/blob/074d8d5a5a131501cc0a2a690cc14337395dcb09/examples/complete_cv_example.py#L244

`samples_seen` seems deprecated.

jdegange:
Has this been resolved or is there a workaround?

muellerzr:
Hi @ernestchu and @jdegange, thanks for pointing out this regression issue, apologies!

The solution is to have the evaluation loop look like so:
```python
        accurate = 0
        num_elems = 0
        for step, batch in enumerate(eval_dataloader):
            # We could avoid this line since we set the accelerator with `device_placement=True`.
            batch = {k: v.to(accelerator.device) for k, v in batch.items()}
            inputs = (batch["image"] - mean) / std
            with torch.no_grad():
                outputs = model(inputs)
            predictions = outputs.argmax(dim=-1)
            predictions, references = accelerator.gather_for_metrics((predictions, batch["label"]))
            accurate_preds = predictions == references
            num_elems += accurate_preds.shape[0]
            accurate += accurate_preds.long().sum()

        eval_metric = accurate.item() / num_elems
        # Use accelerator.print to print only on the main process.
        accelerator.print(f"epoch {epoch}: {100 * eval_metric:.2f}")
```
A PR to fix this will go up shortly, as the regular CV example has this but the complete one did not. Thanks for the flag!

PR #848, opened 2022-11-14
Fix complete_cv example regression

muellerzr:
Fixes https://github.com/huggingface/accelerate/issues/844

There was one regression in the examples that did not get trickled up between `cv_example` and `complete_cv_example` for calculating the metrics.


pacman100 (approved):
Thanks for fixing. LGTM!

Issue #843, opened 2022-11-11
The multi process lr scheduler implementation doesn't seem right


ernestchu:
https://github.com/huggingface/accelerate/blob/074d8d5a5a131501cc0a2a690cc14337395dcb09/src/accelerate/scheduler.py#L72-L75

The scheduler should not step if `self.scheduler._step_count > self.scheduler.total_steps`.

muellerzr:
Thanks for the notice! Will be fixed by #849

PR #849, opened 2022-11-14
Fix if/else in AcceleratedScheduler

muellerzr:
Fixes https://github.com/huggingface/accelerate/issues/843

The issue arises when using `OneCyclePolicy` and the `scheduler.step_count` > `scheduler.total_steps`. Current behavior is it will just move to the `else` statement and call `step()` regardless, which is not what we want. Instead this adds it in as a second if statement with no else.


sgugger (approved):
Thanks for fixing!

Issue #832, opened 2022-11-09
Feature Request: Update config files with a new `update_config` command


muellerzr:
With updates to the `config.yaml`, even though there are backwards compatibility tests I think a command such as `accelerate config update` would be a good option which would fill the config yaml with the basic options (defaulting to no for everything) that's expected.

PR #851, opened 2022-11-14
Refactor Accelerate config and introduce a multi-argument CLI interface

muellerzr:
## Motivation

As I'm focusing more on the CLI, eventually we wanted these three commands:

- `accelerate config`
- `accelerate default-config`
- `accelerate update-config` (from https://github.com/huggingface/accelerate/issues/832)

It did not quite make sense to have the CLI in this way, the actions should read start -> end. 

After doing a ton of digging I figured out that there's a neat way to add extra subparsers to existing subparsers by adding them as Argparsers directly. What does this mean? We can now do:

```bash
accelerate config -h
accelerate config default -h
```
And soon:
```bash
accelerate config update -h
```

## How does it work?

This operates by having a single empty parent parser to start. This one has `add_help=False` and doesn't do much explicitly there.

`parent_parser = argparse.ArgumentParser(add_help=False)`

From there we add the main `config` parser like we did before as a subparser of the `parent_parser`.
`config_parser = config_command_parser(subparsers)`

Finally where the real magic happens is the change to the *default_command_parser*. 

Instead of just adding it as a subparser, we instead create a new subparser first that will act as the "help" for the command, and add a full parser to it, linking it to that initial parser through `parents=[parent_parser]`:

```python
default_parser = parser.add_subparsers(title="subcommand {default}", dest="default", description=description)
parser = default_parser.add_parser("default", parents=parents)
```

The `title` here is so that during `accelerate config --help` it will show up as:

```bash
subcommand {default}:
  Create a default config file for Accelerate with only a few flags set.
```

The `dest` is what makes it easy for us to separate what comes from `accelerate config` vs `accelerate config default` etc. 

Lastly there is a `SubcommandHelpFormatter` class. This essentially makes it so that the subcommand is not repeated at the end, and since the user shouldn't be doing something like `accelerate config --config_file {} default` it will remove it from the usage directions.

Here is the full `--help` from `accelerate config`:

```bash
usage: accelerate config [-h] [--config_file CONFIG_FILE] {default} ...

Launches a series of prompts to create and save a `default_config.yaml` configuration file for your training system. Should always be ran first on your machine

optional arguments:
  --config_file CONFIG_FILE
                        The path to use to store the config file. Will default to a file named default_config.yaml in the cache
                        location, which is the content of the environment `HF_HOME` suffixed with 'accelerate', or if you don't have
                        such an environment variable, your cache directory ('~/.cache' or the content of `XDG_CACHE_HOME`) suffixed with
                        'huggingface'.

subcommand {default}:
  Create a default config file for Accelerate with only a few flags set.
```

## Other changes

To avoid a cyclical import error the `write_basic_config` function was moved from `utils` to `commands.config.default`. `from accelerate.utils import write_basic_config` still works however as it's linked. there. After this PR I will be working on the `update` part of this.

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -14,12 +14,10 @@
 
 import os
 from contextlib import contextmanager
-from pathlib import Path
 
 import torch
 
-from ..commands.config.cluster import ClusterConfig
-from ..commands.config.config_args import default_json_config_file
+from ..commands.config.default import write_basic_config  # noqa: F401
 from ..state import AcceleratorState
 from .dataclasses import DistributedType
 from .imports import is_deepspeed_available, is_tpu_available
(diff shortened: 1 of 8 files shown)


src/accelerate/utils/other.py
```diff
@@ -14,12 +14,10 @@
 
 import os
 from contextlib import contextmanager
-from pathlib import Path
 
 import torch
 
-from ..commands.config.cluster import ClusterConfig
-from ..commands.config.config_args import default_json_config_file
+from ..commands.config.default import write_basic_config  # noqa: F401
```
pacman100, line 20: is this required here anymore?

pacman100 (approved):
Thanks for working on this. LGTM 🤗! Couple suggestions and comments.

src/accelerate/utils/other.py
muellerzr: Yep, this is the nit about making sure we can keep `from accelerate.utils import write_basic_config`

sgugger (approved):
Thanks for working on this! It makes the CLI more natural.

PR #853, opened 2022-11-15
Clean up, add update command

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/832

Last of the CLI updates for now. This one refactors the current way that the CLI funcs are called to mimic `accelerate launch`. 

I also discovered with subcommands we just need to do `parser.add_parser`, we don't need to `add_subparsers` nor need as complex of a `SubcommandHelpFormatter`.

Now the help will just automatically show the subcommands in a list with their help:
```bash
usage: accelerate <command> [<args>] config [-h] [--config_file CONFIG_FILE] {default,update} ...

Launches a series of prompts to create and save a `default_config.yaml` configuration file for your training system. Should always be
ran first on your machine

optional arguments:
  --config_file CONFIG_FILE
                        The path to use to store the config file. Will default to a file named default_config.yaml in the cache
                        location, which is the content of the environment `HF_HOME` suffixed with 'accelerate', or if you don't have
                        such an environment variable, your cache directory ('~/.cache' or the content of `XDG_CACHE_HOME`) suffixed with
                        'huggingface'.

subcommands:
  {default,update}
    default             Create a default config file for Accelerate with only a few flags set.
    update              Update an existing config file with the latest defaults while maintaining the old configuration.
```

(Note that the command-specific help usage still looks the same, this was indeed needed)

### DIFF
--- a/docs/source/package_reference/cli.mdx
+++ b/docs/source/package_reference/cli.mdx
@@ -46,7 +46,7 @@ Create a default config file for Accelerate with only a few flags set.
 **Usage**: 
 
 ```bash
-accelerate default-config [arguments]
+accelerate config default [arguments]
 ```
 
 **Optional Arguments**:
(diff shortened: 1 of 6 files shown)


muellerzr:
The DeepSpeed CI failures are unrelated and look like a startup issue

sgugger (approved):
Not sure I understand everything, but LGTM!

PR #859, opened 2022-11-16
Switch default log to warn

muellerzr:
Was noticing a bit more logs than expected post the log change, this is because by default the logging library uses WARN, not INFO (as stated down here: https://docs.python.org/3/howto/logging.html#when-to-use-logging)


sgugger (approved):
Nice catch!

PR #869, opened 2022-11-18
fix dynamo_backend issue when launching using config file

pacman100:
### What does this PR do?
1. fix dynamo_backend issue when launching using config file

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -221,7 +221,9 @@ class Accelerator:
             )
             mixed_precision = "fp16"
 
-        if dynamo_backend is not None:
+        if dynamo_backend is None:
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -221,7 +221,9 @@ def __init__(
             )
             mixed_precision = "fp16"
 
-        if dynamo_backend is not None:
+        if dynamo_backend is None:
+            dynamo_backend = DynamoBackend(os.environ.get("DYNAMO_BACKEND", DynamoBackend.NO))
```
sgugger, line 225: This is done in the state [here](https://github.com/huggingface/accelerate/blob/8a96b0bfb856f37cf5013384454aaecc680ca23d/src/accelerate/state.py#L79) so I don't understand why this line is necessary.

PR #871, opened 2022-11-18
Even more log level refined, leave alone if not explicitly set

muellerzr:
Still not quite pleased with this implementation, learning more about the logging library + good practices. For example the transformers tests the output of `accelerator.state()` won't actually be logged unless the env var is set currently. 

So this PR makes us take a different approach of only filter *if we explicitly choose to filter* otherwise do nothing. This also limits it to the logger for the module, before it was setting the global log level which was leading to bloated traces


sgugger (approved):
Works for me, thanks!

Issue #866, opened 2022-11-17
[Include example code] mixed_precision="fp16" will break torch.save function.


BurguerJohn:
### System Info

```Shell
accelerate-0.14.0
Python 3.7.15
Pytorch 1.12.1+cu113
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```
from accelerate import Accelerator
import torch
import torch.nn as nn

class ExampleModule(torch.nn.Module):
  def __init__(self):
    super().__init__()
    self.conv = nn.Conv2d(3, 3, kernel_size=1)

model = ExampleModule()

#mixed_precision="fp16" will give error on torch.save
#mixed_precision="no" will work with torch.save
accelerator = Accelerator(
        gradient_accumulation_steps=1,
        mixed_precision="fp16",
        log_with="tensorboard",
        logging_dir=".",
    )

#Always work
torch.save(model,  "/model_original.model")

#Will break torch.save if the model if mixed_precision="fp16" 
model = accelerator.prepare(model)

#Error with mixed_precision="fp16" 
torch.save(model,  "/model_acc.model")
#Error as well with mixed_precision="fp16" 
torch.save(accelerator.unwrap_model(model),  "/model_unwrap.sd")
```

It will return this error if mixed_precision="fp16"

```
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
[<ipython-input-1-5ce45839c137>](https://localhost:8080/#) in <module>
     27 
     28 #Error
---> 29 torch.save(model,  "/model_acc.model")
     30 #Error as well
     31 torch.save(accelerator.unwrap_model(model),  "/model_unwrap.sd")

1 frames
[/usr/local/lib/python3.7/dist-packages/torch/serialization.py](https://localhost:8080/#) in _save(obj, zip_file, pickle_module, pickle_protocol)
    587     pickler = pickle_module.Pickler(data_buf, protocol=pickle_protocol)
    588     pickler.persistent_id = persistent_id
--> 589     pickler.dump(obj)
    590     data_value = data_buf.getvalue()
    591     zip_file.write_record('data.pkl', data_value, len(data_value))

[...]

PR #872, opened 2022-11-18
Solve pickling issues

muellerzr:
Fixes https://github.com/huggingface/accelerate/issues/866

## Proposal:

In order to fully unwrap the amount of chaining done in https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L928-L936 and support pickling we have to explore each and every single `__wrapped__` that exists inside `model.forward` until we get back to the original forward implementation and present that. 

To know just when we've hit the original version, the model should store the original function somewhere so a `==` check works and we don't just remove all of the decorators in existence (say if someone wrapped `torch.no_grad` on the forward func)

This then lets you perform:

```python
torch.save(accelerator.unwrap_model(model))
```

Also changes the `test_pickle` because technically that test always failed since it was testing the wrong thing. I'll be adding a test to check that we can unwrap fp16 as well

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -933,7 +933,7 @@ class Accelerator:
                 model.forward = torch.autocast(device_type=device_type, dtype=torch.bfloat16)(model.forward)
             else:
                 model.forward = torch.cuda.amp.autocast()(model.forward)
-            model.forward = partial(convert_outputs_to_fp32, model_forward=model.forward)
+            model.forward = convert_outputs_to_fp32(model.forward)
         if self.distributed_type == DistributedType.TPU and self.state.fork_launched:
             model = xmp.MpModelWrapper(model).to(self.device)
         return model
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -19,7 +19,7 @@ import os
 import sys
 import warnings
 from contextlib import contextmanager
-from functools import partial, wraps
+from functools import wraps
 from typing import List, Optional, Union
 
 import torch
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -933,7 +933,8 @@ class Accelerator:
                 model.forward = torch.autocast(device_type=device_type, dtype=torch.bfloat16)(model.forward)
             else:
                 model.forward = torch.cuda.amp.autocast()(model.forward)
-            model.forward = partial(convert_outputs_to_fp32, model_forward=model.forward)
+            model._original_forward = model.forward
+            model.forward = partial(convert_outputs_to_fp32, model_forward=model._original_forward)
         if self.distributed_type == DistributedType.TPU and self.state.fork_launched:
             model = xmp.MpModelWrapper(model).to(self.device)
         return model
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -19,7 +19,7 @@ import os
 import sys
 import warnings
 from contextlib import contextmanager
-from functools import wraps
+from functools import partial, wraps
 from typing import List, Optional, Union
 
 import torch
@@ -933,7 +933,7 @@ class Accelerator:
                 model.forward = torch.autocast(device_type=device_type, dtype=torch.bfloat16)(model.forward)
             else:
                 model.forward = torch.cuda.amp.autocast()(model.forward)
-            model.forward = convert_outputs_to_fp32(model.forward)
+            model.forward = partial(convert_outputs_to_fp32, model_forward=model.forward)
         if self.distributed_type == DistributedType.TPU and self.state.fork_launched:
             model = xmp.MpModelWrapper(model).to(self.device)
         return model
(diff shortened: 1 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -926,6 +926,7 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
             kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
             model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
         if self.native_amp:
+            model._original_forward = model.forward
```
muellerzr, line 929: @sgugger I know how you feel about `_` but I felt this was a decent time for one because the user really *shouldn't* need to know this exists at all or try to accidentally call it in any way

sgugger: Agree in this case :-)

src/accelerate/utils/operations.py
```diff
@@ -468,13 +468,11 @@ def _is_fp16_bf16_tensor(tensor):
...
 
 
-class ConvertOutputsToFp32:
+def convert_outputs_to_fp32(model_forward):
     """
     Decorator to apply to a function outputing tensors (like a model forward pass) that ensures the outputs in FP16
     precision will be convert back to FP32.
 
-    Use a class instead of a decorator because otherwise, the prepared model can no longer be pickled (issue #273).
-
```
sgugger, line 477: Are we sure this does not make a regression with this issue?

sgugger (approved):
Looks cleaner than before, thanks a lot! My only worry is to make sure the issue that created a class for this conversion is still solved by removing the class.

src/accelerate/utils/operations.py
muellerzr: The issue will always exist because it can never be pickled if we've wrapped it in a mixed precision decorator. So it will always have to go through `unwrap_model`. And this only ever comes up if we do mixed precision. See the nb shown in the bug report here: https://colab.research.google.com/drive/11fvrk1Jslw2VIRTF6h0pdGgJx5AkAIMv?usp=sharing

Where specifically our wrapping around `torch.cuda.amp` causes issues.

muellerzr: I can try to see if it's possible to inject a warning before trying to pickle/save it and tell the user they need to unwrap the model first.

src/accelerate/utils/other.py
```diff
@@ -53,8 +52,13 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = False):
...
     if not keep_fp32_wrapper:
         forward = getattr(model, "forward")
-        if isinstance(forward, ConvertOutputsToFp32):
-            setattr(model, "forward", forward.model_forward)
+        original_forward = model.__dict__.pop("_original_forward", None)
+        if original_forward is not None:
+            while hasattr(forward, "__wrapped__"):
+                forward = forward.__wrapped__
+                if forward == original_forward:
+                    break
```
sgugger, line 60: Let's put your comment on the PR here as well, to explain what this code is doing 🙏

muellerzr:
Some notes on the final version (which will also be trickled into the code of the PR):

- I kept this as a class so that we can implement a custom `__getstate__` which will return an explicit pickling error if someone tries to pickle a wrapped model with automatic mixed precision on it
- When a function is wrapped using functools (such that we have `torch.cuda.amp.autocast(dtype=torch.float16)(model.forward)` the wrapped function (the original `model.forward`) exists in the `__wrapped__` attribute. When using `functools.update_wrapper` (such as in `ConvertOutputsToFp32`) this will also in turn add a new `__wrapped__` function, allowing us to follow the chain of `__wrapped__` until we return to the original model forward function, which was stored away earlier.

src/accelerate/utils/operations.py
```diff
@@ -490,6 +488,11 @@ def __init__(self, model_forward):
     def __call__(self, *args, **kwargs):
         return convert_to_fp32(self.model_forward(*args, **kwargs))
 
+    def __getstate__(self):
+        raise pickle.PicklingError(
+            "Cannot pickle a prepared model with automatic mixed precision, please unwrap the model with `Accelerator.unwrap_model(model)` before pickling it."
+        )
+
```
muellerzr, line 495: A custom `__getstate__` func let's us raise an error when the user tries to prepare a model that had automatic mixed precision performed, ensuring with a clear error what the proper behavior needs to be, and the user doesn't have to know that this is due to PyTorch *and* accelerate and their behaviors

src/accelerate/utils/other.py
muellerzr, line 60: When a function is wrapped using functools (such that we have `torch.cuda.amp.autocast(dtype=torch.float16)(model.forward)` the wrapped function (the original `model.forward`) exists in the `__wrapped__` attribute. When using `functools.update_wrapper` (such as in `ConvertOutputsToFp32`) this will also in turn add a new `__wrapped__` function, allowing us to follow the chain of `__wrapped__` until we return to the original model forward function, which was stored away earlier.

sgugger (approved):

PR #873, opened 2022-11-18
Allow safetensors offload

sgugger:
This PR reworks a bit the internals of the offload hooks to allow for the case where the checkpoint is saved with `safetensors`. In this case, there is no need to have a separate save of the weight as a Numpy memory-mapped array, as we can directly access the weight using `safe_open`.

Goes along with https://github.com/huggingface/transformers/pull/20321

### DIFF
--- a/src/accelerate/utils/offload.py
+++ b/src/accelerate/utils/offload.py
@@ -20,6 +20,12 @@ from typing import Dict, List, Optional, Union
 import numpy as np
 import torch
 
+from ..logging import get_logger
+from .imports import is_safetensors_available
+
+
+logger = get_logger(__name__)
+
 
 def offload_weight(weight, weight_name, offload_folder, index=None):
     dtype = None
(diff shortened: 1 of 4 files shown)


src/accelerate/utils/offload.py
```diff
@@ -152,12 +153,25 @@ def __init__(
...
         self.all_keys.extend([key for key in self.index if key not in self.all_keys])
+        self.device = device
 
     def __getitem__(self, key: str):
         # State dict gets priority
         if key in self.state_dict:
             return self.state_dict[key]
         weight_info = self.index[key]
+        if weight_info.get("safetensors_file") is not None:
+            from safetensors import safe_open
```
muellerzr, line 164: We need a guard here for `is_safetensors_available` and raise an error if not :)

muellerzr (commented):
Looks easy enough! Just one comment to make sure we add a `safetensors` check and it should be added to the test requirements most likely 😄

src/accelerate/utils/offload.py
sgugger: Indeed, will clean that up on Monday!

LysandreJik (approved):
:fire:

PR #881, opened 2022-11-22
Update deprecated `logging.warn()`

SHi-ON:
This PR, replaces `logging.warn()` with `logging.warning()` as the former has been deprecated since Python 3.3.


sgugger (approved):
Thanks!

Issue #888, opened 2022-11-27
`DISABLE_RICH` => `ACCELERATE_DISABLE_RICH`


stas00:
## Feature Update

I'm working on adding `M4_DISABLE_RICH` and noticed Accelerate didn't add the `ACCELERATE_` prefix to the same functionality, and uses `DISABLE_RICH`.

IMHO, it'd be good if each project kept its env vars contained to its "namespace", so users of each project can turn on/off different flags w/o impacting other projects. I am aware that this in no way interferes with my work. I just thought that we could sync across HF projects at how we manage the env var name space if you're open to that, of course.

Incidentally this also helps when reading the output of `printenv` - `ACCELERATE_DISABLE_RICH` tells me that the flag disables RICH in Accelerate. `DISABLE_RICH` on the other hand is very ambiguous, especially since I don't think `rich` supports it (at least I couldn't find it in its docs).

Thank you for considering my proposition.

muellerzr:
Agreed, this is a very good point and *all* env variables I think should use `ACCELERATE` prefix so it's clear what's being set (unless specifically `torchrun` etc needs it). WDYT @sgugger @pacman100. We already did this with `ACCELERATE_LOG_LEVEL` so it would make sense to trickle this up and keep things consistent

sgugger:
Works for me!

stas00:
Thank you for the fix, @muellerzr! and syncing the rest of the env vars too - awesome!

PR #890, opened 2022-11-28
Prefix all accelerate env vars with ACCELERATE

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/888 and follows the current pattern for `LOG_LEVEL` by prefixing all env variables Accelerate uses with `ACCELERATOR`

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -119,9 +119,9 @@ class Accelerator:
             in your script multiplied by the number of processes.
         mixed_precision (`str`, *optional*):
             Whether or not to use mixed precision training (fp16 or bfloat16). Choose from 'no','fp16','bf16'. Will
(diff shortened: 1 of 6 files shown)


src/accelerate/utils/deprecate.py
```diff
@@ -0,0 +1,68 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import warnings
+
+
+def deprecate_value(old=None, new=None, version=None, warn=True, deprecate_str=None):
```
sgugger, line 19: This feels over-engineered, and outside the scope of this PR.

sgugger (approved):

PR #891, opened 2022-11-29
fix prefix issues in tests

pacman100:
### What does this PR do?
1. fix prefix issues in tests introduced #890


muellerzr (approved):
Thanks! Sneaky failure 🙃

sgugger (approved):
Thanks for the fixes!

PR #892, opened 2022-11-29
Support bfloat16 in load_offloaded_weight

sgugger:
Currently loading offloaded weights via `load_offloaded_weights` (as is done in Transformers in `from_pretrained` with `offload_state_dict=True`) does not work in bfloat16 because Numpy does not support that floating type.

This PR addresses that by reusing the function `load_offloaded_weight` which has the logic.


muellerzr (approved):
Excellent! Always fun finding out when code already works 😉

Issue #867, opened 2022-11-17
keymap.py import termios and tty, that work only on Unix


BurguerJohn:
### System Info

```Shell
accelerate-0.15.0.dev0
python 3.9.13
Windows 64 bit (AMD64)
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

On windows use:
`from accelerate import Accelerator`
Will return:
`ModuleNotFoundError: No module named 'termios'`

`import tty`
Will also try to import termios, giving the same error.

They are both called on keymap.py

https://docs.python.org/3/library/termios.html
https://docs.python.org/3/library/tty.html

### Expected behavior

```Shell
Should import the library using windows.
```

muellerzr:
Hi @BurguerJohn I don't have a way to check windows currently, hoping to have something setup this week. In the interim can you try building via: `pip install git+https://github.com/huggingface/accelerate@fix-windows-cli` to see if that works for you?

PR #893, opened 2022-11-29
Fix windows cli selector

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/867

Mild fix for right now, need to eventually see why windows just prints the history of arrow keys such as:
```bash
Do you wish to use FP16 or BF16 (mixed precision)?
Please select a choice using the arrow or number keys, and selecting with enter
 * no
    fp16
    bf16
[3A                                                                                                                    [1A                                                                                                                    [1A                                                                                                                    [1A                                                                                                                    [1A                                                                                                                    fp16
```

But the CLI works and the import issue has been resolved with this

Also windows in general doesn't support pretty colors, so adds an option to do `*` if on windows

### DIFF
--- a/src/accelerate/commands/menu/keymap.py
+++ b/src/accelerate/commands/menu/keymap.py
@@ -35,10 +35,27 @@ KEYMAP = {
     "mod_int": 91,
     "undefined": sys.maxsize,
     "interrupt": 3,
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for the fix!

Issue #870, opened 2022-11-18
Question: How to I use the kwargs handler to pass DDP params?


antopost:
I would like to set broacast_buffers=False when wrapping DDP.
The [documentation](https://huggingface.co/docs/accelerate/v0.2.0/kwargs.html) suggests using a kwargs handler class, but I have no idea how to pass it to Accelerator.

This is how I imagined it:

```
import copy
from dataclasses import dataclass


class KwargsHandler:

    def to_dict(self):
        return copy.deepcopy(self.__dict__)

    def to_kwargs(self):
        default_dict = self.__class__().to_dict()
        this_dict = self.to_dict()
        return {k: v for k, v in this_dict.items() if default_dict[k] != v}

@dataclass
class DistributedDataParallelKwargs(KwargsHandler):

    broadcast_buffers: bool = False

# Here I create an instance of KwargsHandler
kw = DistributedDataParallelKwargs()

# But passing it like this throws an error
accelerator = Accelerator(kwargs_handlers=kw)
```

Any help would be much appreciated!

sgugger:
You need to use the kwarg handles in the library:
```py
from accelerate import Accelerator
from accelerate.utils import DistributedDataParallelKwargs

accelerator = Accelerator(kwargs_handlers=[DistributedDataParallelKwargs(broadcast_buffers=False)])
```

antopost:
Many thanks, @sgugger !

muellerzr:
@antopost alright if I reopen this and flag docs so I can improve it? 😄

antopost:
You're very welcome to do so ^^

antopost:
@sgugger
Actually, when I try instantiate accelerator using more than one gpu, I get the following error:
```

  File "train.py", line 46, in __init__
    self.accelerator = Accelerator(gradient_accumulation_steps=self.gas,
  File "/opt/conda/lib/python3.8/site-packages/accelerate/accelerator.py", line 227, in __init__
    assert isinstance(handler, KwargsHandler), f"Unsupported kwargs handler passed: {handler}."
AssertionError: Unsupported kwargs handler passed: DistributedDataParallelKwargs(broadcast_buffers=False).
Traceback (most recent call last):
  File "train.py", line 645, in <module>
    main(config, output_dir, args)
  File "train.py", line 610, in main
    TA = TrainAgent(config, output_dir, args)
  File "train.py", line 46, in __init__
[...]

PR #894, opened 2022-11-29
Better description for improper kwargs

muellerzr:
Ties into https://github.com/huggingface/accelerate/issues/870, writes a more descriptive assert flag that states it must inherit `KwargHandler`


sgugger (approved):
Thanks!

PR #896, opened 2022-11-30
Adds a utility function to install correct version of torch XLA

### LINKED ISSUE #586: Feature Request: Install the right XLA wheels based on the torch version and environment
With the torch xla wheels now having a separate wheel directly for Google Colab, I think it'd be useful to have an xla installer as part of a utility in Accelerate. 

In my head I imagine it would look something like so:
1. Check if we are in Colab or not. (there are ways to do this, see the `notebook_launcher` code)
2. Grab the torch XLA wheel based on the current installed torch version
3. Install said wheel

An `upgrade=True` could be passed that will just install the latest torch and torch_xla wheels as well.

API wise it'd look something like so:

```python
# from the notebook
from accelerate.utils import install_xla
install_xla()
```

```bash
# from the CLI
accelerate install_xla
```

bipinKrishnan:
The PR adds a function ```install_xla``` to install the correct version of XLA wheels based on the version of torch installed on the system. The function also takes in a boolean argument ```upgrade``` which allow the user to install the latest available version of torch and XLA.

Closes #586

### DIFF
--- a/src/accelerate/utils/torch_xla.py
+++ b/src/accelerate/utils/torch_xla.py
@@ -35,7 +35,7 @@ def install_xla(upgrade: bool = False):
             subprocess.run(torch_install_cmd, check=True)
         # get the current version of torch
         torch_version = pkg_resources.get_distribution("torch").version
-        torch_version_trunc = torch_version[:torch_version.rindex(".")]
+        torch_version_trunc = torch_version[: torch_version.rindex(".")]
         xla_wheel = f"https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch_xla-{torch_version_trunc}-cp37-cp37m-linux_x86_64.whl"
         xla_install_cmd = ["pip", "install", xla_wheel]
         subprocess.run(xla_install_cmd, check=True)
--- a/src/accelerate/utils/torch_xla.py
+++ b/src/accelerate/utils/torch_xla.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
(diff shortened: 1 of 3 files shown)


muellerzr (approved):
*Excellent* work! Could we add this to the `utils` doc located in `docs/source/package_reference/utilities.mdx`? You can make a new section for it and follow the autodoc to ensure it gets there.

Lastly can you do `pip install -e .[quality]` and `make style; make quality;` from the root of the repo to pass the quality failure? Thanks!

src/accelerate/utils/torch_xla.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Args:
+        upgrade (`bool`, *optional*, defaults to `False`):
+            Whether to upgrade `torch` and install the latest `torch_xla` wheels.
+    """
+    in_colab = False
+    if "IPython" in sys.modules:
+        in_colab = "google.colab" in str(sys.modules["IPython"].get_ipython())
+
+    if in_colab:
```
sgugger, line 32: Is there a reason to limit this? Everything should work fine in any other env no? Or is it that the wheels can only work on Colab?

sgugger (commented):
Very nice addition! I just have a quick question on why it's restricted to colab.

src/accelerate/utils/torch_xla.py
muellerzr: The FR was for colab specifically because they have their own wheels for it https://github.com/huggingface/accelerate/issues/586

sgugger (approved):
Thanks for your contribution!

Issue #903, opened 2022-12-05
CLI should guard against bad input parameters


muellerzr:
Look into how we can guard against someone running a command such as:
```bash
accelerate launch --multi_gpu --gpu_ids 0,1,2,3 --mixed_precision no --num_machines 1 --num_processes 1 --num_cpu_threads_per_process=1
```
As `--multi_gpu` and `--num_processes 1` are mutually exclusive based on the fact that `--num_processes < 2` and `--num_machines < 2`

PR #905, opened 2022-12-06
Expand sanity checks

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/903 by adding more sanity checks to `launch_command` including more compute type options with the latest ones and checking if we are doing `multi_gpu` num processes > 2


sgugger (approved):
Thanks for the fixes!

PR #906, opened 2022-12-07
fix issue that amp bf16 does not work for cpu in env with cuda.

sywangyi:
and num_cpu_threads_per_process is not reset for better performance in cpu only case

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -365,7 +365,10 @@ class Accelerator:
             and self.distributed_type != DistributedType.FSDP
             and self.distributed_type != DistributedType.MEGATRON_LM
         ):
-            self.native_amp = is_bf16_available(True)
+            if self.device.type == "cpu":
+                self.native_amp = is_torch_version(">=", "1.10")
+            else:
+                self.native_amp = is_bf16_available(True)
             if mixed_precision == "bf16" and not self.native_amp and not is_tpu_available():
                 raise ValueError(err.format(mode="bf16", requirement="PyTorch >= 1.10 and a supported device."))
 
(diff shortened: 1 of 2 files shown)


sywangyi:
@sgugger @yao-matrix  @jiqing-feng  please notice the issue

src/accelerate/accelerator.py
```diff
@@ -365,7 +365,10 @@ def __init__(
             and self.distributed_type != DistributedType.FSDP
             and self.distributed_type != DistributedType.MEGATRON_LM
         ):
-            self.native_amp = is_bf16_available(True)
+            if self.device.type == "cpu":
+                self.native_amp = is_torch_version(">=", "1.10")
+            else:
+                self.native_amp = is_bf16_available(True)
```
muellerzr, line 371: I don't understand the need for this, `is_bf16_available` already does this check as shown here: https://github.com/huggingface/accelerate/blob/6ebaab71bbf87960583bc17339c9da9ec2e4ea91/src/accelerate/utils/imports.py#L89-L92

src/accelerate/accelerator.py
```diff
@@ -948,7 +951,7 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
             if self.mixed_precision == "fp16" and is_torch_version(">=", "1.10"):
                 model.forward = torch.cuda.amp.autocast(dtype=torch.float16)(model.forward)
             elif self.mixed_precision == "bf16" and self.distributed_type != DistributedType.TPU:
-                device_type = "cuda" if torch.cuda.is_available() else "cpu"
+                device_type = "cuda" if self.device.type != "cpu" and torch.cuda.is_available() else "cpu"
```
muellerzr, line 954: This line (and the next occurrence) can be removed entirely and just pass in `self.device.type` to `torch.autocast`

muellerzr (commented):
Thanks for the enhancement! Rather than having an ever-growing conditional, I think we should just pass `self.device.type` to `autocast` instead directly to avoid this, since it is already setup based on what was done in `accelerate.prepare`

src/accelerate/accelerator.py
sywangyi: good point. done

sywangyi: done

muellerzr (approved):

PR #909, opened 2022-12-09
fix accelerate test failure with cpu config

sywangyi:
*failure occurs when testing FP16
*autocast fail to work for cpu bf16 in some gpu+cpu platform, no need to use is_bf16_available logic. because native_amp already contains such logic.


sgugger (approved):
Thanks for the fix! @muellerzr can you also have a look?

muellerzr (approved):
Thanks! Makes sense to me

Issue #914, opened 2022-12-12
[Feature Request] Allow only saving the last `n` states


patrickvonplaten:
The checkpointing [utilities](https://huggingface.co/docs/accelerate/v0.15.0/en/usage_guides/checkpoint#checkpointing) of `accelerate` is really nice! 

 We're trying to fully adopt to `accelerate` to save and load intermediate checkpoints when training in `diffusers`. One feature request was to allow to only save the last "n" checkpoints and always overwrite the ones before. We have such functionality in `transformers`' Trainer here: https://huggingface.co/docs/transformers/v4.25.1/en/main_classes/trainer#transformers.Seq2SeqTrainingArguments.save_total_limit
This can greatly reduce the required hard drive space when training models.

Would it make sense to add such functionality directly to `accelerate` ?

PR #916, opened 2022-12-12
Introduce `project_dir` and limit the number of saved checkpoints

muellerzr:
# Introduce `project_dir` and limit the number of saved checkpoints

## What does this add?

- Makes `logging_dir` optional and centralizes anything written to be saved to a `project_dir`
- Allows for a set limit to the number of saved checkpoints
- Introduces a `checkpoints` folder when saving with `Accelerator.save_state`

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/914

## Why is it needed?

As more needs to customize directories with Accelerate arises, rather than having 2+ different folders a user can save things into it would be better to have a centralized "Project Directory" to work out of instead where the users checkpoints will always save to and loggers that require a directory will point to. 

### User-facing:

- A `SaveConfiguration` dataclass has been made which will store customizations when saving states through `Accelerator.save_state`
- `save_state` will no longer require a `save_path`, instead a `project_dir` can passed into the `Accelerator` object and it uses this.
- If combined with `automatic_checkpoint_naming=True` checkpoints are saved to `save_path/checkpoint_{number}` instead 
- `logging_dir` and `project_dir` when it comes to tracking will act the same if a `logging_dir` was not passed in
- A limit to the number of saved checkpoints can be performed and the oldest version (by name) will be removed through `Accelerator(save_total_limit={n})`

### Internal structure:

- `project_dir` when calling `save_state` will be used to create a new directory called `checkpoints` that checkpoints will be saved to. If a `project_dir` was not passed and instead `save_state` got a directory it will save to that new directory / `checkpoint_{n}`.
- The current number of checkpoints is saved in the `Accelerator` object. We could potentially save this in `AcceleratorState` instead, let me know if that is of interest.

## Basic Usage Example(s):

To fully utilize the new behavior: 

```diff
  accelerator = Accelerator(
-   logging_dir="some_dir",
+   project_dir="some_dir",
+   save_total_limit=3,
  )
```

The limitation of `save_state` should be used when lots of checkpoints are being performed and a limit should be set on just how many should be saved.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -214,7 +214,7 @@ class Accelerator:
         rng_types: Optional[List[Union[str, RNGType]]] = None,
         log_with: Optional[List[Union[str, LoggerType, GeneralTracker]]] = None,
         project_dir: Optional[Union[str, os.PathLike]] = None,
-        save_configuration: Optional[SaveConfiguration] = None,
+        save_config: Optional[SaveConfiguration] = None,
         logging_dir: Optional[Union[str, os.PathLike]] = None,
         dispatch_batches: Optional[bool] = None,
         even_batches: bool = True,
@@ -222,8 +222,8 @@ class Accelerator:
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
         dynamo_backend: Union[DynamoBackend, str] = None,
     ):
-        if save_configuration is not None:
-            self.save_configuration = save_configuration
+        if save_config is not None:
+            self.save_configuration = save_config
         else:
             self.save_configuration = SaveConfiguration(project_dir=project_dir)
         if self.project_dir is not None and logging_dir is None:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -16,6 +16,7 @@ import contextlib
 import gc
 import math
 import os
+import shutil
 import sys
 import warnings
 from contextlib import contextmanager
@@ -159,6 +160,9 @@ class Accelerator:
             also accept implementations of `GeneralTracker` for custom trackers, and can be combined with `"all"`.
         logging_dir (`str`, `os.PathLike`, *optional*):
             A path to a directory for storing logs of locally-compatible loggers.
+        save_dir (`str`, `os.PathLike`, *optional*):
+            A path to a directory for storing checkpoints. Each checkpoint will be saved in a subdirectory named
+            `checkpoint_<save_iteration>`.
         dispatch_batches (`bool`, *optional*):
             If set to `True`, the dataloader prepared by the Accelerator is only iterated through on the main process
             and then the batches are split and broadcast to each process. Will default to `True` for `DataLoader` whose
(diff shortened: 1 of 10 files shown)


muellerzr:
@Cyberes there is a deprecation cycle for `logging_dir` of two full versions and when it comes to trackers `project_dir` acts the exact same as the `logging_dir` parameter, so for your particular setup if I understand correctly would just need to change `logging_dir` to `project_dir` and nothing else.  

Is this reasonable enough for you? Or is there something more specific/drastic I've misunderstood.

muellerzr:
Checkpoints are saved to `project_dir/checkpoints`, logs are saved to `project_dir/[name]`

Cyberes:
Yeah, I'd like to have the option to put them in completely separate directories. For example, save the checkpoint states to `project_dir/checkpoints` and TB logs to `/mnt/tensorboard`.

If you're against the seperate directories thing I could always override `TensorBoardTracker`, I guess.

muellerzr:
That use case makes sense, will do as you suggested and override only if not passed. Thanks @Cyberes :)

sgugger (approved):
LGTM, but let's bump all deprecations to 0.18.0 (we will release 0.16.0 as next version, so don't want to remove everything directly at 0.17.0). Thanks!

Cyberes:
Sweeeeeeeeeet, thanks @muellerzr.

src/accelerate/accelerator.py
```diff
@@ -158,7 +159,16 @@ class Accelerator:
...
+            A path to a directory for storing logs of locally-compatible loggers. If not passed will save in
+            `project_dir` by default.
+        project_dir (`str`, `os.PathLike`, *optional*):
+            A path to a directory for storing data such as logs of locally-compatible loggers and potentially saved
+            checkpoints.
+        automatic_checkpoint_naming (`bool`, *optional*, defaults to `False`):
+            Whether saved states should be stored in `project_location` and be automatically iteratively named.
+        save_total_limit (`int`, *optional*):
+            The maximum number of checkpoints to keep if performing `automatic_checkpoint_naming`, will default to all
+            of them.
```
sgugger, line 171: This is starting to make too many arguments.

muellerzr: Fixed with a `SaveConfiguration` class

src/accelerate/accelerator.py
```diff
@@ -205,13 +213,23 @@ def __init__(
...
         kwargs_handlers: Optional[List[KwargsHandler]] = None,
         dynamo_backend: Union[DynamoBackend, str] = None,
     ):
+        if save_config is not None:
+            self.save_configuration = save_config
+        else:
+            self.save_configuration = SaveConfiguration()
+        if project_dir is not None and self.project_dir is None:
+            self.save_configuration.project_dir = project_dir
+        if self.project_dir is not None and logging_dir is None:
```
sgugger, line 231: Let's also deprecate logging_dir to include it in the SaveConfig? Wdyt?

sgugger (approved):

src/accelerate/accelerator.py
muellerzr: I was debating on that, if we did so I think it'd be better to name it `ProjectConfiguration` instead since it's all relative to the total "project" at that point. But can make adjustments for it. The only "con" I see is given the conversation above for the user the code would change, but the end-result would not so I think it'd be fine

pacman100 (approved):
Great enhancement! LGTM! 🤗

PR #920, opened 2022-12-13
Honor model dtype in `load_checkpoint`

sgugger:
This PR fixes a standing bug where we have a different behavior than PyTorch. In torch, loading a `state_dict` inside a model will never change the model's dtype:

```py
import torch

model = torch.nn.Linear(5, 6)
model_in_half = model.half()
state_dict_in_half = model.state_dict()

model = torch.nn.Linear(5, 6).to("meta")
model.load_state_dict(state_dict_in_half)
model.weight.dtype
# return torch.float32
```

Currently in Accelerate, `load_checkpoint` does the opposite and when loading a model, it converts it to the dtype of the state dict. This PR addresses that.

This PR only contains the fix for now, we have to discuss how to maybe maintain backward compatibility (even if this is a bug fix), because `diffusers` might be relying on this behavior, cc @patrickvonplaten

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -85,7 +85,11 @@ def dtype_byte_size(dtype: torch.dtype):
 
 
 def set_module_tensor_to_device(
-    module: nn.Module, tensor_name: str, device: Union[int, str, torch.device], value: Optional[torch.Tensor] = None
+    module: nn.Module,
+    tensor_name: str,
+    device: Union[int, str, torch.device],
+    value: Optional[torch.Tensor] = None,
+    dtype: Optional[Union[str, torch.dtype]] = None,
 ):
     """
     A helper function to set a given tensor (parameter of buffer) of a module on a specific device (note that doing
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks, that makes sense. I don't particularly think there is a "harm" in silently pushing it out (i.e. don't advertise the bad behavior but let it still pass) in this particular case. If we do care about phasing that out perhaps leave it for a 1.0.0? (Similar to some optimizer bits we have)

patrickvonplaten:
Actually before merging, should be maybe not better handle this in `set_module_tensor_to_device` ? E.g. add a `dtype` argument to the function there? This would be easier for `diffusers` to by in line with `accelerate` I think - see: https://github.com/huggingface/diffusers/blob/727434c206f6c22b746e460293035a1324f0bc13/src/diffusers/modeling_utils.py#L491

src/accelerate/utils/modeling.py
```diff
@@ -680,8 +694,7 @@ def load_checkpoint_in_model(
         else:
             for param_name, param in checkpoint.items():
                 module_name = param_name
-                if dtype is not None and not str(param.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
```
sgugger, line 683: This is moved to `set_module_tensor_to_device`.

patrickvonplaten (approved):
Thanks a lot for adapting!

PR #926, opened 2022-12-16
Support `init_on_device`

thomasw21:
Similarly to `init_empty_weights`, we support a context manager defaulting all constructors to a specified device:

```python
with init_on_device(torch.device("cuda")):
   model = MyModel() # Directly loaded in the device we want
```

Additionally we now suppport `require_mps` that allows one to start testing on other backends provided by `torch`


src/accelerate/big_modeling.py
```diff
@@ -62,6 +62,32 @@ def init_empty_weights(include_buffers: bool = False):
...
+
+    Example:
+
+    ```python
+    import torch.nn as nn
+    from accelerate import init_empty_weights
+
+    # Initialize a model with 100 billions parameters in no time and without using any RAM.
+    with init_on_device(device=torch.device("cuda")):
+        tst = nn.Liner(100, 100)  # on `cuda` device
```
sgugger, line 88: Ah ah, that example needs adapting ;-p

sgugger (approved):
Nice and clean, thanks a lot for the added tests as well! There is just a copy-paste to fix in a docstring.

thomasw21:
I guess we can also extend to specific dtypes as well now.

sgugger:
Which dtype did you have in mind? For floating types there is a default you can set in PyTorch directly.

thomasw21:
You mean https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html ? Does fp16/bf16 also work? If so my bad.

sgugger:
Yes it works pretty well. We use it in Transformers [here](https://github.com/huggingface/transformers/blob/17292440c069118fbdb992b9a17da2098fab5b87/src/transformers/modeling_utils.py#L1109) for instance.

PR #935, opened 2022-12-21
add `mixed_precision_type` property to `AcceleratorState`

pacman100:
### What does this PR do?
1. add `mixed_precision` property to `AcceleratorState`.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -473,7 +473,7 @@ class Accelerator:
 
     @property
     def mixed_precision(self):
-        return self.state.mixed_precision_type
+        return self.state.mixed_precision
 
     def on_main_process(func):
         """
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -473,17 +473,7 @@ class Accelerator:
 
     @property
     def mixed_precision(self):
-        if self.distributed_type == DistributedType.DEEPSPEED:
-            config = self.state.deepspeed_plugin.deepspeed_config
-            if config.get("fp16", {}).get("enabled", False):
-                mixed_precision = "fp16"
-            elif config.get("bf16", {}).get("enabled", False):
-                mixed_precision = "bf16"
-            else:
-                mixed_precision = "no"
-        else:
-            mixed_precision = self.state.mixed_precision
-        return mixed_precision
+        return self.state.mixed_precision_type
 
     def on_main_process(func):
         """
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
Thanks! Let's also add it to the state's `repr` as well please :)

sgugger (commented):
This is not ideal in terms of naming. I'm wondering if we shouldn't rename the current attribute `_mixed_precision` and have the property named `mixed_precision` which would be clearer. There might be a bit of code to adapt, but I think it's worth the effort, no?

Thanks for working on this in any case!

pacman100:
> This is not ideal in terms of naming. I'm wondering if we shouldn't rename the current attribute `_mixed_precision` and have the property named `mixed_precision` which would be clearer. There might be a bit of code to adapt, but I think it's worth the effort, no?

Yes, I thought of that but changed my mind in order to be least invasive in terms of the changes required. Addressed this in latest commit. Thank you.

pacman100:
Example with recent changes:

Code:
```
from accelerate.state import AcceleratorState

def main():
    accelerator = Accelerator()
    accelerator.print(f"{AcceleratorState()}")
    accelerator.print(f"{AcceleratorState().mixed_precision=}")
    accelerator.print(f"{accelerator.mixed_precision=}")
```

config:
```yaml
command_file: null
commands: null
deepspeed_config:
  gradient_clipping: 1.0
  offload_optimizer_device: 'cpu'
  offload_param_device: 'cpu'
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
distributed_type: DEEPSPEED
downcast_bf16: 'no'
dynamo_backend: 'NO'
fsdp_config: {}
gpu_ids: null
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'bf16'
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_name: null
tpu_zone: null
use_cpu: false
```

Output
```bash
Distributed environment: DEEPSPEED  Backend: nccl
Num processes: 2
Process index: 0
Local process index: 0
Device: cuda:0
Mixed precision type: bf16
ds_config: {'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 1, 'zero_optimization': {'stage': 3, 'offload_optimizer': {'device': 'cpu'}, 'offload_param': {'device': 'cpu'}, 'stage3_gather_16bit_weights_on_model_save': True}, 'gradient_clipping': 1.0, 'steps_per_print': inf, 'bf16': {'enabled': True}}

AcceleratorState().mixed_precision='bf16'
accelerator.mixed_precision='bf16'
```

sgugger (approved):
Much better, thanks a lot!

PR #937, opened 2022-12-22
fix batch size in prepare_dataloader for iterable datasets

sanderland:
There are a lot of cases, so I may be missing something, but I believe this is a bug.
I am setting batch_size=16 with num_processes=2 and using an IterableDataset.
My collator is receiving batches of size 8, and the dataloader is yielding batches of size 4 on each process.

The DataLoaderDispatcher has batch size 8 due to this line, which further gets split later on to 4.
This PR works in my case, but I am not sure it does not break another.


sgugger:
Looks like the right fix to me, thanks!

sgugger (approved):

Issue #936, opened 2022-12-21
[feature] assert on ambiguity config when using deepspeed


stas00:
OK, so please consider this setup:

```
[...]
deepspeed_config:
  deepspeed_multinode_launcher: standard
  deepspeed_config_file: ds_config.json
  gradient_accumulation_steps: 2
  zero3_init_flag: true
  zero_stage: 2
  zero3_save_16bit_model: true
distributed_type: DEEPSPEED
[...]
```
and `ds_config.json`:
```
{
    "fp16": {
        "enabled": true
    },
    "zero_optimization": {
        "stage": 3,
        "stage3_gather_16bit_weights_on_model_save": false,
        "offload_optimizer": {
            "device": "none"
        },
        "offload_param": {
            "device": "none"
        }
    },
    "gradient_clipping": 1.0,
    "train_batch_size": "auto",
    "train_micro_batch_size_per_gpu": "auto",
    "gradient_accumulation_steps": 10,
    "steps_per_print": 2000000
}
```

as you can see we have a potential difficult situation here as there are the same configs defined in multiple places. e.g. `zero3_save_16bit_model`, `gradient_accumulation_steps`, `zero_stage`, etc. with values that don't match! And the user isn't the wiser which of them prevail.

Here is my proposal: 

If `deepspeed_config_file` is provided, Accelerate should assert on any other configs that will collide with it. And unless I'm missing some details, I think the only valid (unique) entries of `deepspeed_config` that aren't in the DS config file are:
```
deepspeed_config:
  deepspeed_multinode_launcher: standard
  deepspeed_config_file: ds_config.json
  zero3_init_flag: true
distributed_type: DEEPSPEED
```

and any other deepspeed_config entries create ambiguity and should be asserted on.

And of course if `deepspeed_config_file` is not provided, then of course all those other options are a must.

OK, but then we have the `auto` feature which is useful when things are set elsewhere. So the categorical proposition I made above is probably incorrect. Yet, as you can see from the test case I presented - there is a massive mismatch between the 2 config files and accelerate doesn't assert on that at all. This situation makes debugging extremely painful since I can never be sure if I changed the right config and I'm testing the right combination. And yes, I can look at the logs, but this is not conducive for easy debug.

for context: I spent several days debugging a bad memory leak in m4, triggered by `zero.Init` and I had to switch multiple flags on/off while doing the debug many times again and again.

Thank you!

sgugger:
cc @pacman100

[...]

Issue #922, opened 2022-12-14
[feature request] `zero3_init_flag` should be `True` by default for stage 3


stas00:
This feature request is about deepspeed's plugin's `zero3_init_flag` here (and a few related issues):

https://github.com/huggingface/accelerate/blob/7889ba6b6dd094e72fa222194146a40426fb3d5a/src/accelerate/utils/dataclasses.py#L381-L387

1. IMHO, this flag should be `True` by default as zero stage 3 is for large models, it's very unlikely the user will be able to load those models w/o `zero.Init`, so why not help users to just have one less thing to figure out.

   I realize that one may want not to use `deepspeed.zero.Init` to save the hassle of adding ` deepspeed.zero.GatheredParameters` during `init_weights` and some other early setup stage in some special models so that the model remains unsharded till the last moment. But this is very rare situation. And I guess in those cases a user may want to turn it off sometimes, but it's unlikely to work for any largish model.

2. Also there is a documentation ambiguity: what does: `default=None,` mean when defining the default behavior in the docs? This is super confusing - is it Off by default? after reading the code I see later it's set to `False` but how is the user to know that? The doc here says `None`:
https://huggingface.co/docs/accelerate/package_reference/deepspeed#accelerate.DeepSpeedPlugin

   The relevant code that I needed to decipher to find out the default is here:

https://github.com/huggingface/accelerate/blob/7889ba6b6dd094e72fa222194146a40426fb3d5a/src/accelerate/utils/dataclasses.py#L452-L456

3. same ambiguity applies to  `zero3_save_16bit_model`  (says `None`) as well - the user has no way of telling what the default is.

4. (already going into possibly another Issue: all the Accelerate `DEEPSPEED_*` env vars probably should be `ACCELERATE_DEEPSPEED_*` env vars as in a recent rename of all Accelerate env vars, but these got missed) - please let me know if you prefer a separate issue about it.

Thank you.

stas00:
btw, actually, I wanted to also say thanks for making the `zero.Init` functionality configurable, as I'm using it right now while debugging z3 with m4 which doesn't yet quite work due to nested `zero.Init` calls (via nested `from_pretrained`).

In the HF Trainer it's always on for Z3.

stas00:
Thank you for the PR, @pacman100 

How does one `accelerate config` command address many different projects with various needs? Unless one runs this for each project but then the need can change through the project as well, so I'm not sure how that would substitute clear documentation.

[...]

PR #941, opened 2022-12-23
raise error for duplicate accelerate config  values when using `deepspeed_config_file`

pacman100:
### What dos this PR do?
1. Fixes: #936 

Example:
1. accelerate config manually tweaked to have both `deepspeed_config_file` and other ds config entries that are available in the json config file:
```yaml
command_file: null
commands: null
deepspeed_config:
  gradient_clipping: 1.0
  offload_optimizer_device: 'cpu'
  offload_param_device: 'cpu'
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
  deepspeed_config_file: 'ds_config.json'
distributed_type: DEEPSPEED
downcast_bf16: 'no'
dynamo_backend: 'NO'
fsdp_config: {}
gpu_ids: null
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'bf16'
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_name: null
tpu_zone: null
use_cpu: false
```

2. ds_config.json:
```json
{
    "fp16": {
        "enabled": true
    },
    "zero_optimization": {
        "stage": 3,
        "stage3_gather_16bit_weights_on_model_save": false,
        "offload_optimizer": {
            "device": "none"
        },
        "offload_param": {
            "device": "none"
        }
    },
    "gradient_clipping": 1.0,
    "train_batch_size": "auto",
    "train_micro_batch_size_per_gpu": "auto",
    "gradient_accumulation_steps": 10,
    "steps_per_print": 2000000
}
```

3. Code:
```python

def main():
    accelerator = Accelerator()

if __name__ == "__main__":
    main()
```

4. output:
```bash
ValueError: When using `deepspeed_config_file`, the following accelerate config variables will be 
ignored: ['gradient_accumulation_steps', 'gradient_clipping', 'zero_stage', 
'offload_optimizer_device', 'offload_param_device', 'zero3_save_16bit_model', 'mixed_precision'].
Please specify them appropriately in the DeepSpeed config file.
If you are using accelerate config file, set `mixed_precision=no` and remove others config variables
mentioned in the above specified list; else don't specify these config variables in `accelerate 
launch` command. 
The easiest method is to create new config following the questionnaire via  `accelerate config`.
It will only ask for the necessary config variables when using `deepspeed_config_file`.
```

### DIFF
--- a/docs/source/usage_guides/deepspeed.mdx
+++ b/docs/source/usage_guides/deepspeed.mdx
@@ -397,7 +397,7 @@ Only the `auto` fields specified in above examples are handled by `prepare` meth
 
 **Things to note when using DeepSpeed Config File**
 
-Below is a smaple script using `deepspeed_config_file` in different scenarios.
+Below is a sample script using `deepspeed_config_file` in different scenarios.
 
 Code `test.py`:
 
@@ -417,7 +417,7 @@ if __name__ == "__main__":
 
 **Scenario 1**: Manually tampered accelerate config file having `deepspeed_config_file` along with other entries.
 
-1. `accelerate config`:
+1. Content of the `accelerate` config:
 
 ```yaml
 command_file: null
@@ -508,7 +508,7 @@ How many GPU(s) should be used for distributed training? [1]:4
 accelerate configuration saved at ds_config_sample.yaml
 ```
 
-2. `accelerate config`:
+2. Content of the `accelerate` config:
 
 ```yaml
 compute_environment: LOCAL_MACHINE
@@ -541,9 +541,9 @@ Mixed precision type: bf16
 ds_config: {'bf16': {'enabled': True}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': False, 'offload_optimizer': {'device': 'none'}, 'offload_param': {'device': 'none'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 10, 'steps_per_print': inf, 'fp16': {'enabled': False}}
 ```
 
-**Scenario 3**: Setting the `accelerate launch` cmd args related to deepspeed as `auto` in  `deepspeed_config_file` and check that things work as expected.
+**Scenario 3**: Setting the `accelerate launch` command arguments related to DeepSpeed as `"auto"` in the DeepSpeed` configuration file and check that things work as expected.
 
-1. New `ds_config.json` with `auto` for the `accelerate launch` deepspeed cmd args:
+1. New `ds_config.json` with `"auto"` for the `accelerate launch` DeepSpeed command arguments:
 
 ```json
 {
@@ -580,7 +580,7 @@ Mixed precision type: fp16
 ds_config: {'bf16': {'enabled': False}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': True, 'offload_optimizer': {'device': 'nvme'}, 'offload_param': {'device': 'cpu'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 5, 'steps_per_print': inf, 'fp16': {'enabled': True, 'auto_cast': True}}
 ```
 
-**Note**: Remaining `auto` values are handled in `accelerator.prepare()` call as explained in point 2 of 
+**Note**: Remaining `"auto"` values are handled in `accelerator.prepare()` call as explained in point 2 of 
 `Important code changes when using DeepSpeed Config File`.
 
 ## Saving and loading
--- a/docs/source/usage_guides/deepspeed.mdx
+++ b/docs/source/usage_guides/deepspeed.mdx
@@ -494,9 +494,11 @@ It will only ask for the necessary config variables when using `deepspeed_config
 
 ```bash
 $ accelerate config
--------------------------------------------------------------------------------------------------------------------------------In which compute environment are you running?
+-------------------------------------------------------------------------------------------------------------------------------
+In which compute environment are you running?
 This machine                                                                                                                   
--------------------------------------------------------------------------------------------------------------------------------Which type of machine are you using?                                                                                           
+-------------------------------------------------------------------------------------------------------------------------------
+Which type of machine are you using?                                                                                           
 multi-GPU                                                                                                                      
 How many different machines will you use (use more than 1 for multi-node training)? [1]:                                       
 Do you wish to optimize your script with torch dynamo?[yes/NO]:                                                                
--- a/docs/source/usage_guides/deepspeed.mdx
+++ b/docs/source/usage_guides/deepspeed.mdx
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
 based on model, dataloaders, dummy optimizer and dummy schedulers provided to `prepare` method. 
 Only the `auto` fields specified in above examples are handled by `prepare` method and the rest have to be explicitly specified by the user.
 
+**Things to note when using DeepSpeed Config File**
+
+Below is a smaple script using `deepspeed_config_file` in different scenarios.
+
+Code `test.py`:
+
+```python
+from accelerate import Accelerator
+from accelerate.state import AcceleratorState
+
+
+def main():
+    accelerator = Accelerator()
+    accelerator.print(f"{AcceleratorState()}")
+
+
+if __name__ == "__main__":
+    main()
+```
+
+**Scenario 1**: Manually tampered accelerate config file having `deepspeed_config_file` along with other entries.
+
+1. `accelerate config`:
+
+```yaml
+command_file: null
+commands: null
+compute_environment: LOCAL_MACHINE
+deepspeed_config:
+  gradient_accumulation_steps: 1
+  gradient_clipping: 1.0
+  offload_optimizer_device: 'cpu'
+  offload_param_device: 'cpu'
+  zero3_init_flag: true
+  zero3_save_16bit_model: true
+  zero_stage: 3
+  deepspeed_config_file: 'ds_config.json'
+distributed_type: DEEPSPEED
+downcast_bf16: 'no'
+dynamo_backend: 'NO'
+fsdp_config: {}
+gpu_ids: null
+machine_rank: 0
+main_process_ip: null
+main_process_port: null
+main_training_function: main
+megatron_lm_config: {}
+num_machines: 1
+num_processes: 2
+rdzv_backend: static
+same_network: true
+tpu_name: null
+tpu_zone: null
+use_cpu: false
+```
+
+2. `ds_config.json`:
+
+```json
+{
+    "bf16": {
+        "enabled": true
+    },
+    "zero_optimization": {
+        "stage": 3,
+        "stage3_gather_16bit_weights_on_model_save": false,
+        "offload_optimizer": {
+            "device": "none"
+        },
+        "offload_param": {
+            "device": "none"
+        }
+    },
+    "gradient_clipping": 1.0,
+    "train_batch_size": "auto",
+    "train_micro_batch_size_per_gpu": "auto",
+    "gradient_accumulation_steps": 10,
+    "steps_per_print": 2000000
+}
+```
+
+3. Output of `accelerate launch test.py`:
+
+```bash
+ValueError: When using `deepspeed_config_file`, the following accelerate config variables will be ignored: 
+['gradient_accumulation_steps', 'gradient_clipping', 'zero_stage', 'offload_optimizer_device', 'offload_param_device', 
+'zero3_save_16bit_model', 'mixed_precision'].
+Please specify them appropriately in the DeepSpeed config file.
+If you are using an accelerate config file, remove others config variables mentioned in the above specified list.
+The easiest method is to create a new config following the questionnaire via `accelerate config`.
+It will only ask for the necessary config variables when using `deepspeed_config_file`.
+```
+
+**Scenario 2**: Use the solution of the error to create new accelerate config and check that no ambiguity error is now thrown.
+
+1. Run `accelerate config`:
+
+```bash
+$ accelerate config
+-------------------------------------------------------------------------------------------------------------------------------In which compute environment are you running?
+This machine                                                                                                                   
+-------------------------------------------------------------------------------------------------------------------------------Which type of machine are you using?                                                                                           
+multi-GPU                                                                                                                      
+How many different machines will you use (use more than 1 for multi-node training)? [1]:                                       
+Do you wish to optimize your script with torch dynamo?[yes/NO]:                                                                
+Do you want to use DeepSpeed? [yes/NO]: yes                                                                                    
+Do you want to specify a json file to a DeepSpeed config? [yes/NO]: yes                                                        
+Please enter the path to the json DeepSpeed config file: ds_config.json                                                        
+Do you want to enable `deepspeed.zero.Init` when using ZeRO Stage-3 for constructing massive models? [yes/NO]: yes
+How many GPU(s) should be used for distributed training? [1]:4
+accelerate configuration saved at ds_config_sample.yaml
+```
+
+2. `accelerate config`:
+
+```yaml
+compute_environment: LOCAL_MACHINE
+deepspeed_config:
+  deepspeed_config_file: ds_config.json
+  zero3_init_flag: true
+distributed_type: DEEPSPEED
+downcast_bf16: 'no'
+dynamo_backend: 'NO'
+fsdp_config: {}
+machine_rank: 0
+main_training_function: main
+megatron_lm_config: {}
+num_machines: 1
+num_processes: 4
+rdzv_backend: static
+same_network: true
+use_cpu: false
+```
+
+3. Output of `accelerate launch test.py`:
+
+```bash
+Distributed environment: DEEPSPEED  Backend: nccl
+Num processes: 4
+Process index: 0
+Local process index: 0
+Device: cuda:0
+Mixed precision type: bf16
+ds_config: {'bf16': {'enabled': True}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': False, 'offload_optimizer': {'device': 'none'}, 'offload_param': {'device': 'none'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 10, 'steps_per_print': inf, 'fp16': {'enabled': False}}
+```
+
+**Scenario 3**: Setting the `accelerate launch` cmd args related to deepspeed as `auto` in  `deepspeed_config_file` and check that things work as expected.
+
+1. New `ds_config.json` with `auto` for the `accelerate launch` deepspeed cmd args:
+
+```json
+{
+    "bf16": {
+        "enabled": "auto"
+    },
+    "zero_optimization": {
+        "stage": "auto",
+        "stage3_gather_16bit_weights_on_model_save": "auto",
+        "offload_optimizer": {
+            "device": "auto"
+        },
+        "offload_param": {
+            "device": "auto"
+        }
+    },
+    "gradient_clipping": "auto",
+    "train_batch_size": "auto",
+    "train_micro_batch_size_per_gpu": "auto",
+    "gradient_accumulation_steps": "auto",
+    "steps_per_print": 2000000
+}
+```
+
+2. Output of `accelerate launch --mixed_precision="fp16" --zero_stage=3 --gradient_accumulation_steps=5 --gradient_clipping=1.0 --offload_param_device="cpu" --offload_optimizer_device="nvme" --zero3_save_16bit_model="true" test.py`:
+
+```bash
+Distributed environment: DEEPSPEED  Backend: nccl
+Num processes: 4
+Process index: 0
+Local process index: 0
+Device: cuda:0
+Mixed precision type: fp16
+ds_config: {'bf16': {'enabled': False}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': True, 'offload_optimizer': {'device': 'nvme'}, 'offload_param': {'device': 'cpu'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 5, 'steps_per_print': inf, 'fp16': {'enabled': True, 'auto_cast': True}}
+```
+
+**Note**: Remaining `auto` values are handled in `accelerator.prepare()` call as explained in point 2 of 
+`Important code changes when using DeepSpeed Config File`.
+
 ## Saving and loading
 
 1. Saving and loading of models is unchanged for ZeRO Stage-1 and Stage-2.
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -722,6 +722,7 @@ def deepspeed_launcher(args):
 
     current_env["PYTHONPATH"] = env_var_path_add("PYTHONPATH", os.path.abspath("."))
     current_env["ACCELERATE_MIXED_PRECISION"] = str(mixed_precision)
+    current_env["ACCELERATE_CONFIG_DS_FIELDS"] = str(args.deepspeed_fields_from_accelerate_config).lower()
     current_env["ACCELERATE_USE_DEEPSPEED"] = "true"
     if args.zero_stage is not None:
         current_env["ACCELERATE_DEEPSPEED_ZERO_STAGE"] = str(args.zero_stage)
@@ -1025,7 +1026,13 @@ def launch_command(args):
                 ):
                     setattr(args, name, attr)
         if not args.mixed_precision:
-            args.mixed_precision = defaults.mixed_precision
+            if defaults.mixed_precision is None:
+                args.mixed_precision = "no"
+                mp_from_config_flag = False
+            else:
+                args.mixed_precision = defaults.mixed_precision
+                mp_from_config_flag = True
+
         if args.dynamo_backend is None:
             warned.append("\t`--dynamo_backend` was set to a value of `'no'`")
             args.dynamo_backend = "no"
@@ -1068,6 +1075,10 @@ def launch_command(args):
 
     # Use the proper launcher
     if args.use_deepspeed and not args.cpu:
+        args.deepspeed_fields_from_accelerate_config = list(defaults.deepspeed_config.keys())
+        if mp_from_config_flag:
+            args.deepspeed_fields_from_accelerate_config.append("mixed_precision")
+        args.deepspeed_fields_from_accelerate_config = ",".join(args.deepspeed_fields_from_accelerate_config)
         deepspeed_launcher(args)
     elif args.use_fsdp and not args.cpu:
         multi_gpu_launcher(args)
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -979,6 +979,7 @@ def launch_command(args):
 
     defaults = None
     warned = []
+    mp_from_config_flag = False
     # Get the default from the config file.
     if args.config_file is not None or os.path.isfile(default_config_file) and not args.cpu:
         defaults = load_config_from_file(args.config_file)
@@ -1028,7 +1029,6 @@ def launch_command(args):
         if not args.mixed_precision:
             if defaults.mixed_precision is None:
                 args.mixed_precision = "no"
-                mp_from_config_flag = False
             else:
                 args.mixed_precision = defaults.mixed_precision
                 mp_from_config_flag = True
@@ -1075,7 +1075,7 @@ def launch_command(args):
 
     # Use the proper launcher
     if args.use_deepspeed and not args.cpu:
-        args.deepspeed_fields_from_accelerate_config = list(defaults.deepspeed_config.keys())
+        args.deepspeed_fields_from_accelerate_config = list(defaults.deepspeed_config.keys()) if defaults else []
         if mp_from_config_flag:
             args.deepspeed_fields_from_accelerate_config.append("mixed_precision")
         args.deepspeed_fields_from_accelerate_config = ",".join(args.deepspeed_fields_from_accelerate_config)
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -309,14 +309,14 @@ def launch_command_parser(subparsers=None):
         default=None,
         type=str,
         help="Decides where (none|cpu|nvme) to offload optimizer states (useful only when `use_deepspeed` flag is passed). "
-        "If unspecified, will default to `none`.",
+        "If unspecified, will default to 'none'.",
     )
     deepspeed_args.add_argument(
         "--offload_param_device",
         default=None,
         type=str,
         help="Decides where (none|cpu|nvme) to offload parameters (useful only when `use_deepspeed` flag is passed). "
-        "If unspecified, will default to `none`.",
+        "If unspecified, will default to 'none'.",
     )
     deepspeed_args.add_argument(
         "--gradient_accumulation_steps",
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -301,45 +301,50 @@ def launch_command_parser(subparsers=None):
         "--zero_stage",
         default=None,
         type=int,
-        help="DeepSpeed's ZeRO optimization stage (useful only when `use_deepspeed` flag is passed).",
+        help="DeepSpeed's ZeRO optimization stage (useful only when `use_deepspeed` flag is passed). "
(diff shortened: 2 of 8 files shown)


sgugger (approved):
LGTM but let's also wait for @stas00 to check this.
Thanks for working on it!

stas00 (commented):
Thank you very much for working on this, Sourab!

I think perhaps the Issue I posted didn't succeed to communicate what the main problem is.

The mismatch is one issue, and your PR fantastically taking care of it - but really what we want is one definitive source of information.

The problem is that unless the user spent enough time with this code base they will try to maintain 2 sets of configs, which makes work very painful.

What I propose is that if ds_config is used any config duplicates are outright refused. As in:

> The ds_config is already defining the value for gradient_accumulation, please remove the ambiguous gradient accumulation setting from the accelerate config file.


Or alternatively require a single source in either way as we don't want to restrict users freedom of how they do thing.

In other words every time a duplicity is detected Accelerate will assert and say:

> Detected gradient_accumulation defined in both ds_config and accelerate config file, as this creates ambiguity please remove one of the settings.

Does my thinking make sense?

stas00:
I tried this branch, getting:

```
  File "m4/training/main.py", line 47, in <module>
    accelerator = Accelerator(
  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/accelerator.py", line 246, in __init__
    DeepSpeedPlugin() if os.environ.get("ACCELERATE_USE_DEEPSPEED", "false") == "true" else None
  File "<string>", line 12, in __init__
  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 412, in __post_init__
    self._deepspeed_config_checks()
  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 560, in _deepspeed_config_checks
    if ds_gradient_accumulation_steps != int(accelerate_gradient_accumulation_steps):
ValueError: invalid literal for int() with base 10: 'None'
```

configs:

```
{
    "fp16": {
        "enabled": true,
        "auto_cast": true,
        "loss_scale": 0.0,
        "initial_scale_power": 10,
        "loss_scale_window": 1000,
        "hysteresis": 2,
        "min_loss_scale": 1
    },
    "zero_optimization": {
        "stage": 3,
        "allgather_partitions": true,
        "allgather_bucket_size": 5e8,
        "overlap_comm": false,
        "reduce_scatter": true,
        "reduce_bucket_size": "auto",
        "contiguous_gradients": true,
        "stage3_gather_16bit_weights_on_model_save": false,
        "offload_optimizer": {
            "device": "none"
        },
        "offload_param": {
            "device": "none"
        }
    },
    "gradient_clipping": 1.0,
    "gradient_accumulation_steps": 2,
    "train_batch_size": "auto",
    "train_micro_batch_size_per_gpu": "auto",
    "steps_per_print": 2000000
}

deepspeed_config:
  deepspeed_multinode_launcher: standard
  deepspeed_config_file: ./configs/vopt-large-z3/ds_config.json
  zero3_init_flag: true
distributed_type: DEEPSPEED
fsdp_config: {}
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
num_machines: 1
num_processes: 1
#num_machines: 20
#num_processes: 80
use_cpu: false


```

muellerzr:
@pacman100 default needs to be 1 instead of `None` :)

pacman100:
Now, `accelerate launch --use_deepspeed --help` shows below output wherein default value info is specified `If unspecified, will default to *`:
```
...

DeepSpeed Arguments:
  Arguments related to DeepSpeed.

  --deepspeed_config_file DEEPSPEED_CONFIG_FILE
                        DeepSpeed config file.
  --zero_stage ZERO_STAGE
                        DeepSpeed's ZeRO optimization stage (useful only when `use_deepspeed` flag
                        is passed). If unspecified, will default to `2`.
  --offload_optimizer_device OFFLOAD_OPTIMIZER_DEVICE
                        Decides where (none|cpu|nvme) to offload optimizer states (useful only
                        when `use_deepspeed` flag is passed). If unspecified, will default to
                        `none`.
  --offload_param_device OFFLOAD_PARAM_DEVICE
                        Decides where (none|cpu|nvme) to offload parameters (useful only when
                        `use_deepspeed` flag is passed). If unspecified, will default to `none`.
  --gradient_accumulation_steps GRADIENT_ACCUMULATION_STEPS
                        No of gradient_accumulation_steps used in your training script (useful
                        only when `use_deepspeed` flag is passed). If unspecified, will default to
                        `1`.
  --gradient_clipping GRADIENT_CLIPPING
                        gradient clipping value used in your training script (useful only when
                        `use_deepspeed` flag is passed). If unspecified, will default to `1.0`.
  --zero3_init_flag ZERO3_INIT_FLAG
                        Decides Whether (true|false) to enable `deepspeed.zero.Init` for
                        constructing massive models. Only applicable with DeepSpeed ZeRO Stage-3.
                        If unspecified, will default to `true`.
  --zero3_save_16bit_model ZERO3_SAVE_16BIT_MODEL
                        Decides Whether (true|false) to save 16-bit model weights when using ZeRO
                        Stage-3. Only applicable with DeepSpeed ZeRO Stage-3. If unspecified, will
                        default to `false`.
  --deepspeed_hostfile DEEPSPEED_HOSTFILE
                        DeepSpeed hostfile for configuring multi-node compute resources.
  --deepspeed_exclusion_filter DEEPSPEED_EXCLUSION_FILTER
                        DeepSpeed exclusion filter string when using mutli-node setup.
  --deepspeed_inclusion_filter DEEPSPEED_INCLUSION_FILTER
                        DeepSpeed inclusion filter string when using mutli-node setup.
  --deepspeed_multinode_launcher DEEPSPEED_MULTINODE_LAUNCHER
                        DeepSpeed multi-node launcher to use. If unspecified, will default to
                        `pdsh`.

```

pacman100:
> 7\. output:
> 
> ```shell
> ValueError: When using `deepspeed_config_file`, the following accelerate config variables will be 
> ignored: ['gradient_accumulation_steps', 'gradient_clipping', 'zero_stage', 
> 'offload_optimizer_device', 'offload_param_device', 'zero3_save_16bit_model', 'mixed_precision'].
> Please specify them appropriately in the DeepSpeed config file. In accelerate config file, set 
> `mixed_precision=no` and remove others config variables mentioned in the above specified list.
> The easiest method is to create new config following the questionnaire via  `accelerate config`.
> It will only ask for the necessary config variables when using `deepspeed_config_file`.
> ```

Now, the error will be this.

pacman100:
> ```shell
> ValueError: When using `deepspeed_config_file`, the following accelerate config variables will be 
> ignored: ['gradient_accumulation_steps', 'gradient_clipping', 'zero_stage', 
> 'offload_optimizer_device', 'offload_param_device', 'zero3_save_16bit_model', 'mixed_precision'].
> Please specify them appropriately in the DeepSpeed config file.
> If you are using accelerate config file, set `mixed_precision=no` and remove others config variables
> mentioned in the above specified list; else don't specify these config variables in `accelerate 
> launch` command. 
> The easiest method is to create new config following the questionnaire via  `accelerate config`.
> It will only ask for the necessary config variables when using `deepspeed_config_file`.
> ```

Now, the error will be this

pacman100:
> but really what we want is one definitive source of information.

Hello @stas00, let us know if this addresses the issue. When `deepspeed_config_file` is specified, it is the single definitive source of information and error is raised when duplicates are found either in accelerate config file or through the arguments of `accelerate launch` command

src/accelerate/commands/launch.py
```diff
@@ -301,45 +301,50 @@ def launch_command_parser(subparsers=None):
...
+        help="DeepSpeed's ZeRO optimization stage (useful only when `use_deepspeed` flag is passed). "
+        "If unspecified, will default to `2`.",
     )
     deepspeed_args.add_argument(
         "--offload_optimizer_device",
         default=None,
         type=str,
-        help="Decides where (none|cpu|nvme) to offload optimizer states (useful only when `use_deepspeed` flag is passed).",
+        help="Decides where (none|cpu|nvme) to offload optimizer states (useful only when `use_deepspeed` flag is passed). "
+        "If unspecified, will default to `none`.",
```
sgugger, line 312: Nit: all `none`->`None`

src/accelerate/utils/dataclasses.py
```diff
@@ -549,6 +551,40 @@ def zero3_init_context_manager(self, enable=False):
...
+
+        if duplicate_values_flag:
+            env_variable_names_to_ignore = [
+                name.replace("ACCELERATE_", "").replace("DEEPSPEED_", "").lower()
+                for name in env_variable_names_to_ignore
+            ]
+            raise ValueError(
+                f"When using `deepspeed_config_file`, the following accelerate config variables will be ignored: {env_variable_names_to_ignore}.\n"
+                "Please specify them appropriately in the DeepSpeed config file.\n"
+                "If you are using accelerate config file, set `mixed_precision=no` "
```
sgugger, line 581:
```suggestion
                "If you are using an accelerate config file, set `mixed_precision=no` "
```

src/accelerate/utils/dataclasses.py
```diff
@@ -549,6 +551,40 @@ def zero3_init_context_manager(self, enable=False):
...
+                "and remove others config variables mentioned in the above specified list; "
+                "else don't specify these config variables in `accelerate launch` command. \n"
```
sgugger, line 583:
```suggestion
                "and make sure to not specify these config variables in `accelerate launch` command. \n"
```

sgugger (approved):
Thanks for working on this! Left a couple of nits.

src/accelerate/commands/launch.py
pacman100: Hello, `none` is the strong option as possible values are `cpu`|`nvme`l`none`

sgugger: Oh then add quotes?

pacman100: Done.

src/accelerate/utils/dataclasses.py
```diff
@@ -549,6 +551,40 @@ def zero3_init_context_manager(self, enable=False):
...
+
+        if duplicate_values_flag:
+            env_variable_names_to_ignore = [
+                name.replace("ACCELERATE_", "").replace("DEEPSPEED_", "").lower()
+                for name in env_variable_names_to_ignore
+            ]
+            raise ValueError(
+                f"When using `deepspeed_config_file`, the following accelerate config variables will be ignored: {env_variable_names_to_ignore}.\n"
+                "Please specify them appropriately in the DeepSpeed config file.\n"
+                "If you are using an accelerate config file, set `mixed_precision=no` "
```
stas00, line 581: May I ask to resolve this config as well. Did I understand it right that here you suggest to the user to set `mixed_precision=no` even if they use mixed precision in ds_config.json? If so this now is worse than before, as now you're proposing to force 2 totally opposite settings in the same config entry.

Unless I'm misunderstanding the wording that is.

stas00 (commented):
Thank you for further improvements, Sourab.

Will this still work with `auto` entries in `ds_config.json`, which were designed to be filled by either the system when they need to be calculated at run time or via cmd line args? I'm yet to try it out, but I have a feeling that the latter will no longer work, would it.

pacman100:
> Will this still work with `auto` entries in `ds_config.json`

Hello Stas, they will work as before because nothing is changing other than the fact that an error is raised if their is any duplicate entry in `accelerate config` with things that are in `deepspeed_config_file`. Here, these fields which can be duplicates are never `auto` in either case and hence fields that had `auto` support won't be impacted.

stas00:
Thank you for confirming that, @pacman100 - let me try it out as it's easier to see things in action.

src/accelerate/utils/dataclasses.py
```diff
@@ -578,10 +578,9 @@ def _deepspeed_config_checks(self):
             raise ValueError(
                 f"When using `deepspeed_config_file`, the following accelerate config variables will be ignored: {env_variable_names_to_ignore}.\n"
                 "Please specify them appropriately in the DeepSpeed config file.\n"
-                "If you are using an accelerate config file, set `mixed_precision=no` "
-                "and remove others config variables mentioned in the above specified list; "
+                "If you are using an accelerate config file, remove others config variables mentioned in the above specified list; "
                 "and make sure to not specify these config variables in `accelerate launch` command. \n"
-                "The easiest method is to create new config following the questionnaire via  `accelerate config`.\n"
+                "The easiest method is to create new config following the questionnaire via `accelerate config`.\n"
```
stas00, line 583:
```suggestion
                "The easiest method is to create a new config following the questionnaire via `accelerate config`.\n"
```

src/accelerate/utils/dataclasses.py
pacman100: Done. I've simplified the error message wherein user can just remove all the ambiguous entries from  accelerate config file and not specify them in launch command

stas00: to focus we are dealing with `mixed_precision` duplicity here:

I've just tried:

```
deepspeed_config:
  deepspeed_multinode_launcher: standard
  deepspeed_config_file: ./configs/vopt-large-z3/ds_config.json
  zero3_init_flag: true
distributed_type: DEEPSPEED
mixed_precision: 'no'
```

```
{
    "fp16": {
        "enabled": true,
        [...]
```
and it doesn't assert.

stas00: (incidentally ``mixed_precision`: fp16|bf16|'no'` is odd - why do different values don't follow the same style - i.e 'no' in quotes but not the other values)

stas00:
But this shows otherwise:
```
  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 409, in __post_init__
    raise ValueError("gradient_accumulation_steps cannot be set to 'auto' in the DeepSpeed config.")
ValueError: gradient_accumulation_steps cannot be set to 'auto' in the DeepSpeed config.
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid
```

why can't it be auto? I'd expect to set it to the default value or the value passed via the cmd line arg.

src/accelerate/utils/dataclasses.py
pacman100: Hello @stas, this case can't be helped as 'no' is the default value and if one doesn't specify the entry in accelerate config file, it will get default value of 'no' and thereby we have no way to check if user has given the default in config file or was it set to default when reading the config file. It would mean a lot of code rewriting as `mixed_precision` is used in various other non DeepSpeed parts.

The default value of `mixed_precision` would be overriden by that in `deepspeed_config_file`. Now, if you answer the questionnaire via `accelerate config`, the entry of `mixed_precision` won't be there when using `deepspeed_config_file`. 


@sgugger, what are your thoughts around this.

pacman100:
Regarding `gradient_accumulation_steps` and `auto`, that piece of code wasn't changed at all and behaviour is the same as before. If one doesn't specify that entry in the config file, it is set to default of `1`, if it is `auto` error is raised else thale given value is used. When not using `deepspeed_config_file`, it is asked when using `accelerate config` cmd.

src/accelerate/utils/dataclasses.py
stas00: understood! in which case I think this should work:

1. if the `mixed_precision` config isn't there all is good
2. if it is there and ds_config is on it must match  ds_config

stas00:
> Regarding `gradient_accumulation_steps` and `auto`, that piece of code wasn't changed at all and behaviour is the same as before. If one doesn't specify that entry in the config file, it is set to default of `1`, if it is `auto` error is raised else thale given value is used. When not using `deepspeed_config_file`, it is asked when using `accelerate config` cmd.

OK, so your logic is different from HF Trainer then.

The HF Trainer was directing cmd args into the config file's `auto` values so that the user could override them via cmd args.

I'm not saying the 2 logicss have to match. If I am not mistaken the accelerate logic is less flexible, but it's ok if you prefer it that way.

pacman100:
As accelerate is meant to work with all models apart from Transformers and user being in control of the training loop, they are in charge of all the arguments and the naming convention of arguments will be different across different users. On the other hand, in Trainer, users are restricted to a given args set and as such those can be used to fill the DeepSpeed config due to clear mapping between args and DS config params. The idea is that artifacts sent to `accelerator.prepare` have the params required by DS config and we exactly know the mapping between them and makes filling of the params independent of the user's training loop and their arguments naming convention.

In accelerate, the `auto` values are those that can be filled via artifacts being sent to `accelerator.prepare` as all the other places,  the  user has complete control over the training loop, argument naming and filling. The user still has flexibility to fill in all the `auto` values themselves as mentioned here https://github.com/huggingface/accelerate/pull/676#issuecomment-1317990966.

stas00:
> As accelerate is meant to work with all models apart from Transformers and user being in control of the training loop, they are in charge of all the arguments and the naming convention of arguments will be different across different users. On the other hand, in Trainer, users are restricted to a given args set and as such those can be used to fill the DeepSpeed config due to clear mapping between args and DS config params. The idea is that artifacts sent to `accelerator.prepare` have the params required by DS config and we exactly know the mapping between them and makes filling of the params independent of the user's training loop and their arguments naming convention.

Thank you for explaining this to me, Sourab, but I'm having a hard time following how Accelerate is any different from HF Trainer wrt sending cmd line arg values to the unfilled out config values in ds_config. e.g.  the Accelerate launcher provides an explicit list of cmd line args for the deepspeed use. There is a 1:1 mapping here as well. Could you please explain how is this different from the HF Trainer?

But as I said above it's totally fine if you prefer to do it this way, Sourab. This doesn't prevent users from doing what they need.

> In accelerate, the auto values are those that can be filled via artifacts being sent to accelerator.prepare as all the other places, the user has complete control over the training loop, argument naming and filling. The user still has flexibility to fill in all the auto values themselves as mentioned here https://github.com/huggingface/accelerate/pull/676#issuecomment-1318139353

Understood. more work, but doable. Thank you for the explanations.

pacman100:
> Could you please explain how is this different from the HF Trainer?

Users can have `bs`\`batch_size` as cmd arguments in their code and as such we can't fill ds config's `train_micro_batch_size_per_gpu` whereas Trainer always maps `args.per_device_train_batch_size` to it. The same reason can go for other configs. 

Please note that `accelerate launch` cmd args are primarily used for setting accelerate config's deespseed fields rather than setting `auto` values of `deepspeed_config_file`. Now I understood that you meant using `accelerate launch` cmd args for filling in auto` values of `deepspeed_config_file` and have made respective changes.

Latest changes:

Code `test.py`:
```python
from accelerate.state import AcceleratorState

def main():
    accelerator = Accelerator()
    accelerator.print(f"{AcceleratorState()}")
if __name__ == "__main__":
    main()
```

Scenario 1: manually tampered accelerate config file having `deepspeed_config_file` along with other entries.

1. `accelerate config`:
```yaml
command_file: null
commands: null
deepspeed_config:
  gradient_clipping: 1.0
  offload_optimizer_device: 'cpu'
  offload_param_device: 'cpu'
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
  deepspeed_config_file: 'ds_config.json'
distributed_type: DEEPSPEED
downcast_bf16: 'no'
dynamo_backend: 'NO'
fsdp_config: {}
gpu_ids: null
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
megatron_lm_config: {}
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_name: null
tpu_zone: null
use_cpu: false
```

2. `ds_config.json`:
```json
{
    "bf16": {
        "enabled": true
    },
    "zero_optimization": {
        "stage": 3,
        "stage3_gather_16bit_weights_on_model_save": false,
        "offload_optimizer": {
            "device": "none"
        },
        "offload_param": {
            "device": "none"
        }
    },
    "gradient_clipping": 1.0,
    "train_batch_size": "auto",
    "train_micro_batch_size_per_gpu": "auto",
    "gradient_accumulation_steps": 10,
    "steps_per_print": 2000000
}
```

3. Output of `accelerate launch test.py`:
```
ValueError: When using `deepspeed_config_file`, the following accelerate config variables will be ignored: 
['gradient_accumulation_steps', 'gradient_clipping', 'zero_stage', 'offload_optimizer_device', 'offload_param_device', 
'zero3_save_16bit_model', 'mixed_precision'].
Please specify them appropriately in the DeepSpeed config file.
If you are using an accelerate config file, remove others config variables mentioned in the above specified list.
The easiest method is to create a new config following the questionnaire via `accelerate config`.
It will only ask for the necessary config variables when using `deepspeed_config_file`.
```

Scenario 2: Use the solution of the error to create new accelerate config and check that no ambiguity error is now thrown.
1. Run `accelerate config`:
```
$ accelerate config --config_file ds_config_sample.yaml
-------------------------------------------------------------------------------------------------------------------------------In which compute environment are you running?
This machine                                                                                                                   
-------------------------------------------------------------------------------------------------------------------------------Which type of machine are you using?                                                                                           
multi-GPU                                                                                                                      
How many different machines will you use (use more than 1 for multi-node training)? [1]:                                       
Do you wish to optimize your script with torch dynamo?[yes/NO]:                                                                
Do you want to use DeepSpeed? [yes/NO]: yes                                                                                    
Do you want to specify a json file to a DeepSpeed config? [yes/NO]: yes                                                        
Please enter the path to the json DeepSpeed config file: ds_config.json                                                        
Do you want to enable `deepspeed.zero.Init` when using ZeRO Stage-3 for constructing massive models? [yes/NO]: yes
How many GPU(s) should be used for distributed training? [1]:4
accelerate configuration saved at ds_config_sample.yaml
```

2. `accelerate config`:
```yaml
deepspeed_config:
  deepspeed_config_file: ds_config.json
  zero3_init_flag: true
distributed_type: DEEPSPEED
downcast_bf16: 'no'
dynamo_backend: 'NO'
fsdp_config: {}
machine_rank: 0
main_training_function: main
megatron_lm_config: {}
num_machines: 1
num_processes: 4
rdzv_backend: static
same_network: true
use_cpu: false
```

3. Output of `accelerate launch test.py`:
```
Distributed environment: DEEPSPEED  Backend: nccl
Num processes: 4
Process index: 0
Local process index: 0
Device: cuda:0
Mixed precision type: bf16
ds_config: {'bf16': {'enabled': True}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': False, 'offload_optimizer': {'device': 'none'}, 'offload_param': {'device': 'none'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 10, 'steps_per_print': inf, 'fp16': {'enabled': False}}
```

Scenario 3: Setting the `accelerate launch` cmd args related to deepspeed as `auto` in  `deepspeed_config_file` and check that things work as expected.
1. new `ds_config.json` with `auto` for the `accelerate launch` deepspeed cmd args:
```json
{
    "bf16": {
        "enabled": "auto"
    },
    "zero_optimization": {
        "stage": "auto",
        "stage3_gather_16bit_weights_on_model_save": "auto",
        "offload_optimizer": {
            "device": "auto"
        },
        "offload_param": {
            "device": "auto"
        }
    },
    "gradient_clipping": "auto",
    "train_batch_size": "auto",
    "train_micro_batch_size_per_gpu": "auto",
    "gradient_accumulation_steps": "auto",
    "steps_per_print": 2000000
}
```

2. Output of `accelerate launch --mixed_precision="fp16" --zero_stage=3 --gradient_accumulation_steps=5 --gradient_clipping=1.0 --offload_param_device="cpu" --offload_optimizer_device="nvme" --zero3_save_16bit_model="true" test.py`:
```bash
Distributed environment: DEEPSPEED  Backend: nccl
Num processes: 4
Process index: 0
Local process index: 0
Device: cuda:0
Mixed precision type: fp16
ds_config: {'bf16': {'enabled': False}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': True, 'offload_optimizer': {'device': 'nvme'}, 'offload_param': {'device': 'cpu'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 5, 'steps_per_print': inf, 'fp16': {'enabled': True, 'auto_cast': True}}
```

stas00:
Looks fantastic, Sourab! Thank you for the improvements and taking time to layout out the different scenarios - if I'm not mistaken those would make for perfect additions to the documentation if it resonates.

BTW, the config generates things like:
```
fsdp_config: {}
megatron_lm_config: {}
```
why not just skip parts that the user hasn't asked for? It just makes the config scarier than it is, no? I'm asking since when I first looked at it I wasn't a all sure which of the empty placeholders were safe to remove and which aren't. My personal preference is for active config - that is to only ever list config entries that I work with and any defaults should be just that defaults and not be listed at all. Which I suppose isn't the case with typical configs where everything is listed out whether it's being used or not.

And I can of course remove all those, so definitely it's not an issue, I'm just asking if my thinking resonates with you.

stas00 (approved):

stas00:
Sourab, I found one more in one of our tests:

```
  zero3_init_flag: true
  zero_stage: 2
```

This combo is quietly getting accepted. I'm concerned that a developer may see `zero3_init_flag: true` and think it's zero3?

Do you think accelerate should assert when `zero3_init_flag` is used with `stage < 3`?

pacman100:
Hello @stas00, with current setup below warning is given which I think is fine:
```bash
/home/sourab/accelerate/src/accelerate/utils/dataclasses.py:470: UserWarning: DeepSpeed Zero3 Init flag is only applicable for ZeRO Stage 3. Setting it to False.
```

stas00:
oh boy. I didn't see it. :(

If a tree falls in a forest and no one is around to hear it, does it make a sound?

I guess I need to start using this pragma to turn warnings into errors, but then some warnings can't be acted upon :(

```
import warnings
warnings.filterwarnings("error")
```

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
 based on model, dataloaders, dummy optimizer and dummy schedulers provided to `prepare` method. 
 Only the `auto` fields specified in above examples are handled by `prepare` method and the rest have to be explicitly specified by the user.
 
+**Things to note when using DeepSpeed Config File**
+
+Below is a smaple script using `deepspeed_config_file` in different scenarios.
```
sgugger, line 400:
```suggestion
Below is a sample script using `deepspeed_config_file` in different scenarios.
```

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+
+**Scenario 2**: Use the solution of the error to create new accelerate config and check that no ambiguity error is now thrown.
+
+1. Run `accelerate config`:
+
+```bash
+$ accelerate config
+-------------------------------------------------------------------------------------------------------------------------------In which compute environment are you running?
+This machine                                                                                                                   
+-------------------------------------------------------------------------------------------------------------------------------Which type of machine are you using?
```
sgugger, line 499: Are the lines of `-` intended? Should there be a new line before the text?

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+    accelerator.print(f"{AcceleratorState()}")
+
+
+if __name__ == "__main__":
+    main()
+```
+
+**Scenario 1**: Manually tampered accelerate config file having `deepspeed_config_file` along with other entries.
+
+1. `accelerate config`:
```
sgugger, line 420:
```suggestion
1. Content of the `accelerate` config:
```
(to not mistake this with the command `accelerate config`

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+Do you wish to optimize your script with torch dynamo?[yes/NO]:                                                                
+Do you want to use DeepSpeed? [yes/NO]: yes                                                                                    
+Do you want to specify a json file to a DeepSpeed config? [yes/NO]: yes                                                        
+Please enter the path to the json DeepSpeed config file: ds_config.json                                                        
+Do you want to enable `deepspeed.zero.Init` when using ZeRO Stage-3 for constructing massive models? [yes/NO]: yes
+How many GPU(s) should be used for distributed training? [1]:4
+accelerate configuration saved at ds_config_sample.yaml
+```
+
+2. `accelerate config`:
```
sgugger, line 511:
```suggestion
2. Content of the `accelerate` config:
```

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+Distributed environment: DEEPSPEED  Backend: nccl
+Num processes: 4
+Process index: 0
+Local process index: 0
+Device: cuda:0
+Mixed precision type: bf16
+ds_config: {'bf16': {'enabled': True}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': False, 'offload_optimizer': {'device': 'none'}, 'offload_param': {'device': 'none'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 10, 'steps_per_print': inf, 'fp16': {'enabled': False}}
+```
+
+**Scenario 3**: Setting the `accelerate launch` cmd args related to deepspeed as `auto` in  `deepspeed_config_file` and check that things work as expected.
```
sgugger, line 544:
```suggestion
**Scenario 3**: Setting the `accelerate launch` command arguments related to DeepSpeed as `"auto"` in the DeepSpeed` configuration file and check that things work as expected.
```

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+
+1. New `ds_config.json` with `auto` for the `accelerate launch` deepspeed cmd args:
```
sgugger, line 546:
```suggestion
1. New `ds_config.json` with `"auto"` for the `accelerate launch` DeepSpeed command arguments:
```

docs/source/usage_guides/deepspeed.mdx
```diff
@@ -395,6 +395,194 @@ We will look at the changes needed in the code when using these.
...
+Distributed environment: DEEPSPEED  Backend: nccl
+Num processes: 4
+Process index: 0
+Local process index: 0
+Device: cuda:0
+Mixed precision type: fp16
+ds_config: {'bf16': {'enabled': False}, 'zero_optimization': {'stage': 3, 'stage3_gather_16bit_weights_on_model_save': True, 'offload_optimizer': {'device': 'nvme'}, 'offload_param': {'device': 'cpu'}}, 'gradient_clipping': 1.0, 'train_batch_size': 'auto', 'train_micro_batch_size_per_gpu': 'auto', 'gradient_accumulation_steps': 5, 'steps_per_print': inf, 'fp16': {'enabled': True, 'auto_cast': True}}
+```
+
+**Note**: Remaining `auto` values are handled in `accelerator.prepare()` call as explained in point 2 of
```
sgugger, line 583:
```suggestion
**Note**: Remaining `"auto"` values are handled in `accelerator.prepare()` call as explained in point 2 of 
```

src/accelerate/utils/dataclasses.py
```diff
@@ -450,15 +461,17 @@ def __post_init__(self):
             if self.gradient_clipping:
                 config["gradient_clipping"] = self.gradient_clipping
             self.hf_ds_config = HfDeepSpeedConfig(config)
+
         self.deepspeed_config = self.hf_ds_config.config
         self.deepspeed_config["steps_per_print"] = float("inf")  # this will stop deepspeed from logging @ stdout
         if self.zero3_init_flag is None:
-            self.zero3_init_flag = os.environ.get("ACCELERATE_DEEPSPEED_ZERO3_INIT", "false") == "true"
+            self.zero3_init_flag = os.environ.get("ACCELERATE_DEEPSPEED_ZERO3_INIT", "true") == "true"
```
sgugger, line 468: Why is this set to True by default now?

sgugger (commented):
Left some nits on the doc. Why did we switch the default for `zero3_init_flag` to `True`?

docs/source/usage_guides/deepspeed.mdx
pacman100: these are the exact outputs from the new config #830. Added new lines.

src/accelerate/utils/dataclasses.py
pacman100: Hello, as stas mentioned in Issue #922 

> this flag should be True by default as zero stage 3 is for large models, it's very unlikely the user will be able to load those models w/o zero.Init

sgugger: Is this only used if zero3 is enabled in the rest of the code? It should at least only default to True when Zero-3 is enabled otherwise.

pacman100: Done.

sgugger (approved):

PR #943, opened 2022-12-23
fix mp related test fails

pacman100:
### What does this PR do?
1. Fixes bugs introduced in #935


muellerzr (approved):
Thanks! I'm okay with merging but let's keep an eye on the CI post merge to make sure it all passes. (I can't run it myself since it was made via a fork and not a branch of `accelerate`)

Issue #938, opened 2022-12-22
`load_checkpoint_and_dispatch` fails for GPTNeoX 20B


johnPertoft:
### System Info

```Shell
- `Accelerate` version: 0.15.0
- Platform: Linux-4.19.0-22-cloud-amd64-x86_64-with-glibc2.35
- Python version: 3.10.6
- Numpy version: 1.23.5
- PyTorch version (GPU?): 1.13.0+cu117 (True)
- `Accelerate` default config:
        Not found

- 1xNvidia T4, 510.47.03, cuda 11.7
- 30gb ram
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Run the following script (slightly adapted from https://huggingface.co/docs/accelerate/usage_guides/big_modeling) but with the GPTNeoX model.

```
from accelerate import init_empty_weights
from accelerate import load_checkpoint_and_dispatch
from huggingface_hub import snapshot_download
from transformers import GPTNeoXTokenizerFast
from transformers import AutoConfig
from transformers import AutoModelForCausalLM

model_path = snapshot_download("EleutherAI/gpt-neox-20b")
tokenizer = GPTNeoXTokenizerFast.from_pretrained(model_path)

prompt = "GPTNeoX20B is a 20B-parameter autoregressive Transformer model developed by EleutherAI."
inputs = tokenizer(prompt, return_tensors="pt")
inputs = inputs.to("cuda")

print("Creating empty model")
config = AutoConfig.from_pretrained(model_path)
tokenizer = GPTNeoXTokenizerFast.from_pretrained(model_path)
with init_empty_weights():
    model = AutoModelForCausalLM.from_config(config)

print("Loading checkpoint")
model = load_checkpoint_and_dispatch(
    model,
    checkpoint=model_path,
    device_map="auto",
    no_split_module_classes=["GPTNeoXLayer"],
    offload_folder="/tmp/gpt-neox-20b-offload-accelerate",
)

for k, v in model.hf_device_map.items():
    print(k, v)

print("Running forward pass")
outputs = model(**inputs)
```

It crashes in `load_checkpoint_and_dispatch` with the following trace
```
Traceback (most recent call last):
[...]

PR #945, opened 2022-12-27
Fix offload when weights are on the GPU

sgugger:
This was reported in #938 and is a very small and easy fix: some checkpoints may have weights on the GPU and if the user does not pass along `map_location="cpu"` (which also our `load_checkpoint` function does not) we end with an error at this line.


pacman100 (approved):
Thanks for the fix! LGTM 🤗

Issue #933, opened 2022-12-20
[feature request] add `accelerate.state.is_initialized()`


stas00:
asking to add `accelerate.state.is_initialized()`
thread: https://huggingface.slack.com/archives/C01Q6JPP6NA/p1671149816440139
cc: @muellerzr

PR #949, opened 2023-01-03
Add is_initialized method and refactor

muellerzr:
# Add a static `is_initialized` method and refactor existing logic for initialization check

## What does this add?

This PR introduces a static `is_initialized` method which will check  if the state's dictionary is empty or not. It also refactors current methods to have `initialized` be a `@property` check of the same to increase code clarity towards what is happening under the hood.

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/933

## Why is it needed?

The feature request asked for a similar API to what `torch` uses for `cuda`, i.e. `torch.cuda.is_initialized()` for `Accelerate`, so a user can do `import accelerate; accelerate.state.is_initialized()`

### User-facing:

A new `is_initialized` method was added to the `accelerate.state` module

### Internal structure:

`initialized` was changed to be a dictionary check to mimic how the new `is_initialized()` function operates

## Basic Usage Example(s):

```python
# One way:
import accelerate
accelerate.state.is_initialized()

# Or:
accelerator = Accelerator()
accelerator.state.initialized
```

When a static method is needed to check if the `AcceleratorState` has been initialized without doing a `getattr`, or a cleaner interface to present to the user

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -78,7 +78,6 @@ class AcceleratorState:
         if parse_flag_from_env("ACCELERATE_USE_CPU"):
             cpu = True
         self._check_initialized(mixed_precision, cpu)
-        self.fork_launched = parse_flag_from_env("FORK_LAUNCHED", 0)
         if not self.initialized:
             self.backend = None
             self.deepspeed_plugin = None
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for adding this!

PR #950, opened 2023-01-03
Fix DeepSpeed tests

muellerzr:
The `DeepSpeed` tests manually set `AcceleratorState.initialized` when trying to test the mixed precision types when it should use `_reset_state()`, leading to a failure on the recent API change introduced in #949.

IIUC this should be the proper fix, let me know if there was a specific reason for this particular setup @pacman100 when testing and I can modify this fix differently.

Proof tests pass can be seen here: https://github.com/huggingface/accelerate/actions/runs/3830884007


pacman100 (approved):
LGTM! Thank you for the fixes 🤗

PR #955, opened 2023-01-06
Typo fix in src/accelerate/utils/modeling.py

ryderwishart:
Simple string change (typo fix) I happened to notice and figured I should just fix while I'm looking at it.


sgugger (approved):
Thanks for the fix!

PR #962, opened 2023-01-10
Log Images and other types to wandb

tcapelle:
This PR adds logging of images, tables and artifacts to the `WandBTracker` class. It also adds the functionality to Tensorboard. This enables to simplify the code used on https://github.com/huggingface/diffusers for logging images like [here](https://github.com/huggingface/diffusers/blob/aba2a65d6ab47c0d1c12fa47e9b238c1d3e34512/examples/unconditional_image_generation/train_unconditional.py#L491)

I am opening the PR as draft to discuss on the API changes needed, and what should be on the `Accelerator` class or the `Tracker`.

I refer to this [PR](https://github.com/huggingface/diffusers/pull/1245) that asked to delegate the `wandb` functionalities to the underlying accelerate codebase.


sgugger:
cc @muellerzr

muellerzr:
@tcapelle I believe that something like this we'd want to leave on the `wandb` side (and other tracking classes) for other folks to use, and keep the core of it as general purpose as possible (otherwise this explodes rather quickly, rather than being *flexible* using inheritance by users). Users can easily enough perform logging with custom trackers in the framework, and I'm looking at making it even easier (without checks to `is_main_process` explicitly).

tcapelle:
I was pushing here as from the diffusers library, there was a suggestion to [integrate into accelerate](https://github.com/huggingface/diffusers/pull/1245#pullrequestreview-1184657800).
The idea is that generative libraries require some more sophisticated logging than just the metrics, as one wants to monitor image generation over time to assess model perf.
Right now there is code to log images to tensorboard and we could unify that into a single call into accelerate.
We have some scripts that call `wandb` and do the integration in diffusers like [here](https://github.com/huggingface/diffusers/blob/a66f2baeb782e091dde4e1e6394e46f169e5ba58/examples/text_to_image/train_text_to_image_lora.py#L761)
or others that [do tensorboard](https://github.com/huggingface/diffusers/blob/a66f2baeb782e091dde4e1e6394e46f169e5ba58/examples/unconditional_image_generation/train_unconditional.py#L546).
If we had a `log_images` baked into accelerate, both would have been possible.

muellerzr (approved):
@tcapelle thinking on this more, I believe that this all looks good to me. The caveat with this all being that in `Accelerator` it should still only remain as `log`.

Using the API I mentioned earlier is how users should access these internal `log_images` etc functions, however the `Accelerator` won't see an explicit `log_images` function. For convenience this is something diffusers should be doing instead if they want to wrap around it with their own tracker etc. We will just provide the low-level aspect of it

muellerzr:
Checking with @sgugger to make sure we're in alignment on it!

src/accelerate/tracking.py
```diff
@@ -119,6 +119,19 @@ def log(self, values: dict, step: Optional[int], **kwargs):
         """
         pass
 
+    def log_images(self, values: dict, step: Optional[int], **kwargs):
```
sgugger, line 122: There is no need to add this here.

sgugger (commented):
It should only be added to the trackers that actually implement it though. I don't think we need to have it on the base class (this is what I meant with not using abstract class @muellerzr ).

src/accelerate/tracking.py
```diff
@@ -272,6 +301,64 @@ def log(self, values: dict, step: Optional[int] = None, **kwargs):
...
+    def log_table(
+        self,
+        table_name: str,
+        columns: List[str] = None,
+        data: List[List[Any]] = None,
+        dataframe: Any = None,
+        step: Optional[int] = None,
+        **kwargs,
+    ):
+        """Log a Table containing any object type (text, image, audio, video, molecule, html, etc).
```
muellerzr, line 330:
```suggestion
        """
        Log a Table containing any object type (text, image, audio, video, molecule, html, etc).
```

muellerzr (commented):
One more nit here along with Sylvain's suggestion there and we should be okay to merge :)

tcapelle:
@muellerzr  I am curious how to access the underlying `log_images`, as the accelerator does not give access.

muellerzr:
@tcapelle there's a `get_tracker` api: https://huggingface.co/docs/accelerate/package_reference/accelerator#accelerate.Accelerator.get_tracker

You then just wrap it in:

```python
tracker = accelerator.get_tracker('wandb')
    tracker.xyz()
```

Looking at changing that API to be:

```python
tracker = accelerator.get_tracker('wandb')
tracker.xyz()
```
sometime this week

muellerzr:
cc @tcapelle :)

tcapelle:
I am working on this tomorrow =)

muellerzr:
@tcapelle is this good for a review? :)

muellerzr:
@tcapelle considering it's meant to save any object, not just the model, so I don't think that would be a wanted behavior. And that's too magical for this library I think.

tcapelle:
Do you think we can do something about this line, it makes my eyes bleed: https://github.com/huggingface/accelerate/blob/78151f87a4e35fe1a1ae6de2d6bbaff8038b71a4/src/accelerate/tracking.py#L277

muellerzr:
What would you have us do with it exactly?

tcapelle:
You already call `init_tackers` with `project_name` variable, so make it consistent.
```
I would love calling wandb.init(project=project_name)
```

muellerzr:
@tcapelle the API is built upon `project_name==run_name`. It would require an extensive rewrite/expansion to change this behavior and the change would need to be backward compatible for at least 3 versions. If you wish to do that, it can happen in a follow-up PR.

tcapelle:
I think this is good for review @muellerzr

muellerzr:
Sorry @tcapelle! This looks great to me, thanks for all your work!

PR #966, opened 2023-01-11
Fix OOM when moving tensor to designated device

johnPertoft:


src/accelerate/utils/modeling.py
```diff
@@ -126,17 +126,15 @@ def set_module_tensor_to_device(
     if value is not None:
         if dtype is None:
             # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
-            value = value.to(old_value.dtype)
-        elif str(value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
-            value = value.to(dtype)
```
sgugger, line 131: We need to keep this logic as models sometimes have a couple of weights in int/uint/bool that should not be converted to floating dtypes.

sgugger (commented):
Thanks for your PR. I don't understand how this changes anything in terms of memory use.

johnPertoft:
> I don't understand how this changes anything in terms of memory use.

I'm not entirely sure of the implementation details, but I guess the dtype cast is allocating memory for a new tensor which maybe caused my OOM issue in the linked issue.

PR #968, opened 2023-01-12
When use accelerator with deepspeed, We must import from accelerate package.

Don9wanKim:
In README.md file, explain "from accelerator import Accelerator, DeepSpeedPlugin"
But, Our package name is accelerate ( no accelerator )
When I use previous guidance, statement for this is failed.


pacman100 (approved):
Hello @Don9wanKim, Thank you for the fix!

Issue #847, opened 2022-11-14
Multi-GPU doesn't work


liu6381810:
Hi, when i run `accelerate launch my_scripy.py`  it seems do nothing even `my_scripy.py` only contain `print(1)`
But if i use one-gpu setting, `print(1)` work as expected

here is my accelerate env；
- `Accelerate` version: 0.14.0
- Platform: Linux-5.4.143.bsk.7-amd64-x86_64-with-debian-10.13
- Python version: 3.7.3
- Numpy version: 1.21.5
- PyTorch version (GPU?): 1.10.0 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - main_process_ip: None
        - main_process_port: None
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_name: None
        - tpu_zone: None
        - command_file: None
        - commands: None

muellerzr:
Can you share the entire script you are running?

liu6381810:
> Can you share the entire script you are running?

I run 
`accelerate launch test.py`
the test.py is
`from accelerate import Accelerator 
print(1)`
It's weird this command report nothing even the test.py is not exist

muellerzr:
@liu6381810 I'm afraid I can't recreate this. I have done:

```python
# Contents of test.py
from accelerate import Accelerator

print(1)
```

The config:
```yaml
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: MULTI_GPU
downcast_bf16: 'no'
fsdp_config: {}
gpu_ids: all
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'no'
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
use_cpu: false
```

The launch command:

```bash
accelerate launch test.py
```

Can you try again with the above setup?

tobigue:
We had a similar issue when trying to use `accelerate==0.15` with `pytorch==1.9.0`. When adding the `--multi_gpu` flag `accelerate launch` just exits without outputs or errors.

[...]

PR #978, opened 2023-01-17
Raise minimum version for distrib launch

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/847

TL;DR there was a neat bug in torch 1.9.0 where `torchrun` wouldn't always *actually run*. This PR changes the flag for checking to be 1.9.1, which I found *did* work after successfully recreating the 1.9.0 issue. 

It also modifies the `if/else` flag as it was always going to raise if the value was `<` 1.9.0 but gives it a much more clear answer as to what is truly happening here.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -555,8 +555,10 @@ def simple_launcher(args):
 
 
 def multi_gpu_launcher(args):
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for the fix!

Issue #952, opened 2023-01-04
gather_for_metrics ignores the last batch of data


rajcscw:
### System Info

```Shell
- `Accelerate` version: 0.15.0
- Platform: Linux-3.10.0-1160.45.1.el7.x86_64-x86_64-with-debian-buster-sid
- Python version: 3.7.15
- Numpy version: 1.21.6
- PyTorch version (GPU?): 1.12.0+cu102 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - dynamo_backend: NO
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: 4,5
        - main_process_ip: None
        - main_process_port: None
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_name: None
        - tpu_zone: None
        - command_file: None
        - commands: None
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

In the following example, the dataset is of size 512 and a batch size of 64. On gathering using ´gather_for_metrics` with 2 GPUs, the last batch seems to be ignored with a size of 0.

The output of the following code is:

Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([0, 20])

But the expected output is:
Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([128, 20])
Rank: 0 torch.Size([128, 20])

```python
from transformers import AutoModelForCausalLM, AutoTokenizer
from accelerate import Accelerator
from torch.utils.data import DataLoader
from tqdm import tqdm
import torch


class TestDataset(torch.utils.data.Dataset):
    def __init__(self, data):
        self.inputs = data
        self.output = data

    def __getitem__(self, index):
        return {"inputs": self.inputs[index], "targets": self.output[index]}

    def __len__(self):
[...]

PR #982, opened 2023-01-18
Ensure that last batch doesn't get dropped if perfectly even in gather_for_metrics

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/952 

TL;DR in a perfectly distributed batch setup (say dset 512, bs of 16), we need to ensure not to drop any samples and delegate to the `else` statement. Added a test to ensure this can get caught as well.


sgugger (approved):
Thanks for fixing!

PR #983, opened 2023-01-18
Fix test for converting tensor to proper dtype

sgugger:
I made a typo in #920 and forgot a `not` while copy-pasting some code :grimacing: 
This PR fixes that.


muellerzr (approved):
Oopsie! Is it possible we could have written a test to catch this? Or not as much and it's just a silent-never-called bug

PR #988, opened 2023-01-19
in sync with transformers, removing style_doc utils and using doc-builder instead

pacman100:
### What does this PR do?
1. In sync with transformers, removing `style_doc` utils and using `doc-builder` instead


sgugger (approved):
Thanks a lot!

PR #989, opened 2023-01-19
adding support for kwargs in `load_state`

pacman100:
### What does this PR do?
1. Fixes: https://huggingface.slack.com/archives/C01Q6JPP6NA/p1672362244049489. Adds `kwargs` keyword arguments to the `load_state` method of accelerate.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1704,7 +1704,7 @@ class Accelerator:
         self.project_configuration.iteration += 1
         return save_location
 
-    def load_state(self, input_dir: str):
+    def load_state(self, input_dir: str, **model_load_kwargs):
         """
         Loads the current states of the model, optimizer, scaler, RNG generators, and registered objects.
 
@@ -1717,6 +1717,8 @@ class Accelerator:
         Args:
             input_dir (`str` or `os.PathLike`):
                 The name of the folder all relevant weights and states were saved in.
+            model_load_kwargs (`dict`): Additional keyword arguments required for loading model, e.g.,
+            deepspeed's optional args for `load_checkpoint`.
         """
         # Check if folder exists
         input_dir = os.path.expanduser(input_dir)
@@ -1729,16 +1731,16 @@ class Accelerator:
         for i, model in enumerate(self._models):
             if self.distributed_type == DistributedType.FSDP:
                 logger.info("Loading FSDP model")
-                self.state.fsdp_plugin.load_model(self, model, input_dir, i)
+                self.state.fsdp_plugin.load_model(self, model, input_dir, i, **model_load_kwargs)
                 logger.info(f"FSDP Model loaded from input dir {input_dir}")
             elif self.distributed_type == DistributedType.DEEPSPEED:
                 logger.info("Loading DeepSpeed Model and Optimizer")
                 ckpt_id = f"{MODEL_NAME}" if i == 0 else f"{MODEL_NAME}_{i}"
-                model.load_checkpoint(input_dir, ckpt_id)
+                model.load_checkpoint(input_dir, ckpt_id, **model_load_kwargs)
                 logger.info(f"DeepSpeed Model and Optimizer loaded from input dir {os.path.join(input_dir, ckpt_id)}")
             elif self.distributed_type == DistributedType.MEGATRON_LM:
                 logger.info("Loading Megatron-LM Model, Optimizer and Scheduler")
-                model.load_checkpoint(input_dir)
+                model.load_checkpoint(input_dir, **model_load_kwargs)
                 logger.info(f"Megatron-LM Model , Optimizer and Scheduler loaded from input dir {input_dir}")
             else:
                 models.append(model)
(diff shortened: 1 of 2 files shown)


src/accelerate/accelerator.py
```diff
@@ -1717,6 +1717,8 @@ def load_state(self, input_dir: str):
         Args:
             input_dir (`str` or `os.PathLike`):
                 The name of the folder all relevant weights and states were saved in.
+            model_load_kwargs (`dict`): Additional keyword arguments required for loading model, e.g.,
+            deepspeed's optional args for `load_checkpoint`.
```
sgugger, line 1721:
```suggestion
            model_load_kwargs (`dict`):
                Additional keyword arguments required for loading model, e.g., deepspeed's optional args for `load_checkpoint`.
```

sgugger (approved):
Thanks for working on this!

pacman100:
cc @stas00

src/accelerate/accelerator.py
```diff
@@ -1704,7 +1704,7 @@ def save_state(self, output_dir: str = None):
         self.project_configuration.iteration += 1
         return save_location
 
-    def load_state(self, input_dir: str):
+    def load_state(self, input_dir: str, **model_load_kwargs):
```
muellerzr, line 1707:
```suggestion
    def load_state(self, input_dir: str, **load_func_kwargs):
```

I think we should rename this to `load_func_kwargs` which is a bit more explicit in what these are actually going towards.

src/accelerate/accelerator.py
```diff
@@ -1717,6 +1717,9 @@ def load_state(self, input_dir: str):
         Args:
             input_dir (`str` or `os.PathLike`):
                 The name of the folder all relevant weights and states were saved in.
+            model_load_kwargs (`dict`):
+                Additional keyword arguments required for loading model, e.g., deepspeed's optional args for
+                `load_checkpoint`.
```
muellerzr, line 1722:
```suggestion
                Additional keyword arguments required for loading model which are passed to the underlying load function, such as optional arguments for `DeepSpeed`'s `load_checkpoint` function.
```

Let's be a bit more explicit here

src/accelerate/checkpointing.py
```diff
@@ -126,12 +126,13 @@ def load_accelerator_state(input_dir, models, optimizers, schedulers, process_in
             The current process index in the Accelerator state
         scaler (`torch.cuda.amp.GradScaler`, *optional*):
             An optional *GradScaler* instance to load
+        model_load_kwargs (`dict`): Additional arguments to pass to the model's `load_state_dict` method.
```
muellerzr, line 129: Also bring the earlier suggestions into this section (including the spacing Sylvain suggested earlier)

muellerzr (approved):
Nice addition! Just a few requests for naming and documentation clarity

muellerzr (approved):
Let's also try to add a test or three (since we have FSDP, DeepSpeed, and Megatron) to ensure that everything is working properly if possible :)

src/accelerate/accelerator.py
pacman100: hello, but `load_func` can mean optimizer/scheduler loading too, and hence chose this name.

muellerzr: True, `load_model_func_kwargs` perhaps then as a medium

stas00:
Thank you very much for adding this functionality, Sourab et al.

I don't really need it right now, but do you want to add the `save_checkpoint` kwargs for symmetry? e.g. deepspeed's is:

```
    def save_checkpoint(self, save_dir, tag=None, client_state={}, save_latest=True):
        r"""Save training checkpoint
        Arguments:
            save_dir: Required. Directory for saving the checkpoint
            tag: Optional. Checkpoint tag used as a unique identifier for the checkpoint, global step is
                used if not provided. Tag name must be the same across all ranks.
            client_state: Optional. State dictionary used for saving required training states in the client code.
            save_latest: Optional. Save a file 'latest' pointing to the latest saved checkpoint.
```

but as I said this is totally a suggestion and not a need.

Thank you!

src/accelerate/accelerator.py

src/accelerate/accelerator.py
```diff
@@ -1634,6 +1634,9 @@ def save_state(self, output_dir: str = None):
         Args:
             output_dir (`str` or `os.PathLike`):
                 The name of the folder to save all relevant weights and states.
+            save_model_func_kwargs (`dict`, *optional*):
+                Additional keyword arguments for saving model which can be passed to the underlying save function, such
+                as optional arguments for `DeepSpeed`'s `save_checkpoint` function.
```
stas00, line 1639:
```suggestion
                as optional arguments for DeepSpeed's `save_checkpoint` function.
```

either  `deepspeed` (package) or  DeepSpeed (name / no backquotes)

src/accelerate/accelerator.py
```diff
@@ -1717,6 +1720,9 @@ def load_state(self, input_dir: str):
         Args:
             input_dir (`str` or `os.PathLike`):
                 The name of the folder all relevant weights and states were saved in.
+            load_model_func_kwargs (`dict`, *optional*):
+                Additional keyword arguments for loading model which can be passed to the underlying load function,
+                such as optional arguments for `DeepSpeed`'s `load_checkpoint` function.
```
stas00, line 1725:
```suggestion
                such as optional arguments for DeepSpeed's `load_checkpoint` function.
```

stas00 (approved):
Thank you, Sourab - looking great.

I left one small suggestion (2x)

muellerzr (approved):
Great!

PR #990, opened 2023-01-19
Add new release_memory util

muellerzr:
Helps https://github.com/huggingface/transformers/issues/21094 by introducing a lower-level convenience API for releasing memory. It's quite similar to `Accelerator.free_memory` but let's the user pass in a list of objects to remove.

Example usage:

```python
        >>> import torch
        >>> from accelerate.utils import release_memory

        >>> a = torch.ones(1000, 1000).cuda()
        >>> b = torch.ones(1000, 1000).cuda()
        >>> a, b = release_memory(a, b)
```

Extracting it outside into a new util let's it be used by more usecases, such as big model inference

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import contextlib
-import gc
 import math
 import os
 import shutil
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Thanks for adding this!

muellerzr:
@asyu17 it won’t quite work with model offloading iirc, there’s some issues we’re still working out but torch won’t free that memory

Issue #976, opened 2023-01-16
[Discussion] `save_state` / `load_state` don't use custom saving/loading methods.


patrickvonplaten:
`save_state` currently serializes models always as a simple `pytorch_model.bin` file, but no additional metadata is saved such as the model config.

Could it make sense to add functionality to `save_state` and `load_state` that would allow to customize saving & loading more? e.g. allow models to be loaded and saved with their unique `save_pretrained` and `from_pretrained(...)` methods? 

The problem we're currently experiencing in `diffusers` is that intermediate saved checkpoints are difficult to load & use for inference as explained here: https://github.com/huggingface/diffusers/issues/732#issuecomment-1383006354 and shown here: https://github.com/huggingface/diffusers/blob/main/docs/source/en/training/dreambooth.mdx#performing-inference-using-a-saved-checkpoint

During training we save both a `transformers` and `diffusers` model without a config simply under `"pytorch_model{i}.bin"` which then makes it quite hard / unintuitive to load those checkpoints afterward for inference (one has to run this code: https://github.com/huggingface/diffusers/blob/main/docs/source/en/training/dreambooth.mdx#performing-inference-using-a-saved-checkpoint).

Currently, I don't see a clean & easy solution though to add this to `save_state` so maybe it'd be best to just tell users to use directly the customized `save_pretrained(...)` method? We could possible add a mapping such as the following to `accelerate`: https://github.com/huggingface/diffusers/blob/b2ea8a84e9db71c207d28687722d241447bde81e/src/diffusers/pipelines/pipeline_utils.py#L70 to allow for customized loading and saving.

This way we could maybe get the best of both worlds: Leverage `load/save_state` for all saving/loading in the training script **and** create an easy to re-use folder structure when saving? 

Wdyt @sgugger ?

patrickvonplaten:
Also cc @pcuenca FYI

sgugger:
I have nothing against having a mapping like the one you mention that would use `save_pretrained`/`from_pretrained` for diffusers/transformers models in the `save_state` method (and we can support third-party special functions the same way if users want it). It looks like the easiest way to get the best of both worlds.

pacman100:
[...]

PR #991, opened 2023-01-19
Saving and loading state hooks

patrickvonplaten:
With this design we would have the necessary freedom to save & load models as wished in `diffusers`. 

Possible solution to: https://github.com/huggingface/accelerate/issues/976
The idea would be for loading:
- loop over the `models` and if a certain type of model, e.g. UNet, we call `from_pretrained(...)` and remove it from the list so it's not called after.

For saving:
- loop over the passed `self._models` and passed `weights` and if `self_models` if of certain type (e.g. UNet) we call `save_pretrained(...)` and remove only from the weights list so it's not called after. 

Design is heavily inspired by the hooks in: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module

If this is ok, happy to write tests and better docs and try it out. 

cc @sgugger @pcuenca

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1637,7 +1637,7 @@ class Accelerator:
 
         <Tip>
 
-        Should only be used in conjunction with [`Accelerator.register_save_state_pre_hook`]. Can be useful to save
+        Should only be used in conjunction with [`Accelerator.register_load_state_pre_hook`]. Can be useful to save
         configurations in addition to model weights. Can also be used to overwrite model saving with a customized
         method. In this case, make sure to remove already loaded weights from the weights list.
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -21,7 +21,7 @@ import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
 from functools import wraps
-from typing import List, Optional, Union, Callable
+from typing import Callable, List, Optional, Union
 
 import torch
 import torch.utils.hooks as hooks
@@ -1620,6 +1620,29 @@ class Accelerator:
         save(obj, f)
 
     def register_save_state_pre_hook(self, hook: Callable[..., None]) -> hooks.RemovableHandle:
+        """
+        Registers a pre hook to be run before `save_checkpoint` is called in [`Accelerator.save_state`].
+
+        Args:
+            hook (`Callable`):
+                A function to be called in [`Accelerator.save_state`] before `save_checkpoint`. The hook should have
+                the following signature: \n `hook(models: List[torch.nn.Module], weigths: List[Dict[str,
+                torch.Tensor]], input_dir: str) -> None` \n The `models` argument are the models as saved in the
+                accelerator state under `accelerator._models`, `weigths` argument are the state dicts of the `models`,
+                and the `input_dir` argument is the `input_dir` argument passed to [`Accelerator.load_state`].
+
+        <Tip>
+
+        Should only be used in conjunction with [`Accelerator.register_save_state_pre_hook`]. Can be useful to save
+        configurations in addition to model weights. Can also be used to overwrite model saving with a customized
+        method. In this case, make sure to remove already loaded weights from the weights list.
+
+        </Tip>
+
+        Returns:
+            `torch.utils.hooks.RemovableHandle`: a handle that can be used to remove the added hook by calling
+            `handle.remove()`
+        """
         handle = hooks.RemovableHandle(self._save_model_state_pre_hook)
         self._save_model_state_pre_hook[handle.id] = hook
         return handle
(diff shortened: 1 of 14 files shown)


src/accelerate/accelerator.py
```diff
@@ -18,11 +18,14 @@
...
 import warnings
+from collections import OrderedDict
 from contextlib import contextmanager
 from functools import wraps
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Callable
 
 import torch
+import torch.utils.hooks as hooks
+from ...utils.hooks import RemovableHandle
```
sgugger, line 28: I don't think this object exists in accelerate.

sgugger (commented):
I'd be interested to see how it solves your problem in practice: the existing code to load/save is still executed so I'm not sure how this actually helps. I can review more the design afterward (though I have no objection right now, just nits).

patrickvonplaten:
The use case would then look like this in diffusers: https://github.com/huggingface/diffusers/pull/2048/files . Because we're doing loading and saving directly in the hooks, the weights and models have to be popped from the respective lists so that nothing is loaded / saved twice. 

Think there are also use cases where the model doesn't need to be popped, e.g. if someone wants to save the config for transformers in addition to the checkpoint the hooks could just save and load the config but then let accelerate do the saving / loading of the weights. 

Think this design is quite minimal for accelerate. Wdyt? 
Happy to add a test to accelerate that shows how one could use the hooks to save a config of a transformer model in addition to saving/loading the weights.

src/accelerate/accelerator.py
```diff
@@ -1613,6 +1619,11 @@ def save(self, obj, f):
         """
         save(obj, f)
 
+    def register_save_state_pre_hook(self, hook: Callable[..., None]) -> hooks.RemovableHandle:
+        handle = hooks.RemovableHandle(self._save_model_state_pre_hook)
+        self._save_model_state_pre_hook[handle.id] = hook
+        return handle
+
```
pacman100, line 1626: Can we combine `register_save_state_pre_hook` and `register_load_state_pre_hook` using a bool parameter `load_state`

```suggestion
    def register_pre_hook(self, hook: Callable[..., None], load_state: bool) -> hooks.RemovableHandle:
        if load_state:
            pre_hook = self._load_model_state_pre_hook
        else:
            pre_hook = self._save_model_state_pre_hook
        handle = hooks.RemovableHandle(pre_hook)
        pre_hook[handle.id] = hook
        return handle

```

pacman100 (commented):
Great way to enable custom saving and loading, pretty neat! ✨

Left a comment.

src/accelerate/accelerator.py
muellerzr: @pacman100 WDYT of instead mimicking how we have `prepare` and have a `register_state_pre_hook` func that calls internally `register_save_state_pre_hook` or `register_load_state_pre_hook`? I also think the explicit naming on `save` vs `load` pre-hook is important here for users to understand what it does, but have `save_state_pre_hook` for convenience
potentially.

tests/test_accelerator.py
```diff
@@ -49,3 +61,83 @@ def test_free_memory_dereferences_prepared_components(self):
         self.assertTrue(len(accelerator._schedulers) == 0)
         self.assertTrue(len(accelerator._dataloaders) == 0)
         AcceleratorState._reset_state()
+
+    def test_save_load_model(self):
```
patrickvonplaten, line 65: This test is a bit unrelated to the PR, but think this is a useful test for the accelerator class in general.

patrickvonplaten:
Pretty sure test failures are unrelated. Would be great to get another review here :-)

src/accelerate/accelerator.py
```diff
@@ -1613,6 +1619,38 @@ def save(self, obj, f):
...
+        `hook(models: List[torch.nn.Module], weights: List[Dict[str, torch.Tensor]], input_dir: str) -> None`
+
+        The `models` argument are the models as saved in the accelerator state under `accelerator._models`, `weigths`
+        argument are the state dicts of the `models`, and the `input_dir` argument is the `input_dir` argument passed
+        to [`Accelerator.load_state`].
+
+        <Tip>
+
+        Should only be used in conjunction with [`Accelerator.register_save_state_pre_hook`]. Can be useful to save
```
sgugger, line 1640:
```suggestion
        Should only be used in conjunction with [`Accelerator.register_load_state_pre_hook`]. Can be useful to save
```

sgugger (approved):
Nice, thanks a lot for iterating on this!

muellerzr:
@patrickvonplaten only was now able to look at this today, I can confirm that your PR does indeed break that test, it works prior to this and is only on the CPU when it fails. If you have some time please try to look into it otherwise I'll look into this later today. cc @sgugger

Issue #963, opened 2023-01-10
Scheduler always steps when training with gradient accumulation


markovalexander:
### System Info

```Shell
- Accelerate version: 0.15.0
- Platform: macOS-10.16-x86_64-i386-64bit
- Python version: 3.8.11
- Numpy version: 1.23.0
- PyTorch version (GPU?): 1.9.0 (False)
- Accelerate default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

When training a model with gradient accumulation scheduler modifies learning rate at every step, even when `optimiser.step` was not performed.

To reproduce:
1. Add debugging prints 🙂  [here](https://github.com/huggingface/accelerate/blob/9b7ef9679f3a3d1c3e579562568f6fc637e62091/src/accelerate/optimizer.py#L129) and [here](https://github.com/huggingface/accelerate/blob/9b7ef9679f3a3d1c3e579562568f6fc637e62091/src/accelerate/scheduler.py#L53) after `.step()` calls to track them
2. Run the following ([pastebin](https://pastebin.com/MZimGGwc)) script with `accelerate launch test.py`
```
from accelerate import Accelerator
from transformers import (
    get_cosine_schedule_with_warmup,
)
import torch
 
accelerator = Accelerator(
    device_placement=True,
    gradient_accumulation_steps=4,
    split_batches=True,
)
 
model = torch.nn.Linear(1, 1)
optimizer = torch.optim.AdamW(model.parameters())
scheduler = get_cosine_schedule_with_warmup(optimizer, num_warmup_steps=10, num_training_steps=50)
model, optimizer, scheduler = accelerator.prepare(model, optimizer, scheduler)
 
for i in range(50):
    with accelerator.accumulate(model):
        print('----------------------------- batch -----------------------------')
        optimizer.zero_grad()
        loss = model(torch.randn(10, 1))
        accelerator.backward(loss.mean())
 
        optimizer.step()
        scheduler.step()
```
Output: 
![telegram-cloud-photo-size-2-5325883662245414381-y](https://user-images.githubusercontent.com/22663468/211563928-bdb5093a-7817-4e46-a499-801f618756fb.jpg)

### Expected behavior

[...]

PR #999, opened 2023-01-23
Fix scheduler incorrect steps when gradient accumulation enabled

markovalexander:
Fix https://github.com/huggingface/accelerate/issues/963 issue

### DIFF
--- a/src/accelerate/test_utils/scripts/test_sync.py
+++ b/src/accelerate/test_utils/scripts/test_sync.py
@@ -202,11 +202,13 @@ def test_gradient_accumulation_with_opt_and_scheduler(split_batches=False, dispa
         ddp_model.train()
         step_model(model, input, target, accelerator, False)
         opt.step()
-        if split_batches:
-            sched.step()
-        else:
-            for _ in range(accelerator.num_processes):
+
+        if ((iteration + 1) % 2 == 0) or ((iteration + 1) == len(dataloader)):
+            if split_batches:
                 sched.step()
+            else:
+                for _ in range(accelerator.num_processes):
+                    sched.step()
         opt.zero_grad()
         # Perform gradient accumulation under wrapper
         with accelerator.accumulate(ddp_model):
--- a/src/accelerate/scheduler.py
+++ b/src/accelerate/scheduler.py
@@ -16,7 +16,7 @@
 
 import warnings
 
-from .state import AcceleratorState
+from .state import AcceleratorState, GradientState
 
 
 warnings.filterwarnings("ignore", category=UserWarning, module="torch.optim.lr_scheduler")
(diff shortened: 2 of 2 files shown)


src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -202,11 +202,13 @@ def test_gradient_accumulation_with_opt_and_scheduler(split_batches=False, dispa
...
-            sched.step()
-        else:
-            for _ in range(accelerator.num_processes):
+
+        if ((iteration + 1) % 2 == 0) or ((iteration + 1) == len(dataloader)):
+            if split_batches:
                 sched.step()
+            else:
+                for _ in range(accelerator.num_processes):
+                    sched.step()
```
sgugger, line 211: Could you explain he logic here?

sgugger (commented):
Thanks for your PR. The firs half LGTM as discussed on the issue, but I don't fully understand the changes on the test file. Could you elaborate?

src/accelerate/test_utils/scripts/test_sync.py
markovalexander: Yes, basically I added if-statement to check  that optimizer step was made, as in line 224: `did_step = (((iteration + 1) % 2) == 0) or ((iteration + 1) == len(dataloader)) `. Because AcceleratedScheduler now steps every `gradient_accumulation_steps` steps, I added same logic to the ground truth scheduler. May be I should have moved `opt.step()` under this if statement too.

sgugger (approved):
Thanks for the explanation! @muellerzr can you also have a look here?

muellerzr (approved):
Thanks! This all looks good to me as well. I'll keep an eye post-merge to make sure our multi-GPU tests all pass (as those are only ran on merges/nightly) but great work @markovalexander! Thanks! 🤗

yjkim3:
From this commit therefore,

Do I don't need to set the steps of scheduler with calculating on my own?
Is scheduler steps internally adjusted? (or Is sheduler.step() runed every steps in regardless to gradient_accumulation_steps)

If I try to configure below,
training steps = 1000
scheduler steps = 1000
gradient_accumulation_steps = 4

then, is the scheduler.step() runed total 250 steps (1000 / 4)?

image

muellerzr:
Hi!

> Is scheduler.step() run every steps in regardless to gradient_accumulation_steps
No it's run only on steps that actually "step", and then it is run `n` times for `n` GPUs. 

It will still run `1000` steps, we only step the scheduler when we step, as I mentioned before

yjkim3:
> Hi!
> 
> > Is scheduler.step() run every steps in regardless to gradient_accumulation_steps
> > No it's run only on steps that actually "step", and then it is run `n` times for `n` GPUs.
> 
> It will still run `1000` steps, we only step the scheduler when we step, as I mentioned before (aka update the weights on all processes/synchronize)

Thanks for explanation :)

PR #1001, opened 2023-01-23
Start of adding examples

muellerzr:
Over the next few days I will be adding more examples into the docstrings of major functions called within `accelerate` that are not there. This is to help in anticipation for the 1.0.0 release and to have a well-documented codebase along with a well-documented library as a whole.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -564,7 +564,7 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.main_process_first():
-        ...     # This will be printed by process 0 then in a seemingly
+        ...     # This will be printed first by process 0 then in a seemingly
         ...     # random order by the other processes.
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -564,6 +564,7 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.main_process_first():
+        ...     # This will be printed by process 0 then by process 1, etc.
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -556,6 +556,16 @@ def main_process_first(self):
...
         The other processes will enter the with block after the main process exits.
+
+        Example:
+
+        ```python
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> with accelerator.main_process_first():
+        ...     print(f"This will be printed by process {accelerator.process_index}")
```
sgugger, line 567: Maybe put something like "This will only be printed on process 0."?

src/accelerate/accelerator.py
```diff
@@ -565,6 +575,16 @@ def local_main_process_first(self):
...
         The other processes will enter the with block after the main process exits.
+
+        Example:
+
+        ```python
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> with accelerator.local_main_process_first():
+        ...     print(f"This will be printed by process {accelerator.local_process_index}")
```
sgugger, line 586: Same here.

sgugger (approved):
Thanks for adding those!

src/accelerate/accelerator.py
muellerzr: Since it's `main_process_first` and `local_process_first`, it will start with 0 and go up. I added a comment in the example stating something similar.

src/accelerate/accelerator.py
muellerzr: See response to other comment

src/accelerate/accelerator.py
sgugger: It's actually a bit more complicated: it will be printed on process 0 first, then in a (seemingly) random order by the others.

muellerzr: Good point, will update stating something similar

sgugger (approved):

PR #1002, opened 2023-01-23
Efficiently skip batches in a dataloader

sgugger:
This PR adds a new function to efficiently skip batches in a given DataLoader when the underlying dataset is not an `IterableDataset`. It works by replacing the `batch_sampler` with one that skips the first n batches, since iterating though the batch sampler is fast.

Missing tests and examples for now.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -25,7 +25,7 @@ from typing import List, Optional, Union
 import torch
 
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
-from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_batches
+from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_first_batches
 from .logging import get_logger
 from .optimizer import AcceleratedOptimizer
 from .scheduler import AcceleratedScheduler
@@ -1919,7 +1919,7 @@ class Accelerator:
                 return True
         return False
 
-    def skip_batches(self, dataloader, num_batches: int = 0):
+    def skip_first_batches(self, dataloader, num_batches: int = 0):
         """
         Creates a new `torch.utils.data.DataLoader` that will efficiently skip the first `num_batches`.
 
@@ -1927,4 +1927,4 @@ class Accelerator:
             dataloader (`torch.utils.data.DataLoader`): The data loader in which to skip batches.
             num_batches (`int`, *optional*, defaults to 0): The number of batches to skip
         """
-        return skip_batches(dataloader, num_batches=num_batches)
+        return skip_first_batches(dataloader, num_batches=num_batches)
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -25,7 +25,7 @@ from typing import List, Optional, Union
 import torch
 
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
-from .data_loader import DataLoaderDispatcher, prepare_data_loader
+from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_batches
 from .logging import get_logger
 from .optimizer import AcceleratedOptimizer
 from .scheduler import AcceleratedScheduler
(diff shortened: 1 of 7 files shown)


src/accelerate/accelerator.py
```diff
@@ -1918,3 +1918,13 @@ def optimizer_step_was_skipped(self):
...
                 return True
         return False
+
+    def skip_batches(self, dataloader, num_batches: int = 0):
+        """
+        Creates a new `torch.utils.data.DataLoader` that will efficiently skip the first `num_batches`.
+
+        Args:
+            dataloader (`torch.utils.data.DataLoader`): The data loader in which to skip batches.
+            num_batches (`int`, *optional*, defaults to 0): The number of batches to skip
```
muellerzr, line 1928: Docstring format nit
```suggestion
            dataloader (`torch.utils.data.DataLoader`): 
                The data loader in which to skip batches.
            num_batches (`int`, *optional*, defaults to 0): 
                The number of batches to skip
```

src/accelerate/data_loader.py
```diff
@@ -340,8 +340,10 @@ class DataLoaderShard(DataLoader):
             - `"cuda"`: the CUDA random number generator (GPU only)
             - `"xla"`: the XLA random number generator (TPU only)
             - `"generator"`: an optional `torch.Generator`
-        generator (`torch.Generator`, *optional*):
+        synchronized_generator (`torch.Generator`, *optional*):
             A random number generator to keep synchronized across processes.
+        split_batches (`int`, *optional*, defaults to 0):
+            The number of batches to skip at the beginning.
```
muellerzr, line 346:
```suggestion
        skip_n_batches (`int`, *optional*, defaults to 0):
            The number of batches to skip at the beginning.
```

I think having `skip_n_batches` instead would be better on the user as it's more descriptive.

src/accelerate/data_loader.py
```diff
@@ -354,11 +356,12 @@ class DataLoaderShard(DataLoader):
         - **total_dataset_length** (`int`) -- Total length of the inner dataset across all processes.
     """
 
-    def __init__(self, dataset, device=None, rng_types=None, synchronized_generator=None, **kwargs):
+    def __init__(self, dataset, device=None, rng_types=None, synchronized_generator=None, skip_batches=0, **kwargs):
```
muellerzr, line 359:
```suggestion
    def __init__(self, dataset, device=None, rng_types=None, synchronized_generator=None, skip_n_batches=0, **kwargs):
```

Suggestion based on previous comment

src/accelerate/data_loader.py
```diff
@@ -354,11 +356,12 @@ class DataLoaderShard(DataLoader):
...
         super().__init__(dataset, **kwargs)
         self.device = device
         self.rng_types = rng_types
         self.synchronized_generator = synchronized_generator
+        self.skip_batches = skip_batches
```
muellerzr, line 364:
```suggestion
        self.skip_n_batches = skip_n_batches
```
Suggestion based on previous comment

src/accelerate/data_loader.py
```diff
@@ -407,16 +415,19 @@ def total_dataset_length(self):
...
+    Args:
         split_batches (`bool`, *optional*, defaults to `False`):
             Whether the resulting `DataLoader` should split the batches of the original data loader across devices or
             yield full batches (in which case it will yield batches starting at the `process_index`-th and advancing of
             `num_processes` batches at each iteration). Another way to see this is that the observed batch size will be
             the same as the initial `dataloader` if this option is set to `True`, the batch size of the initial
             `dataloader` multiplied by `num_processes` otherwise. Setting this option to `True` requires that the batch
             size of the `dataloader` is a round multiple of `batch_size`.
+        skip_batches (`int`, *optional*, defaults to 0):
+            The number of batches to skip at the beginning.
```
muellerzr, line 430:
```suggestion
            The number of batches to skip at the beginning of an iteration.
```

src/accelerate/data_loader.py
```diff
@@ -747,3 +762,119 @@ def prepare_data_loader(
     if state.distributed_type == DistributedType.TPU:
         return MpDeviceLoaderWrapper(dataloader, device)
     return dataloader
+
+
+class SkipBatchSampler(BatchSampler):
+    """
+    A `torch.utils.data.BatchSampler` that skips the first batches of another `torch.utils.data.BatchSampler`.
```
muellerzr, line 769:
```suggestion
    A `torch.utils.data.BatchSampler` that skips the first `n` batches of another `torch.utils.data.BatchSampler`.
```

muellerzr (approved):
Looks great!

Main feedback is I think we should call it `skip_n_batches` to be more descriptive.

Nice work 🤗

src/accelerate/data_loader.py
sgugger: Not really fan of the longer name, and those are not really public classes anyway (in the sense that they are not in the init) so will keep it as is.

muellerzr (approved):
Great!

PR #1008, opened 2023-01-25
Update toctree

muellerzr:
Reorganizes everything slightly so it reads better top -> down when looking at the TOC tree. I'll write a Basic tutorial for big model inference so it can be featured in the basic tutorial section as otherwise it's a bit buried.


sgugger (approved):
Thanks!

PR #1009, opened 2023-01-25
Allow the torch device to be set with an env var

Yard1:
Signed-off-by: Antoni Baum <antoni.baum@protonmail.com>

# Why

Currently, `AccelerateState` will always use the Torch device corresponding to the local rank. This is fine in most cases, but may be limiting for advanced uses (eg. fractional GPUs) or compatibility with other libraries which manage resources themselves like Ray Train.

# How

This PR adds logic to first derive the device from `ACCELERATE_TORCH_DEVICE` env var and falling back to local rank if that env var is not set. As this is an advanced use case, no documentation changes are introduced and there is no expectation that most users will make use of it.

# Notes

I wasn't sure if this required tests and if so, what file they should be put into. Would appreciate some pointers here :)

This would make usage of Accelerate on Ray Train less error-prone.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -92,6 +92,8 @@ class AcceleratorState:
             env_device = os.environ.get("ACCELERATE_TORCH_DEVICE", None)
             if env_device:
                 self.device = torch.device(env_device)
+            else:
+                self.device = None
             self.dynamo_backend = DynamoBackend(dynamo_backend.upper())
             if not _from_accelerator:
                 raise ValueError(
@@ -113,7 +115,7 @@ class AcceleratorState:
                     self.num_processes = torch.distributed.get_world_size()
                     self.process_index = torch.distributed.get_rank()
                     self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
-                    if not env_device:
+                    if not self.device:
                         self.device = torch.device("cuda", self.local_process_index)
                     torch.cuda.set_device(self.device)
                     self._mixed_precision = mixed_precision
@@ -152,7 +154,7 @@ class AcceleratorState:
                 self.num_processes = torch.distributed.get_world_size()
                 self.process_index = torch.distributed.get_rank()
                 self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
-                if not env_device:
+                if not self.device:
                     self.device = torch.device("cuda", self.local_process_index)
                 torch.cuda.set_device(self.device)
                 self._mixed_precision = "no"  # deepspeed handles mixed_precision using deepspeed_config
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
I think this all makes sense to me and looks alright. @Yard1 can you test on two GPUs to ensure that everything still works? As I can't run our multi-GPU CI until after it's merged in this case.

Yard1:
Will do and get back to you

Yard1:
I spotted one issue. Fixing!

src/accelerate/state.py
```diff
@@ -89,6 +89,7 @@ def __init__(
             dynamo_backend = (
                 parse_choice_from_env("ACCELERATE_DYNAMO_BACKEND", "no") if dynamo_backend is None else dynamo_backend
             )
+            env_device = os.environ.get("ACCELERATE_TORCH_DEVICE", None)
```
muellerzr, line 92: This can be refactored as:

```suggestion
            env_device = os.environ.get("ACCELERATE_TORCH_DEVICE", None)
            if  env_device is not None:
                self.device = torch.device(env_device)
```

And then in each `self.device` check you can reduce it to

```python
if env_device is None:
   self.device = ...
```

muellerzr (commented):
Looking better! We can simplify this logic a bit though to make it easier to read :)

src/accelerate/state.py
sgugger: Or `self.device = torch.device(env_device) if env_device is not None else None` here and then all checks are
```
if self.device is None:
    self.device = ...
```

sgugger (commented):
I agree with the generic idea behind the PR, let's just simplify a bit the code as suggested.

Could you just add one quick test that double checks the env variable takes priority over the default device?

src/accelerate/state.py
muellerzr: Agreed :) Was debating back and forth on that option, but if you were thinking that as well I'm in agreement

Yard1: Sounds good!

Yard1:
Sure thing @sgugger. Can you let me know where it should be added? I was unsure which test file tests `AcceleratorState` behavior.

src/accelerate/state.py
```diff
@@ -89,6 +89,11 @@ def __init__(
...
+            if env_device:
+                self.device = torch.device(env_device)
+            else:
+                self.device = None
```
muellerzr, line 96: Let's use Sylvains suggestion of:

```suggestion
            self.device = torch.device(env_device) if env_device is not None else None
```

And then in all the checks do `if self.device is None: ...`

muellerzr:
@Yard1 these exist in `accelerate/src/test_utils/scripts/test_script.py` and are a pytest-like test. You would add a new function to it utilizing the `Accelerator()` to create the new test.

Yard1:
@muellerzr @sgugger I have added a test and ran the full test suite without issues on an instance with 2 GPUs.

sgugger:
Looks like there is an error in the tests:
```
        with patch("torch.cuda.set_device", noop), patch_environment(ACCELERATE_TORCH_DEVICE="cuda:1234"):
            accelerator = Accelerator()
>           self.assertEqual(str(accelerator.state.device), "cuda:1234")
E           AssertionError: 'cuda:-46' != 'cuda:1234'
E           - cuda:-46
E           + cuda:1234
```

Yard1:
Hmm, it was passing for me locally... maybe a dependency version mismatch or something. Will check

muellerzr:
Thanks for iterating @Yard1! Nice work 🤗

PR #1010, opened 2023-01-25
More improvements to docstrings + examples

muellerzr:
Includes the rest of the docstring examples where needed on the major API parts of the codebase.

**Should not be merged until after** #1002

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -564,7 +564,7 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.main_process_first():
-        ...     # This will be printed by process 0 then in a seemingly
+        ...     # This will be printed first by process 0 then in a seemingly
         ...     # random order by the other processes.
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -564,6 +564,7 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.main_process_first():
+        ...     # This will be printed by process 0 then by process 1, etc.
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
@@ -583,6 +584,7 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.local_main_process_first():
+        ...     # This will be printed by local process 0 then by process 1, etc.
         ...     print(f"This will be printed by process {accelerator.local_process_index}")
         ```
         """
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -564,7 +564,8 @@ class Accelerator:
 
         >>> accelerator = Accelerator()
         >>> with accelerator.main_process_first():
-        ...     # This will be printed by process 0 then by process 1, etc.
+        ...     # This will be printed by process 0 then in a seemingly
+        ...     # random order by the other processes.
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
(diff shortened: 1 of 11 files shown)


src/accelerate/big_modeling.py
```diff
@@ -380,6 +380,24 @@ def load_checkpoint_and_dispatch(
...
+
+    >>> # Create a model and initialize it
+    >>> checkpoint = "EleutherAI/gpt-j-6B"
+    >>> config = AutoConfig.from_pretrained(checkpoint)
+    >>> with init_empty_weights():
+    ...     model = AutoModelForCausalLM.from_config(config)
+
+    >>> # Load the checkpoint and dispatch it to the right devices
+    >>> model = load_checkpoint_and_dispatch(
+    ...     model, "sharded-gpt-j-6B", device_map="auto", no_split_module_classes=["GPTJBlock"]
```
sgugger, line 398: The second argument needs to point to the location in the cache where the checkpoint is. You should probably use `hf_hub_download` to download it.

src/accelerate/accelerator.py
```diff
@@ -1445,6 +1570,21 @@ def gather_for_metrics(self, tensor):
...
+        >>> # Assuming two processes, with a batch size of 5 on a dataset with 9 samples
+        >>> import torch
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> dataloader = torch.utils.data.DataLoader(range(9), batch_size=5)
+        >>> dataloader = accelerator.prepare(dataloader)
+        >>> batch = next(iter(dataloader))
+        >>> gathered_items = accelerator.gather_for_metrics(batch)
+        >>> assert len(gathered_items) == 9
```
sgugger, line 1586:
```suggestion
        >>> len(gathered_items)
        9
```
 The goal of this syntax is to be able to show results.

src/accelerate/accelerator.py
```diff
@@ -1483,7 +1623,22 @@ def reduce(self, tensor, reduction="sum"):
...
+        ```python
+        >>> # Assuming two processes
+        >>> import torch
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> process_tensor = torch.arange(accelerator.num_processes) + 1 + (2 * accelerator.process_index)
+        >>> process_tensor = process_tensor.to(accelerator.device)
+        >>> reduced_tensor = accelerator.reduce(process_tensor, reduction="sum")
+        >>> assert reduced_tensor == torch.tensor([4, 6])
```
sgugger, line 1640: Same here.

src/accelerate/accelerator.py
```diff
@@ -1501,6 +1656,23 @@ def pad_across_processes(self, tensor, dim=0, pad_index=0, pad_first=False):
...
+
+        ```python
+        >>> # Assuming two processes, with the first processes having a tensor of size 1 and the second of size 2
+        >>> import torch
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> process_tensor = torch.arange(accelerator.process_index + 1).to(accelerator.device)
+        >>> padded_tensor = accelerator.pad_across_processes(process_tensor)
+        >>> assert padded_tensor.shape == torch.Size([2])
```
sgugger, line 1674: And there.

src/accelerate/accelerator.py
```diff
@@ -1514,13 +1686,47 @@ def unwrap_model(self, model, keep_fp32_wrapper: bool = False):
...
+        ```python
+        >>> # Assuming two GPU processes
+        >>> from torch.nn.parallel import DistributedDataParallel
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> model = accelerator.prepare(MyModel())
+        >>> assert isinstance(model, DistributedDataParallel)
+        >>> model = accelerator.unwrap_model(model)
+        >>> assert not isinstance(model, DistributedDataParallel)
```
sgugger, line 1704: Remove asserts and just ptint the class name.

sgugger (commented):
Looking good. Make sure to take advantage of the doctest syntax to avoid asserts in the code examples.

src/accelerate/big_modeling.py
muellerzr: Noted, got it working. The official docs also need to be updated with this. Will do so in a separate PR

sgugger (approved):
Thanks a lot!

PR #1011, opened 2023-01-25
Maintain accumulation steps environment variable

muellerzr:
I noticed the `DeepSpeed` config was setting an environment variable of `ACCELERATE_GRADIENT_ACCUMULATION_STEPS`. This PR maintains consistency and lets the user have that specification when using the `Accelerator` if desired.


sgugger (approved):
Thanks!

PR #1013, opened 2023-01-26
Include steppage in performance docs

muellerzr:
Based on this [discussion](https://discuss.huggingface.co/t/learning-rate-scheduler-distributed-training/30453) writes a small note in the performance doc about how `accelerate` will step the scheduler multiple times and how the "batch size" is treated as a result

### DIFF
--- a/docs/source/concept_guides/performance.mdx
+++ b/docs/source/concept_guides/performance.mdx
@@ -89,9 +89,6 @@ learning_rate *= accelerator.num_processes
 optimizer = AdamW(params=model.parameters(), lr=learning_rate)
 ```
 
-You will also find that `accelerate` will step the learning rate based upon the number of processes, so if you have a learning rate scheduler
-that steps every 1000 steps, it will step every 1000 * `num_processes` steps. You should adjust the number of steps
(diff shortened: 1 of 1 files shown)


docs/source/concept_guides/performance.mdx
```diff
@@ -89,3 +89,9 @@ learning_rate *= accelerator.num_processes
 optimizer = AdamW(params=model.parameters(), lr=learning_rate)
 ```
 
+You will also find that `accelerate` will step the learning rate based upon the number of processes, so if you have a learning rate scheduler
+that steps every 1000 steps, it will step every 1000 * `num_processes` steps. You should adjust the number of steps
```
sgugger, line 93: No there is no adjusting to be done that is the whole point. This is to account for the fact the dataloader length has been divided by `num_processes`.

muellerzr: @sgugger I changed this to:

You will also find that `accelerate` will step the learning rate based on the number of processes being trained on. This is because 
of the observed batch size noted earlier. So in a case of 2 GPUs, the learning rate will be stepped twice as often as a single GPU
to account for the batch size being twice as large (if no changes to the batch size on the single GPU instance are made).

That better? 😄

PR #1015, opened 2023-01-26
Deepspeed param check

dhar174:
On line 146, in set_module_tensor_to_device(), adding a check for deepspeed parameters in the kwargs object, and not passing them solved the error I was receiving regarding the ds parameters not being recognized by torch.nn.Parameter.__new__(). With my admittedly limited knowledge, it seemed to me that the kwargs are not necessary to pass in the case of using Deepspeed+ Accelerate, and this bears out since the model loaded fine with zero-3 cpu parameter and buffer offload on a single-GPU machine, and performed perfectly comprehensible inference outputs (slowly) using the GPU.

The error, in my case, was occurring here as called from accelerator's dispatch_model().

Please let me know if my thinking on this is in anyway wrong! This fix worked for me. 

 `transformers` version: 4.26.0
- Platform: Linux-5.15.83.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
- Python version: 3.10.6
- Huggingface_hub version: 0.11.1
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using GPU in script?: Yes
- Using distributed or parallel set-up in script?: Yes and no (zero-3 on single machine)

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -47,19 +47,19 @@ def convert_file_size_to_int(size: Union[int, str]):
     if isinstance(size, int):
         return size
     if size.upper().endswith("GIB"):
-        return int(size[:-3]) * (2**30)
+        return int(size[:-3]) * (2 ** 30)
     if size.upper().endswith("MIB"):
-        return int(size[:-3]) * (2**20)
+        return int(size[:-3]) * (2 ** 20)
     if size.upper().endswith("KIB"):
-        return int(size[:-3]) * (2**10)
+        return int(size[:-3]) * (2 ** 10)
     if size.upper().endswith("GB"):
-        int_size = int(size[:-2]) * (10**9)
+        int_size = int(size[:-2]) * (10 ** 9)
         return int_size // 8 if size.endswith("b") else int_size
     if size.upper().endswith("MB"):
-        int_size = int(size[:-2]) * (10**6)
+        int_size = int(size[:-2]) * (10 ** 6)
         return int_size // 8 if size.endswith("b") else int_size
     if size.upper().endswith("KB"):
-        int_size = int(size[:-2]) * (10**3)
+        int_size = int(size[:-2]) * (10 ** 3)
         return int_size // 8 if size.endswith("b") else int_size
     raise ValueError("`size` is not in a valid format. Use an integer followed by the unit, e.g., '5GB'.")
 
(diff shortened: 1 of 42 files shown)


sgugger:
Thanks for your PR!

@younesbelkada the kwargs added to support bnb and 8-bit load seem to clash with DeepSpeed, so we should probably change the test to only pass them along when the `param_cls` is one of the bnb class to be on the safe side?

younesbelkada:
Thanks for the heads up! 
I tried to run the `bnb` slow tests with this patch and everything seems to work fine. 
This is because in `transformers` we call [`set_8bit_module_tensor_to_device`](https://github.com/huggingface/transformers/blob/73a2ff69740123ef85343580cbfa9ee8ce5e6fd5/src/transformers/utils/bitsandbytes.py#L17) before calling `dispatch_model`. 
In `dispatch_model` [this line](https://github.com/huggingface/accelerate/blob/ac970148cd03c63fbb3bf3966576d39fb66ebaef/src/accelerate/hooks.py#L235):
```python
set_module_tensor_to_device(module, name, self.execution_device)
```
is called recursively, hence the condition [`elif value is not None` is never met. ](https://github.com/huggingface/accelerate/blob/ac970148cd03c63fbb3bf3966576d39fb66ebaef/src/accelerate/utils/modeling.py#L143)

In any case I agree with @sgugger and we should add a safety checker to be on the safe side

src/accelerate/utils/modeling.py
```diff
@@ -143,7 +143,12 @@ def set_module_tensor_to_device(
         elif value is not None or torch.device(device) != module._parameters[tensor_name].device:
             param_cls = type(module._parameters[tensor_name])
             kwargs = module._parameters[tensor_name].__dict__
-            new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
+            if (kwargs.get("ds_tensor") is not None):
```
sgugger, line 146: So let's make this check a bit stronger, like the `param_cls.__name__` is bnb8 parameter class or not, since we only want to pass the kwargs for them.

sgugger (commented):
Noted. @dhar174 can you edit a bit the test then? Should still fix since it's a stronger one.

src/accelerate/utils/modeling.py
dhar174: Yea, no problem.

younesbelkada: FYI the bnb parameter name is `Int8Params` ;) !

dhar174 (commented):
Simple change like this look good?

src/accelerate/utils/modeling.py
```diff
@@ -143,7 +143,12 @@ def set_module_tensor_to_device(
         elif value is not None or torch.device(device) != module._parameters[tensor_name].device:
             param_cls = type(module._parameters[tensor_name])
             kwargs = module._parameters[tensor_name].__dict__
-            new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
+            if (param_cls.__name__ == "Int8Params"):
```
dhar174, line 146: Changed to test param_cls.__name__ for Int8 parameters ("Int8Params") to only pass kwargs if true, else do not send kwargs

sgugger: Perfect, thanks!

sgugger (approved):
Thanks for iterating! Can you just run `make style` on your branch to fix the formatting?

dhar174:
Apologies for that. I've run make style, and it changed 5 files. However, it still seems to fail the quality test. Did I do something wrong when using make style?

dhar174:
Ah yes, that must be the issue.

PR #1016, opened 2023-01-26
Fix test introduced in PR and introduce AcceleratorTestCase

muellerzr:
Fixes test introduced in https://github.com/huggingface/accelerate/pull/991

It needed another call to `AcceleratorState._reset_state()` so I just moved it to the `tearDown` since every test in it reset the state. 

Also introduces an `AcceleratorTestCase` class to handle the `_reset_state()` for users since it's very easy to forget it happens and then a cryptic stack trace follows

### DIFF
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -223,6 +223,19 @@ class TempDirTestCase(unittest.TestCase):
                     shutil.rmtree(path)
 
 
+class AccelerateTestCase(unittest.TestCase):
+    """
+    A TestCase class that will reset the accelerator state at the end of every test. Every test that checks or utilizes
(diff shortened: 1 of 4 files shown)


src/accelerate/test_utils/testing.py
```diff
@@ -223,6 +223,19 @@ def setUp(self):
                     shutil.rmtree(path)
 
 
+class AccelerateTestCase(unittest.TestCase):
```
muellerzr, line 226: Open to calling this `AcceleratorStateTestCase` or `AcceleratorTestCase`, just wanted it to be something obvious where users know *oh if I touch AcceleratorState` somehow they should inherit this

sgugger (approved):
Thanks for the fix and refactor!

patrickvonplaten:
Thanks!

PR #1018, opened 2023-01-26
Don't force mixed precision as no in examples

sgugger:
Currently all examples have a default of `mixed_precision="no"` when defining the `Accelerator`, which overrides whatever the user has in their config. This is a bit annoying, so this PR switches to `None` so that:
- the user can still force any mixed precision they want with `--mixed_precision`
- but otherwise the default value is read in the Accelerate config.


muellerzr (approved):
Makes sense! Thanks for the catch :)

pacman100 (approved):
Thank you!

PR #1020, opened 2023-01-28
Fix for BLOOM models which has slightly different model structure

zsc:
Without this fix, there will be exceptions like
```
AttributeError: 'BloomForCausalLM' object has no attribute 'word_embeddings'
```
raised from /data/accelerate/src/accelerate/utils/modeling.py:112

This fix is to handle for the special case of BloomForCausalLM where there is a ".transformer" indirection.


sgugger:
Looks like you are trying to load a checkpoint of `BloomModel` inside a `BloomForCausalLM` without using the `from_pretrained` method of Transformers. Accelerate is not meant to deal with this logic, it can only handle a checkpoint with the same weight names as the model you are trying to use.

Issue #1022, opened 2023-01-30
Unable to use MPS device


tcapelle:
### System Info

Hello, I am unable to use the `mps` backend with accelerate anymore (it used to work)

```bash
- `Accelerate` version: 0.15.0
- Platform: macOS-12.5.1-arm64-arm-64bit
- Python version: 3.10.8
- Numpy version: 1.24.1
- PyTorch version (GPU?): 1.13.1 (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MPS
        - mixed_precision: no
        - use_cpu: False
        - dynamo_backend: NO
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: None
        - main_process_ip: None
        - main_process_port: None
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_name: None
        - tpu_zone: None
        - command_file: None
        - commands: None
```

After configuring accelerate, if I run:

```python
from accelerate import Accelerator
a = Accelerator()
a.device
>> device(type='cpu')
```
I was expecting that device would be `torch.device("mps")` but no.

Is there a wat to force the device into the accelerator?

pacman100:
Hello @tcapelle, can you refer to the discussion in #998 and let us know if that solves the issue

tcapelle:
Thanks @pacman100 , actually using the launcher works. But I want to be able to use accelerate interactively in the notebook, without the launcher.
We are also using accelerate in the new fastai course and it does not work on a Mac GPU as we are running the code cell by cell.
Is there a workaround to tell the accelerator to use the device `mps`?

muellerzr:
@tcapelle do you happen to know the last time it worked successfully for you prior to v0.15?

tcapelle:
It's fixed in dev

muellerzr:
Yep! #1024 set the right env after the switch, great investigating!

PR #1024, opened 2023-01-30
Fix env var for MPS device on notebook_launcher

muellerzr:
When adding the `ACCELERATE` prefix in https://github.com/huggingface/accelerate/pull/890, the `notebook_launcher` needs to check `ACCELERATE_USE_MPS_DEVICE`

Closes https://github.com/huggingface/accelerate/issues/1022


pacman100 (approved):
Thank you for fixing this! 🤗

PR #1025, opened 2023-01-30
Change default for keep_fp32_wrapper

muellerzr:
> If a user does an evaluation loop at the end of each epoch, then they have to use unwrap_model with the keep_wrapper at True for FP16 integration. We made a breaking change with that when introducing the arg.
> For instance the no trainer translation and summarization scripts have stopped working with FP16.

Switches regression and makes the default `True`


sgugger (approved):
Thanks for taking care of it!

PR #1026, opened 2023-01-30
Fix slow test by keeping tied weights on the same GPU

sgugger:
In #1000 we fixed the weight tying in `dispatch_model`, but one of the slow tests relied on the fact it was broken (by mistake) by putting tied weights of GPT-2 on two different devices. This PR fixes that.


muellerzr (approved):
Great! Thanks for the fix

PR #1027, opened 2023-01-30
Add example usage to `skip_batches`

muellerzr:


sgugger (approved):
Thanks for adding this!

PR #1028, opened 2023-01-30
Load tensors directly on device

sgugger:
This PR refines `load_checkpoint_in_model` to directly load weights on the GPUs when the checkpoint file is a safetensors file, since `safetensors` can load very fast directly there. It also introduces a new `load_state_dict` function to do this loading.

@patrickvonplaten is this enough for your use case in `diffusers`?

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -24,8 +24,13 @@ from typing import Dict, List, Optional, Tuple, Union
 import torch
 import torch.nn as nn
 
+from .imports import is_safetensors_available
 from .offload import load_offloaded_weight, offload_weight, save_offload_index
 
+
+if is_safetensors_available():
+    from safetensors import safe_open
+    from safetensors.torch import load_file as safe_load_file
 
 WEIGHTS_INDEX_NAME = "pytorch_model.bin.index.json"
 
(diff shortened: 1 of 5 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -621,6 +626,55 @@ def check_device_map(model: nn.Module, device_map: Dict[str, Union[int, str, tor
...
+    Args:
+        checkpoint_file (`str`): The path to the checkpoint to load.
+        device_map (`Dict[str, Union[int, str, torch.device]]`, *optional*):
+            A map that specifies where each submodule should go. It doesn't need to be refined to each parameter/buffer
+            name, once a given module name is inside, every submodule of it will be sent to the same device.
+    """
+    if checkpoint_file.endswith(".safetensors"):
+        if not is_safetensors_available():
+            raise ImportError(
+                f"To load {checkpoint_file}, the `safetensors` library is necessary `pip install safetensors`>"
```
muellerzr, line 643:
```suggestion
                f"To load {checkpoint_file}, the `safetensors` library is necessary `pip install safetensors`."
```

patrickvonplaten (approved):
Cool that works well with `diffusers` ! Like in `transformers`, we use `load_checkpoint_and_dispatch` so it should work out of the box :-)

pacman100 (approved):
Thank you for extending the support to safetensors and the related tests. LGTM! 🤗

PR #1030, opened 2023-01-31
enabling `mps` device by default and removing related config

pacman100:
### What does this PR do?
1. As `mps` and `cuda` won't be concurrently available/set, we can remove the `mps` config and have it enabled by default when it is available. To disable it use `--cpu` arg. This PR does that.
2. Supports backward compatibility till version 0.18.0. Post that `mps` related config will be removed.

### DIFF
--- a/src/accelerate/commands/config/cluster.py
+++ b/src/accelerate/commands/config/cluster.py
@@ -44,7 +44,7 @@ from .config_utils import (
 def get_cluster_input():
     distributed_type = _ask_options(
         "Which type of machine are you using?",
-        ["No distributed training", "multi-CPU", "multi-GPU", "TPU", "MPS"],
+        ["No distributed training", "multi-CPU", "multi-GPU", "TPU"],
         _convert_distributed_mode,
     )
 
@@ -92,7 +92,7 @@ def get_cluster_input():
 
     if distributed_type == DistributedType.NO:
         use_cpu = _ask_field(
-            "Do you want to run your training on CPU only (even if a GPU is available)? [yes/NO]:",
+            "Do you want to run your training on CPU only (even if a GPU / Apple Silicon device is available)? [yes/NO]:",
             _convert_yes_no_to_bool,
             default=False,
             error_message="Please enter yes or no.",
@@ -129,8 +129,16 @@ def get_cluster_input():
     else:
         dynamo_backend = DynamoBackend.NO
 
+    use_mps = False
+    if distributed_type == DistributedType.NO:
+        use_mps = _ask_field(
+            "Do you want to use Apple Silicon GPUs via `mps` device backend? [yes/NO]:",
+            _convert_yes_no_to_bool,
+            default=False,
+            error_message="Please enter yes or no.",
+        )
     deepspeed_config = {}
-    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO]:
+    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO] and not use_mps:
         use_deepspeed = _ask_field(
             "Do you want to use DeepSpeed? [yes/NO]: ",
             _convert_yes_no_to_bool,
@@ -452,7 +460,7 @@ def get_cluster_input():
     else:
         num_processes = 1
 
-    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO] and not use_cpu:
+    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO] and not use_cpu and not use_mps:
         gpu_ids = _ask_field(
             "What GPU(s) (by id) should be used for training on this machine as a comma-seperated list? [all]:",
             default="all",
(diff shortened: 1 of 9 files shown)


src/accelerate/commands/config/cluster.py
```diff
@@ -129,8 +129,16 @@ def get_cluster_input():
...
         dynamo_backend = DynamoBackend.NO
 
+    use_mps = False
+    if distributed_type == DistributedType.NO:
+        use_mps = _ask_field(
+            "Do you want to use Apple Silicon GPUs via `mps` device backend? [yes/NO]:",
+            _convert_yes_no_to_bool,
+            default=False,
+            error_message="Please enter yes or no.",
+        )
```
sgugger, line 139: I thought we were removing this entirely? Here `use_mps` should be `not use_cpu` when the MPS device is available, `False` otherwise.

src/accelerate/state.py
```diff
@@ -223,34 +223,34 @@ def __init__(
...
+                        raise AssertionError(
+                            "MPS not available because PyTorch version is < 1.12.0 or MacOS version is < 12.3 "
+                            "and/or you do not have an MPS-enabled device on this machine."
+                        )
+
+                if self.device is None:
+                    if (
+                        cpu
+                        or not torch.cuda.is_available()
+                        or not (torch.backends.mps.is_available() and torch.backends.mps.is_built())
```
sgugger, line 248: Nit, this condition appears a lot. We could have a property of the `AcceleratorState` that stores it (in something like `is_mps_available`) so we can simplify all those tests.

sgugger (commented):
Thanks for working on this! I left a couple of comments.

src/accelerate/commands/config/cluster.py
pacman100: `not use_cpu` won't work. As I mentioned offline, it would ask for `gpu_ids` if it is `not use_cpu` which isn't applicable for `mps`.

sgugger: I have trouble understanding why, since if you set `use_mps` to `not use_cpu and is_mps_available()` here, at line 463 `use_mps=True` will prevent the question about GPUs.

pacman100: Yup, makes sense, addressed it, thanks for the pointer

sgugger (approved):
Nice!

src/accelerate/commands/config/cluster.py
```diff
@@ -92,7 +93,7 @@ def get_cluster_input():
 
     if distributed_type == DistributedType.NO:
         use_cpu = _ask_field(
-            "Do you want to run your training on CPU only (even if a GPU is available)? [yes/NO]:",
+            "Do you want to run your training on CPU only (even if a GPU / Apple Silicon device is available)? [yes/NO]:",
```
muellerzr, line 96:
```suggestion
            "Do you want to run your training on CPU only (even if a GPU or Apple Silicon device is available)? [yes/NO]:",
```
Ideally we should do a small check here to get if it's GPU or Apple Silicon

src/accelerate/commands/config/cluster.py
```diff
@@ -129,8 +130,9 @@ def get_cluster_input():
     else:
         dynamo_backend = DynamoBackend.NO
 
+    use_mps = not use_cpu and is_mps_available()
     deepspeed_config = {}
-    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO]:
+    if distributed_type in [DistributedType.MULTI_GPU, DistributedType.NO] and not use_mps:
```
muellerzr, line 135: It also makes sense if we keep `MPS` to make this logic here the same. It adds some more complexity in the refactor/improvement I think

src/accelerate/commands/launch.py
```diff
@@ -152,7 +153,8 @@ def launch_command_parser(subparsers=None):
         "--mps",
         default=False,
         action="store_true",
-        help="Whether or not this should use MPS-enabled GPU device on MacOS machines.",
+        help="This argument is deprecated. MPS device will be enabled by default when available and can be disabled via `--cpu`."
+        " Whether or not this should use MPS-enabled GPU device on MacOS machines.",
```
muellerzr, line 157: Same thought here, keep MPS since it's separate from `--cpu`

muellerzr (changes requested):
Let me know your thoughts on this redesign. I think doing it like this keeps it more in-tune with how I view devices when it comes to the library and its presentation, but open to how others feel about it cc @sgugger

pacman100:
Here is a quick summary of my thoughts:

As I understood from internal discussions and user feedback, the aim was to be inline with the experience users currently have for cuda devices. Users never state they want to use CUDA devices and it's detected automatically without any env or config as you mentioned wrt user feedback. The PR achieves this.

It avoids asking irrelevant questions when running accelerate config by detecting whether `mps` is available or not.

wrt code complexity, I think it's fine because we currently already have `use_cpu`  at a few places. These are places it can't be avoided but at the same time, it reduces the load on users by having them one less thing to specify. The user experience of having to specify less config to have things work is really cool and impactful and that is why I was very interested in quickly working on this PR.

PR #1036, opened 2023-02-02
do not scale gradient in bf16 mode

kashif:
`GradScaler` only makes sense in float16 mode and not for bf16 on the GPU. Removed block of code which did this for bf16.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -392,11 +392,6 @@ class Accelerator:
             if mixed_precision == "bf16" and not self.native_amp and not is_tpu_available():
                 raise ValueError(err.format(mode="bf16", requirement="PyTorch >= 1.10 and a supported device."))
 
-            # Only on the GPU do we care about scaling the gradients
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Let's just wait a little bit for the conversation in the internal thread to be concluded before merging, but otherwise LGTM!
Thanks for making the fix!

kashif:
would be great if @stas00 can also have a look 🙇🏽

stas00:
I confirm.

bf16 has an almost the same dynamic range as fp32, so no scaling is needed.

the same needs to be fixed in HF Trainer

https://github.com/huggingface/transformers/blob/98d88b23f54e5a23e741833f1e973fdf600cc2c5/src/transformers/trainer.py#L573

probably it should be:
```
                if self.amp_dtype == torch.float16:
                    self.do_grad_scaling = True
```

stas00 (approved):

kashif:
sure @stas00 let me get the Trainer fixed

PR #1037, opened 2023-02-03
Fixing of a bug connected with DeepSpeed

DistinctVision:
Hello. I tried to use DeepSpeed for my project and met a small bug. The problem was connected with backward function - i had to pass an argument 'retain_graph'. But we lose it by passing to internal DeepSpeed function.
I fixed it and these are my changes.


pacman100 (approved):
LGTM! 😄
Thank you @DistinctVision for the fix.

PR #1039, opened 2023-02-03
Refactor process executors to be in AcceleratorState

muellerzr:
Based on conversation in slack, refactors the logic for process executors to be handled in the `AcceleratorState` and delegated to the `Accelerator` instead of in the `Accelerator` itself so that the `State` can be used by individuals that want to do process delegation without needing multiple `Accelerator` objects in their codebase. 

Still TODO is to move the decorators over, but those need some more improvements due to confusions in them and tests so that is a much later to-get-to

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -770,8 +770,7 @@ class Accelerator:
         >>> accelerator.print("Hello world!")
         ```
         """
-        if self.is_local_main_process:
-            print(*args, **kwargs)
+        self.state.print(*args, **kwargs)
 
     def _prepare_one(self, obj, first_pass=False, device_placement=None):
         # First pass of preparation: DataLoader, model, optimizer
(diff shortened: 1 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -573,7 +558,7 @@ def main_process_first(self):
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
-        yield from self._goes_first(self.is_main_process)
+        yield self.state.main_process_first()
```
sgugger, line 561: Shoudn't this be `yield from`?

sgugger (approved):
Never sure between `yield` and `yield from` but otherwise LGTM!

src/accelerate/accelerator.py
muellerzr: Using `yield from` actually will raise an error and using `yield` with the inner calling `yield from` is enough. I'll add a test with this first but the multi-GPU tests passed :)

sgugger (approved):

PR #1042, opened 2023-02-06
Add daily slack notifier for nightlies

muellerzr:
Introduces a slackbot that posts in #accelerate-ci-daily with the status of how the nightly tests ran each day.


sgugger (approved):
Thanks for adding this!

PR #1045, opened 2023-02-07
Add `cpu_offload_with_hook`

sgugger:
This is a feature request for Diffusers. This PR adds a new `cpu_offload_with_hook` function that will offload the model to CPU, then put it back on the GPU once executed, but without offloading it just after the forward like `cpu_offload` does. Instead, it's up to the user to call the `hook.offload()` to offload it again.

Example:
```py
import torch
form accelerate import cpu_offload_with_hook

model = nn.Linear(4, 5)
model, hook = cpu_offload_with_hook(model)
print(model.weight.device) # Always cpu

outputs = model(inputs)
print(outputs.device) # Outputs are on GPU, execution was done on GPU
print(model.weight.device) # Stays on the GPU until hook.offload() is called

hook.offload()
print(model.weight.device) # Back to cpu
```

cc @pcuenca @patrickvonplaten

### DIFF
--- a/src/accelerate/__init__.py
+++ b/src/accelerate/__init__.py
@@ -7,6 +7,7 @@ __version__ = "0.17.0.dev0"
 from .accelerator import Accelerator
 from .big_modeling import (
     cpu_offload,
+    cpu_offload_with_hook,
     disk_offload,
     dispatch_model,
     init_empty_weights,
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -19,7 +19,14 @@ from typing import Dict, List, Optional, Union
 import torch
 import torch.nn as nn
 
-from .hooks import AlignDevicesHook, add_hook_to_module, attach_align_device_hook, attach_align_device_hook_on_blocks, CpuOffload, UserCpuOffloadHook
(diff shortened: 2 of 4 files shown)


patrickvonplaten (commented):
This looks very nice! If possible we were also envisioning the possibility to pass a previous user_hook to the CpuOffload hook so that the previous model can be offloaded at forward. E.g. so that the following could be possible:

hook_1 = cpu_offload_with_hook(model_1, cuda_device)
hook_2 = cpu_offload_with_hook(model_2, cuda_device, user_hook=hook_1)
hook_3 = cpu_offload_with_hook(model_3, cuda_device, user_hook=hook_2)

so that the following would automatically work:

```py
hid_1 = model_1(input)
for i in range(50):
     hid_2 = model_2(hid_1)
hid_3 = model_3(hid_3)
```

Alternatively, we could also call the hooks ourselves if preferred, but if we could directly pass the hooks as shown in the code review that would make diffusers code much nicer I think :-)

pacman100:
Really Cool!

patrickvonplaten (approved):
Awesome! Thanks a mille for the super fast PR ❤️

PR #1049, opened 2023-02-08
Introduce TPU Pod launching to `accelerate launch`

muellerzr:
## Proposed API:

`accelerate launch` now allows for a configured pod setup through three new params/config items:
- `use_cluster`, whether to use a TPU cluster
- `vm`, this mimics `xla_dist`'s `vm` argument and is a list of single compute VM names if you are not using an instance group. (generally not needed)
- `env`, this is a list of environment variables to set on each of the compute VM instances

Currently we only support non-Docker, as GCP doesn't support docker yet on the larger pods. 

To launch a script on a TPU pod, the API will look like such:

### Fully configured:
```bash
accelerate launch myscript.py --arg1 --arg2 ...
```
### No configuration:
```bash
accelerate launch --tpu --use_cluster myscript.py --arg1 --arg2 ...
```

Note that if file permissions are given on your GCP cluster, there is a `--tpu_use_sudo` flag for this, and I recommend also doing `sudo accelerate launch ...` plus `sudo pip install git+https://github.com/huggingface/accelerate`. On my specific configuration it was needed, but other users reported they didn't so it's optional. 

Closes https://github.com/huggingface/accelerate/issues/501 and closes https://github.com/huggingface/accelerate/issues/471


sgugger (approved):
Thanks for your work on this!

PR #1053, opened 2023-02-09
Fix args failure by adding in the defaults

muellerzr:
The post-filtered GPU params to `launch` need to maintain the script and arg parameters. This PR adds them in. 

Running on the nightly-ci to be sure before merging


sgugger (approved):
Thanks for the fix!

PR #1055, opened 2023-02-10
Introduce PartialState

muellerzr:
This PR introduces a `PartialState` class, which allows for process control while being *fully disconnected* from the `Accelerator` class. This allows users to utilize `print`, `on_main_process`, and more on their library integrations without needing to create an `Accelerator` object, and leave that to the user to create themselves. 

The `AcceleratorState` class does not inherit from this class, as otherwise the `__dict__`'s would be the same for `PartialState` and `AcceleratorState`, which we don't want. So it's referenced instead without explicit inheritance with passthroughs.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -73,6 +73,7 @@ class PartialState:
         self.__dict__ = self._shared_state
         self._check_initialized(cpu)
         if not self.initialized:
+            self.backend = None
             env_device = os.environ.get("ACCELERATE_TORCH_DEVICE", None)
             self.device = torch.device(env_device) if env_device is not None else None
             if (
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -74,8 +74,9 @@ class PartialState:
         # Raise an error if the user tries to reinitialize on a different device setup in the same launch
         if self.initialized and (self._cpu != cpu):
             raise AssertionError(
-                "You cannot reinitialize the state with a different device setup in the same script,"
-                "please launch the script again using the desired device setup."
+                "The current device and desired device are not the same. If the `PartialState` was generated "
+                "before the `AcceleratorState` has been made, ensure the `cpu` flag is the same for both. In this case, "
+                f"the `PartialState` has {self._cpu} and the desired device is {cpu}. Please use `cpu={self._cpu}`."
             )
         if not self.initialized:
             self._cpu = cpu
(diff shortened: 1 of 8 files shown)


sgugger (commented):
The object does not respect its promises. If you use a `PartialState` then instantiates an `Accelerator` with `cpu=True`, you will get an error.

tests/test_scheduler.py
```diff
@@ -82,15 +83,19 @@ def test_one_cycle_scheduler_not_step_with_optimizer_single_process(self):
         debug_launcher(partial(one_cycle_test, num_processes=1, step_scheduler_with_optimizer=False), num_processes=1)
 
     def test_lambda_scheduler_steps_with_optimizer_multiprocess(self):
+        AcceleratorState._reset_state(True)
```
muellerzr, line 86: Since it switches from single cpu to multi-CPU within the same process/time period, the state needs to be reset

sgugger (commented):
I have one problem with the implementation: if the user uses a `PartialState` without setting `cpu=False` and then instantiates an `Accelerator` with `cpu=True`, all the info taken from the `PartialState` before the `Accelerator` is reinitialized will be wrong no?
So maybe we should error on the change (e.g. the user can use `PartialState` before instantiating an `Accelerator`, but in this case it needs to be done with consistent cpu vlaues).

What do you think?

muellerzr:
I agree, because it's highly unlikely a user would be trying to mix/match the two together off a single launch/instance, and if they were they're probably doing something bad. I'll update with this

src/accelerate/state.py
```diff
@@ -46,59 +46,43 @@ def is_initialized() -> bool:
...
+    def __init__(self, cpu: bool = False, **kwargs):
         self.__dict__ = self._shared_state
-        if parse_flag_from_env("ACCELERATE_USE_CPU"):
-            cpu = True
-        self._check_initialized(mixed_precision, cpu)
+        # Raise an error if the user tries to reinitialize on a different device setup in the same launch
+        if self.initialized and (self._cpu != cpu):
+            raise AssertionError(
+                "The current device and desired device are not the same. If the `PartialState` was generated "
+                "before the `AcceleratorState` has been made, ensure the `cpu` flag is the same for both. In this case, "
```
sgugger, line 78:
```suggestion
                "before the `Accelerator` has been instantiated, ensure the `cpu` flag is the same for both. In this case, "
```

sgugger (approved):
Good to go like this!

Issue #1054, opened 2023-02-09
All layer weights not getting saved when using accelerate + FSDP for OPT 2.7B


lvnair3:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-5.15.0-60-generic-x86_64-with-glibc2.35
- Python version: 3.10.6
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.0+cu116 (True)

`Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: FSDP
        - mixed_precision: no
        - use_cpu: False
        - dynamo_backend: NO
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {'fsdp_auto_wrap_policy': 'SIZE_BASED_WRAP', 'fsdp_backward_prefetch_policy': 'BACKWARD_PRE', 'fsdp_min_num_params': 1000000, 'fsdp_offload_params': True, 'fsdp_sharding_strategy': 1, 'fsdp_state_dict_type': 'FULL_STATE_DICT'}
        - megatron_lm_config: {}
        - downcast_bf16: no
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

**SETUP**

Task: Finetuning `facebook/opt-2.7b` for Wikitext2

Code is: https://github.com/huggingface/transformers/blob/main/examples/pytorch/language-modeling/run_clm_no_trainer.py

Dataset is: Wikitext2

The command used to run this is as follows:
```
accelerate launch run_clm_no_trainer.py \
    --model_name_or_path facebook/opt-2.7b \
    --dataset_name wikitext \
    --dataset_config_name wikitext-2-raw-v1 \
    --per_device_train_batch_size 2 \
    --per_device_eval_batch_size 8 \
    --output_dir ./opt-2.7B-FP32 \
    --checkpointing_steps epoch \
    --max_train_steps 3
```

**ERROR**
[...]

PR #1059, opened 2023-02-13
update fsdp docs and removing deepspeed version pinning

pacman100:
### What does this PR do?
1. Adds docs on correct way to save FSDP models as per the resolution of issue #1054 
2. Removes deepspeed version pinning

### DIFF
--- a/docs/source/usage_guides/fsdp.mdx
+++ b/docs/source/usage_guides/fsdp.mdx
@@ -79,11 +79,11 @@ Currently, `Accelerate` supports the following config through the CLI:
   Below is an example:
 
 ```diff
-unwrapped_model.save_pretrained(
-    args.output_dir,
-    is_main_process=accelerator.is_main_process,
-    save_function=accelerator.save,
-+   state_dict=accelerator.get_state_dict(model),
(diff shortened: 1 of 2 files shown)


docs/source/usage_guides/fsdp.mdx
```diff
@@ -73,6 +73,20 @@ Currently, `Accelerate` supports the following config through the CLI:
...
+1. When using transformers `save_pretrained`, pass `state_dict=accelerator.get_state_dict(model)` to save the model state dict. 
+  Below is an example:
+
+```diff
+unwrapped_model.save_pretrained(
+    args.output_dir,
+    is_main_process=accelerator.is_main_process,
+    save_function=accelerator.save,
++   state_dict=accelerator.get_state_dict(model),
+)
```
sgugger, line 87: You need to add two space at each line, and then just + and a space on the line added for formatting :-)

sgugger (approved):
Thanks for the update!

docs/source/usage_guides/fsdp.mdx
pacman100: Done. Thank you.

PR #1060, opened 2023-02-13
deepspeed `hidden_size` auto value default fixes

pacman100:
### What does this PR do?
1. handle models with `hidden_sizes` attributes for default values of deepspeed config which will be used if user has specified these as `auto`. Reference PR: https://github.com/huggingface/transformers/pull/21504


sgugger (approved):
Thanks for aligning the APIs!

PR #1061, opened 2023-02-13
Flag use_fp16 for deprecation in 1.0

muellerzr:
As discussed in https://huggingface.slack.com/archives/C01Q6JPP6NA/p1676299264125009, flags `use_fp16` for deprecation once v1.0.0 of Accelerate is released.


sgugger (approved):
Thanks!

PR #1062, opened 2023-02-13
Fix failed deepspeed tests with proper reset of state

muellerzr:
Same comment as before, `reset_state` needs `True` to reset the `PartialState` for these tests. Ran locally on two GPUs to prove it fixes the failures.


sgugger (approved):
Thanks for the fix!

PR #1063, opened 2023-02-13
Update integrations

muellerzr:
Updates the integrations in the README to be a bit more general allowing for us to include high visibility repos utilizing the library. Since the integration zoo is based on python frameworks with examples using them, instead of very, very high level frameworks (like Invoke and Open Assistant, where it's mostly through a web interface things happen) they were not added to this list. 

This list was generated looking at the most forked and utilized repos based on [this dataset](https://huggingface.co/datasets/open-source-metrics/accelerate-dependents) for February 12th, 2023

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,13 +208,16 @@ You shouldn't use 🤗 Accelerate if you don't want to write a training loop you
 
 ## Frameworks using 🤗 Accelerate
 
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks!

PR #1064, opened 2023-02-13
Update deffering exec docs

muellerzr:
Updates the relevant docs to showcase the new `PartialState` as a more viable option depending on the use case.

### DIFF
--- a/docs/source/concept_guides/deferring_execution.mdx
+++ b/docs/source/concept_guides/deferring_execution.mdx
@@ -35,7 +35,7 @@ These two objects will be referenced as such throughout the rest of this tutoria
 
 <Tip>
 
-    Every function that is shown from the `Accelerator` object in this tutorial can be found in the `PartialState` object as well.
+    Every function that is shown from `PartialState` in this tutorial can be found in the `Accelerator` class as well.
 
 </Tip>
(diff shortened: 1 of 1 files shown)


docs/source/concept_guides/deferring_execution.mdx
```diff
@@ -16,14 +16,43 @@ When you run your usual script, instructions are executed in order. Using 🤗 A
...
+partial_state = PartialState()
+```
+
+These two objects will be referenced as such throughout the rest of this tutorial.
+
+<Tip>
+
+    Every function that is shown from the `Accelerator` object in this tutorial can be found in the `PartialState` object as well.
+
+</Tip>
```
sgugger, line 40: Isn't it the other way round? Also no need for indentation.

sgugger (approved):
Thanks!

docs/source/concept_guides/deferring_execution.mdx
muellerzr: I suppose it would make more sense to show off the `PartialState` since we want users using that, will adjust!

muellerzr:
Closing to open a new PR based on the new changes introduced today

PR #1065, opened 2023-02-13
Swap utils over to use PartialState

muellerzr:
Swaps over any utils that do not directly relate to the `Accelerator` object to use the `PartialState` instead (so utils on the optimizer, dataloader, etc all remain the same because they directly work with `accelerator.prepare`)


sgugger (approved):
Thanks for your work on this!

Issue #923, opened 2022-12-15
TypeError: on_main_process() takes 1 positional argument but 2 were given


ernestchu:
Hi, I'm not sure if I misunderstood the usage or not. But according to the doc, one can use ` @accelerator.on_main_process` decorator on a function.

https://github.com/huggingface/accelerate/blob/447ad0e6359a5b6f686e36c45cfa411a1cbe3997/docs/source/package_reference/accelerator.mdx?plain=1#L85-L93

Yet, I get the error `TypeError: on_main_process() takes 1 positional argument but 2 were given`. I guess the reason is that the method do not take `self` as argument, so one cannot really use it as described in the documentation :(

https://github.com/huggingface/accelerate/blob/f002ce2ae97f0e04d9081102dfd81b792a4cfcf9/src/accelerate/accelerator.py#L488-L491

If I got it wrong, please correct me. Thanks!

PR #1070, opened 2023-02-14
First get decorator working

muellerzr:
# Fix process decorators and make them more flexible

## What does this add?

This PR completely refactors/reworks the decorator methodology introduced in https://github.com/huggingface/accelerate/pull/488 into something that fully completes the proposed API

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/923

## Why is it needed?

The decorators that were introduced did not work as intended, and only worked to have the `Accelerator` object itself actually use it. To use it outside the `Accelerator` was impossible. As a result, this PR completely reworks how the existing decorators were designed by making them utilize the `state` instead, and ensures that a special internal use case is added when using the decorators back on the `Accelerator` class itself, with a raised error if a user tries to do the same. 

### User-facing:

A user can now actually perform `@accelerator.on_main_process` as a decorator, along with the rest of the existing ones.
The `AcceleratorState` also now stores the `use_distributed` value originally in the `Accelerator` object

### Internal structure:

The existing implementation in `Accelerator` was gutted, and reworked into a usable fashion.

Specifically, this logic has now been added to the decorators:
```python
    def on_local_process(self, function: Callable[..., Any] = None, local_process_index: int = None):
        ...
        if function is None:
            if "Accelerator." in self.__qualname__:
                function = self
            else:
                raise ValueError(
                    "The `on_main_process` decorator must be called with a function on an instantiated `Accelerator` object."
                )
```
This check is to ensure if `self` is a function *from* the `Accelerator` itself, or the behavior that occurs when performing:

```python
# Inside the Accelerator object definition
    @on_main_process
    def log(self, ...)
        ...
```
Instead of requiring complex workarounds. As you can see in the above, a proper error will also be raised if a user breaks this "magic" with a clear description of what is happening.

Inside the `on_process*` functions there also exists this snippet:

```python
    def on_process(self, function: Callable[..., Any] = None, process_index: int = None):
        # Initial construction of the decorator.
        if (self is not None) and (process_index is not None) and (function is None):
            return partial(self.on_process, process_index=process_index)
```
This is to check if an overloaded decorator was used, potentially again on the `Accelerator` class, before moving on to the proper constructor. 

## Basic Usage Example(s):

```python
accelerator = Accelerator()
@accelerator.on_local_main_process
def print_local_main():
    print(f"Printing from the local main process {state.local_process_index}")
print_local_main()
```

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -20,8 +20,8 @@ import sys
 import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
-from functools import wraps
-from typing import Callable, List, Optional, Union
+from functools import partial
+from typing import Any, Callable, List, Optional, Union
 
 import torch
 import torch.utils.hooks as hooks
@@ -31,7 +31,7 @@ from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_first_b
 from .logging import get_logger
 from .optimizer import AcceleratedOptimizer
 from .scheduler import AcceleratedScheduler
-from .state import AcceleratorState, GradientState, parse_flag_from_env
+from .state import AcceleratorState, GradientState, PartialState, parse_flag_from_env
 from .tracking import LOGGER_TYPE_TO_CLASS, GeneralTracker, filter_trackers
 from .utils import (
     MODEL_NAME,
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Thanks for splitting the work in two. LGTM!

PR #1071, opened 2023-02-14
Refactor state and make `PartialState` first class citizen 

muellerzr:
Refactors the `PartialState` and `Accelerator` to have the core decorator process-logic happen inside `PartialState` and then upstreams it to `AcceleratorState` and `Accelerator`, as `PartialState` should act as the main processes controller for users if they choose not to use the `Accelerator`. As a result, `PartialState` has complete docs for the functions while the `AcceleratorState` does not. I can also make a note on the version in `AcceleratorState` to point this out to users potentially.


sgugger (approved):
Thanks a lot!

Issue #1073, opened 2023-02-15
No such file or directory: '/root/.cache/huggingface/accelerate/default_config.yaml'


LazyFyh:
### System Info

```Shell
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /usr/local/bin/accelerate:8 in <module>                                                          │
│                                                                                                  │
│   5 from accelerate.commands.accelerate_cli import main                                          │
│   6 if __name__ == '__main__':                                                                   │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                         │
│ ❱ 8 │   sys.exit(main())                                                                         │
│   9                                                                                              │
│                                                                                                  │
│ /usr/local/lib/python3.7/dist-packages/accelerate/commands/accelerate_cli.py:45 in main          │
│                                                                                                  │
│   42 │   │   exit(1)                                                                             │
│   43 │                                                                                           │
│   44 │   # Run                                                                                   │
│ ❱ 45 │   args.func(args)                                                                         │
│   46                                                                                             │
│   47                                                                                             │
│   48 if __name__ == "__main__":                                                                  │
│                                                                                                  │
│ /usr/local/lib/python3.7/dist-packages/accelerate/commands/launch.py:985 in launch_command       │
│                                                                                                  │
│    982 │   mp_from_config_flag = False                                                           │
│    983 │   # Get the default from the config file.                                               │
│    984 │   if args.config_file is not None or os.path.isfile(default_config_file) and not args.  │
│ ❱  985 │   │   defaults = load_config_from_file(args.config_file)                                │
│    986 │   │   if (                                                                              │
│    987 │   │   │   not args.multi_gpu                                                            │
│    988 │   │   │   and not args.tpu                                                              │
│                                                                                                  │
│ /usr/local/lib/python3.7/dist-packages/accelerate/commands/config/config_args.py:46 in           │
│ load_config_from_file                                                                            │
│                                                                                                  │
│    43 def load_config_from_file(config_file):                                                    │
│    44 │   config_file_exists = config_file is not None and os.path.isfile(config_file)           │
│    45 │   config_file = config_file if config_file_exists else default_config_file               │
│ ❱  46 │   with open(config_file, "r", encoding="utf-8") as f:                                    │
│    47 │   │   if config_file.endswith(".json"):                                                  │
│    48 │   │   │   if (                                                                           │
[...]

PR #1074, opened 2023-02-15
Add error when trying to load non-existant config.yaml

muellerzr:
As noted in https://github.com/huggingface/accelerate/issues/1073, the user tried to perform `accelerate launch --config_file MY_CONFIG.yaml ...` and if the file does not exist then a non-intuitive trace is given:
```python
FileNotFoundError: [Errno 2] No such file or directory: '/root/.cache/huggingface/accelerate/default_config.yaml'
```
Even though the user specified their own custom configuration.

This PR introduces a guard now instead of a magic-always-passing dropback when using `accelerate launch` that originally would then check if the default configuration could be used instead. 

New error message:
```python
(accelerate) zach@workstation:~/accelerate/accelerate$ accelerate launch --config_file "config.yml" test.py
  File "/home/zach/miniconda3/envs/accelerate/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/home/zach/accelerate/accelerate/src/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/home/zach/accelerate/accelerate/src/accelerate/commands/launch.py", line 1082, in launch_command
  File "/home/zach/accelerate/accelerate/src/accelerate/commands/config/config_args.py", line 46, in load_config_from_file
    raise FileNotFoundError(
FileNotFoundError: The passed configuration file `config.yml` does not exist. Please pass an existing file to `accelerate launch`, or use the the default one created through `accelerate config` and run `accelerate launch` without the `--config_file` argument.
```

### DIFF
--- a/src/accelerate/commands/config/config_args.py
+++ b/src/accelerate/commands/config/config_args.py
@@ -41,8 +41,16 @@ else:
 
 
 def load_config_from_file(config_file):
-    config_file_exists = config_file is not None and os.path.isfile(config_file)
-    config_file = config_file if config_file_exists else default_config_file
+    if config_file is not None:
+        if not os.path.isfile(config_file):
+            raise FileNotFoundError(
+                f"The passed configuration file `{config_file}` does not exist. "
+                "Please pass an existing file to `accelerate launch`, or use the the default one "
+                "created through `accelerate config` and run `accelerate launch` "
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks a lot!

PR #1075, opened 2023-02-15
devcontainer: "extensions" has been removed and replaced by customizations

dbpprt:
"extensions" has been moved to "customizations/vscode"

```
"customizations": {
        "vscode": {
            "extensions": [
                // Ensure we have IntelliSense in VSCode when running inside container
                "ms-python.python"
            ]
        }
    }
```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,10 +15,14 @@
     "remoteEnv": {
         "PYTHONPATH": "${containerEnv:PATH}:${containerWorkspaceFolder}"
     },
-    "extensions": [
-        // Ensure we have IntelliSense in VSCode when running inside container
-        "ms-python.python"
-    ],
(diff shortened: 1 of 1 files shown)


sgugger:
Let's cc @Chris-hughes10 who authored this.

Chris-hughes10:
This is a response to a breaking change in the remote container extension and is the change that I would make. @dennisbappert have you verified that the dev container builds correctly with the new version?

dennisbappert:
I tested the "default" configuration, so CPU image. 

image


I only tested it and do not use this configuration for development, I'm using my own dev container boilerplate.

dennisbappert:
Let me investigate further.

dennisbappert:
#1087

sgugger:
That's fine by me if it also works for @dbpprt

Issue #1076, opened 2023-02-15
SageMakerConfig missing 1 required positional argument: 'image_uri'


dennisbappert:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-6.0.12-76060006-generic-x86_64-with-glibc2.35
- Python version: 3.10.9
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- `Accelerate` default config:
        Not found
```

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Steps to reproduce:
- Run `accelerate config`
- Choose `SageMaker`
- Do not specify a custom image uri

The generated config cannot be parsed because the field image_uri is missing.

```
Traceback (most recent call last):
  File "/opt/conda/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/env.py", line 50, in env_command
    accelerate_config = load_config_from_file(args.config_file).to_dict()
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/config/config_args.py", line 64, in load_config_from_file
    return config_class.from_yaml_file(yaml_file=config_file)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/config/config_args.py", line 123, in from_yaml_file
    return cls(**config_dict)
TypeError: SageMakerConfig.__init__() missing 1 required positional argument: 'image_uri'
```

### Expected behavior

```
image_uri needs to be present in the config and set to `None`
```

sgugger:
cc @philschmid

PR #1077, opened 2023-02-15
#1076 SageMaker image_uri is now optional

dbpprt:
Addressing #1076


sgugger (approved):
Thanks for the fix! LGTM but would love to have @philschmid have a look as well.

philschmid (approved):
LGTM! Thank you for the fix

PR #1079, opened 2023-02-15
Tracker rewrite and lazy process checker

muellerzr:
This PR rewrites the tracker classes to be a regular class with init checks rather than an abstract class, letting them be flexible enough but also verifying the information that *must* be there is included.

This PR also makes the process execution for trackers happen at the tracker level, rather than at the `Accelerator` level, so that trackers can be utilized on their own. To do so, a new `on_main_process` decorator was added in the `tracking.py` specifically which checks if the tracker class has a value for `on_main_process` to `True` or `False`. Depending on this the logging function (and/or init) will then execute. This tracker is executed lazily, so that the `PartialState` will have been created not at import time.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -78,7 +78,7 @@ class on_main_process:
 
     def __set_name__(self, owner, name):
         if getattr(self, "main_process_only", False):
-            setattr(owner, name, on_main_process(self.function))
+            setattr(owner, name, state.on_main_process(self.function))
         else:
             setattr(owner, name, self.function)
 
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -18,6 +18,7 @@
 import json
 import os
 import time
+from functools import wraps
 from typing import Any, Dict, List, Optional, Union
 
 import yaml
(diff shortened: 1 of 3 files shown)


muellerzr:
Post approval I'll ping the folks at W&B to ensure they agree and like the proposed API

src/accelerate/tracking.py
```diff
@@ -67,32 +68,70 @@
...
+    Implementations can also include a `main_process_only` (`bool`) attribute to toggle if relevent logging, init, and
+    other functions should occur on the main process or across all processes (by default will use `True`)
+    """
+
+    main_process_only = True
+
+    def __init__(self):
+        err = ""
+        if not hasattr(self, "name"):
+            err += "`requires_logging_directory`"
```
sgugger, line 117: Should be name here, no?

sgugger (approved):

muellerzr:
As discussed offline:

I put back in `@on_main_process` on the `Accelerator` object to maintain its API, however the new API can still exist so that a user can grab the tracker from the `Accelerator` object and have it work OOTB without the `if accelerator.is_main_process` block

src/accelerate/tracking.py
```diff
@@ -67,32 +68,71 @@
...
-    @abstractproperty
-    def requires_logging_directory(self):
-        """
-        Whether the logger requires a directory to store their logs. Should either return `True` or `False`.
-        """
-        pass
+    `name` (`str`): String representation of the tracker class name, such as "TensorBoard" `requires_logging_directory`
+    (`bool`): Whether the logger requires a directory to store their logs. `tracker` (`object`): Should return internal
+    tracking mechanism used by a tracker class (such as the `run` for wandb)
+
```
pacman100, line 107:
```suggestion
    `name` (`str`): String representation of the tracker class name, such as "TensorBoard"
    `requires_logging_directory` (`bool`): Whether the logger requires a directory to store their logs.
    `tracker` (`object`): Should return internal tracking mechanism used by a tracker class (such as the `run` for wandb)

```

pacman100 (approved):
Thank you for the refactoring of the trackers! Could you please put an example code highlighting the importance of changes in this PR for clarity?

PR #1080, opened 2023-02-15
Allow custom SageMaker Estimator arguments

dennisbappert:
The current implementation for SageMaker only allows a limited set of arguments, however the `Estimator` and `HuggingFace` Estimator support many additional arguments. With the current config wizard, this would be very hard to implement, hence I added a new field to `SageMakerConfig` to allow overrides to be set.

This e.g. allows the usage of Spot Instances.

### DIFF
--- a/docs/source/usage_guides/sagemaker.mdx
+++ b/docs/source/usage_guides/sagemaker.mdx
@@ -186,4 +186,4 @@ additional_args:
   max_wait: 86400
 ```
 
-*Note: Spot Instances are subject to be terminated and training to be continued from a checkpoint. This is not handled in 🤗 Accelerate out of the box. Contact us if you would like this feature.
+*Note: Spot Instances are subject to be terminated and training to be continued from a checkpoint. This is not handled in 🤗 Accelerate out of the box. Contact us if you would like this feature.*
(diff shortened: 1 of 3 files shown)


pacman100:
Hello @dennisbappert, how will user specify `additional_args` as I don't see it being part of the `accelerate config` questionarrie ?

dennisbappert:
I purposefully kept it out of the questionnaire, as it might be hard to enter in an interactive console session. It is mentioned in the documentation and for more advanced use-cases. I can add a hint in the questionnaire to guide the user towards the documentation. 

Do you think it will be valuable as an interactive input?

dennisbappert:
Yes it has to be added directly. I modified the docs to include an example for Spot Instances, let me make that a bit more clear in the docs.

sgugger:
cc @philschmid

dennisbappert:
see ^^^

philschmid (approved):
LGTM! Good idea on adding more configurations

dennisbappert:
I'm working on some more configuration right now: adding custom docker container support (Dockerfile in the repository), enhancing the documentation to cover local mode (which works fine, would just be nice as part of the documentation).

sgugger (approved):
LGTM, please ping us when you're happy with the documentation you're adding :-)

dennisbappert:
You can merge this one, I will create additional PRs for the other changes.

Issue #1078, opened 2023-02-15
'SageMakerConfig' object has no attribute 'tpu_cluster'


dennisbappert:
### System Info

```Shell
- `Accelerate` version: 0.17.0.dev0
- Platform: Linux-6.0.12-76060006-generic-x86_64-with-glibc2.35
- Python version: 3.10.9
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- `Accelerate` default config:
        - compute_environment: AMAZON_SAGEMAKER
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - dynamo_backend: NO
        - ec2_instance_type: ml.g4dn.xlarge
        - iam_role_name: accelerate_sagemaker_execution_role
        - profile: default
        - region: eu-west-1
        - num_machines: 1
        - gpu_ids: all
        - base_job_name: accelerate-sagemaker-1
        - pytorch_version: 1.10.2
        - transformers_version: 4.17.0
        - py_version: py38
```

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Steps to reproduce:

- Run `accelerate config`
- Choose SageMaker
- Use default values for the configuration

The newly added field `tpu_cluster` is not available but accessed.

```
Traceback (most recent call last):
  File "/opt/conda/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/launch.py", line 1095, in launch_command
    args.tpu_cluster = defaults.tpu_cluster and args.tpu
AttributeError: 'SageMakerConfig' object has no attribute 'tpu_cluster'
```

![image](https://user-images.githubusercontent.com/7000188/218992737-83dc6325-2bec-4b37-9dd5-507aa45d89f3.png)

### Expected behavior

```Shell
Works in 0.16.0, broken in 0.17.0.dev0
```

muellerzr:
Hi @dennisbappert, can you try again please using Accelerate via:

```bash
[...]

PR #1081, opened 2023-02-15
Fix tpu_cluster arg for sagemaker config

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/1078, and is reported that the solution works. Will merge after https://github.com/huggingface/accelerate/pull/1080 so that I get the merge-conflict headaches and not the community contributor 🤗


dennisbappert:
I can also squash if necessary, my fault :)

muellerzr:
Not necessary at all! :)

PR #1084, opened 2023-02-15
Added SageMaker local mode doc section

dennisbappert:
SageMaker supports local mode to test using the SageMaker DLC locally before starting the training job. This works out of the box, I added it to the configuration, I think it is valuable for users.


sgugger (approved):
Thanks!

PR #1086, opened 2023-02-15
Fp8 integration

sgugger:
This PR brings FP8 mixed precision training to Accelerate. Using this requires a GPU Hopper or higher (hard to find at the moment!) and the `transformers_engine` library.


src/accelerate/accelerator.py
```diff
@@ -123,10 +132,11 @@ class Accelerator:
             round multiple of the `num_processes` you are using. If `False`, actual batch size used will be the one set
             in your script multiplied by the number of processes.
         mixed_precision (`str`, *optional*):
-            Whether or not to use mixed precision training (fp16 or bfloat16). Choose from 'no','fp16','bf16'. Will
-            default to the value in the environment variable `ACCELERATE_MIXED_PRECISION`, which will use the default
-            value in the accelerate config of the current system or the flag passed with the `accelerate.launch`
-            command. 'fp16' requires pytorch 1.6 or higher. 'bf16' requires pytorch 1.10 or higher.
+            Whether or not to use mixed precision training (fp16 or bfloat16). Choose from 'no','fp16','bf16 or 'fp8'.
```
muellerzr, line 135:
```suggestion
            Whether or not to use mixed precision training (fp8, fp16, or bfloat16). Choose from 'no','fp16','bf16 or 'fp8'.
```

src/accelerate/accelerator.py
```diff
@@ -1144,6 +1160,25 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
...
+            kwargs = self.fp8_recipe_handler.to_kwargs() if self.fp8_recipe_handler is not None else {}
+            if "fp8_format" in kwargs:
+                kwargs["fp8_format"] = getattr(te_recipe.Format, kwargs["fp8_format"])
+            fp8_recipe = te_recipe.DelayedScaling(**kwargs)
+            fp8_enabled = torch.cuda.get_device_capability()[0] >= 9
+            if not fp8_enabled:
+                logger.warn(
+                    f"The current device has compute capability of {torch.cuda.get_device_capability()} which is "
+                    "insufficient for FP8 mixed precision training (requires a GPU Hopper or higher, compute "
+                    "capability of 9 or higher). Will using FP16 instead."
```
muellerzr, line 1179: Should this only warn? Or should it not explicitly raise an error?

src/accelerate/utils/transformer_engine.py
```diff
@@ -0,0 +1,73 @@
...
+import torch.nn as nn
+
+from .imports import is_fp8_available
+
+
+if is_fp8_available():
+    import transformer_engine.pytorch as te
+
+
+def convert_model(model, to_transformer_engine=True, _convert_linear=True, _convert_ln=True):
```
muellerzr, line 24: Should this run an explicit try/catch for `is_fp8_available` and raise an error if not?

examples/complete_cv_example.py
```diff
@@ -272,7 +272,7 @@ def main():
         "--mixed_precision",
         type=str,
         default=None,
-        choices=["no", "fp16", "bf16"],
+        choices=["no", "fp16", "bf16", "fp8"],
```
muellerzr, line 275:
```suggestion
        choices=["no", "fp8", "fp16", "bf16"],
```
Should we sort them all alphabetically/precision-ly?

muellerzr (approved):
Awesome job! I left a few notes and general questions, but great work :)

examples/complete_cv_example.py
sgugger: I like the order I put, which is by GPU generation (the later in the list, the more recent you need it, Volta, Ampere, Hopper)

muellerzr: Makes sense! :)

src/accelerate/accelerator.py
muellerzr: Based on the earlier comment, this could be "fp16, bfloat16, or fp8", or we remove the () and just have "Choose from"

pacman100 (approved):
Awesome! This is going to be a game-changer for LLM training. LGTM 🚀

src/accelerate/accelerator.py
sgugger: It still uses transformer engine instead of the regular model, so useful for testing on A100s

src/accelerate/utils/transformer_engine.py
sgugger: It shouldn't be called if `is_fp8_available` is `False`.

src/accelerate/accelerator.py
sgugger: Removing the parenthesis entirely.

JulesGM:
I just made two new issues with fp8.
https://github.com/huggingface/accelerate/issues/1276
https://github.com/huggingface/accelerate/issues/1275

Issue #1083, opened 2023-02-15
ClusterConfig.__init__() got an unexpected keyword argument 'tpu_downcast_bf16'


dennisbappert:
### System Info

```Shell
(current master)

- `Accelerate` version: 0.17.0.dev0
- Platform: Linux-6.0.12-76060006-generic-x86_64-with-glibc2.35
- Python version: 3.10.9
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (False)
- `Accelerate` default config:
        Not found
```

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

- `accelerate config`
- Use default settings everywhere

![image](https://user-images.githubusercontent.com/7000188/219116784-fe9ebcad-e3ad-44a0-8069-62997a0f4083.png)

```
Traceback (most recent call last):                                                                                                                                  
  File "/opt/conda/bin/accelerate", line 8, in <module>                                                                                                             
    sys.exit(main())                                                                                                                                                
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/config/config.py", line 67, in config_command
    config = get_user_input()
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/config/config.py", line 40, in get_user_input
    config = get_cluster_input()
  File "/opt/conda/lib/python3.10/site-packages/accelerate/commands/config/cluster.py", line 502, in get_cluster_input
    return ClusterConfig(
TypeError: ClusterConfig.__init__() got an unexpected keyword argument 'tpu_downcast_bf16'
```

### Expected behavior

```Shell
It should work :)
```

PR #1090, opened 2023-02-17
Fix config

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/1083

Sadly to my knowledge there's not a decent way to test for the CLI use case without running some extensive test-check specific cases inside the config questionnaire logic. I think it'd be worth it later to do so but not right now.


sgugger (approved):
Thanks for the fix!

PR #1091, opened 2023-02-17
Adds missing "lfs" in instructions for using git-lfs

CSchoel:
The instructions suggested to use `git pull`, but the command to get the actual files from git-lfs is `git lfs pull`.


sgugger (approved):
Thanks!

PR #1094, opened 2023-02-19
add multi_cpu support to reduce

### LINKED ISSUE #1058: Reduce support for multi_cpu
As far as I can tell, reduce for multi_cpu actually just performs a within-process instead of an across-process reduction. 

https://github.com/huggingface/accelerate/blob/36beea9a1be9251a3ad07f7c10fd425c96954f46/src/accelerate/utils/operations.py#L430-L443

alex-hh:
Fixes #1058 

N.B. for consistency with existing code this always performs sum, ignoring the reduction kwarg (c.f. #745)


sgugger (approved):
Thanks a lot!

PR #1100, opened 2023-02-22
Changed accelerator's fp16 parameter

BM-K:
From version 0.16.0, `mixed_precision='fp16'` must be given as a parameter when using accelerator fp16.


muellerzr:
Thanks for the fix!

PR #1108, opened 2023-02-23
Attempt to fix import error when PyTorch is build without `torch.distributed` module

mfuntowicz:
When torch is build without distributed support (non pypi release), importing `accelerate` crashs because it tries to import `torch.distributed.ReduceOp`.

This PR attempts to only import when torch has support for distributed setups along with trying to not have too much impact on CCL call performances.

_I'm not quite sure how to test this setup as it, if you have any suggestion ..._

Root error stacktrace:

```python
  File "/usr/local/lib/python3.10/dist-packages/accelerate/utils/operations.py", line 24, in <module>
    from torch.distributed import ReduceOp
ImportError: cannot import name 'ReduceOp' from 'torch.distributed' (/usr/local/lib/python3.10/dist-packages/torch/distributed/__init__.py)
```

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -224,8 +224,8 @@ def gather(tensor):
     """
     if not is_torch_distributed_available():
         raise ModuleNotFoundError(
-            "torch was compiled without distributed support (torch.distributed.is_available() = False). "
-            "Please use a version which includes distributed support in order to call #gather()"
+            "`torch` was compiled without distributed support (`torch.distributed.is_available() == False`). "
+            "Please install a version which includes distributed support in order to call gather."
         )
 
     if PartialState().distributed_type == DistributedType.TPU:
@@ -263,8 +263,8 @@ def gather_object(object: Any):
     """
     if not is_torch_distributed_available():
         raise ModuleNotFoundError(
-            "torch was compiled without distributed support (torch.distributed.is_available() = False). "
-            "Please use a version which includes distributed support in order to call #gather_object()"
+            "`torch` was compiled without distributed support (`torch.distributed.is_available() == False`). "
+            "Please install a version which includes distributed support in order to call gather_object."
         )
 
     if PartialState().distributed_type == DistributedType.TPU:
@@ -308,8 +308,8 @@ def broadcast(tensor, from_process: int = 0):
     """
     if not is_torch_distributed_available():
         raise ModuleNotFoundError(
-            "torch was compiled without distributed support (torch.distributed.is_available() = False). "
-            "Please use a version which includes distributed support in order to call #broadcast()"
+            "`torch` was compiled without distributed support (`torch.distributed.is_available() == False`). "
+            "Please install a version which includes distributed support in order to call broadcast."
         )
 
     if PartialState().distributed_type == DistributedType.TPU:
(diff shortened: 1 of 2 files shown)


muellerzr:
I *think* it's a bit more complicated than this check (unless you've checked this on a non-cuda built system), as I ran into this as well when I was trying to refactor argparse. To test we want to build using the CPU-only python wheels, which the CPU CI should technically do. 

If so, I think adding it to `imports.py` instead would be better because there's more areas we can use it in. (It can still raise the flag if we want to raise the err)

mfuntowicz:
Thanks for additional context and sorry for this not being complete, I'm discovering the accelerate codebase 😅.

Within `imports.py` you'd like to see a `is_torch_distributed_available` function rather than having it only in `operations.py`? Did I get it right?

Regarding testing, CPU-only build can have distributed enabled too, not sure this is possible without building torch ourselve? (`gloo` / `oneCCL` notably are CPU compatible / only CCL providers)

muellerzr:
Yep! Exactly :) It'd need to be a disabled torch version install, I'll see if it's easy to do that via `light-the-torch`

mfuntowicz:
Let me know if this is what you were expecting 👍🏻

src/accelerate/utils/operations.py
```diff
@@ -219,6 +222,12 @@ def gather(tensor):
     Returns:
         The same data structure as `tensor` with all tensors sent to the proper device.
     """
+    if not is_torch_distributed_available():
+        raise ModuleNotFoundError(
+            "torch was compiled without distributed support (torch.distributed.is_available() = False). "
+            "Please use a version which includes distributed support in order to call #gather()"
```
muellerzr, line 228:
```suggestion
            "`torch` was compiled without distributed support (`torch.distributed.is_available() == False`). "
            "Please install a version which includes distributed support in order to call `gather`."
```
Let's format these errs a little differently, otherwise LG2M!

muellerzr (approved):

muellerzr:
cc @sgugger to make sure it all looks fine to you as well :)

muellerzr:
(Also, we do install the non-distributed torch wheels during the CI, that's how I know this issue exists so the fact we're passing means it works :) )

mfuntowicz:
Thanks for checking on the tests side 🤗

src/accelerate/utils/operations.py
```diff
@@ -219,6 +222,12 @@ def gather(tensor):
     Returns:
         The same data structure as `tensor` with all tensors sent to the proper device.
     """
+    if not is_torch_distributed_available():
+        raise ModuleNotFoundError(
+            "`torch` was compiled without distributed support (`torch.distributed.is_available() == False`). "
+            "Please install a version which includes distributed support in order to call gather."
+        )
```
sgugger, line 229: Why do we need to raise here? Users can definitely call `gather` on a single process with distributed support. That's tje promise of Accelerate: the script will run **the same and without errors** on any setup when you use accelerate primitives.

Since we can't be launched distributed-ly without distributed support, we should just leave this as is and the last elese with return the `tensor` or am I missing something?

src/accelerate/utils/operations.py
```diff
@@ -252,6 +261,12 @@ def gather_object(object: Any):
     Returns:
         The same data structure as `object` with all the objects sent to every device.
     """
+    if not is_torch_distributed_available():
```
sgugger, line 264: Same here and all below

sgugger (commented):
Thanks for working on this! I don't understand the newly added errors which shouldn't be raised.

src/accelerate/utils/operations.py
muellerzr: Great point, yes I believe we should just return the tensor and `noop`

mfuntowicz: Didnt have this philosophy in mind, will do the changes 👌

sgugger (approved):
Perfect, thanks!

PR #1118, opened 2023-02-26
Intel GPU support initialization

abhilash1910:
Intel GPU Support for Accelerate Framework .
Initial Draft PR :
-Feature addition for XPU devices backend
- CCL bindings for XPU devices
- Sanity checks
PR is in progress


src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -53,6 +54,7 @@ def print_on(state, process_idx):
 def process_execution_check():
     accelerator = Accelerator()
     num_processes = accelerator.num_processes
+    print(num_processes,'num_process')
```
sgugger, line 57: To clean up

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -23,7 +23,7 @@
 from accelerate import Accelerator
 from accelerate.test_utils import RegressionDataset, RegressionModel
 from accelerate.utils import DistributedType, set_seed
-
+import intel_extension_for_pytorch as ipex
```
sgugger, line 26: Should be removed.

src/accelerate/accelerator.py
```diff
@@ -380,6 +381,8 @@ def __init__(
                 self.scaler = ShardedGradScaler(**kwargs)
             else:
                 self.scaler = torch.cuda.amp.GradScaler(**kwargs)
+                if is_xpu_available():
+                    self.scaler = torch.xpu.amp.GradScaler(**kwargs)
```
sgugger, line 385:
```suggestion
                if is_xpu_available():
                    self.scaler = torch.xpu.amp.GradScaler(**kwargs)
                else:
                    self.scaler = torch.cuda.amp.GradScaler(**kwargs)
```

src/accelerate/accelerator.py
```diff
@@ -2451,11 +2454,17 @@ def autocast(self):
         if self.native_amp:
             if self.mixed_precision == "fp16" and is_torch_version(">=", "1.10"):
                 autocast_context = torch.cuda.amp.autocast(dtype=torch.float16)
+                if is_xpu_available():
+                    autocast_context = torch.xpu.amp.autocast(dtype=torch.float16)
```
sgugger, line 2458: Same logic here, don't set `autocast_context` to then change it.

src/accelerate/accelerator.py
```diff
@@ -2451,11 +2454,17 @@ def autocast(self):
...
             elif self.mixed_precision == "bf16":
                 if self.distributed_type in [DistributedType.NO, DistributedType.MULTI_CPU, DistributedType.MULTI_GPU]:
                     autocast_context = torch.autocast(dtype=torch.bfloat16, device_type=self.device.type)
+                    if is_xpu_available():
```
sgugger, line 2462: Same here.

src/accelerate/accelerator.py
```diff
@@ -2451,11 +2454,17 @@ def autocast(self):
...
+                        autocast_context = torch.xpu.amp.autocast(dtype=torch.bfloat16, device_type=self.device.type)
             else:
                 autocast_context = torch.cuda.amp.autocast()
+                if is_xpu_available():
```
sgugger, line 2466: Same there/

src/accelerate/hooks.py
```diff
@@ -531,6 +532,8 @@ def __init__(
             self.execution_device = torch.device("mps")
         elif torch.cuda.is_available():
             self.execution_device = torch.device(0)
+        elif is_xpu_available():
+            self.execution_device = torch.device("xpu")
```
sgugger, line 536: Should probably be before the CUDA line.

src/accelerate/state.py
```diff
@@ -107,7 +107,11 @@ def __init__(self, cpu: bool = False, **kwargs):
                     self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                     if self.device is None:
                         self.device = torch.device("cuda", self.local_process_index)
+                        if is_xpu_available():
+                            self.device = torch.device("xpu", self.local_process_index)
                     torch.cuda.set_device(self.device)
+                    if is_xpu_available():
+                        torch.xpu.set_device(self.device)
```
sgugger, line 114:
```suggestion
                    if is_xpu_available():
                        torch.xpu.set_device(self.device)
                    else:
                        torch.cuda.set_device(self.device)
```

src/accelerate/state.py
```diff
@@ -107,7 +107,11 @@ def __init__(self, cpu: bool = False, **kwargs):
                     self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                     if self.device is None:
                         self.device = torch.device("cuda", self.local_process_index)
+                        if is_xpu_available():
+                            self.device = torch.device("xpu", self.local_process_index)
```
sgugger, line 111:
```suggestion
                        if is_xpu_available():
                            self.device = torch.device("xpu", self.local_process_index)
                        else:
                            self.device = torch.device("cuda", self.local_process_index)
```

src/accelerate/state.py
```diff
@@ -121,33 +125,45 @@ def __init__(self, cpu: bool = False, **kwargs):
...
                 self.num_processes = torch.distributed.get_world_size()
                 self.process_index = torch.distributed.get_rank()
                 self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                 if self.device is None:
                     self.device = torch.device("cuda", self.local_process_index)
+                    if is_xpu_available():
+                        self.device = torch.device("xpu", self.local_process_index)
                 torch.cuda.set_device(self.device)
+                if is_xpu_available():
+                    torch.xpu.set_device(self.device)
```
sgugger, line 147: See comments above.

src/accelerate/test_utils/scripts/test_cli.py
```diff
@@ -1,12 +1,19 @@
 import torch
-
+import intel_extension_for_pytorch as ipex
```
sgugger, line 2: This import is not used, and we can't make `accelerate test` have a hard dependency on ipex, so this should be removed.

src/accelerate/test_utils/scripts/test_cli.py
```diff
@@ -1,12 +1,19 @@
...
         num_gpus = torch.cuda.device_count()
     else:
         num_gpus = 0
     print(f"Successfully ran on {num_gpus} GPUs")
+    
+    if torch.xpu.is_available():
+        num_xpus = torch.xpu.device_count()
+    else:
+        num_xpus = 0
+    print(f"Successfully ran on {num_xpus} XPUs")
```
sgugger, line 15: Should be removed as well, probably.

src/accelerate/utils/megatron_lm.py
```diff
@@ -300,8 +300,12 @@ def cyclic_iter(iter):
             do_test = test_dataloader is not None and args.eval_iters > 0
             # Need to broadcast num_tokens and num_type_tokens.
             flags = torch.cuda.LongTensor([int(do_train), int(do_valid), int(do_test)])
+            if is_xpu_available():
+                flags = torch.xpu.LongTensor([int(do_train), int(do_valid), int(do_test)])
         else:
             flags = torch.cuda.LongTensor([0, 0, 0])
+            if is_xpu_available():
+                flags = torch.xpu.LongTensor([0, 0, 0])
```
sgugger, line 308: Same comment as before. Don't initialize a variable on GPU to then erase it on put on XPU. Also not sure Megatron LM will support torch xpu so I don't think this is necesary.

src/accelerate/utils/memory.py
```diff
@@ -52,6 +52,10 @@ def release_memory(*objects):
         objects[i] = None
     gc.collect()
     torch.cuda.empty_cache()
+    try:
+        torch.xpu.empty_cache()
+    except AttributeError:
+        pass
```
sgugger, line 58: Protect this by a `is_xpu_available` instead (here and the two other below).

src/accelerate/utils/modeling.py
```diff
@@ -322,15 +322,22 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
     Get the maximum memory available if nothing is passed, converts string to int otherwise.
     """
     import psutil
-
+     
     if max_memory is None:
         if not torch.cuda.is_available():
             max_memory = {}
+        if not is_xpu_available():
+            max_memory = {}
```
sgugger, line 330: This will replace `max_memory` on any setup having GPUs but not XPUs, so we can't change the file like this.

sgugger (commented):
Thnaks for opening this PR. I think it tries to do way too much at the same time and should be split in smaller blocks: we don't need to have support for DeepSpeed +XPUs (is that even a thing?) or MegatronLM + XPUs (same) or big model inference on XPUs all in the same PR as enabling XPUs for distributed training.

I suggest starting smaller by just pushing support for distributed training, then we can add the other features if there is interest from the community.

abhilash1910:
Hi @sgugger , the PR is still in debug state, we are working internally to ensure it functionally works. I will make the recommended changes in the meantime , before I assign a reviewer.

src/accelerate/utils/constants.py
```diff
@@ -28,7 +28,7 @@
 FSDP_AUTO_WRAP_POLICY = ["TRANSFORMER_BASED_WRAP", "SIZE_BASED_WRAP", "NO_WRAP"]
 FSDP_BACKWARD_PREFETCH = ["BACKWARD_PRE", "BACKWARD_POST", "NO_PREFETCH"]
 FSDP_STATE_DICT_TYPE = ["FULL_STATE_DICT", "LOCAL_STATE_DICT", "SHARDED_STATE_DICT"]
-DEEPSPEED_MULTINODE_LAUNCHERS = ["pdsh", "standard", "openmpi", "mvapich"]
+DEEPSPEED_MULTINODE_LAUNCHERS = ["pdsh", "standard", "openmpi", "mvapich","mpich"]
```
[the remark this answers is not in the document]
abhilash1910: yes this should be cleaned

src/accelerate/test_utils/scripts/test_cli.py
abhilash1910: The test scripts will be  cleaned. (includes test code changes inside test_utils)

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -24,7 +24,7 @@
 
 from accelerate.accelerator import Accelerator
 from accelerate.utils.dataclasses import DistributedType
-
+import intel_extension_for_pytorch as ipex
```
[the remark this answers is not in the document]
abhilash1910: Yes this would be removed and cleaned.

src/accelerate/state.py
```diff
@@ -106,8 +106,12 @@ def __init__(self, cpu: bool = False, **kwargs):
                     self.process_index = torch.distributed.get_rank()
                     self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                     if self.device is None:
-                        self.device = torch.device("cuda", self.local_process_index),
+                        self.device = torch.device("cuda", self.local_process_index)
+                        if is_xpu_available():
+                            self.device = torch.device("xpu", self.local_process_index)
```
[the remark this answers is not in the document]
abhilash1910: So something like :
self.device = torch.device("xpu",self.local_process_index) if is_xpu_available() else torch.device("cuda",self.local_process_index)

src/accelerate/accelerator.py
mingxiaoh: sorry but there is no this api "torch.xpu.amp.GradScaler"

abhilash1910:
@sgugger do you suggest to remove Megatron LM support for XPU ? I guess that seems ok considering the PR first tries to enable distributed training

src/accelerate/utils/modeling.py
abhilash1910: Resolved for now.

src/accelerate/commands/env.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+[#!/usr/bin/env python
```
mingxiaoh, line 1: remove "[" please,

src/accelerate/utils/launch.py
```diff
@@ -101,7 +104,9 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> Tuple[List[str]
     if args.cpu or args.use_cpu:
         if args.ipex_enabled:
             current_env["IPEX_ENABLED"] = str(args.ipex_enabled).lower()
-
+     elif is_xpu_available():
```
mingxiaoh, line 107: IndentationError: unindent does not match any outer indentation level

src/accelerate/utils/imports.py
```diff
@@ -101,6 +110,8 @@ def is_bf16_available(ignore_tpu=False):
     if is_torch_version(">=", "1.10"):
         if torch.cuda.is_available():
             return torch.cuda.is_bf16_supported()
+        elif torch.xpu.available():
```
mingxiaoh, line 113: should be torch.xpu.is_available()

src/accelerate/utils/imports.py
```diff
@@ -101,6 +110,8 @@ def is_bf16_available(ignore_tpu=False):
...
+            return torch.xpu.is_bf16_supported()
```
mingxiaoh, line 114: there is no this api  torch.xpu.is_bf16_supported() at all

mingxiaoh (approved):

src/accelerate/hooks.py
```diff
@@ -22,6 +22,8 @@
 from .utils import (
     PrefixedDataset,
     find_device,
+    is_mps_available,
```
mingxiaoh, line 25: why we need to touch mps?

abhilash1910: This was already there ?

sgugger:
As I said before, please keep the changes minimal to progressively enable support in small increments that are easy to review. The first integration should focus only on a small training on XPU. It shouldn't touch things like big model inference, megatron LM etc.

abhilash1910:
@sgugger @pacman100   Is there a plan to support comm backend of deepspeed for deepspeed invocation ("ACCELERATE_USE_DEEPSPEED") ? Or only through torch.distributed permanently? ( https://github.com/huggingface/accelerate/pull/1138 , https://github.com/microsoft/DeepSpeed/issues/2117)

sgugger:
@abhilash1910 We can discuss this in another PR. I feel you are not listening to the feedback we are given you here. Let's first support simple trainings on XPUs **then** we can iterate to add support for more integrations.

abhilash1910:
sure @sgugger  , we are finalizing the changes on our end to just support simple trainings on XPU. Since state.py is foundational , hence thought of asking about it.

abhilash1910:
@sgugger requesting review . As per the directives, only simplified distributed training on XPUs is enabled on this PR; do let us know on this.

src/accelerate/accelerator.py
```diff
@@ -1608,6 +1628,38 @@ def _prepare_ipex(self, *args):
...
+        result = [obj for obj in args]
+        for obj in result:
+            if isinstance(obj, torch.nn.Module):
+                model = obj
+            elif isinstance(obj, (torch.optim.Optimizer)):
+                optimizer = obj
+        model.to(self.device)
+        if optimizer is not None:
+            if not model.training:
+                model.train()
```
sgugger, line 1650: This is not something we can do magically behind the scenes for the user. You should raise an error her if this can't be used for evaluation.

src/accelerate/accelerator.py
```diff
@@ -1608,6 +1628,38 @@ def _prepare_ipex(self, *args):
...
+            model, optimizer = torch.xpu.optimize(
+                model, dtype=xpu_plugin.dtype, optimizer=optimizer, inplace=True, level="O1"
+            )
+            model.forward = torch.xpu.amp.autocast(dtype=xpu_plugin.dtype)(model.forward)
```
sgugger, line 1654: We should re-use the normal mixed precision argument here.

src/accelerate/checkpointing.py
```diff
@@ -100,7 +101,10 @@ def save_accelerator_state(
     states["random_state"] = random.getstate()
     states["numpy_random_seed"] = np.random.get_state()
     states["torch_manual_seed"] = torch.get_rng_state()
-    states["torch_cuda_manual_seed"] = torch.cuda.get_rng_state_all()
+    if is_xpu_available():
+        states["torch_xpu_manual_seed"] = torch.xpu.get_rng_state_all()
+    else:
+        states["torch_cuda_manual_seed"] = torch.cuda.get_rng_state_all()
     # ^^ safe to call this function even if cuda is not available
```
sgugger, line 108: This comment should be removed if it's not true anymore.

src/accelerate/checkpointing.py
```diff
@@ -183,7 +187,10 @@ def load_accelerator_state(
         random.setstate(states["random_state"])
         np.random.set_state(states["numpy_random_seed"])
         torch.set_rng_state(states["torch_manual_seed"])
-        torch.cuda.set_rng_state_all(states["torch_cuda_manual_seed"])
+        if is_xpu_available():
+            torch.xpu.set_rng_state_all(states["torch_xpu_manual_seed"])
+        else:
+            torch.cuda.set_rng_state_all(states["torch_cuda_manual_seed"])
         # ^^ safe to call this function even if cuda is not available
```
sgugger, line 194: Same here.

src/accelerate/commands/config/default.py
```diff
@@ -62,6 +62,14 @@ def write_basic_config(mixed_precision="no", save_location: str = default_json_c
             config["distributed_type"] = "MULTI_GPU"
         else:
             config["distributed_type"] = "NO"
+    if is_xpu_available():
+        num_gpus = torch.xpu.device_count()
```
sgugger, line 66:
```suggestion
        num_xpus = torch.xpu.device_count()
```
Since they are called XPU everywhere...

src/accelerate/commands/env.py
```diff
@@ -55,6 +56,7 @@ def env_command(args):
         "Python version": platform.python_version(),
         "Numpy version": np.__version__,
         "PyTorch version (GPU?)": f"{pt_version} ({pt_cuda_available})",
+        "PyTorch version (GPU?)": f"{pt_version} ({pt_xpu_available})",
```
sgugger, line 59:
```suggestion
        "PyTorch XPU available": str(pt_xpu_available),
```
No need to repeat the version here.

src/accelerate/utils/dataclasses.py
```diff
@@ -1346,3 +1349,23 @@ def set_mixed_precision(self, mixed_precision):
...
             self.dtype = torch.bfloat16
+
+@dataclass
+class XPUPlugin:
+    """
+    This plugin is used to enable XPU plugin with IPEX.
+    """
+
+    use_xpu: bool = field(default=None, metadata={"help": "Enable XPU Plugin"})
+    dtype: torch.dtype = field(default=torch.float32, metadata={"help": "Enable mixed precision on XPU"})
```
sgugger, line 1360: The dtype should be the mixed_precision argument we already have so shouldn't be here. Do you expect to have other arguments coming here? If no, there is no need for a plugin and we can simplify the integration.

src/accelerate/utils/imports.py
```diff
@@ -191,3 +200,22 @@ def get_major_and_minor_from_version(full_version):
...
+
+@lru_cache()
+def is_xpu_available(check_device=True):
+    "Checks if `intel_extension_for_pytorch` is installed and potentially if a XPU is in the environment"
+    if is_ipex_available():
+        import torch
+        if version.parse(version.parse(torch.__version__).base_version) < version.parse("1.13"):
+            return False
+    else:
+        return False
```
sgugger, line 212: Let's use importlib here to have the check with `check_device=False` be cheap to execute.

sgugger (commented):
Thanks a lot for all your work here. I have flagged in comments a couple of places where we have breaking changes that really need to be addressed and left other comments on the style.

Could you also run `make style` on your branch to make sure the quality jobs pass?

src/accelerate/utils/imports.py
abhilash1910: Ok there is already a ```is_torch_version``` in versions.py which invokes the importlib which can be used.

src/accelerate/utils/dataclasses.py
abhilash1910: FP16 mixed precision is not supported yet through xpu . Yes so for args if it comes in as FP16 it will raise an issue .

abhilash1910:
@sgugger ran make style and addressed reviews. Could you take a look again ? (Have a doubt in one comment regarding model.train - https://github.com/huggingface/accelerate/pull/1118/#discussion_r1178174712) Thanks

src/accelerate/accelerator.py
sgugger: Yes, so that the user has to manually do the `model.train()` and not be surprised.

sgugger (commented):
Clarified my comment on the training, and I do insist on re-using the existing mixed_precision. You can check when the state is initialized and raise an error if there is a non-supported value for XPUs there (mixed precision cannot change once the state has been initialized).

src/accelerate/accelerator.py
abhilash1910: Got it , makes sense. I think  I can remove the model.train() aspect since it is based  on users' discretion. 
Update: Added commits

src/accelerate/accelerator.py
abhilash1910: Update: Yes should add normal Mix precision; added.

src/accelerate/utils/dataclasses.py
abhilash1910: Yes resolved in new commit. I agree the default mix precision should be used and flagged accordingly for device support.

abhilash1910:
Thanks @sgugger  for the detailed review. updated the commits as per suggestions. (+ test npr /styling).
Could you take a look ?

sgugger (commented):
Thanks a lot for addressing all of those comments every thing looks good apart from the XPUPlugin which was a question in my review from a couple of days ago: why have a plugin (both for XPU or IPEX) that only has one flag? And like for the MPS device, we want the XPUs to be used automatically if there are available (unless the user specified `use_cpu=True`) so that flag `use_xpu` should probably not be there as well to make the API easier to use (there initially was a `use_mps` flag we deprecated for context).

I think the whole `XPUPlugin` should just be removed, and if XPUs are detected, they are used (just like CUDA GPUs or MPS devices are automatically used when available). Unless there is a potential conflict between XPUs and CUDA GPUs? But it seems from what I read they are mutually exclusive.

src/accelerate/test_utils/scripts/check_main_process_first.txt
```diff
@@ -0,0 +1 @@
+Currently in the main process
```
sgugger, line 1: Is this used anywhere?

src/accelerate/state.py
```diff
@@ -130,8 +131,12 @@ def __init__(self, cpu: bool = False, **kwargs):
...
                 self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
                 if self.device is None:
-                    self.device = torch.device("cuda", self.local_process_index)
-                torch.cuda.set_device(self.device)
+                    if is_xpu_available():
+                        self.device = torch.device("xpu", self.local_process_index)
+                        torch.xpu.set_device(self.device)
+                    else:
+                        self.device = torch.device("cuda", self.local_process_index)
+                        torch.cuda.set_device(self.device)
```
sgugger, line 139: This needs to be executed is `self.device` is not `None` and we have a CUDA device.
Likewise, the `torch.xpu.set_device(self.device)` line above probably needs to be executed if `self.device` is not `None` and we have an XPU device.

src/accelerate/state.py
```diff
@@ -539,6 +550,7 @@ def __init__(
         fsdp_plugin=None,
         megatron_lm_plugin=None,
         ipex_plugin=None,
+        xpu_plugin=None,
```
sgugger, line 553: This isn't passed anymore, so should be removed, no?

src/accelerate/state.py
```diff
@@ -594,6 +607,11 @@ def __init__(
                     self.ipex_plugin = ipex_plugin if ipex_plugin.use_ipex else None
                     if self.ipex_plugin is not None:
                         self.ipex_plugin.set_mixed_precision(mixed_precision)
+            if self.distributed_type in [DistributedType.MULTI_XPU, DistributedType.NO]:
+                if self.device.type == "xpu" and xpu_plugin is not None:
+                    self.xpu_plugin = xpu_plugin if xpu_plugin.use_xpu else None
+                    if self.xpu_plugin is not None:
+                        self.xpu_plugin.set_mixed_precision(mixed_precision)
```
sgugger, line 614: Same here.

sgugger (commented):
Thanks a lot for iterating! We're very close to have this merged :-)
There are a couple of lines of code that are left with the XPU plugin in the state module that need to be cleaned up.

At this stage, I'd also like if @muellerzr could have a second look on the PR.

src/accelerate/test_utils/scripts/check_main_process_first.txt
muellerzr: This is the extra txt file generated that https://github.com/huggingface/accelerate/pull/1381 got rid of. So should be removed fully before merging.

src/accelerate/accelerator.py
```diff
@@ -430,12 +432,15 @@ def __init__(
...
                 self.scaler = torch.cuda.amp.GradScaler(**kwargs)
+
         elif self.state.mixed_precision == "bf16" and self.distributed_type not in (
             DistributedType.DEEPSPEED,
             DistributedType.MEGATRON_LM,
         ):
             if self.device.type == "cpu":
                 self.native_amp = is_torch_version(">=", "1.10")
+            elif self.device.type == "xpu":
+                self.native_amp = is_torch_version(">=", "1.10")
```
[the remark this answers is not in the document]
abhilash1910: resolved

src/accelerate/state.py
abhilash1910: Resolved

src/accelerate/accelerator.py
```diff
@@ -1135,6 +1140,9 @@ def prepare(self, *args, device_placement=None):
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.NO]:
             if self.device.type == "cpu" and self.state.ipex_plugin is not None:
                 args = self._prepare_ipex(*args)
+        elif self.distributed_type in [DistributedType.MULTI_XPU]:
+            if self.device.type == "xpu" and self.state.ipex_plugin is not None:
+                args = self._prepare_ipex(*args)
```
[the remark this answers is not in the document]
abhilash1910: Added the change

abhilash1910:
Thanks @sgugger  @muellerzr for pointing out the details. Let me know if this is correct?

muellerzr (approved):
This all looks good to me, however can we address the failing tests? :)

abhilash1910:
@muellerzr  Thanks for reviewing . Fixed the error in checkpointing tests, but the other tests fail since ipex is not installed .
I also added a small change to allow xpu amp to be used if xpu devices are detected rather (prepare_ipex method). With ipex installed (with or without xpus) test passes.
Could you please review ?

src/accelerate/accelerator.py
```diff
@@ -1135,6 +1138,9 @@ def prepare(self, *args, device_placement=None):
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.NO]:
             if self.device.type == "cpu" and self.state.ipex_plugin is not None:
                 args = self._prepare_ipex(*args)
+        elif self.distributed_type == DistributedType.MULTI_XPU:
+            if self.device.type == "xpu" and self.state.ipex_plugin is not None:
+                args = self._prepare_ipex(*args)
```
sgugger, line 1143: Those checks now need to be adapted since the plugin is always there.

sgugger (commented):
The tests shouldn't fail if `ipex` is not installed. The goal is not to add a hard dependency on `ipex` but to use it when it's available. On CPU, Ipex should be used if it's available only. On XPUs I think the error could be changed to a warning to install Ipex.

abhilash1910:
@sgugger  @muellerzr seems like this commit should fix the tests associated (waiting for CI, but locally this CI test passes). 
Thanks again for the comments.

jgong5:
> The tests shouldn't fail if `ipex` is not installed. The goal is not to add a hard dependency on `ipex` but to use it when it's available. On CPU, Ipex should be used if it's available only. On XPUs I think the error could be changed to a warning to install Ipex.

@sgugger XPU is still out of PyTorch tree and supported via IPEX. Before we upstream XPU support to PyTorch, its functionality depends on the availability of IPEX. Can we skip XPU tests if IPEX is not installed?

abhilash1910:
@sgugger @muellerzr  seems like tests pass now ( no longer a hard dependency on ipex installation). 
Let us know on this? :)

sgugger:
Yes, let us just run all slow tests on this and if all goes well it can be merged. Thanks a lot for all your efforts on this!

muellerzr:
@abhilash1910 we have a number of tests failing:

On single GPU:
```
 FAILED tests/test_modeling_utils.py::ModelingUtilsTester::test_get_balanced_memory - AssertionError: {0: 215, 1: 300} != {0: 300, 1: 300}
- {0: 215, 1: 300}
?     ^^^

+ {0: 300, 1: 300}
?     ^^^
```

On multi-GPU:
```
FAILED tests/test_big_modeling.py::BigModelingTester::test_dispatch_model_bnb - AssertionError: False is not true
```
And most concerning, when testing the example scripts, it looks like the Accelerator is prioritizing the CPU over the GPU, even when ipex isn't available. This was tested on two T4's running the following:

```bash
CUDA_VISIBLE_DEVICES="0" pytest -sv tests/test_examples.py -k test_checkpointing_by_epoch
```
And in this script (located at `examples/by_feature/checkpointing.py` I changed lines 124-130 to be:
```diff
def training_function(config, args):
    # For testing only
    if os.environ.get("TESTING_MOCKED_DATALOADERS", None) == "1":
        config["num_epochs"] = 2
    # Initialize accelerator
    accelerator = Accelerator(cpu=args.cpu, mixed_precision=args.mixed_precision)
+    assert accelerator.device.type == "cuda", f'Device: {accelerator.device}, type: {accelerator.device.type}'
```

abhilash1910:
@muellerzr is the failure only on tests with modelling.py ? Prioritisation is strange ,maybe due to checking for xpu before cuda in some places . I will see why this fails .

muellerzr:
@abhilash1910 the two test files are there in the trace I gave, `test_modeling_utils` and `test_big_modeling`. All the example tests are failing silently because what should take 6 minutes to run to completion takes 30+ minutes

src/accelerate/utils/modeling.py
```diff
@@ -482,10 +488,14 @@ def get_balanced_memory(
     # Get default / clean up max_memory
     max_memory = get_max_memory(max_memory)
 
-    if not torch.cuda.is_available():
+    if not (torch.cuda.is_available() and is_xpu_available()):
```
mingxiaoh, line 491: should change to if not (torch.cuda.is_available() or is_xpu_available()):

mingxiaoh (changes requested):

abhilash1910:
@muellerzr  does this still fail in the slow tests?

muellerzr:
@abhilash1910 we're now failing our CPU tests (notice that `test_checkpoint_step` has been running for 2+ hours and not passing. Can you try running `make test_checkpoint_step` to see if it passes? I'll run the slow tests after we get an all green again. (This should take *no more* than 2-3 minutes to run)

muellerzr:
@abhilash1910 still not using CUDA on the example tests. Did you try making my modification to the checkpointing script to make sure it was using CDUA in practice? (If you don't have a GPU to use, you could try using Google Colab to test)

abhilash1910:
Hi @muellerzr  , I did some testing on colab for checkpointing.py (using the assertion for cuda which you gave) with the xpu wheel build(1.19.9 -dev build) , and when I specify args.cpu=None it does pick up the GPU cuda device . I am sharing a snapshot: 
image
The colab is here: https://colab.research.google.com/drive/15NZd13igA_S-jresSiKVcb6p2qXatziI?usp=sharing
This is also seen with the default version 1.19 of accelerate . So with cpu=None , it does pick up the cuda and assertion passes. Could you please check if anything is missing ? I will re-check again .

muellerzr:
@abhilash1910 this also needs to work *without* the XPU build, which is where I believe we're facing the breaking issue :) As it still shows `cpu` for me

abhilash1910:
@muellerzr  I had tried with the same notebook with the default public accelerate(1.19.0)  but saw the same results  on colab(standard public - 1.19.0 ; xpu wheel - 1.19.0-dev 0). (as mentioned here: https://github.com/huggingface/accelerate/pull/1118#issuecomment-1539592556) . Both the public and the xpu build picks up the cuda properly.  I put args.cpu=None so as to use GPU cuda device in both cases.(since colab has no xpu ). Could you suggest which scripts to check further? Thanks

abhilash1910:
Thanks @muellerzr  for sharing the test notebook. It was a tricky issue  to resolve, and now it passes the slow tests (both checkpoint & and modeling utils) on T4 with cuda getting used.  Let me know on this ?

muellerzr:
Great, slow tests are running

muellerzr (approved):
Tests are passing fine here! cc @sgugger for a final look

muellerzr:
@abhilash1910 it seems this has broken something in diffusers, please see this issue: https://github.com/huggingface/accelerate/issues/1420

Any insight towards what could be going wrong here? This is breaking some of our other libraries

src/accelerate/accelerator.py
```diff
@@ -322,7 +323,7 @@ def __init__(
                 raise ImportError("Megatron is not installed. please build it from source.")
 
         if ipex_plugin is None:  # init from env variables
-            ipex_plugin = IntelPyTorchExtensionPlugin() if os.environ.get("IPEX_ENABLED", "false") == "true" else None
```
pacman100, line 325: Hello, removing this is leading to various test failures as we are trying to integrate accelerate in transformers trainer. We are planning to revert this line and wanted to check if there was a reason for removing this.

PR #1121, opened 2023-02-27
[Big model loading] Correct GPU only loading

patrickvonplaten:
There was a bug when loading safetensors directly on GPU via `device_map="auto"` in `diffusers`.

The test shows the failure case.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -678,10 +678,10 @@ def load_state_dict(checkpoint_file, device_map=None):
             return safe_load_file(checkpoint_file)
         else:
             devices = [device for device in device_map.values() if device not in ["disk"]]
-            
+
             # if we only have one device we can load everything directly
             if len(devices) == 1:
-                return safe_load_file(checkpoint_file, device=devices[0]) 
+                return safe_load_file(checkpoint_file, device=devices[0])
             # cpu device should always exist as fallback option
             devices = set(devices + ["cpu"])
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -678,6 +678,10 @@ def load_state_dict(checkpoint_file, device_map=None):
             return safe_load_file(checkpoint_file)
         else:
             devices = [device for device in device_map.values() if device not in ["disk"]]
+            
+            # if we only have one device we can load everything directly
+            if len(devices) == 1:
+                return safe_load_file(checkpoint_file, device=devices[0]) 
             # cpu device should always exist as fallback option
             devices = set(devices + ["cpu"])
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -682,6 +682,7 @@ def load_state_dict(checkpoint_file, device_map=None):
             # if we only have one device we can load everything directly
             if len(devices) == 1:
                 return safe_load_file(checkpoint_file, device=devices[0])
+                
             # cpu device should always exist as fallback option
             devices = set(devices + ["cpu"])
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -678,14 +678,17 @@ def load_state_dict(checkpoint_file, device_map=None):
             return safe_load_file(checkpoint_file)
         else:
             devices = [device for device in device_map.values() if device not in ["disk"]]
+            # cpu device should always exist as fallback option
+            devices = set(devices + ["cpu"])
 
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -678,14 +678,17 @@ def load_state_dict(checkpoint_file, device_map=None):
             return safe_load_file(checkpoint_file)
         else:
             devices = [device for device in device_map.values() if device not in ["disk"]]
```
patrickvonplaten, line 680:
```suggestion
            devices = [device for device in device_map.values() if device not in ["disk"]]
            
            # if we only have one device we can load everything directly
            if len(devices) == 1:
                return safe_load_file(checkpoint_file, device=devices[0]) 
```

src/accelerate/utils/modeling.py
```diff
@@ -679,13 +679,20 @@ def load_state_dict(checkpoint_file, device_map=None):
...
 
+            # if we only have one device we can load everything directly
+            if len(devices) == 1:
+                return safe_load_file(checkpoint_file, device=devices[0])
```
patrickvonplaten, line 684:
```suggestion
                return safe_load_file(checkpoint_file, device=devices[0])
                
```

src/accelerate/utils/modeling.py
```diff
@@ -679,13 +679,21 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+
+            # cpu device should always exist as fallback option
+            devices = set(devices + ["cpu"])
```
sgugger, line 687:
```suggestion
            if "cpu" not in devices:
                devices.append("cpu")
```

src/accelerate/utils/modeling.py
```diff
@@ -679,13 +679,21 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+
             # For each device, get the weights that go there
             device_weights = {device: [] for device in devices}
             for module_name, device in device_map.items():
                 if device in devices:
                     device_weights[device].extend([k for k in weight_names if k.startswith(module_name)])
-            device_weights["cpu"].extend([k for k in weight_names if k not in device_weights])
 
+            # all weights that haven't defined a device sholud be loaded on CPU
```
sgugger, line 695:
```suggestion
            # all weights that haven't defined a device should be loaded on CPU
```

tests/test_modeling_utils.py
```diff
@@ -421,13 +421,15 @@ def test_load_state_dict(self):
...
 
-            loaded_state_dict = load_state_dict(checkpoint_file, device_map=device_map)
+                loaded_state_dict = load_state_dict(checkpoint_file, device_map=device_map)
 
-        self.assertEqual(loaded_state_dict["a"].device, torch.device("cpu"))
-        self.assertEqual(loaded_state_dict["b"].device, torch.device(0))
-        self.assertEqual(loaded_state_dict["c"].device, torch.device("cpu"))
+            for k, v in device_map.items():
+                v = v if v != "disk" else "cpu"
+                self.assertEqual(loaded_state_dict[k].device, torch.device(v))
```
sgugger, line 435: Let's use proper names here instead of k, v (like param, device).

sgugger (approved):
Thanks for fixing the bug! I left a couple of comments.

tests/test_modeling_utils.py
```diff
@@ -421,13 +421,15 @@ def test_load_state_dict(self):
...
+                checkpoint_file = os.path.join(tmp_dir, "model.safetensors")
+                save_file(state_dict, checkpoint_file, metadata={"format": "pt"})
 
-            loaded_state_dict = load_state_dict(checkpoint_file, device_map=device_map)
+                loaded_state_dict = load_state_dict(checkpoint_file, device_map=device_map)
 
-        self.assertEqual(loaded_state_dict["a"].device, torch.device("cpu"))
-        self.assertEqual(loaded_state_dict["b"].device, torch.device(0))
-        self.assertEqual(loaded_state_dict["c"].device, torch.device("cpu"))
+            for k, v in device_map.items():
```
patrickvonplaten, line 433:
```suggestion
            for param, device in device_map.items():
```

tests/test_modeling_utils.py
```diff
@@ -421,13 +421,15 @@ def test_load_state_dict(self):
...
+                v = v if v != "disk" else "cpu"
```
patrickvonplaten, line 434:
```suggestion
                device = device if device != "disk" else "cpu"
```

patrickvonplaten:
Merging as Sylvain has :heavy_check_mark:

PR #1123, opened 2023-02-27
Refactor `launch` for greater extensibility

Yard1:
This PR refactors `commands.launch` to break up the logic into smaller functions, allowing for the environment preparation logic and actual launch logic to be decoupled. Therefore, a developer can now prepare an environment dict (and command list, where applicable) without having to launch any processes.

There have been no changes to the logic - this is purely a refactor of an existing structure.

As discussed offline, this will unblock Ray Train and other distributed frameworks which launch their own jobs.

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -12,13 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import argparse
 import os
 import sys
+import warnings
+from ast import literal_eval
+from typing import Any, Dict, List, Tuple
 
 import torch
 
-from ..utils import is_torch_version
-from .dataclasses import DistributedType
+from ..commands.config.config_args import SageMakerConfig
+from ..commands.config.config_utils import DYNAMO_BACKENDS
+from ..utils import DynamoBackend, PrecisionType, is_torch_version
+from ..utils.constants import DEEPSPEED_MULTINODE_LAUNCHERS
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Thanks a lot for your PR! This looks great to me. Mind having a second look @muellerzr ?

Can you just solve the merge conflict @Yard1 ?

muellerzr (approved):
LG2M, great work! Once the merge conflict is fixed we can go ahead and bring it in 🤗

Yard1:
@muellerzr Thanks, should be good now!

src/accelerate/utils/launch.py
```diff
@@ -23,8 +23,9 @@
 
 from ..commands.config.config_args import SageMakerConfig
 from ..commands.config.config_utils import DYNAMO_BACKENDS
-from ..utils import DynamoBackend, PrecisionType, is_torch_version, merge_dicts
```
dbpprt: Ideally it should be imported from ..utils and not using the direct path to `.other`, following the same mechanism used for all other utils, isn't it?

Yard1: That results in a circular import, unfortunately

dbpprt: Got it, then LGTM.

LouisCastricato:
FYI @natolambert this was the PR that i had mentioned to you ages ago. It'll be useful if you guys plan to do sweeps in TRL.

PR #1128, opened 2023-02-28
Move dynamo.optimize to the end of model preparation

ymwangg:
Dynamo no longer supports adding amp wrapper after calling compile/optimize as described in https://github.com/pytorch/pytorch/issues/95641.


muellerzr:
CC @sgugger

sgugger (approved):
Nice catch, thanks for the fix! (We also need to move this to `torch.compile` somewhat soon-ish!)

PR #1132, opened 2023-03-01
Expose tee and role in Launch

muellerzr:
Adds `--tee` and `--role` to `launch`, once https://github.com/huggingface/accelerate/pull/1108 is closed a better method will be used that doesn't require us to redefine each arg like this (and it can be done dynamically). 

Requested by @stas00


sgugger (approved):
Thanks for adding those!

stas00 (approved):
tested - all works - thanks a lot, Zach!

PR #1134, opened 2023-03-01
Expand warning and grab all GPUs available by default

muellerzr:
Modifies the default for `launch` to grab all of the GPUs available if called without specifying anything. Will then also set `multi_gpu` to `True`.


sgugger (approved):
SGTM!

PR #1140, opened 2023-03-02
fix lr scheduler issue

### LINKED ISSUE #1139: lr schedulers aren't prepared when using 2.0 nightlies
### System Info

```Shell
- `Accelerate` version: 0.17.0.dev0
- Platform: Linux-5.4.0-125-generic-x86_64-with-glibc2.31
- Python version: 3.10.4
- Numpy version: 1.23.5
- PyTorch version (GPU?): 2.0.0.dev20230208 (True)
	Not found
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)

### Reproduction

Steps to reproduce:
1. Add following snippet post `accelerator.prepare` in the nlp_example.py
```
from accelerate.scheduler import AcceleratedScheduler
if not isinstance(lr_scheduler, AcceleratedScheduler):
        raise ValueError("Oh no! lr scheduler isn't prepared")
```
2. logs will have:
```
ValueError: Oh no! lr scheduler isn't prepared
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid: 386705
```

Reason: 
1. Because of PR #1017, `_LRScheduler` is being imported instead of  `LRScheduler` as both are present. However, lr schedulers now extend `LRScheduler`, `_LRScheduler` also extends `LRScheduler`. Because of this below line fails and lr scheduler doesn't get prepared:

https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L918

 

### Expected behavior

```Shell
lr scheduler is properly prepared
```

pacman100:
### What does this PR do?
1. Fixes #1139

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -102,7 +102,7 @@ if is_tpu_available(check_device=False):
 
 
 try:
-    from torch.optim.lr_scheduler import LRScheduler as LRScheduler
+    from torch.optim.lr_scheduler import LRScheduler
 except ImportError:
     from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -102,9 +102,9 @@ if is_tpu_available(check_device=False):
 
 
 try:
-    from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
-except ImportError:
     from torch.optim.lr_scheduler import LRScheduler as LRScheduler
+except ImportError:
+    from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 
 logger = get_logger(__name__)
 


src/accelerate/accelerator.py
```diff
@@ -102,9 +102,9 @@
 
 
 try:
-    from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
-except ImportError:
     from torch.optim.lr_scheduler import LRScheduler as LRScheduler
```
sgugger, line 105:
```suggestion
    from torch.optim.lr_scheduler import LRScheduler
```

sgugger (approved):
Thanks for the fix!

muellerzr (approved):
Thanks!

Issue #1119, opened 2023-02-26
TPU cannot be ran in Colab anymore


arkadiusz-czerwinski:
### System Info

```Shell
Current accelerate pytorch example in readme doesnt work.
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Go to https://github.com/huggingface/accelerate#launching-your-training-from-a-notebook
Run a notebook
Xla version wont match.
Upon updating first cell to:
```
  !pip install datasets transformers
  !pip install cloud-tpu-client==0.10 torch==1.13.0 https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch_xla-1.13-cp38-cp38-linux_x86_64.whl
  !pip install git+https://github.com/huggingface/accelerate
```
Following error is generated:
```
Exception in device=TPU:2: Cannot replicate if number of devices (1) is different from 8  File "/usr/local/lib/python3.8/dist-packages/torch_xla/distributed/xla_multiprocessing.py", line 331, in _mp_start_fn
```

It seems like an error regarding distirbution of data and models across different tpu cores, running the examples from terminal doesnt solve the issue, at least for me for current colab version, although this statement should be verified, as I might have done something wrong. 

Unfortunatelly, for python 3.8, which is default python for colab only this wheel is avaliable for pytorch_xla as per [link](https://github.com/pytorch/xla#wheel).

### Expected behavior
Pytorch xla version should be updated, i.e using following command, although unlike command below which is only exemplary, it shouldn't throw an error above. 
```Shell
!pip install cloud-tpu-client==0.10 torch==1.13.0 https://storage.googleapis.com/tpu-pytorch/wheels/colab/torch_xla-1.13-cp38-cp38-linux_x86_64.whl
```

sgugger:
Thanks for the report! Would you like to open a PR with your suggested change?

muellerzr:
This actually should probably be changed to utilize `install_xla` instead, I'll make a PR today for this. [source code](https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/torch_xla.py#L21-52)

arkadiusz-czerwinski:
[...]

PR #1141, opened 2023-03-02
Fix notebook_launcher

muellerzr:
The decorators in the Accelerator object still were initialing the Accelerator on import, leading to issues when using the `notebook_launcher` (as seen in https://github.com/huggingface/accelerate/issues/1119). This PR fixes https://github.com/huggingface/accelerate/issues/1119 by rewriting the decorators again to follow the `_inner` pattern. Note that this is only needed on the `Accelerator`, for the self-decorators. This isn't needed for the `PartialState` elsewhere. 

(And this problem is mostly apparent on `notebook_launcher`, which takes a fork of the current process)

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -511,10 +511,10 @@ class Accelerator:
                     "The `on_main_process` decorator must be called with a function on an instantiated `Accelerator` object."
                 )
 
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for the fix!

PR #1144, opened 2023-03-03
fix partial state

pacman100:
### What does this PR do?
1. FSDP and Megatron-LM integrations are broken since the introduction of `PartialState` #1055. In prepare_data_loader , state obj is created again via `state = AcceleratorState()` . Now, it causes everything to go haywire, `self.__dict__.update(PartialState._shared_state)`  reverts FSDP  in AcceleratorState to `MULTI_GPU`  present in PartialState. This PR fixes it.


muellerzr (approved):
Thanks for the catch, just to be sure is this why the nightlies are failing? (Aka they caught this?)

muellerzr:
Thanks, sorry for the breaking change 🙃 should've also been checking the slow tests with such a large PR before merging

sgugger (approved):
LGTM!

PR #1145, opened 2023-03-03
FSDP enhancements and fixes

pacman100:
### What does this PR do?
1. Adds support for multiple transformer class layer wrapping to be in sync with latest changes in Trainer FSDP support.
2. `limit_all_gathers` is supported from 1.13 torch version, fixing that to be backward compatible


sgugger (approved):
Thanks for the fixes!

PR #1146, opened 2023-03-03
Fixed typos in notebook

SamuelLarkin:
Looks like a typo where `eval_split == train_split`.  The intend was most likely a 80%/20% split.


sgugger (approved):
Thanks for the fix!

PR #1151, opened 2023-03-06
[Safetensors] Relax missing metadata constraint

patrickvonplaten:
Many diffusers checkpoint are sadly saved without `metadata["format"] = "pt"` which then creates the following issues: https://github.com/huggingface/diffusers/issues/2445

@sgugger could we maybe relax the constraint here a bit that if the metadata is missing we assume it's PT (also since accelerate is only made for PT)

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -674,7 +674,8 @@ def load_state_dict(checkpoint_file, device_map=None):
 
         if metadata is None:
             logger.warn(
-                f"The safetensors archive passed at {checkpoint_file} does not contain metadata. Make sure to save your model with the `save_pretrained` method. Defaulting to 'pt' metadata."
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -667,6 +671,13 @@ def load_state_dict(checkpoint_file, device_map=None):
         with safe_open(checkpoint_file, framework="pt") as f:
             metadata = f.metadata()
             weight_names = f.keys()
+
+        if metadata is None:
+            logger.warn(
+                f"The safetensors archive passed at {checkpoint_file} does not contain metadata. Make sure to save your model with the `save_pretrained` method. Defaulting to 'pt' metadata."
```
sgugger, line 677: Can you just split this to respect the 119 char limit? Thanks!

sgugger (approved):
LGTM, thanks!

Issue #1147, opened 2023-03-04
Incompatible with 8bit models


sabetAI:
### System Info

```Shell
If I load a model in 8bit, it has to be mapped automatically to gpus. However, when `accelerate` tries to load the model to several gpus, I get the following error.

`ValueError: `.to` is not supported for `8-bit` models. Please use the model as it is, since the model has already been set to the correct devices and casted to the correct `dtype`.`

How to fix this?
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1.. Load model in 8bit

```
model = GPTJForCausalLM.from_pretrained(
    hyperparameters["model_checkpoint"],
    revision="float16",
    load_in_8bit=True,
    device_map="auto",
)
```

2. Run with pytorch accelerate
```
import torch
import torch.nn.functional as F
from datasets import load_dataset
from accelerate import Accelerator

accelerator = Accelerator()
optimizer = torch.optim.Adam(model.parameters())

dataset = load_dataset('my_dataset')
data = torch.utils.data.DataLoader(dataset, shuffle=True)

model, optimizer, data = accelerator.prepare(model, optimizer, data)

model.train()
for epoch in range(10):
    for source, targets in data:
         optimizer.zero_grad()
         output = model(source)
         loss = F.cross_entropy(output, targets)
         accelerator.backward(loss)
         optimizer.step()
```

### Expected behavior

```Shell
accelerate loads model to gpus.
```

sgugger:
cc @younesbelkada Could you have the same test you did in the Trainer here in Accelerate?

younesbelkada:
Hi @sabetAI 
Thanks for the issue!
Actually the snippet won't work even if we fix the issue on `Accelerator` side because you cannot backpropagate the gradients through an 8bit model! 
If you are interested in fine-tuning a model that is loaded in 8bit, you should have a look at what is done on `peft` as you can train adapters on top of 8-bit models. Note that the script below works fine on the current `main` branch:

```python
import torch

from transformers import AutoModelForCausalLM
from accelerate import Accelerator

from peft import get_peft_model, prepare_model_for_int8_training, LoraConfig


model = AutoModelForCausalLM.from_pretrained(
[...]

PR #1155, opened 2023-03-06
[`Accelerator`] Fix issue with 8bit models

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/1147

In theory it is not possible to fine-tune 8-bit models except if you use adapters that can be used only is a `PeftModel` is used for training (I also need to test the snippet below with a `PeftModel` to make sure this is relevant or not)

But in some systems you can use the accelerator to load an 8bit model and use it out of the training scope 

I am not sure if we should support 8-bit models using `Accelerator`, but if so, I propose the following changes in this PR
Happy also to revert the tests / `bnb` dependency 

To reproduce: 
```python
import torch
import torch.nn.functional as F

    "EleutherAI/gpt-neo-125m",
    load_in_8bit=True,
    device_map="balanced",
)

accelerator = Accelerator()


```

cc @sgugger 

Ran all the slow tests and got errors on DeepSpeed and FSDP tests but not sure if the failing is related to my PR

Screenshot 2023-03-06 at 18 24 12

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1119,8 +1119,27 @@ class Accelerator:
         if device_placement is None:
             device_placement = self.device_placement and self.distributed_type != DistributedType.FSDP
         self._models.append(model)
-        if device_placement:
+        if getattr(model, "is_loaded_in_8bit", False):
+            model_devices = set(model.hf_device_map.values())
+            if len(model_devices) > 1:
+                raise ValueError(
+                    "You can't train a model that has been loaded in 8-bit precision on multiple devices."
+                )
+
+            current_device_index = list(model_devices)[0]
+            if torch.device(current_device_index) != self.device:
+                # if on the first device (GPU 0) we don't care
+                if (self.device.index is not None) or (current_device_index != 0):
+                    raise ValueError(
+                        "You can't train a model that has been loaded in 8-bit precision on a different device than the one "
+                        "you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device()}"
+                    )
+
+            if "cpu" in model_devices or "disk" in model_devices:
+                raise ValueError("You can't train a model that has been loaded in 8-bit precision on CPU or disk.")
+        elif device_placement:
             model = model.to(self.device)
+
         if self.distributed_type == DistributedType.MULTI_GPU:
             if any(p.requires_grad for p in model.parameters()):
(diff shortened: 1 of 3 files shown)


younesbelkada:
I think this PR is only necessary in case people want to design systems that uses 8-bit models in their training loop without backpropagating on the 8-bit model (for example RLHF), as using adapters works already out of the box right now

src/accelerate/accelerator.py
```diff
@@ -1119,8 +1119,27 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
...
+            if torch.device(current_device_index) != self.device:
+                # if on the first device (GPU 0) we don't care
+                if (self.device.index is not None) or (current_device_index != 0):
+                    raise ValueError(
+                        "You can't train a model that has been loaded in 8-bit precision on a different device than the one "
+                        "you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device()}"
+                    )
+
+            if "cpu" in model_devices or "disk" in model_devices:
+                raise ValueError("You can't train a model that has been loaded in 8-bit precision on CPU or disk.")
```
sgugger, line 1139:
```suggestion
                raise ValueError("You can't train a model that has been loaded in 8-bit precision with CPU or disk offload.")
```

sgugger (approved):
Thanks for the fix!

tests/test_accelerator.py
```diff
@@ -160,3 +162,57 @@ def load_config(models, input_dir):
...
+        with init_empty_weights():
+            model = AutoModelForCausalLM.from_pretrained(
+                "EleutherAI/gpt-neo-125m",
+            )
+            device_map = infer_auto_device_map(model)
+            device_map["lm_head"] = "cpu"
+
+        model = AutoModelForCausalLM.from_pretrained(
+            "EleutherAI/gpt-neo-125m", device_map=device_map, load_in_8bit=True, llm_int8_enable_fp32_cpu_offload=True
+        )
```
younesbelkada, line 190: @sgugger I think this will fail as the docker image is not using the `main` branch of `transformers` no?

younesbelkada: happy to skip it until the next release of `transformers`

sgugger: Yes we are not installing from source.

sgugger (approved):
Thanks again!

PR #1156, opened 2023-03-06
Load custom state to cpu

Guangxuan-Xiao:
The current implementation loads custom states to GPUs, leading to OOM. I add `map_location="cpu"` to the `torch.load` function, which is similar to the strategy in `load_accelerator_state`.


sgugger (approved):
Thanks for fixing this!

Issue #1050, opened 2023-02-09
A BUG? when performing gradient accumulation


jcyk:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-4.14.105-1-tlinux3-0013-x86_64-with-glibc2.2.5
- Python version: 3.8.12
- Numpy version: 1.21.4
- PyTorch version (GPU?): 1.10.0 (True)
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```python
step = 0
while True:
    for batch in training_dataloader:
        step += 1
        with accelerator.accumulate(model):
            inputs, targets = batch
            outputs = model(inputs)
            loss = loss_function(outputs, targets)
            accelerator.backward(loss)
            optimizer.step()
            scheduler.step()
            optimizer.zero_grad()
        if step % 100 == 0:
            for batch in eval_dataloader:
                inputs, targets = batch
                outputs = model(inputs)
                loss = loss_function(outputs, targets)
```

### Expected behavior

As seen, we shuffle from training and evaluating (This is common as we monitor its performance on dev set while training the model ).

but the scheduler's step will be ***wrong***

It seems the forward steps in evaluating also account for gradient accumulation.

jcyk:
I think I found the reason:

The loop over eval_dataloader make `self.gradient_state.end_of_dataloader` to be `True`
Then self.gradient_state.sync_gradients is always set to `True` according to 
https://github.com/huggingface/accelerate/blob/978dfc38ea91aae645d780d393d5bc0e33ac8306/src/accelerate/accelerator.py#L627


Do you feel this is a bug?

sgugger:
cc @muellerzr

jcyk:
I would like to share my current solution to this problem.

```python
step = 0
while True:
    for batch in training_dataloader:
        step += 1
        with accelerator.accumulate(model):
            inputs, targets = batch
            outputs = model(inputs)
            loss = loss_function(outputs, targets)
            accelerator.backward(loss)
            optimizer.step()
            scheduler.step()
            optimizer.zero_grad()
        if step % 100 == 0:
            ######## we save the state of the outer dataloader
[...]

Issue #960, opened 2023-01-09
`gather_for_metrics` and `GradientState` don't work with arbitrary evaluation intervals


ernestchu:
I try to run evaluation on the evaluation_dataloader (entire eval epoch) every given steps (instead of running after an entire train epoch). After evaluation,  the `GradientState` used by `gather_for_metrics` seems to stuck as-is in the last epoch of evaluation, which is
```
Sync Gradients: True                                                                                  
At end of current dataloader: True                                                                     
Extra samples added: 20
```
and it cause `gather_for_metrics` drop samples on train_loader for the following training until it reaches the end of the train epoch.

I noticed that the `gradient_state` in `Accelerator` main instance and both `DataLoaderShard` are synced, in a way I can't find in the accelerate source code.

Can someone elaborate what the design philosophy of `GradientState` and how it syncs between the accelerator and dataloaders. Additionally, if it is possible to do evaluation in the middle of the train epoch like I mentioned above. Thanks!

ernestchu:
After some investigations, I know the `GradientState` is synced by the **singleton class** design. My current workaround is manually setting `accelerator.gradient_state.end_of_dataloader` to `False` after the evaluation, so the following training would not incorrectly drop samples.

sgugger:
cc @muellerzr

muellerzr:
@ernestchu `GradientState` is a singleton designed to work with a single dataloader that is currently being iterated upon, which is why you're experiencing this trouble. One thing I could think of is setting an api such as `pause` and `resume` which will "turn off" `GradientState` and/or reset it. So essentially the API would be:

```python
dl1, dl2 = accelerator.prepare(dl1, dl2)
for i,batch in enumerate(dl1):
    ...
    if i == 5:
        accelerator.gradient_state.pause()
        for batch in dl2:
            ....
        accelerator.gradient_state.resume()
    ...
```

Does this seem reasonable to you both @ernestchu @sgugger

muellerzr:
Should be closed with https://github.com/huggingface/accelerate/pull/1162, doesn't require any modification in the code

PR #1162, opened 2023-03-07
Let GradientState know active dataloaders and reset the remainder

muellerzr:
Adjusts the `GradientState` so it can know:
- What dataloader is entered/exited on `__iter__`
- The total dataloader list that we are currently iterating over
- Changes the logic for setting the end of the dataloader to exist in `GradientState._add_dataloader` and `GradientState._remove_dataloader`

Solves https://github.com/huggingface/accelerate/issues/960 and https://github.com/huggingface/accelerate/issues/1050


sgugger (approved):
LGTM, thanks!

Ethan-yt:
This doesn't seem to be enough to solve the problem.https://github.com/huggingface/accelerate/issues/1050
When the second dataloader finishes, it removes it and sets the state to "True", which affects the external state.

Issue #1163, opened 2023-03-08
save_state deletes wrong checkpoint when total_limit is set


Ttl:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-5.19.0-32-generic-x86_64-with-glibc2.35
- Python version: 3.10.6
- Numpy version: 1.23.3
- PyTorch version (GPU?): 1.12.1+cu116 (True)
- `Accelerate` default config:
	- compute_environment: LOCAL_MACHINE
	- distributed_type: NO
	- mixed_precision: fp16
	- use_cpu: False
	- dynamo_backend: NO
	- num_processes: 1
	- machine_rank: 0
	- num_machines: 1
	- gpu_ids: all
	- rdzv_backend: static
	- same_network: True
	- main_training_function: main
	- deepspeed_config: {}
	- fsdp_config: {}
	- megatron_lm_config: {}
	- downcast_bf16: no
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```python
from accelerate import Accelerator
from accelerate.utils import ProjectConfiguration
import torch

config = ProjectConfiguration(total_limit=3, automatic_checkpoint_naming=True)
accelerator = Accelerator(project_dir='./total_limit_test', project_config=config)

model = torch.nn.Linear(1, 1)
optimizer = torch.optim.Adam(params=model.parameters())
dataloader = torch.utils.data.DataLoader(None)
scheduler = torch.optim.lr_scheduler.ConstantLR(optimizer)

model, optimizer, dataloader, scheduler = accelerator.prepare(
     model, optimizer, dataloader, scheduler
)

for _ in range(20):
    accelerator.save_state()
```

### Expected behavior

[...]

PR #1165, opened 2023-03-08
Fix what files get deleted through `total_limit`

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/1163 by adding in a more explicit regex to help with sorting.


sgugger (commented):
Do you think there is a way to add a regression test?

sgugger (approved):
Thanks!

PR #1166, opened 2023-03-08
Remove outdated command directions and use in tests

muellerzr:
Since `torch.distributed.launch` is deprecated, finds all cases of `launch` and replaces with `run`/`torchrun`

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Want to learn more? Check out the [documentation](https://huggingface.co/docs/ac
 
 ## Launching script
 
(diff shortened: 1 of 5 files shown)


sgugger (approved):
Thanks! In the examples we should mention that for older PyTorch versions, user should use torch.distributed.launch no?

muellerzr:
@sgugger changed it to be:
````
    * With PyTorch launcher only (`torch.distributed.launch` can be used in older versions of PyTorch)
        ```bash
        python -m torchrun --nproc_per_node 2 \
            --use_env \
            --node_rank 0 \
            --master_addr master_node_ip_address \
            ./nlp_example.py  # On the first server
       ```
````
in the respective areas in the README.md

sgugger (approved):

PR #1168, opened 2023-03-09
Fix backward compatibility in configs wrt dynamo backend

sgugger:
Currently if a user tries to load a config saved with 0.16.0, they will get
```
TypeError: __init__() got an unexpected keyword argument 'dynamo_backend'
```

This PR ensures backward compatibility.


muellerzr (approved):
Thanks!

pacman100 (approved):
Thank you!  LGTM 🤗

PR #1169, opened 2023-03-09
Fix accelerate test with new config_file errors

sgugger:
`accelerate test` errors out on main, this fixes it.


muellerzr (approved):
LG2M! Will investigate why this isn't being ran as part of the CI

PR #1172, opened 2023-03-10
[`Accelerator`] We should not call `to` on modules that wraps `accelerate` loaded models

younesbelkada:
# What does this PR do?

In some training setups such as in `trl`: https://github.com/lvwerra/trl sometimes users create a new module that wraps modules that are loaded with `accelerate`. Check for example [this class](https://github.com/lvwerra/trl/blob/ddb6df367da4a7374287f35ae37d092f7c32a815/trl/models/modeling_value_head.py#L57)
I propose a simple fix in the accelerator to check if there is any immediate child module that has the `hf_device_map` attribute

cc @sgugger 

I can confirm all slow tests pass except the similar ones as in https://github.com/huggingface/accelerate/pull/1155 - which I believe are related to the way I am creating my env

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1137,6 +1137,14 @@ class Accelerator:
             device_placement = self.device_placement and self.distributed_type != DistributedType.FSDP
         self._models.append(model)
         # We check only for models loaded with `accelerate`
+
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Works for me, thanks!

Issue #1174, opened 2023-03-10
Accelerate.prepare fails on optimizer prepare when forcing CPU


muellerzr:
### System Info

```Shell
Accelerate v0.17.0
torch 1.13.1
python 3.10
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```python
import torch
from accelerate import Accelerator

model = torch.nn.Linear(10, 10)
sgd = torch.optim.SGD(model.parameters(), lr=0.01)

accelerator = Accelerator(cpu=True)

accelerator.prepare(sgd)
# this works: accelerator.prepare(model)

# this also fails
# accelerator = Accelerator(device_placement=False, cpu=True)
# accelerator.prepare(sgd)
```

Error:
```
AssertionError: The current device and desired device are not the same. If the `PartialState` was generated before the `Accelerator` has been instantiated, ensure the `cpu` flag is the same for both. In this case, the `PartialState` has True and the desired device is False. Please use `cpu=True`.
```

### Expected behavior

```Shell
Should pass
```

PR #1175, opened 2023-03-10
Fix CPU error always being raised

muellerzr:
Turns out I've discovered a neat behavior in the library, that was "broken" with the State rewrite. With how `Accelerate` will work, if the user defines `cpu=False` initially then goes `cpu=True`, then a value error will be raised and the state will be changed. *But* if `cpu=True` is passed first then `cpu=False`, nothing will change and no error will be raised. The reason for this is simply because its the default and its a way for `State` to not override the first choice, so we assume CPU all the time. 

Also brings back the `ACCELERATE_USE_CPU` env flag, as this was removed in the `State` refactor. 

This brings back this behavior and adds a test to the `AcceleratorTester` class. 

Closes https://github.com/huggingface/accelerate/issues/1174

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -78,13 +78,6 @@ class PartialState:
 
     def __init__(self, cpu: bool = False, **kwargs):
         self.__dict__ = self._shared_state
-        # Raise an error if the user tries to reinitialize on a different device setup in the same launch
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for fixing!

PR #1176, opened 2023-03-10
Run accelerate_test in cli

muellerzr:
`accelerate test` was not part of the test suites, so it was never being ran. This PR adds it to `test_cli`


sgugger (approved):
Nice, thanks!

PR #1177, opened 2023-03-10
Add repr to AlignHook for easier debugging.

sgugger:
This PR adds a repr to AlignHook, which makes life easier for debugging issues with big model inference.


muellerzr (approved):
Looks useful! Thanks!

PR #1179, opened 2023-03-10
Support special mapping of dtypes when preparing device map

sgugger:
This PR will go along with [PR to come in Transformers] that fixes big model inference for T5 models. T5 models now have some weights that are kept in float32, which interferes with the computation done by `infer_auto_device_map`. On my setup, it gets me OOM when trying to load any big T5 model.

To fix that, on the Accelerate side we now accept a new argument `special_dtypes` which maps weight names to dtypes to use (and for the others we use the `dtype` provided).


muellerzr (approved):
Thanks!

younesbelkada (approved):
Thanks a mile for this!

PR #1181, opened 2023-03-10
Add documentation around FSDP state dict save behavior

VikParuchuri:
By default, `accelerator.get_state_dict` will call the underlying `model.state_dict` implementation.  With FSDP, this gathers all of the state onto the rank 0 device.  If the parameters don't fit, this can cause CUDA OOM errors.

This PR adds documentation around how to use the PyTorch state_dict_type context manager to configure the behavior of the state dict.

### DIFF
--- a/docs/source/usage_guides/fsdp.mdx
+++ b/docs/source/usage_guides/fsdp.mdx
@@ -88,6 +88,22 @@ Currently, `Accelerate` supports the following config through the CLI:
 )
 ```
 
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
Thank you @VikParuchuri for improving the FSDP integration docs, LGTM!

sgugger (approved):
Thanks for documenting this!

Issue #1170, opened 2023-03-09
Feature Request: have `AcceleratedScheduler` take into account the gradient accumulation steps automatically


muellerzr:
Based on the code [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/scheduler.py#L54-L62), during gradient accumulation the scheduler is stepped as though gradient accumulation doesn't exist, and the scheduler length should be adjusted based on the gradient accumulation'd fraction

Issue #1160, opened 2023-03-07
Diffucult to understand the behavior of `lr_scheduler` when using `gradient_accumulation`


kingnobro:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-5.4.0-91-generic-x86_64-with-glibc2.31
- Python version: 3.9.15
- Numpy version: 1.23.5
- PyTorch version (GPU?): 1.12.1 (True)
- `Accelerate` default config:
        Not found
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Add the following codes after this line: [link](https://github.com/huggingface/accelerate/blob/1bfde6b963f449a2a5c52762fe429308e6d4a9f0/examples/by_feature/gradient_accumulation.py#L184)

```python
if accelerator.is_local_main_process:
    accelerator.print('step', step, 'lr', lr_scheduler.get_last_lr()[0])
```

I run the `gradient_accumulation.py` using the following command provided by document.
```shell
accelerate launch ./gradient_accumulation.py --gradient_accumulation_steps 5
```

### Expected behavior

```Shell
Expected: The learning rate is 0 at the end of training.

However, the result is not 0.

Epoch 3
step 222 lr 1.8779661016949152e-05
step 223 lr 1.8779661016949152e-05
step 224 lr 1.8745762711864407e-05
step 225 lr 1.8745762711864407e-05
step 226 lr 1.8745762711864407e-05
step 227 lr 1.8745762711864407e-05
step 228 lr 1.8745762711864407e-05
step 229 lr 1.8711864406779663e-05  # the last line of log
```

muellerzr:
For gradient accumulation, the number of actual steps need to be accounted for. (It does not need to be usually for multi-GPU etc). So for example, you would make your scheduler like so:

```python
    lr_scheduler = get_linear_schedule_with_warmup(
        optimizer=optimizer,
        num_warmup_steps=0,
        num_training_steps=(len(train_dataloader) * num_epochs) / gradient_accumulation_steps,
    )
```

PR #1187, opened 2023-03-13
Make the Scheduler adjust the steps taken relative to the gradient accumulation steps

muellerzr:
# Let the `AcceleratedScheduler` handle gradient accumulation steppage

## What does this add?

This PR adjusts the logic in the `AcceleratedScheduler` to take into account gradient accumulation steps. 

## Who is it for?

Closes https://github.com/huggingface/accelerate/issues/1170
Closes https://github.com/huggingface/accelerate/issues/1160

## Why is it needed?

Currently a behavior does not exist that will automatically "cut" the LR scheduler of a user if they pass in `gradient_accumulation_steps`, so unless they are careful and adjust their LR scheduler beforehand, they're not actually stepping properly with the lr scheduler.

### User-facing:

None, explicitly

### Internal structure:

`AcceleratedScheduler`'s `step` function now will run `n*num_processes` where `n==gradient_accumulation_steps` to account for the difference. 

To test performance, I ran the equivalent training of `gradient_accumulation_steps==2` on a bs of 16, and a regular batch size of 32, with negligible performance differences (accuracy += 0.5%, F1 of 0.3). Likely due to batch norm layers

## Usage Example:

When building the `Accelerator`, pass in `adjust_scheduler_to_accumulation` (default `False`) to enable this behavior:

```python
    accelerator = Accelerator(
        cpu=args.cpu, 
        mixed_precision=args.mixed_precision, 
        gradient_accumulation_steps=2,
        adjust_scheduler_to_accumulation=True,
    )
```

### DIFF
--- a/docs/source/usage_guides/gradient_accumulation.mdx
+++ b/docs/source/usage_guides/gradient_accumulation.mdx
@@ -111,6 +111,11 @@ You can remove all the special checks for the step number and the loss adjustmen
 
 As you can see the [`Accelerator`] is able to keep track of the batch number you are on and it will automatically know whether to step through the prepared optimizer and how to adjust the loss. 
 
+<Tip>
+  Typically with gradient accumulation, you would need to adjust the number of steps to reflect the change in total batches you are 
+  training on. 🤗 Accelerate will automatically do this for you, so long as you pass `adjust_scheduler_to_accumulation` to the [`Accelerator`] object's `__init__`.
+</Tip>
+
 ## The finished code
 
 Below is the finished implementation for performing gradient accumulation with 🤗 Accelerate
--- a/docs/source/usage_guides/gradient_accumulation.mdx
+++ b/docs/source/usage_guides/gradient_accumulation.mdx
@@ -112,8 +112,10 @@ You can remove all the special checks for the step number and the loss adjustmen
 As you can see the [`Accelerator`] is able to keep track of the batch number you are on and it will automatically know whether to step through the prepared optimizer and how to adjust the loss. 
 
 <Tip>
-  Typically with gradient accumulation, you would need to adjust the number of steps to reflect the change in total batches you are 
-  training on. 🤗 Accelerate will automatically do this for you, so long as you pass `adjust_scheduler_to_accumulation` to the [`Accelerator`] object's `__init__`.
+
+Typically with gradient accumulation, you would need to adjust the number of steps to reflect the change in total batches you are 
+training on. 🤗 Accelerate will automatically do this for you, so long as you pass `adjust_scheduler_to_accumulation` to the [`Accelerator`] object's `__init__`.
+
 </Tip>
 
 ## The finished code
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -27,6 +27,8 @@ from typing import Any, Callable, List, Optional, Union
 import torch
 import torch.utils.hooks as hooks
 
+from accelerate.utils.environment import parse_choice_from_env
+
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
 from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_first_batches
 from .logging import get_logger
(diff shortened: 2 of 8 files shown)


docs/source/usage_guides/gradient_accumulation.mdx
```diff
@@ -111,6 +111,11 @@ You can remove all the special checks for the step number and the loss adjustmen
 
 As you can see the [`Accelerator`] is able to keep track of the batch number you are on and it will automatically know whether to step through the prepared optimizer and how to adjust the loss. 
 
+<Tip>
+  Typically with gradient accumulation, you would need to adjust the number of steps to reflect the change in total batches you are 
+  training on. 🤗 Accelerate will automatically do this for you, so long as you pass `adjust_scheduler_to_accumulation` to the [`Accelerator`] object's `__init__`.
```
sgugger, line 116:
```suggestion

Typically with gradient accumulation, you would need to adjust the number of steps to reflect the change in total batches you are 
training on. 🤗 Accelerate will automatically do this for you, so long as you pass `adjust_scheduler_to_accumulation` to the [`Accelerator`] object's `__init__`.

```

src/accelerate/accelerator.py
```diff
@@ -143,6 +143,10 @@ class Accelerator:
             The number of steps that should pass before gradients are accumulated. A number > 1 should be combined with
             `Accelerator.accumulate`. If not passed, will default to the value in the environment variable
             `ACCELERATE_GRADIENT_ACCUMULATION_STEPS`.
+        adjust_scheduler_to_accumulation (`bool`, *optional*, defaults to `False`):
+            Whether or not the scheduler should be adjusted to the number of gradient accumulation steps. If `True`,
+            the scheduler's length will be reduced by `gradient_accumulation_steps`. If the scheduler already accounts
+            for the number of gradient accumulation steps, this should be set to `False`.
```
sgugger, line 149: We're starting to have too many arguments to `Accelerator`. This one should be documented waaaay below as it's not as important, and we should think of grouping it with the gradient_accumulation_steps in a `GradientAccumulationPlugin` maybe.

src/accelerate/scheduler.py
```diff
@@ -42,10 +42,24 @@ class AcceleratedScheduler:
             Whether or not the dataloaders split one batch across the different processes (so batch size is the same
             regardless of the number of processes) or create batches on each process (so batch size is the original
             batch size multiplied by the number of processes).
+        adjust_scheduler_to_accumulation (`bool`, *optional*, defaults to `False`):
```
sgugger, line 45:
```suggestion
        adjust_to_accumulation (`bool`, *optional*, defaults to `False`):
```
We are in a scheduler.

sgugger (commented):
Thanks for working on this. Just not sure about the API (adding yet another argument to the Accelerator) for this.

muellerzr:
@sgugger good for review, a few notes:

- You'll notice that in the test for grad_sync changes, for it to pass (and for everywhere to pass in general), having accelerator use the new stepped scheduler to `True` is used, and on *every* script we have this behavior of what should exist is what we've advertised using. As such, I think it would be good to change the default to `True` not `False` here, as users are expecting one thing rather than the other
- Should we keep `gradient_accumulation_steps` as an arg in `Accelerator`? I'm leaning towards yes *if* we change the default to `True` for reasons above, though we can also just keep it there for usability. (In my eyes, users would likely not *need* to adjust the other param, hence we only need to expose `gradient_accumulation_steps` there

src/accelerate/utils/dataclasses.py
```diff
@@ -380,6 +380,25 @@ def __post_init__(self):
...
 
+@dataclass
+class GradientAccumulationPlugin(KwargsHandler):
+    """
+    A plugin to configure gradient accumulation behavior.
+    """
+
+    num_steps: int = field(default=None, metadata={"help": "The number of steps to accumulate gradients for."})
+    adjust_scheduler: bool = field(
+        default=False,
```
sgugger, line 391: This should be True by default I think.

sgugger (approved):
Thanks for iterating! I agree that the default should be `True` for `adjust_scheduler`. The gradient accumulation API is quite new anyway, so it's okay if it's adjusted a bit like this.

sgugger (approved):

PR #1188, opened 2023-03-13
Only convert linear layers with weights multiple of 16

sgugger:
In the FP8 integration, we don't want to convert linear layers that have weights that are not multiple of 16 (for instance the classifier head) as they won't work with transformer_engine.


muellerzr (approved):
Thanks! Makes sense

PR #1189, opened 2023-03-13
Set drop last to ensure modulo16 restriction for fp8

ksivaman:


examples/nlp_example.py
```diff
@@ -98,10 +98,14 @@ def collate_fn(examples):
...
-        tokenized_datasets["train"], shuffle=True, collate_fn=collate_fn, batch_size=batch_size
+        tokenized_datasets["train"], shuffle=True, collate_fn=collate_fn, batch_size=batch_size, drop_last=True
     )
     eval_dataloader = DataLoader(
-        tokenized_datasets["validation"], shuffle=False, collate_fn=collate_fn, batch_size=EVAL_BATCH_SIZE
+        tokenized_datasets["validation"],
+        shuffle=False,
+        collate_fn=collate_fn,
+        batch_size=EVAL_BATCH_SIZE,
+        drop_last=True,
```
sgugger, line 108: The one for the training set is not a problem, but this one will give a slightly wrong result, so let's only set it when mixed_precision is FP8.

sgugger (commented):
Thanks for the fixes! Left one comment.

sgugger (approved):
Thanks again!

Issue #1195, opened 2023-03-15
accelerate launch not working with trainer 


pacman100:
### System Info

```Shell
- `Accelerate` version: 0.18.0.dev0
- Platform: Linux-5.4.0-125-generic-x86_64-with-glibc2.31
- Python version: 3.10.4
- Numpy version: 1.24.1
- PyTorch version (GPU?): 2.0.0.dev20230301+cu117 (True)
- `Accelerate` default config:
	Not found
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Steps:

1. run following command:
```
accelerate launch --multi_gpu --num_processes 4 --main_process_port 9991 run_clm.py --model_name_or_path "facebook/opt-1.3b" --per_device_train_batch_size 1 --output_dir "/workspace/workfile/Downloads/output" --overwrite_output_dir --fp16 --do_train --max_train_samples 500 --num_train_epochs 1 --dataset_name "wikitext" --dataset_config "wikitext-2-v1" --fsdp "full_shard offload auto_wrap" --fsdp_transformer_layer_cls_to_wrap "OPTDecoderLayer" --optim "adamw_torch" --block_size 1024
```

2. error:
```
usage: accelerate <command> [<args>] launch [-h] [--config_file CONFIG_FILE] [--quiet] [--cpu]
                                            [--mps] [--multi_gpu] [--tpu]
                                            [--mixed_precision {no,fp16,bf16}]
                                            [--num_processes NUM_PROCESSES]
                                            [--num_machines NUM_MACHINES]
                                            [--num_cpu_threads_per_process NUM_CPU_THREADS_PER_PROCESS]
                                            [--dynamo_backend {no,eager,aot_eager,inductor,nvfuser,aot_nvfuser,aot_cudagraphs,ofi,fx2trt,onnxrt,ipex}]
                                            [--dynamo_mode {default,reduce-overhead,max-autotune}]
                                            [--dynamo_use_fullgraph] [--dynamo_use_dynamic]
                                            [--use_deepspeed] [--use_fsdp] [--use_megatron_lm]
                                            [--gpu_ids GPU_IDS] [--same_network]
                                            [--machine_rank MACHINE_RANK]
[...]

PR #1198, opened 2023-03-15
Fixup passing overlapping args to the script

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1195

TL;DR argparse has a behavior where by default if the passed args to the script align partially to the args in the parser, it will raise the error described in the issue


pacman100 (approved):
Thank you! 🤗

Issue #1196, opened 2023-03-15
AttributeError: 'ClusterConfig' object has no attribute 'tpu_cluster' on 


Beomi:
### System Info

```Shell
- Accelerate 0.17.1
- TPU v4-8 (TPUVM)
- tpu-vm-v4-pt-1.13

Sys env:

# Tpu v4 PyTorch XLA .env
export XRT_TPU_CONFIG="localservice;0;localhost:51011"

# Set the environment variable to visible devices*
export TPU_NUM_DEVICES=4

# Allow LIBTPU LOAD by multiple processes
export ALLOW_MULTIPLE_LIBTPU_LOAD=1
```

`accelerate env`:

```
- `Accelerate` version: 0.17.1
- Platform: Linux-5.13.0-1023-gcp-x86_64-with-glibc2.17
- Python version: 3.8.16
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: TPU
        - mixed_precision: fp16
        - use_cpu: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Accelerate Config:

```yaml
# default_config.yaml
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: TPU
downcast_bf16: 'no'
dynamo_config: {}
fsdp_config: {}
machine_rank: 0
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'no'
num_machines: 1
num_processes: 4
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

Train Command:

```bash
accelerate launch run_clm_no_trainer.py \
[...]

PR #1202, opened 2023-03-16
Fix typo in TPU config

muellerzr:
Should solve https://github.com/huggingface/accelerate/issues/1196 (will merge once user has confirmed or on Monday, whichever is sooner)


sgugger (approved):
Thanks for the fix!

Beomi:
Seems working!

PR #1208, opened 2023-03-18
ds offload optim fix to use CPUAdam

pacman100:
### What does this PR do?
1. Use `DeepSpeedCPUAdam` when offloading optimizer to CPU/NVME


sgugger (approved):
Thanks for the fix!

PR #1211, opened 2023-03-19
Fix example in accumulate method documentation

VikParuchuri:
The accumulate method has an example of how to do gradient accumulation that has the context manager and the for loop in the wrong order.  The _do_sync method is called only when the context manager is entered, so it needs to go inside the loop.

(I was wondering why my model wasn't converging well!)


muellerzr:
Thanks for catching this!

Issue #1217, opened 2023-03-20
When using accelerate launch, `num_machines` is not loaded from `config_file`


neumyor:
### System Info

```Shell
- `Accelerate` version: 0.17.1
- Platform: Linux-5.4.0-48-generic-x86_64-with-glibc2.17
- Python version: 3.8.10
- Numpy version: 1.22.4
- PyTorch version (GPU?): 1.13.0+cu117 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

To reproduce the bug, 
1. cd `/examples/text_to_image`
2. set up `config.yaml` as 
```yaml
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: MULTI_GPU
downcast_bf16: 'no'
dynamo_config: {}
fsdp_config: {}
gpu_ids: o
machine_rank: 0
main_process_ip: 172.17.0.5
main_process_port: 10000
main_training_function: main
megatron_lm_config: {}
mixed_precision: 'no'
num_machines: 4
num_processes: 4
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```
[...]

PR #1218, opened 2023-03-20
Fix bug in loading launch config

neumyor:
associated with issue #1217 
fix bug in loading custom config file in `accelerate launch` by adding 

```python
if (args.multi_gpu and args.num_machines is None):
    args.num_machines = defaults.num_machines
```
in `/command/launch.py` line 805

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -211,10 +211,7 @@ def launch_command_parser(subparsers=None):
         help="Whether to use full graph mode for dynamo or it is ok to break model into several subgraphs",
     )
     resource_args.add_argument(
-        "--dynamo_use_dynamic",
-        default=False,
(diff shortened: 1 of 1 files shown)


muellerzr:
@neumyor thanks for the PR, to fix the quality checks please do the following:

```python
cd accelerate
pip install -e .[quality]
make style; make quality
```
This should fix that error

neumyor:
> @neumyor thanks for the PR, to fix the quality checks please do the following:
> 
> ```python
> cd accelerate
> pip install -e .[quality]
> make style; make quality
> ```
> 
> This should fix that error

Thanks for your advice, I think it's my mistake to use `black` instead of `make style`. Now the code should be clean.

muellerzr (approved):
Thanks! This fix looks good to me 🤗

Issue #1215, opened 2023-03-20
AttributeError: 'NoneType' object has no attribute 'num_steps'


cahya-wirawan:
### System Info

```Shell
- `Accelerate` version: 0.18.0.dev0
- Platform: Linux-5.13.0-1027-gcp-x86_64-with-glibc2.31
- Python version: 3.9.16
- Numpy version: 1.24.2
- PyTorch version (GPU?): 2.0.0+cu117 (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: TPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {}
```

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction
```
>>> from accelerate import Accelerator
>>> accelerator = Accelerator()
WARNING:root:XRT configuration not detected. Defaulting to preview PJRT runtime. To silence this warning and continue using PJRT, explicitly set PJRT_DEVICE to a supported device or configure XRT. To disable default device selection, set PJRT_SELECT_DEFAULT_DEVICE=0
WARNING:root:For more information about the status of PJRT, see https://github.com/pytorch/xla/blob/master/docs/pjrt.md
WARNING:root:Defaulting to PJRT_DEVICE=CPU
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ <stdin>:1 in <module>                                                                            │
│                                                                                                  │
│ /mnt/data-1/home/cahya/Work/miniconda3/envs/lora/lib/python3.9/site-packages/accelerate/accelera │
│ tor.py:358 in __init__                                                                           │
│                                                                                                  │
│    355 │   │   )                                                                                 │
│    356 │   │                                                                                     │
│    357 │   │   if self.state.distributed_type == DistributedType.TPU:                            │
[...]

PR #1219, opened 2023-03-20
Move when the GradientState test is performed so that it is not None

muellerzr:
Closes https://github.com/huggingface/accelerate/issues/1215


sgugger (approved):
Thanks for the fix!

Issue #1210, opened 2023-03-18
Detected 1 oom-kill event(s) in StepId=125757.batch. Some of your processes may have been killed by the cgroup out-of-memory handler.


QishengL:
### System Info

```Shell
- `Accelerate` version: 0.15.0
- Platform: Linux-5.4.199-ql-generic-12.0-18-x86_64-with-glibc2.31
- Python version: 3.10.4
- Numpy version: 1.23.1
- PyTorch version (GPU?): 1.12.1 (False)
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

The error: Detected 1 oom-kill event(s) in StepId=125757.batch. Some of your processes may have been killed by the cgroup out-of-memory handler.
Try to load optimizer and models but got CPU OOM with code below
optimizer.load_state_dict(torch.load("optim.pt",map_location='cpu'))
of
accelerator.load_state("my_checkpoint")

If I use the code below then everything is fine.
optimizer.load_state_dict(torch.load("optim.pt",map_location=accelerator.device))

### Expected behavior

```Shell
Is it possible to use accelerator.load_state("my_checkpoint") and load the state to accelerator.device directly?
```

sgugger:
I think this comes from the load of the optimizer state on multiple processes at once. @muellerzr we should probably load the optimizer on the device when in multi-GPU setting with num_processes>1

muellerzr:
Should be fixed with #1220, can you try it out via `pip install git+https://github.com/huggingface/accelerate` and let us know? :)

PR #1220, opened 2023-03-20
Set the state device dependant to Accelerator on multigpu

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1210 by setting the optimizer state to `accelerator.device` when `num_processes > 1` and calling `load_state`.

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -110,7 +110,14 @@ def save_accelerator_state(
 
 
 def load_accelerator_state(
-    input_dir, models, optimizers, schedulers, process_index, scaler=None, **load_model_func_kwargs
+    input_dir,
+    models,
+    optimizers,
+    schedulers,
+    process_index,
+    scaler=None,
+    optimizer_map_location="cpu",
+    **load_model_func_kwargs,
 ):
     """
     Loads states of the models, optimizers, scaler, and RNG generators from a given directory.
@@ -128,6 +135,8 @@ def load_accelerator_state(
             The current process index in the Accelerator state
         scaler (`torch.cuda.amp.GradScaler`, *optional*):
             An optional *GradScaler* instance to load
+        optimizer_map_location (`torch.device`, *optional*):
+            What device to load the optimizer state onto. By default uses "cpu".
         load_model_func_kwargs (`dict`, *optional*):
             Additional arguments that can be passed to the model's `load_state_dict` method.
     """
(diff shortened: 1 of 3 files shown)


src/accelerate/checkpointing.py
```diff
@@ -128,6 +135,8 @@ def load_accelerator_state(
             The current process index in the Accelerator state
         scaler (`torch.cuda.amp.GradScaler`, *optional*):
             An optional *GradScaler* instance to load
+        optimizer_map_location (`torch.device`, *optional*):
+            What device to load the optimizer state onto. By default uses "cpu".
```
sgugger, line 139: Should be None by default then either be cpu or "on_device" (which gets interpreted as accelerator.device after).

sgugger (commented):
Mmm, if we add a new flag to the load method, we should make it default smartly. Also not sure if that new flag needs to be a string since it has two states (apart from unset): CPU or device. So maybe an optional bool would suffice?

src/accelerate/checkpointing.py
```diff
@@ -139,10 +149,18 @@ def load_accelerator_state(
     logger.info("All model weights loaded successfully")
 
     # Optimizer states
+    if optimizer_map_location not in [None, "cpu", "on_device"]:
+        raise TypeError(
+            "Unsupported optimizer map location passed, please choose one of `None`, `cpu`, or `on_device`"
+        )
+    if optimizer_map_location is None:
+        optimizer_map_location = "cpu"
+    elif optimizer_map_location == "on_device":
```
sgugger, line 158: I still think we should default to `"on_device"` for distributed trainings on GPU, otherwise we require num_processes times the optimizer state available in CPU RAM.

src/accelerate/checkpointing.py
```diff
@@ -128,21 +136,34 @@ def load_accelerator_state(
...
+    elif map_location == "on_device":
+        map_location = PartialState().device
     # Model states
     for i, model in enumerate(models):
         weights_name = f"{MODEL_NAME}.bin" if i == 0 else f"{MODEL_NAME}_{i}.bin"
         input_model_file = os.path.join(input_dir, weights_name)
-        models[i].load_state_dict(torch.load(input_model_file, map_location="cpu"), **load_model_func_kwargs)
+        if map_location != "cpu":
+            models[i].to(map_location)
+        models[i].load_state_dict(torch.load(input_model_file, map_location=map_location), **load_model_func_kwargs)
```
muellerzr, line 158: PyTorch will load the optimizer state based on the mapping to the models parameters, so the model needs to be on the `map_location` first if it's not CPU for it to work.

PR #1222, opened 2023-03-22
Fix get_logger kwarg documentation issue

bcol23:
The `get_logger` example's keyword argument is not aligned with the function, resulting in inconsistency. This pull request aims to fix this issue by adjusting the keyword argument used in the example to match the function.


sgugger (approved):
Thanks for the fix!

Issue #1223, opened 2023-03-22
[who is using accelerate] add finetuner to readme


bwanglzu:
Hi guys i really want to thanks everyone who contributed and maintained the project! 

At @jina-ai we have a project called Finetuner: https://github.com/jina-ai/finetuner, the project was an open sourced python project then turned into a cloud sdk, for representation learning and information retrieval.

previously we implement everything by ourselves, such as DDP/half precision training, around a year ago we removed everything and start to dependent on accelerate. Now it has powered us to fine-tune 2000+ models for our users. 

Currently we're testing the accelerate with deepspeed plugin for language model training. And I realised in readme there is a session called [who is using accelerate](https://github.com/huggingface/accelerate#frameworks-using--accelerate). If you are interested, we are more than happy to add our project here. 

thanks again for your great work and hopefully i can contribute back to it!

muellerzr:
Hi @bwanglzu! Awesome library, yes we'd love to list your package on there! Feel free to open a PR for it if you'd like 😄

PR #1224, opened 2023-03-22
docs: add finetuner to ppl who use accelerate

bwanglzu:
this PR adds Finetuner to accelerate readme, as was mentioned in [this](https://github.com/huggingface/accelerate/issues/1223) issue.


sgugger:
Thanks for adding this!

PR #1229, opened 2023-03-22
Add additional check when patching env

Chris-hughes10:
This PR adds an additional check to ensure that an environment variable exists before deleting it when patching an environment. 
This prevents a `KeyError` being raised when training Multi-GPU using AzureML with the latest version of accelerate.

It would be awesome if this could be released quickly as a patch!

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -116,7 +116,8 @@ def patch_environment(**kwargs):
     yield
 
     for key in kwargs:
-        del os.environ[key.upper()]
+        if key.upper() in os.environ:
+            del os.environ[key.upper()]
 
 
 def get_pretty_name(obj):


sgugger (commented):
Out of curiosity, which variable gets deleted during the yield and is thus throwing an error after?

Chris-hughes10:
When you mount an AzureML data source it creates an environment variable with the mount path; if the registered dataset was called `imagenette` it would be `AZURE_ML_INPUT_IMAGENETTE`. At the moment, this means that every multi GPU jib we are running is failing.

Apologies for the lack of issue and spartan amount of detail on this btw, I just thought that it is a rather specific case and hard to reproduce without AzureML; all of the jobs I am currently running are internal so cannot be shared. I thought it is easier just to fix it, as it shouldn't interfere with anything else :D

sgugger (approved):
The fix is completely reasonable, I was just curious. Thanks!

Chris-hughes10:
Thanks for the speedy turnaround! When were you planning to next release?

sgugger:
We just add one recently, but I can put this in a patch tomorrow.

Chris-hughes10:
That would be amazing, much appreciated!

Issue #1230, opened 2023-03-22
main broke `accelerator.gradient_accumulation_steps`


stas00:
### System Info

```Shell
accelerate@main
```

### Reproduction

Breaks with `main`:
```
File "/mnt/nvme0/code/huggingface/m4-zero-stage3-resume/m4/training/trainer.py", line 290, in update_gas_and_gbs 
   self.accelerator.gradient_accumulation_steps = grad_acc_size_current 
AttributeError: can't set attribute
```
But it worked in the last release. 

And I was told this is the correct way to override `gradient_accumulation_steps`

### Expected behavior

perhaps there needs to be a public method to override `gradient_accumulation_steps` if the attributes are unstable.

Thank you.

PR #1233, opened 2023-03-22
Make grad accum steps mutable on the Accelerator object

muellerzr:
Mild regression and expansion on the new gradient accumulation API change. 

Keeps all main areas of the `gradient_state` to be mutable if needed/wanted from the `Accelerator` object that users may wish to do (so `sync_gradients` and `gradient_accumulation_steps`)

Solves https://github.com/huggingface/accelerate/issues/1230

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -819,9 +819,17 @@ class Accelerator:
     def sync_gradients(self):
         return self.gradient_state.sync_gradients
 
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks for fixing and adding a test!

stas00:
Thank you for the quick fix and the test, @muellerzr - I validated it's back to working again.

Issue #1232, opened 2023-03-22
[feature request] api to get deepspeed zero stage


stas00:
I'm struggling to figure out how to get the zero stage from `AcceleratorState().deepspeed_plugin` - my first attempt [failed](https://github.com/huggingface/accelerate/issues/1231) - would it be possible to provide a public API to return the stage regardless of which of the 2 config styles were used?

Thanks a lot

@pacman100

pacman100:
Hello Stas, please let us know if the above PR serves the purpose.  `AcceleratorState().deepspeed_plugin.zero_stage` should now work properly when using ds_config

PR #1235, opened 2023-03-23
backfill ds plugin attributes when using ds_config

### LINKED ISSUE #1231: invalid `deepspeed_plugin.zero_stage` when using ds_config file
### System Info

```Shell
`accelerate@main`
```

### Reproduction

when using a stage 3 config with ds_config file, this returns `2` instead of expected `3`:
```
    print(f"STAGE {AcceleratorState().deepspeed_plugin.zero_stage}")
```

I think this issue comes from:

https://github.com/huggingface/accelerate/blob/6e4e8702030222979d7ff54a2d8b0406cdf96393/src/accelerate/utils/dataclasses.py#L489-L490

why does it use the default of 2? this would be incorrect in 3 out of 4 cases as there are 4 stages.

If `zero_stage` isn't explicitly passed there is the config file to derive this value from. Bur perhaps you'd want to use this API to explicitly express the value that was passed via `zero_stage` - so the other possibility is to return `None` if it wasn't set by the API or the env var.

But the default should definitely not be `2`. 

@pacman100

pacman100:
### What does this PR do?
1. Fixes #1231  and #1232

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -515,28 +515,25 @@ class DeepSpeedPlugin:
                 raise ValueError("Please specify the ZeRO optimization config in the DeepSpeed config.")
 
             self._deepspeed_config_checks()
-            kwargs = {
-                "gradient_accumulation_steps": self.gradient_accumulation_steps,
-                "gradient_clipping": self.gradient_clipping if self.gradient_clipping else 1.0,
-                "zero_optimization.stage": self.zero_stage,
-                "zero_optimization.offload_optimizer.device": self.offload_optimizer_device,
-                "zero_optimization.offload_param.device": self.offload_param_device,
-                "zero_optimization.stage3_gather_16bit_weights_on_model_save": self.zero3_save_16bit_model,
+            plugin_to_config_mapping = {
+                "gradient_accumulation_steps": "gradient_accumulation_steps",
+                "gradient_clipping": "gradient_clipping",
+                "zero_stage": "zero_optimization.stage",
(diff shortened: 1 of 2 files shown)


stas00:
Thank you for working on this, Sourab.

So this changes the configuration quite a bit. Where does this leave env var config-wise? I'm asking since normally env vars have a last say, but this is no longer the case (but I don't think it ever was)

So currently "the chain of command" appears to be:

```
accelerate_config is overridden by env var config is overridden by ds_config file
```

if ds_config has an auto value, then it becomes:

```
accelerate_config is overridden by env var config
```

Is that the idea?

pacman100:
Hello @stas00, below are the ways DeepSpeed integration is meant to be used by users and most of them are indeed following this:
1. Use the `accelerate config` questionnaire to set the DeepSpeed config. Here, accelerate config is then used for filling the env variables which are subsequently used when running the `accelerate config`. This is for the majority of users that don't have the need for finer control of ds_config. 
2. Use the ds_config field in accelerate questionnaire with all the DS config with `auto` values. Now, the DeepSpeed plugin attributes will not be used as they are available through the ds_config. `auto` values are filled via cmd arguments and from the `prepare` call. To keep things in sync, I have now made the plugin attributes to pickup the values from ds_config if now passed through cmd args (cmd args set the env variables), this helps with the use case you have in a easier way.
3. The other way is to use `AccelerateState().deepspeedplugin.deepspeed_config` which is the dict object that will be used by DeepSpeedEngine.

The point of env variables was to pass the values from accelerate config or cmd arguments on the distributed setup while instantiating `AccelerateState`.

PR #1236, opened 2023-03-23
remove empty dicts while saving accelerate config

pacman100:
### What does this PR do?
1. Accelerate config is flooding with many empty dict fields as the number of integrations is increasing. This help accelerate config to only contain the necessary config


sgugger (approved):
Nice!

muellerzr (approved):
Great!

Issue #1228, opened 2023-03-22
Dispatching int8 model doesn't quantize the tensors.


BlackSamorez:
### System Info

```Shell
accelerate==0.17.1
bitsandbytes==0.26.0
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

tldr: [This kaggle notebook](https://www.kaggle.com/code/blacksamorez/accelerate-int8-bad-dispatch/notebook)

When using [load_checkpoint_and_dispatch](https://github.com/huggingface/accelerate/blob/a3065e18422260ad24cb4d0354d6f3aaae1074ad/src/accelerate/utils/modeling.py#L746) it [moves parameters to device](https://github.com/huggingface/accelerate/blob/a3065e18422260ad24cb4d0354d6f3aaae1074ad/src/accelerate/utils/modeling.py#L149) before loading it.
BitsAndBytes only [quantizes the weight when using](https://github.com/TimDettmers/bitsandbytes/blob/49a04253fb1f3e195cb0e9e79bdb01db7a490774/bitsandbytes/nn/modules.py#L191) `.to()` if it's **moved from cpu** .
Which means the weight never get quantized during dispatch.

### Expected behavior

```Shell
It would be great if it was possible to use load_checkpoint_and_dispatch to dispatch an int8 model without 2x model size GPU memory consumption.
```

sgugger:
Quantization is not supported yet directly in Accelerate, only on the Transformers side via the `from_pretrained` method. cc @younesbelkada

younesbelkada:
I see you shared a notebook, will have a look

pacman100:
@younesbelkada , Solving this would solve issues of PEFT + INT8 on multi-gpu loading too as many issues are being reported regarding this.

younesbelkada:
Awesome, having a look

younesbelkada:
Hi @BlackSamorez 
I made a PR https://github.com/huggingface/accelerate/pull/1237 that should fix your issue. Now the script: 
```python
from transformers import AutoConfig, AutoModel
from transformers.utils.bitsandbytes import replace_8bit_linear

from accelerate import init_empty_weights
from huggingface_hub import hf_hub_download

[...]

PR #1237, opened 2023-03-23
[`core`] Add Quantization support for `dispatch_model`

younesbelkada:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/1228

Before this PR, if a user wants to call `dispatch_model` on a customly created model, the model remains unquantized and leads to some errors that are hard to understand and debug. This issue has been also observed several times on other libraries that uses `accelerate` + `bitsandbytes` such as in `PEFT` that calls `dispatch_model` under the hood as well.: https://github.com/huggingface/peft/issues/187 

Now with this PR, the script below returns the expected output: 
```python
import torch
from transformers import AutoConfig, AutoModel
from transformers.utils.bitsandbytes import replace_8bit_linear

from accelerate import init_empty_weights
from huggingface_hub import hf_hub_download

from accelerate import load_checkpoint_and_dispatch
from itertools import chain

with init_empty_weights():
    model = AutoModel.from_config(AutoConfig.from_pretrained("bigscience/bloom-560m"))


model = replace_8bit_linear(model)

# For some reason replace_8bit_linear creates parameters with requires_grad=True but it's irrelevant rn
for p in model.parameters():
    p.requires_grad = False

model_path = hf_hub_download("bigscience/bloom-560m", "pytorch_model.bin")

model = load_checkpoint_and_dispatch(
    model,
    checkpoint=model_path,
    device_map="auto",
)

print(f"{model.h[0].self_attention.query_key_value.weight.device}\n{model.h[0].self_attention.query_key_value.weight.dtype}")
```

Also, this PR is fully backward compatible with `transformers` + `bitsandbytes` integration, I ran all the bnb slow tests and they pass.

cc @sgugger @pacman100

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -166,14 +166,14 @@ def set_module_tensor_to_device(
 
             if module.__class__.__name__ == "Linear8bitLt" and getattr(module.weight, "SCB", None) is None:
                 # quantize only if necessary
-                if not getattr(module.weight, "SCB", None) and isinstance(device, int):
+                device_index = torch.device(device).index if torch.device(device).type == "cuda" else None
+                if not getattr(module.weight, "SCB", None) and device_index is not None:
                     if module.bias is not None and module.bias.device.type != "meta":
                         # if a bias exists, we need to wait until the bias is set on the correct device
-                        module = module.cuda(device)
+                        module = module.cuda(device_index)
                     elif module.bias is None:
                         # if no bias exists, we can quantize right away
-                        module = module.cuda(device)
-
+                        module = module.cuda(device_index)
 
 def named_module_tensors(module: nn.Module, include_buffers: bool = True, recurse: bool = False):
     """
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -175,6 +175,7 @@ def set_module_tensor_to_device(
                         # if no bias exists, we can quantize right away
                         module = module.cuda(device_index)
 
+
 def named_module_tensors(module: nn.Module, include_buffers: bool = True, recurse: bool = False):
     """
     A helper function that gathers all the tensors (parameters + buffers) of a given module. If `include_buffers=True`
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -171,7 +171,7 @@ def set_module_tensor_to_device(
             ):
                 # quantize only if necessary
                 if module.bias.device.type != "meta" and not getattr(module.weight, "SCB", None):
-                    module = module.cuda()
+                    module = module.cuda(device)
 
 
 def named_module_tensors(module: nn.Module, include_buffers: bool = True, recurse: bool = False):
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -156,10 +156,22 @@ def set_module_tensor_to_device(
             param_cls = type(module._parameters[tensor_name])
             kwargs = module._parameters[tensor_name].__dict__
             if param_cls.__name__ == "Int8Params":
+                # downcast to fp16 if any
+                if new_value.dtype == torch.float32:
+                    new_value = new_value.to(torch.float16)
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
             else:
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad).to(device)
             module._parameters[tensor_name] = new_value
+
+            if (
+                module.__class__.__name__ == "Linear8bitLt"
+                and getattr(module.weight, "SCB", None) is None
+                and module.bias is not None
+            ):
+                # quantize only if necessary
+                if module.bias.device.type != "meta" and not getattr(module.weight, "SCB", None):
+                    module = module.cuda()
 
 
 def named_module_tensors(module: nn.Module, include_buffers: bool = True, recurse: bool = False):
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -170,7 +170,11 @@ def set_module_tensor_to_device(
                 and module.bias is not None
             ):
                 # quantize only if necessary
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -156,11 +156,27 @@ def set_module_tensor_to_device(
...
+                    new_value = new_value.to(torch.float16)
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
             else:
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad).to(device)
             module._parameters[tensor_name] = new_value
 
+            if (
+                module.__class__.__name__ == "Linear8bitLt"
+                and getattr(module.weight, "SCB", None) is None
+                and module.bias is not None
```
BlackSamorez, line 170: Why require bias?

younesbelkada: Good point, this seems to be not needed indeed

BlackSamorez:
```suggestion
```

younesbelkada: I think you need more checks, I will push something now

younesbelkada: f7e9fd3

younesbelkada: Btw you can test it right away on your Kaggle notebook with:
```bash
pip install git+https://github.com/younesbelkada/accelerate.git@fix-dispatch-int9
```

src/accelerate/utils/modeling.py
```diff
@@ -156,11 +156,24 @@ def set_module_tensor_to_device(
...
+                if new_value.dtype == torch.float32:
+                    new_value = new_value.to(torch.float16)
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad, **kwargs).to(device)
             else:
                 new_value = param_cls(new_value, requires_grad=old_value.requires_grad).to(device)
             module._parameters[tensor_name] = new_value
 
+            if module.__class__.__name__ == "Linear8bitLt" and getattr(module.weight, "SCB", None) is None:
+                # quantize only if necessary
+                if not getattr(module.weight, "SCB", None) and isinstance(device, int):
```
BlackSamorez, line 169: It doesn't work when device is `"cuda:0"` of `torch.device.cuda`.

younesbelkada: I think in this method `device` is always an int for GPUs, can you share a reproducible script?

BlackSamorez: It look like it isn't, when using a custom `device_map`. I'll try and create a smaller script because right now I'm working with 13B model with custom wrappers.
But from what I've seen it work if and only if devices in a custom `device_map` are `int`s

src/accelerate/utils/modeling.py
```diff
@@ -156,11 +156,24 @@ def set_module_tensor_to_device(
...
+                    if module.bias is not None and module.bias.device.type != "meta":
+                        # if a bias exists, we need to wait until the bias is set on the correct device
+                        module = module.cuda(device)
```
BlackSamorez, line 172: This will fail if weight requires grad with `data set to a tensor that requires gradients must be floating point or complex dtype`. Maybe check for module.weight.requires_grad or throw own error.

src/accelerate/utils/modeling.py
sgugger: I think you want to test device.type here (after converting the device received to a torch.device)

src/accelerate/utils/modeling.py
sgugger, line 172: Let's use module.to(...)

src/accelerate/utils/modeling.py
```diff
@@ -156,11 +156,24 @@ def set_module_tensor_to_device(
...
+                    elif module.bias is None:
+                        # if no bias exists, we can quantize right away
+                        module = module.cuda(device)
```
sgugger, line 175: Here too

sgugger (commented):
Thanks for working on this!

src/accelerate/utils/modeling.py
```diff
@@ -156,11 +156,24 @@ def set_module_tensor_to_device(
...
+
```
BlackSamorez, line 176: This both checks `device.type` and gets correct index from any device representation
```suggestion
            if module.__class__.__name__ == "Linear8bitLt" and getattr(module.weight, "SCB", None) is None:
                # quantize only if necessary
                device_index = torch.device(device).index if torch.device(device).type == "cuda" else None
                if not getattr(module.weight, "SCB", None) and device_index is not None:
                    if module.bias is not None and module.bias.device.type != "meta":
                        # if a bias exists, we need to wait until the bias is set on the correct device
                        module = module.cuda(device_index)
                    elif module.bias is None:
                        # if no bias exists, we can quantize right away
                        module = module.cuda(device_index)
```

src/accelerate/utils/modeling.py
younesbelkada: Yes, since we place the model weights on gpu already, we need to "force quantize" the model by calling `cuda()`

src/accelerate/utils/modeling.py
younesbelkada: I believe this has been addressed in [bfd7e1a](https://github.com/huggingface/accelerate/pull/1237/commits/bfd7e1afea3cee100dcbed91ae6ffd89dee09432) (also added some tests later on)

src/accelerate/utils/modeling.py
younesbelkada: Let's force `module.requires_grad` to be `False` in `replace_8bit_linear`, I will make a follow up PR on `transformers` for this

younesbelkada: https://github.com/huggingface/transformers/pull/22396

sgugger (approved):
Looks good on my side. @BlackSamorez can you confirm this solves all your issues?

BlackSamorez:
@sgugger yes, everything works great on my end.
Thanks everyone!

PR #1240, opened 2023-03-23
Make model conversion compatible with TE < v0.4

abuccts:
In Transformer Engine version less than v0.4, te.LayerNorm uses `layer_norm_weight/layer_norm_bias` attributes instead of `weight/bias`, which will break `convert_model` to/from transformer engine models.

This PR adds the backward compatibility for layer norm in lower TE versions, [reference](https://github.com/NVIDIA/TransformerEngine/blob/v0.4/transformer_engine/pytorch/module.py#L2580-L2584).


sgugger:
This version of Transformer Engine has many bugs that yielded to a degradation in performance, so we do not support it on purpose. You should use the most recent release for FP8 mixed precision training.

PR #1241, opened 2023-03-23
Handle multiple tied parameters

sgugger:
This PR reworks the internal of big model inference to handle multiple tied parameters (instead of just a pair). It contains a breaking change in the util function `find_tied_parameters` (which is used once in Transformers, so we will need an update there). This is because its results as a dictionary doesn't really make sense when we can have groups of three/four tied parameters all together.

The logic in `infer_auto_device_map` is getting a lot more complicated, so I added a `verbose` argument to leave plenty of print statements and help future me or others debug what's happening. Tests and integration tests are added as well.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -198,7 +198,7 @@ def find_tied_parameters(model: nn.Module, **kwargs):
         model (`torch.nn.Module`): The model to inspect.
 
     Returns:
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
Thanks! This all looks good to me. Test failure is because `test_infer_auto_device_map_on_t0pp(self)` needs to use the `@require_huggingface_suite` decorator I believe.

PR #1249, opened 2023-03-24
Fix pypi image

muellerzr:
Needs to be a full `raw.githubusercontent.com/*` for the image to show up on pypi, solves the logo not showing here: https://pypi.org/project/accelerate


sgugger (approved):
Thanks for the fix!

PR #1251, opened 2023-03-24
Handle missing deepspeed config

HeyangQin:
This PR improves the handling of cases where users provide a non-existing deepspeed config path

- Parsing non-base64 strings with `base64.urlsafe_b64decode` raises `binascii.Error`, which is a subclass of `ValueError`. The previous error handling code won't catch it
- `config` is not defined if this piece of code errors out so we need to use `config_file_or_dict` instead in the error handling


sgugger:
cc @pacman100

pacman100 (approved):
Thank you @HeyangQin for all the fixes, LGTM! 🤗

Issue #700, opened 2022-09-14
Integrate Intel Extension for Pytorch into accelerate to get out-of-box optimizations on intel platform.


tangleintel:
### Motivation
Intel Extension for Pytorch provides extra optimization and performance boost on intel platform for both inference and training. Integrate it into accelerate can make users who use distributed training or evaluation get out-of-box performance boost on intel hardware.

#### User interface

- The first thing is how we tell accelerate to enable ipex. We design to extend the accelerate class's constructor with an additional argument 'use_ipex' to indicate wether we will use ipex in accelerate and the default value of this argument is False, thus explicitly tell accelerate if we want to use IPEX. Such as:

```
class Accelerator:
    def __init__(
        self,
        device_placement: bool = True,
        split_batches: bool = False,
        fp16: bool = None,
        mixed_precision: Union[PrecisionType, str] = None,
        gradient_accumulation_steps: int = 1,
        cpu: bool = False,
        deepspeed_plugin: DeepSpeedPlugin = None,
        fsdp_plugin: FullyShardedDataParallelPlugin = None,
        rng_types: Optional[List[Union[str, RNGType]]] = None,
        log_with: Optional[List[Union[str, LoggerType, GeneralTracker]]] = None,
        logging_dir: Optional[Union[str, os.PathLike]] = None,
        dispatch_batches: Optional[bool] = None,
        step_scheduler_with_optimizer: bool = True,
        kwargs_handlers: Optional[List[KwargsHandler]] = None,
        use_ipex: bool = False,
    ):

accelerate = Accecerator(use_ipex=True)
```
If we set use_ipex equals True, we will import the package in the constructor if available. 
- Model and optimizer wrapper for distributed training. 
```
model, optim, data = accelerator.prepare(model, optim, data)
```
Accelerator.prepare() is the main method where most magic happens, IPEX can also hidden optimization magic via a similar front end API called ipex.optimize(). If we choose use ipex, then we can automatically invoke ipex's API inside prepare() according to different distribute configuration. So, no code change in the user interface side.

- Auto Mixed Precision
In CPU side, currently we only support bf16 half precision, we can further extend the ability of AMP on CPU side.

pacman100:
Hello @tangleintel, Thank you for the detailed feature request. I've made the above draft PR. I currently don't have access to instance with latest Intel CPUs to perform the initial testing. Could you please try it out on your end and let us know.
You can enable ipex via below options:
1. Through  `accelerate config` as shown below:
```
[...]

PR #1255, opened 2023-03-27
ipex intel extension for pytorch integration

sywangyi:
What does this PR do?
Implements feature request https://github.com/huggingface/accelerate/issues/700


src/accelerate/accelerator.py
```diff
@@ -1556,6 +1571,44 @@ def _prepare_megatron_lm(self, *args):
...
+        optimizer = None
+        dataloader = None
+        result = [obj for obj in args]
+        for obj in result:
+            if isinstance(obj, torch.utils.data.DataLoader) and dataloader is None:
+                dataloader = obj
+            elif isinstance(obj, torch.nn.Module):
+                model = obj
+            elif isinstance(obj, (torch.optim.Optimizer)):
+                optimizer = obj
```
sgugger, line 1593: This assumes there is only one model/one optimizer and one dataloader. Many users pass along their training and validation dataloader together for instance, so there should be asserts and clear error message for those situations.

src/accelerate/accelerator.py
```diff
@@ -1556,6 +1571,44 @@ def _prepare_megatron_lm(self, *args):
...
+        if ipex_plugin.do_fusion and optimizer is None:
+            model.eval()
+            model = ipex.optimize(model, dtype=ipex_plugin.dtype, level="O1")
+            model = ipex_plugin.torch_jit_model_eval(model, dataloader, training=False)
+        else:
+            if not model.training:
+                model.train()
+            model, optimizer = ipex.optimize(
+                model, dtype=ipex_plugin.dtype, optimizer=optimizer, inplace=True, level="O1"
+            )
```
sgugger, line 1603: This assumes that the user will always pass along the optimizer with the model when training it, and that when they don't pass an optimizer, they are doing evaluation only. As said above, many users prepare everything together. Those assumptions should be clearly documented as they break from the rest of the library.

src/accelerate/utils/dataclasses.py
```diff
@@ -1324,3 +1324,65 @@ def set_tensorboard_logging_options(self):
...
+            self.do_fusion = False
+
+    def set_mixed_precision(self, mixed_precision):
+        if mixed_precision == "fp16":
+            raise ValueError("Tried to use `fp16` but it is not supported on cpu")
+        elif mixed_precision == "bf16":
+            self.dtype = torch.bfloat16
+
+    def torch_jit_model_eval(self, model, dataloader, training=False):
+        if not training:
```
sgugger, line 1357: This is only used with `training=False` so there is no point having this argument and then this test.

sgugger (commented):
Thanks for opening a fresh PR. I have some comments, and we would also need to add tests for this.

src/accelerate/accelerator.py
```diff
@@ -1260,7 +1275,7 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
         if self.distributed_type == DistributedType.TPU and self.state.fork_launched:
             model = xmp.MpModelWrapper(model).to(self.device)
         # torch.compile should be called last.
-        if self.state.dynamo_plugin.backend != DynamoBackend.NO:
+        if self.state.dynamo_plugin and self.state.dynamo_plugin.backend != DynamoBackend.NO:
```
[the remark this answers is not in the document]
sywangyi: fix it

src/accelerate/accelerator.py
sywangyi: yes, we should not take the assumption like this. ipex will be only used for training part optimization as described above

sywangyi:
> Thanks for opening a fresh PR. I have some comments, and we would also need to add tests for this.

I add testcase, please help review, later, I will add doc about this

sgugger (approved):
Thanks for your work on this! @pacman100 could you have a second look?

sywangyi:
@pacman100 could you help with the review?

pacman100 (commented):
Hello, Thank you @sywangyi for adding ipex support and iterating on the previous PR 🤗. I believe adding docs with the below details would help users to get a better idea about when and how to use it:
1. Brief intro on what ipex does and what is integrated currently in accelerate (only training for now)
2. How to use the current integration
3. Link to blog/original docs showing benefits when using ipex on CPU in comparison to without it

For reference, please find the DeepSpeed integration docs here: https://huggingface.co/docs/accelerate/usage_guides/deepspeed

sywangyi:
agree, I will create a new PR for the document. For this PR, could you help with the merge?

pacman100 (approved):

PR #1262, opened 2023-03-29
Adds `in_order` argument that defaults to False, to log in order.

JulesGM:
Ads `in_order` argument that defaults to `False`, to log in order.  It really helps with readability.  Defaults to false to not break backwards comp.

### DIFF
--- a/src/accelerate/logging.py
+++ b/src/accelerate/logging.py
@@ -14,7 +14,6 @@
 
 import logging
 import os
-import torch
 
 from .state import PartialState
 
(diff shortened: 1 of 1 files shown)


sgugger:
cc @muellerzr

src/accelerate/logging.py
```diff
@@ -40,11 +41,26 @@ def log(self, level, msg, *args, **kwargs):
...
         if self.isEnabledFor(level) and self._should_log(main_process_only):
             msg, kwargs = self.process(msg, kwargs)
             self.logger.log(level, msg, *args, **kwargs)
+        elif self.isEnabledFor(level) and in_order:
+            state = PartialState()
+            for i in range(state.num_processes):
+                if i == state.process_index:
+                    msg, kwargs = self.process(msg, kwargs)
+                    self.logger.log(level, msg, *args, **kwargs)
+                torch.distributed.barrier()
```
muellerzr, line 63: This should be replaced with `state.wait_for_everyone()` so it can work on all distributed types

src/accelerate/logging.py
```diff
@@ -53,6 +69,8 @@ def get_logger(name: str, log_level: str = None):
 
     If a log should be called on all processes, pass `main_process_only=False`
 
+    If a lot should be called on all processes in order, also pass `in_order=True`
```
muellerzr, line 72:
```suggestion
    If a log should be called on all processes, pass `main_process_only=False`
    If a log should be called on all processes and in order, also pass `in_order=True`
```

src/accelerate/logging.py
```diff
@@ -40,11 +41,26 @@ def log(self, level, msg, *args, **kwargs):
...
+
+        `in_order` is ignored if `main_process_only` is passed.
         """
         main_process_only = kwargs.pop("main_process_only", True)
+        in_order = kwargs.pop("in_order", False)
+
         if self.isEnabledFor(level) and self._should_log(main_process_only):
             msg, kwargs = self.process(msg, kwargs)
             self.logger.log(level, msg, *args, **kwargs)
+        elif self.isEnabledFor(level) and in_order:
```
muellerzr, line 57: Let's refactor this under one `if self.isEnabledFor()` and then have our checks for `main_process_only` and `in_order`. Likely this check should be first and then move to the `main_process_only` (so reverse what we have here)

muellerzr (commented):
Thanks for the PR! I certainly can see the usefulness in this idea. I left some review comments that should be addressed, such as a more distributed-friendly alternative to using `barrier` for the TPU folks :) 

Great work!

muellerzr:
Also, to pass the quality check do `pip install -e .[quality]` followed by `make style; make quality;`

muellerzr:
Thanks for your work!

PR #1264, opened 2023-03-30
fix issue template

stas00:
Expected behavior most likely is a normal text and not a code section


sgugger (approved):
Thanks for the fix!

PR #1267, opened 2023-03-30
Change error raised to ValueError

sgugger:
There is something weird going on with `TypeError` not being properly raised in the PyTorch distributed launcher. This PR thus changes the type of the error raised when one tries to gather/reduce something that is not a Tensor.


LysandreJik (approved):
Cool!

PR #1278, opened 2023-04-03
fix attribute error in DataloaderShared

ZhiyuanChen:
When running in single GPU, the `batch_sampler` of `DataLoaderShared` is a `torch.utils.data.sampler.BatchSampler` object instead of `DataSamplerShared ` object, which does not contain necessary attributes to calculate `total_batch_size`.


sgugger (approved):
Thanks for the fix!

PR #1279, opened 2023-04-03
fix DataLoaderShared.batch_size is None

ZhiyuanChen:


ZhiyuanChen:
> I would leave the `batch_size` attribute to PyTorch. If it's `None`, then it was `None` before the dataloader was prepared.

Hmmm, I really need to access batch_size before prepare. 
Is there any other workaround? 
Maybe get batch_size from batch_sampler before preparation?

Issue #1276, opened 2023-04-02
FP8 causes exception: name `te` not defined


JulesGM:
### System Info

```Shell
Accelerate 0.18.0
Transformers 4.27.4
Torch 2.0.0
```

### Reproduction

When launching with `mixed_precision=fp8` in the config_file, I get

![name `te` is not defined](https://i.imgur.com/JX6HeI0.png)

muellerzr:
Hi @JulesGM, this should be solved via https://github.com/huggingface/accelerate/pull/1283

Can you do `pip install git+https://github.com/huggingface/accelerate` to double check? :)

JulesGM:
hello @muellerzr , I get `ImportError: Using `has_transformer_engine_layers` requires transformer_engine to be installed.`

JulesGM:
<img width="412" alt="Screen Shot 2023-04-03 at 12 48 19 PM" src="https://user-images.githubusercontent.com/3231217/229575197-a5cf166d-1f59-48c8-8fbe-62647d48318d.png">

muellerzr:
That is correct, have you installed the transformer engine on your system? :)

JulesGM:
hmm also, it's only on `H100`s, ok, sad. I was hoping it would also be ok with A100 somehow.

JulesGM:
> No I have not. But it is a weird error to get. I did not know I needed transformer_engine to use fp8, I just selected the option because I saw it in accelerate config.

I suggest that the library checks if `transformer_engine` is installed, and raises if it's not, explaining that it's required. It would be more clean than this.

JulesGM:
Like this, it makes one think it could be a bug.

muellerzr:
Thanks @JulesGM, that does indeed make sense. Should be addressed via https://github.com/huggingface/accelerate/pull/1287, need to see with Sylvain if we want to raise that in the CLI as well (since config files can be swapped/passed through machines). However this will raise the right error when running the script now more clearly

JulesGM:
cool thanks. I can close this if you want

muellerzr:
Sure :)

PR #1283, opened 2023-04-03
Raise import error if fp8 not available in `has_transfomer_engine_layers`

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1276 by also including a check for `is_fp8_available`


sgugger (approved):
Thanks for the fix!

Issue #1275, opened 2023-04-02
fp 8 doesn't work with accelerate launch command line args


JulesGM:
### System Info

```Shell
Accelerate 0.18.0, shows `fp8` as a choice with `accelerate config`, but doesn't accept that as a command line argument with `accelerate launch --mixed_precision=fp8`.

It accepts the argument when specified in the `--config_file` yaml though
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

`accelerate launch --mixed_precision=fp8 {script.py}`

### Expected behavior

accept the parameter

PR #1284, opened 2023-04-03
Add missing FP8 options to CLI

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1275


sgugger (approved):
Thanks for adding those!

Issue #1263, opened 2023-03-30
`accelerator.gather` silently fails if any item in the list is not a tensor


stas00:
### System Info

```Shell
main
```

### Reproduction

```
$ cat test.py
import torch
from accelerate import Accelerator

accelerator = Accelerator()
process_tensor = torch.tensor([accelerator.process_index], device="cuda")
gathered_tensor, i = accelerator.gather((process_tensor, 1))
```

run:

```
$ python -m torch.distributed.run --nproc_per_node=2 test.py 
WARNING:__main__:
*****************************************
Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
*****************************************
Traceback (most recent call last):
Traceback (most recent call last):
  File "test.py", line 6, in <module>
  File "test.py", line 6, in <module>
        gathered_tensor, i = accelerator.gather((process_tensor, 1))gathered_tensor, i = accelerator.gather((process_tensor, 1))

ValueErrorValueError: : not enough values to unpack (expected 2, got 0)not enough values to unpack (expected 2, got 0)
```

### Expected behavior

`accelerator.gather` doesn't validating the incoming argument and fails with a misleading error that doesn't tell the user what's wrong.

Expecting for it to tell the user that one or more items on the list aren't tensors and not return an empty list.

Please note that it works fine on a single gpu, so the test needs to be 2 gpus or more.

Thank you.

sgugger:
The error is there, but it looks like `torchrun` (or `torch.distributed.run`) ignores TypeErrors... I'll change the error to a `ValueError`, which will provide the (more helpful) error message
```
ValueError: Can't apply _gpu_gather_one on object of type <class 'int'>, only of nested list/tuple/dicts of objects that satisfy is_torch_tensor.
```

stas00:
That works, if you think the user can connect the dots as this is not the function the user calls. But hopefully they will figure it out from the traceback.

I think checking the inputs are all tensors is probably expensive, since it's the same function called again and again and this overhead will add up.

So your proposal works for me.

Thank you, Sylvain

gebalassa:
> The error is there, but it looks like `torchrun` (or `torch.distributed.run`) ignores TypeErrors... I'll change the error to a `ValueError`, which will provide the (more helpful) error message
> 
> ```
[...]

PR #1285, opened 2023-04-03
Use a new Exception to catch second type error

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1263

The basic issue relies on the fact that a `TypeError` was raised initially, and in turn can't really be raised "again" for a second test of if something can be done, due to the `else` in `honor_type`:
```python
def honor_type(obj, generator):
    """
    Cast a generator to the same type as obj (list, tuple, or namedtuple)
    """
    try:
        return type(obj)(generator)
    except TypeError: <- Our culprit
        # Some objects may not be able to instantiate from a generator directly
        return type(obj)(*list(generator))
```
This PR solves it by adding a `TypedApplyException` which will raise the true error better.

This PR also added a test.

Proof it works with the example @stas00 showed:

```bash
  File "/home/zach_mueller_huggingface_co/test_distrib.py", line 6, in <module>
    gathered_tensor, i = accelerator.gather((process_tensor, 1))
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/accelerator.py", line 1869, in gather
    return gather(tensor)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 229, in gather
    return _gpu_gather(tensor)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 209, in _gpu_gather
    return recursively_apply(_gpu_gather_one, tensor, error_on_other_type=True)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 83, in recursively_apply
    return honor_type(
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 54, in honor_type
    return type(obj)(generator)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 86, in <genexpr>
    recursively_apply(
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/operations.py", line 104, in recursively_apply
    raise TypedApplyException(
accelerate.utils.exceptions.TypedApplyException: Can't apply _gpu_gather_one on object of type <class 'int'>, only of nested list/tuple/dicts of objects that satisfy is_torch_tensor.
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: 1) local_rank: 0 (pid: 4009209) of binary: 
...
```

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -25,6 +25,7 @@ import torch
 from ..state import PartialState
 from .constants import CUDA_DISTRIBUTED_TYPES
 from .dataclasses import DistributedType, TensorInformation
+from .exceptions import TypedApplyException
 from .imports import is_torch_distributed_available, is_tpu_available
 from .versions import is_torch_version
 
@@ -100,7 +101,7 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
     elif test_type(data):
         return func(data, *args, **kwargs)
     elif error_on_other_type:
-        raise ValueError(
+        raise TypedApplyException(
             f"Can't apply {func.__name__} on object of type {type(data)}, only of nested list/tuple/dicts of objects "
             f"that satisfy {test_type.__name__}."
         )
--- a/src/accelerate/utils/__init__.py
+++ b/src/accelerate/utils/__init__.py
@@ -22,6 +22,7 @@ from .dataclasses import (
     TorchDynamoPlugin,
 )
 from .environment import get_int_from_env, parse_choice_from_env, parse_flag_from_env
+from .exceptions import TypedApplyException
 from .imports import (
     get_ccl_version,
     is_aim_available,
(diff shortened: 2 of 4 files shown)


sgugger (commented):
Mmm the goal is to fix us gobbling all `TypeError`, not creating a new one as a patch.

sgugger (approved):
Thanks for the fix!

src/accelerate/utils/operations.py
```diff
@@ -100,7 +115,7 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
     elif test_type(data):
         return func(data, *args, **kwargs)
     elif error_on_other_type:
-        raise ValueError(
+        raise TypeError(
             f"Can't apply {func.__name__} on object of type {type(data)}, only of nested list/tuple/dicts of objects "
             f"that satisfy {test_type.__name__}."
```
stas00, line 120: while at it could this message be reworded to make sense "only of nested" can't be parsed, IMHO.

stas00: perhaps, `"Expected all elements of list/tuple/dict to be a tensor, but got {type(data)}"`

muellerzr: I'll do this in a follow up PR as it's in more than one place.

PR #1287, opened 2023-04-03
Raise more explicit error when transformer_engine isn't installed

muellerzr:
Addresses https://github.com/huggingface/accelerate/issues/1276 by raising an error in `PartialState` when passing in `fp8` to the precision and its unsupported. I don't believe we should address this in `accelerate config` as well, since config files can be passed around. If you disagree @sgugger and it would be good to include that check during the CLI questionnaire, I can do so.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -552,8 +552,7 @@ class AcceleratorState:
             )
             if mixed_precision == "fp8" and not is_fp8_available():
                 raise ValueError(
-                    "Using `fp8` precision requires `transformer_engine` to be installed and utilizing supported hardware. "
-                    "Currently your system is not supported."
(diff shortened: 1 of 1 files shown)


src/accelerate/state.py
```diff
@@ -549,6 +550,10 @@ def __init__(
                 if mixed_precision is None
                 else mixed_precision.lower()
             )
+            if mixed_precision == "fp8" and not is_fp8_available():
+                raise ValueError(
+                    "Using `fp8` precision requires `transformer_engine` to be installed and utilizing supported hardware."
```
sgugger, line 555:
```suggestion
                    "Using `fp8` precision requires `transformer_engine` to be installed."
```

sgugger (approved):
Thanks for adding this!

Issue #1246, opened 2023-03-24
AttributeError: 'dict' object has no attribute 'dtype'


shaohm:
### System Info

```Shell
- `Accelerate` version: 0.16.0
- Platform: Linux-5.19.0-35-generic-x86_64-with-glibc2.35
- Python version: 3.9.16
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. Create a model extends `nn.Module`
2. implements `get_extra_state_dict`, and `set_extra_state_dict`
3. let `get_extra_state_dict` return something other than a tensor. mine returned an ordinary python dict
4. try saving your model by call `accelerator.get_state_dict(self.model)`
5. you will see

>   File "/home/bernard/diffusions/my_diffusion/my_trainer.py", line 155, in save
    'model': self.accelerator.get_state_dict(self.model),
  File "/home/bernard/miniconda3/envs/ddpm/lib/python3.9/site-packages/accelerate/accelerator.py", line 2274, in get_state_dict
    if state_dict[k].dtype == torch.float16:
AttributeError: 'dict' object has no attribute 'dtype'

### Expected behavior

```Shell
File "/home/bernard/diffusions/my_diffusion/my_trainer.py", line 155, in save
'model': self.accelerator.get_state_dict(self.model),
File "/home/bernard/miniconda3/envs/ddpm/lib/python3.9/site-packages/accelerate/accelerator.py", line 2274, in get_state_dict
if state_dict[k].dtype == torch.float16:
AttributeError: 'dict' object has no attribute 'dtype'
```

sgugger:
cc @muellerzr

muellerzr:
@shaohm could you clarify a few notes for me please?

- Does calling `model.state_dict()` call your `get_extra_state_dict` to get the model? 
- Are you expecting `Accelerator.get_state_dict()` to return said extra thing

Changing the `dtype` part is something doable, I just want to know if you're expecting more behavior than we provide

shaohm:
[...]

PR #1288, opened 2023-04-03
Only check for dtype if it has it in get_state_dict

muellerzr:
Partial solution to https://github.com/huggingface/accelerate/issues/1246 (I believe), which will always let the user pull the model, and we only check for fp16 if the state dict has the capability to do so


sgugger (approved):
Thanks for the fix!

PR #1292, opened 2023-04-04
[`bnb`] fix bnb slow test

younesbelkada:
# What does this PR do?

This PR fails the bnb slow tests that are currently failing. 
I was testing the tests on the `main` branch of `transformers` that included https://github.com/huggingface/transformers/pull/22238 . That fix was necessary to correctly call `replace_8bit_linear` that expects a list of strings as input for `modules_not_to_replace`, otherwise some checks will fail, leading to modules not being properly converted in int8

Afterwards I hit another issue where the device map missed the parameters of the model `*.SCB` (the quantization parameters) in `check_device_map`. Since these modules are not registered as a proper submodule the correct fix is to add a device_map that puts the entire model on a given device. 

This time ran the tests on the latest transformers release and they all pass

cc @sgugger

### DIFF
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -527,7 +527,8 @@ class BigModelingTester(unittest.TestCase):
         with init_empty_weights():
             model = AutoModel.from_config(AutoConfig.from_pretrained("bigscience/bloom-560m"))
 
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for the fix!

Issue #1277, opened 2023-04-03
The performance of DDP acceleration is lower than single GPU training.


JiuFengSC:
### System Info

```Shell
- `Accelerate` version: 0.15.0
- Platform: Linux-5.4.0-81-generic-x86_64-with-glibc2.17
- Python version: 3.8.13
- Numpy version: 1.23.5
- PyTorch version (GPU?): 1.13.1 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: fp16
        - use_cpu: False
        - dynamo_backend: NO
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - main_process_ip: None
        - main_process_port: None
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_name: None
        - tpu_zone: None
        - command_file: None
        - commands: None
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I was finetuning the CLIP model with another dataset.
I used the HuggingFace Accelerate library to speed up, but the problem is that DDP training’s performance is lower than single GPU training. Some web pages said Batch Normalization might cause it, since every single batch is divided into different GPUs, we need to synchronize the BN value.
But there is no BN in CLIP/transformer-based architecture.

The Loss gap is like this figure Upper one is hugging face DDP and the bottom one is single GPU training. Also, there is a gap in performance.
<img width="416" alt="image" src="https://user-images.githubusercontent.com/80580308/229421395-c729cdec-0188-41f0-933a-a06cbf7753f0.png">

To simplify, here is only main() function and train() function. 

```python
def train(clip_model, train_dataloader, Max_Epoch, loss_img, loss_txt, optimizer, device, start_epoch=0):
    clip_model.train()
    logfile = open(args.exp_dir+"/finetuning_log.txt", "a+")
    time_str = time.strftime('%Y-%m-%d %H:%M:%S', time.localtime())
    logfile.write(' Start Training{}\n'.format(time_str))
[...]

PR #1296, opened 2023-04-06
Make note about grad accum and prec in performance documentation

muellerzr:
Helps https://github.com/huggingface/accelerate/issues/1277 by adding a note based on my tests yesterday. tl;dr: When trying to compare performance you won't see a 1:1 if you use gradient accumulation or mixed precision due to averaging of the gradients and their overall precision, though the end result should be similar.


sgugger (approved):
Thanks for the note!

PR #1297, opened 2023-04-06
fix for load_checkpoint_and_dispatch(device_map=None)

anentropic:
The `load_checkpoint_and_dispatch` method has this kwarg type:

`device_map: Optional[Union[str, Dict[str, Union[int, str, torch.device]]]] = None,`

But if you pass `device_map=None` you get an error:

```
accelerate/big_modeling.py", line 477, in load_checkpoint_and_dispatch
    if offload_state_dict is None and "disk" in device_map.values():
AttributeError: 'NoneType' object has no attribute 'values'
```


sgugger (approved):
Thanks for the fix!

Issue #1303, opened 2023-04-08
main_process_first() and local_main_process_first() decorators do not work


flukeskywalker:
### System Info

```Shell
- `Accelerate` version: 0.18.0
- Platform: Linux-4.9.0-6-amd64-x86_64-with-glibc2.24
- Python version: 3.11.2
- Numpy version: 1.24.2
- PyTorch version (GPU?): 2.0.0+cu117 (True)
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

test.py:
```python
from accelerate import Accelerator
from time import sleep
import os


def main():
    accelerator = Accelerator()
    with accelerator.main_process_first():
        print("HEY", accelerator.is_main_process)
        sleep(3)
        print("BYE", accelerator.is_main_process)


if __name__ == "__main__":
    main()
```
Run with:
```accelerate launch --multi_gpu --num_processes=2 test.py```

Output:
```
HEY True
HEY False
BYE True
BYE False
```
Thanks to @engintoklu for minimal example this is based on.

### Expected behavior

Should output BYE from main process before HEY from non-main processes.

flukeskywalker:
I have a working fix. Will submit PR.

sgugger:
cc @muellerzr

PR #1304, opened 2023-04-09
Fix nested context manager for main_process_first()

flukeskywalker:
This PR proposes a fix for #1303.

The bug seems to be in delegation from one context manager (for example from `Accelerator`) to another one (from `AccelerateState`). Currently, the following pattern is used:

```python
@contextmanager
def main_process_first(self):
    yield self.state.main_process_first()
```
However, when executed, this context manager simply yields without entering `self.state.main_process_first()`, causing the bug. The solution is to correctly enter the inner context manager using `with`.

This was a very interesting bug. Since I haven't defined nested context managers like this before, I wanted to understand what the right way to do this is, and whether the current implementation in the library was correct. Like the cool kids I even asked ChatGPT, which explained what the current implementation was trying to do and how to use `yield context_manager()` or `yield from context_manager()`. Unfortunately it turned out that ChatGPT was hallucinating its explanation and use of these concepts, and even the code examples which it produced did not run and produced an error instead. 

In summary, according to ChatGPT the current implementation is correct, though it's not. So I'm very curious if I'm understanding it correctly now, and whether GPT-3/Copilot/ChatGPT had anything to do with the original implementation :)

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -755,7 +755,8 @@ class Accelerator:
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
-        yield self.state.main_process_first()
+        with self.state.main_process_first():
+            yield
 
     @contextmanager
     def local_main_process_first(self):
@@ -776,7 +777,8 @@ class Accelerator:
         ...     print(f"This will be printed by process {accelerator.local_process_index}")
         ```
         """
-        yield self.state.local_main_process_first()
+        with self.state.local_main_process_first():
+            yield
 
     @contextmanager
     def no_sync(self, model):
--- a/src/accelerate/test_utils/scripts/test_script.py
+++ b/src/accelerate/test_utils/scripts/test_script.py
@@ -18,6 +18,7 @@ import contextlib
 import io
 import os
 import time
+from pathlib import Path
 
 import torch
 from torch.utils.data import DataLoader
@@ -56,14 +57,31 @@ def print_on(state, process_idx):
 def process_execution_check():
     accelerator = Accelerator()
     num_processes = accelerator.num_processes
+
+    # Test main_process_first context manager
+    path = Path("check_main_process_first.txt")
+    if path.exists():
+        path.unlink()
     with accelerator.main_process_first():
         if accelerator.is_main_process:
             time.sleep(0.1)  # ensure main process takes longest
-            open("check_main_process_first.txt", "w").close()
-        assert os.path.isfile("check_main_process_first.txt"), "Main process did not create check file yet."
+            with open(path, "a+") as f:
+                f.write("Currently in the main process\n")
+        else:
+            with open(path, "a+") as f:
+                f.write("Now on another process\n")
     accelerator.wait_for_everyone()
     if accelerator.is_main_process:
-        os.remove("check_main_process_first.txt")
+        with open(path, "r") as f:
+            text = "".join(f.readlines())
+        try:
+            assert text.startswith("Currently in the main process\n"), f"Main process was not first"
+            assert text.endswith("Now on another process\n"), f"Main process was not first"
+            assert text.count("Now on another process\n") == num_processes - 1, \
+                f"Only wrote to file {text.count('Now on another process') + 1} times, not {num_processes}"
+        except AssertionError:
+            path.unlink()
+            raise
 
     # Test the decorators
     f = io.StringIO()
--- a/src/accelerate/test_utils/scripts/test_script.py
+++ b/src/accelerate/test_utils/scripts/test_script.py
@@ -16,7 +16,6 @@
 
 import contextlib
 import io
-import os
 import time
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


muellerzr:
Hi @flukeskywalker, it did not actually. But great job identifying the bug! What concerns me is when I was writing this I tried my best to come up with a test, and clearly that test does not work. So if you feel inclined to do so, it'd be great to see perhaps what went wrong with them/fix them (see accelerate/test_utils/scripts/test_script.py), otherwise I will look into it later next week.

flukeskywalker:
Thanks @muellerzr. Indeed, the current test doesn't work because the `process_index` does not depend on whether the main process executes the code block first or not. 

We need a way to test not only that the code block is entered first in the main process, but also that it is exited in the main process before it is entered in other processes. This will ensure that there isn't a false positive when the main process happened to enter the code block first. A simple test would be to create a file if in main process and attempt to read it if not in main process. Perhaps there could be something more elegant?

muellerzr:
I think that test would work quite well for the time being!

muellerzr:
@flukeskywalker would you like to include it in this PR? Otherwise I can write one in afterwards

flukeskywalker:
Added, ready for review.

src/accelerate/accelerator.py
```diff
@@ -755,7 +755,8 @@ def main_process_first(self):
         ...     print(f"This will be printed by process {accelerator.process_index}")
         ```
         """
-        yield self.state.main_process_first()
```
lebrice, line 758: Hey there, just a thought: you also could have written this as `yield from self.state.main_process_first()`.
I personally prefer your way of doing it, it feels a bit more explicit.

muellerzr (approved):
Excellent work! (Sorry it took a few days, was on leave). 

Just a small nit/idea for the test to make it a bit more rigid. Otherwise great 🤗

src/accelerate/accelerator.py
muellerzr: Agreed, explicit is great here. Especially with something advanced such as this

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -55,10 +57,13 @@ def process_execution_check():
...
-    idxs = accelerator.gather(idx)
-    if num_processes > 1:
-        assert idxs[0] == 0, "Main process was not first."
+        if accelerator.is_main_process:
+            time.sleep(0.1)  # ensure main process takes longest
+            open("check_main_process_first.txt", "w").close()
+        assert os.path.isfile("check_main_process_first.txt"), "Main process did not create check file yet."
+    accelerator.wait_for_everyone()
+    if accelerator.is_main_process:
+        os.remove("check_main_process_first.txt")
```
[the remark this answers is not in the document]
flukeskywalker: Thanks! I like the expansion idea. I should point out a couple of issues though:
- `PartialState().main_process_first()` actually works fine in the current release. The issue is in the delegation of the context manager in `Accelerator` to that in `State` and so on.
- If we switch to `Accelerator` the test doesn't do what we want because the main process writes first due to the time delays, causing the test to succeed with the current release. This is why I had a delay in the main process _before_ creating the file and no delay in other processes, to ensure that the full block of code in the context must be executed on main process before being entered in other processes. I hope I'm thinking correctly; does that makes sense? :)

I'll try to modify the test to incorporate this idea tomorrow.

src/accelerate/accelerator.py
flukeskywalker: Hi @lebrice!

Interestingly, I think what you suggest doesn't work, because `State.main_process_first()` is not a generator but another context manager. BTW this is why `PartialState.main_process_first()` does work -- it delegates to a generator.

This is related to what I found very interesting in the issue as I mentioned above: it _feels like_ `yield` or at least `yield from` should work here, but it doesn't. I think this is why ChatGPT is confident it works :D It suggested the following example should work, which actually produces an error (you can remove the decorator from  `inner_context_manager()` to make it work).

```python

@contextmanager
def outer_context_manager():
    print("Entering outer context")
    yield
    print("Exiting outer context")

@contextmanager
def inner_context_manager():
    print("Entering inner context")
    yield
    print("Exiting inner context")

@contextmanager
def combined_context_manager():
    print("Entering combined context")
    yield from inner_context_manager()
    print("Exiting combined context")

with outer_context_manager():
    with combined_context_manager():
        print("Inside combined context")
```

src/accelerate/test_utils/scripts/test_script.py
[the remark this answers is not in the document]
flukeskywalker: Modified the test to incorporate the suggestion. Confirmed that it currently fails on 0.18.0 and passes with the fix. Should be ready for review.

muellerzr:
@flukeskywalker do `make style; make quality` :)

flukeskywalker:
Hi @muellerzr, do you think this PR is ready?

muellerzr (approved):
Yep! Thanks for the fix and iterating with it!

PR #1308, opened 2023-04-11
Simplify MPS implementation

sgugger:
This PR simplifies the MPS implementation by removing all deprecated code (we did mention it would be good until v0.18). It also paves the way for an easier integration of other default devices later on by adding a `default_device` property on `PartialState`.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -181,31 +181,8 @@ class PartialState:
                 self.num_processes = 1
                 self.process_index = self.local_process_index = 0
 
-                # the below block using env variable for `mps` will be removed in version 0.18.0
-                if parse_flag_from_env("ACCELERATE_USE_MPS_DEVICE") and not cpu:
-                    from .utils import is_torch_version
-
-                    if is_mps_available():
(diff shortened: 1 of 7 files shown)


pacman100 (approved):
Thank you @sgugger for greatly simplifying the logic of the default device, LGTM! 🤗

src/accelerate/state.py
```diff
@@ -500,6 +477,22 @@ def print(self, *args, **kwargs):
         if self.is_local_main_process:
             print(*args, **kwargs)
 
+    @property
+    def default_device(self) -> torch.device:
+        """
+        Returns the default device which is:
+        - MPS if `torch.backends.mps.is_available()` and `torch.backends.mps.is_built()` bother return True.
```
muellerzr, line 484:
```suggestion
        - MPS if `torch.backends.mps.is_available()` and `torch.backends.mps.is_built()` both return True.
```

muellerzr (approved):
Thanks for the refactoring! Looks much easier now to add other integrations as time goes on. Just one typo fix

PR #1313, opened 2023-04-13
Allow xpu backend

muellerzr:
Let's a backend be passed to the `PartialState`. For now internal only while we migrate over the `Trainer`, afterwards will be configurable through the CLI/config (and the `Accelerator` object?)


sgugger (commented):
This should be inside the corresponding KwargHandler instead.

PR #1314, opened 2023-04-14
Default to nccl

muellerzr:
Defaults the `kwargs` to `nccl`, as the kwarg handler will sometimes not be used and we still need to use `nccl` as the default


sgugger (approved):
Good catch!

PR #1315, opened 2023-04-15
offload the previous module hook before the current module is moved to…

williamberman:
… the execution device

Moving the previous module afterwards unnecessarily puts both on the device at the same time


sgugger (approved):
Thanks for the fix!

PR #1322, opened 2023-04-17
Add test flag and import check for dynamo

muellerzr:
Solves failing version check on `main` by adding a test that `torch.dynamo` is available.


sgugger (commented):
Should be named something around torch_greater_than_2.0 instead, since it doesn't test for dynamo.

sgugger (approved):
Thanks!

PR #1329, opened 2023-04-18
Verbosity, Progress Bar for Loading

xloem:
This changes adds a progress bar to `load_state_dict` and size information to the verbosity output of `infer_auto_device_map`.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -27,6 +27,7 @@ import torch.nn as nn
 
 from .imports import is_safetensors_available
 from .offload import load_offloaded_weight, offload_weight, save_offload_index
+from .tqdm import is_tqdm_available, tqdm
 
 
 if is_safetensors_available():
@@ -749,7 +750,9 @@ def infer_auto_device_map(
 
         else:
             if verbose:
-                print(f"Putting {name} on {devices[current_device]}.")
+                print(
+                    f"Putting {name} (size={module_size}) on {devices[current_device]} (available={current_max_size-current_memory_used})."
+                )
             current_memory_used += module_size
             device_map[name] = devices[current_device]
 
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -840,10 +843,23 @@ def load_state_dict(checkpoint_file, device_map=None):
             # all weights that haven't defined a device should be loaded on CPU
             device_weights["cpu"].extend([k for k in weight_names if k not in sum(device_weights.values(), [])])
             tensors = {}
+            pbar = is_tqdm_available() and tqdm(
```
sgugger, line 846: We do not ever rely on Python magic conversion og objects to bool. This should be a test `if is_tqdm_availbale`with two branches. Also `pbar` should be named `progress_bar` to be explicit.

src/accelerate/utils/modeling.py
```diff
@@ -840,10 +843,23 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+                main_process_only=False,
+                total=sum([len(device_weights[device]) for device in devices]),
+                unit="w",
+                smoothing=0,
+                leave=False,
+            )
             for device in devices:
                 with safe_open(checkpoint_file, framework="pt", device=device) as f:
                     for key in device_weights[device]:
+                        if pbar:
```
sgugger, line 856:
```suggestion
                        if progress_bar is not None:
```
then update pbar to progress_bar

src/accelerate/utils/modeling.py
```diff
@@ -840,10 +843,23 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+                            pbar.set_postfix(dev=device, refresh=False)
+                            pbar.set_description(key, refresh=False)
+                            pbar.update()
                         tensors[key] = f.get_tensor(key)
+            if pbar:
+                pbar.close()
```
sgugger, line 862: Same comments here.

sgugger (commented):
Thanks for your PR! Left a couple of comments to adapt the style so that it matches the style of the current codebase.

sgugger (approved):
Thanks for iterating!

PR #1339, opened 2023-04-19
Skip failing torch 2.0+ test

muellerzr:
This PR temporarily skips the gradient accumulation test on torch 2.0+ due to an active issue/investigation on it. The performance did not drop unreasonably and the API is still performing comparably in torch 2.0, so it may be a test-specific issue, however in terms of raw performance from our library the main implementation is fine.


sgugger (approved):
Thanks!

PR #1341, opened 2023-04-19
Small progress bar fix

xloem:
Fixes an off-by-one appearance issue at 0% and 100% with the safetensors device_map loading progress bar.

I had this tacked onto my previous progress bar contribution but lost internet before the merge happened.


sgugger (approved):
SGTM, thanks!

PR #1342, opened 2023-04-20
Pop more backend options

muellerzr:
This PR modifies the rest of the kwargs to only use `backend` when the user can set it (aka when doing just raw multi-GPU). Helps with the failing tests on transformers main.


sgugger (approved):
Works for me, thanks!

PR #1349, opened 2023-04-22
Adding support for local SGD.

searchivarius:
Hi,

As discussed with @muellerzr , I am adding a non-breaking functionality to support local SGD, which enables efficient multi-GPU training in the case when no fast interconnect is possible. It is a more stable and reliable alternative to gradient accumulation (they can be used jointly when needed too). I ran rather extensive tests to compare both. Unfortunately, even with all the proper code, gradient accumulation is quite finicky and you do not want to use it as a speed up approach.

[Here is a link to the summary of results and benchmarking code.](https://github.com/searchivarius/BlogCode/blob/master/2023/test_accelerate_followup/README.md)

It is not clear what kind of unit test can test this kind of functionality. The key ingredient (disabling gradient synchronization is being tested already).

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -25,6 +25,8 @@
     title: How perform inference on large models with small resources
   - local: usage_guides/gradient_accumulation
     title: Performing gradient accumulation
+  - local: usage_guides/local_sgd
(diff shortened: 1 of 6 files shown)


searchivarius:
Closing due to errors in the documentation building.

PR #1352, opened 2023-04-24
using deepspeed.comm for distrbiuted init

pacman100:
### What does this PR do?
1. Uses DeepSpeed comm for initializing distributed setup when using DeepSpeed. This resolves the world size issues when using accelerate PartialState in Trainer with zero stage 3 + zero_init. This PR needs to be merged before https://github.com/huggingface/transformers/pull/22921


muellerzr (approved):
Awesome! Thanks for finding this fix!

sgugger (approved):
Thanks for the fix!

PR #1355, opened 2023-04-25
[`bnb`] Fix bnb slow test

younesbelkada:
# What does this PR do?

Link to failing tests: https://github.com/huggingface/accelerate/actions/runs/4792971906

Since the latest `transformers` release, a fix has been introduced to make use of the correct behavior of `device_map="auto"` and 8bit models. 
As for now you can't train a model using the `Accelerator` + 8bit models that have been trained on multiple devices, the fix should be to force the model to be on the first GPU, so that the multi GPU test passes

I also removed a test skip since the Docker image uses the latest `transformers` and I made sure it passes

cc @muellerzr

### DIFF
--- a/tests/test_accelerator.py
+++ b/tests/test_accelerator.py
@@ -1,7 +1,6 @@
 import json
 import os
 import tempfile
-import unittest
 from unittest.mock import patch
 
 import torch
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks for the fix!

Issue #1359, opened 2023-04-26
`wait_for_everyone` with timeout


kiyoon:
### System Info

I want to make the main process download a dataset. Other processes need to wait until it is finished.
```Shell

# Pseudo code
if accelerator.is_local_main_process:
    download_dataset()

accelerator.wait_for_everyone()
```

However, it will only wait for 30 minutes and crash after that. It can take more than 30 minutes to download the dataset.  
Is there any way to add timeout to this function?

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

```python
# Pseudo code
if accelerator.is_local_main_process:
    download_dataset()

accelerator.wait_for_everyone()
```

### Expected behavior

All processes have to wait until the main process finishes downloading the dataset, even if it's longer than 30 minutes

sgugger:
Yes, you can specify the maximum timeout in `InitProcessGroupKwargs` and then pass it along when creating your `Accelerator`:
```py
from accelerate.utils import InitProcessGroupKwargs

accelerator = Accelerator(kwargs_handler=InitProcessGroupKwargs(timeout=3600))
```

kiyoon:
Thank you. I think it will work although I haven't tested.

By the way, there are lots of type errors when I use Pyright LSP. It would be nice to explicitly set type hints for this function.

![image](https://user-images.githubusercontent.com/12980409/234839793-9e9e677e-3520-4f9e-9fbf-9752cc69a07d.png)

sgugger:
As long as those type hints are not too complex, we welcome any PR.

kiyoon:
My recommendation is to use `|` instead of `Optional[]`, `Union[]` etc. However this will only work from Python 3.7. It will make the type hint much more simplified. Because of the compatibility, I wonder if you would approve this change

sgugger:
I'm happy to look at a PR with small changes. We do not support Python versions older than 3.7, so happy to see what you have in mind!

PR #1363, opened 2023-04-27
fix: typing issues, and replace deprecated python typing (Optional, Union) to `|`

kiyoon:
Fix typing issue in #1359 partially. For now, it only changes one file, but we'll need to gradually change all files to follow the new format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ If you like the simplicity of 🤗 Accelerate but would prefer a higher-level ab
 
 ## Installation
 
-This repository is tested on Python 3.6+ and PyTorch 1.4.0+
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Nice, thanks a lot! @muellerzr could you also have a look?

muellerzr (approved):
This all looks good to me! Definitely smaller PR's modifying a file or two at a time would be more feasible for us to review if you would like to. Thanks for updating it all, looks much cleaner! (So, what you have just here works great)

sgugger:
No those changes are good. It's an oversight on our side.

PR #1361, opened 2023-04-27
Update dataclasses.py

sam-hieken:
Hello,

I received the following error after trying to use the default minimum parameters option for FSDP:

```
Traceback (most recent call last):
  File "/home/hiekense/GPT2/ChatGPTDetector/train-accel.py", line 71, in <module>
    model, optimizer, train_dl, scheduler = accelerator.prepare(
  File "/home/hiekense/.local/lib/python3.9/site-packages/accelerate/accelerator.py", line 1122, in prepare
    result = tuple(
  File "/home/hiekense/.local/lib/python3.9/site-packages/accelerate/accelerator.py", line 1123, in <genexpr>
    self._prepare_one(obj, first_pass=True, device_placement=d) for obj, d in zip(args, device_placement)
  File "/home/hiekense/.local/lib/python3.9/site-packages/accelerate/accelerator.py", line 977, in _prepare_one
    return self.prepare_model(obj, device_placement=device_placement)
  File "/home/hiekense/.local/lib/python3.9/site-packages/accelerate/accelerator.py", line 1211, in prepare_model
    self.state.fsdp_plugin.set_auto_wrap_policy(model)
  File "/home/hiekense/.local/lib/python3.9/site-packages/accelerate/utils/dataclasses.py", line 846, in set_auto_wrap_policy
        min_num_params = int(os.environ.get("FSDP_MIN_NUM_PARAMS", 0))
min_num_params = int(os.environ.get("FSDP_MIN_NUM_PARAMS", 0))ValueError
ValueError: : invalid literal for int() with base 10: '100000000.0'invalid literal for int() with base 10: '100000000.0'
```

It can be solved by just wrapping the string in a float first, so I figured I'd just skip an issue and open a PR.

Thanks

PR #1367, opened 2023-04-27
Fix default FSDP_MIN_NUM_PARAMS so it's an int

sam-hieken:
FSDP_MIN_NUM_PARAMS default changed from 1e8 to 100000000 (no floats allowed), see #1361


pacman100 (approved):
Thank you for the quick fix, LGTM! 🤗

PR #1378, opened 2023-05-01
Adding support for local SGD.

searchivarius:
Hi,

As discussed with @muellerzr , I am adding a non-breaking functionality to support local SGD, which enables efficient multi-GPU training in the case when no fast interconnect is possible. 

This replaces a previous [pull request](https://github.com/huggingface/accelerate/pull/1350)

I incorporated recommendation to change the per device batch size (in the separate benchmark) and I also removed the erroneous code (borrowed from the examples) that reduced the size of the dataset (as well as code-related recommendations) and retested everything. There is still some unexplained behavior, but it looks quite reasonable now:

[Here is a link to the summary of results and benchmarking code.](https://github.com/searchivarius/BlogCode/blob/master/2023/test_accelerate_followup/README.md)
 
**Importantly** @sgugger @muellerzr I think it is possibly every **no-trainer** example [has this line](https://github.com/huggingface/transformers/blob/main/examples/pytorch/text-classification/run_glue_no_trainer.py#L448) that reduces the dataset size, but I think it's a bug. Please double check!

It is not clear what kind of unit test can test this kind of functionality. The key ingredient (disabling gradient synchronization is being tested already).

### DIFF
--- a/src/accelerate/local_sgd.py
+++ b/src/accelerate/local_sgd.py
@@ -1,4 +1,4 @@
-# Copyright 2021 The HuggingFace Team. All rights reserved.
+# Copyright 2023 The HuggingFace Team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/src/accelerate/local_sgd.py
+++ b/src/accelerate/local_sgd.py
@@ -98,7 +98,5 @@ class LocalSGD:
 
         self.accelerator.wait_for_everyone()
         with self.accelerator.autocast():
-            qty = float(dist.get_world_size())
             for param in self.model.parameters():
-                dist.all_reduce(param.data, op=torch.distributed.ReduceOp.SUM)
-                param.data /= qty
+                self.accelerator.reduce(param.data, reduction="mean")
--- a/src/accelerate/local_sgd.py
+++ b/src/accelerate/local_sgd.py
@@ -98,4 +98,4 @@ class LocalSGD:
         self.accelerator.wait_for_everyone()
         with self.accelerator.autocast():
             for param in self.model.parameters():
-                self.accelerator.reduce(param.data, reduction="mean")
+                param.data = self.accelerator.reduce(param.data, reduction="mean")
--- a/src/accelerate/local_sgd.py
+++ b/src/accelerate/local_sgd.py
@@ -24,7 +24,6 @@ class LocalSGD:
     It should be used only in the multi-GPU (or multi-CPU) setup without extensions such as DeepSpeed. In particular,
     this is a simple implementation that cannot support scenarios such as model parallelism.
 
-    Contributed by Leo (Leonid) Boytsov.
 
     Although we are not aware of the true origins of this simple approach, the idea of local SGD is quite old and goes
     back to at least:
--- a/src/accelerate/local_sgd.py
+++ b/src/accelerate/local_sgd.py
@@ -1,0 +1,104 @@
+# Copyright 2021 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import torch
+
+from accelerate import Accelerator, DistributedType
+
+
+class LocalSGD:
+    """
+    A helper class to support local SGD on top of Accelerator. It simply runs a given number of updates independently
+    on each device, and averages model weights every K synchronization step.
+
+    It should be used only in the multi-GPU (or multi-CPU) setup without extensions such as DeepSpeed. In particular,
+    this is a simple implementation that cannot support scenarios such as model parallelism.
+
+    Contributed by Leo (Leonid) Boytsov.
+
+    Although we are not aware of the true origins of this simple approach, the idea of local SGD is quite old and goes
+    back to at least:
+
+    Zhang, J., De Sa, C., Mitliagkas, I., & Ré, C. (2016). [Parallel SGD: When does averaging help?. arXiv preprint
+    arXiv:1606.07365.](https://arxiv.org/abs/1606.07365)
+
+    We credit the term Local SGD to the following paper (but there might be earlier references we are not aware of).
+
+    Stich, Sebastian Urban. ["Local SGD Converges Fast and Communicates Little." ICLR 2019-International Conference on
+    Learning Representations. No. CONF. 2019.](https://arxiv.org/abs/1805.09767)
+
+    """
+
+    def __enter__(self):
+        if self.enabled:
+            self.model_sync_obj = self.model.no_sync()
+            self.model_sync_obj.__enter__()
+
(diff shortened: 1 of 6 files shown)


tests/test_examples.py
```diff
@@ -34,6 +34,7 @@
 EXCLUDE_EXAMPLES = [
     "cross_validation.py",
     "gradient_accumulation.py",
+    "local_sgd.py",
```
muellerzr, line 37: While this is accurate, to test your script we should also include a test similar to https://github.com/huggingface/accelerate/blob/daf76a7d385b4c56e82acd653bd9e720084b6dff/tests/test_examples.py#L215-L217, to ensure your script runs fine. That would be all the tests that it would need I believe :)

src/accelerate/local_sgd.py
```diff
@@ -0,0 +1,104 @@
...
+        Synchronize + Average model parameters across all GPUs
+        """
+        import torch.distributed as dist
+
+        self.accelerator.wait_for_everyone()
+        with self.accelerator.autocast():
+            qty = float(dist.get_world_size())
+            for param in self.model.parameters():
+                dist.all_reduce(param.data, op=torch.distributed.ReduceOp.SUM)
+                param.data /= qty
```
muellerzr, line 104: All of this should be able to be replaced with:

```suggestion
            for param in self.model.parameters():
                self.accelerator.reduce(param.data)
```

src/accelerate/local_sgd.py
```diff
@@ -0,0 +1,104 @@
...
+
+class LocalSGD:
+    """
+    A helper class to support local SGD on top of Accelerator. It simply runs a given number of updates independently
+    on each device, and averages model weights every K synchronization step.
+
+    It should be used only in the multi-GPU (or multi-CPU) setup without extensions such as DeepSpeed. In particular,
+    this is a simple implementation that cannot support scenarios such as model parallelism.
+
+    Contributed by Leo (Leonid) Boytsov.
```
muellerzr, line 27:
```suggestion
```
You can put this as a comment in the code if you'd like to, but generally we rely on the GitHub for where the history of contributions lie.

src/accelerate/local_sgd.py
```diff
@@ -0,0 +1,104 @@
...
+            return
+
+        if self.num_steps % self.local_sgd_steps == 0:
+            self._sync_and_avg_model_params()
+
+    def _sync_and_avg_model_params(self):
+        """
+        Synchronize + Average model parameters across all GPUs
+        """
+        import torch.distributed as dist
```
muellerzr, line 97:
```suggestion
        
```
This should also be removed, based on the suggestion below

muellerzr (commented):
Thanks for the PR! I've made a few suggestions to simply the code, and one request re; attribution. The suggestions can be done here or I can do it in a follow up PR, whichever is easier. Thanks again!

src/accelerate/local_sgd.py
muellerzr: @searchivarius the only thing I can think of off the top of my head is:

Do you still experience this if you do:
```python
            qty = int(dist.get_world_size())
            for param in self.model.parameters():
                dist.all_reduce(param.data, op=torch.distributed.ReduceOp.SUM)
                param.data /= qty
```

(Notice `qty` is `int` instead of `float`)

muellerzr: As otherwise, the two should be the exact same.

muellerzr: @searchivarius I see what's up, you need to pass `reduction="mean"` during `accelerator.reduce`

muellerzr:
@searchivarius any particular reason for closing/reopening later rather than modifying on the PR branch and having a trail to follow back through on the PR comments?

tests/test_examples.py
searchivarius: Regarding this one, I am a bit unsure how this is invoked. I add a file `examples/by_feature/local_sgd.py`, I can add a similar function test_local_sgd(self). A couple of Qs though:
1. will it be called automatically?
2. which testargs will be used?

muellerzr: It gets called when the example tests are through our CI. `testargs` are whatever need to be used to make your script run with the feature enabled. For instance this reads as:

```bash
accelerate launch examples/by_feature/gradient_accumulation.py
```
(`accelerate launch comes from `self._launch_args`)

Yes, add a similar `test_local_sgd()` file which does this, as we make sure all of our example tests get run. to see when they are fully evoked, see the `Makefile` under `test_examples`

src/accelerate/local_sgd.py
searchivarius: Hi @muellerzr this still doesn't work this way. However, after looking more carefully at the code, I realized that the operation is done on the cloned tensor. So the suggested reduction operation does nothing. The reduced tensor needs to be reassigned back. Fortunately, this seems to be finally working:
```
-                self.accelerator.reduce(param.data, reduction="mean")
+                param.data = self.accelerator.reduce(param.data, reduction="mean")
```

searchivarius: PS: also yet another piece of evidence of how Python is really bad language that doesn't do useful type checks for you.

searchivarius:
@muellerzr I just wanted to squash commits. If this is not an issue, I can reopen the PR & make additional commits, just let me know.

muellerzr:
Not an issue at all, we squash all commits upon merge anyways

searchivarius:
PS: @muellerzr  not quite ready yet, I am still making changes and testing.

tests/test_examples.py
muellerzr: You need `test_examples`, which was further down from where you looked at... https://github.com/huggingface/accelerate/blob/main/Makefile#L45-L46

Yes, the core tests ignore the examples as they often require GPUs or take a long time..

searchivarius:
@muellerzr it should be ready, I am not sure why the quality check fails, `make quality` or 'make style` do not suggest any changes when run locally. Perhaps, there's some version mismatch.

muellerzr:
@searchivarius you should try doing `pip install -e .[quality]`, as we pin what versions we use which could be part of it? :)

searchivarius:
@muellerzr it doesn't make a difference.

muellerzr:
@searchivarius perhaps uninstall then reinstall everything in the `docs` dep? Otherwise I can fork + put a PR in to fix it. Aka:

```bash
pip uninstall black flake8 hf-doc-builder -y
```

```bash
pip install -e .[quality]
```
Note that you'll also need `urllib3` to be < 2.0.0

searchivarius:
Hi @muellerzr this didn't help: moreover, `make style` and `make quality` do not reformat the code. I suspect it's something about the version of black and ruff that is being installed. Do you know which versions do you use?

BTW, regarding urllib3:
```
>>> urllib3.__version__
'1.26.14'
```

muellerzr:
Perfect, that's the right version of urllib3.

For the other deps, try installing these versions:

```bash
pip install black==23.1 ruff==0.0.241 hf-doc-builder==0.4.0
```

searchivarius:
@muellerzr thank you, now the file is reformated properly. This is a chance that versions were not an issue, but I was rather working on a wrong branch. There's no way to know this for sure though.

muellerzr (approved):
LG2M! Thanks for all the back and forth, and hard work on this! 🤗 

cc @sgugger for final review

sgugger (approved):
Just have a couple of nits in the Copyright years, but LGTM otherwise! Thanks!

searchivarius:
@muellerzr @sgugger NP and thank you for the guidance. Do not forget to remove the division by the number of steps in examples like this one:
https://github.com/huggingface/transformers/blob/main/examples/pytorch/question-answering/run_qa_no_trainer.py#L758

PR #1379, opened 2023-05-02
Seperate out contextmanager generation

muellerzr:
This PR refactors the `autocast` capability by introducing a new `get_mixed_precision_context_manager`.

For the general user, this changes nothing. However for the `transformers` integration this allows us to fully remove the `autocast_smart_context_manager` function and let's `Accelerate` handle/generate it instead. 

See this chunk of code we're replacing:

https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L2661-L2678

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -61,6 +61,7 @@ from .utils import (
     convert_outputs_to_fp32,
     extract_model_from_parallel,
     gather,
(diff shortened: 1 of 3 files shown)


muellerzr:
@sgugger I'll do that in a follow up PR if that's alright? As there's more functions in `operations` that should go to `modeling` instead.

sgugger (approved):

PR #1380, opened 2023-05-02
Handle auto choice in Trainer for mixed precision

muellerzr:
The `Trainer` has a `self.half_precision_backend == "auto"` option, which will pick `fp16` or `bf16` based on what is available. This PR refactors `AcceleratorState` to allow for a late-init of mixed precision, so that if `auto` is chosen we can look at the `AcceleratorState` which was generated to pick what device precision we should use (`bf16` if CPU or `fp16` if GPU).

Another option is for us to support `"auto"` directly in Accelerate instead, let me know if you'd prefer this @sgugger

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -580,10 +580,9 @@ class AcceleratorState:
                     self.ipex_plugin = ipex_plugin if ipex_plugin.use_ipex else None
             self._init_mixed_precision(mixed_precision=mixed_precision)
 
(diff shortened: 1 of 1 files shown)


sgugger (approved):
LGTM, make sure to run all slow tests though!

muellerzr:
Status of slow tests: [Self-hosted runner with slow tests (scheduled)](https://github.com/huggingface/accelerate/actions/workflows/nightly.yml)

PR #1385, opened 2023-05-03
Support TPU v2 and v3 on new PyTorch/XLA TPU runtime

will-cromar:
Depends on #1393

I've been working on migrating PyTorch/XLA from our legacy XRT runtime to PJRT. We have detailed documentation on the differences and changes here: https://github.com/pytorch/xla/blob/master/docs/pjrt.md

Due to TPU design constraints, PJRT must use multithreading on TPU v2 and v3 ([docs](https://github.com/pytorch/xla/blob/master/docs/pjrt.md#multithreading-on-tpu-v2v3)). 

- This means we cannot use global state for `AcceleratorState`, since two replicas would end up sharing the same `accelerator.device` and `accelerator.process_index`. I implemented a [descriptor](https://docs.python.org/3/howto/descriptor.html) that wraps the global `_shared_state` dicts in `state.py` and optionally makes it thread-local when using XLA.
- Use numpy `Generator`s in the `accelerate test` script instead of the global `numpy.random.seed`. This makes the test thread-safe for TPU v2 and v3. Numpy recommends using `Generators` as a best practice in their docs: https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html

Tested:

- `accelerate test` on TPU v2-8 with XRT and PJRT

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import threading
 import warnings
 from contextlib import contextmanager
 from functools import partial
@@ -54,6 +55,21 @@ def do_nothing(*args, **kwargs):
     return None
 
 
+class GlobalSharedDict:
+  def __init__(self, thread_local: bool = False):
+    self._storage = {}
+
+  def __get__(self, obj, objtype=None):
+    return self._storage
+
+  def __set__(self, obj, value):
+    self._storage = value
+
+class ThreadLocalSharedDict(GlobalSharedDict, threading.local):
+  pass
+
+SharedDict = GlobalSharedDict if not is_tpu_available(check_device=False) else ThreadLocalSharedDict
+
 # Inspired by Alex Martelli's 'Borg'.
 class PartialState:
     """
@@ -75,8 +91,7 @@ class PartialState:
         - **is_main_process** (`bool`) -- Whether or not the current process is the main one.
         - **is_local_main_process** (`bool`) -- Whether or not the current process is the main one on the local node.
     """
-
-    _shared_state = {}
+    _shared_state = SharedDict()
 
     def __init__(self, cpu: bool = False, **kwargs):
         self.__dict__ = self._shared_state
@@ -211,7 +226,7 @@ class PartialState:
     @staticmethod
     def _reset_state():
         "Resets `_shared_state`, is used internally and should not be called"
-        PartialState._shared_state = {}
+        PartialState._shared_state.clear()
 
     @property
     def initialized(self) -> bool:
@@ -527,8 +542,7 @@ class AcceleratorState:
         - **is_main_process** (`bool`) -- Whether or not the current process is the main one.
         - **is_local_main_process** (`bool`) -- Whether or not the current process is the main one on the local node.
     """
-
-    _shared_state = {}
+    _shared_state = SharedDict()
 
     def __init__(
         self,
@@ -652,7 +666,7 @@ class AcceleratorState:
     @staticmethod
     def _reset_state(reset_partial_state: bool = False):
         "Resets `_shared_state`, is used internally and should not be called"
-        AcceleratorState._shared_state = {}
+        AcceleratorState._shared_state.clear()
         if reset_partial_state:
             PartialState._reset_state()
 
@@ -721,8 +735,7 @@ class GradientState:
         - **adjust_scheduler** (`bool`) -- Whether the scheduler should be adjusted to account for the gradient
           accumulation
     """
-
-    _shared_state = {}
+    _shared_state = SharedDict()
 
     def __init__(self, gradient_accumulation_plugin: Optional[GradientAccumulationPlugin] = None):
         self.__dict__ = self._shared_state
@@ -793,4 +806,4 @@ class GradientState:
     @staticmethod
     def _reset_state():
         "Resets `_shared_state`, is used internally and should not be called"
-        GradientState._shared_state = {}
+        GradientState._shared_state.clear()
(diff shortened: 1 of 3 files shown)


JackCaoG:
@muellerzr We are trying to enable the Accelerate with new TPUVM and the PJRT runtime. Would someone in your team be able to review this change? We are working on deprecating the old XRT runtime and TPU node architure, so it is important for us to keep the hf support going forward.

src/accelerate/state.py
```diff
@@ -54,6 +55,21 @@ def do_nothing(*args, **kwargs):
...
+  def __get__(self, obj, objtype=None):
+    return self._storage
+
+  def __set__(self, obj, value):
+    self._storage = value
+
+class ThreadLocalSharedDict(GlobalSharedDict, threading.local):
+  pass
+
+SharedDict = GlobalSharedDict if not is_tpu_available(check_device=False) else ThreadLocalSharedDict
```
sgugger, line 71: Let's not break the existing ang keep using a regular dict when not on TPU.

sgugger (commented):
Thanks for opening this PR! If possible let's have it focused on TPUs only and leave the changes non-related to that for other PRs, so it's easier to inspect potential regressions with git blames in the future.

Note that we use four spaces as indentation and not two.

From what I understand in your comments, a user will need to have PyTorch 2.0+ to use TPUs after this is merged? So let's have a proper error raised if we detect an earlier version telling the user to upgrade, what do you think?

will-cromar (commented):
Thanks for taking a look!

I can break this down into 3 reviews:

1. Suppress `FileNotFound` when deleting the test file. This is trivial and not directly related to XLA.
2. Changes to get Accelerate working on TPU v4 and XLA:GPU (which don't use multithreading). These are straightforward refactors of XLA-specific code.
3. Changes to support TPU v2 and v3 with multithreading. These change some common code paths and are more likely to need a rollback.

Unless I missed something, I don't think I added a dependency on anything introduced in PT/XLA 2.0. You can switch to PJRT by setting `PJRT_DEVICE=TPU` in the environment, which will have no effect before PJRT was added. I did add a call to `xm.collective_broadcast`, which is in the 1.13 release branch.

Let me know if this works for you. I can re-request your review and take this PR out of draft when I'm done shuffling commits and fixing code style issues.

will-cromar:
(1) is already fixed on `main` (thanks!), so dropping that commit here.

Sent the commits to make TPU v4 work (2) in #1393. We can review the last set of changes (for TPU v2 and v3) in this PR since we've already started the conversation.

I don't think GitHub has a nice way for me to make this PR diff against #1393, since my changes are coming from a fork. The relevant changes here are in `training.py`, `test_script.py`, and `state.py`.

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -280,7 +287,7 @@ def central_dl_preparation_check():
 def mock_training(length, batch_size, generator):
     set_seed(42)
     generator.manual_seed(42)
-    train_set = RegressionDataset(length=length)
+    train_set = RegressionDataset(length=length, seed=42)
```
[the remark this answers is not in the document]
will-cromar: See https://github.com/huggingface/accelerate/pull/1385#discussion_r1186414407

Each thread needs to make a `Generator` with a common seed, rather than using the global random state.

src/accelerate/state.py
```diff
@@ -211,7 +226,7 @@ def __repr__(self) -> str:
     @staticmethod
     def _reset_state():
         "Resets `_shared_state`, is used internally and should not be called"
-        PartialState._shared_state = {}
+        PartialState._shared_state.clear()
```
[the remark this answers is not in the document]
sgugger: My bad.

src/accelerate/state.py
sgugger: Leaving a dict and calling clear seems easier to me, as long as it works. We don't have to worry about the state being multithreaded usually (as it's just launched on different processes) so I believe this is the first time it's coming up.

will-cromar: Done.

will-cromar:
Rebased after #1393 and re-tested with XRT and PJRT on TPU v2-8.

src/accelerate/test_utils/training.py
```diff
@@ -21,11 +21,10 @@
...
 class RegressionDataset:
     def __init__(self, a=2, b=3, length=64, seed=None):
-        if seed is not None:
-            np.random.seed(seed)
+        rng = np.random.default_rng(seed)
         self.length = length
-        self.x = np.random.normal(size=(length,)).astype(np.float32)
-        self.y = a * self.x + b + np.random.normal(scale=0.1, size=(length,)).astype(np.float32)
+        self.x = rng.normal(size=(length,)).astype(np.float32)
+        self.y = a * self.x + b + rng.normal(scale=0.1, size=(length,)).astype(np.float32)
```
muellerzr, line 27: Alternative suggestion here, what if we modified `set_seed` to give it a `default_rng` bool flag, and when `True` will call `np.random.default_rng` instead and return the object.

Then, `RegressionDataset` checks if `seed` is an `int` or a generator, and then modifies accordingly towards what we want to do. 

E.g. something like:

```python
rng = np.random
if seed is not None:
    if isinstance(seed, int):
        np.random.seed(seed)
    else:
        rng = seed
```

wdyt @sgugger? This way we maintain the original behavior, and allow for this when needed (aka, only TPUs)

sgugger: That doesn't quite work as `set_seed` sets multiple seeds. Why would it return a Numpy RNG and not a torch rng?

sgugger (approved):
Thanks for iterating on this, LGTM!

PR #1387, opened 2023-05-04
Fix flakey thread issue

muellerzr:
Should fix the thread issue with `pathlib.unlink()`, moves the removal under the first check that passes the conditional and waits for everyone, and has the main process only remove it. Going to rerun the CI multiple times to ensure it's no longer flaky

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ from setuptools import setup
 from setuptools import find_packages
 
 extras = {}
-extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
 from setuptools import find_packages
 
 extras = {}
-extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0"]
+extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0", "urllib3 <= 2.0.0"]
```
sgugger, line 19:
```suggestion
extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0", "urllib3 < 2.0.0"]
```
2.0.0 was yanked but technically we don't want it either.

sgugger (approved):

Issue #1383, opened 2023-05-02
[Feature Request] gather non-tensor objects


Hannibal046:
Considering that PyTorch has already  support non-tensor objects. Can we just make accelerate to support this? This is so important to calculate metrics for NLP tasks since we don't want to decoder(encode(labels)).

https://pytorch.org/docs/stable/_modules/torch/distributed/distributed_c10d.html#all_gather_object

relevant issue: https://github.com/huggingface/accelerate/issues/476

muellerzr:
This already exists, please use `accelerate.utils.gather_objects`

Hannibal046:
Oh, i just found it. Thanks ! 

BTW, would you mind checking this issue https://github.com/huggingface/accelerate/issues/1382, it confuse me for a while.

Hannibal046:
Hi, there seems to have some bugs about `gather_object`

For tensors we gather:
```python
## CUDA_VISIBLE_DEVICES=0,1 accelerate run --num_processes 2 test.py
import torch
from accelerate import Accelerator

accelerator = Accelerator()
dataloader = torch.utils.data.DataLoader(range(9), batch_size=5)
dataloader = accelerator.prepare(dataloader)
batch = next(iter(dataloader))
gathered_items = accelerator.gather_for_metrics(batch)
if accelerator.is_local_main_process:
    print(gathered_items)
```
This would output:
```
tensor([0, 1, 2, 3, 4, 5, 6, 7, 8], device='cuda:0')
```

For objects we gather:
```python
## CUDA_VISIBLE_DEVICES=0,1 accelerate run --num_processes 2 test.py
import torch
from accelerate import Accelerator
from accelerate.utils import gather_object

accelerator = Accelerator()
obj = [accelerator.local_process_index]
obj = gather_object(obj)
if accelerator.is_local_main_process:
    print(obj)
```
this would only return:
```
[0]
```

I using `torch.distributed` to rewrite this, and it works fine:
```python
from accelerate import Accelerator
import torch.distributed as dist

accelerator = Accelerator()
obj = [accelerator.local_process_index]
all_rank_outputs = [None for _ in range(accelerator.num_processes)]    
dist.all_gather_object(all_rank_outputs,obj) ## would create new proce
obj = [x for y in all_rank_outputs for x in y]
if accelerator.is_local_main_process:
    print(obj)
```
returns 
```
[0,1]
```

muellerzr:
[...]

PR #1391, opened 2023-05-05
Fix gather_obj

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1383

Currently `gather_object` doesn't actually work properly, because pytorch will gather the whole object as a whole. As a result, using the nested execution that is currently there will force pytorch to not actually return what the user wants. This PR fixes the implementation, based on the suggestion in #1383, and adds tests for it as well.

I also noticed that occasionally just doing `torchrun` wouldn't force the tests to be ran with multiple GPUs/all gpus available, so this PR adds `--nproc_per_node` manually

### DIFF
--- a/src/accelerate/test_utils/scripts/test_ops.py
+++ b/src/accelerate/test_utils/scripts/test_ops.py
@@ -17,6 +17,7 @@
 import torch
 
 from accelerate import PartialState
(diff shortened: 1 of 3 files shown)


sgugger (approved):
Thanks for fixing!

PR #1392, opened 2023-05-05
Bump torch version

muellerzr:
Per https://github.com/huggingface/accelerate/pull/638, it never actually bumped the minimal torch version to 1.6.0, which is the minimum we test on now and support.


sgugger (approved):
Good catch!

PR #1393, opened 2023-05-05
Support TPU v4 with new PyTorch/XLA TPU runtime

will-cromar:
I've been working on migrating PyTorch/XLA from our legacy XRT runtime to PJRT. We have detailed documentation on the differences and changes here: https://github.com/pytorch/xla/blob/master/docs/pjrt.md

This PR collects all of the fixes that I've made so far to support XRT and PJRT interchangeably through Accelerate. In order of significance:

1. Update implementation of `synchronize_rng_types` to use `xm.collective_broadcast` to broadcast the RNG state tensor. Collective operations in general should be called from the main thread of each process to avoid unpredictable behaviors in XLA. But, our implementation of [`MpDeviceLoader` calls `accelerate.DataLoaderShard`'s `__iter__`](https://github.com/pytorch/xla/blob/f763d0d2a4dac0872e2e1670332fd490f524b8be/torch_xla/distributed/parallel_loader.py#L138) (which [synchronizes the RNG](https://github.com/huggingface/accelerate/blob/75a693c9b47635074a72b716af14bb55b0033750/src/accelerate/data_loader.py#L368-L369)) in all of the preloading threads. To ensure that the RNG is synchronized once from the main thread, call it in the `MpDeviceLoaderWrapper`'s `__iter__` instead.
2. In general, you should call `xm.mark_step` to finish any remaining steps before checkpointing. `MpDeviceLoader` is responsible for calling `xm.mark_step` at the beginning of each [new step and at the end of the dataset iterator](https://github.com/pytorch/xla/blob/f763d0d2a4dac0872e2e1670332fd490f524b8be/torch_xla/distributed/parallel_loader.py#L40-L48). If you checkpoint in the middle of iteration, replica 0 won't reach the `mark_step` at the beginning of the next iteration before it tries to checkpoint. To avoid making the user call `mark_step` themselves, call it for them on replica 0 before checkpointing in `save_state`.
3. Use XLA's `all_gather` in `gather` instead of `mesh_reduce` + `torch.cat`. With PJRT, `rendezvous` and `mesh_reduce` both use XLA collective ops to broadcast pickled data ([docs](https://github.com/pytorch/xla/blob/master/docs/pjrt.md#changes-to-xmrendezvous)) and move it back to the CPU. Batching all of the recursive `all_gather` calls together and calling `xm.mark_step` once reduces the number of transfers between host and device.
4. Don't override `XLA_USE_BF16` when `mixed_precision` is not set. This isn't really mixed precision as much as it is a mechanism to silently convert _all_ `torch.floats` to BF16 on the TPU. Real mixed precision support in PT/XLA is still a WIP, and we can update back here when it's stable.
5. Remove obsolete/unused `_mp_fn` in test script.

Tested:

- `accelerate test` on v4-8 with XRT and PJRT
- [`diffusers` Stable Diffusion fine-tuning example](https://github.com/huggingface/diffusers/blob/main/examples/text_to_image/README.md) on TPU v4-8

Accelerate will not work on TPU v2 and v3 with this PR, because both of them use use multithreading due to TPU design constraints. I'll follow up with the remaining fixes for TPU v2 and v3 in #1385.

cc @sgugger @jackcaog

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -116,6 +116,7 @@ if is_torch_version(">", "1.10.0"):
 
 
 if is_tpu_available(check_device=False):
+    import torch_xla.core.xla_model as xm
     import torch_xla.distributed.xla_multiprocessing as xmp
 
 
@@ -2293,6 +2294,10 @@ class Accelerator:
                 )
         os.makedirs(output_dir, exist_ok=True)
         logger.info(f"Saving current state to {output_dir}")
+
+        if self.distributed_type == DistributedType.TPU:
+            # Finish running the previous step before checkpointing
+            xm.mark_step()
 
         # Save the models taking care of FSDP and DeepSpeed nuances
         weights = []
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -36,31 +36,6 @@ from .utils import (
 )
 
 
-if is_tpu_available(check_device=False):
-    import torch_xla.distributed.parallel_loader as xpl
-
-    class MpDeviceLoaderWrapper(xpl.MpDeviceLoader):
-        """
-        Wrapper for the xpl.MpDeviceLoader class that knows the total batch size.
-
-        **Available attributes:**
-
-        - **total_batch_size** (`int`) -- Total batch size of the dataloader across all processes.
-            Equal to the original batch size when `split_batches=True`; otherwise the original batch size * the total
-            number of processes
-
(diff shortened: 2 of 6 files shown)


sgugger (approved):
Thanks for adding support for TPU v4! All the changes LGTM! Can you just run `make style` on your branch so that the quality check passes?

sgugger:
Thanks again!

JackCaoG:
Thanks both, this is super exciting!

PR #1403, opened 2023-05-10
use existing mlflow experiment if exists

Rusteam:
If mlflow experiment already exists, we can re-use its id for tracking. (otherwise it will fail)


muellerzr (approved):
Thanks! Makes sense to me!

PR #1406, opened 2023-05-10
changes required for DS integration

pacman100:
### What does this PR do?
1. Changes to be made for integrating accelerate in Transformers

PR #1411, opened 2023-05-11
fix deepspeed failing tests

pacman100:
### What does this PR do?
1. Fixes https://github.com/huggingface/accelerate/pull/1406#issuecomment-1542848943
2. Changing the default of `zero_force_ds_cpu_optimizer` to True
3. Changing the max lr logic of the scheduler  in ds config


pacman100:
Checked offline that all tests pass

muellerzr (approved):
Thanks for fixing!

PR #1413, opened 2023-05-11
Make mlflow logging dir optional

mattplo-decath:
If we use an already existing experiment, having to specify a logging directory doesn't make sense.
If we create a new experiment, the logging dir is optional too and is picked automatically by mlflow if not provided.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -558,13 +558,13 @@ class MLflowTracker(GeneralTracker):
     """
 
     name = "mlflow"
-    requires_logging_directory = True
+    requires_logging_directory = False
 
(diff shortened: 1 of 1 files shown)


muellerzr:
This should be fixed via https://github.com/huggingface/accelerate/pull/1403, correct me if I'm wrong. Try this by doing `pip install git+https://github.com/huggingface/accelerate`

mattplo-decath:
#1403 doesn't address this issue. The logging_dir argument should not be mandatory for the mlflow tracker (both if we use an already existing experiment or create a new one)

muellerzr (approved):
Makes sense then! Thanks!

Issue #1420, opened 2023-05-12
[Casting] RuntimeError: Found dtype Float but expected BFloat16 on a CPU


sayakpaul:
## Issue

When I run the following (on CPUs) using the from-source installation of `accelerate`: 

```bash
accelerate launch --config_file default_config.yml examples/unconditional_image_generation/train_unconditional.py \
    --dataset_name hf-internal-testing/dummy_image_class_data \
    --model_config_name_or_path diffusers/ddpm_dummy \
    --resolution 64 \
    --output_dir /temp \
    --train_batch_size 2 \
    --num_epochs 1 \
    --gradient_accumulation_steps 1 \
    --ddpm_num_inference_steps 2 \
    --learning_rate 1e-3 \
    --lr_warmup_steps 5
```

it throws:

```bash
Epoch 0:   0%|                                                                                                          | 0/3 [00:00<?, ?it/s]Traceback (most recent call last):
  File "examples/unconditional_image_generation/train_unconditional.py", line 692, in <module>
    main(args)
  File "examples/unconditional_image_generation/train_unconditional.py", line 594, in main
    accelerator.backward(loss)
  File "/opt/venv/lib/python3.8/site-packages/accelerate/accelerator.py", line 1761, in backward
    loss.backward(**kwargs)
  File "/opt/venv/lib/python3.8/site-packages/torch/_tensor.py", line 487, in backward
    torch.autograd.backward(
  File "/opt/venv/lib/python3.8/site-packages/torch/autograd/__init__.py", line 200, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: Found dtype Float but expected BFloat16
Epoch 0:   0%|                                                                                                          | 0/3 [00:04<?, ?it/s]
Traceback (most recent call last):
  File "/opt/venv/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/opt/venv/lib/python3.8/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/opt/venv/lib/python3.8/site-packages/accelerate/commands/launch.py", line 928, in launch_command
    simple_launcher(args)
  File "/opt/venv/lib/python3.8/site-packages/accelerate/commands/launch.py", line 588, in simple_launcher
    raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
[...]

PR #1426, opened 2023-05-12
Fix bug on ipex for diffusers

abhilash1910:
Fix for https://github.com/huggingface/diffusers/issues/3411 and #1420 owing to ipex changes for cpu.  (prelims seem ok )
cc for additional tests: @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1137,7 +1137,9 @@ class Accelerator:
             old_named_params = self._get_named_parameters(*args)
 
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
-            if self.device.type in ["cpu", "xpu"]:
+            if self.device.type == "cpu" and self.state.ipex_plugin is not None:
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks! Solution looks correct to me :)

src/accelerate/accelerator.py
```diff
@@ -1137,7 +1137,9 @@ def prepare(self, *args, device_placement=None):
             old_named_params = self._get_named_parameters(*args)
 
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
-            if self.device.type in ["cpu", "xpu"]:
+            if self.device.type == "cpu" and self.state.ipex_plugin is not None:
+                args = self._prepare_ipex(*args)
+            elif self.device.type == "xpu" and self.state.ipex_plugin is not None and is_xpu_available():
                 args = self._prepare_ipex(*args)
```
jgong5, line 1143: It might be helpful to add assertions in `self._prepare_ipex` for these checks too.

PR #1434, opened 2023-05-15
[`core`] Introducing `CustomDtype` enum for custom dtypes

younesbelkada:
# What does this PR do?

This PR adds a new utility class, termed as `CustomDtype` to handle custom dtypes that does not exists natively in PyTorch for future users that want to use `infer_auto_device_map` on exotic dtypes, such as FP8, int4, etc.

cc @sgugger

### DIFF
--- a/src/accelerate/utils/__init__.py
+++ b/src/accelerate/utils/__init__.py
@@ -46,6 +46,7 @@ from .imports import (
     is_xpu_available,
 )
 from .modeling import (
+    CustomDtype,
     check_device_map,
     compute_module_sizes,
     convert_file_size_to_int,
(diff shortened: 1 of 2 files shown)


sgugger (commented):
It's good to have something to deal with those dtypes indeed. I'm just wondering how we can convert existing weights in those dtypes to this custom enum. We probably will need a class that regroups regular torch.dtype and this enum, with a `from_weight` method that can do the dispatching depending on the type of the weight.

younesbelkada:
I think for some exotic dtype, the conversion should be handled by external libraries (e.g. `bitsandbytes`) as for example int4 or any 4bit precision dtypes are not supported natively by `torch` (maybe I am wrong here?) - I see the main purpose of this util is just to compute correctly the theoretical device map when calling `infer_auto_device_map` for a model that uses exotic dtypes

sgugger (approved):

PR #1437, opened 2023-05-16
Make sure torch compiled model can also be unwrapped

patrickvonplaten:
When torch compile is used in addition to multi-GPU training the PyTorch module is wrapped twice and looks more or less as follows:
```
model: torch._dynamo.eval_frame.OptimizedModule(torch.nn.parallel.DistributedDataParallel(model))
```

The model is not unwrapped from distributed data parallel to just model. This PR adds more logic to `unwrap_model()` so that calling:

```py
accelerator.unwrap_model(model)
```

results in:
```
model: torch._dynamo.eval_frame.OptimizedModule(model)
```

which I think is a better fit here

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -19,6 +19,7 @@ import torch
 
 from ..commands.config.default import write_basic_config  # noqa: F401
 from ..state import PartialState
+from .versions import is_torch_version
 from .dataclasses import DistributedType
 from .imports import is_deepspeed_available, is_tpu_available
 from .transformer_engine import convert_model
@@ -29,6 +30,13 @@ if is_deepspeed_available():
 
 if is_tpu_available(check_device=False):
     import torch_xla.core.xla_model as xm
+
+
+def is_compiled_module(module):
+    """Check whether the module was compiled with torch.compile()"""
+    if is_torch_version("<", "2.0.0") or not hasattr(torch, "_dynamo"):
+        return False
+    return isinstance(module, torch._dynamo.eval_frame.OptimizedModule)
 
 
 def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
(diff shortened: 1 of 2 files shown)


patrickvonplaten:
This issue came up as part of this `diffusers` issue: https://github.com/huggingface/diffusers/issues/3403

Happy to add tests and improve doc string if ok for you @sgugger @muellerzr

src/accelerate/utils/other.py
```diff
@@ -31,6 +32,13 @@
     import torch_xla.core.xla_model as xm
 
 
+def is_compiled_module(module):
+    """Check whether the module was compiled with torch.compile()"""
```
muellerzr, line 36: Mild nit, space out the `"""` to be under
```suggestion
    """
    Check whether the module was compiled with torch.compile()
    """
```

muellerzr (approved):
Solution makes sense to me, @sgugger for another check to make sure it looks fine for you as well.

We should also add tests for this, the relevant tests are in `tests/test_utils.py`

src/accelerate/utils/other.py
```diff
@@ -31,6 +32,13 @@
...
+    if is_torch_version("<", "2.0.0") or not hasattr(torch, "_dynamo"):
+        return False
+    return isinstance(module, torch._dynamo.eval_frame.OptimizedModule)
```
sgugger, line 39: I'm guessing we don't have a public class we could use? This is going to get ugly if PyTorch decides to move this around.

sgugger (approved):
Thanks a  lot for adding this! Sounds great to me!

src/accelerate/utils/other.py
patrickvonplaten: Agree, we had the same question in `diffusers`. Will also ask PyTorch about this

patrickvonplaten: https://huggingface.slack.com/archives/C047QHKH944/p1684250821270959

patrickvonplaten: Backwards comp will be kept, so think this is good for now. Merging.

PR #1440, opened 2023-05-16
Fix XPU being used in default config

muellerzr:
Currently we have tests failing on transformers due to XPU, due to unsupported operations being performed. This PR makes it so that `write_basic_config` has an XPU flag, and sets it to `False` to not interfere with the plan CPU runners in transformers (also removes unused `dynamo_backend` option)


sgugger (approved):
Thanks for fixing!

PR #1444, opened 2023-05-17
fix potential OOM when resuming with multi-GPU training

exhyy:
When resuming from a checkpoint, only the model will be loaded to different GPUs, which might cause OOM.

Background: 
I encounter an OOM error when attempting to resume training ControlNet from a checkpoint using 4 GPUs 12GB each. After making modifications to the code, this error (OOM) has been fixed.


sgugger (commented):
This only applies to the optimizer state. Scheduler, grad scaler or RNG states do not need this.

Also cc @muellerzr

exhyy:
Thanks for replay! Code updated and it works fine for me!

sgugger (approved):
Thanks! Waiting for a second look from @muellerzr and then should be good to merge.

muellerzr (approved):
Makes sense to me! I did wonder at some point if the optimizer state needed to be mapped as well, nice catch :)

Issue #1445, opened 2023-05-17
CPU never used when MPS is available, if logger is used beforehand


antonioalegria:
### System Info

```Shell
- `Accelerate` version: 0.19.0
- Platform: macOS-13.3.1-arm64-arm-64bit
- Python version: 3.11.3
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1 (False)
- System RAM: 32.00 GB
- `Accelerate` default config:
	- compute_environment: LOCAL_MACHINE
	- distributed_type: NO
	- mixed_precision: no
	- use_cpu: False
	- num_processes: 1
	- machine_rank: 0
	- num_machines: 1
	- rdzv_backend: static
	- same_network: True
	- main_training_function: main
	- downcast_bf16: no
	- tpu_use_cluster: False
	- tpu_use_sudo: False
	- tpu_env: []
	- dynamo_config: {'dynamo_backend': 'INDUCTOR'}
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Logger used beforehand

Device is set to MPS, although we set `cpu` to `True`.

```python
from accelerate import Accelerator
from accelerate.logging import get_logger

logger = get_logger(__name__)
logger.warning(f"Starting test")
accelerator = Accelerator(cpu=True)
print(f"Device: {accelerator.state.device}")
```

### Output:

```
Starting test
Device: mps
```

---

### Logger NOT used beforehand

Device is set to CPU, as expected.

```python
from accelerate import Accelerator
from accelerate.logging import get_logger

logger = get_logger(__name__)
# logger.warning(f"Starting test")
accelerator = Accelerator(cpu=True)
print(f"Device: {accelerator.state.device}")
```

### Output:

```
Device: cpu
```

### Expected behavior

I would expect that whenever cpu=True then device should always be CPU. Also, if I pass a certain device it should use that device. I don't see why using the logger beforehand would have any impact on that.

muellerzr:
The accelerate logger needs to know about your distributed setup when it begins logging, as a result when doing `log()` it spins up an `AcceleratorState`, forcing it to be whatever is in there. For you to do what you would like, the `Accelerator` must be initialized *first* then the logger.

[...]

PR #1446, opened 2023-05-17
Raise error when logging improperly

muellerzr:
A user cannot call the logging utility in `Accelerate` without first building the state in some form. This PR introduces an error raising it as such and fixes https://github.com/huggingface/accelerate/issues/1445


sgugger (approved):
Thanks for adding this!

PR #1449, opened 2023-05-17
[docs] Minor typo in distributed inference

stevhliu:
I think `split_across_processes` should be [`split_between_processes`](https://huggingface.co/docs/accelerate/main/en/package_reference/state#accelerate.PartialState.split_between_processes) right?


muellerzr (approved):
Thanks!

PR #1450, opened 2023-05-18
[docs] Replace `state.rank` -> `process_index`

pcuenca:
I couldn't find a rank property in `PartialState`.


muellerzr (approved):
Thanks 🤦

PR #1452, opened 2023-05-18
Auto multigpu logic

muellerzr:
It was reported with an unexpected behavior (both on the forums [here](https://discuss.huggingface.co/t/num-processes-1-even-when-i-set-it-to-num-processes-2/40122) and in slack) where if a user specified `--num_processes` but didn't specify multi-gpu, multiple GPUs wouldn't be used by default.

This PR modifies the default checks when doing `accelerate launch` to see if `num_processes>1` and if multiple GPUs are available, specifies multi-gpu with a warning. cc @pcuenca


sgugger (approved):
Nice catch!

PR #1453, opened 2023-05-18
Update with cli instructions

muellerzr:
Updates the distributed inference docs to a: show how to use `accelerate launch`, and b: update to include the change in #1452


sgugger (approved):
Thanks for adding!

PR #1456, opened 2023-05-19
Refactor and simplify xpu device in state

abhilash1910:
Simplify device selection in state for xpus & cpus  (also avoid certain boundary cases for xpu)
cc @muellerzr

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -708,13 +708,12 @@ class AcceleratorState:
                     self.distributed_type = DistributedType.MEGATRON_LM
                     megatron_lm_plugin.set_mixed_precision(self._mixed_precision)
                     self.megatron_lm_plugin = megatron_lm_plugin
-            elif self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.NO]:
+            elif self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
                 if self.device.type == "cpu" and ipex_plugin is not None:
                     self.ipex_plugin = ipex_plugin
                     if self.ipex_plugin is not None:
                         self.ipex_plugin.set_mixed_precision(mixed_precision)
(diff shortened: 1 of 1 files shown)


sgugger (commented):
LGTM, but would love to have @muellerzr have a second look.
Thanks!

src/accelerate/state.py
```diff
@@ -708,13 +708,12 @@ def __init__(
...
                 if self.device.type == "cpu" and ipex_plugin is not None:
                     self.ipex_plugin = ipex_plugin
                     if self.ipex_plugin is not None:
                         self.ipex_plugin.set_mixed_precision(mixed_precision)
-            if self.distributed_type in [DistributedType.MULTI_XPU, DistributedType.NO]:
-                if self.device.type == "xpu" and ipex_plugin is not None:
+                elif self.device.type == "xpu" and ipex_plugin is not None and is_xpu_available():
                     self.ipex_plugin = ipex_plugin
                     if self.ipex_plugin is not None:
                         self.ipex_plugin.set_mixed_precision(mixed_precision)
```
muellerzr, line 719: This could all be simplified further, refactoring all of L711->719 to be:

```python
elif self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
    if (
        self.device.type == "cpu" 
        or (self.device.type == "xpu" and is_xpu_available())
        and ipex_plugin is not None
    ):
        self.ipex_plugin = ipex_plugin
        self.ipex_plugin.set_mixed_precision(mixed_precision)
```

muellerzr (commented):
Thanks! We can refactor this further to remove duplicate logic

src/accelerate/state.py
abhilash1910: Yes had thought of doing so 😄

muellerzr (approved):
Thanks for the cleanup!

PR #1457, opened 2023-05-19
Document how to use commands with python module instead of argparse

muellerzr:
Documents how to use `python -m` when running launch instead of using the CLI. cc @stas00 

Also documents the other available module-calling option: `env`


sgugger (approved):
Thanks for adding this!

PR #1458, opened 2023-05-19
4-bit QLoRA via bitsandbytes (4-bit base model + LoRA)

TimDettmers:
This adds QLoRA support to accelerate. More information about QLoRA from our abstract:

>We develop QLoRA tuning, a method that finetunes by backpropagating gradients through a frozen 4-bit base model into low rank adapters (LoRA). With QLoRA tuning we can finetune 30B/65B parameter models on 24/48GB GPUs while preserving regular 16-bit full finetuning runtime and task performance. We achieve the memory efficiency and quantization precision through a combination of new methods: nested quantization to reduce the average memory footprint from 4.5 to 4.1 bits per parameter, paged optimizers to manage gradient checkpointing memory spikes, and a new data type, 4-bit NormalFloat (NF4), which is information theoretically and empirically optimal for normally distributed weights. To demonstrate the effectiveness and ease of use of QLoRA tuning we finetune more than 1,000 models to create a detailed dissection of instruction following performance across datasets (FLAN, Alpaca, Chip2, SuperNatural Instructions, Chip2, AnthropicHH), models types (LLaMA, T5), and model scales (125M to 65B). A discussion of the results is forthcoming in our paper.

Also see the matching PR for PEFT: https://github.com/huggingface/peft/pull/476

@sgugger @younesbelkada


sgugger (commented):
Is it only the new version of bitsandbytes that will support creating INT8Params from a torch.flaot32 or has it been around a bit more?

src/accelerate/utils/modeling.py
```diff
@@ -172,10 +172,7 @@ def set_module_tensor_to_device(
         elif value is not None or torch.device(device) != module._parameters[tensor_name].device:
             param_cls = type(module._parameters[tensor_name])
             kwargs = module._parameters[tensor_name].__dict__
-            if param_cls.__name__ == "Int8Params":
-                # downcast to fp16 if any
-                if new_value.dtype == torch.float32:
-                    new_value = new_value.to(torch.float16)
```
younesbelkada, line 178: I think these changes were needed for `Int8Params` for the serialization - can we add a check to add these only for `Int8Params`

younesbelkada (commented):
Thanks for the addition! I left one comment
also cc @SunMarc for visibility!

younesbelkada (approved):
Thanks a lot Tim ! 
let me know if the proposed changes looks fine, I did that to keep the backward compatibility I belive with 8bit serialization (I suspect your fork was not in sync with `main`)

src/accelerate/utils/modeling.py
TimDettmers: This looks great to me for backward compatibility!

TimDettmers:
This all looks great to me. Thanks for thinking about backward compatibility, @younesbelkada!

PR #1465, opened 2023-05-22
Update `register_empty_buffer` to match torch args

NouamaneTazi:
Fixes the following case
```python
import torch
from torch import nn
from accelerate import init_on_device
with init_on_device(device="meta", include_buffers=True):
    dummy_module = nn.Module()
    dummy_module.register_buffer("dummy_buffer", torch.tensor([1.0]), persistent=False)
```

cc @thomasw21


thomasw21 (approved):
Thanks! LGTM!

sgugger (approved):
Thanks for adding this!

PR #1468, opened 2023-05-23
Add AssertionError in accelerate.prepare() using Deepspeed.

tensimiku:
Dear contributers and maintainers.

- Include your OS type and version, the versions of Python and PyTorch.
```
> uname -a
Linux arch-cglab-2 6.2.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 30 Mar 2023 14:51:14 +0000 x86_64 GNU/Linux

> pip list
accelerate         0.20.0.dev0
...
deepspeed          0.9.2
...
torch              2.0.1+cu118
```

- A short, self-contained, code snippet that allows us to reproduce the bug in less than 30s;

config
```
deepspeed_config:
  gradient_clipping: 1.0
  zero3_init_flag: false
  zero_stage: 0
distributed_type: DEEPSPEED
downcast_bf16: 'no'
machine_rank: 0
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 1
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```
code
```
import torch
from torch import nn
from torch.utils.data import TensorDataset


def main():
    accelerator = Accelerator(log_with="all")

    class net0(nn.Module):
        def __init__(self):
            super().__init__()
            self.model = torch.nn.Sequential(
                torch.nn.Linear(3, 16),
                torch.nn.Linear(16, 3)
            )
        
        def forward(self, x):
            return self.model(x)
    
    class net1(nn.Module):
        def __init__(self):
            super().__init__()
            self.model = torch.nn.Sequential(
                torch.nn.Linear(3, 32),
                torch.nn.Linear(32, 3)
            )
        
        def forward(self, x):
            return self.model(x)


    model0 = net0()
    model1 = net1()


    dset = TensorDataset(torch.rand([40, 3]))
    train_dataloader = torch.utils.data.DataLoader(dset, batch_size=4, shuffle=True, num_workers=4, pin_memory=True, drop_last=True, persistent_workers=True)

    optimizer = torch.optim.Adam(model0.parameters(), lr=0.01)
    lr_scheduler = torch.optim.lr_scheduler.OneCycleLR(optimizer, 0.05, epochs=10, steps_per_epoch=len(train_dataloader))

    optimizer_1 = torch.optim.Adam(model1.parameters(), lr=0.01)
    lr_scheduler_1 = torch.optim.lr_scheduler.OneCycleLR(optimizer_1, 0.05, epochs=10, steps_per_epoch=len(train_dataloader))

    print("---model class", type(model0), type(model1))

    model0, model1, optimizer, optimizer_1, train_dataloader, lr_scheduler, lr_scheduler_1 = accelerator.prepare(
        model0, model1, optimizer, optimizer_1, train_dataloader, lr_scheduler, lr_scheduler_1
    )

    print("---model class", type(model0.module), type(model1.module))
    assert type(model0.module) !=  type(model1.module), "should assert!"

    for epoch in range(10):
        model0.train()
        model1.train()
        for step, batch in enumerate(train_dataloader):
            batch = batch[0]
            outputs = model0(batch)
            loss = torch.nn.functional.mse_loss(batch, outputs)
            accelerator.backward(loss)
            optimizer.step()
            lr_scheduler.step()
            optimizer.zero_grad()

            outputs = model1(batch)
            loss = torch.nn.functional.mse_loss(batch, outputs)
            accelerator.backward(loss)
            optimizer_1.step()
            lr_scheduler_1.step()
            optimizer_1.zero_grad()
        
if __name__ == "__main__":
    main()
```

If i use more than one nn.module in accelerate.prepare(), it should assert(like when i call accelerate.prepare() for each module)but it copies second module to first one.
(type(model0) or type(model0.module) should be net0 but it becomes net1 after prepare.)

I think accelerate.prepare MEGATRON_LM```DistributedType.MEGATRON_LM``` also has similar issue. but i'm not sure.

Thank you.


sgugger:
cc @pacman100

Issue #788, opened 2022-10-25
Script hangs on `set_seed` and `accelerator.backward(loss)` with A100 GPU


ProGamerGov:
### System Info

```Shell
- `Accelerate` version: 0.12.0
- Platform: Linux-4.19.0-21-cloud-amd64-x86_64-with-debian-10.13
- Python version: 3.7.12
- Numpy version: 1.21.6
- PyTorch version (GPU?): 1.12.1 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - main_process_ip: None
        - main_process_port: None
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - downcast_bf16: no
```

I also tested using `Accelerate` version: 0.13.2 and can confirm the issue is still present.

```
OS: Debian GNU/Linux 10 (buster) (x86_64)
GCC version: (Debian 8.3.0-6) 8.3.0
Clang version: Could not collect
CMake version: version 3.24.1
Libc version: glibc-2.10

Python version: 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 06:08:53)  [GCC 9.4.0] (64-bit runtime)
Python platform: Linux-4.19.0-20-cloud-amd64-x86_64-with-debian-10.13
Is CUDA available: True
CUDA runtime version: 11.3.109
CUDA_MODULE_LOADING set to: 
GPU models and configuration: GPU 0: NVIDIA A100-SXM4-40GB
Nvidia driver version: 470.57.02
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
```

I'm using the `Debian 10 based Deep Learning VM for PyTorch CPU/GPU with CUDA 11.3 M98` image on Google Cloud Compute.

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```
pip install git+https://github.com/ShivamShrirao/diffusers.git
[...]

PR #1472, opened 2023-05-23
*Critical Fix* Check for xpu specifically and disable when using GPUs

muellerzr:
Should solve https://github.com/huggingface/accelerate/issues/788 (*finally*) by disabling the use of torch-xla when on a GPU. While they have GPU support currently it will break amp series GPUs, so as a safety precaution until it becomes more stable we are disabling it entirely unless they are running on a TPU.

Will be testing in a moment on some A100's as well.


sgugger (approved):
You meant TPU in the title right? Thanks for the fix!

muellerzr:
@sgugger nope, GPUs. XLA has GPU support, but it breaks when using amp series GPUs. We want to explicitly disable it when this situation happens

Issue #1400, opened 2023-05-09
Issue when using a dataloader on MULTI_GPU


ArmelRandy:
### System Info

```Shell
- `Accelerate` version: 0.18.0
- Platform: Linux-5.15.0-1023-aws-x86_64-with-glibc2.31
- Python version: 3.10.11
- Numpy version: 1.24.3
- PyTorch version (GPU?): 1.13.1 (False)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: 0, 1, 2, 3
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am using accelerate to speed up the inference on a set of prompts. I sequentially receive batches of prompts, that I turn into a torch IterableDataset.  From that dataset I build a dataloader that I prepare as well as the model I am using. I am trying to run the inference by going through the prepared dataloader. As a matter of fact, I expect accelerate to dispatch my prompts through all my GPUs but I end up having an issue.

The problem arises in with the following code 

```python
from transformers import AutoTokenizer, AutoModelForCausalLM, pipeline
from accelerate import Accelerator

import tqdm
import torch
from torch.utils.data import IterableDataset
from torch.utils.data.dataloader import DataLoader
import os


class TokenizedDataset(IterableDataset):
    """Tokenize and preprocess the dataset, where the dataset is a list of instructions (str)
    """
    def __init__(self, tokenizer, dataset):
        self.tokenizer = tokenizer
        self.dataset = dataset
        self.outputs = self.tokenizer(self.dataset, padding=True, return_tensors="pt")
    def __iter__(self):
        for i in range(len(self.dataset)):
            yield {
                "input_ids" : self.outputs.input_ids[i],
                "attention_mask" : self.outputs.attention_mask[i],
                "prompt" : self.dataset[i]
            }



[...]

PR #1477, opened 2023-05-26
Split tensors as part of `split_between_processes`

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1400 by also letting the tensors be split if added. Note that similar to the list version, the tensors padded will have the value of the last item in the tensor being used. Such that on four processes:

```python
t = torch.tensor([[0,1,2,3], [4,5,6,7]])
with PartialState().split_between_processes(t) as t:
    print(t)
# Process 1:
tensor([[0,1,2,3]])
# Process 2:
tensor([[4,5,6,7]])
# Process 3:
tensor([[4,5,6,7]])
# Process 4:
tensor([[4,5,6,7]])
```


sgugger (approved):
LGTM, can you add a test of this use case as well?

PR #1483, opened 2023-05-28
Fix gradient state bugs in multiple dataloader

Ethan-yt:
1. Add `DataLoaderStateMixin`: Each dataloader should have it own states like `end_of_dataloader` and `remainder`
2. Changed `_remove_dataloader`: This should be called after last `yield`. Otherwise the active dataloader will be wrong in the last iteration.
3. Changed test: active_dataloader should be the current dataloader whether is the last iteration or not.

Solves #1050

@muellerzr

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -300,13 +300,16 @@ class IterableDatasetShard(IterableDataset):
 
 class DataLoaderStateMixin:
     """
-    Mixin class that adds a state to a `DataLoader` to keep track of the status inside the dataloader such as at the end of the iteration, the number of items in the dataset in the last batch relative to the batch size, and other useful information that might be needed.
-    
+    Mixin class that adds a state to a `DataLoader` to keep track of the status inside the dataloader such as at the
+    end of the iteration, the number of items in the dataset in the last batch relative to the batch size, and other
+    useful information that might be needed.
+
     **Available attributes:**
-    
+
         - **end_of_dataloader** (`bool`) -- Whether at the last iteration or batch
-        - **remainder** (`int`) -- The number of items that are remaining in the last batch, relative to the total batch size
-    
+        - **remainder** (`int`) -- The number of items that are remaining in the last batch, relative to the total
+          batch size
+
     """
 
     def __init_subclass__(cls, **kwargs):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -298,7 +298,18 @@ class IterableDatasetShard(IterableDataset):
                 yield current_batch[i]
 
 
-class DataLoaderShard(DataLoader):
+class DataLoaderStateMixin:
+    """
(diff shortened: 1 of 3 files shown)


src/accelerate/data_loader.py
```diff
@@ -298,7 +298,18 @@ def __iter__(self):
                 yield current_batch[i]
 
 
-class DataLoaderShard(DataLoader):
+class DataLoaderStateMixin:
+    """
+    Mixin class that adds state to a `DataLoader` to keep track of the current state such as is end of dataloader or
+    not and the remainder of the current batch, etc.
```
muellerzr, line 304:
```suggestion
    Mixin class that adds a state to a `DataLoader` to keep track of the status inside the dataloader such as at the end of the iteration, the number of items in the dataset in the last batch relative to the batch size, and other useful information that might be needed.
    
    **Available attributes:**
    
        - **end_of_dataloader** (`bool`) -- Whether at the last iteration or batch
        - **remainder** (`int`) -- The number of items that are remaining in the last batch, relative to the total batch size
    
```

src/accelerate/state.py
```diff
@@ -930,6 +928,18 @@ def initialized(self) -> bool:
...
         return GradientState._shared_state != {}
 
+    @property
+    def end_of_dataloader(self) -> bool:
+        if not self.in_dataloader:
+            return False
+        return self.active_dataloader.end_of_dataloader
+
+    @property
+    def remainder(self) -> int:
```
muellerzr, line 938: As well as here

muellerzr (approved):
Excellent! This is a much better way to handle the `DataLoader` situation. Left a few comments mostly about documentation prose to make sure everything is clear and concise for the readers. 

We also need to add `DataLoaderStateMixin` to the docs at `docs/source/package_reference/torch_wrappers` (With the other `data_loader` docs)

src/accelerate/state.py
```diff
@@ -930,6 +928,18 @@ def initialized(self) -> bool:
         "Returns whether the `GradientState` has been initialized"
         return GradientState._shared_state != {}
 
+    @property
+    def end_of_dataloader(self) -> bool:
```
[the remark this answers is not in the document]
Ethan-yt: Done

muellerzr:
Thanks again!

muellerzr:
Will try to look at this tommorow, but FYI @Ethan-yt this caused a failing test on main 😢 

If you have access to multi-gpu system, it's:

```bash
CUDA_VISIBLE_DEVICES="0,1" pytest -sv tests/test_accelerator.py::AcceleratorTester::test_save_load_model
```

Ethan-yt:
@muellerzr 
image

muellerzr:
Thanks, yep I see it passing in our nighties now too suddenly. Must have been something weird yesterday with the runner (or something we fixed today?) Thanks for verifying!

Issue #1394, opened 2023-05-06
5x faster text generation on multi-GPU setups (+ lower VRAM consumption)


emvw7yf:
TL;DR: the patch below makes multi-GPU inference 5x faster.

I noticed that text-generation is significantly slower on multi-GPU vs. single-GPU. Some results (using llama models and utilizing the full 2048 context window, I also tested with GPT-J and the results are similar):

| model size     | quantization | GPU configuration | seconds/token        | total peak VRAM usage  |
|--------------|-------------|--------------------|--------------------|--------------------|
| 7B                   | 16                 | 1x4090                    | 0.025                        | 14996 MiB              |
| 7B                   | 16                 | 2x4090                    | 0.22  (8x slower)     | 16484 MiB              |
| 7B w/patch    | 16                 | 2x4090                    | 0.026                        | 15891 MiB               |
| 13B                 | 8 w/bnb       | 1x4090                    | 0.067                        | 16076 MiB              |
| 13B                 | 8 w/bnb       | 2x4090                    | 0.34  (5x slower)     | 18087 MiB              |
| 13B w/patch  | 8 w/bnb        | 2x4090                    | 0.070                        | 16883 MiB              |
| 13B                 | 4 w/gptq      | 1x4090                    | 0.036                        | 10219 MB               |
| 13B                 | 4 w/gptq      | 2x4090                    | 0.228 (6x slower)    | 11091 MB               |
| 13B w/patch  | 4 w/gptq       | 2x4090                   | 0.038                        | 10336 MB               |

This makes using large models prohibitively slow: running 65b model on 4x3090 results in more than 4 seconds/token, which is quite unusable for interactive applications.

After some profiling and debugging, I narrowed this down to a simple problem:
1. accelerate wraps the `LlamaModel` and moves the HUGE `past_key_values` tensor to the `execution_device` of `LlamaModel` (which is GPU 0) at the beginning of the inference
2. it also wraps each `LlamaDecoderLayer` and move the same `past_key_values` (which stays constant during the entire inference pass) from GPU 0 to the execution device for each layer — repeatedly moving it between GPUs for every layer that is not on GPU 0.

This unnecessary repeated moving consumes up to 85% of the inference time. Furthermore, because it makes a copy of the `past_key_values` on each GPU, it significantly increases VRAM usage (although I didn't measure the exact number).

[...]

PR #1491, opened 2023-05-30
Allow key skipping in big model inference

sgugger:
This PR adds an option to skip keys when moving inputs/outputs during the big model inference. The main motivation is #1394


muellerzr (approved):
Nice! Implementation makes sense to me :)

PR #1492, opened 2023-05-31
adjust overriding of model's forward function

prathikr:
During ONNX Runtime optimization, the forward function's signature is inspected to perform downstream tasks. The current way of overriding the model's forward function results in the following error:

`AttributeError: 'function' object has no attribute '__func__'`

This PR improves the function signature override by using `types.MethodType` which doesn't impact the function signature for downstream tasks.


sgugger (approved):
LGTM but would love to have @muellerzr have a look as well.

Edit: looks like this breaks in mixed precision settings.

muellerzr (commented):
The solution presented here makes sense, but let's make sure tests pass before we can merge please :) (And then I'll give the ✅ 😉 )

sgugger (approved):

PR #1493, opened 2023-05-31
add torchkeras reference

lyhue1991:
Add torchkeras reference. Maybe someone will like this cute tool.


sgugger (approved):
Thanks for adding!

PR #1495, opened 2023-05-31
Only use IPEX if available

muellerzr:
Currently will fail with the transformers integration because IPEX is *always* used due to the presence of the plugin, when it should only exist when being used/actually available. 

<s>In a follow-up PR I will change this to use an env variable as well, as the "create a plugin by default" breaks our norm</s>

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -322,11 +322,12 @@ class Accelerator:
             if not is_megatron_lm_available():
                 raise ImportError("Megatron is not installed. please build it from source.")
 
-        if ipex_plugin is None and is_ipex_available():  # init from env variables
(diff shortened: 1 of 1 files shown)


sgugger (approved):
No to the follow-up PR however. We use CUDA when it's there by default. We use the MPS device when it's there by default. It should be the same for the specific IPEX device (but only `if is_ipex_available()` as you just did).

Note that I tried to have the Intel folks remove the plugin as I don't think it's necessary, that would be a nice followup PR if you want :-)

Thanks for the fix!

abhilash1910 (commented):
The plugin will be used only when there is ipex in the system (for either CPU or XPU) . Thanks @muellerzr  for the fix in this

PR #1497, opened 2023-05-31
Use empty like when we only need to create buffers

thomasw21:


muellerzr (approved):
Thanks! That should save on ram/time some I imagine :) 

cc @sgugger

sgugger (approved):
Thanks for the fix!

PR #1502, opened 2023-06-01
fix crash when ipex is installed and torch has no xpu

sywangyi:
also when cpu flag is set. should use cpu instead of XPU


sywangyi:
@sgugger

sywangyi:
@abhilash1910  @muellerzr please help review

abhilash1910 (approved):
LGTM . cc @muellerzr for a second look

sgugger (approved):
Thanks for the fix! @muellerzr I'll let you mere if you're happy with the changes.

PR #1503, opened 2023-06-01
fix the issue that ipex is always imported if it's installed in python env.

sywangyi:
even we answer "NO" to "Do you want to use Intel PyTorch Extension (IPEX) to speed up training on CPU?" in accelerate config

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -49,7 +49,6 @@ from .utils import (
     GradientAccumulationPlugin,
     GradScalerKwargs,
     InitProcessGroupKwargs,
-    IntelPyTorchExtensionPlugin,
     KwargsHandler,
     LoggerType,
     MegatronLMPlugin,
@@ -165,9 +164,6 @@ class Accelerator:
         megatron_lm_plugin (`MegatronLMPlugin`, *optional*):
             Tweak your MegatronLM related args using this argument. This argument is optional and can be configured
             directly using *accelerate config*
-        ipex_plugin (`IntelExtensionPlugin`, *optional*):
-            Tweak your Intel Extension for PyTorch related args using this argument for CPU and XPU. This argument is
-            optional and can be configured directly using *accelerate config*
         rng_types (list of `str` or [`~utils.RNGType`]):
             The list of random number generators to synchronize at the beginning of each iteration in your prepared
             dataloaders. Should be one or several of:
@@ -239,7 +235,6 @@ class Accelerator:
         deepspeed_plugin: DeepSpeedPlugin | None = None,
         fsdp_plugin: FullyShardedDataParallelPlugin | None = None,
         megatron_lm_plugin: MegatronLMPlugin | None = None,
-        ipex_plugin: IntelPyTorchExtensionPlugin | None = None,
         rng_types: list[str | RNGType] | None = None,
         log_with: str | LoggerType | GeneralTracker | list[str | LoggerType | GeneralTracker] | None = None,
         project_dir: str | os.PathLike | None = None,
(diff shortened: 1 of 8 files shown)


sywangyi:
@abhilash1910 @muellerzr please help review

src/accelerate/accelerator.py
```diff
@@ -324,10 +324,13 @@ def __init__(
 
         if is_ipex_available():
             if ipex_plugin is None:  # init from env variables
-                ipex_plugin = IntelPyTorchExtensionPlugin()
+                ipex_plugin = (
+                    IntelPyTorchExtensionPlugin() if os.environ.get("IPEX_ENABLED", "false") == "true" else None
+                )
```
muellerzr, line 329: Considering ipex is a device and not really a plugin like megatron, we want it to create the plugin to automatically pick up on using xpu. cc @sgugger to verify.

sgugger: yes, by default it should be used if available unless the user explicitly said no.

sywangyi: agree, I change logic like ipex is used unless the user explicitly said no by export IPEX_ENABLED or set it in accelerate config

sgugger (approved):
Thanks!

src/accelerate/accelerator.py
```diff
@@ -324,10 +324,13 @@ def __init__(
 
         if is_ipex_available():
             if ipex_plugin is None:  # init from env variables
-                ipex_plugin = IntelPyTorchExtensionPlugin()
+                ipex_plugin = (
+                    IntelPyTorchExtensionPlugin() if os.environ.get("IPEX_ENABLED", "true") == "true" else None
+                )
```
muellerzr, line 329: Still not entirely sure I agree with this.

Is it possible to have `is_ipex_available` check if the device supports IPEX specifically, not just if it's installed?

Because we *should* pick up `ipex` by default if it's usable and available on the system, that is how all of accelerate's device configurations work. cc @abhilash1910 

(So basically if this PR is needed, we have other issues I think)

muellerzr: If we want to do this, we should store a "disable IPEX" or something similar in the `PartialState` that gets set/read during `accelerate launch` and stored in the config file. 

In a follow-up PR I'll also adjust the env vars to act similar to Accelerate's

sywangyi: @muellerzr , do you mean remove the env check to is is_ipex_available()?

muellerzr: Yes, so this ipex plugin checking and creation should happen in the state, not in the `Accelerator` object. The `Accelerator` object should only handle the case of if the user wants to pass their own one in. And if a plugin is passed in and ipex is *not* enabled explicitly, we need to raise an error stating this and not just enable it magically on the backend. 

If needed, I can put in a PR today also changing this. Let me know if so, perfectly fine doing that.

sywangyi: Hi, @muellerzr I get your point, do you think we should remove ipex_plugin input in Accelerator object and control ipex plugin in Accelerate state. the accelerate state only has ipex plugin if there's ipex installed and users do not disable it explicitly? I so, I could help to change like this and modify testcase as well.

sgugger: In general the Ipex plugin is there with just one flag... There is no need for a plugin for just one boolean flag, we can have the flag in the state :-)

muellerzr: Agreed w/ Sylvains note, let's move forward with that @sywangyi, you can keep it in this PR if you'd like or do another one :) Just let us know when you're ready to review it

sywangyi: I will update the PR. remove ipex plugin. only add enable flag in state

sywangyi:
@mingxiaoh

sywangyi:
@muellerzr @sgugger  I have updated the PR according to the discussion, please help review it.

muellerzr (approved):
Thanks! cc @sgugger

PR #1506, opened 2023-06-01
Fix 4bit model on multiple devices

SunMarc:
This small PR fixes the issue when the user prepare a 4-bit model on multiple devices. It will return an error just like for the 8-bit model. Furthermore, i have fixed how we get the device index as the device_map values can be of type torch.device.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1243,14 +1243,15 @@ class Accelerator:
                 has_hf_device_map = True
                 break
 
-        if getattr(model, "is_loaded_in_8bit", False) and getattr(model, "hf_device_map", False):
+        if (getattr(model, "is_loaded_in_8bit", False) or getattr(model, "is_loaded_in_4bit", False)) and getattr(model, "hf_device_map", False):
             model_devices = set(model.hf_device_map.values())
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1243,14 +1243,17 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
                 has_hf_device_map = True
                 break
 
-        if getattr(model, "is_loaded_in_8bit", False) and getattr(model, "hf_device_map", False):
+        if (getattr(model, "is_loaded_in_8bit", False) or getattr(model, "is_loaded_in_4bit", False)) and getattr(
```
sgugger, line 1246: Isn't there an `is_quantized` attribute that groups those two? @younesbelkada

sgugger (approved):
Nice, just one question on the test.

src/accelerate/accelerator.py
younesbelkada: yes there is, but since it has been introduced for transformers >= 4.30.0 maybe better to keep that logic for backward compatibility

younesbelkada (approved):
Thanks a lot for your work on this!

PR #1508, opened 2023-06-02
fix the bug in is_xpu_available

mingxiaoh:
only when import intel_extension_for_pytorch will torch has xpu attribute.


abhilash1910 (approved):
@muellerzr  could you please review once

abhilash1910:
@mingxiaoh  could you run  ```make style ``` on your branch to pass quality tests

src/accelerate/utils/imports.py
```diff
@@ -219,6 +219,7 @@ def is_xpu_available(check_device=False):
     else:
         return False
 
+    import intel_extension_for_pytorch
```
sgugger, line 222: This needs to be in a try/except block.

sgugger (commented):
Thanks for the fix.

src/accelerate/utils/imports.py
mingxiaoh: thanks for the review, I just finished the format error, as for try/block, I think we have no need to add a try block for "import  intel_extension_for_pytorch", because in there is check in "is_ipex_available()", if is_ipex_available is true, "import intel_extension_for_pytorch" will always be okay.

mingxiaoh: > This import is not under that test/

what does test/ you mean? if you meant I should put it into the check block, I don't think so, when check=False(actually, in this repo, I don't see any caller set it as True), you will go to  line 231 "hasattr(torch, "xpu") and torch.xpu.is_available()", if you don't  import intel_extension_for_pytorch, line 231 will always return False, which is unexpected.

sgugger: The import `import intel_extension_for_pytorch` is not under anything testing that the library is actually available. So this will make the code break for anyone who doesn't have it installed.

mingxiaoh: > The import `import intel_extension_for_pytorch` is not under anything testing that the library is actually available. So this will make the code break for anyone who doesn't have it installed.
thanks for the comment. But, 
in line 217(if is_ipex_available()), it will check whether intel_extension_for_pytorch is installed or not, if it is not installed, it will go to line 223, that is "return False" directly.

sgugger: Oh, sorry. I don't know how I missed that else.

src/accelerate/commands/launch.py
```diff
@@ -812,7 +813,9 @@ def _validate_launch_command(args):
             and not args.use_megatron_lm
         ):
             args.use_deepspeed = defaults.distributed_type == DistributedType.DEEPSPEED
-            args.multi_gpu = defaults.distributed_type == DistributedType.MULTI_GPU
+            args.multi_gpu = (
+                True if defaults.distributed_type in (DistributedType.MULTI_GPU, DistributedType.MULTI_XPU) else False
+            )
```
muellerzr, line 818:
```suggestion
            args.multi_gpu = defaults.distributed_type in (DistributedType.MULTI_GPU, DistributedType.MULTI_XPU)
```

We can clean this up a bit to keep in line with the rest of the logic

src/accelerate/commands/launch.py
```diff
@@ -869,8 +872,16 @@ def _validate_launch_command(args):
     else:
         if args.num_processes is None:
             args.num_processes = torch.cuda.device_count()
+            if args.use_xpu and is_xpu_available():
+                args.num_processes = torch.xpu.device_count()
```
muellerzr, line 876: I think I'd rather have an `if/else` here:

```suggestion
            if args.use_xpu and is_xpu_available():
                args.num_processes = torch.xpu.device_count()
            else:
                args.num_processes = torch.cuda.device_count()
```

src/accelerate/commands/launch.py
```diff
@@ -869,8 +872,16 @@ def _validate_launch_command(args):
...
+        if args.use_xpu and is_xpu_available() and torch.xpu.device_count() > 1 and not args.multi_gpu:
+            warned.append(
+                "\t\tMore than one GPU was found, enabling multi-GPU training.\n"
+                "\t\tIf this was unintended please pass in `--num_processes=1`."
+            )
+            args.multi_gpu = True
+        elif torch.cuda.device_count() > 1 and not args.multi_gpu:
             warned.append(
                 "\t\tMore than one GPU was found, enabling multi-GPU training.\n"
                 "\t\tIf this was unintended please pass in `--num_processes=1`."
```
muellerzr, line 887:
```suggestion
        if not args.multi_gpu and ((args.use_xpu and is_xpu_available() and torch.xpu.device_count() > 1) or (torch.cuda.device_count() > 1)):
            warned.append(
                "\t\tMore than one GPU was found, enabling multi-GPU training.\n"
                "\t\tIf this was unintended please pass in `--num_processes=1`."
            )
            args.multi_gpu = True
```

We should just change the if statement here. (Note: style will probably fail with this, be sure to run `make style; make quality;`)

src/accelerate/state.py
```diff
@@ -240,7 +240,12 @@ def __init__(self, cpu: bool = False, **kwargs):
                 self.process_index = torch.distributed.get_rank()
                 self.local_process_index = local_rank
                 if self.device is None:
-                    self.device = torch.device("cpu") if cpu else self.default_device
+                    if not cpu and is_xpu_available():
+                        self.device = torch.device("xpu", self.local_process_index)
+                        if self.device is not None:
+                            torch.xpu.set_device(self.device)
+                    else:
+                        self.device = torch.device("cpu") if cpu else self.default_device
```
muellerzr, line 248:
```suggestion
                    if cpu:
                        self.device = torch.device("cpu")
                    elif is_xpu_available():
                        self.device = torch.device("xpu", self.local_process_index)
                        if self.device is not None:
                            torch.xpu.set_device(self.device)
                    else:
                        self.device = self.default_device
```

Also, is it a real possibility `torch.device("xpu", self.local_process_index)` could be `None`? If so, that seems very breakable and something we should raise an error for instead? Otherwise, we should just always call `set_device`

muellerzr (approved):
Thanks, left a number of comments and some refactoring code suggestions so we can keep the codebase easily readable.

muellerzr (commented):
(Also meant to be a comment, not approval until changes are discussed)

src/accelerate/commands/launch.py
mingxiaoh: thanks a lot, good comment, I just modified it.

src/accelerate/state.py
mingxiaoh: thanks a lot, I just cleaned and refactored it according to your comment.

src/accelerate/commands/launch.py
mingxiaoh: suggestion makes sense, I just modified it, thanks.

mingxiaoh:
> (Also meant to be a comment, not approval until changes are discussed)

@muellerzr  @sgugger  I just push a new commit to avoid the bug exposed in accelerate due to limit in CCL. may you please help double review it? Thanks.

src/accelerate/state.py
```diff
@@ -238,9 +241,13 @@ def __init__(self, cpu: bool = False, **kwargs):
                     torch.distributed.init_process_group(self.backend, rank=rank, world_size=size, **kwargs)
                 self.num_processes = torch.distributed.get_world_size()
                 self.process_index = torch.distributed.get_rank()
-                self.local_process_index = local_rank
```
sgugger, line 241: This shouldn't be removed, no?

sgugger (approved):

sgugger:
Thanks for iterating!

src/accelerate/state.py
mingxiaoh: > This shouldn't be removed, no?

thanks for review, I forgot to mention that I removed it is because it is deprecated in line 225, https://github.com/huggingface/accelerate/pull/1508/files#diff-e54d1f9cdd342420827fcabd4280eceb68da266c9ca6df120d0432c6f6f2e10aR225

mingxiaoh:
@muellerzr @sgugger  if change is okay, may you please help merge it? Thanks for your time.

mingxiaoh:
@muellerzr is there any concern regarding to this PR?

muellerzr (approved):
Thanks for iterating!

PR #1511, opened 2023-06-02
should set correct dtype to ipex optimize and use amp logic in native…

sywangyi:
…_amp logic in prepare_model

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1643,6 +1643,8 @@ class Accelerator:
                 "Trying to use IPEX but IPEX is not installed or IPEX's version does not match current PyTorch, please refer"
                 " to https://github.com/intel/intel-extension-for-pytorch."
             )
+            result = [obj for obj in args]
+            return tuple(result)
         else:
             import intel_extension_for_pytorch as ipex
 
@@ -1655,19 +1657,16 @@ class Accelerator:
             elif isinstance(obj, (torch.optim.Optimizer)):
                 optimizer = obj
         if optimizer is not None and model is not None:
-            if is_ipex_available():
-                if is_xpu_available() and self.device.type == "xpu":
-                    model = model.to(self.device)
-                    model, optimizer = torch.xpu.optimize(model, optimizer=optimizer, inplace=True, level="O1")
-                    model.forward = torch.xpu.amp.autocast()(model.forward)
-                else:
-                    model, optimizer = ipex.optimize(model, optimizer=optimizer, inplace=True, level="O1")
-                    model.forward = torch.cpu.amp.autocast()(model.forward)
+            dtype = self.state.ipex_plugin.dtype
+            if dtype == "no":
+                dtype = torch.float32
+            if is_xpu_available() and self.device.type == "xpu":
+                model = model.to(self.device)
+                model, optimizer = torch.xpu.optimize(
+                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
+                )
             else:
-                if is_torch_version(">=", "1.10"):
-                    model.forward = torch.autocast(self.device.type)(model.forward)
-                else:
-                    model.forward = convert_outputs_to_fp32(model.forward)
+                model, optimizer = ipex.optimize(model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1")
         for i in range(len(result)):
             if isinstance(result[i], torch.nn.Module):
                 result[i] = model
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -1339,5 +1339,4 @@ class IntelPyTorchExtensionPlugin:
     def set_mixed_precision(self, mixed_precision):
         if mixed_precision == "fp16":
             raise ValueError("Tried to use `fp16` but it is not supported on cpu or xpu")
-        elif mixed_precision == "bf16":
-            self.dtype = torch.bfloat16
+        self.dtype = mixed_precision


sywangyi:
@sgugger @muellerzr @pacman100  please help review

sywangyi:
@abhilash1910 @mingxiaoh @yao-matrix

src/accelerate/accelerator.py
```diff
@@ -1643,6 +1643,8 @@ def _prepare_ipex(self, *args):
                 "Trying to use IPEX but IPEX is not installed or IPEX's version does not match current PyTorch, please refer"
                 " to https://github.com/intel/intel-extension-for-pytorch."
             )
+            result = [obj for obj in args]
+            return tuple(result)
```
abhilash1910, line 1647: Can compress directly:
return tuple(list(args))

abhilash1910 (approved):
@sywangyi  thanks for this fix. A small modification and it is good.

src/accelerate/accelerator.py
sgugger: Could you explain why not return `args` (which is a tuple) directly?

src/accelerate/accelerator.py
```diff
@@ -1655,19 +1657,16 @@ def _prepare_ipex(self, *args):
...
-                if is_xpu_available() and self.device.type == "xpu":
-                    model = model.to(self.device)
-                    model, optimizer = torch.xpu.optimize(model, optimizer=optimizer, inplace=True, level="O1")
-                    model.forward = torch.xpu.amp.autocast()(model.forward)
-                else:
-                    model, optimizer = ipex.optimize(model, optimizer=optimizer, inplace=True, level="O1")
-                    model.forward = torch.cpu.amp.autocast()(model.forward)
+            dtype = self.state.ipex_plugin.dtype
+            if dtype == "no":
+                dtype = torch.float32
```
sgugger, line 1662: The `dtype` should be the one of the Accelerator state. We shouldn't have two different dtypes as the user will not necessarily keep them in sync and this will result in subtle bugs.

src/accelerate/utils/dataclasses.py
```diff
@@ -1339,5 +1339,4 @@ class IntelPyTorchExtensionPlugin:
     def set_mixed_precision(self, mixed_precision):
         if mixed_precision == "fp16":
             raise ValueError("Tried to use `fp16` but it is not supported on cpu or xpu")
-        elif mixed_precision == "bf16":
-            self.dtype = torch.bfloat16
+        self.dtype = mixed_precision
```
sgugger, line 1342: Said this before on other PRs, this `dtype` should not be used (and should be removed ideally).

sgugger (commented):
Thanks for your PR!

src/accelerate/accelerator.py
abhilash1910: yes that can be done too :)

abhilash1910 (approved):

sgugger (approved):
Thanks for iterating! @muellerzr could you also have a look?

src/accelerate/accelerator.py
```diff
@@ -1643,6 +1643,7 @@ def _prepare_ipex(self, *args):
                 "Trying to use IPEX but IPEX is not installed or IPEX's version does not match current PyTorch, please refer"
                 " to https://github.com/intel/intel-extension-for-pytorch."
             )
+            return args
```
muellerzr, line 1646: I personally feel this should raise a full `ImportError` error, rather than just a `warning` and not return anything. Not doing so led to some subtle bugs we didn't know existed during the transformers integration, and iirc transformers also raises an error.

src/accelerate/accelerator.py
```diff
@@ -1655,19 +1656,14 @@ def _prepare_ipex(self, *args):
...
+                model = model.to(self.device)
+                model, optimizer = torch.xpu.optimize(
+                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
+                )
             else:
-                if is_torch_version(">=", "1.10"):
-                    model.forward = torch.autocast(self.device.type)(model.forward)
-                else:
-                    model.forward = convert_outputs_to_fp32(model.forward)
+                model, optimizer = ipex.optimize(model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1")
```
muellerzr, line 1666: We should definitely not be doing this, the native mixed precision autocast logic needs to stay when not using ipex. 

This second check for `ipex.optimize` should be under the xpu available check only or specific to `ipex`.

muellerzr (commented):
Two notes with these changes. Let's address these please, thanks!

src/accelerate/accelerator.py
abhilash1910: I agree during the initial PR I had some tests which were very slow and failing  (in 1118). 
I think we can keep as it is without returning and terminate in a  ImportError .

src/accelerate/accelerator.py
sywangyi: hi, @muellerzr , there's logic in https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L1307. after prepare_ipex. it will go there for the amp disposal.

sywangyi: ipex.optimize does not have device type input, it is for cpu now. for xpu, torch.xpu.optimize is used

muellerzr: Got it, that makes sense there then now, excellent. Sorry for the confusion!

src/accelerate/accelerator.py
sywangyi: currently. _prepare_ipex is only called when self.state.ipex_plugin is not None now. so maybe this import check should be removed

muellerzr: No, it should stay and not warn but instead raise an error. This is how the rest of the checks operate.

src/accelerate/accelerator.py
```diff
@@ -1639,8 +1639,8 @@ def _prepare_megatron_lm(self, *args):
 
     def _prepare_ipex(self, *args):
         if not is_ipex_available():
-            logger.warn(
-                "Trying to use IPEX but IPEX is not installed or IPEX's version does not match current PyTorch, please refer"
+            raise ImportError(
+                "Using IPEX but IPEX is not installed or IPEX's version does not match current PyTorch, please refer"
```
muellerzr, line 1643:
```suggestion
                "IPEX is not installed or IPEX's version does not match current PyTorch version. Please refer"
```

muellerzr (approved):
Small error message nit for consistency, otherwise LG2M! Thanks for iterating!

src/accelerate/accelerator.py
muellerzr, line 1643:
```suggestion
                "IPEX is not installed or IPEX's version does not match current PyTorch version. Please refer"
```

PR #1523, opened 2023-06-05
Officially support naive PP for quantized models + PEFT

### LINKED ISSUE #1515: ValueError: You can't train a model that has been loaded in 8-bit precision on multiple devices.
@younesbelkada 
(Thanks again for developing these great libraries and responding on Github!)

Related issue: https://github.com/huggingface/accelerate/issues/1412

With the bleeding edge `transformers`, I cannot combine PEFT and accelerate to do parameter-efficient fine-tuning with naive pipeline parallelism (i.e., splitting a model loaded on 8-bit across multiple GPUs).
Are both PEFT and accelerate not supporting such use cases? The code is working on earlier `transformers` version so wondering about it. 

```
  File "/home/ec2-user/.local/lib/python3.7/site-packages/transformers/trainer.py", line 1665, in train
    ignore_keys_for_eval=ignore_keys_for_eval,
  File "/home/ec2-user/.local/lib/python3.7/site-packages/transformers/trainer.py", line 1768, in _inner_training_loop
    self.model, self.optimizer, self.lr_scheduler
  File "/home/ec2-user/.local/lib/python3.7/site-packages/accelerate/accelerator.py", line 1144, in prepare
  File "/home/ec2-user/.local/lib/python3.7/site-packages/accelerate/accelerator.py", line 1144, in <genexpr>
  File "/home/ec2-user/.local/lib/python3.7/site-packages/accelerate/accelerator.py", line 995, in _prepare_one
    return self.prepare_model(obj, device_placement=device_placement)
  File "/home/ec2-user/.local/lib/python3.7/site-packages/accelerate/accelerator.py", line 1201, in prepare_model
    "You can't train a model that has been loaded in 8-bit precision on multiple devices."
ValueError: You can't train a model that has been loaded in 8-bit precision on multiple devices.
```

Here is the subset of outcomes from `pip3 list` regarding the package version:
```
Package                  Version
------------------------ -----------
accelerate               0.19.0
transformers             4.30.0.dev0
peft                     0.3.0
```

younesbelkada:
# What does this PR do?

Fixes #1515 

Naive Pipeline Parallelism should be supported by accelerate and should work, if we properly educate users on how to use it. 

## What is NPP? 

It is the simplest paradigm for running a model across multiple GPUs. It tries to evenly fit the model across all available GPUs (e.g. `device_map="auto"`)

npp 

## When to use it and when not to use it?

Initially I added that check because I was afraid users will train 8bit models that are loaded across multiple GPUs **and** under multi-GPU distributed regime. In that case the model will be converted to DDP (which is fine if the model fits in a single GPU and duplicated across multiple GPUs (Data Parallelism)) - which can lead to many breaking behaviours such as https://github.com/huggingface/peft/issues/269#issuecomment-1498776567 . 
The fix is to relax the check constraint and to also check if we are under multi GPU distributed regime (expects to use DDP).

In TRL library, it is possible to use the `PPOTrainer` (that calls `accelerator.prepare` under the hood) to apply Naive Pipeline Parallelism: https://huggingface.co/docs/trl/main/en/lora_tuning_peft#naive-pipeline-parallelism-npp-for-large-models-60b-models to train 60B+ scale models using RLHF. The error was never raised there because I forgot to store the attribute `hf_device_map` inside the model class we use in TRL.

To reproduce (you need PEFT and run this script in a multi-GPU env):
```python
from transformers import AutoModelForCausalLM, AutoTokenizer
from peft import LoraConfig, get_peft_model, prepare_model_for_int8_training

model_id = "facebook/opt-350m"
accelerator = Accelerator()

config = LoraConfig(
    r=16, 
    lora_alpha=32, 
    target_modules=["q_proj", "v_proj"], 
    lora_dropout=0.05, 
    bias="none", 
    task_type="CAUSAL_LM"
)

model = AutoModelForCausalLM.from_pretrained(model_id, device_map="auto", load_in_8bit=True)
model = prepare_model_for_int8_training(model)

print(set(model.hf_device_map.values()))

model = get_peft_model(model, config)

```

cc @sgugger @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1258,7 +1258,7 @@ class Accelerator:
             model, "hf_device_map", False
         ):
             model_devices = set(model.hf_device_map.values())
-            if len(model_devices) > 1 and self.distributed_type == DistributedType.MULTI_GPU:
+            if len(model_devices) > 1 and self.distributed_type != DistributedType.NO:
                 raise ValueError(
                     "You can't train a model that has been loaded in 8-bit precision on multiple devices in any distributed mode."
                     " In order to use 8-bit models that have been loaded across multiple GPUs the solution is to use Naive Pipeline Parallelism."
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1260,9 +1260,9 @@ class Accelerator:
             model_devices = set(model.hf_device_map.values())
             if len(model_devices) > 1 and self.distributed_type == DistributedType.MULTI_GPU:
                 raise ValueError(
-                    "You can't train a model that has been loaded in 8-bit precision on multiple devices in a multi-GPU setup (to use DDP)."
+                    "You can't train a model that has been loaded in 8-bit precision on multiple devices in any distributed mode."
                     " In order to use 8-bit models that have been loaded across multiple GPUs the solution is to use Naive Pipeline Parallelism."
-                    " Therefore you should not specify that you are under mutli GPU regime in your accelerate config."
+                    " Therefore you should not specify that you are under any distributed regime in your accelerate config."
                 )
             current_device = list(model_devices)[0]
             current_device_index = current_device.index if isinstance(current_device, torch.device) else current_device
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1258,9 +1258,11 @@ class Accelerator:
             model, "hf_device_map", False
         ):
             model_devices = set(model.hf_device_map.values())
-            if len(model_devices) > 1:
+            if len(model_devices) > 1 and self.distributed_type == DistributedType.MULTI_GPU:
                 raise ValueError(
-                    "You can't train a model that has been loaded in 8-bit precision on multiple devices."
+                    "You can't train a model that has been loaded in 8-bit precision on multiple devices in a multi-GPU setup (to use DDP)."
+                    " In order to use 8-bit models that have been loaded across multiple GPUs the solution is to use Naive Pipeline Parallelism."
+                    " Therefore you should not specify that you are under mutli GPU regime in your accelerate config."
                 )
             current_device = list(model_devices)[0]
             current_device_index = current_device.index if isinstance(current_device, torch.device) else current_device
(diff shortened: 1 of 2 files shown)


src/accelerate/accelerator.py
```diff
@@ -1258,9 +1258,11 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
             model, "hf_device_map", False
         ):
             model_devices = set(model.hf_device_map.values())
-            if len(model_devices) > 1:
+            if len(model_devices) > 1 and self.distributed_type == DistributedType.MULTI_GPU:
```
younesbelkada, line 1261:
```suggestion
            if len(model_devices) > 1 and not (self.distributed_type != DistributedType.NO):
```
maybe this is a better check

src/accelerate/accelerator.py
```diff
@@ -1258,9 +1258,11 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
...
                 raise ValueError(
-                    "You can't train a model that has been loaded in 8-bit precision on multiple devices."
+                    "You can't train a model that has been loaded in 8-bit precision on multiple devices in a multi-GPU setup (to use DDP)."
```
younesbelkada, line 1263:
```suggestion
                    "You can't train a model that has been loaded in 8-bit precision on multiple devices in any distributed mode."
```

src/accelerate/accelerator.py
```diff
@@ -1258,9 +1258,11 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
...
+                    " In order to use 8-bit models that have been loaded across multiple GPUs the solution is to use Naive Pipeline Parallelism."
+                    " Therefore you should not specify that you are under mutli GPU regime in your accelerate config."
```
younesbelkada, line 1265:
```suggestion
                    " Therefore you should not specify that you are under any distributed regime in your accelerate config."
```

src/accelerate/accelerator.py
sgugger: Yes this is a better check, but you need to also test it is not the (deprecated) `DistributedType.MPS` (although maybe we can remove this one now @muellerzr ?)

sgugger (approved):
Thanks!

src/accelerate/accelerator.py
younesbelkada: I see, I think let's add this suggestion as it also includes that the distributed type is not `DistributedType.MPS`

muellerzr (approved):
Thanks! LG2M :)

PR #1525, opened 2023-06-05
Refactor mp into its own wrapper

muellerzr:
Refactors the mixed precision section of `prepare_model` into its own function, and lets `prepare_model` take in a param to set or not set mixed precision.


pacman100 (approved):
Thank you @muellerzr for the super quick PR 🔥, LGTM!

PR #1527, opened 2023-06-05
Refactor mp into its own wrapper

muellerzr:
Refactors the mixed precision section of prepare_model into its own function, and lets prepare_model take in a param to set or not set mixed precision.

For some reason git history and I were not in agreement today, cleaner version of the same PR

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -25,7 +25,6 @@ import warnings
 from collections import OrderedDict
 from contextlib import contextmanager
 from functools import partial
-from types import MethodType
 from typing import Any, Callable
 
 import torch
@@ -1220,7 +1219,7 @@ class Accelerator:
 
         return result if len(result) > 1 else result[0]
 
-    def prepare_model(self, model: torch.nn.Module, device_placement=None):
+    def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, apply_mixed_precision: bool = True):
         """
         Prepares a PyTorch model for training in any distributed setup. It is recommended to use
         [`Accelerator.prepare`] instead.
@@ -1231,6 +1230,8 @@ class Accelerator:
                 any kind of mixed precision
             device_placement (`bool`, *optional*):
                 Whether or not to place the model on the proper device. Will default to `self.device_placement`.
+            apply_mixed_precision (`bool`, *optional*, defaults to `True`):
+                Whether or not to apply automatic mixed precision to the model based on `self.mixed_precision`.
 
         Example:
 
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1231,6 +1231,8 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
                 any kind of mixed precision
             device_placement (`bool`, *optional*):
                 Whether or not to place the model on the proper device. Will default to `self.device_placement`.
+            apply_mixed_precision (`bool`, *optional*, defaults to `True`):
+                Whether or not to apply automatic mixed precision to the model based on `self.mixed_precision`.
```
sgugger, line 1235: I was thinking more of a `mixed_precision_only` flag, so we don't need a new method.

src/accelerate/accelerator.py
```diff
@@ -1683,6 +1655,58 @@ def _prepare_ipex(self, *args):
                 result[i] = optimizer
         return tuple(result)
 
+    def prepare_mixed_precision(self, model: torch.nn.Module):
```
sgugger, line 1658: See comment above :-)

src/accelerate/accelerator.py
```diff
@@ -1683,6 +1655,58 @@ def _prepare_ipex(self, *args):
...
+        """
+        Args:
+        Prepares a PyTorch model automatic mixed precision by wrapping it's forward layer and converting its output
+        layer to fp32. It is recommended to use [`Accelerator.prepare`] instead.
+            model (`torch.nn.Module`):
+                A PyTorch model to convert to mixed precision
```
sgugger, line 1664: Docstring here is borked.

sgugger (commented):
Thanks for working on this! Left a comment.

src/accelerate/accelerator.py
```diff
@@ -1231,6 +1231,8 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
                 any kind of mixed precision
             device_placement (`bool`, *optional*):
                 Whether or not to place the model on the proper device. Will default to `self.device_placement`.
+            apply_mixed_precision (`bool`, *optional*, defaults to `True`):
```
sgugger: Needs to be updated :-)

sgugger (approved):
Thanks!

src/accelerate/accelerator.py
```diff
@@ -1347,13 +1349,14 @@ def prepare_model(self, model: torch.nn.Module, device_placement=None):
...
-                raise ValueError("Using `torch.compile` requires PyTorch 2.0 or higher.")
-            model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
+        if not mixed_precision_only:
+            if self.distributed_type == DistributedType.TPU and self.state.fork_launched:
+                model = xmp.MpModelWrapper(model).to(self.device)
+            # torch.compile should be called last.
+            if self.state.dynamo_plugin.backend != DynamoBackend.NO:
+                if not is_torch_version(">=", "2.0"):
+                    raise ValueError("Using `torch.compile` requires PyTorch 2.0 or higher.")
+                model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
```
pacman100, line 1359: torch compile should also be allowed in eval-only mode, right?

pacman100 (commented):
Hello, Thank you for adding the `mixed_precision_only` kwarg. Left a comment.

src/accelerate/accelerator.py
muellerzr: In the context of this PR no, because it's mixed precision mode only. If we want to add more disabling options like this, would probably be good to write a config that can be passed to the `Accelerator`. WDYT @sgugger

pacman100: We can rename `mixed_precision_only` to `eval_only` and have it handle cases similar to trainer. This should be easy to do and clear in terms of what is being done.

PR #1529, opened 2023-06-05
Check tied parameters

SunMarc:
# What does this PR do ? 

This PR fixes two issues user can have when using big inference model:
- Use their own device map but forget that parameters that are tied together should be on the same device. We return an error showing which parameters should be on the same device
- Forget to tie the parameters before using infer_auto_device_map() which can create a bad device_map. We also return an error asking to tie the weights before using this function.

# How to test it 

### Issue 1 
```python
import os
import torch
from transformers import AutoTokenizer
from accelerate.utils import find_tied_parameters

checkpoint = "facebook/opt-350m"

device_map_work = {'model.decoder.embed_tokens': 'cpu',
 'model.decoder.embed_positions': 'cpu',
 'model.decoder.project_out': 'cpu',
 'model.decoder.project_in': 'cpu',
 'model.decoder.layers': 'cpu',
 'lm_head': 'cpu'}

device_map_do_no_work = {'model.decoder.embed_tokens': 'cpu',
 'model.decoder.embed_positions': 'cpu',
 'model.decoder.project_out': 'cpu',
 'model.decoder.project_in': 'cpu',
 'model.decoder.layers': 'cpu',
 'lm_head': 'disk'}

model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map = device_map_do_no_work, offload_folder="offload",offload_state_dict = True)
tokenizer = AutoTokenizer.from_pretrained(checkpoint)
```

### Issue 2
```python
import torch
from transformers import AutoConfig,AutoModelForCausalLM
from accelerate import init_empty_weights, infer_auto_device_map, load_checkpoint_and_dispatch

checkpoint = "facebook/opt-350m"

config = AutoConfig.from_pretrained(checkpoint)
with init_empty_weights():
    model = AutoModelForCausalLM.from_config(config)
device_map = infer_auto_device_map(model, no_split_module_classes=["OPTDecoderLayer"])
```

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -281,8 +281,8 @@ def check_tied_parameters_on_same_device(tied_params, device_map):
             tie_param_devices[param] = _get_param_device(param, device_map)
         if len(set(tie_param_devices.values())) > 1:
             raise RuntimeError(
-                f"Tied parameters are in different devices: {tie_param_devices}."
-                "Please modify your custom device map or set device_map = 'auto' "
+                f"Tied parameters are in different devices: {tie_param_devices}. "
+                "Please modify your custom device map or set device_map = 'auto'. "
             )
 
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -680,11 +680,10 @@ def infer_auto_device_map(
     module_sizes = compute_module_sizes(model, dtype=dtype, special_dtypes=special_dtypes)
     tied_parameters = find_tied_parameters(model)
 
-    if check_tied_parameters(model):
-        if len(tied_parameters) == 0:
-            raise RuntimeError(
-                "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
-            )
+    if check_tied_parameters(model) and len(tied_parameters) == 0:
+        raise RuntimeError(
+            "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
+        )
 
     device_map = {}
     current_device = 0
@@ -1011,11 +1010,10 @@ def load_checkpoint_in_model(
     """
     tied_params = find_tied_parameters(model)
 
-    if check_tied_parameters(model):
-        if len(tied_params) == 0:
-            raise RuntimeError(
-                "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
-            )
+    if check_tied_parameters(model) and len(tied_params) == 0:
+        raise RuntimeError(
+            "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
+        )
 
     check_tied_parameters_on_same_device()
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -233,6 +233,59 @@ class FindTiedParametersResult(list):
         return sum([x[1:] for x in self], [])
 
 
+def check_tied_parameters(model: nn.Module):
+    """
+    Check if there is any indication in the given model that some weights should be tied.
+
+    Args:
+        model (nn.Module): The model to inspect
+
+    Returns:
+        bool: True if the model needs to have tied weights
+    """
+
+    # based on model.tie_weights() method
+    return any(
+        [
+            getattr(model.config, "tie_word_embeddings", True) and model.get_output_embeddings(),
+            getattr(model.config, "is_encoder_decoder", False) and getattr(model.config, "tie_encoder_decoder", False),
+            any(hasattr(module, "_tie_weights") for module in model.modules()),
+        ]
+    )
+
+
+def _get_param_device(param, device_map):
+    if param in device_map:
+        return device_map[param]
+    parent_param = ".".join(param.split(".")[:-1])
+    if parent_param == param:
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -233,6 +233,63 @@ def values(self):
...
+        [
+            hasattr(model.config, True)
+            and getattr(model.config, "tie_word_embeddings", False)
+            and model.get_output_embeddings(),
+            hasattr(model.config, True)
+            and getattr(model.config, "is_encoder_decoder", False)
+            and getattr(model.config, "tie_encoder_decoder", False),
+            any(hasattr(module, "_tie_weights") for module in model.modules()),
+        ]
+    )
```
sgugger: Here maybe start with a check on whether one of the base classes if `PreTrainedModel` since this is very Transformers-centric. (`inspect.getmro(model)` will give you the list of super classes of the model)

Also no needs for `and` since you are using the `any`.

src/accelerate/utils/modeling.py
```diff
@@ -233,6 +233,63 @@ def values(self):
...
+
+
+def _get_param_device(param, device_map):
+    if param in device_map:
+        return device_map[param]
+    parent_param = ".".join(param.split(".")[:-1])
+    if parent_param == param:
+        raise ValueError(f"The device_map was do not contain the module {param}.")
```
sgugger, line 266:
```suggestion
        raise ValueError(f"The `device_map` does not contain the module {param}.")
```

src/accelerate/utils/modeling.py
```diff
@@ -233,6 +233,63 @@ def values(self):
...
+        return _get_param_device(parent_param, device_map)
+
+
+def check_tied_parameters_on_same_device(tied_params, device_map):
+    """
+    Check if tied parameters are on the same device
+
+    Args:
+        tied_params (List[List[str]]):  A list of lists of parameter names being all tied together
+        device_map (Dict[str, Union[int, str, torch.device]]): mapping between the parameters and the devices
```
sgugger, line 277: Types should be in code blocks. Also copy the docstring of `device_map` from another function.

src/accelerate/utils/modeling.py
```diff
@@ -233,6 +233,63 @@ def values(self):
...
+    Raises:
+        RuntimeError: If we found tied parameters in different devices
+    """
+    for tie_param in tied_params:
+        tie_param_devices = {}
+        for param in tie_param:
+            tie_param_devices[param] = _get_param_device(param, device_map)
+        if len(set(tie_param_devices.values())) > 1:
+            raise RuntimeError(
+                f"Tied parameters are in different devices: {tie_param_devices}. "
```
sgugger, line 288:
```suggestion
                f"Tied parameters are on different devices: {tie_param_devices}. "
```

src/accelerate/utils/modeling.py
```diff
@@ -233,6 +233,63 @@ def values(self):
...
+                "Please modify your custom device map or set device_map = 'auto'. "
```
sgugger, line 289:
```suggestion
                "Please modify your custom device map or set `device_map='auto'`. "
```

src/accelerate/utils/modeling.py
```diff
@@ -627,6 +684,11 @@ def infer_auto_device_map(
     module_sizes = compute_module_sizes(model, dtype=dtype, special_dtypes=special_dtypes)
     tied_parameters = find_tied_parameters(model)
 
+    if check_tied_parameters(model) and len(tied_parameters) == 0:
+        raise RuntimeError(
+            "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
+        )
```
sgugger, line 690: This should probably be a warning here. An error seems a bit too much.

src/accelerate/utils/modeling.py
```diff
@@ -951,6 +1013,14 @@ def load_checkpoint_in_model(
             Whether or not to include the buffers in the weights offloaded to disk.
     """
     tied_params = find_tied_parameters(model)
+
+    if check_tied_parameters(model) and len(tied_params) == 0:
+        raise RuntimeError(
```
sgugger, line 1018: Same here.

src/accelerate/utils/modeling.py
```diff
@@ -951,6 +1013,14 @@ def load_checkpoint_in_model(
...
+            "The model weights are not tied. Please run model.tie_weights() before using infer_auto_device."
```
sgugger, line 1019: Name needs to be adapted here.

sgugger (commented):
Thanks for those QOL improvements!

src/accelerate/utils/modeling.py
```diff
@@ -233,29 +237,36 @@ def values(self):
...
 
     # based on model.tie_weights() method
-    return any(
-        [
-            hasattr(model,"config")
+    has_tied_word_embedding = False
+    has_tied_encoder_decoder = False
+    has_tied_module = False
+
+    if transformers.modeling_utils.PreTrainedModel in inspect.getmro(model.__class__):
```
sgugger, line 256: I was thinking on testing the class `__name__` to avoid the extra dep on Transformers.

sgugger (approved):

PR #1531, opened 2023-06-05
Raise ValueError on iterable dataset if we've hit the end and attempting to go beyond it

muellerzr:
Replaces the following confusing error:
```python
../accelerate/src/accelerate/data_loader.py:564: in __iter__
    observed_batch_size = find_batch_size(batch)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data = None

    def find_batch_size(data):
        """
        Recursively finds the batch size in a nested list/tuple/dictionary of lists of tensors.
    
        Args:
            data (nested list/tuple/dictionary of `torch.Tensor`): The data from which to find the batch size.
    
        Returns:
            `int`: The batch size.
        """
        if isinstance(data, (tuple, list)):
            return find_batch_size(data[0])
        elif isinstance(data, Mapping):
            for k in data.keys():
                return find_batch_size(data[k])
        elif not isinstance(data, torch.Tensor):
>           raise TypeError(f"Can only find the batch size of tensors but got {type(data)}.")
E           TypeError: Can only find the batch size of tensors but got <class 'NoneType'>.

../accelerate/src/accelerate/utils/operations.py:223: TypeError
```
With what is actually happening:
```python
src/transformers/trainer.py:1531: in train
    return inner_training_loop(
src/transformers/trainer.py:1779: in _inner_training_loop
    for step, inputs in enumerate(epoch_iterator):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <accelerate.data_loader.DataLoaderDispatcher object at 0x7fe22c27b2e0>

    def __iter__(self):
        self.gradient_state._add_dataloader(self)
        main_iterator = None
        if self.state.process_index == 0:
            # We only iterate through the DataLoader on process 0.
            main_iterator = super().__iter__()
        stop_iteration = False
        self._stop_iteration = False
        first_batch = None
        next_batch, next_batch_info = self._fetch_batches(main_iterator)
        batch_index = 0
        while not stop_iteration:
            batch, batch_info = next_batch, next_batch_info
    
            if self.state.process_index != 0:
                # Initialize tensors on other processes than process 0.
                batch = initialize_tensors(batch_info[0])
            batch = send_to_device(batch, self.state.device)
            # Broadcast the batch before splitting it.
            batch = broadcast(batch, from_process=0)
    
            if not self._drop_last and first_batch is None:
                # We keep at least num processes elements of the first batch to be able to complete the last batch
                first_batch = slice_tensors(batch, slice(0, self.state.num_processes))
    
            if batch is None:
>               raise ValueError(f'Batch does not contain any data ({batch}). At the end of all iterable data available before expected stop iteration.')
E               ValueError: Batch does not contain any data (None). At the end of all iterable data available before expected stop iteration.

../accelerate/src/accelerate/data_loader.py:565: ValueError
```

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -561,6 +561,9 @@ class DataLoaderDispatcher(DataLoader, DataLoaderStateMixin):
                 # We keep at least num processes elements of the first batch to be able to complete the last batch
                 first_batch = slice_tensors(batch, slice(0, self.state.num_processes))
 
+            if batch is None:
+                raise ValueError(f'Batch does not contain any data (`{batch}`). At the end of all iterable data available before expected stop iteration.')
+
             observed_batch_size = find_batch_size(batch)
             batch_size = observed_batch_size // self.state.num_processes
 
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -562,7 +562,9 @@ class DataLoaderDispatcher(DataLoader, DataLoaderStateMixin):
                 first_batch = slice_tensors(batch, slice(0, self.state.num_processes))
 
             if batch is None:
-                raise ValueError(f'Batch does not contain any data (`{batch}`). At the end of all iterable data available before expected stop iteration.')
+                raise ValueError(
+                    f"Batch does not contain any data (`{batch}`). At the end of all iterable data available before expected stop iteration."
+                )
 
             observed_batch_size = find_batch_size(batch)
             batch_size = observed_batch_size // self.state.num_processes


sgugger (approved):
Thanks!

cmosguy:
@sgugger  and @muellerzr  - so I've encountered this error when trying to do finetuning.  I do not understand what I am doing wrong to encounter this violation.  I posted my issue here: https://github.com/loubnabnl/santacoder-finetuning/issues/23 but maybe you can help elaborate on this what is the reason for the error and how to workaround the issue with small datasets?

Thank you!

muellerzr:
@cmosguy any chance you can share the dataset you're using? (and as such, a full reproducer) Essentially this means there's likely a streaming dataset or something and your maximum number of steps you're using is larger than the dataset and it's returning empty after awhile

cmosguy:
hey @muellerzr - the issue this is a private dataset that I am working with.   All I did was scoop up a bunch of my python source code then  create a dataset in `json` format for the data loader to consume for training.  I am trying to finetune star coder on my own codebase.  It would take me some time to sort this out.  Is there any other way to debug this without create datasets for you?  Maybe we can selectively choose 295 files from `bigcode/the-stack-dedup`

In other words, we could do something like this:
```bash
--model_path="bigcode/santacoder" \
        --dataset_name="bigcode/the-stack-dedup" \
        --subset="data/python" \
        --data_column "content" \
        --split="train[0:297]" \
```

Would that work for you?

muellerzr:
That would work! Tell me if you still see that issue and I can look at that tommorow :)

cmosguy:
@muellerzr  - had to to increase the test_size from `0.005` to `0.015` to get the training to finally work on my system. Do you have any thoughts why that made a difference?

PR #1535, opened 2023-06-06
Fixed issue with printing device map in Accelerate

de5ash:
This contribution addresses an issue where the device map was not printed correctly in the Accelerate library. The fix ensures that the device map is printed as expected, providing accurate information about the device allocation. This enhances the usability and debugging capabilities of the library


muellerzr:
Hi! Thanks for the PR, could you do `pip install -e .[quality]` followed by `make style; make quality`? This will make the tests pass

sgugger (commented):
Thanks but this is the docstring of `load_checkpoint_and_dispatch`, there is no need to use `infer_auto_device_map` there.

PR #1536, opened 2023-06-06
Prevent using extra VRAM for static device_map

LSerranoPEReN:
When using `load_checkpoint_and_dispatch` with a user provided `device_map`, the function first computes the `max_memory` for each device. This has a side-effect of allocating one tensor, and thus cuda, on each available device, even if the device is not present in the values of the `device_map`.

Since this argument is not used with a user provided `device_map`, this PR simply move the `max_memory` computation out of the execution path, to prevent extra memory allocation on unused devices.


sgugger (approved):
LGTM, thanks for the fix!

PR #1540, opened 2023-06-06
Instead go with `evaluation_mode`

muellerzr:
Sets an `evaluation_mode` instead under `prepare_model`, that will let `torch.compile` remain as well.


pacman100 (approved):
Thank you Zach for adding this! 🤗

PR #1545, opened 2023-06-06
Add mps support to big inference modeling

SunMarc:
# What does this PR do

This PR add the support for M1/M2 MacOS to big inference modeling. 

The main highlights of this PR are as follows:
- device_map = 'auto' will now detect mps device if available and set the device_map correctly. Since memory is shared between the GPU and CPU, modules are placed on the MPS device, and if necessary, offloaded to the disk.
- add tests for mps device

Reminder: mps backend can be slower than cpu in some cases.

Fix [#23916](https://github.com/huggingface/transformers/issues/23916)

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -30,7 +30,7 @@ import torch.nn as nn
 
 from ..state import AcceleratorState
 from .dataclasses import DistributedType
-from .imports import is_safetensors_available, is_torch_version, is_xpu_available
+from .imports import is_mps_available, is_safetensors_available, is_torch_version, is_xpu_available
 from .offload import load_offloaded_weight, offload_weight, save_offload_index
 from .tqdm import is_tqdm_available, tqdm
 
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -688,7 +692,10 @@ def infer_auto_device_map(
         devices.append("disk")
 
     # Devices that need to keep space for a potential offloaded layer.
-    main_devices = [gpus[0], "cpu"] if len(gpus) > 0 else ["cpu"]
+    if "mps" in gpus:
+        main_devices = ["mps"]
+    else:
+        main_devices = [gpus[0], "cpu"] if len(gpus) > 0 else ["cpu"]
```
sgugger, line 698: Let's not mix and match syntaxes
```suggestion
    if "mps" in gpus:
        main_devices = ["mps"]
    elif len(gpus) > 0:
        main_devices = [gpus[0], "cpu"] if len(gpus) > 0
    else:
        main_devices = ["cpu"]
```

sgugger (approved):
Nice, thanks a lot for adding support!

src/accelerate/utils/modeling.py
```diff
@@ -694,8 +694,10 @@ def infer_auto_device_map(
     # Devices that need to keep space for a potential offloaded layer.
     if "mps" in gpus:
         main_devices = ["mps"]
+    elif len(gpus) > 0:
+        main_devices = [gpus[0], "cpu"] if len(gpus) > 0
```
sgugger, line 698:
```suggestion
        main_devices = [gpus[0], "cpu"]
```
Arf ;-)

PR #1552, opened 2023-06-07

muellerzr:

Also resets step to 0 during `free_memory`


sgugger (commented):
That works. Might be cleaner to add an attribute `_is_accelerate_prepared` to all things we return with `prepare` and check for that.

pacman100 (approved):
Thank you @muellerzr for the super quick PR with the fixes for handling `auto_find_batch_size` use case.🤗

muellerzr:
Closing b/c github 🙃

PR #1554, opened 2023-06-07

muellerzr:

Also resets step to 0 during free_memory


pacman100 (approved):
Thank you!

PR #1555, opened 2023-06-07

muellerzr:

Also resets step to 0 during free_memory

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1729,6 +1729,10 @@ class Accelerator:
         >>> optimizer = accelerator.prepare_optimizer(optimizer, device_placement=True)
         ```
         """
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Thanks!

Yura52:
@muellerzr @sgugger Hi! I have zero familiarity with the code base, but is it intentional that the added `setattr()` call sets the "_is_accelerate_prepared" attribute, while the used `getattr()` calls access the different "_accelerate_prepared" (note missing "_is" suffix) attribute?

muellerzr:
Nope, it is not! Would you like to put in a fix @Yura52? Otherwise I can do so quickly :)

Yura52:
> Nope, it is not! Would you like to put in a fix @Yura52? Otherwise I can do so quickly :)

Here is the PR: #1560

PR #1557, opened 2023-06-07
Fix load_state_dict when there is one device and disk

sgugger:
When there is one device + the disk, the state dict will return everything on the device (instead of putting the disk weights on the CPU). This PR addresses that.


muellerzr (approved):
Passes for me locally on single and multi-GPU (being available to be ran on) :) Thanks for the fix!

PR #1558, opened 2023-06-07
Fix tests not being ran on multi-GPU nightly

muellerzr:
It's still... *strange* that the workflows running `make test` (we have a number of them) didn't pick up on https://github.com/huggingface/accelerate/pull/1557... but at least this solves some of it...


sgugger (approved):
Thanks for the fix!

PR #1560, opened 2023-06-07
fix the typo when setting the "_accelerator_prepared" attribute

Yura52:
Fixes the typo from #1555 (setattr and the corresponding getattr currently use different attribute names, which is a bug)


muellerzr:
Thanks! Let's do the opposite though :) (keep `is_accelerate_prepared`

sgugger (approved):

sgugger:
Thanks for the fix!

PR #1561, opened 2023-06-08
[`core`] Fix possibility to pass`NoneType` objects in `prepare`

younesbelkada:
# What does this PR do?

Fixes https://github.com/lvwerra/trl/issues/413 and the current CI in TRL :D https://github.com/lvwerra/trl/actions/runs/5204023029/jobs/9387749684?pr=412 

Before https://github.com/huggingface/accelerate/pull/1555 it was possible to pass `NoneType` objects into prepare and that function would do nothing with it and return it. 

Adding a simple check `if item is not None:`  before setting the `_is_accelerate_prepare` attribute circumvents the bug

cc @sgugger @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1198,7 +1198,8 @@ class Accelerator:
             result = self._prepare_fsdp(*result)
 
         for item in result:
-            setattr(item, "_is_accelerate_prepared", True)
+            if item is not None:
+                setattr(item, "_is_accelerate_prepared", True)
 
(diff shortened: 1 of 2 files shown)


pacman100 (approved):
Thank you for the fix!

sgugger:
Can you also add a test where `None` is passed to `Accelerator.prepare`? If you rely on this behavior, we need to make sure it's tested so there is no future regression.

younesbelkada:
sure, adding it now

sgugger (approved):
Thanks!

PR #1562, opened 2023-06-08
Reset dataloader end_of_datalaoder at each iter

sgugger:
The `end_of_dataloader` from the `DataLoaderMixin` is set to `True` at the end of an iter, but never reset to False...

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -903,7 +903,9 @@ class GradientState:
             self.sync_gradients = True
             self.active_dataloader = None
             self.dataloader_references = [None]
-            self.plugin_kwargs = gradient_accumulation_plugin.to_kwargs()
+            self.plugin_kwargs = (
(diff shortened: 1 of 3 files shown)


src/accelerate/state.py
```diff
@@ -903,7 +903,9 @@ def __init__(self, gradient_accumulation_plugin: Optional[GradientAccumulationPl
             self.sync_gradients = True
             self.active_dataloader = None
             self.dataloader_references = [None]
-            self.plugin_kwargs = gradient_accumulation_plugin.to_kwargs()
+            self.plugin_kwargs = (
+                gradient_accumulation_plugin.to_kwargs() if gradient_accumulation_plugin is not None else {}
```
sgugger, line 907: not strictly related but I could not test as with the added test without this change.

muellerzr (approved):
Thanks! Nice solution, and good test. Esp as it catches the `None` case now.

PR #1567, opened 2023-06-09
[`bnb`] Fix failing int8 tests

younesbelkada:
# What does this PR do?

Fixes failing tests in https://github.com/huggingface/accelerate/actions/runs/5217618552/jobs/9417634363 

The new release of transformers includes the 4bit integration, therefore the behavior of `replace_8bit_linear` has been changed to support 4bit and replaced to `replace_with_bnb_linear` method. This PR addresses corresponding changes to properly call this method

cc @muellerzr @sgugger


sgugger (approved):
Thanks for fixing!

Issue #1563, opened 2023-06-08
LR resume is somewhat buggy


stas00:
### System Info

During m4 training we had to hack the LR resume state file to adjust LR in the middle of the training.

As I wasn't sure if I did the right thing, I saved a back up file next to the resume file so that I had:

```
opt_step-30500/accelerator_state> ls -1
custom_checkpoint_0.pkl
custom_checkpoint_0.pkl.orig
```

Now on resume Accelerate silently chose not to load the LR state and continue loading and restarted the LR cycle. 

```
 0: [ip-26-0-144-35:0]: - 1732205 - accelerate.checkpointing - All random states loaded successfully
 0: [ip-26-0-144-35:0]: - 1732205 - accelerate.accelerator - Warning! Number of found checkpoints does not match the number of registered objects:
 0: [ip-26-0-144-35:0]: Found checkpoints: 2
 0: [ip-26-0-144-35:0]: Registered objects: 1
 0: [ip-26-0-144-35:0]:Skipping.
```

2 Issues here:

1. what do you mean "Skipping"? please please please don't do any silent skipping - if you think something is wrong - assert - don't continue. Unfortunately warnings do not work in the sea of tens of thousands of log lines. 
2. why did this situation happen in the first place? The registered file was `custom_checkpoint_0.pkl` - it says nowhere a user is not allowed to have other files in the `accelerator_state` folder.

Thank you.

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

sgugger:
cc @muellerzr

muellerzr:
> why did this situation happen in the first place? The registered file was custom_checkpoint_0.pkl - it says nowhere a user is not allowed to have other files in the accelerator_state folder.

You did not register it with `Accelerator.register_for_checkpointing`, and as stated clearly in the docstring for `load_state`:

```
        Loads the current states of the model, optimizer, scaler, RNG generators, and registered objects.

        <Tip>

        Should only be used in conjunction with [`Accelerator.save_state`]. <-

        </Tip>
```

It is only meant to be used with just the load and save state paradigm. If there's something else needed in there you must register it for saving, which is mentioned in the docs.

We should probably raise an error instead here, but that's what this means

[...]

PR #1568, opened 2023-06-09
Raise error instead of warn during the check for custom objects in load_state

muellerzr:
Should solve https://github.com/huggingface/accelerate/issues/1563 by raising a more explicit error when users have objects with the `custom_checkpoint` naming convention which is specific to `save/load_state`

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2553,10 +2553,12 @@ class Accelerator:
         )
         custom_checkpoints = [f for f in os.listdir(input_dir) if "custom_checkpoint" in f]
         if len(custom_checkpoints) != len(self._custom_objects):
-            err = "Warning! Number of found checkpoints does not match the number of registered objects:"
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks!

stas00:
great, thank you for changing warning into assertion, Zachary.

This still doesn't address Issue 2 

> why did this situation happen in the first place? The registered file was custom_checkpoint_0.pkl - it says nowhere a user is not allowed to have other files in the accelerator_state folder.

why can't I leave a back up file which is not named `custom_checkpoint_0.pkl` in the `accelerator_state` folder.

muellerzr:
@stas00 it can't have the name `custom_checkpoint` because that is what we expect to read from during saving. Call it something else if you want to keep other files in there.

muellerzr:
Did not know of this, will make a second PR to allow for `.bak`. And will also adjust the docstring further in `load_state`

PR #1569, opened 2023-06-09
Update mixed precision integrations in README

sgugger:
Update the mixed precision integrations (FP16 with Apex is deprecated so removed it, and it was never actually requested).


muellerzr (approved):
Excellent!

PR #1575, opened 2023-06-12
improve oob performance when use mpirun to start DDP finetune without…

sywangyi:
improve oob performance when use mpirun to launch DDP finetune w/o `accelerate launch`

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -231,6 +231,20 @@ class PartialState:
                             "Looks like distributed multinode run but MASTER_ADDR env not set, "
                             "please try exporting rank 0's hostname as MASTER_ADDR"
                         )
+                if (
+                    self.distributed_type == DistributedType.MULTI_CPU
+                    and get_int_from_env(["OMP_NUM_THREADS", "MKL_NUM_THREADS"], 0) == 0
+                ):
+                    import psutil
+
+                    num_cpu_threads_per_process = int(psutil.cpu_count(logical=False) / local_size)
+                    if num_cpu_threads_per_process == 0:
+                        num_cpu_threads_per_process = 1
+                    torch.set_num_threads(num_cpu_threads_per_process)
+                    warnings.warn(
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for your PR!
@muellerzr you can merge if you're happy with the change.

src/accelerate/state.py
```diff
@@ -231,6 +231,20 @@ def __init__(self, cpu: bool = False, **kwargs):
...
+                    import psutil
+
+                    num_cpu_threads_per_process = int(psutil.cpu_count(logical=False) / local_size)
+                    if num_cpu_threads_per_process == 0:
+                        num_cpu_threads_per_process = 1
+                    torch.set_num_threads(num_cpu_threads_per_process)
+                    warnings.warn(
+                        f"OMP_NUM_THREADS/MKL_NUM_THREADS unset, we set it at {num_cpu_threads_per_process} to improve oob"
+                        " performance."
+                    )
```
muellerzr, line 247: This should be unnecessary, as we only set this during `_validate_launch_command`

https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/launch.py#L902-L913

The logic should be updated there only so we don't split this.

sywangyi: @muellerzr ,the logic in https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/launch.py#L902-L913 will be called only if "accelerate launch" is called. but in our multi-cpu guide in https://github.com/huggingface/transformers/blob/main/docs/source/en/perf_train_cpu_many.mdx, accelerate launch will not be used. in this case, if you do no set -genv OMP_NUM_THREADS=xxx explicitly in command, only 1 lcore is used in oob case

sywangyi: also in accelerate guide.
Launching multi-CPU run using MPI
🤗 Here is another way to launch multi-CPU run using MPI. You can learn how to install Open MPI on [this page](https://www.open-mpi.org/faq/?category=building#easy-build). You can use Intel MPI or MVAPICH as well. Once you have MPI setup on your cluster, just run:

mpirun -np 2 python examples/nlp_example.py

the accelerate launch is not used as well.

Issue #1488, opened 2023-05-30
load_checkpoint_and_dispatch compatibility with accelerate FSDP?


eric-mitchell:
### System Info

```Shell
- `Accelerate` version: 0.19.0
- Platform: Linux-5.15.0-41-generic-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1+cu117 (True)
- System RAM: 1007.69 GB
- GPU type: NVIDIA A100-SXM4-80GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - use_cpu: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {'deepspeed_config_file': 'configs/deepspeed_defaults.json', 'zero3_init_flag': False}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am trying to train a large (30B+) model with Accelerate & FSDP on 8 GPUs. I understand a model this large requires using a meta device to avoid instantiating the model 8 times (for each process) in CPU RAM.

However, I can't figure out the right way to use `load_checkpoint_and_dispatch` with `accelerator.prepare`. When I use `load_checkpoint_and_dispatch`, I see to get a copy of the model (split across all GPUs) *for each GPU*. However, I'd like to instantiate *only* the FSDP shards for that process. It feels like `load_checkpoint_and_dispatch` needs an `accelerator` argument, but one doesn't exist. The below code (using a small model for testing) gives the wrong behavior, with two copies of the model being created (i.e., both GPUs have 24GB VRAM used, not 12GB).

```

import transformers
from accelerate import Accelerator
import torch
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
from torch.utils.data import DataLoader, Dataset
import time
import os
import getpass
from accelerate import init_empty_weights, load_checkpoint_and_dispatch
[...]

Issue #1462, opened 2023-05-22
FSDP unable to load checkpoint, state dict, saved weights


conceptofmind:
### System Info

```Shell
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: FSDP
downcast_bf16: 'no'
fsdp_config: {}
gpu_ids: all
machine_rank: 0
main_process_ip: ''
main_process_port: ''
main_training_function: main
megatron_lm_config: {}
mixed_precision: bf16
num_machines: 8
num_processes: 64
rdzv_backend: static
same_network: false
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

### Information

- [X] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Hi all,

When trying to resume from a model checkpoint with `use_orig_params=True`:
```python
        if CFG.RESUME_FROM_CHECKPOINT is not None or CFG.RESUME_FROM_CHECKPOINT != "":
            accelerator.print(f"Resuming from checkpoint {CFG.RESUME_FROM_CHECKPOINT}")
            accelerator.load_state(CFG.RESUME_FROM_CHECKPOINT)
```
I am receiving this error:
```python
    main()
  File "/train_distributed.py", line 545, in main
    accelerator.load_state(CFG.RESUME_FROM_CHECKPOINT)
  File "/lib/python3.9/site-packages/accelerate/accelerator.py", line 2439, in load_state
    self.state.fsdp_plugin.load_optimizer(self, opt, self._models[i], input_dir, i)
  File "/lib/python3.9/site-packages/accelerate/utils/dataclasses.py", line 951, in load_optimizer
    sharded_osd = FSDP.scatter_full_optim_state_dict(full_osd, model)
  File "/lib/python3.9/site-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 1581, in scatter_full_optim_state_dict
    return FullyShardedDataParallel._optim_state_dict_to_load_impl(
  File "/lib/python3.9/site-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 1236, in _optim_state_dict_to_load_impl
    flat_osd = _flatten_optim_state_dict(
[...]

Issue #1442, opened 2023-05-17
Failing to load model using accelerate launch


TZeng20:
Hi,

Without using `accelerate launch`, I am able to load flan-t5-xxl and use it for inference without a problem
```
model_name = "google/flan-t5-xxl"
tokenizer = AutoTokenizer.from_pretrained(model_name)
model = T5ForConditionalGeneration.from_pretrained(model_name, device_map = 'auto')
```

However, when I use an accelerator script like below, I get a sigkill error when the model starts to load.
Why is it that I can load using `from_pretrained` but not when using `accelerate launch`?
Is there some deepspeed configuration that I need to specify?

```
import argparse
import os
from transformers import T5Tokenizer, T5ForConditionalGeneration, AutoTokenizer, set_seed
from accelerate import Accelerator, DistributedType
from accelerate.utils import set_seed
import torch
from datasets import load_dataset, Dataset
from torch.utils.data import DataLoader

def eval_function(args):
    set_seed(args.seed)
    accelerator = Accelerator()
    model_name = "google/flan-t5-xxl"
    tokenizer = AutoTokenizer.from_pretrained(model_name)
    model = T5ForConditionalGeneration.from_pretrained(model_name)
    print('Finished loading model!')

    # load data
    input_text = ['This is the first sentence', 'This is the second sentence']
    ds = Dataset.from_dict({'prompt': input_text})
    ds = ds.map(lambda x: tokenizer(x['prompt']), remove_columns= ['prompt'])
    ds = ds.with_format('torch')
    dataloader = DataLoader(ds, batch_size=1)

    # Prepare everything
    model, eval_dataloader = accelerator.prepare(model, dataloader)
    model.eval()
    dummy_input = "Are you human?"
    batch = tokenizer(dummy_input, return_tensors="pt").to(accelerator.device)
    labels = tokenizer("yes", return_tensors="pt").input_ids.to(accelerator.device)
    outputs = model(**batch, labels = labels)

    print('Beginning prediction!')
    for batch in eval_dataloader:
        input_ids = batch['input_ids']
        torch.cuda.empty_cache()        
        outputs = accelerator.unwrap_model(model).generate(
            input_ids,
            min_length = 100, 
            max_length = 200,
            num_beams = 5, 
            num_return_sequences = 1,
            do_sample = True,
            top_p = 0.9,
            no_repeat_ngram_size=1,
            remove_invalid_values=True,
            synced_gpus=True)
[...]

PR #1212, opened 2023-03-19
Update how FSDP uses state dicts for Torch 2.0

VikParuchuri:
Torch 2.0 has made some changes with FSDP and state dicts:

- `full_optim_state_dict` is [deprecated](https://pytorch.org/docs/stable/_modules/torch/distributed/fsdp/fully_sharded_data_parallel.html#FullyShardedDataParallel.full_optim_state_dict), and should be replaced with `optim_state_dict`.
- `scatter_full_optim_state_dict` is [deprecated](https://pytorch.org/docs/stable/_modules/torch/distributed/fsdp/fully_sharded_data_parallel.html#FullyShardedDataParallel.scatter_full_optim_state_dict), and should be replaced with `optim_state_dict_to_load`
- `optim_state_dict` does [not support](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.optim_state_dict) `optim_input`
- `set_state_dict_type` now takes an [optimizer state dict](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.set_state_dict_type), along with a model state dict

I have updated the FSDP plugin to mirror these changes.  I used `set_state_dict` to request the full state dict as needed.  I tried to keep parity with the existing logic in `save_model`, and only apply these changes for new versions of torch.

This works fine in my limited testing on Torch 2.0.

PR #1576, opened 2023-06-12
FSDP updates

pacman100:
# What does this PR do?
1. Many FSDP related issues due to gap between current implementation and latest torch version 2.0.1
    a. https://github.com/huggingface/transformers/issues/24057
    b. https://github.com/huggingface/accelerate/issues/1488
    c. https://github.com/huggingface/accelerate/issues/1462
    d. https://github.com/huggingface/accelerate/issues/1442
    e. https://github.com/huggingface/accelerate/pull/1212
    f. https://github.com/huggingface/transformers/pull/23481
    g. https://github.com/huggingface/transformers/pull/23812
2. This PR is to address all the above

ToDo:
- [x] Add Code
- [x] Add thorough docs
- [x] Update example
- [x] Add tests
- [ ] Add documentation in Transformers on how to use Accelerate Launcher for FSDP with Trainer


src/accelerate/accelerator.py
```diff
@@ -285,8 +284,8 @@ def __init__(
         if os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" or isinstance(
             fsdp_plugin, FullyShardedDataParallelPlugin
         ):
-            if is_torch_version("<", "1.12.0"):
-                raise ValueError("FSDP requires PyTorch >= 1.12.0")
+            if is_torch_version("<", "2.0.1"):
+                raise ValueError("FSDP requires PyTorch >= 2.0.1")
```
muellerzr, line 288: A thought I had on these, to help users out perhaps we should give these errors context of what specific accelerate version they should be using, if they want to use less than 2.0 (and for any other deps) so that they don't always have to be using the latest version, and know when they should pin things to? (And make it more stable in that regard)

pacman100: FSDP is highly experimental as it is and keeps evolving with every release, for FSDP sadly i can't see a way for users to get the latest and robust features without upgrading Torch

muellerzr: All good!

muellerzr (approved):
Thanks!

src/accelerate/utils/launch.py
```diff
@@ -172,6 +172,9 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
             current_env["FSDP_BACKWARD_PREFETCH"] = str(args.fsdp_backward_prefetch_policy)
         if args.fsdp_state_dict_type is not None:
             current_env["FSDP_STATE_DICT_TYPE"] = str(args.fsdp_state_dict_type)
+        current_env["FSDP_FORWARD_PREFETCH"] = str(args.fsdp_forward_prefetch).lower()
+        current_env["FSDP_USE_ORIG_PARAMS"] = str(args.fsdp_use_orig_params).lower()
+        current_env["FSDP_SYNC_MODULE_STATES"] = str(args.fsdp_sync_module_states).lower()
```
muellerzr, line 177: Question for you here, are these read/used only by us? Or by FSDP/something specific to it

pacman100: Used only bu us for the FSDP Plugin

muellerzr: Got it, eventually we should look at adding the `ACCELERATE_` prefix to keep in line with how the rest of our environmental variables we use/set are, for people to easily see. I can do this in a follow-up PR, no need to worry about this now :)

muellerzr (approved):
Thanks for iterating and the refactor 🙏

docs/source/usage_guides/fsdp.mdx
```diff
@@ -66,17 +66,67 @@ Currently, `Accelerate` supports the following config through the CLI:
...
+Below is an example:
+
+```
+from accelerate import FullyShardedDataParallelPlugin
+from torch.distributed.fsdp.fully_sharded_data_parallel import FullOptimStateDictConfig, FullStateDictConfig
+
+fsdp_plugin = FullyShardedDataParallelPlugin(
+        state_dict_config=FullStateDictConfig(offload_to_cpu=False, rank0_only=False),
+        optim_state_dict_config=FullOptimStateDictConfig(offload_to_cpu=False, rank0_only=False),
+    )
```
sgugger, line 95: This is badly formatted.

docs/source/usage_guides/fsdp.mdx
```diff
@@ -66,17 +66,67 @@ Currently, `Accelerate` supports the following config through the CLI:
...
+`Forward Prefetch`: If True, then FSDP explicitly prefetches the next upcoming all-gather while executing in the forward pass
+```
+
+For additional and more nuanced control, you can specify other FSDP parameters via `FullyShardedDataParallelPlugin`. 
+When creating `FullyShardedDataParallelPlugin` object, pass it the parameters that weren't part of the accelerate config or if you want to override them.
+The FSDP parameters will be picked based on the accelerate config file or launch command arguments and other parameters that you will pass directly through the `FullyShardedDataParallelPlugin` object will set/override that.
+
+Below is an example:
+
+```
```
sgugger, line 88:
````suggestion
```py
````

File: docs/source/usage_guides/fsdp.mdx [hunk 5]
```diff
@@ -66,17 +66,67 @@ Currently, `Accelerate` supports the following config through the CLI:
...
+accelerator = Accelerator(fsdp_plugin=fsdp_plugin)
 ```
 
 ## Saving and loading
 
-1. When using transformers `save_pretrained`, pass `state_dict=accelerator.get_state_dict(model)` to save the model state dict. 
+The new recommended way of checkpointing when using FSDP models is to use `SHARDED_STATE_DICT` as `StateDictType` when setting up the accelerate config.
+Below is the code snippet to save using `save_state` utility of accelerate.
+
+```
```
sgugger, line 105:
````suggestion
```py
````

File: docs/source/usage_guides/fsdp.mdx [hunk 6]
```diff
@@ -66,17 +66,67 @@ Currently, `Accelerate` supports the following config through the CLI:
...
+ls optimizer_0
+# __0_0.distcp  __1_0.distcp
+
+ls pytorch_model_0
+# __0_0.distcp  __1_0.distcp
+```
+
+To load them back for resuming the training, use the `load_state` utility of accelerate
+
+```
```
sgugger, line 125:
````suggestion
```py
````

sgugger (approved):
Just some comments on the doc, but otherwise LGTM, thanks!

PR #1577, opened 2023-06-12
More perminant solution for logging issues

muellerzr:
Rather than have the exception, this PR just checks for an `in` during the CLI tests for the output, since figuring out how to silence the loggers for every dependency that could be installed that decides to add bloat to the outputs is not sustainable.

(The tests are currently failing due to DeepSpeed 0.9.4 which changed from a `print` to a `logging.info`, making it still always log 🙃 )


sgugger (approved):
LGTM, thanks!

Issue #1574, opened 2023-06-10
_is_accelerate_prepared issue


spyroot:
### System Info

Hi Folks,

I was running code and upgrading for three days. Keep hitting this issue.

Before that, it was the only way I manage to run it. (where both are Torch Dataloader)
```python
self.model, self.optimizer, train_dataloader, eval_dataset = accelerator.prepare(
            [self.model, self.optimizer, train_dataloader, eval_dataset],
            device_placement=[True])
```

Note that in many examples in docs, you provide different examples, but in practice, I think you need to handle a bit more concrete.

Now I see in the code you patch 
AttributeError: 'list' object has no attribute '_is_accelerate_prepared'
Now the question is a list of 4 elements, of course, that will fail.

Now if I do
```
  self.model, self.optimizer, train_dataloader, eval_dataset = accelerator.prepare(
            self.model, self.optimizer, train_dataloader, eval_dataset,
            device_placement=[True, True, True, True])"
```
It does work. It blocks inside, prepares to hit 100, and never returns.

Now why did I in the beginning (i.e, in the version I had before) [x,y,z,w]

Because of this, in prepare.
```
  result = tuple(
      self._prepare_one(obj, first_pass=True, device_placement=d) 
  for obj, d in zip(args, device_placement)
  )
  result = tuple(self._prepare_one(obj, device_placement=d) 
for obj, d in zip(result, device_placement))
```

### Information

- [ ] The official example scripts
- [x ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

  self.model, self.optimizer, train_dataloader, eval_dataset = accelerator.prepare(
            [self.model, self.optimizer, train_dataloader, eval_dataset],
            device_placement=[True])"
            
It will try to set attr on the list
           

### Expected behavior

It is not clear what it is trying to do, I tried to understand why when I do   

self.model, self.optimizer, train_dataloader, eval_dataset = accelerator.prepare(
            self.model, self.optimizer, train_dataloader, eval_dataset,
            device_placement=[True, True, True, True])"
            
 Just block.

muellerzr:
[...]

PR #1578, opened 2023-06-12
Make sure that we only set is_accelerator_prepared on items accelerate actually prepares

muellerzr:
Another solution to our `_is_accelerator_prepared`, checking that the object must be in one of the `self._{type}` that is designated for storing

Solves https://github.com/huggingface/accelerate/issues/1574


sgugger (approved):
Thanks for fixing!

Issue #1178, opened 2023-03-10
`logger.info` and `logger.debug` never print anything


vvvm23:
### System Info

```Shell
- `Accelerate` version: 0.17.0
- Platform: Linux-6.1.12-arch1-1-x86_64-with-glibc2.37
- Python version: 3.10.9
- Numpy version: 1.23.5
- PyTorch version (GPU?): 2.1.0.dev20230310+cu118 (True)
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: bf16
        - use_cpu: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {}
        - fsdp_config: {}
        - megatron_lm_config: {}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {'dynamo_backend': 'INDUCTOR'}
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Run the following script:
```python
from accelerate.logging import get_logger
logger = get_logger(__name__, log_level="DEBUG")

if __name__ == '__main__':
    print(__name__)
    logger.debug("🐛", main_process_only=False)
    logger.info("Aloha, World!", main_process_only=False)
    logger.warning("Watch out!", main_process_only=False)
    logger.error("Panic!!!", main_process_only=False)
    logger.critical("D:", main_process_only=False)
```

This will output:
```
>  accelerate launch reproducer.py
__main__
Watch out!
Panic!!!
D:
```

The same behaviour occurs when setting log level via environmental variables.

### Expected behavior

```Shell
We should have two additional lines before the `warning` and after the regular `print`.
```

sgugger:
cc @muellerzr

vvvm23:
Hmm, I did more digging and might be something weird with my env, as the equivalent code with just plain `logging` also does the same thing :/
```
import logging
logger = logging.getLogger(__name__)
logger.setLevel(logging.INFO)

if __name__ == '__main__':
[...]

PR #1579, opened 2023-06-12
fix logger level

caopulan:
[#1178](https://github.com/huggingface/accelerate/issues/1178)
This pull request addresses an issue where the root logger's default level was set to WARNING, making `log_level` in get_logger out of work.


muellerzr (approved):
Thanks!

PR #1580, opened 2023-06-12
Expand `prepare()` doc

muellerzr:
Helps again on a pain-point in https://github.com/huggingface/accelerate/issues/1574#issuecomment-1587719420 by directly stating how to properly send in device placement items and objects to `prepare()` at once

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1092,7 +1092,7 @@ class Accelerator:
 
             device_placement (`list[bool]`, *optional*):
                 Used to customize whether automatic device placement should be performed for each object passed. Needs
-                to be a list of the same length as `args`.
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1109,6 +1109,20 @@ def prepare(self, *args, device_placement=None):
...
         ```
+
+        ```python
+
+        >>> # Assume a model, optimizer, data_loader and scheduler are defined
+        >>> device_placement = [True, True, False, False]
+        >>> # Will place the first to items passed in automatically to the right device
+        >>> # The second two will not
```
sgugger, line 1120: The last sentence is weird here.
```suggestion
        >>> # Will place the first to items passed in automatically to the right device but not the last two.
```

sgugger (approved):
Thanks!

PR #1581, opened 2023-06-12
Fix TPU issues and actually initialize the device

muellerzr:
Per discussion in google chat, the logic in the `is_tpu_available` check is incorrect and actually will not allow XLA to be ran at all. The proper check of just "is cuda available, if so return False" is used instead. This can be relied on as this is still a valid check even on a CPU-only build of torch.

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -88,16 +88,15 @@ def is_fp8_available():
 def is_tpu_available(check_device=True):
     "Checks if `torch_xla` is installed and potentially if a TPU is in the environment"
     # Due to bugs on the amp series GPUs, we disable torch-xla on them
-    if _tpu_available:
-        if _torch_distributed_available:
+    if torch.cuda.is_available():
+        return False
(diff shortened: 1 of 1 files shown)


sgugger (commented):
Are we sure this won't impact things that require xla not to be initialized (like the notebook launcher)?

muellerzr:
It *shouldn't*, because this won't initialize CUDA on a notebook by just checking if CUDA is available (I assume you meant CUDA here). I'll run a notebook to be sure nothing strange happens despite torch cuda not being initialized.

sgugger:
No I mean XLA. The notebook launcher can be used to launch trainings on TPUs in Colab and Kaggle and I'm afraid this will break it.

muellerzr:
Will check and verify, however the logic should still be the same as it was before, just the `_torch_distributed_available` was removed from the if statement

sgugger (commented):
Oh, misread the diff. Yes all good!

sgugger (approved):

Issue #1573, opened 2023-06-10
[BUG] Test failed with accelerate version 0.20.3 and deepspeed version 0.9.3 on single machine with 4x A100 GPU


seraphzl:
### System Info

```Shell
- `Accelerate` version: 0.20.3
- Platform: Linux-3.10.0-1160.71.1.el7.x86_64-x86_64-with-glibc2.17
- Python version: 3.10.9
- Numpy version: 1.23.5
- PyTorch version (GPU?): 2.0.1+cu117 (True)
- PyTorch XPU available: False
- System RAM: 1929.00 GB
- GPU type: NVIDIA A100-SXM4-80GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {'gradient_accumulation_steps': 1, 'offload_optimizer_device': 'none', 'offload_param_device': 'none', 'zero3_init_flag': False, 'zero_stage': 2}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Accelerate version 0.19.0 and deepspeed version 0.9.3:
    4x A100: test OK.
2. Accelerate version 0.20.3 and deepspeed version 0.9.3:
    2x A100: test OK.
    4x A100: test failed.

The output log is as follows.

```Shell
$ accelerate test
Setting ds_accelerator to cuda (auto detect)

Running:  accelerate-launch /home/xxx/apps/miniconda310/envs/ml/lib/python3.10/site-packages/accelerate/test_utils/scripts/test_script.py
stdout: Setting ds_accelerator to cuda (auto detect)
stdout: [09:50:58] WARNING                                                                                                                                                                                          run.py:694
stdout:                     *****************************************
[...]

PR #1586, opened 2023-06-13
Solve failing test on 4 GPU

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1573 by changing the test to:

- Properly give us an extra remainder/odd one
- Use `wait_for_everyone` for a clear trace on what happens if that test fails
- Properly convert the output to a list for the expected output when 4 GPUs are used

(Verified on 4x T4)


sgugger (approved):
Thnaks for the fix!

PR #1593, opened 2023-06-14
Integration tests

muellerzr:
Adds integration CI's as we try to catch issues with accelerate breaking main or released versions for core libraries.

Currently just has `transformers`, will eventually add `skorch`.

Built in a way that lets us quickly add in more tests to run. (and whatever other libs we want to add along the way)

Will monitor the runner integration after it's been merged so I can manually run that afterwards, since it relies on our hosted runner.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -52,6 +52,7 @@ jobs:
     
     - name: Clone and install transformers
       run: |
+        cd ..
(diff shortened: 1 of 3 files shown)


sgugger (approved):
Thanks!

pacman100 (approved):
Thank you @muellerzr for adding this! 🤗

PR #1595, opened 2023-06-14
Bug fix for `check_tied_parameters_in_config`

ain-soph:
In current codes, there is a comma for `has_tied_encoder_decoder`, which makes it a tuple rather than bool. Therefore, the result of `check_tied_parameters_in_config` is always `True`.

For example, `has_tied_encoder_decoder = (False,)`

However, `any([False, (False, ), False]) == True`. We should make it a `bool` rather than `tuple[bool]`


sgugger (approved):
Thanks for the fix (cc @SunMarc )

PR #1596, opened 2023-06-15
Add bitsandbytes quantization

SunMarc:
# What does this PR do?
This PR adds 4-bit and 8-bit quantization from bitsandbytes library in accelerate. For now, it only works with cuda in a distributed setup or not. 
Key features: 
- Quantization config can be set using accelerate config, by passing a `BnbQuantizationPlugin` object to `Accelerator` or passing the args in `accelerate launch `command 
- Replace linear layers by bnb linear layers 
- If we have a `PeftModel` as input model, we replace the targeted linear layers by a quantized linear layer from peft library (useful because we should not train a native quantized model) 
- If the input model is already quantized, we don't modify it. However, we prepare it for training (check `prepare_model_for_kbit_peft_training`) 

## To do
- [ ] write documentation on how to use quantization with accelerate
- [ ] add more tests
- [ ] add examples
- [ ] cleanup


examples/by_feature/quantization.py
```diff
@@ -0,0 +1,88 @@
...
+`accelerate config` and choose your quantization parameters
+`accelerate launch quantization.py`
+"""
+
+
+def main():
+    accelerator = Accelerator()
+    model = AutoModelForCausalLM.from_pretrained(
+        "facebook/opt-1.3b",
+    )
```
sgugger, line 40: Fits in one line.

src/accelerate/utils/bitsandbytes.py
```diff
@@ -0,0 +1,332 @@
...
+    Args:
+        model (`torch.nn.Module`): Input model
+        bnb_config (`dict`'): The bitsandbytes parameters
+
+    Returns:
+        `torch.nn.Module`: The quantized model
+    """
+    if is_transformers_available():
+        from transformers import BitsAndBytesConfig
+        from transformers.utils.bitsandbytes import get_keys_to_not_convert
```
sgugger, line 49: You should raise an error here if transformers is not available. But do we really need the config from Transformers? Can't the plugin be used instead?

Also is `get_keys_to_not_convert` general enough to work for non-transformer models?

src/accelerate/utils/bitsandbytes.py
```diff
@@ -0,0 +1,332 @@
...
+        )
+        torch_dtype = torch.float16
+        model.to(torch_dtype)
+
+    # some modules needs to be in fp32
+    keep_in_fp32_modules = model._keep_in_fp32_modules
+    if keep_in_fp32_modules is not None:
+        for name, param in model.named_parameters():
+            if any(module_to_keep_in_fp32 in name for module_to_keep_in_fp32 in keep_in_fp32_modules):
+                param = param.to(torch.float32)
```
sgugger, line 80: This is specific to Transformers and should be an argument of the function here.

src/accelerate/utils/bitsandbytes.py
```diff
@@ -0,0 +1,332 @@
...
+
+    # set model config
+    config = model.config
```
sgugger, line 83: Same here the model does not necessarily have a config

src/accelerate/utils/bitsandbytes.py
```diff
@@ -0,0 +1,332 @@
...
+
+def has_bnb_layers(model):
+    """Check if we have `bnb.nn.Linear4bit` or `bnb.nn.Linear8bitLt` layers inside our model"""
+    for m in model.modules():
+        if isinstance(m, bnb.nn.Linear4bit) or isinstance(m, bnb.nn.Linear8bitLt):
+            return True
+    return False
+
+
+def prepare_model_for_kbit_peft_training(model, use_gradient_checkpointing=True):
```
sgugger, line 294:
```suggestion
def prepare_model_for_quantized_peft_training(model, use_gradient_checkpointing=True):
```
kbit does not mean anything since k can be equal to 32 and then it's normal training

src/accelerate/utils/bitsandbytes.py
```diff
@@ -0,0 +1,332 @@
...
+    r"""
+    This method wraps the entire protocol for preparing a model before running a training. This includes:
+        1- Cast the layernorm in fp32 2- making output embedding layer require grads 3- Add the upcasting of the lm
+        head to fp32 4- make sure to not modfiy lora layers
```
sgugger, line 298: List is not properly formatted.

sgugger (commented):
Thanks for diving into this. I haven't looked in depth at all the functions yet because I think there is a design issue in the current approach: it requires the model to be fully loaded, which defeats the purpose of saving memory.

I was more thinking of the quantization to be done in a function like `load_checkpoint_and_dispatch` which can take a (potentially empty since on the meta device) model and load the quantized weights inside.
We also shouldn't assume a dependency on Transformers in this integration and some things that Transformers implement as private class attributes for each model should be designed as a function argument here to be more general.

SunMarc:
Thanks for the feedback ! I will close this PR for now and create a new one in the future with the new design as many features introduced here don't make sense with the desired design =)

PR #1597, opened 2023-06-15
Add triggers for CI workflow

muellerzr:
Forgot to add the triggers, pretty sure the `RUN_SLOW=yes` will trickle down through the workflows, will see how the nightly does to check if they're being called.


sgugger (approved):
Thanks!

PR #1603, opened 2023-06-16
Remove GPU safetensors env variable

sgugger:
As pointed out by @Narsil , `safetensors` does not use that env variable anymore, so let's remove it and the log that goes with it.


sgugger:
Merging now to fix the issue, can address any comment in a followup PR.

PR #1609, opened 2023-06-19
reset end_of_dataloader for dataloader_dispatcher

megavaz:
Reset `end_of_dataloader` to `False` at each iteration as in [PR#1562](https://github.com/huggingface/accelerate/pull/1562) but for the `DataLoaderDispatcher`.

### DIFF
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -17,16 +17,16 @@ import unittest
 
 from torch.utils.data import BatchSampler, DataLoader, IterableDataset
 
+from accelerate import Accelerator
 from accelerate.data_loader import (
     BatchSamplerShard,
-    DataLoaderShard,
     DataLoaderDispatcher,
+    DataLoaderShard,
     IterableDatasetShard,
     SkipBatchSampler,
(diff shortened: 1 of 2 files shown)


megavaz:
Without the thing `lr_scheduler.step()` works incorrectly when gradient accumulation is used with `IterableDataset`

tests/test_data_loader.py
```diff
@@ -384,3 +386,13 @@ def test_end_of_dataloader(self):
         # Test it also works on the second iteration
         for idx, _ in enumerate(dataloader):
             self.assertEqual(dataloader.end_of_dataloader, idx == 3)
+
+    def test_end_of_dataloader_dispatcher(self):
+        accelerator = Accelerator()
```
megavaz, line 391: Couldn't make the test work without the line. `AcceleratorState` raises "Please make sure to properly initialize your accelerator via `accelerator = Accelerator()` before using any functionality from the `accelerate` library."

sgugger (approved):
Thanks a lot for catching this!

Issue #1612, opened 2023-06-20
Using ipex: Division by Zero with auto device mapping


TheMrCodes:
# The Problem
The following code dosn't work as aspected it iterate over all devices and looks for "xpu" devices, this is correct for xpu, intel cpu devices (as far as I know) but is a problem for gpu devices because of there simulated device type 'cuda' (this is because of the IPEX for GPUs). The next paragraph contains a test output that confirmed my theory.
https://github.com/huggingface/accelerate/blob/535b52cef22c81a5cd01f971d0a573dd1c73762d/src/accelerate/utils/modeling.py#LL582C1-L585C106

## Programm file thats tests the theory:
``` Python
import torch
import intel_extension_for_pytorch as ipex
print(torch.__version__)
print(ipex.__version__)
[ print(f'[{i}]: {torch.xpu.get_device_properties(i)}') for i in range(torch.xpu.device_count()) ]
print(torch.device(0).type)
```

 ## Produces the following output:
```  Plain Text
No CUDA runtime is found, using CUDA_HOME='/usr'
1.13.0a0+git6c9b55e
1.13.120+xpu
[0]: _DeviceProperties(name='Intel(R) Arc(TM) A770 Graphics', platform_name='Intel(R) Level-Zero', dev_type='gpu, support_fp64=0, total_memory=15473MB, max_compute_units=512)
cuda
```

## My temporary solution
I also tried by implemented a quick and dirty fix that always uses the cuda version and it worked as expected.

# Possible solution
extending the for ... if ... statement in line 585 to check for the device type gpu (probably something like, string torch.xpu.get_device_properties(i).dev_type starting with 'gpu')

sgugger:
Would you like to open a PR to fix this?

abhilash1910:
@TheMrCodes  Thanks for highlighting this issue in arc gpus. I could not find any other fault lines for arc gpus; do let me know if this fixes for you ?

PR #1615, opened 2023-06-20
Fix for arc gpus

abhilash1910:
Addresses #1612  for arc gpus . (Could not find any other arc specific invocation) 
@sgugger


sgugger (approved):
Thanks for fixing!

Issue #1606, opened 2023-06-18
Accelerate test failed on 2 nodes with 8x A800


seraphzl:
### System Info

```Shell
- `Accelerate` version: 0.19.0
- Platform: Linux-3.10.0-1160.81.1.el7.x86_64-x86_64-with-glibc2.17
- Python version: 3.10.10
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1+cu117 (True)
- System RAM: 1007.34 GB
- GPU type: NVIDIA Graphics Device
- `Accelerate` default config:
	- compute_environment: LOCAL_MACHINE
	- distributed_type: DEEPSPEED
	- mixed_precision: no
	- use_cpu: False
	- num_processes: 16
	- machine_rank: 0
	- num_machines: 2
	- main_process_ip: g0001
	- main_process_port: 5001
	- rdzv_backend: static
	- same_network: True
	- main_training_function: main
	- deepspeed_config: {'deepspeed_hostfile': '/data/nfs_share/hostfile', 'deepspeed_multinode_launcher': 'pdsh', 'gradient_accumulation_steps': 1, 'offload_optimizer_device': 'cpu', 'offload_param_device': 'cpu', 'zero3_init_flag': False, 'zero3_save_16bit_model': False, 'zero_stage': 3}
	- downcast_bf16: no
	- tpu_use_cluster: False
	- tpu_use_sudo: False
	- tpu_env: []
```

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

The hostfile is:
```SHELL
g0001 slots=8
g0003 slots=8
```
The command and test message is:
```SHELL
$ accelerate test

[2023-06-18 23:37:09,278] [INFO] [real_accelerator.py:110:get_accelerator] Setting ds_accelerator to cuda (auto detect)

Running:  accelerate-launch /data/nfs_share/xxx/miniconda3/lib/python3.10/site-packages/accelerate/test_utils/scripts/test_script.py
stdout: [2023-06-18 23:37:11,024] [INFO] [real_accelerator.py:110:get_accelerator] Setting ds_accelerator to cuda (auto detect)
[...]

PR #1616, opened 2023-06-20
Fix failing multinode tests

muellerzr:
These particular tests shouldn't be ran on multinode, as the printing etc it relies on is not always able to be recreated well across nodes, and same with the split_between_processes.

Solves https://github.com/huggingface/accelerate/issues/1606

### DIFF
--- a/src/accelerate/test_utils/scripts/test_script.py
+++ b/src/accelerate/test_utils/scripts/test_script.py
@@ -82,21 +82,21 @@ def process_execution_check():
     if accelerator.is_main_process:
         with open(path, "r") as f:
             text = "".join(f.readlines())
-        try:
-            assert text.startswith("Currently in the main process\n"), "Main process was not first"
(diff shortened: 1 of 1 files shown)


src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -526,21 +526,28 @@ def main():
     if state.local_process_index == 0:
         print("**Initialization**")
     init_state_check()
-    if state.local_process_index == 0:
```
sgugger, line 529: Or maybe change this to `process_index == 0` so it's only run on machine 0?

muellerzr: That cleans it up some, but there's still the issue of `state.num_processes` will be 16 on two nodes of 8

sgugger (approved):
Thanks for the fix!

PR #1618, opened 2023-06-20
Doc to md

sgugger:
Like in Transformers. The new GitHub UI is pretty horrible for how it renders MDX (or we are using MDX wrong 😅 ) so this PR puts everything in markdown, so the diff in docs is easier to read :-)


muellerzr (approved):
Sad, but makes sense 😭

Issue #1619, opened 2023-06-21
ValueError: Logging with `tensorboard` requires a `logging_dir` to be passed in


czk32611:
### System Info

```Shell
- `Accelerate` version: 0.20.3
- Platform: Linux-4.14.105-1-tlinux3-0013-x86_64-with-glibc2.17
- Python version: 3.10.6
- Numpy version: 1.23.4
- PyTorch version (GPU?): 2.0.0+cu118 (True)
- PyTorch XPU available: False
- System RAM: 375.51 GB
- GPU type: Tesla V100-SXM2-32GB
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```
from accelerate import Accelerator
accelerator = Accelerator(log_with="tensorboard")
```

I think this error happens because `self.logging_dir` is still referred in https://github.com/huggingface/accelerate/blob/v0.20.3/src/accelerate/accelerator.py#L356, which is supposed to be replaced by `project_dir`?

### Expected behavior

Modify `accelerator.py` so that no error will raise when using other tracker

muellerzr:
Nope, that's expected. You need to create a ProjectConfiguration, or pass in a project_dir. We can clarify this issue, but what it states is still accurate.

czk32611:
Thanks, I see what exactly happened. I got this error if I ran
```
from accelerate import Accelerator
from accelerate.utils import ProjectConfiguration

accelerator_project_config = ProjectConfiguration()
accelerator = Accelerator(log_with="tensorboard", project_dir='./output', project_config=accelerator_project_config)
```
In this way, accelerator.logging_dir is not initiated by  project_dir as in https://github.com/huggingface/accelerate/blob/v0.20.3/src/accelerate/utils/dataclasses.py#L381.


> Nope, that's expected. You need to create a ProjectConfiguration, or pass in a project_dir. We can clarify this issue, but what it states is still accurate.

muellerzr:
Otherwise, to do what you'd like this should fully be fixed via `pip install git+https://github.com/huggingface/accelerate@project-dir`

muellerzr:
[...]

PR #1623, opened 2023-06-21
Fix TensorBoard/logging dir issue

muellerzr:
Fixes an issue where if you passed in a ProjectConfig and a `project_dir` but not a `logging_dir`, the logging_dir wouldn't be set due to th epost_init never being ran. This PR refactors that logic to be a bit more fullproof.

solves https://github.com/huggingface/accelerate/issues/1619


sgugger (approved):
Thanks for the fix!

PR #1624, opened 2023-06-21
Fix transformers sync bug with accumulate

muellerzr:
Adds a new `sync_with_dataloader` arg to the `GradientAccumulationPlugin`, to assist with using it across epochs via `total_batched_samples` (https://github.com/huggingface/transformers/issues/23935#issuecomment-1588134562)

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -857,10 +857,7 @@ class Accelerator:
 
     def _do_sync(self):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
-        if (
-            self.gradient_state.plugin_kwargs.get("sync_with_dataloader", True)
-            and self.gradient_state.end_of_dataloader
-        ):
+        if self.gradient_state.sync_with_dataloader and self.gradient_state.end_of_dataloader:
             self.step = 0
             self.gradient_state._set_sync_gradients(True)
         else:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -857,7 +857,10 @@ class Accelerator:
 
     def _do_sync(self):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
-        if self.gradient_state.end_of_dataloader:
+        if (
+            self.gradient_state.plugin_kwargs.get("sync_with_dataloader", True)
+            and self.gradient_state.end_of_dataloader
+        ):
             self.step = 0
             self.gradient_state._set_sync_gradients(True)
         else:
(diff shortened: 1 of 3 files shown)


sgugger (approved):
SGTM!

src/accelerate/accelerator.py
```diff
@@ -857,7 +857,10 @@ def no_sync(self, model):
 
     def _do_sync(self):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
-        if self.gradient_state.end_of_dataloader:
+        if (
```
pacman100, line 860: This still won't solve the issue. At the`end_of_dataloader`, `_set_sync_gradients` should always be True. The flag should only handle resetting of `self.step`

pacman100 (commented):
Thank you Zach for helping with the issue. I've left a comment

src/accelerate/accelerator.py
muellerzr: With `transformers` we always set `sync_with_dataloader` to `False` so this will actually never happen, and it will only rely on the step

pacman100:
```
# should this be under the accumulate context manager?
                # the `or` condition of `steps_in_epoch <= args.gradient_accumulation_steps` is not covered
                # in accelerate
                if total_batched_samples % args.gradient_accumulation_steps == 0 or (
                    # last step in epoch but step is always smaller than gradient_accumulation_steps
                    steps_in_epoch <= args.gradient_accumulation_steps
                    and (step + 1) == steps_in_epoch
                ):
```

When `steps_in_epoch <= args.gradient_accumulation_steps`, gradients should be synced at the end of dataloader as well as across epochs every gradient_accumulation_steps.

pacman100: E.g. epoch has 3 steps and `gradient_accumulation_steps` is 4. In this case, gradients should be synced at 3 step as that is the end of dataloader as well on 4th step (1st step of second epoch). This is to be on par with Trainer's behaviour.

muellerzr: *just reread your first note*

Will adjust :)

pacman100: No, no, wait, even that would be incorrect for the steps_in_epoch > args.gradient_accumulation_steps. Let this PR be as is. I will explicitly set in Trainer the sync gradients to True for this peculiar case when `steps_in_epoch <= args.gradient_accumulation_steps`

pacman100: Yes, this is still required, I'll raise PR tomorrow using this

muellerzr: Great! Okay. Let me know if we still need https://github.com/huggingface/transformers/pull/24405, otherwise I'll merge this for now thanks 🤗

pacman100 (approved):
Thank you @muellerzr for adding this! 🤗

sumpster:
Fixed the problem for me.
I was running into crashes at epoch boundaries before and installing this pr plus the matching transformers pr (https://github.com/huggingface/transformers/pull/24405) did the trick.

pacman100:
We can close this one as I have added the missing edge case in the above mentioned PR

PR #1625, opened 2023-06-21
Fix workflow

muellerzr:
Fixes the nightly and merge workflows from not running.

Proof here: https://github.com/huggingface/accelerate/actions/runs/5338157693

Which also, we currently have a failure on multi-GPU on both main and pypi versions of accelerate, which I'll look at this week.


sgugger (approved):
Thanks for fixing!

PR #1627, opened 2023-06-22
Set torch.Generator device

lhoestq:
Can help with issues like this one: https://discuss.huggingface.co/t/runtime-error-trainer-api-dataloader-using-cpu-but-expecting-cuda/38746

(not too sure though - I just looked for torch.Generator without device=device in the codebase)

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -740,7 +740,7 @@ def prepare_data_loader(
                 sampler = dataloader.batch_sampler.sampler
             if hasattr(sampler, "generator"):
                 if sampler.generator is None:
-                    sampler.generator = torch.Generator()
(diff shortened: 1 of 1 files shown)


lhoestq:
Maybe we can leave device=None except when the user had set somethign like `torch.set_default_tensor_type('torch.cuda.FloatTensor')`

sgugger:
I don't fully understand how the device of the generator of random numbers for the sampling can impact the device of the batches. Do you have an example that fails before this fix and works after?

lhoestq:
After more thoughts I think the generator needs the device when a user uses `torch.set_default_tensor_type('torch.cuda.FloatTensor')` otherwise it fails.

Don't have a GPU right now to test though

PR #1631, opened 2023-06-22
fixes offload dtype

SunMarc:
# What does this PR do ? 
This PR fixes a problem I had when I tried to offload a tensor to `meta` device using `set_module_tensor_to_device` with a specific dtype. The `dtype` of the meta tensor is not changed by the `dtype` arg. 

When we move back the meta tensor to `cpu` or `cuda`, if we don't indicate the `dtype` ,which is the case when we added the hook, it will take the `dtype` of the `meta` tensor. 
```python
    if value is not None:
        if dtype is None:
            value = value.to(old_value.dtype)
```


sgugger (approved):
Thanks for the fixes!

PR #1633, opened 2023-06-23
Update scheduler.py

mzamini92:
following PR [#23038](https://github.com/huggingface/transformers/pull/23038) @muellerzr I guess we need to add the necessary check for the learning rate scheduler update based on the current batch size. Here's an example of how you can modify the code


muellerzr (commented):
Hi @mzamini92, thanks for your PRs, haven't forgotten about this, I've been thinking on the last few days the best way to go about this. Still not 100% sure this is the right solution, but again some things I'm thinking on. :)

muellerzr:
Please check out this PR, which describes why we can't quite do this :) https://github.com/huggingface/transformers/pull/24521

Issue #1607, opened 2023-06-18
get_balanced_memory() ignores GPU:0 entirely if low_zero=True


poedator:
### System Info

```Shell
ubuntu18, python39, transformers 4292, accelerate=0.20.3
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

There may be a flaw in `get_balanced_memory()` function, which leads to completely ignoring GPU:0 entirely if `low_zero = True`.

TO REPRODUCE:
load Llama 7B model, with accelerate and device_map = "balanced_low_0", 2 GPUS with 8GIB limit on each.
or run:
```
kwargs = {'max_memory': {0: '8GIB', 1: '8GIB'}, 'no_split_module_classes': ['LlamaDecoderLayer'], 'special_dtypes': {}}
get_balanced_memory(model, low_zero=True, dtype=torch.float16, **kwargs)

# returns {0: 0, 1: 8589934592}
# expected {0: ~5GiB, 1: 8589934592}
```
WHAT HAPPENS: 
function location: https://github.com/huggingface/accelerate/blob/v0.20-release/src/accelerate/utils/modeling.py#L541
1) L588: `per_gpu` is not accounting for CUDA:0 capacity.
2) L631-632 sets max_memory[0] to 0
3) L635 attemts to calculate capacity needed on CUDA:0 (although not quite enough)
4) L636 sets max_memory[0] back to 0, because min(min_zero, max_memory[0]) = min(min_zero, 0) = 0
so 0 is returned for CUDA:0 capacity. It leads to no allocation of modules to Cuda:0.

This observation may be related to https://github.com/huggingface/accelerate/issues/1167

I can try to propose a PR if the issue is recognized. However the code for device_map = "balanced_low_0" seems quite convoluted, with logic split between get_balanced_memory() and infer_auto_device_map(). Some guidance from HF would be appreciated.

### Expected behavior

most of the model loaded go GPUs 1+ while GPU0 is utilized, but less, as described in the docs.

sgugger:
`balanced_low_0` will use all the memory of the other GPUs and only put things on GPU 0 if there is not enough, so what you describe doesn't sound like a bug but the intended behavior of this option.

[...]

PR #1634, opened 2023-06-25
fix modeling low zero

abhilash1910:
#1607 
To fill GPU 0 , I think the remaining should be allocated as max(gpu:0 , remaining model memory) [L636 modeling.py]. (Since gpu:0 was initialized as 0, if we use min(), it will always be 0.). I am thinking this should resolve but want an expert opinion.
@sgugger  @muellerzr  need your review .

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -649,7 +649,7 @@ def get_balanced_memory(
 
     if low_zero:
         min_zero = max(0, module_sizes[""] - sum([max_memory[i] for i in range(1, num_devices)]))
-        max_memory[0] = min(min_zero, max_memory[0])
+        max_memory[0] = max(min_zero, max_memory[0])
 
     return max_memory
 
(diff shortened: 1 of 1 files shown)


abhilash1910:
I see, in that case, "max" should pick up the maximum value to be fit under (min_zero,rem model size) ,no? Is there any alternative to the issue 1607?

sgugger (commented):
Actually after re-reading I think the right fix is to put `max_memory[0]` at line 648 when initializing and keep that min.

abhilash1910:
> Actually after re-reading I think the right fix is to put `max_memory[0]` at line 648 when initializing and keep that min.

Are you suggesting like this L648?
```
max_memory[i] = min(0 if low_zero and i == 0 else per_gpu, max_memory[0])
```

sgugger:
No
```
max_memory[i] = min(max_memory[0] if low_zero and i == 0 else per_gpu, max_memory[i])
```
or just ignore i == 0 in the loop if `low_zero` is True.

abhilash1910:
Thanks, this makes more sense now. The max_memory[0] was initializing as 0 which was forcing the issue.

sgugger (approved):
Thanks for iterating!

PR #1637, opened 2023-06-26
fix autocasting bug

pacman100:
# What does this PR do?
1. Fixes https://github.com/huggingface/transformers/issues/24431

Use `MethodType` to preserve the forward function's signature when the forward is a `method` else skip it if is a `function`.  For example, `model.foward` object is a function when using `device_map` and currently it is leading to `AttributeError: 'function' object has no attribute '__func__'`. This PR fixes it.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1331,12 +1331,12 @@ class Accelerator:
                 new_forward = torch.autocast(device_type=self.device.type, dtype=torch.bfloat16)(model_forward_func)
             else:
                 new_forward = torch.cuda.amp.autocast()(model_forward_func)
-            model.forward = MethodType(new_forward, model) if hasattr(model.forward, "__func__") else new_forward
-            model.forward = (
-                MethodType(convert_outputs_to_fp32(model.forward.__func__), model)
-                if hasattr(model.forward, "__func__")
-                else convert_outputs_to_fp32(model.forward)
-            )
+
+            if hasattr(model.forward, "__func__"):
+                model.forward = MethodType(new_forward, model)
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1324,15 +1324,19 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
             else:
-                model.forward = MethodType(torch.cuda.amp.autocast()(model.forward.__func__), model)
-            model.forward = MethodType(convert_outputs_to_fp32(model.forward.__func__), model)
+                new_forward = torch.cuda.amp.autocast()(model_forward_func)
+            model.forward = MethodType(new_forward, model) if hasattr(model.forward, "__func__") else new_forward
+            model.forward = (
+                MethodType(convert_outputs_to_fp32(model.forward.__func__), model)
+                if hasattr(model.forward, "__func__")
+                else convert_outputs_to_fp32(model.forward)
+            )
```
sgugger, line 1339: Can we do this on `new_forward` before wrapping? Would avoid doing the test twice.

sgugger (approved):
Thanks for the fix. Can you put a description that contains more than the link to the issue so that users coming over this PR later understand its goal? Thanks.

src/accelerate/accelerator.py
pacman100: Hello, done.

StevenSong:
thanks for the fix, sorry if I missed the patch schedule somewhere but will the next patch include this/when is the next patch going to be?

February24-Lee:
I really have been waiting for it. ty for fixing it. 🙏

PR #1638, opened 2023-06-26
deepspeed z2/z1 state_dict bloating fix

pacman100:
# What does this PR do?
1. Fixes https://github.com/huggingface/transformers/issues/22822


sgugger (approved):
Thanks!

PR #1639, opened 2023-06-26
Arc 107 gpu fix offloading

abhilash1910:
This mandatory fix is for offloading from cpu to respective xpu /arc devices .
Since ipex is out of pytorch tree , converting tensors back from cpu to devices must require this fix for xpu/arc graphics card devices (cuda is implicit).
@sgugger @muellerzr

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -162,28 +162,34 @@ def set_module_tensor_to_device(
 
     with torch.no_grad():
         if value is None:
-            if is_xpu_available():
-                device = torch.device("xpu")
-            new_value = old_value.to(device)
+            if is_xpu_available() and device != "cpu":
+                xpu_device = torch.device(f"xpu:{device}")
+                new_value = old_value.to(xpu_device)
+            else:
+                new_value = old_value.to(device)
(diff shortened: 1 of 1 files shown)


sgugger (commented):
This function accepts a `device` argument. It should not be changed behind the user's back like this, but called with the appropriate device.

abhilash1910:
Yes but in pytorch there is always a scope of implicit conversion to cuda (from or to meta) because pytorch supports cuda by default. But in case of xpu devices alone, (only for intel devices) if there is a need to offload from cpu to xpu devices this is required. Hence updated the check for only xpu devices.

abhilash1910:
Ok this is a more critical issue related to intel arc gpus (Exclusively) similar to the PR #1615 . Since the device "name" appears as cuda it is throwing the exception. I am working on a proper fix for this and this implies that a condition needs to be applied to check if it is really cuda or not.  Even though arc gpus are xpus, the torch.device(device_id) equals "cuda" . This is an ipex design change following meta's/stock pytorch directives.  Agree the device is automatically specified but a check needs to be there to ensure that it does not break for arc gpus specifically. For core data center gpus (core xpus) it is fine. 
@sgugger  just for context. Keeping it in draft state for now.

Issue #1630, opened 2023-06-22
Disable rich traceback configuration by default


aaron11496:
I am requesting that the behavior of `ACCELERATE_DISABLE_RICH=1` be made the default, so that developers can opt-in instead of opt-out of `rich` traceback decorating. This would avoid the confusion, the work investigate, and need for extra configuration in user's projects.

---

`accelerate` and `rich` are both transitive dependencies of my project: 

* flair > transformers > accelerate
* apache-airflow > rich

Seeing tracebacks formatted with colors while not having done anything to enable that is surprising and confusing. Then tracking down the cause to Accelerate is not intuitive. I was able to by learning by chance that `sys.excepthook` exists, grepping my entire virtual environment for code that sets it, and then following the trail back to `accelerate/utils/rich.py`.

I was able to confirm that Accelerate is the cause with these commands:
```shell
$ python -c 'import flair; import does_not_exist'                        
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ in <module>:1                                                                                    │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ModuleNotFoundError: No module named 'does_not_exist'
```
```shell
$ ACCELERATE_DISABLE_RICH=1 python -c 'import flair; import does_not_exist'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'does_not_exist'
```

`ACCELERATE_DISABLE_RICH` is not documented as far as I can tell. And even if it was, it would be a lot to task to have to set (and _remember_ to set) that env var in every environment forever to satisfy the quirk of one transitive dependency.

muellerzr:
Agreed, I thought we'd made that the default, (disabling) but it was just in Jupyter, not the CLI. Would you like to change that? Otherwise I'll put in a PR tommorow!

aaron11496:
I can wait for your PR and would appreciate it. Thank you!

aaron11496:
LGTM. Thanks @muellerzr!

PR #1640, opened 2023-06-26
Swap disable rich

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1630 by disabling rich by *default*, and deprecates the env var


sgugger (approved):
Thanks!

PR #1641, opened 2023-06-26
add save model

SunMarc:
# What does this PR do ? 

This PR adds the method `save_model()` to `Accelerator` class. This enable user to save any pytorch model :
- with safetensors (set `safe_serialization=True`)
- into multiple checkpoint shards (set `max_shard_size`)

PS1: This will be very useful for users who want to use big model inference or `load_checkpoint_in_model` general with any pytorch model

PS2: This is heavily based on the `save_pretrained` method from `transformers` library (without the `push_to_hub` argument, the `.config` attribute of transformer models and sagemaker flag). I will do a follow PR in transformers to remove duplicates.


src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
     load_location = f"{path}/custom_checkpoint_{index}.pkl"
     logger.info(f"Loading the state of {get_pretty_name(obj)} from {load_location}")
     obj.load_state_dict(torch.load(load_location, map_location="cpu"))
+
+
+def dtype_byte_size(dtype):
```
sgugger, line 235: I think this already exists in utils/modeling ;-)

src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
...
+    """
+    Unique identifier to a tensor storage. Multiple different tensors can share the same underlying storage. For
+    example, "meta" tensors all share the same storage, and thus their identifier will all be equal. This identifier is
+    guaranteed to be unique and constant for this tensor's storage during its lifetime. Two tensor storages with
+    non-overlapping lifetimes may have the same id.
+    """
+    return tensor.device, storage_ptr(tensor), storage_size(tensor)
+
+
+def convert_file_size_to_int(size: Union[int, str]):
```
sgugger, line 274: Same for this one.

src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
...
+        for key in shard.keys():
+            weight_map[key] = shard_file
+
+    # Add the metadata
+    metadata = {"total_size": total_size}
+    index = {"metadata": metadata, "weight_map": weight_map}
+    return shards, index
+
+
+def unwrap_model(model: nn.Module) -> nn.Module:
```
sgugger, line 384: This is in Accelerate already. `save_model` should assume an unwrapped model.

src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
...
+        raise ImportError("`safe_serialization` requires the `safetensors library: `pip install safetensors`.")
+
+    if os.path.isfile(save_directory):
+        logger_simple.error(f"Provided path ({save_directory}) should be a directory, not a file")
+        return
+
+    os.makedirs(save_directory, exist_ok=True)
+
+    # Only save the model itself if we are using distributed training
+    model_to_save = unwrap_model(model)
```
sgugger, line 455: Should be removed. The user should do
```
save_model(accelerator.unwrap_model(model))
```

src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
...
+        save_directory (`str` or `os.PathLike`):
+            Directory to which to save. Will be created if it doesn't exist.
+        is_main_process (`bool`, *optional*, defaults to `True`):
+            Whether the process calling this is the main process or not. Useful when in distributed training like TPUs
+            and need to call this function on all processes. In this case, set `is_main_process=True` only on the main
+            process to avoid race conditions.
+        state_dict (nested dictionary of `torch.Tensor`):
+            The state dictionary of the model to save. Will default to `self.state_dict()`, but can be used to only
+            save parts of the model or if special precautions need to be taken when recovering the state dictionary of
+            a model (like when using model parallelism).
```
sgugger, line 422: I don't think we should support this argument here. This is specific to some things in the Trainer.

src/accelerate/checkpointing.py
```diff
@@ -214,3 +230,322 @@ def load_custom_state(obj, path, index: int = 0):
...
+
+            keys_to_ignore_on_load_missing = None
+            if keys_to_ignore_on_load_missing is not None:
+                found = 0
+                for name in sorted(names):
+                    matches_pattern = any(re.search(pat, name) for pat in keys_to_ignore_on_load_missing)
+                    if matches_pattern and name in state_dict:
+                        found += 1
+                        if found < len(names):
+                            del state_dict[name]
```
sgugger, line 488: This part should be removed here, this is specific to Transformers.

sgugger (commented):
Thanks for adding this! Left a couple of comments.

src/accelerate/checkpointing.py
SunMarc: Fixed [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
SunMarc: Fixed [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
SunMarc: Thx for the fix. I've added that in the description of the function. See fix [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
SunMarc: Fixed [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
SunMarc: my bad ;-) [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
SunMarc: my bad ;-) [8d853b4](https://github.com/huggingface/accelerate/pull/1641/commits/8d853b4c7918360b91877737c363b5722b4c8e48)

src/accelerate/checkpointing.py
```diff
@@ -214,3 +232,236 @@ def load_custom_state(obj, path, index: int = 0):
...
+def _add_variant(weights_name: str, variant: Optional[str] = None) -> str:
+    if variant is not None:
+        splits = weights_name.split(".")
+        splits = splits[:-1] + [variant] + splits[-1:]
+        weights_name = ".".join(splits)
+
+    return weights_name
+
+
+def id_tensor_storage(tensor: torch.Tensor) -> Tuple[torch.device, int, int]:
```
sgugger, line 246: Should probably go in utils/modeling

src/accelerate/checkpointing.py
```diff
@@ -214,3 +232,236 @@ def load_custom_state(obj, path, index: int = 0):
...
+    index = {"metadata": metadata, "weight_map": weight_map}
+    return shards, index
+
+
+def save_model(
+    model: nn.Module,
+    save_directory: Union[str, os.PathLike],
+    is_main_process: bool = True,
+    max_shard_size: Union[int, str] = "10GB",
+    save_function: Callable = torch.save,
```
sgugger, line 338: Same here, we can use `accelerate.save`.

pacman100:
A high level comment. This should be capable of taking `state_dict` too as these are useful for saving models when using DeepSpeed or FSDP.

So, instead of `torch.save(accelerator.get_state_dict(model), "pytorch_model.bin")`, I can do `save_model(accelerator.get_state_dict(model), safe_serialization=True, max_shard_size="1GB")`

sgugger:
In that case maybe it should just be simple and take the wrapped model (not the unwrapped one) so that user can just do `save_model(model, ...)` all the time without having to worry about FSDP/DeepSpeed etc. The method might need to go on the `Accelerator` object to have all the info then.

The goal of Accelerate is to provide a unified API.

src/accelerate/checkpointing.py
```diff
@@ -214,3 +232,236 @@ def load_custom_state(obj, path, index: int = 0):
...
+    safe_serialization: bool = False,
+    variant: Optional[str] = None,
+    keys_to_ignore_on_save: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
SunMarc: removed

src/accelerate/checkpointing.py
SunMarc: moved to function to Accelerator class

src/accelerate/checkpointing.py
```diff
@@ -214,3 +232,236 @@ def load_custom_state(obj, path, index: int = 0):
...
+    # Add the metadata
+    metadata = {"total_size": total_size}
+    index = {"metadata": metadata, "weight_map": weight_map}
+    return shards, index
+
+
+def save_model(
+    model: nn.Module,
+    save_directory: Union[str, os.PathLike],
+    is_main_process: bool = True,
```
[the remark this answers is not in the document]
SunMarc: Used `PartialState() `instead

src/accelerate/checkpointing.py
SunMarc: moved

src/accelerate/accelerator.py
```diff
@@ -2299,6 +2308,124 @@ def save(self, obj, f):
...
+
+        if safe_serialization and not is_safetensors_available():
+            raise ImportError("`safe_serialization` requires the `safetensors library: `pip install safetensors`.")
+
+        if os.path.isfile(save_directory):
+            logger.error(f"Provided path ({save_directory}) should be a directory, not a file")
+            return
+
+        # unwrap model
+        model_to_save = self.unwrap_model(model)
```
sgugger, line 2349: This won't be necessary since `self.get_state_dict` will do it.

sgugger (approved):
Nice, thanks a lot for iterating!

src/accelerate/accelerator.py
SunMarc: Oups ;)

PR #1642, opened 2023-06-27
the model should be in training mode, or else ipex.optimize will thro…

sywangyi:
…w error

like "AssertionError: The optimizer should not be given for inference mode"

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1677,6 +1677,8 @@ class Accelerator:
             elif isinstance(obj, (torch.optim.Optimizer)):
                 optimizer = obj
         if optimizer is not None and model is not None:
+            if not model.training:
+                model.train()
             dtype = torch.bfloat16 if self.state.mixed_precision == "bf16" else torch.float32
             if self.device.type == "xpu" and is_xpu_available():
                 model = model.to(self.device)


sywangyi:
Now accelerate.prepare is called in trainer API. when I use trainer api for finetune, ipex is used, but model.training is false.
which will cause assertion failure.
callstack like
  warnings.warn(
  File "/home/ywan171/alpaca/transformers/examples/pytorch/question-answering/run_qa.py", line 685, in <module>
    main()
  File "/home/ywan171/alpaca/transformers/examples/pytorch/question-answering/run_qa.py", line 626, in main
    train_result = trainer.train(resume_from_checkpoint=checkpoint)
  File "/home/ywan171/alpaca/transformers/src/transformers/trainer.py", line 1536, in train
    return inner_training_loop(
  File "/home/ywan171/alpaca/transformers/src/transformers/trainer.py", line 1647, in _inner_training_loop
    model, self.optimizer = self.accelerator.prepare(self.model, self.optimizer)
  File "/home/ywan171/alpaca/accelerate/src/accelerate/accelerator.py", line 1190, in prepare
  File "/home/ywan171/alpaca/accelerate/src/accelerate/accelerator.py", line 1689, in _prepare_ipex
    model, optimizer = ipex.optimize(model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1")
  File "/home/ywan171/miniconda3/envs/alpaca/lib/python3.9/site-packages/intel_extension_for_pytorch/frontend.py", line 387, in optimize
    assert optimizer is None, "The optimizer should not be given for inference mode"
AssertionError: The optimizer should not be given for inference mode

sywangyi:
@muellerzr

abhilash1910 (approved):
@sgugger  this was initially there in pr 1118, but had asked to remove model.train() since it was happening behind the user. For ipex, we need to have ipex.optimize only in training mode. 
 This check seems correct ,right?

sgugger:
I am not in favor of switching the model in training behing the user's back. You can raise an error if the model is not in training mode, but you cannot change this like that.

sywangyi:
@sgugger could I added model.train() in inner_training_loop before self.accelerator.prepare to avoid such error?

sgugger:
In the `Trainer`, you can definitely add it there yes.

sywangyi:
ok, I will drop the PR and add it in trainer. So, User should make sure the model in correct mode before calling accelerate.prepare.

sywangyi:
pr in transformers merged

Ankur-singh:
I tried setting the model in `train` mode but now I am getting this error `AssertionError: _ipex_optimizer does not support load_state_dict`. I am using pytorch1.13.0+cpu and ipex1.13.0+cpu

Any idea on how do I mitigate this issue?

abhilash1910:
@Ankur-singh are you using transformers installed from source with #24520 ? Secondly I see you are using CPU build of ipex, so could you share your accelerate config( I'm guessing multi cpu?) .

Ankur-singh:
I am using transformers=4.24.0. Yes, CPU build of IPEX. Here is my config file

```
distributed_type: MULTI_CPU
downcast_bf16: 'no'
ipex_config:
  ipex: true
machine_rank: 0
main_process_ip: 12.34.456.789
main_process_port: 29500
main_training_function: main
mixed_precision: bf16
num_machines: 3
num_processes: 1
rdzv_backend: static
same_network: false
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: true
```

abhilash1910:
@Ankur-singh  can you try with ipex 2.0 release for cpu (https://github.com/intel/intel-extension-for-pytorch/releases/tag/v2.0.100%2Bcpu) . The load_state_dict api should already be supported.

PR #1646, opened 2023-06-27
Add skorch to runners

muellerzr:
Adds [skorch](https://github.com/skorch-dev/skortch) to the merge and nightly runners. (all are passing 😄 )

cc @BenjaminBossan


sgugger (approved):
Thanks for adding!

PR #1648, opened 2023-06-27
Change dispatch_model when we have only one device

SunMarc:
# What does this PR do ? 
This PR changes the behavior of the `dispatch_model` function when the `device_map` have only one device. We don't attach the hooks anymore and the user will be able to move the model with `.to()`. This change breaks the case where the `device_map` is `{"":'cpu'}` or `{"":'disk'}` where the hooks will dispatch the weights to the `main_device`. However, this fix will lead to less issues from users who want to use `device_map` as a regular `device` parameter.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -390,7 +390,9 @@ def dispatch_model(
         if device != "disk":
             model.to(device)
         else:
-            raise ValueError("You are trying to offload the whole model to the disk. Please use the `disk_offload` function instead.")
+            raise ValueError(
+                "You are trying to offload the whole model to the disk. Please use the `disk_offload` function instead."
+            )
     model.hf_device_map = device_map
     return model
 
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -390,7 +390,7 @@ def dispatch_model(
         if device != "disk":
             model.to(device)
         else:
-            raise ValueError("You can't offload the whole model to the disk")
+            raise ValueError("You are trying to offload the whole model to the disk. Please use the `disk_offload` function instead.")
     model.hf_device_map = device_map
     return model
 
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -387,12 +387,10 @@ def dispatch_model(
         retie_parameters(model, tied_params)
     else:
         device = device_map.values()[0]
-        if device!="disk":
+        if device != "disk":
             model.to(device)
         else:
-            raise ValueError(
-                "You can't offload the whole model to the disk"
-            )
+            raise ValueError("You can't offload the whole model to the disk")
     model.hf_device_map = device_map
     return model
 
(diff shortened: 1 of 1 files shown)


src/accelerate/big_modeling.py
```diff
@@ -324,63 +324,67 @@ def dispatch_model(
...
-        raise NotImplementedError("Model dispatching requires torch >= 1.9.0")
     # Error early if the device map is incomplete.
     check_device_map(model, device_map)
 
-    if main_device is None:
-        if set(device_map.values()) == {"cpu"} or set(device_map.values()) == {"cpu", "disk"}:
-            main_device = "cpu"
+    # We attach hooks if the device_map have at least 2 different devices. Otherwise, the model in already loaded
+    # in the unique device and the user can decide where to dispatch the model.
+    if len(device_map.values()) > 1:
```
sgugger, line 332: I think it's missing an else with `model.to()` the only device in the dict.

sgugger (commented):
Thanks for working on this!

src/accelerate/big_modeling.py
SunMarc: Oh yeah, we can add that to be sure. However, the model should already be on the right device if the user uses `load_checkpoint_in_model` or load the model from` from_pretrained` method

sgugger: Yes, but the user might be calling `dispatch_model` by themselves.

src/accelerate/big_modeling.py
```diff
@@ -324,63 +324,73 @@ def dispatch_model(
...
-        skip_keys=skip_keys,
-        preload_module_classes=preload_module_classes,
-    )
-    # Attaching the hook may break tied weights, so we retie them
-    retie_parameters(model, tied_params)
+        device = device_map.values()[0]
+        if device != "disk":
+            model.to(device)
+        else:
+            raise ValueError("You can't offload the whole model to the disk")
```
sgugger, line 393: Maybe point the user to the `offload_disk` function in this error message?

sgugger (approved):

src/accelerate/big_modeling.py
SunMarc: nice

zero1zero:
Just for googlers, this presents as this exception when the change is pulled down and using `device_map` like below:
```python
TypeError                                 Traceback (most recent call last)

[<ipython-input-3-34cc1538f877>](https://localhost:8080/#) in <cell line: 14>()
     12 
     13 tokenizer = AutoTokenizer.from_pretrained(model_id)
---> 14 model = AutoModelForCausalLM.from_pretrained(model_id, 
     15                                              quantization_config=bnb_config,
     16                                              device_map={"":0}, trust_remote_code=True)

2 frames

[/usr/local/lib/python3.10/dist-packages/accelerate/big_modeling.py](https://localhost:8080/#) in dispatch_model(model, device_map, main_device, state_dict, offload_dir, offload_index, offload_buffers, skip_keys, preload_module_classes)
    387         retie_parameters(model, tied_params)
    388     else:
--> 389         device = device_map.values()[0]
    390         if device != "disk":
    391             model.to(device)

TypeError: 'dict_values' object is not subscriptable
```

SunMarc:
Thx for testing. Should be fixed in this [PR](https://github.com/huggingface/accelerate/pull/1651).

PR #1653, opened 2023-06-28
change the import place to avoid import error

pacman100:
### What does this PR do?
1. DeepSpeed version check wasn't happening before an import which requires versions above or equal to it. 
2. This PR changes the place of such import so that the version error is raised appropriately. 
3. Fixes https://huggingface.slack.com/archives/C02CH2YP4EQ/p1687909933322059

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -97,7 +97,6 @@ from .utils.constants import FSDP_PYTORCH_VERSION
 
 if is_deepspeed_available():
     import deepspeed
-    from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
 
     from .utils import (
         DeepSpeedEngineWrapper,
@@ -2834,6 +2833,8 @@ class Accelerator:
                         "To save the full checkpoint, run `model.save_checkpoint(save_dir)` and use `zero_to_fp32.py` to recover weights."
                     )
             else:
+                from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
+
                 state_dict = clone_tensors_for_torch_save(self.unwrap_model(model).state_dict())
         else:
             if unwrap:


src/accelerate/accelerator.py
```diff
@@ -2834,6 +2833,8 @@ def get_state_dict(self, model, unwrap=True):
                         "To save the full checkpoint, run `model.save_checkpoint(save_dir)` and use `zero_to_fp32.py` to recover weights."
                     )
             else:
+                from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
```
ydshieh, line 2836: I am wondering what is ther relation between `self.deepspeed_config["zero_optimization"]["stage"] == 3` and `deepspeed == 0.9.3`.

The PR description says:

> This PR changes the place of such import so that the version error is raised appropriately.

But isn't the goal to avoid such `import` and `usage` of `clone_tensors_for_torch_save` when `deepspeed < 0.9.3`?

ydshieh:
BTW, we should avoid the links to slack threads. Either copy paste (part of) the discussion or link to a failing job page.

Here is the erro copied

```bash
src/transformers/testing_utils.py:42: in <module>
    from .deepspeed import is_deepspeed_available
src/transformers/deepspeed.py:37: in <module>
    from accelerate.utils.deepspeed import HfDeepSpeedConfig as DeepSpeedConfig
/usr/local/lib/python3.8/dist-packages/accelerate/__init__.py:3: in <module>
    from .accelerator import Accelerator
/usr/local/lib/python3.8/dist-packages/accelerate/accelerator.py:91: in <module>
    from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
E   ImportError: cannot import name 'clone_tensors_for_torch_save' from 'deepspeed.checkpoint.utils' (/usr/local/lib/python3.8/dist-packages/deepspeed/checkpoint/utils.py)
```

sgugger (approved):
Thanks for the fix!

PR #1654, opened 2023-06-28
Update dataloader

yuxinyuan:
This PR does the following:

1. The current ``SkipDataLoader`` and ``SkipBatchSampler`` keep skipping batches every epoch. I think it's more useful if they only skip batches for the first epoch

2. PyTorch DataLoader adds forward compatibilities for DataPipes. In particular, it broadcasts a shared seed across all dist processes to shuffle a datapipe when ``__iter__`` is called (See [here](https://github.com/pytorch/pytorch/blob/89fcfc1b8c449f072dbd694c55d790e33c19b781/torch/utils/data/dataloader.py#L574)). This causes ``DataLoaderDispatcher`` to fall into an infinite wait, since ``DataLoaderDispatcher`` only creates an iterator for the main process.

3. Currently, ``prepare_data_loader`` re-builds a new DataLoader and completely discards the user's DataLoader. This could be problematic when a user subclass ``DataLoader`` and adds some custom behavior to the DataLoader. This PR proposes to turn ``DataLoaderShard`` and ``DataLoaderDispatcher`` into a wrapper that wraps user's DataLoader. They should hehave exactly the same as before except that ``DataLoaderShard`` and ``DataLoaderDispatcher`` are no longer ``DataLoader`` instances (e.g. isinstance check will fail)

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -347,6 +347,10 @@ class _DataLoaderWrapper:
             self.__dict__[attr] = val
         setattr(self.dataloader, attr, val)
 
+    def __len__(self):
+        # Make `len()` work with dataloader wrapper
+        return len(self.dataloader)
+
 
 class DataLoaderShard(_DataLoaderWrapper, DataLoaderStateMixin):
     """
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -833,7 +833,12 @@ def prepare_data_loader(
 def _dataloader_forcibly_setattr(dataloader, name, val):
     # NOTE After initialization, `DataLoader` prevents one from updating its internal attrs
     # We need to be careful of what we are doing here.
-    super(DataLoader, dataloader).__setattr__(name, val)
+    if isinstance(dataloader, DataLoader):
+        super(DataLoader, dataloader).__setattr__(name, val)
+    elif isinstance(dataloader, _DataLoaderWrapper):
+        _dataloader_forcibly_setattr(dataloader.dataloader, name, val)
+    else:
+        raise NotImplementedError
 
 
 class SkipBatchSampler(BatchSampler):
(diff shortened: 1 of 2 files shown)


muellerzr:
@yuxinyuan could you run `make style; make quality`? This will make the quality check pass

sgugger:
This PR changes way too many things to be merged as is. In particular we never advertise that we support custom subclasses of DataLoader from users (we don't, we support PyTorch dataloaders) so 3 is not a road we want to get in.

For 1, the skip_batches function builds a new dataloader that is shorter, this is the way the API was released/advertised but we won't break backward compatibility to change its behavior.

As for 2, we can explore a targetted fix in its own PR, with a clear regression test showing us the problem it fixes.

yuxinyuan:
> This PR changes way too many things to be merged as is. In particular we never advertise that we support custom subclasses of DataLoader from users (we don't, we support PyTorch dataloaders) so 3 is not a road we want to get in.
> 
> For 1, the skip_batches function builds a new dataloader that is shorter, this is the way the API was released/advertised but we won't break backward compatibility to change its behavior.
> 
> As for 2, we can explore a targetted fix in its own PR, with a clear regression test showing us the problem it fixes.



For 1, yes it's breaking backward compatibility. However, looking at the [doc](https://huggingface.co/docs/accelerate/v0.20.3/en/package_reference/accelerator#accelerate.Accelerator.skip_first_batches), It seems to me that ``skipped_dataloader`` is only useful for one epoch (do people skip dataloader besides resuming from a checkpoint ?)

So, from my point of view, changing the behavior of ``skip_first_batches`` can help users simplify their training code. e.g users can just do the following:
```
loader = DataLoader(...)
loader = accelerator.prepare(loader)
loader = accelerator.skip_first_batches(loader, num_batches=...)

for epoch in range(last_epoch, total_training_epoch):
    for batch in loader:
        # Training logic
```

Does this make sense to your team?

Issue #1649, opened 2023-06-27
Multi-GPU training issue - Failing to specify which GPUs to use (socket error: errno 98)


ShashwatNigam99:
### System Info

```Shell
- `Accelerate` version: 0.20.3
- Platform: Linux-5.3.18-57-default-x86_64-with-glibc2.31
- Python version: 3.10.11
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1+cu117 (True)
- PyTorch XPU available: False
- System RAM: 503.54 GB
- GPU type: NVIDIA A100 80GB PCIe
- `Accelerate` default config:
        Not found
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am running example scripts in the diffusers library which use the accelerate library.

1. Suppose, I have a multi-gpu training running on 2 GPUs out of the 4 available to me (launched using accelerate, working well).
Launched using-
`accelerate launch --mixed_precision="fp16" --multi_gpu --num_processes 2 train_controlnet.py`
So nvidia-smi looks something like this-
![Capture](https://github.com/huggingface/accelerate/assets/30972206/dcaf0d1a-b501-4576-9a82-48e1347b0e14)

2. Now I try to run another multi-gpu process, by specifying the GPUs to specify (wish to use the idle gpus: 2 and 3):
`accelerate launch --mixed_precision="fp16" --multi_gpu --gpu_ids="2,3" --num_processes 2  train_controlnet.py `
I get this error-
`RuntimeError: The server socket has failed to listen on any local network address. The server socket has failed to bind to [::]:29500 (errno: 98 - Address already in use). The server socket has failed to bind to 0.0.0.0:29500 (errno: 98 - Address already in use).`

Have also tried explicitly specifying:
`export CUDA_VISIBLE_DEVICES="2,3"` before calling accelerate, getting the same error.

I am not able to launch single-gpu training as well. For example if I run any of these:
`accelerate launch --mixed_precision="fp16"  train_controlnet.py`
`accelerate launch --mixed_precision="fp16" --gpu_ids="2,3" train_controlnet.py `
`accelerate launch --mixed_precision="fp16" --gpu_ids="2" train_controlnet.py `

[...]

PR #1656, opened 2023-06-28
Check for port usage before launch

muellerzr:
Helps https://github.com/huggingface/accelerate/issues/1649 by checking under the following circumstances:

- If a user has multiple `accelerate launch` they wish to perform and is training on multi-GPU (or deepspeed)
- See if the connection port `torch.distributed` relies on for communication is already in use
- If so, direct them to specify a different communication port

This directly helps catch the following error before it occurs:

```python
RuntimeError: The server socket has failed to listen on any local network address. The server socket has failed to bind to [::]:29500 (errno: 98 - Address already in use). The server socket has failed to bind to 0.0.0.0:29500 (errno: 98 - Address already in use).
```

And replaces it with:
```python
  File "/home/zach_mueller_huggingface_co/miniconda3/envs/accelerate/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/commands/launch.py", line 960, in launch_command
    multi_gpu_launcher(args)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/commands/launch.py", line 639, in multi_gpu_launcher
    current_env = prepare_multi_gpu_env(args)
  File "/home/zach_mueller_huggingface_co/accelerate/src/accelerate/utils/launch.py", line 133, in prepare_multi_gpu_env
    raise ConnectionError(
ConnectionError: Tried to launch distributed communication on port `29500`, but another process is utilizing it. Please specify a different port (such as using the `--master_port` flag or specifying a different `main_process_port` in your config file) and rerun your script.
```

### DIFF
--- a/src/accelerate/utils/__init__.py
+++ b/src/accelerate/utils/__init__.py
@@ -156,6 +156,7 @@ from .memory import find_executable_batch_size, release_memory
 from .other import (
     extract_model_from_parallel,
     get_pretty_name,
+    is_port_in_use,
     merge_dicts,
     patch_environment,
     save,
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -24,7 +24,7 @@ from ..commands.config.config_args import SageMakerConfig
 from ..commands.config.config_utils import DYNAMO_BACKENDS
 from ..utils import DynamoBackend, PrecisionType, is_ipex_available, is_torch_version, is_xpu_available
 from ..utils.constants import DEEPSPEED_MULTINODE_LAUNCHERS
-from ..utils.other import merge_dicts
+from ..utils.other import is_port_in_use, merge_dicts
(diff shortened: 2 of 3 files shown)


sgugger (approved):
Nice idea, thanks!

liu-zichen:
Thanks for your work! 
But I encountered some problems and had some doubts.
When multi-machines share the same local_host, they can never launch successfully (gpus aren't shared across machines but network is shared, so I need to launch multi-times on the same localhost).

Here are my doubts:
Why do all machines check port use in local_host?
Isn't it only machine_rank 0 that needs to be checked? 
Also, why check local_host instead of main_process_ip?

Very much looking forward to your reply. Thanks!

PR #1657, opened 2023-06-28
Also hint at using the next port directly

muellerzr:
Ross gave a wonderful tip:

> You can avoid this if you explicitly specify the rdzv backend with `localhost:0`

So for us, we can tell the user they can specify it to `0`, and the next port will be used.


sgugger (approved):
Thanks!

PR #1658, opened 2023-06-28
Bump integration python version tests

muellerzr:
Transformers removed support for 3.7 today. Until we decide if Accelerate should do something similar, this PR changes the workflow to use 3.8 for the integration tests


pacman100 (approved):
Thank you @muellerzr for quickly fixing this! LGTM!

PR #1659, opened 2023-06-28
docker cpu py version

muellerzr:
Bumps the cpu python version of the docker image to 3.8 as well (GPU was already there)


sgugger (approved):
Thanks!

PR #1660, opened 2023-06-28
[`BigModeling`] Final fix for dispatch int8 and fp4 models

younesbelkada:
# What does this PR do?

Fixes a silent bug regarding buffers that are not in the state dict of the model and when using quantized models. Currently the following snippet fails:

```python
from transformers import AutoModelForCausalLM, AutoConfig, AutoTokenizer
import torch

model_path="ybelkada/gpt2-xl-8bit"

config = AutoConfig.from_pretrained(model_path)
model = AutoModelForCausalLM.from_pretrained(model_path, load_in_8bit=True)

tokenizer = AutoTokenizer.from_pretrained(model_path)

input_text = "Describe the solar system."
input_ids = tokenizer(input_text, return_tensors="pt").input_ids.to("cuda")

outputs = model.generate(input_ids, max_length=10)
print(tokenizer.decode(outputs[0]))
```

With the following error:
```bash
  File "/home/younes_huggingface_co/code/transformers/src/transformers/models/gpt2/modeling_gpt2.py", line 203, in _attn
    attn_weights = torch.where(causal_mask, attn_weights.to(attn_weights.dtype), mask_value)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

Because in https://github.com/huggingface/accelerate/pull/1652 we introduced a check to do nothing in case the model is quantized. This leads to buffers that are not in the state dict being on CPU.

As the `.to` operation is not supported for quantized models, the fix is to always force-dispatch the model if the model is quantized. 

cc @sgugger @SunMarc 

Currently the corresponding test is implemented in https://github.com/huggingface/transformers/pull/24543 as it relies on the main branch of transformers.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -327,9 +327,13 @@ def dispatch_model(
     # Error early if the device map is incomplete.
     check_device_map(model, device_map)
 
+    # for backward compatibility
+    is_quantized = getattr(model, "is_quantized", False) or getattr(model, "is_loaded_in_8bit", False)
+
     # We attach hooks if the device_map have at least 2 different devices. Otherwise, the model in already loaded
     # in the unique device and the user can decide where to dispatch the model.
-    if len(device_map.values()) > 1:
+    # If the model is quantized, we always force-dispatch the model
+    if (len(set(device_map.values())) > 1) or is_quantized:
         if not is_torch_version(">=", "1.9.0"):
             raise NotImplementedError("Model dispatching requires torch >= 1.9.0")
 
@@ -387,11 +391,9 @@ def dispatch_model(
         retie_parameters(model, tied_params)
     else:
         device = list(device_map.values())[0]
-        # for backward compatibility
-        is_quantized = getattr(model, "is_quantized", False) or getattr(model, "is_loaded_in_8bit", False)
         if device != "disk" and not is_quantized:
             model.to(device)
-        elif not is_quantized:
+        else:
             raise ValueError(
                 "You are trying to offload the whole model to the disk. Please use the `disk_offload` function instead."
             )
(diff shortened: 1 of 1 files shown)


src/accelerate/big_modeling.py
```diff
@@ -327,9 +327,13 @@ def dispatch_model(
...
     check_device_map(model, device_map)
 
+    # for backward compatibility
+    is_quantized = getattr(model, "is_quantized", False) or getattr(model, "is_loaded_in_8bit", False)
+
     # We attach hooks if the device_map have at least 2 different devices. Otherwise, the model in already loaded
     # in the unique device and the user can decide where to dispatch the model.
-    if len(device_map.values()) > 1:
+    # If the model is quantized, we always force-dispatch the model
+    if (len(set(device_map.values())) > 1) or is_quantized:
```
younesbelkada, line 336: Just to confirm, if you want to check the number of devices maybe `set(xxx)` was more appropriate? (reading from the comments above)

sgugger (approved):
Thanks for the fix!

src/accelerate/big_modeling.py
SunMarc: works for me

src/accelerate/big_modeling.py
```diff
@@ -387,11 +391,9 @@ def dispatch_model(
         retie_parameters(model, tied_params)
     else:
         device = list(device_map.values())[0]
-        # for backward compatibility
-        is_quantized = getattr(model, "is_quantized", False) or getattr(model, "is_loaded_in_8bit", False)
         if device != "disk" and not is_quantized:
```
SunMarc, line 394:
```suggestion
        if device != "disk":
```

SunMarc: The condition above should be enough

SunMarc (approved):
LGTM !

src/accelerate/big_modeling.py
younesbelkada: Awesome, thanks for double checking!

PR #1664, opened 2023-06-29
remove safetensor dep on shard_checkpoint

SunMarc:
# What does this PR do ? 
This PR removes the dependency on `safetensors` when people uses the method `save_model` from `Accelerator` even when the attribute `safe_serialization` was set to `False`


sgugger (commented):
Since we have to copy them, can we group them in the function that uses them (so we do only one try except for the storage.

sgugger (approved):
Very clean, thanks!

PR #1670, opened 2023-06-30
Doc big model inference

SunMarc:
# What does this PR do? 
This PR modifies replaces the example in big model inference documentation so that we use a model that is not from the transformers library. Tested the new example in this [colab](https://colab.research.google.com/drive/1Mj6YOrETdrCzLWjEBjpC-pt1DvquMEW0?usp=sharing).

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -22,7 +22,7 @@
   - local: usage_guides/training_zoo
     title: Example Zoo
   - local: usage_guides/big_modeling
-    title: How perform inference on large models with small resources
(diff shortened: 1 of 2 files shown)


sgugger (commented):
LGTM, thanks! Can you add a link to the doc in Transformers for Transformer models, so users know to use `from_pretrained` with `device_map="auto"`?

SunMarc:
> LGTM, thanks! Can you add a link to the doc in Transformers for Transformer models, so users know to use `from_pretrained` with `device_map="auto"`?

Ofc =D

sgugger (approved):

PR #1672, opened 2023-06-30
replace save funct in doc

SunMarc:
# What does this PR do ? 
This PR adds the `save_model` method in the `Accelerator` documentation.


sgugger (approved):
Thanks!

PR #1678, opened 2023-07-03
update readme in examples

statelesshz:
This PR updates the readme in exmaples, when using fp16(mixed-precision), we should pass in `--mixed_precision fp16` rather than `--fp16`

If you run `nlp_examle.py` with `--fp16`
 ```python
python ./nlp_example.py --fp16
 ```
will get the following error message:
```text
usage: nlp_example.py [-h] [--mixed_precision {no,fp16,bf16,fp8}] [--cpu]
nlp_example.py: error: unrecognized arguments: --fp16
```
@sgugger


muellerzr (approved):
Thanks!

PR #1679, opened 2023-07-03
fix bnb tests

SunMarc:
# What does this PR do ? 
This PR fixes the bnb tests that did not pass the Accelerate nightly single_gpu and multi_gpu test. I added decorators to function that should only be run on multi_gpu. I also fixed the 8-bit serialization tests because of the new version of bnb (check this PR for more [details](https://github.com/huggingface/transformers/pull/24416))


muellerzr (commented):
Rather than skip outright, can we set a maximum version requirement for now? And probably modify the docker image (or setup.py(?)) To have a lower bnb requirement?

muellerzr (approved):
Thanks for the fix!

PR #1682, opened 2023-07-04
Move mixed precision wrapping ahead of DDP/FSDP wrapping

ChenWu98:
This PR aims to fix a potential issue in `Accelerator.prepare_model`. In the current code, `_original_forward` is stored before mixed precision wrapping is applied, right after DDP/FSDP wrapping; therefore, `_original_forward` is no longer an attribute of the original model. This may cause an issue in the `extract_model_from_parallel`, which tries to get the `_original_forward` attribute of the original model. 

I ran a quick test for DDP + fp16 training after this commit, and it looked good to me; however, I'm not sure if it will have unexpected effects in other cases. Looking forward to hearing your opinion!

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1287,10 +1287,6 @@ class Accelerator:
                 )
         elif device_placement and not has_hf_device_map:
             model = model.to(self.device)
-        
-        # Store the original forward function before DDP/FSDP wraps it
-        if self.native_amp or self.mixed_precision == "fp8":
-            model._original_forward = model.forward
 
         if not evaluation_mode:
             if self.distributed_type in (DistributedType.MULTI_GPU, DistributedType.MULTI_XPU):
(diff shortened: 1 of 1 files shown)


pacman100:
Hello, thank you for the PR, can you add a test which should be fixed by this PR, i.e., which would fail without this PR?

ChenWu98:
Sure, the following steps can demonstrate the issue. A runtime error `RuntimeError: expected scalar type Half but found Float` should be raised without this PR.

1. Create a file named `unwrap_test.py` with content:

```python
import torch

accelerator = Accelerator(mixed_precision="fp16")

model = torch.nn.Linear(2, 4)
model_with_fp32_wrapper = accelerator.unwrap_model(model, keep_fp32_wrapper=True)

# Run forward with fp16 as input.
# When the model is with mixed precision wrapper, no error will be raised.
input_tensor = torch.Tensor([1, 2]).to(dtype=torch.float16, device=accelerator.device)
output = model_with_fp32_wrapper(input_tensor)
```

2. Run launch (with at least two GPUs):

```bash
accelerate launch --multi_gpu --num_processes=2 unwrap_test.py
```

abhilash1910 (commented):
This doesnt break xpus, lgtm. I guess test_utils.py can be modified to add in ``` convert_outputs_to_fp32=True```.

ChenWu98:
@muellerzr Thanks for the explanation! I modified `src/accelerate/test_utils/scripts/test_script.py` to add this test.

pacman100 (approved):
Thank you @ChenWu98 for adding this, LGTM!

sgugger (approved):
Thanks for your contribution!

PR #1689, opened 2023-07-05
Update Apex build command

janEbert:
Due to pip having deprecated `--global-option` and `--build-option`, the new build command will be highly relevant in the future (removal is planned for 23.3, see https://github.com/pypa/pip/issues/11859).

See also https://github.com/NVIDIA/apex/pull/1690 for why the command is different from the Apex README.


muellerzr:
CC @pacman100

janEbert:
Actually this may not be the correct command after all, see https://github.com/NVIDIA/apex/issues/1679#issuecomment-1622863439.
It suffers from the same deprecations.

janEbert:
If NVIDIA/apex#1692 is accepted as the official solution, the current state will be fine to use.

PR #1690, opened 2023-07-06
Fix workflow CI

muellerzr:
Replaces fetch with pull so that it's fully up-to-date

### DIFF
--- a/.github/workflows/self_hosted_integration_tests.yml
+++ b/.github/workflows/self_hosted_integration_tests.yml
@@ -43,7 +43,7 @@ jobs:
         run: 
           source activate accelerate;
           git config --global --add safe.directory '*';
-          git checkout main && git fetch && git checkout ${{ github.sha }};
+          git checkout main && git pull && git checkout ${{ github.sha }};
           pip install -e .;
 
       - name: Update transformers clone & pip install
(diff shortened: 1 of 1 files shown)


muellerzr:
proof: https://github.com/huggingface/accelerate/actions/runs/5478977182

muellerzr:
cc @ydshieh just so we can get someones eyes on this soon and merge so main can be green again :)

muellerzr:
As for why this fails: on the self hosted runners it has a static/persistent version of the repo it uses, so fetch isn't actually doing much to it, vs when we have a typical gh actions storage setup where it pulls down and clones it each time

muellerzr:
@ydshieh here we want the latest main commit always :) These are for our outside integrations, where we want to make sure that e.g. the integration with transformers should always pass on main for the tests we wish to run, esp after putting in a fix, so thus we build from main

ydshieh:
OK if we only test on `main`. But does the CI will be canceled if a new commit is pushed to `main`?

If not (i.e. not cancelled automatically), we should stick to the commit sha that triggers that CI.

ydshieh:
Nice!

(Sorry, I didn't realize there are 3 libraries involved here. I was just thinking your are talking about `accelerate`)

muellerzr:
No worries, that's the fun with these new test suites 😅 Otherwise looks good to you for merge? 😄

ydshieh (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #1626, opened 2023-06-22
Bnb quantization

SunMarc:
# What does this PR do?
This PR adds 4-bit and 8-bit quantization from bitsandbytes library in accelerate. You can now quantize any pytorch model using the `quantize_model` function. This function can take as input an empty model or a loaded model.

```python
model_name =  "marcsun13/bloom-1b7_with_lm_head" 
with init_empty_weights():
     model_8bit = AutoModelForCausalLM.from_config(AutoConfig.from_pretrained(model_name))
weights_location = hf_hub_download(model_name, "pytorch_model.bin")
bnb_quantization_config = BnbQuantizationConfig(load_in_8bit=True)

model_8bit = quantize_model(model_8bit, 
                            bnb_quantization_config, 
                            weights_location=weights_location, 
                            device_map="auto",
                            no_split_module_classes=['BloomBlock']) 
  ```

Furthermore, you can also save and load a 8-bit model (4-bit is currently not supported)
```python
save_model(model_8bit,"path/to/save")
with init_empty_weights():
      # same config as the one you passed before
      model_8bit_from_save = AutoModelForCausalLM.from_config(AutoConfig.from_pretrained(model_name))
model_8bit_from_save = quantize_model(model_8bit_from_save,  
                                      bnb_quantization_config, 
                                      weights_location="path/to/save", 
                                      device_map="auto")
  ```

# Work to do
- [x] - 8bit serialization 
- [x] - disk offload issue

PR #1691, opened 2023-07-06
Fix 4bit test

SunMarc:
# What does this PR do ? 
This PR fixes a bug introduced in this [PR](https://github.com/huggingface/accelerate/pull/1626). The following test `tests/bnb/test_4bit.py::Bnb4bitTestMultiGpu::test_multi_gpu_loading` in `transformers` library fails because the model is not dispatched on multiple gpus when using` device_map='balanced'.` This behavior happens because I forgot to remove the duplicated class `CustomDtype` (both in dataclass and modeling). Hence, we were not able to get the right size occupied by the 4-bit module in `dtype_byte_size`, resulting in an unbalanced device_map.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import contextlib
-import enum
 import gc
 import inspect
 import json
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Thanks a lot for fixing @SunMarc !

Issue #1683, opened 2023-07-04
failed to run accelerate config in colab


Erickrus:
### System Info

```Shell
accelerate version: 0.20.3, OS: ubuntu, python version: 3.11, Google Colab Environment
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

In the colab notebook:

```shell
%cd /content
!git clone https://github.com/SuLvXiangXin/zipnerf-pytorch

%cd /content/zipnerf-pytorch
!pip3 install -r requirements.txt
!pip3 install ./gridencoder
!git clone https://github.com/NVlabs/nvdiffrast
!pip3 install ./nvdiffrast
CUDA='cu118'
!pip3 install torch-scatter -f https://data.pyg.org/whl/torch-2.0.0+{CUDA}.html


!mkdir -p /content/zipnerf-pytorch/data
!cd data && wget http://storage.googleapis.com/gresearch/refraw360/360_v2.zip
!cd data && unzip 360_v2.zip

!accelerate config
```

### Expected behavior

It is impossible to run `accelerate config` on colab. This means following training step could not be completed. 
Other solution is `/root/.cache/huggingface/accelerate/default_config.yaml`, need understand the configuration in advance.
After some works, I got this solution finally:

accelerate/commands/menu/selection_menu.py

L18: 
```python3
import sys
input2 = input
in_colab = False
try:
    from google.colab import drive
    in_colab = True
except:
    pass
```

L118:
```python3
                if in_colab:
                    choice = int(input2())
                else:
                    choice = self.handle_input()
```

muellerzr:
Correct, at the current time there isn't a way to get it working, however your solution seems to work alright. Would you like to open a PR with it @Erickrus? Otherwise I can work on a solution with it over the next few days.

Erickrus:
I dont know what's wrong with the quality check. It seems stuck here

Erickrus:
This issue is resolved

PR #1692, opened 2023-07-07
Fixes for issue #1683: failed to run accelerate config in colab

Erickrus:
Fixes for issue #1683: failed to run accelerate config in colab

### DIFF
--- a/src/accelerate/commands/menu/selection_menu.py
+++ b/src/accelerate/commands/menu/selection_menu.py
@@ -16,7 +16,14 @@
 Main driver for the selection menu, based on https://github.com/bchao1/bullet
 """
 import sys
-
+input2 = input
+in_colab = False
+try:
+    from google.colab import drive
+    in_colab = True
+except:
+    pass
+    
 from . import cursor, input
 from .helpers import Direction, clear_line, forceWrite, linebreak, move_cursor, reset_cursor, writeColor
 from .keymap import KEYMAP
@@ -115,7 +122,10 @@ class BulletMenu:
         move_cursor(len(self.choices) - self.position, "UP")
         with cursor.hide():
             while True:
-                choice = self.handle_input()
+                if in_colab:
+                    choice = int(input2())
+                else:
+                    choice = self.handle_input()
                 if choice is not None:
                     reset_cursor()
                     for _ in range(len(self.choices) + 1):
--- a/src/accelerate/commands/menu/selection_menu.py
+++ b/src/accelerate/commands/menu/selection_menu.py
@@ -24,7 +24,11 @@ from .helpers import Direction, clear_line, forceWrite, linebreak, move_cursor, 
 from .keymap import KEYMAP
 
 
-in_colab = _is_package_available("google.colab")
+in_colab = False
+try:
+    in_colab = _is_package_available("google.colab")
+except ModuleNotFoundError:
+    pass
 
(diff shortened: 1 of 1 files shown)


Erickrus:
How to reformat a file? which tool or standard did you use ?

muellerzr:
@Erickrus please do `pip install -e .[quality]; make style; make quality;`. This will install the right versions of black, ruff, and hf-doc-builder, and run the right quality setup

Erickrus:
For you to decide it can be merged to the base branch @muellerzr

src/accelerate/commands/menu/selection_menu.py
```diff
@@ -15,13 +15,22 @@
...
 import sys
 
 from . import cursor, input
 from .helpers import Direction, clear_line, forceWrite, linebreak, move_cursor, reset_cursor, writeColor
 from .keymap import KEYMAP
 
 
+in_colab = False
+try:
+    in_colab = pkgutil.find_loader("google.colab") is not None
```
sgugger, line 29: Please use [this helper function](https://github.com/huggingface/accelerate/blob/64d7b58c4415a5714c25caa5acb428bec26f94e1/src/accelerate/utils/imports.py#L42) here to be consistent with the rest of the library code.

Erickrus: > Please use [this helper function](https://github.com/huggingface/accelerate/blob/64d7b58c4415a5714c25caa5acb428bec26f94e1/src/accelerate/utils/imports.py#L42) here to be consistent with the rest of the library code.

Thanks for your advice. I'm looking into it. 
a) Can I use _is_package_available() outside accelerate/utils/imports.py ? 
```python
import builtins
import sys

from ...utils.imports import _is_package_available
from . import cursor, input
from .helpers import Direction, clear_line, forceWrite, linebreak, move_cursor, reset_cursor, writeColor
from .keymap import KEYMAP


in_colab = _is_package_available("google.colab")
```

b) Or I need create another method in imports.py , e.g. 
```
def is_colab_available():
    return _is_package_available("google.colab")
```
puzzled ...

sgugger: Yes you can do a) :-)

src/accelerate/commands/menu/selection_menu.py
```diff
@@ -15,13 +15,18 @@
...
+import builtins
 import sys
 
+from ...utils.imports import _is_package_available
 from . import cursor, input
 from .helpers import Direction, clear_line, forceWrite, linebreak, move_cursor, reset_cursor, writeColor
 from .keymap import KEYMAP
 
 
+in_colab = _is_package_available("google.colab")
```
sgugger, line 27: Should be maybe be in a try except as it seems to fail instead of returning `False`.

muellerzr (approved):
Thanks for this! Fantastic! Let's address Sylvain's note here and then we'll be good to merge 😄

src/accelerate/commands/menu/selection_menu.py
Erickrus: Yes, in some environment setup the code is not safe. When I test it in local console and colab, it works. 
I should wrap it with a try except.

sgugger (approved):
LGTM, thanks for iterating!

PR #1696, opened 2023-07-10
Fix nightly tests

muellerzr:
Solves the issue of the failing nightly by checking for the right dictionary to be parsed, and ignoring other outputs (the cause of the issue)


pacman100 (approved):
Thank you @muellerzr for fixing the tests, LGTM!

PR #1698, opened 2023-07-10
Improve quality errors

muellerzr:
When quality checks fail, provides a clear error in the job summary, as seen here: https://github.com/huggingface/accelerate/actions/runs/5509213335

> Quality check failed. Please ensure the right dependency versions are installed with 'pip install -e .[quality]' and rerun 'make style; make quality;'

Many users have issues with knowing what to do or how to install with it, so this is one solution I have in mind, otherwise go as far as a github bot that points to the different job summaries and their statuses/errors that automatically gets updated each push. As github doesn't just let you go "Details" and automatically see the summary

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -18,5 +18,5 @@ jobs:
     - name: Check if failure
       if: ${{ failure() }}
       run: |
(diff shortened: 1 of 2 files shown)


pacman100 (approved):
Thanks a bunch, @muellerzr, this is nice! 🤗

sgugger (approved):
Thanks a lot!

PR #1701, opened 2023-07-10
fix bnb version

SunMarc:
# What does this PR do ? 
This PR updates the version of bitsandbytes required for the tutorial since we fixed the [issue](https://github.com/huggingface/accelerate/pull/1679) with the serialization of 8bit model.

### DIFF
--- a/docs/source/usage_guides/quantization.md
+++ b/docs/source/usage_guides/quantization.md
@@ -28,7 +28,7 @@ You will need to install the following requirements:
 
 - Install `bitsandbytes` library
 ```bash
-pip install bitsandbytes==0.39.0
+pip install bitsandbytes
 ```
 - Install latest `accelerate` from source
 ```bash


pacman100 (commented):
Thank you @SunMarc for the fixes. However, PEFT bnb tests are failing with the latest release. Can you confirm if the same is observed with accelerate's bnb tests

SunMarc:
On accelerate's side, the tests are passing with `bitsandbytes == 0.39.1`.

SunMarc:
The failing tests are passing after the new release today =)

docs/source/usage_guides/quantization.md
```diff
@@ -28,7 +28,7 @@ You will need to install the following requirements:
 
 - Install `bitsandbytes` library
 ```bash
-pip install bitsandbytes==0.39.0
+pip install bitsandbytes
```
muellerzr, line 31: Considering we're facing issues with versions, are we marking that down in the source of accelerate for checking minimum `bnb` versions when something like this happens?

SunMarc: We do have check for minimum version of bnb when the user tries to do 4-bit quant or 8-bit quant:

muellerzr: These need to be updated for 0.40.0 for the latest breakage though, right? (Unrelated to this exact PR, just something I'm wondering about with the fixes. Otherwise PR works fine for me :) )

We also might want to consider `pip install accelerate[bnb]` at some point potentially since it is technically an extra

muellerzr: Excellent 🥳

Issue #1693, opened 2023-07-07
Mixed precision BF16 silently ignored on unsupported device


lukasHoel:
### System Info

```Shell
- `Accelerate` version: 0.20.3
- Platform: Linux-5.4.0-105.119.1.ubuntu.x86_64-x86_64-with-glibc2.31
- Python version: 3.10.11
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1 (True)
- PyTorch XPU available: False
- System RAM: 503.78 GB
- GPU type: Tesla V100-SXM2-32GB
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

A Tesla V100-SXM2-32GB GPU does not support the bf16 mode. I would expect accelerate to give a comprehensive error message in this case. However, the following scenario does not work.

1. This version has expected behaviour:

```
from accelerate import Accelerator
from accelerate.utils.imports import is_bf16_available

accelerator = Accelerator(mixed_precision="bf16")
print(accelerator.native_amp, accelerator.state.mixed_precision, accelerator.device.type, is_bf16_available())
```

Run with `accelerate launch test.py` and the output is: `ValueError: bf16 mixed precision requires PyTorch >= 1.10 and a supported device.  `

2. This version does not raise an error:
```
from accelerate import Accelerator
from accelerate.utils.imports import is_bf16_available

accelerator = Accelerator()
print(accelerator.native_amp, accelerator.state.mixed_precision, accelerator.device.type, is_bf16_available())
```

Run with `accelerate launch --mixed_precision="bf16" test.py` and the output is: `False bf16 cuda False`.

In other words: we silently ignore the requested precision instead of throwing the same error message.
In follow-up code, this may lead to undefined behaviour, e.g.:

```
model, dataloader = accelerator.prepare(model, dataloader)
for batch in dataloader:
    model(batch)
```

I got error messages like ```RuntimeError: expected scalar type BFloat16 but found Float``` during the model forward pass.

### Expected behavior

[...]

PR #1705, opened 2023-07-11
Fix launcher validation 

abhilash1910:
Possible fix for #1693  (bf16 silent error ignorance) . 
There might be  a  better design than throwing a value error, but resorted to default accelerator code for throwing the error.
@muellerzr could you take a look ?


muellerzr (approved):
Thanks! Makes sense to me :) Nice error

PR #1706, opened 2023-07-11
Skip tests when bnb isn't available

muellerzr:
Noticed some tests in Accelerator that fail when bnb isn't available, rather than use the conditional check for if it is when running the slow tests locally


SunMarc (approved):
Thanks for adding !

sgugger (approved):
Thanks!

PR #1709, opened 2023-07-12
Fix the bug where DataLoaderDispatcher gets stuck in an infinite wait…

yuxinyuan:
Fix the bug where ``DataLoaderDispatcher`` gets stuck in an infinite wait when the dataset is an ``IterDataPipe`` during multi-process training. 

In newer version of pytorch, the iterator of ``DataLoader`` will try to broadcast a shared seed across all distributed processes (see [here](https://github.com/pytorch/pytorch/blob/7c98e70d44abc7a1aead68b6ea6c8adc8c554db5/torch/utils/data/dataloader.py#L120C14-L120C14)). In the current implementation of ``DataLoaderDispatcher``, the iterator is only created in the main process. This causes the training to hang when dataset is an ``IterDataPipe``.

One can try out the script below to see the effect.

```
import accelerate
from torch.utils.data.dataloader import DataLoader
from torch.utils.data.datapipes.iter import IterableWrapper


def print_rank_by_rank(*args, **kwargs):
    for rank in range(accelerator.num_processes):
        if rank == accelerator.process_index:
            print(*args, **kwargs)


if __name__ == "__main__":
    accelerator = accelerate.Accelerator()
    accelerator.print(accelerate.__version__)

    dp = IterableWrapper(range(21)).shuffle().map(lambda x: x + 1)

    loader = DataLoader(dp, batch_size=4, shuffle=True)
    loader = accelerator.prepare(loader)

    accelerator.print(type(loader))

    print_rank_by_rank(list(loader))
    print_rank_by_rank(list(loader))
```

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -495,10 +495,6 @@ class DataLoaderDispatcher(DataLoader, DataLoaderStateMixin):
         self.state = AcceleratorState()
         self._drop_last = _drop_last
         self.skip_batches = skip_batches
-        # We can safely pass because the default is -1
-        with suppress(Exception):
-            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
-            self.remainder = length % self.total_batch_size
 
     def _fetch_batches(self, iterator):
         batches, batch = None, None
@@ -540,10 +536,14 @@ class DataLoaderDispatcher(DataLoader, DataLoaderStateMixin):
     def __iter__(self):
         self.reset()
         self.gradient_state._add_dataloader(self)
-        main_iterator = None
-        if self.state.process_index == 0:
-            # We only iterate through the DataLoader on process 0.
-            main_iterator = super().__iter__()
+        # We can safely pass because the default is -1
+        with suppress(Exception):
+            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
+            self.remainder = length % self.total_batch_size
+        # NOTE PyTorch DataLoader adds forward compatibilities for DataPipes, which broadcasts
+        # shared seed to all dist processes. Thus, we need to create iterator for all dist processes.
+        # But, we only iterate through the DataLoader on process 0.
+        main_iterator = super().__iter__()
         stop_iteration = False
         self._stop_iteration = False
         first_batch = None


src/accelerate/data_loader.py
```diff
@@ -495,10 +495,6 @@ def __init__(self, dataset, split_batches: bool = False, skip_batches=0, _drop_l
         self.state = AcceleratorState()
         self._drop_last = _drop_last
         self.skip_batches = skip_batches
-        # We can safely pass because the default is -1
-        with suppress(Exception):
-            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
-            self.remainder = length % self.total_batch_size
```
muellerzr, line 501: Is there a reason this has been moved into `__iter__` and not `__init__`?

muellerzr (approved):
This looks great, thanks! Left one comment on moving a chunk of code.

src/accelerate/data_loader.py
yuxinyuan: I didn't go through the rest of this repo to figure out how ``remainder`` is used. However, since ``self.reset()`` will always set it to -1, it just makes more sense (looking at data_loader.py) to follow ``DataLoaderShard`` and set the remainder in ``__iter__``. Otherwise, it should be safe to remove this code completely.

muellerzr: It's used for Gradient accumulation. Here it makes sense to have it in `__iter__` as there's no need to calculate it twice

sgugger: Agreed

sgugger (commented):
Thanks for your PR! Let's just keep the remainder logic in the init as mentioned by @muellerzr and we should be good to go.

src/accelerate/data_loader.py
yuxinyuan: Well, I might be missing something, but, in ``__iter__``, ``self.reset()`` will set ``self.remainder`` to -1. So, ``self.remainder`` won't be useful once we start iterating through the dataloader/dataset. Is that correct?

sgugger: Looking more, it is also updated at the end of the iter loop. @muellerzr do we actually need those lines?

But while we investigate, I agree that it's safer to just copy this after the reset.

muellerzr: Hmmm... will look into this today.

PR #1712, opened 2023-07-12
Fix bnb test by adding multi_gpu decorator

SunMarc:
# What does this PR do ? 
This PR adds a decorator to a test that should require multi gpu. Otherwise, it will fail during the nightly single_gpu test.


sgugger (approved):
Thanks for the fix!

PR #1716, opened 2023-07-12
Keep old behavior on dispatch dataloader

muellerzr:
Follow up to https://github.com/huggingface/accelerate/pull/1709, which did not include guards for when the torch version is `< 2.0.1`, which is the version that introduced this (see [here](https://github.com/pytorch/pytorch/commit/f1a6f32b72b7c2b73277f89bbf7e7459a400d80a))


sgugger (approved):
Thanks!

PR #1717, opened 2023-07-12
Remove duplicate code and check the remainder only once

muellerzr:
Second part of a followup from https://github.com/huggingface/accelerate/pull/1709, given a dataloader never really should change its remainder after iterating, the remainder isn't re-set after the end of an `__iter__`, nor is it recalculated *during* an `__iter__`. Instead we check for the default (`-1`), then recalculate.


sgugger (approved):
LGTM!

PR #1718, opened 2023-07-12
Optimize `get_scale` to reduce async calls

muellerzr:
Reduces the number of times `scaler.get_scale` is called during `step()` from 2 to 1 by storing the last scale, and overriding it after a check for `_is_overflow` is ran.


sgugger (approved):
Good idead!

PR #1719, opened 2023-07-12
Add `begin` and `end` setup/teardown to mixin, fix test

muellerzr:
There's a failing multi-GPU test currently from the prior PR because after `stop_iteration` we need to set `self.remainder` to `observed_batch_size`, which is critical. Upon thinking about this more, I came to a conclusion that this logic is better suited to live inside the mixin, as its designed specifically for bits such as this. As a result `DataLoaderStateMixin` gets two new functions: `begin()` and `end()`, which replicate the same logic done in the `DataLoader` class, but as their more internal (such as using private functions), it makes more sense to keep them in here. If we want to keep the current behavior and un-isolate, do let me know.

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -320,6 +320,18 @@ class DataLoaderStateMixin:
         self.end_of_dataloader = False
         self.remainder = -1
 
+    def begin(self):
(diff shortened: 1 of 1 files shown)


sgugger (approved):
Thanks for the fix!

Issue #1714, opened 2023-07-12
self.scaler.get_scale() forces cpu and gpu to synchronize


StarCycle:
### System Info

https://github.com/huggingface/accelerate/blob/c33adecc9f92808cbaeb0d34928e52e34180f964/src/accelerate/optimizer.py#L125C32-L125C55

In this line, "scale_before = self.scaler.get_scale()" will force a synchronization between CPU and GPU, and may make the training process slower. You can see this behaviour in pytorch profiler

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Run the following code in single gpu without mixed precision, and read the profile trace with tensorboard or chrome, you can see the cpu and gpu does not need to synchronize (Dataloader will do cpu data prefetching and DataPrefetcher will do gpu data prefetching).
2. Run the following code in single gpu with mixed precision. In the trace, you can see the cpu and gpu must synchronize before updating the network weights, because Accelerate calls "scale_before = self.scaler.get_scale()" and "scale_after = self.scaler.get_scale()" in optimizer.py.
3. Disable the 2 lines and run the following again, you can see the cpu and gpu does NOT need to synchronize and the GPU utilization becomes higher.

```python
import torch
import torchvision
import torchvision.transforms as T
from torch.profiler import profile, record_function, ProfilerActivity, tensorboard_trace_handler, schedule

def train(inputs, labels, scaler):
    with torch.autocast(device_type='cuda', dtype=torch.float16):
        outputs = model(inputs)
        loss = criterion(outputs, labels)
    optimizer.zero_grad()
    # loss.backward()
    # optimizer.step()
    scaler.scale(loss).backward()
    scaler.step(optimizer)
    scaler.update()

class DataPrefetcher():
    def __init__(self, loader, device):
        self.device = device
        self.loader = iter(loader)
        self.stream = torch.cuda.Stream()
        self.preload()

    def preload(self):
        try:
            self.next_input, self.next_label = next(self.loader)
        except StopIteration:
            self.next_input = None
            self.next_label = None
            return
        with torch.cuda.stream(self.stream):
            self.next_input = self.next_input.to(device, non_blocking=True)
            self.next_label = self.next_label.to(device, non_blocking=True)

[...]

PR #1720, opened 2023-07-13
Get rid of calling get_scale() by patching the step method of optimizer.

yuxinyuan:
Follow up of #1714 . This avoids constantly synchronization between cpu and gpu

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -60,17 +60,9 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
         self.device_placement = device_placement
         self._is_overflow = False
 
-        def patch_optimizer_step(method):
-            def patched_step(*args, **kwargs):
-                self._accelerate_num_step_called += 1
-                return method(*args, **kwargs)
-
-            return patched_step
-
         self._accelerate_num_step_called = 0
         self._optimizer_original_step_method = self.optimizer.step
-        # Bind the patched step method to self.optimizer
-        self._optimizer_patched_step_method = patch_optimizer_step(self.optimizer.step)
+        self._optimizer_patched_step_method = patch_optimizer_step(self, self.optimizer.step)
 
         # Handle device placement
         if device_placement:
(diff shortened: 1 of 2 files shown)


src/accelerate/optimizer.py
```diff
@@ -161,7 +165,26 @@ def step_was_skipped(self):
...
-        return self.__dict__.copy()
+        ret = {}
+        for k, v in self.__dict__.items():
+            if k not in [
+                "_accelerate_num_step_called",
+                "_optimizer_original_step_method",
+                "_optimizer_patched_step_method",
+            ]:
+                ret[k] = v
+        return ret
```
sgugger, line 176: A dict comprehension here would be more efficient:
```py
ret = {k: v for k, v in self.__dict__.items() if k not in ...}`
```

sgugger (approved):
Very clever! I left one comment and I think the `_accelerate_num_step_called` should become a bool called `_accelerate_step_called` since it only has two values with care about (0 or 1), but love that workaround!

src/accelerate/optimizer.py
```diff
@@ -161,7 +165,26 @@ def step_was_skipped(self):
...
 
     def __setstate__(self, state):
         self.__dict__.update(state)
+        self._accelerate_num_step_called = 0
+        self._optimizer_original_step_method = self.optimizer.step
+        self._optimizer_patched_step_method = patch_optimizer_step(self, self.optimizer.step)
+
+
+def patch_optimizer_step(accelerated_optimizer: "AcceleratedOptimizer", method):
```
muellerzr, line 185: We can use the class directly for typehints

```suggestion
def patch_optimizer_step(accelerated_optimizer: AcceleratedOptimizer, method):
```

muellerzr (approved):
This is an excellent workaround! Great work and exceedingly clever :) What Sylvain said, and I just made a small nit for typehinting, as we can use `AcceleratedOptimizer` directly here

src/accelerate/optimizer.py
yuxinyuan: I update this in the latest commit

muellerzr (approved):

sgugger:
Great! Can you just do the changes on `_accelerate_num_step_called` (see comment above) and we can merge this?

yuxinyuan:
@sgugger oops missed that comment

sgugger (approved):

muellerzr:
Looks like this is breaking when using deepspeed: https://github.com/huggingface/transformers/actions/runs/5565434638/jobs/10165822075

cc @pacman100 for any guidance?

yuxinyuan:
@muellerzr Seems to me that deepspeed is returning a non-optimizer object and, when we wrap this object with ``AcceleratedOptimizer``, an error will pop up.

I guess a simple way of fixing this is to do the following:
```
if self.scalar is not None:
    self._accelerate_step_called = False
    self._optimizer_original_step_method = self.optimizer.step
    self._optimizer_patched_step_method = patch_optimizer_step(self, self.optimizer.step)
```
Because we only need patching when we are using native amp.

Issue #1722, opened 2023-07-13
【BUG】Test failed with accelerate version  0.21.0.dev0 on single machine with 8x A100 GPU


statelesshz:
### System Info

```Shell
- `Accelerate` version: 0.21.0.dev0
- Platform: Linux-4.15.0-184-generic-x86_64-with-glibc2.31
- Python version: 3.9.12
- Numpy version: 1.21.5
- PyTorch version (GPU?): 1.11.0+cu113 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 754.43 GB
- GPU type: NVIDIA A100-PCIE-40GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Install the accelerate of the main branch
2. execute `accelerate test` with 8xA100, test failed.

The output log is as follows.
```test
Running:  accelerate-launch /home/huggingface/accelerate/src/accelerate/test_utils/scripts/test_script.py
stdout: Distributed environment: MULTI_GPU  Backend: nccl
stdout: Num processes: 8
stdout: Process index: 6
stdout: Local process index: 6
stdout: Device: cuda:6
stdout:
stdout: Mixed precision type: no
stdout:
stdout: Distributed environment: MULTI_GPU  Backend: nccl
stdout: Num processes: 8
stdout: Process index: 1
stdout: Local process index: 1
stdout: Device: cuda:1
stdout:
stdout: Mixed precision type: no
stdout:
stdout: Distributed environment: MULTI_GPU  Backend: nccl
stdout: Num processes: 8
stdout: Process index: 7
stdout: Local process index: 7
stdout: Device: cuda:7
stdout:
stdout: Mixed precision type: no
stdout:
[...]

PR #1724, opened 2023-07-13
fix failing test on 8GPU

statelesshz:
Solves #1722 by changing the test to：
- Properly give us an extra remainder/odd one

Verified on (2, 4, 8)xA100


sgugger:
cc @muellerzr

muellerzr (approved):
Thanks for adding support for up to 8 gpus!

PR #1725, opened 2023-07-13
add compatibility with peft

SunMarc:
# What does this PR do  ?
This PR fixes an issue when people quantize transformers model with the quantization api from accelerate and convert them into peft model for training. In order to fix this issue, we need to add the attributes  `model.is_loaded_in_4bit` and `model.is_loaded_in_8bit`. However, as indicated by the docs, the user should use the transformers integration of bnb when using transformers model. I've also updated the training section of the doc. 

related issue #1713


sgugger (approved):
Thanks for adding this!

PR #1726, opened 2023-07-13
Better control over DDP's `no_sync`

NouamaneTazi:
This PR is a suggestion for a solution for the following observation regarding torch's DDP behaviour using `no_sync`:

Here’s the example from the docs (uses forward-backward)
```python
ddp = torch.nn.parallel.DistributedDataParallel(model, pg)
with ddp.no_sync():
    ddp(input).backward()  # no synchronization, accumulate grads
ddp(another_input).backward()  # synchronize grads
```

Let’s separate forward from backward. It seems like `no_sync` doesn't affect backward(), it must wrap the forward() call instead.
```python
ddp = torch.nn.parallel.DistributedDataParallel(model, pg)
with ddp.no_sync(): # forward under no_sync -> works
    loss = ddp(input) 
loss.backward() # no synchronization, accumulate grads
ddp(another_input).backward()  # synchronize grads
```

```python
ddp = torch.nn.parallel.DistributedDataParallel(model, pg)
loss = ddp(input) 
with ddp.no_sync(): # backward under no_sync -> doesn't work
    loss.backward() # synchronize grads
ddp(another_input).backward()  # synchronize grads
# This means that no_sync doesn't affect backward() calls, only forward() calls
```

This PR aims at solving the following: How to do multiple forwards then multiple backwards, and only sync on the last backward
```python
ddp = torch.nn.parallel.DistributedDataParallel(model, pg)
with ddp.no_sync():
    loss_1 = ddp(input) 
loss_2 = ddp(another_input)
loss_1.backward() # synchronizes grad, but I expect it to no sync (how to do this?)
loss_2.backward()  # synchronize grads
```

My hacky solution was to have a trigger mechanism, that triggers a sync on the last backward:
```python
def ddp_trigger_sync_in_bwd(model_ddp):
    """Trigger the sync of the gradients in the next backward pass of DDP model."""
    assert isinstance(model_ddp, torch.nn.parallel.DistributedDataParallel)
    model_ddp.require_backward_grad_sync = True
    model_ddp.require_forward_param_sync = True
    model_ddp.reducer.prepare_for_backward([])
```

```python
ddp = torch.nn.parallel.DistributedDataParallel(model, pg)
with ddp.no_sync():
    loss_1 = ddp(input) 
    loss_2 = ddp(another_input)
loss_1.backward() # doesn't sync
ddp_trigger_sync_in_bwd(ddp)
loss_2.backward()  # synchronize grads
```

Also as this gives better control over which model to sync grads for, and exactly on which bwd pass. I think it should unlock much more usecases for grad accumulation than the example I gave 🚀
cc @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -866,31 +866,31 @@ class Accelerator:
     @contextmanager
     def trigger_sync_in_backward(model):
         """Trigger the sync of the gradients in the next backward pass of the model after multiple forward passes under
-`Accelerator.no_sync` (only applicable in multi-GPU scenarios).
+        `Accelerator.no_sync` (only applicable in multi-GPU scenarios).
 
-        If the script is not launched in distributed mode, this context manager does nothing.
+                If the script is not launched in distributed mode, this context manager does nothing.
 
-        Args:
-            model (`torch.nn.Module`):
-                The model for which to trigger the gradient synchronization.
+                Args:
+                    model (`torch.nn.Module`):
+                        The model for which to trigger the gradient synchronization.
 
-        Example:
+                Example:
 
-        ```python
-        >>> from accelerate import Accelerator
+                ```python
+                >>> from accelerate import Accelerator
 
-        >>> accelerator = Accelerator()
-        >>> dataloader, model, optimizer = accelerator.prepare(dataloader, model, optimizer)
+                >>> accelerator = Accelerator()
+                >>> dataloader, model, optimizer = accelerator.prepare(dataloader, model, optimizer)
 
-        >>> with accelerator.no_sync():
-        ...     loss_a = loss_func(model(input_a))  # first forward pass
-        ...     loss_b = loss_func(model(input_b))  # second forward pass
-        >>> accelerator.backward(loss_a)  # No synchronization across processes, only accumulate gradients
-        >>> with accelerator.trigger_sync_in_backward(model):
-        ...     accelerator.backward(loss_b)  # Synchronization across all processes
-        >>> optimizer.step()
-        >>> optimizer.zero_grad()
-        ```
+                >>> with accelerator.no_sync():
+                ...     loss_a = loss_func(model(input_a))  # first forward pass
+                ...     loss_b = loss_func(model(input_b))  # second forward pass
+                >>> accelerator.backward(loss_a)  # No synchronization across processes, only accumulate gradients
+                >>> with accelerator.trigger_sync_in_backward(model):
+                ...     accelerator.backward(loss_b)  # Synchronization across all processes
+                >>> optimizer.step()
+                >>> optimizer.zero_grad()
+                ```
         """
         if not isinstance(model, torch.nn.parallel.DistributedDataParallel):
             yield
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -868,10 +868,10 @@ class Accelerator:
         """Trigger the sync of the gradients in the next backward pass of DDP model.
 
         If `model` is not in DDP, this context manager does nothing
-        
+
         Args:
             model_ddp (`torch.nn.parallel.DistributedDataParallel`):
-                PyTorch Module wrapped in `DistributedDataParallel` 
+                PyTorch Module wrapped in `DistributedDataParallel`
 
         Example:
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -865,7 +865,26 @@ class Accelerator:
     @staticmethod
     @contextmanager
     def ddp_trigger_sync_in_bwd(model_ddp):
-        """Trigger the sync of the gradients in the next backward pass of DDP model."""
+        """Trigger the sync of the gradients in the next backward pass of DDP model.
+
+        Example:
+
+        ```python
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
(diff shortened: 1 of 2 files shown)


NouamaneTazi:
Updated the PR description

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
...
+        >>> accelerator.backward(loss_a) # No synchronization across processes, only accumulate gradients
+        >>> with accelerator.ddp_trigger_sync_in_bwd(model):
+        ...     accelerator.backward(loss_b) # Synchronization across all processes
+        >>> optimizer.step()
+        >>> optimizer.zero_grad()
+        ```
+        """
+        if not isinstance(model_ddp, torch.nn.parallel.DistributedDataParallel):
+            yield
+            return
```
sgugger, line 896: Not sure you can put a `return` statement since there is a `yield`. You might need to do an else branch.

NouamaneTazi: I just tested it and it works.
But if you prefer the `if.. else` I can switch to that

sgugger: As long as Python is happy, I'm happy.

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
         with context():
             yield
 
+    @staticmethod
+    @contextmanager
+    def ddp_trigger_sync_in_bwd(model_ddp):
```
sgugger, line 867:
```suggestion
    def trigger_sync_in_backward(model):
```
I don't think we need to put the DDP in the name.

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
...
+        """Trigger the sync of the gradients in the next backward pass of DDP model.
+
+        If `model` is not in DDP, this context manager does nothing
+
+        Args:
+            model_ddp (`torch.nn.parallel.DistributedDataParallel`):
```
sgugger, line 873:
```suggestion
            model (`torch.nn.Module`):
```

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
...
+        >>> from accelerate import Accelerator
+
+        >>> accelerator = Accelerator()
+        >>> dataloader, model, optimizer = accelerator.prepare(dataloader, model, optimizer)
+
+        >>> with accelerator.no_sync():
+        ...     loss_a = loss_func(model(input_a)) # first forward pass
+        ...     loss_b = loss_func(model(input_b)) # second forward pass
+        >>> accelerator.backward(loss_a) # No synchronization across processes, only accumulate gradients
+        >>> with accelerator.ddp_trigger_sync_in_bwd(model):
```
sgugger, line 888:
```suggestion
        >>> with accelerator.trigger_sync_in_backward(model):
```

sgugger (commented):
Thanks! Added a couple more comments on the style.

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
...
+        ...     accelerator.backward(loss_b) # Synchronization across all processes
+        >>> optimizer.step()
+        >>> optimizer.zero_grad()
+        ```
+        """
+        if not isinstance(model_ddp, torch.nn.parallel.DistributedDataParallel):
```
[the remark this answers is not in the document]
NouamaneTazi: fixed!

sgugger (approved):
Thanks for iterating! @muellerzr I'll let you have a second look.

muellerzr (approved):
This is excellent! Thanks for doing this 😄 

As a next step, we should update the gradient synchronization docs to discuss the why and how of this technique: https://github.com/huggingface/accelerate/blob/main/docs/source/concept_guides/gradient_synchronization.md

(Can be done as a follow-up)

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
         with context():
             yield
 
+    @staticmethod
+    @contextmanager
+    def trigger_sync_in_backward(model):
+        """Trigger the sync of the gradients in the next backward pass of the model (only applicable in distributed mode).
```
muellerzr, line 868:
```suggestion
        """Trigger the sync of the gradients in the next backward pass of the model after multiple forward passes under `Accelerator.no_sync` (only applicable in multi-GPU scenarios).
```

src/accelerate/accelerator.py
```diff
@@ -862,6 +862,52 @@ def no_sync(self, model):
...
+        if not isinstance(model, torch.nn.parallel.DistributedDataParallel):
+            yield
+            return
+
+        old_require_backward_grad_sync = model.require_backward_grad_sync
+        old_require_forward_param_sync = model.require_forward_param_sync
+
+        model.require_backward_grad_sync = True
+        model.require_forward_param_sync = True
+        # https://github.com/pytorch/pytorch/blob/e1502c0cdbfd17548c612f25d5a65b1e4b86224d/torch/csrc/distributed/c10d/reducer.cpp#L1371-L1402
```
muellerzr, line 903: Can we add a bit more context to this note for folks who will want to look here and read more towards what we're doing? (Just a sentence or two)

muellerzr (approved):
Thanks again, also left a few nits looking closer

src/accelerate/accelerator.py
NouamaneTazi: Added comment!

PR #1731, opened 2023-07-14
Adding a shape check for `set_module_tensor_to_device`.

### LINKED ISSUE #1729: `set_module_tensor_to_device` has wrong shape
(created by @Narsil who was stuck outside github)
```python 
from accelerate.utils import set_module_tensor_to_device
import torch

model = AutoModel.from_pretrained("gpt2")

print(model.wte.weight.shape)
A = torch.zeros((2, 2))
set_module_tensor_to_device(model, 'wte.weight', "cpu", value=A)
print(model.wte.weight.shape)
```
Potential bug:
`set_module_tensor_to_device` doesn't check shapes, meaning you can potentially load invalid checkpoints onto an already loaded model.
This is triggered here in transformers : https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L717
Is it normal ? Is the bug here, or in transformers where we should do the shape checking ?
In anycase, if you have an invalid checkpoint vs config, there will be no error. I'm not sure this is OK.

Narsil:
Fixes #1729

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -268,7 +268,7 @@ def set_module_tensor_to_device(
 
     if value is not None:
         if old_value.shape != value.shape:
-            raise ValueError(f"Trying to set a tensor of shape {value.shape} in {tensor_name} (which has shape {old_value.shape}), this look incorrect.")
+            raise ValueError(f"Trying to set a tensor of shape {value.shape} in \"{tensor_name}\" (which has shape {old_value.shape}), this look incorrect.")
 
         if dtype is None:
             # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -267,6 +267,9 @@ def set_module_tensor_to_device(
         raise ValueError(f"{tensor_name} is on the meta device, we need a `value` to put in on {device}.")
 
     if value is not None:
+        if old_value.shape != value.shape:
+            raise ValueError(f"{tensor_name} has shape {old_value.shape}, new value has shape {value.shape}, this look incorrect.")
```
sgugger, line 271:
```suggestion
            raise ValueError(f"Trying to set a tensor of shape {value.shape} in {tensor_name} (which has shape {old_value.shape}), this look incorrect.")
```

sgugger (approved):
Thanks a lot! Just a suggestion on the error message since it's unlikely the error will come when the user calls `set_module_tensor_to_device` manually, but more from a higher-level API.

PR #1733, opened 2023-07-17
Fix errors when optimizer is not a Pytorch optimizer.

yuxinyuan:
https://github.com/huggingface/accelerate/pull/1720#issuecomment-1637219672


muellerzr (approved):
Looks good to me, but also would like @pacman100 to take a peek and double check this is right for when we deal with deepspeed :)

pacman100 (approved):
Hello @yuxinyuan for fixing the issue, LGTM!

sgugger (approved):
Thanks for the fix!

PR #1736, opened 2023-07-17
Update docs

muellerzr:
Updates docs to add references to user-facing APIs that individuals could possibly use


sgugger (approved):
Thanks!

PR #1738, opened 2023-07-18
fix KwargsHandler.to_kwargs not working with os.environ initialization in __post_init__

CyCle1024:
[https://github.com/huggingface/accelerate/blob/653ba110d31c86d3527bb88bf6209441c176ce11/src/accelerate/accelerator.py#L1423-L1427](https://github.com/huggingface/accelerate/blob/653ba110d31c86d3527bb88bf6209441c176ce11/src/accelerate/accelerator.py#L1423-L1427)
Before this patch, `self.state.dynamo_plugin.to_kwargs()` in above code always results in `{}` if torch.compile kwargs are set through environment variables. This behaviour is used by transformers:
[https://github.com/huggingface/transformers/blob/66fd3a8d626a32989f4569260db32785c6cbf42a/src/transformers/training_args.py#L1375-L1382](https://github.com/huggingface/transformers/blob/66fd3a8d626a32989f4569260db32785c6cbf42a/src/transformers/training_args.py#L1375-L1382)

In this patch, we fix it by unset all data in os.environ when default_dict is evaluated.

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -48,6 +48,7 @@ class KwargsHandler:
         """
         # import clear_environment here to avoid circular import problem
         from .other import clear_environment
+
         with clear_environment():
             default_dict = self.__class__().to_dict()
         this_dict = self.to_dict()
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -34,6 +34,14 @@ import torch
 from .constants import FSDP_AUTO_WRAP_POLICY, FSDP_BACKWARD_PREFETCH, FSDP_STATE_DICT_TYPE
 
 
+@contextmanager
+def clear_os_environ():
+    _old_os_environ = os.environ
+    os.environ = dict()
+    yield
+    os.environ = _old_os_environ
+
+
 class KwargsHandler:
     """
     Internal mixin that implements a `to_kwargs()` method for a dataclass.
@@ -46,7 +54,8 @@ class KwargsHandler:
         """
         Returns a dictionary containing the attributes with values different from the default of this class.
         """
-        default_dict = self.__class__().to_dict()
+        with clear_os_environ():
+            default_dict = self.__class__().to_dict()
         this_dict = self.to_dict()
         return {k: v for k, v in this_dict.items() if default_dict[k] != v}
 
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -34,14 +34,6 @@ import torch
 from .constants import FSDP_AUTO_WRAP_POLICY, FSDP_BACKWARD_PREFETCH, FSDP_STATE_DICT_TYPE
 
 
-@contextmanager
-def clear_os_environ():
-    _old_os_environ = os.environ
-    os.environ = dict()
-    yield
-    os.environ = _old_os_environ
-
-
 class KwargsHandler:
     """
     Internal mixin that implements a `to_kwargs()` method for a dataclass.
@@ -54,7 +46,9 @@ class KwargsHandler:
         """
         Returns a dictionary containing the attributes with values different from the default of this class.
         """
-        with clear_os_environ():
+        # import clear_environment here to avoid circular import problem
+        from .other import clear_environment
+        with clear_environment():
             default_dict = self.__class__().to_dict()
         this_dict = self.to_dict()
         return {k: v for k, v in this_dict.items() if default_dict[k] != v}
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -129,7 +129,7 @@ def clear_environment():
 
     >>> os.environ["FOO"] = "bar"
     >>> with clear_environment():
-    ...     print(os.environ) # prints "{}"
+    ...     print(os.environ)  # prints "{}"
     ...     os.environ["FOO"] = "new_bar"
     ...     print(os.environ.get("FOO", None))  # prints "new_bar"
     >>> print(os.environ["FOO"])  # prints "bar"
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -115,6 +115,35 @@ def save(obj, f):
 
 
 @contextmanager
+def clear_environment():
+    """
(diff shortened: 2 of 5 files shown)


sgugger (approved):
LGTM, thanks! Could you have a second look @muellerzr ?

CyCle1024:
> LGTM, thanks! Could you have a second look @muellerzr ?

Thanks for your review, I found a side effect of the test in this commit. It causes ci failure. I'm working on it.

CyCle1024:
@sgugger @muellerzr I pushed a commit to fix ci, anyone plz approves the workflows, thanks!

src/accelerate/utils/dataclasses.py
```diff
@@ -34,6 +34,14 @@
 from .constants import FSDP_AUTO_WRAP_POLICY, FSDP_BACKWARD_PREFETCH, FSDP_STATE_DICT_TYPE
 
 
+@contextmanager
+def clear_os_environ():
+    _old_os_environ = os.environ
+    os.environ = dict()
+    yield
+    os.environ = _old_os_environ
```
muellerzr, line 42: Let's move this to utils/other so it's by `patch_environment`: https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/other.py#L117

muellerzr (approved):
Looks good to me, bar moving a util somewhere else

src/accelerate/utils/dataclasses.py
CyCle1024: @muellerzr I pushed a new commit:
[move clear_os_environ func to utils/other and rename it](https://github.com/huggingface/accelerate/pull/1738/commits/80c282ffa17e8d759e37862d885b4b7e4761d96f)

In this commit, I find something hard to handle nicely (circular import):
1. I want to use a func `clear_environment` defined in utils.other in utils.dataclasses, utils.dataclasses import utils.other
2. utils.other requires import commands.config.default, thus requires import commands.config
[https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/utils/other.py#L21](https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/utils/other.py#L21)
3. commands.config requires import commands.config.config
[https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/commands/config/__init__.py#L19](https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/commands/config/__init__.py#L19)
4. commands.config.config requires from accelerate.utils import ComputeEnvironment
[https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/commands/config/config.py#L20](https://github.com/huggingface/accelerate/blob/daa1952f4703e1f22cb3788309a5efd88af6f208/src/accelerate/commands/config/config.py#L20)
5. this causes circular import: we are in the process of import utils.dataclasses, but ComputeEnvironment is [visiable](https://github.com/huggingface/accelerate/blob/69e4c3c54da3201eda288b500d138761e7a5221c/src/accelerate/utils/__init__.py#L3) after utils.dataclasses is imported completely.

sgugger (approved):
Thanks for iterating! @muellerzr I'll let you merge if you are happy with the changes.

muellerzr (approved):
Thanks! Good workaround for the circular import. Can you run `make style; make quality` one more time and we can merge this?

CyCle1024:
> Thanks! Good workaround for the circular import. Can you run `make style; make quality` one more time and we can merge this?

The latest commit is tested by running `pytest tests/test_kwargs_handlers.py; make style; make quality` locally, you can review it if you have time. Sorry for not going through CONTRIBUTING.md first.

CyCle1024:
I don't know whether I should add `[[autodoc]] utils.clear_environment` here: [https://github.com/huggingface/accelerate/blob/efc1fa83769ee24367837960afc27df2d7799515/docs/source/package_reference/utilities.md?plain=1#L103-L110](https://github.com/huggingface/accelerate/blob/efc1fa83769ee24367837960afc27df2d7799515/docs/source/package_reference/utilities.md?plain=1#L103-L110)

muellerzr:
@CyCle1024 yes you can put it there under or above `patch_environment` if you would like! 😄

src/accelerate/utils/other.py
```diff
@@ -114,6 +114,35 @@ def save(obj, f):
...
+    >>> import os
+    >>> from accelerate.utils import clear_environment
+
+    >>> os.environ["FOO"] = "bar"
+    >>> with clear_environment():
+    ...     print(os.environ)  # prints "{}"
+    ...     os.environ["FOO"] = "new_bar"
+    ...     print(os.environ.get("FOO", None))  # prints "new_bar"
+    >>> print(os.environ["FOO"])  # prints "bar"
+    ```
```
muellerzr, line 136:
```suggestion
    >>> os.environ["FOO"] = "bar"
    >>> with clear_environment():
    ...     print(os.environ)
    "bar"
    ...     os.environ["FOO"] = "new_bar"
    ...     print(os.environ.get("FOO", None))
    "new_bar"
    >>> print(os.environ["FOO"])
    "bar"
    ```
```

muellerzr (commented):
Some nits in terms of docstring format with print

src/accelerate/utils/other.py
CyCle1024: Thanks for your suggestion, I tried what your suggest, but unfortunately it would fail the doc style check.
I tried another way in my latest commit, you can have a look: [https://github.com/huggingface/accelerate/blob/a9c7cbb35aa0d0a7c736b1c41d1ef3712e43d792/src/accelerate/utils/other.py#L126-L141](https://github.com/huggingface/accelerate/blob/a9c7cbb35aa0d0a7c736b1c41d1ef3712e43d792/src/accelerate/utils/other.py#L126-L141)
Is it appropriate?

muellerzr: Yep I'd say that is appropriate enough :) Good work!

akx:
[`os.environ = ...` doesn't actually do anything to the process's environment variables, though](https://docs.astral.sh/ruff/rules/assignment-to-os-environ/) (this is being followed up on in #2450).

Issue #1737, opened 2023-07-18
Expose the ``cache_enabled`` parameter option to users in accelerate


yuxinyuan:
As is shown [here](https://github.com/huggingface/accelerate/blob/653ba110d31c86d3527bb88bf6209441c176ce11/src/accelerate/accelerator.py#L1350C54-L1350C54), right now, when users want to use ``native_amp``, accelerate will initialize ``torch.autocast`` with default parameters and wrap the forward method of user's models. However, this prevents users from customizing the ``enable_cache`` parameter inside autocast. Considering that disabling weight cache can save more GPU memories, it would be beneficial to expose the "enable_cache" option to users.

sgugger:
Sounds like something we could do with a KwargHandler like for DDP, cc @muellerzr

muellerzr:
Note that you can do this if you use `Accelerator.autocast(cache_enabled=True)`, however I'll add a kwarg handler for customizing autocast

PR #1740, opened 2023-07-18
Expose autocast kwargs and simplify autocast wrapper

muellerzr:
This PR introduces `AutocastKwargs`, which help specify a few parameters to `torch.autocast` for the user to use. With this, we also change the internal logic in `Accelerator` to use `get_mixed_precision_context_manager` when grabbing the context manager, rather than duplicating the logic so these kwargs can be passed down. 

`enabled` is there so that users can do something such as:

```python
        kwargs = AutocastKwargs(enabled=False)
        AcceleratorState._reset_state()
        accelerator = Accelerator(mixed_precision="fp16")

        a_float32 = torch.rand((8, 8), device=accelerator.device)
        b_float32 = torch.rand((8, 8), device=accelerator.device)
        c_float32 = torch.rand((8, 8), device=accelerator.device)
        d_float32 = torch.rand((8, 8), device=accelerator.device)

        with accelerator.autocast():
            e_float16 = torch.mm(a_float32, b_float32)
            assert e_float16.dtype == torch.float16

            with accelerator.autocast(autocast_handler=kwargs):
                # Convert e_float16 to float32
                f_float32 = torch.mm(c_float32, e_float16.float())
                assert f_float32.dtype == torch.float32

            g_float16 = torch.mm(d_float32, f_float32)
            # We should be back in fp16
            assert g_float16.dtype == torch.float16
```

Solves https://github.com/huggingface/accelerate/issues/1737


sgugger (approved):
LGTM, thanks!

src/accelerate/utils/dataclasses.py
```diff
@@ -51,6 +51,28 @@ def to_kwargs(self):
...
+    from accelerate import Accelerator
+    from accelerate.utils import AutocastKwargs
+
+    kwargs = AutocastKwargs(cache_enabled=True)
+    accelerator = Accelerator(kwargs_handlers=[kwargs])
+    ```
+    """
+
+    enabled: bool = False
+    cache_enabled: bool = False
```
yuxinyuan, line 73: Shouldn't the default value of ``cache_enabled`` be True, as that's the default value of pytorch

muellerzr: The default value is `None`, but yes. I will change this. Nice catch.

sgugger (approved):
Thanks!

Issue #1601, opened 2023-06-16
If save_state support automatic_checkpoint_naming, then load_state should be able to load latest checkpoint


usamec:
`save_state` support automatic checkpoint naming (as `checkpoint_0`, `checkpoint_1`) and also deleting old checkpoints.
However, when one wants to load_state, he needs to first parse which checkpoint is the latest and also manually set current iteration in project config.

This is quite inconsistent in my opinion.

sgugger:
cc @muellerzr

mshvartsman:
> needs to first parse which checkpoint is the latest and also manually set current iteration in project config.

I think the PR fixed the first issue (checkpoints names are parsed automatically) but not the second one (current iteration is not set). So if I save a checkpoint, load it, and try to save again, I get `Checkpoint directory {output_dir} ({self.save_iteration}) already exists...` error. Is that correct? If so, I wonder if we should reopen the issue, and I can PR in a fix (or if I'm misunderstanding something, please let me know how to use this functionality correctly).

PR #1741, opened 2023-07-18
Let load_state automatically grab the latest save

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1601 by having `Accelerator.load_state` grab the latest save folder and use it if no filename is passed.


sgugger (approved):
Nice improvement, thanks!

PR #1742, opened 2023-07-18
Fixed the bug that split dict incorrectly

yuangpeng:


muellerzr (approved):
Thanks! Could you add a test for what you fixed? It's in `test_utils/test_script.py`

sgugger (approved):
Thanks!

muellerzr (approved):
Thanks for adding the test!

PR #1745, opened 2023-07-19
Fix FSDP related issues

pacman100:
### What does this PR do?
1. Fixes `load_result` never being passed when using FSDP in Trainer
2. Fixes https://github.com/huggingface/transformers/issues/24874
3. In torch nightlies they have removed the support for `ignored_parameters`, updating the same.


sgugger (approved):
Thanks for the fixes!

Issue #1743, opened 2023-07-18
what is the possible reason for accelerate running on cuda 12.2 8xA100 with error accelerate multiprocessing.api:failed (exitcode: -9)


garychan22:
### System Info

```Shell
ubuntu 22.04
gpu A100 80G
cuda version 12.2
accelerate version 0.21.0
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

running the demo script from diffusers [train_text_to_image.py](https://github.com/huggingface/diffusers/tree/main/examples/text_to_image) for 100k iterations with batch size 8 each gpu, 8 A100 gpus in total

### Expected behavior

successful training without any problem

muellerzr:
We need a full stack trace to know more

muellerzr:
And as asked, the output of `accelerate env`

garychan22:
@muellerzr thanks for your reply!
the full stack trace is
```
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 43837 closing signal SIGTERM
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 43839 closing signal SIGTERM
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 43841 closing signal SIGTERM
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 43843 closing signal SIGTERM
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 43844 closing signal SIGTERM
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: -9) local_rank: 1 (pid: 43838) of binary: /home/gary/miniconda3/envs/blip2/bin/python
Traceback (most recent call last):
  File "/home/gary/miniconda3/envs/blip2/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/home/gary/miniconda3/envs/blip2/lib/python3.8/site-packages/accelerate/commands/accelerate_cli.py", line 45, in main
    args.func(args)
  File "/home/gary/miniconda3/envs/blip2/lib/python3.8/site-packages/accelerate/commands/launch.py", line 970, in launch_command
    multi_gpu_launcher(args)
[...]

PR #1746, opened 2023-07-19
Check for misconfiguration of single node & single GPU

muellerzr:
It's possible to end up in a situation following the prompts where `num_machines == 1`, `num_processes == 1`, and `distributed_type == DistributedType.MULTI_GPU` if you select "multi-GPU" to the first question. This PR adds a check and raises the right error if done. Helps https://github.com/huggingface/accelerate/issues/1743

(Don't review until tomorrow 😉 )


sgugger (approved):
Thanks for this check!

PR #1748, opened 2023-07-19
Rework new constant for operations

muellerzr:
While working on the operation debugging framework, I noticed that our operation utils have a *lot* of if/else checks for the same thing, along with the fact many are using `in` instead of `==` when checking a `DistributedType`. This PR refactors the logic to check for a new `TORCH_DISTRIBUTED_OPERATION_TYPES`, which should be used when the same `torch.distributed.{something}` should be called in reference to the compute environment we have. Since it's not just CUDA related, hence the naming.


muellerzr:
I also noticed that we did not keep the `_cpu_{operation}` naming for broadcasting. Should we keep this I assume for readability? (So it's `_cpu_broadcast` and `_gpu_broadcast`)? Or is it okay to just get rid of those references and just keep `_gpu_broadcast`

sgugger (approved):
Thanks for the refactor!

PR #1749, opened 2023-07-19
Remove unused constant

muellerzr:
While going through the code, I noticed `XPU_DISTRIBUTED_TYPES` was unused in the framework. This PR cleans that up, unless @abhilash1910 do you have a use for this?


sgugger (approved):
Thanks!

PR #1753, opened 2023-07-21
FSDP enhancements and fixes

pacman100:
### What does this PR do?

1. If the model is already an instance of FSDP, bypass the warning and additional FSDP preparation logic. This is to enable PR https://github.com/huggingface/transformers/pull/24980 which Fixes https://github.com/huggingface/transformers/issues/24724
2. allow usage of _no_split_modules to simplify UX when using FSDP. Fixes https://github.com/huggingface/transformers/issues/24568 via https://github.com/huggingface/transformers/pull/24980
3. `extract_from_parallel` how unwraps FSDP models too but these will lead to FSDP issues if there are modules as part of the global FSDP unit as the parameters of those modules won't be gathered if one isn't calling global FSDP wrapper's forward call. This is wrt issue https://github.com/huggingface/diffusers/issues/4037

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1196,11 +1196,15 @@ class Accelerator:
             )
 
         if self.distributed_type == DistributedType.FSDP:
+            from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
+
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Makes sense to me, thanks!

PR #1756, opened 2023-07-21
Introduce an experimental distributed operations framework

muellerzr:
# Distributed Operations debug framework

## What does this add?

This PR introduces a wrapper framework around each distributed op in Accelerate that lets us verify if the operation will raise a `timeout` error or not by using `torch.distributed.gather_object` to gather all the tensors involved (which will not lead to a timeout if they are different shapes) and run a check that they are compatible shapes

## Why is it needed?

A very frustrating part about distributed training debugging is knowing when incompatible shapes lead to these timeout issues, which almost never actually have a decent traceback showing what the heck happened and why

### User-facing:

A new argument can be added to the `config.yaml`, during `accelerate config`, and `accelerate launch`* called `debug` which will enable this feature.

*: `launch` already has a `debug` option which enables this. 

### Internal structure:

Two decorators were introduced:

- `verify_operation`: This will gather the shapes of tensors across devices, and check if they have the same shape. If not, it will raise a `DistributedOperationException` explaining what went wrong
- `chained_operation`: This is useful when you have operations that call an internal operation with this checker (such as `reduce` calls `gather`) to provide a clearer traceback on the original problematic operation

And a custom Exception as it helps describe more clearly what went wrong:
- `DistributedOperationException`

## Basic Usage Example(s):

When introducing a new operation that calls a `torch.distributed` operation under the hood:

```python
@verify_operation
def my_operation
```

As a user to enable it:

Enable it through the questionarre during `accelerate config` (recommended)

From the CLI: 

```
accelerate launch --debug {my_script.py} --arg1 --arg2
```

As an environmental variable (which avoids the need for `accelerate launch`):

```
ACCELERATE_DEBUG_MODE="1" python {my_script.py} --arg1 --arg2
```

Manually changing the `config.yaml` file:

```diff
 compute_environment: LOCAL_MACHINE
+debug: true
```

You want to enable this if you need to be extra careful about timeouts, or if you experience timeouts.

**This is not enabled by default, as it will add extra calls to `torch.distributed.gather_object`, which can lead to more communications and thus slower speeds**


sgugger (approved):
LGTm, thanks a lot for diving into this!

pacman100 (approved):
Thank you @muellerzr, very helpful, LGTM!

PR #1759, opened 2023-07-23
Fix error when max_memory argument is in unexpected order

ranchlai:
# What does this PR do?
This PR fixes some errors when user-provided `max_memory` argument in `infer_auto_device_map` is not in expected order or contains unexpected devices. 
- [Add argument-checking for max_memory](https://github.com/huggingface/accelerate/commit/301c034ac935509b2b6c706fa47de997c0aab20a)
- [Sort the user-provided max_memory keys in gpu-cpu-disk order](https://github.com/huggingface/accelerate/commit/f84ac36bcda4525f6bbc1596770453e34923ee29)
- [Fixed the bug that adds "disk" to gpus if it is provided in max_memory argument](https://github.com/huggingface/accelerate/commit/394d937fc344bae9e7a370cd4b200948a22736cd)

For some large models that I tested,  using device_map="auto" might still lead to OOM problem( and I'd like to look into it soon). So we need to explicitly call `infer_auto_device_map`  and specify `max_memory` argument to get a device_map. This PR makes the function more user-friendly and less error-prune.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -660,7 +660,7 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
     for k in max_memory.keys():
         if k not in all_devices:
             raise ValueError(
-                f"Device {k} is not recognized, avaiable device are integers(for GPU/XPU), 'mps', 'cpu' and 'disk'"
+                f"Device {k} is not recognized, available device are integers(for GPU/XPU), 'mps', 'cpu' and 'disk'"
             )
     max_memory = {k: max_memory[k] for k in all_devices}
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -647,20 +647,18 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
     # As gpu/xpu are represented by int, we need to sort them first.
     gpu_devices = [k for k in max_memory.keys() if isinstance(k, int)]
     gpu_devices.sort()
-    # check if gpu/xgpu device are available and if not, throw a warning
+    # check if gpu/xgpu devices are available and if not, throw a warning
     num_devices = torch.xpu.device_count() if is_xpu_available() else torch.cuda.device_count()
     for device in gpu_devices:
         if device >= num_devices or device < 0:
-            logger.warning(
-                f"Device {device} is not available, available device are {list(range(num_devices))}"
-            )
+            logger.warning(f"Device {device} is not available, available devices are {list(range(num_devices))}")
     # Add the other devices in the preset order if they are available
     all_devices = gpu_devices + [k for k in ["mps", "cpu", "disk"] if k in max_memory.keys()]
     # Raise an error if a device is not recognized
     for k in max_memory.keys():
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -642,6 +642,28 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
     for key in max_memory:
         if isinstance(max_memory[key], str):
             max_memory[key] = convert_file_size_to_int(max_memory[key])
+
+    # Need to sort the device by type to make sure that we allocate the gpu first.
+    # As gpu/xpu are represented by int, we need to sort them first.
+    gpu_devices = [k for k in max_memory.keys() if isinstance(k, int)]
+    gpu_devices.sort()
+    # check if gpu/xgpu device are available and if not, throw a warning
+    num_deivces = torch.xpu.device_count() if is_xpu_available() else torch.cuda.device_count()
```
sgugger, line 651:
```suggestion
    num_devices = torch.xpu.device_count() if is_xpu_available() else torch.cuda.device_count()
```

src/accelerate/utils/modeling.py
```diff
@@ -642,6 +642,28 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+    for device in gpu_devices:
+        if device >= num_deivces or device < 0:
```
sgugger, line 653:
```suggestion
        if device >= num_devices or device < 0:
```

src/accelerate/utils/modeling.py
```diff
@@ -642,6 +642,28 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+            logger.warning(
+                f"Device {max(gpu_devices)} is not available, available device are {list(range(num_deivces))}"
+            )
```
sgugger, line 656:
```suggestion
            logger.warning(
                f"Device {device} is not available, available device are {list(range(num_devices))}"
            )
```
This will print several times the same warning instead of printing one per wrong device.

src/accelerate/utils/modeling.py
```diff
@@ -642,6 +642,28 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+    # Add the other devices in the preset order if they are available
+    all_devices = gpu_devices + [k for k in ["mps", "cpu", "disk"] if k in max_memory.keys()]
+    # Raise an error if a device is not recognized
+    for k in max_memory.keys():
+        if k not in all_devices:
+            raise ValueError(
+                f"Device {k} is not recognized, avaiable device are integers(for GPU/XPU), 'mps', 'cpu' and 'disk'"
```
sgugger, line 663:
```suggestion
                f"Device {k} is not recognized, available device are integers(for GPU/XPU), 'mps', 'cpu' and 'disk'"
```

sgugger (commented):
Thanks for working on this! I left a couple of comments.

ranchlai:
> Thanks for working on this! I left a couple of comments.
thanks very much for commenting! @sgugger

sgugger (approved):

sgugger:
Thanks for iterating!

PR #1762, opened 2023-07-24
Fix offload on disk when executing on CPU

sgugger:
This PR fixes the disk offload of models when using the CPU as a main device.


osanseviero:
Thanks so much for the debugging and fixing <3

muellerzr (approved):
Nice catch!

PR #1769, opened 2023-07-25
Change is_aim_available() function to not match aim >= 4.0.0

alberttorosyan:
With the upcoming major version of Aim, package structure and SDK will change without providing backwards compatibility.

In order to make sure nothing fails on the HF accelerate side, modified the `is_aim_available()` check to return True value only for Aim versions 3.x, by adding the availability check for `aim.sdk` sub-package.

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -39,13 +39,9 @@ except ImportError:
 _torch_distributed_available = torch.distributed.is_available()
 
 
-def _package_exists(pkg_name):
-    return importlib.util.find_spec(pkg_name) is not None
-
-
 def _is_package_available(pkg_name):
     # Check we're not importing a "pkg_name" directory somewhere but the actual library by trying to grab the version
-    package_exists = _package_exists(pkg_name)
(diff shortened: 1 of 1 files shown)


sgugger:
cc @muellerzr

muellerzr (approved):
Thanks!

muellerzr:
@alberttorosyan should this only be merged after your release then?

alberttorosyan:
> @alberttorosyan should this only be merged after your release then?

@muellerzr it should be merged before Aim release, to prevent failures. If we do other way around, in the time between releasing Aim and merging this PR `import accelerate` will fail, due to the following line:
https://github.com/huggingface/accelerate/blob/0b36ca6e64245b123e9455521c1bd985f4b72679/src/accelerate/tracking.py#L60

alberttorosyan:
> If this PR is just limiting them to < 3.0.0, let's use the available version checker code then via utils.imports.compare_versions

@muellerzr, yes, such check would be perfect! Will push appropriate changes shortly.

muellerzr (approved):
Great! Thanks

Issue #1764, opened 2023-07-24
RuntimeError('Tensors must be contiguous') 


adildahlan:
### System Info

```Shell
Libraries:
pytorch                   2.0.1
pytorch-cuda              11.8
numpy                     1.24.3
accelerate                0.21.0
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

When using the following accelerate config my code crashes as shown in the screenshot below:
```
  1 compute_environment: LOCAL_MACHINE
  2 distributed_type: MULTI_GPU
  3 downcast_bf16: 'no'
  4 gpu_ids: all
  5 machine_rank: 0
  6 main_training_function: main
  7 mixed_precision: 'no'
  8 num_machines: 1
  9 num_processes: 1
 10 rdzv_backend: static
 11 same_network: true
 12 tpu_env: []
 13 tpu_use_cluster: false
 14 tpu_use_sudo: false
 15 use_cpu: false
 16 main_process_port: 8080
```

<img width="817" alt="image" src="https://github.com/huggingface/accelerate/assets/65319353/1518ffc5-a5d4-4296-a4ef-b6a3427f8ebd">

### Expected behavior

Whereas when using the following accelerate config, the code runs perfectly:

```
  1 compute_environment: LOCAL_MACHINE
  2 distributed_type: MULTI_GPU
  3 downcast_bf16: 'no'
  4 gpu_ids: all
  5 machine_rank: 0
  6 main_training_function: main
  7 mixed_precision: fp16
  8 num_machines: 1
  9 num_processes: 1
 10 rdzv_backend: static
 11 same_network: true
 12 tpu_env: []
 13 tpu_use_cluster: false
 14 tpu_use_sudo: false
 15 use_cpu: false
 16 main_process_port: 8080
```

Doing some online search, I came across this issue https://github.com/huggingface/transformers/issues/20942 from the transformers repo. 

Something funny is happening with mixed precision setup. I hope you could help me sort this bug out!

Thank you

muellerzr:
Please provide us a reproducer, because the part of code mentioned there is still in the original codebase.

adildahlan:
[...]

PR #1771, opened 2023-07-25
Contigous on gather

muellerzr:
Solves https://github.com/huggingface/transformers/issues/25076
Solves https://github.com/huggingface/accelerate/issues/1764

We need to ensure tensors are `contiguous` before gathering, else torch will raise an error.


sgugger (approved):
Thanks, nice test!

Issue #1767, opened 2023-07-25
gradient_accumulation with deepspeed 


yoctta:
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-3.10.0-957.el7.x86_64-x86_64-with-glibc2.17
- Python version: 3.10.12
- Numpy version: 1.25.0
- PyTorch version (GPU?): 2.0.1 (False)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 723.61 GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - mixed_precision: bf16
        - use_cpu: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - deepspeed_config: {'gradient_accumulation_steps': 16, 'gradient_clipping': 3.0, 'offload_optimizer_device': 'none', 'offload_param_device': 'none', 'zero3_init_flag': False, 'zero_stage': 2}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {'dynamo_backend': 'INDUCTOR', 'dynamo_mode': 'reduce-overhead', 'dynamo_use_dynamic': False, 'dynamo_use_fullgraph': False}
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

examples/by_feature/deepspeed_with_config_support.py

Line 643             loss = loss / args.gradient_accumulation_steps

~~~~~~~~~~~~~~~~~~~~~

https://github.com/huggingface/accelerate/blob/0b36ca6e64245b123e9455521c1bd985f4b72679/src/accelerate/accelerator.py#L1901

        if self.distributed_type != DistributedType.DEEPSPEED:
            # deepspeed handles loss scaling by gradient_accumulation_steps in its `backward`
            loss = loss / self.gradient_accumulation_steps
        if self.distributed_type == DistributedType.DEEPSPEED:
            self.deepspeed_engine_wrapped.backward(loss, **kwargs)

### Expected behavior

[...]

Issue #1773, opened 2023-07-26
Loss in deepspeed_with_config_support.py in examples


sakshidotvakilsearch:
In `accelerate/examples/by_feature/deepspeed_with_config_support.py` at line 643 (https://github.com/huggingface/accelerate/blob/a2d8f540c3ab37c8f84d616be1300a0572b69cf8/examples/by_feature/deepspeed_with_config_support.py#L643) loss is being divided with gradient accumulation steps before accelerator.backward()

Is this happening because accelerate or deepspeed internally (before or during backpass) are performing a multiplicative operation with loss & gradient accumulation steps in distributed settings or is this a mistake?

Tried checking out source code but frankly couldn't get much

pacman100:
Hello, this is duplicate of https://github.com/huggingface/accelerate/issues/1767

sakshidotvakilsearch:
@pacman100 when you do accelerator.accumulate() & accelerate.backward(), you are also stepping optimiser & scheduler without checking if global % grad_steps == 0 or not. Is that because accelerator.accumulate() handles it internally? If it does, does that mean accelerate tracks the global step and epochs internally?

PR #1775, opened 2023-07-26
simplify and correct the deepspeed example

pacman100:
### What does this PR do?

1. Fixes https://github.com/huggingface/accelerate/issues/1767 and https://github.com/huggingface/accelerate/issues/1773
2. Simplifies the example by removing the redundant saving/loading utils for DeepSpeed as they are supported via `accelerator.save_state()` and `accelerator.load_state()`.

### DIFF
--- a/examples/by_feature/deepspeed_with_config_support.py
+++ b/examples/by_feature/deepspeed_with_config_support.py
@@ -649,7 +649,7 @@ def main():
                 {
                     "perplexity": perplexity,
                     "eval_loss": eval_loss,
-                    "train_loss": total_loss.item() / len(train_dataloader),
+                    "train_loss": total_loss / len(train_dataloader),
                     "epoch": epoch,
                     "step": completed_steps,
                 },


sgugger (approved):
Thanks!

sakshidotvakilsearch:
Hi can someone explain how these 5 lines of code really work in terms of accelerate + deepspeed?

```
          outputs = model(**batch)
          loss = outputs.loss
          accelerator.backward(loss)
          optimizer.step()
          lr_scheduler.step()
          optimizer.zero_grad()
```

line 618-624 in https://github.com/huggingface/accelerate/blob/main/examples/by_feature/deepspeed_with_config_support.py

Looking at the source code, accelerator.backward() will call the deepspeed backward(), which maybe handling some gather-reduce-scatter operations internally (please feel free to elaborate on this too). But post that, the optimizer & scheduler are stepped & optimiser is reset too. Is it a bug that the global step value is not checked before resetting the optimiser such that `global_step % gradient_accumulation_steps == 0`

As far as I understand, deepspeed & accelerate are handling gradient accumulation internally but is the global step tracked by accelerate such that when accelerate.accumulate() is called, it checks whether it is the right step to perform backpass or not? Also, is accumulate triggering the gather operation?

sakshidotvakilsearch:
@pacman100 the completed_steps calculation when training is resumed from a checkpoint that as saved with steps, resumes training from a lot earlier than the step from which it should be resuming. The resume step is supposed to be calculated for epoch while completed_steps is global. so I think this is how it should look like
```
if 'epoch' in training_difference:
    starting_epoch = int(training_difference.replace('epoch_', '')) + 1
    resume_step = None
    completed_steps = starting_epoch * num_update_steps_per_epoch
else:
    resume_step = int(training_difference.replace('step_', ''))
    completed_steps = resume_step
    starting_epoch = resume_step // num_update_steps_per_epoch
    resume_step -= starting_epoch * num_update_steps_per_epoch
```

Your code assigns completed_steps to resume_steps at the end after resume_step calculation

https://github.com/huggingface/accelerate/blob/727d624322c67db66a43c559d8c86414d5ffb537/examples/by_feature/deepspeed_with_config_support.py#L603C20-L603C20

Issue #1774, opened 2023-07-26
Error  on macOS: AttributeError: 'AcceleratorState' object has no attribute 'use_ipex'. Did you mean: 'use_fp16'?


dyzdyz010:
CPU： M2 Max
accelerate: 0.21.0
OS: macOS Ventura 13.14.1
python version: 3.10
numpy version: 1.24.0
torch version: 2.1.0.dev20230722

I'm using [Llama2-Chinese training script](https://github.com/FlagAlpha/Llama2-Chinese/tree/main/train/sft) to train my llama2-based model, but it shows the error: 

```
Traceback (most recent call last):
  File "/Users/dyz/Developer/llm/Llama2-Chinese/train/sft/finetune_clm_lora.py", line 691, in <module>
    main()
  File "/Users/dyz/Developer/llm/Llama2-Chinese/train/sft/finetune_clm_lora.py", line 652, in main
    train_result = trainer.train(resume_from_checkpoint=checkpoint)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/gpt/lib/python3.10/site-packages/transformers/trainer.py", line 1539, in train
    return inner_training_loop(
  File "/opt/homebrew/Caskroom/miniconda/base/envs/gpt/lib/python3.10/site-packages/transformers/trainer.py", line 1553, in _inner_training_loop
    train_dataloader = self.get_train_dataloader()
  File "/opt/homebrew/Caskroom/miniconda/base/envs/gpt/lib/python3.10/site-packages/transformers/trainer.py", line 854, in get_train_dataloader
    return self.accelerator.prepare(DataLoader(train_dataset, **dataloader_params))
  File "/opt/homebrew/Caskroom/miniconda/base/envs/gpt/lib/python3.10/site-packages/accelerate/accelerator.py", line 1193, in prepare
    if self.device.type == "cpu" and self.state.use_ipex:
AttributeError: 'AcceleratorState' object has no attribute 'use_ipex'. Did you mean: 'use_fp16'?
```

muellerzr:
Hi! Can you install accelerate via `pip install git+https://github.com/huggingface/accelerate@ipex-err` and let me know if that solves it for you? Thanks!

dyzdyz010:
This seems to fix my issue, thanks!😊

PR #1776, opened 2023-07-26
Set ipex default in state

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1774 by ensuring we have a default in the `State` for `self.use_ipex`


sgugger (approved):
Thanks for the fix!

abhilash1910 (commented):
Thanks for this fix @muellerzr !

PR #1790, opened 2023-07-28
add warning when using to and cuda

SunMarc:
# What does this PR do ? 
This PR wrap `to` and `cuda` method when we dispatch a model on multiple devices. We send a send a warning when these methods are called as they should not move the model for inference/training. We also give a better error message when the user tries to move a model that has some modules offloaded to cpu or disk.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 from contextlib import contextmanager
 from typing import Dict, List, Optional, Union
@@ -38,6 +39,9 @@ from .utils import (
     offload_state_dict,
     retie_parameters,
 )
+
+
+logger = logging.getLogger(__name__)
 
 
 @contextmanager
@@ -379,6 +383,21 @@ def dispatch_model(
         )
         # Attaching the hook may break tied weights, so we retie them
         retie_parameters(model, tied_params)
+
+        # add warning to cuda and to method
+        def add_warning(fn):
+            def wrapper(*args, **kwargs):
+                logger.warning(
+                    "You can't use the model anymore for training or inference as you moved the model."
+                    "You should not move the model when it is dispatched on multiples devices. "
+                )
+                return fn(*args, **kwargs)
+
+            return wrapper
+
+        model.to = add_warning(model.to)
+        model.cuda = add_warning(model.cuda)
+
     else:
         device = list(device_map.values())[0]
         if device != "disk":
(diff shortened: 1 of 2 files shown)


src/accelerate/big_modeling.py
```diff
@@ -379,6 +384,25 @@ def dispatch_model(
...
         # Attaching the hook may break tied weights, so we retie them
         retie_parameters(model, tied_params)
+
+        # add warning to cuda and to method
+        def add_warning(fn):
+            @wraps(fn)
+            def wrapper(*args, **kwargs):
+                logger.warning(
+                    "You can't use the model anymore for training or inference as you moved the model."
+                    "You should not move the model when it is dispatched on multiples devices. "
```
sgugger, line 394: This warning is not super clear. I would say something more like "You can't move a model when it is dispatched on multiple device." The model can still be used for inference and training.

sgugger (commented):
Thanks for adding this!

src/accelerate/big_modeling.py
SunMarc: changed !

src/accelerate/big_modeling.py
```diff
@@ -379,6 +384,22 @@ def dispatch_model(
...
+
+        # add warning to cuda and to method
+        def add_warning(fn):
+            @wraps(fn)
+            def wrapper(*args, **kwargs):
+                logger.warning("You can't move a model when it is dispatched on multiple devices.")
+                try:
+                    return fn(*args, **kwargs)
+                except NotImplementedError as e:
+                    raise RuntimeError("You can't move a model that has some modules offloaded to cpu or disk.") from e
```
sgugger, line 396: Maybe instead of the try except here we could detect if the model has some offloaded parameters and raise the error then.

sgugger (approved):
Just one last suggestion, but LGTM otherwise

src/accelerate/big_modeling.py
SunMarc: I do that by checking if any parameter is on `meta` device. Let me know that works for you. I don't check the device_map because people might remove the hooks after the dispatching the model.

src/accelerate/big_modeling.py
```diff
@@ -379,6 +384,22 @@ def dispatch_model(
         )
         # Attaching the hook may break tied weights, so we retie them
         retie_parameters(model, tied_params)
+
+        # add warning to cuda and to method
+        def add_warning(fn, model):
+            @wraps(fn)
+            def wrapper(*args, **kwargs):
+                logger.warning("You can't move a model when it is dispatched on multiple devices.")
```
sgugger, line 392:
```suggestion
                logger.warning("You shouldn't move a model when it is dispatched on multiple devices.")
```

sgugger (approved):
One last nit and we should be good to go!

PR #1798, opened 2023-07-31
reserve 10% GPU in `get_balanced_memory` to avoid OOM

ranchlai:
# What does this PR do?
This PR fixes OOM problem in the case of 1 GPU and device_map="auto", by reserving 10% memory for buffer(e.g. , storing the text embeddings). Manually setting max_memory or device_map will override this change. 

This is a patch to `get_balanced_memory()` in the case of `num_devices==1`.

The OOM problem happens for some LLMs, such as chatglm, llama, etc. 

Hope it's acceptable.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -779,7 +779,7 @@ def get_balanced_memory(
             for key in max_memory.keys():
                 if isinstance(key, int):
                     max_memory[key] *= 0.9  # 90% is a good compromise
-                    logger.warning(
+                    logger.info(
                         f"We will use 90% of the memory on device {key} for storing the model, and 10% for the buffer to avoid OOM. "
                         "You can set `max_memory` in to a higher value to use more memory (at your own risk)."
                     )
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -896,8 +896,6 @@ def infer_auto_device_map(
         devices.append("disk")
     gpus = [device for device in devices if device not in ["cpu", "disk"]]
 
-    # find the lowest cardinal of gpus, lower
-
     # Devices that need to keep space for a potential offloaded layer.
     if "mps" in gpus:
         main_devices = ["mps"]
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -771,8 +772,18 @@ def get_balanced_memory(
...
     if num_devices == 1:
-        # We cannot do low_zero on just one GPU
+        # We cannot do low_zero on just one GPU, but we will still reserve some memory for the buffer
         low_zero = False
+        # If user just asked us to handle memory usage, we should avoid OOM
+        if user_not_set_max_memory:
+            for key in max_memory.keys():
+                if isinstance(key, int):
+                    max_memory[key] *= 0.9  # 90% is a good compromise
+                    logger.warning(
```
sgugger, line 782: Let's do an info here, as users are scared of warnings.

src/accelerate/utils/modeling.py
```diff
@@ -771,8 +772,18 @@ def get_balanced_memory(
...
+                        f"We will use 90% of the memory on device {key} for storing the model, and 10% for the buffer to avoid OOM. "
+                        "You can set `max_memory` in to a higher value to use more memory."
```
sgugger, line 784:
```suggestion
                        "You can set `max_memory` in to a higher value to use more memory (at your own risk)."
```

src/accelerate/utils/modeling.py
```diff
@@ -885,6 +896,8 @@ def infer_auto_device_map(
         devices.append("disk")
     gpus = [device for device in devices if device not in ["cpu", "disk"]]
 
+    # find the lowest cardinal of gpus, lower
```
sgugger, line 899: Not sure what this comment is about?

sgugger (commented):
Thanks for your PR! That does seem like a good idea!

src/accelerate/utils/modeling.py
ranchlai: sorry. removed.

src/accelerate/utils/modeling.py
ranchlai: done~

ranchlai:
@sgugger thanks for your kind reviews.

sgugger (approved):
Thanks again!

PR #1799, opened 2023-07-31
add support of float memory size in convert_file_size_to_int

ranchlai:
# What does this PR do
This PR add support for more flexible control of  memory size such as  "5.0GB, 0.5GB" . 
I also add some code to handle invalid memory size values.

Hope this is useful.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -59,24 +59,34 @@ def convert_file_size_to_int(size: Union[int, str]):
     1048576
     ```
     """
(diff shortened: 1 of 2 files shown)


sgugger (approved):
Nice improvement! Could you just add some unit tests in `test_modeling_utils` (I know there is none, adding one would be super great)

ranchlai:
> test_modeling_utils

Sure thing. Already done~ thanks again for reviewing.

sgugger (approved):
Thank you so much!

PR #1808, opened 2023-08-03
Fix pytest import

muellerzr:
Guards pytest import in test util file


sgugger (commented):
Not sure this solves the problem since there are going to be errors later on when `pytest` is not imported, no?

muellerzr:
@sgugger worked around it by writing our own `pytest.Raises` instead

Issue #1802, opened 2023-07-31
accelerator.prepare() error: DistributedDataParallel device_ids and output_device arguments only work with single-device/multiple-device GPU modules or CPU modules


nebrelbug:
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-5.4.0-58-generic-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.24.4
- PyTorch version (GPU?): 2.0.1+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 1510.57 GB
- GPU type: Tesla V100-SXM3-32GB
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Here's some code that loads LLaMA weights and then calls `accelerator.prepare(model, optimizer, dataloader)`. It's part of a training loop I wrote, but the sample below is sufficient to cause an error. I'm just a beginner :sweat_smile:, so I may have made a mistake somehow, but the code works fine without using Accelerate.

**Training Code**

```python
from transformers import LlamaForCausalLM, LlamaTokenizer
from torch.utils.data import Dataset, DataLoader
from torch.optim import AdamW

from accelerate import Accelerator
import json

accelerator = Accelerator()

# MODEL LOADING

MODEL_PATH = "//mnt/pccfs2/backed_up/models/llama/hf/llama-7b-hf"

tokenizer = LlamaTokenizer.from_pretrained(MODEL_PATH, legacy=False)

model = LlamaForCausalLM.from_pretrained(MODEL_PATH, device_map="auto")

optimizer = AdamW(model.parameters(), lr=3e-5)

model.train()  # tell model you're training

# DATALOADER


class StringDataset(Dataset):
    def __init__(self, string_list, tokenizer):
        self.string_list = string_list
        self.tokenizer = tokenizer

    def __len__(self):
        return len(self.string_list)

    def __getitem__(self, idx):
        text = self.string_list[idx]
        tokens = self.tokenizer(
            text,
            padding="max_length",
            truncation=True,
            return_tensors="pt",
        ).to("cuda")

[...]

PR #1810, opened 2023-08-04
Detect device map auto and raise a helpful error when trying to not use model parallelism

muellerzr:
Solves https://github.com/huggingface/accelerate/issues/1802 by raising an error if a user tries doing `.prepare()` on a model that contains a device map and we are in something other than `Distributed.NO

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1325,6 +1325,12 @@ class Accelerator:
                 has_hf_device_map = True
                 break
 
+        if has_hf_device_map and self.distributed_type != DistributedType.NO:
+            raise ValueError(
+                "You can't train a model that has been loaded with `device_map='auto'` in any distributed mode."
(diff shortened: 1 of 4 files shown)


sgugger (approved):
Thanks for adding this!

0xDing:
@muellerzr This change looks like it will cause [qlora](https://github.com/artidoro/qlora/blob/7f4e95a68dc076bea9b3a413d2b512eca6d004e5/qlora.py#L289) to not work properly, qlora will set the device map and max memory per device. it might be better to change it to only throw errors when using deepspeed or fsdp.

sgugger:
@0xDing The error is just moved to be clearer: if you try to put a model on several devices (as returned with `device_map="auto"`) in a DDP wrapper, PyTorch will error out.

The only thing we could refine is only error when the `hf_device_map` contains more than one device, but `dispatch_model` should not set it in that case.

Issue #1807, opened 2023-08-03
Conflict between notebook_launcher function and bitsandbytes package.


looperEit:
When I was fine-tuning baichuan13b, I found a problem. If I want to use accelerate for training, if I import the bitsandbytes package, an error will be reported when using accelerate for training.
This is because the `notebook_launcher` function checks` torch.cuda.is_initialized()`. If this variable is true, an error will be reported:

ckpt_path = 'baichuan13b_ner'

optimizer = bnb.optim.adamw.AdamW(peft_model.parameters(),
                                  lr=6e-05,is_paged=True)  #'paged_adamw'

keras_model = KerasModel(peft_model, loss_fn =None,  optimizer=optimizer)

keras_model.load_ckpt(ckpt_path)

keras_model.fit_ddp(num_processes=2,
                    train_data=dl_train,
                    val_data=dl_val,
                    epochs=100,
                    patience=10,
                    monitor='val_loss',
                    mode='min',
                    ckpt_path=ckpt_path)

>  ---------------------------------------------------------------------------
>  ValueError                                Traceback (most recent call last)
>  Cell In[30], line 12
>        9 keras_model.load_ckpt(ckpt_path)
>       11 # 使用多GPU训练
>  ---> 12 keras_model.fit_ddp(num_processes=2,
>       13                     train_data=dl_train,
>       14                     val_data=dl_val,
>       15                     epochs=100,
>       16                     patience=10,
>       17                     monitor='val_loss',
>       18                     mode='min',
>       19                     ckpt_path=ckpt_path)
>  
>  File ~/anaconda3/envs/baichuan13b/lib/python3.9/site-packages/torchkeras/kerasmodel.py:282, in KerasModel.fit_ddp(self, num_processes, train_data, val_data, epochs, ckpt_path, patience, monitor, mode, callbacks, plot, wandb, quiet, mixed_precision, cpu, gradient_accumulation_steps)
>      279 from accelerate import notebook_launcher
>      280 args = (train_data,val_data,epochs,ckpt_path,patience,monitor,mode,
>      281     callbacks,plot,wandb,quiet,mixed_precision,cpu,gradient_accumulation_steps)
>  --> 282 notebook_launcher(self.fit, args, num_processes=num_processes)
>  
>  File ~/anaconda3/envs/baichuan13b/lib/python3.9/site-packages/accelerate/launchers.py:116, in notebook_launcher(function, args, num_processes, mixed_precision, use_port)
[...]

PR #1813, opened 2023-08-04
Fix bnb import

muellerzr:
`bitsandbytes` will always initialize CUDA on import as it's designed to be imported/ran from the CLI. As a result this makes it fully incompatible with the `notebook_launcher`. To try and mitigate this, I've moved all of the import statements to be inside the functions that require the items to use them. Not the cleanest solution, but as this behavior has been consistent in bnb for many versions I don't believe this is likely to change in the future, so we need to adapt for it as a result. 

Solves https://github.com/huggingface/accelerate/issues/1807


sgugger (approved):
Nice catch!

PR #1817, opened 2023-08-07
Improve docs on grad accumulation

vwxyzjn:
Also added a self-contained runnable example.

### DIFF
--- a/docs/source/usage_guides/gradient_accumulation.md
+++ b/docs/source/usage_guides/gradient_accumulation.md
@@ -127,6 +127,11 @@ training on. 🤗 Accelerate automagically does this for you by default. Behind 
 Below is the finished implementation for performing gradient accumulation with 🤗 Accelerate
 
 ```python
+from accelerate import Accelerator
+accelerator = Accelerator(gradient_accumulation_steps=2)
+model, optimizer, training_dataloader, scheduler = accelerator.prepare(
+    model, optimizer, training_dataloader, scheduler
+)
 for batch in training_dataloader:
     with accelerator.accumulate(model):
         inputs, targets = batch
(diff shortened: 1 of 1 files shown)


docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -138,4 +143,77 @@ for batch in training_dataloader:
         optimizer.zero_grad()
 ```
 
+<Tip warning={true}>
+  It's important that you enter the `with accelerator.accumulate(model)` context manager before you perform the forward and backward pass of your model. In other words, do **not** do this:
```
sgugger, line 147: In the example, you do enter the contextmanager before performing the forward and backward pass, so I'd rephrase this ;-)

Also no indent in the Tip, just new lines after/before the <Tip>/</Tip> :-)

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -138,4 +143,77 @@ for batch in training_dataloader:
...
+Here is a self-contained example that you can run to see gradient accumulation in action with 🤗 Accelerate:
+
+```python
+import torch
+from torch.utils.data import TensorDataset, DataLoader
+import copy
+from accelerate import Accelerator
+
+# seed
+torch.manual_seed(0)
```
muellerzr, line 171:
```suggestion
import torch
import copy
from accelerate import Accelerator
from accelerate.utils import set_seed
from torch.utils.data import TensorDataset, DataLoader

# seed
set_seed(0)
```

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -138,4 +143,77 @@ for batch in training_dataloader:
...
+
+# define toy inputs and labels
+x = torch.tensor([1., 2., 3., 4., 5., 6., 7., 8.])
+y = torch.tensor([2., 4., 6., 8., 10., 12., 14., 16.])
+gradient_accumulation_steps = 4
+batch_size = int(len(x) // gradient_accumulation_steps)
```
muellerzr, line 177: Not sure the cast to `int` is needed here since `//` will always return an integer

muellerzr (commented):
Thanks! Along with Sylvain's comments a few style nits in your example.

docs/source/usage_guides/gradient_accumulation.md
vwxyzjn: Thanks for the suggestion. I added this tip because it was not obvious to me that the context manager needs to ve 1:1 with the forward and backward pass. I had something like below, which was wrong. What would be your suggested phrasing? 

```
import torch
from torch.utils.data import TensorDataset, DataLoader
import copy

# seed
torch.manual_seed(0)

# define toy inputs and labels
x = torch.tensor([1., 2., 3., 4., 5., 6., 7., 8.])
y = torch.tensor([2., 4., 6., 8., 10., 12., 14., 16.])
gradient_accumulation_steps = 4
batch_size = int(len(x) // gradient_accumulation_steps)

# define model, optimizer and loss function
model = torch.zeros((1, 1), requires_grad=True)
model_clone = copy.deepcopy(model)
criterion = torch.nn.MSELoss()
model_optimizer = torch.optim.SGD([model], lr=0.02)
accelerator = Accelerator(gradient_accumulation_steps=gradient_accumulation_steps)
model, model_optimizer = accelerator.prepare(model, model_optimizer)
model_clone_optimizer = torch.optim.SGD([model_clone], lr=0.02)
print(f"initial model weight is {model.mean().item():.5f}")
print(f"initial model weight is {model_clone.mean().item():.5f}")
for _ in range(2):
    # loop over batches
        for mb_start in range(0, len(x), batch_size):
            inputs = x[mb_start:mb_start+batch_size].view(-1, 1)
            labels = y[mb_start:mb_start+batch_size].view(-1, 1)
            print(mb_start, inputs.flatten())
            outputs = inputs @ model
            loss = criterion(outputs, labels)
            accelerator.backward(loss)
            model_optimizer.step()
            model_optimizer.zero_grad()
    print(f"w/ accumulation, the final model weight is {model.mean().item():.5f}")
    loss = criterion(x.view(-1, 1) @ model_clone, y.view(-1, 1))
    model_clone_optimizer.zero_grad()
    loss.backward()
    model_clone_optimizer.step()
    print(f"w/o accumulation, the final model weight is {model_clone.mean().item():.5f}")
```
```
initial model weight is 0.00000
initial model weight is 0.00000
0 tensor([1., 2.])
2 tensor([3., 4.])
4 tensor([5., 6.])
6 tensor([7., 8.])
w/ accumulation, the final model weight is 0.00000
0 tensor([1., 2.])
2 tensor([3., 4.])
4 tensor([5., 6.])
6 tensor([7., 8.])
w/ accumulation, the final model weight is 0.00000
w/o accumulation, the final model weight is 1.99920
```

sgugger: You want to say that only one forward/backward should be done inside the context manager I think.

vwxyzjn: Addressed. Thanks!

vwxyzjn:
Hi @muellerzr and @sgugger thanks for the prompt review. Let me know if further changes are needed :)

muellerzr (approved):
Thanks! (With Sylvain's spacing nit for our docs :) )

PR #1822, opened 2023-08-07
Update docs and docstrings to match `load_and_quantize_model` arg

JonathanRayner:
`load_and_quantize_model` [has an arg](https://github.com/huggingface/accelerate/blob/v0.21.0/src/accelerate/utils/bnb.py#L52) `bnb_quantization_config`, but slightly different versions of this were listed in the docs and docstrings.


sgugger (approved):
Thanks! Can you also have a look and merge if you're happy @SunMarc ?

SunMarc (approved):
Thanks for updating the docs and docstings!

PR #1824, opened 2023-08-08
Expose a bit of args/docstring fixup

muellerzr:
Applies some feedback from an in-depth doc review, and removes chunks about minimal torch requirements since we now require torch 1.10+

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -106,7 +106,7 @@ class PartialState:
           in use.
         - **local_process_index** (`int`) -- The index of the current process on the current server.
         - **mixed_precision** (`str`) -- Whether or not the current script will use mixed precision, and if so the type
-          of mixed precision being performed.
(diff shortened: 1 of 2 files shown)


src/accelerate/state.py
```diff
@@ -106,7 +106,7 @@ class PartialState:
           in use.
         - **local_process_index** (`int`) -- The index of the current process on the current server.
         - **mixed_precision** (`str`) -- Whether or not the current script will use mixed precision, and if so the type
-          of mixed precision being performed.
+          of mixed precision being performed. (One of `"fp16"`, `"bf16"`, or `"fp8"`).
```
sgugger, line 109: Can't be no here? If that's the case the sentence before should be rephrased.

sgugger (approved):
Thanks!

src/accelerate/state.py
muellerzr: Rephrasedd to fully mimic the version in the accelerator

PR #1825, opened 2023-08-08
Better test

muellerzr:
Writes a better test that doesn't depend on the order the test is called in for us to check torch cuda (why main has been failing). Verified by removing the hack of having it go first and test via `pytest -sv tests/test_accelerator.py` on a multi-GPU machine. 

Helpful tip: if for whatever reason you care about the execution of a test, send it to a new test `.py` file in `test_utils/scripts` to run off of an isolated new process cc @SunMarc @BenjaminBossan


sgugger (approved):
Thanks for adding!

PR #1826, opened 2023-08-08
Use device as context manager for init_on_device

### LINKED ISSUE #1814: `big_modeling.init_on_device` not working with WrapperTensor class like `FakeTensor`
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-5.19.0-45-generic-x86_64-with-glibc2.17
- Python version: 3.8.16
- Numpy version: 1.21.2
- PyTorch version (GPU?): 2.1.0.dev20230725+cu118 (True)
- System RAM: 62.70 GB
- GPU type: NVIDIA GeForce RTX 3070
        Not found
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

My repro:

```
import torch
from torch._subclasses.fake_tensor import FakeTensorMode
from accelerate.big_modeling import init_on_device

fake_mode = FakeTensorMode()
with fake_mode:
    with init_on_device(torch.device('cuda')):
        net = torch.nn.Linear(3,4)
```

### Expected behavior

As torch introduced WrapperTensor class and TensorDispatchMode, it is possible that user may want to create custom tensors/parameters while loading from transformers/diffusers. Ideally the above code should work with any WrapperTensor class. However, `big_modeling.init_on_device` assume that everything in `WrapperTensor.__dict__` should be keyword argument of `WrapperTensor.new()`, which may not be the case for `WrapperTensor` classes.

Output:
```
  File "ex.py", line 16, in <module>
    net = torch.nn.Linear(3,4)
  File "python3.8/site-packages/torch/nn/modules/linear.py", line 96, in __init__
    self.weight = Parameter(torch.empty((out_features, in_features), **factory_kwargs))
  File "python3.8/site-packages/torch/nn/modules/module.py", line 1712, in __setattr__
    self.register_parameter(name, value)
  File "python3.8/site-packages/accelerate/big_modeling.py", line 105, in register_empty_parameter
    module._parameters[name] = param_cls(module._parameters[name].to(device), **kwargs)
TypeError: __new__() got an unexpected keyword argument 'fake_device'
```

shingjan:
fixes #1814 

cc: @sgugger

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -34,6 +34,7 @@ from .utils import (
     find_tied_parameters,
     get_balanced_memory,
     infer_auto_device_map,
+    is_torch_version,
     load_checkpoint_in_model,
     offload_state_dict,
     retie_parameters,
@@ -93,6 +94,12 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
         tst = nn.Liner(100, 100)  # on `cuda` device
     ```
     """
+    # TODO(shingjan): remove the torch version check once older versions are deprecated
+    if is_torch_version(">=", "2.0") and include_buffers:
+        with device:
+            yield
+        return
+
     old_register_parameter = nn.Module.register_parameter
     if include_buffers:
         old_register_buffer = nn.Module.register_buffer
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -31,7 +31,7 @@ from accelerate.big_modeling import (
 )
 from accelerate.hooks import remove_hook_from_submodules
 from accelerate.test_utils import require_bnb, require_cuda, require_mps, require_multi_gpu, slow
-from accelerate.utils import offload_state_dict
+from accelerate.utils import is_torch_version, offload_state_dict
 
 
 class ModelForTest(nn.Module):
@@ -106,8 +106,14 @@ class BigModelingTester(unittest.TestCase):
         self.assertEqual(module.running_mean.device, torch.device("cpu"))
 
         # Use with include_buffers=True
+        register_parameter_func = nn.Module.register_parameter
+        register_buffer_func = nn.Module.register_buffer
         with init_empty_weights(include_buffers=True):
             module = nn.BatchNorm1d(4)
+            # nn.Module.register_parameter/buffer shouldn't be changed with torch >= 2.0
+            if is_torch_version(">=", "2.0"):
+                self.assertEqual(register_parameter_func, nn.Module.register_parameter)
+                self.assertEqual(register_buffer_func, nn.Module.register_buffer)
         self.assertEqual(module.weight.device, torch.device("meta"))
         self.assertEqual(module.running_mean.device, torch.device("meta"))
 


src/accelerate/big_modeling.py
```diff
@@ -41,7 +41,7 @@
 
 
 @contextmanager
-def init_empty_weights(include_buffers: bool = False):
+def init_empty_weights(include_buffers: bool = True):
```
sgugger, line 44: We can't change this, it would break the integration with Transformers :-)

src/accelerate/big_modeling.py
```diff
@@ -73,7 +73,7 @@ def init_empty_weights(include_buffers: bool = False):
 
 
 @contextmanager
-def init_on_device(device: torch.device, include_buffers: bool = False):
+def init_on_device(device: torch.device, include_buffers: bool = True):
```
sgugger, line 76: Same here.

src/accelerate/big_modeling.py
```diff
@@ -93,6 +93,13 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
         tst = nn.Liner(100, 100)  # on `cuda` device
     ```
     """
+    if include_buffers and hasattr(device, "__enter__") and callable(getattr(device, "__enter__")):
```
sgugger, line 96: Here it would be better to use a version check on PyTorch.

tests/test_big_modeling.py
```diff
@@ -95,12 +95,12 @@ def forward(self, x):
 class BigModelingTester(unittest.TestCase):
     def test_init_empty_weights(self):
         # base use
-        with init_empty_weights():
+        with init_empty_weights(include_buffers=False):
```
sgugger, line 98: This shouldn't be changed (above below).

src/accelerate/big_modeling.py
```diff
@@ -93,6 +93,13 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
...
+        try:
+            device.__enter__()
+            yield
+        finally:
+            return device.__exit__()
```
BenjaminBossan, line 101: Cool, I didn't know that `device` supports this now.

Instead of entering and exiting explicitly, would it be possible to use it like this?

```python
with device:
    yield
return
```

src/accelerate/big_modeling.py
BenjaminBossan: + a `TODO` comment to remove the version check once older PyTorch versions are no longer supported.

BenjaminBossan (commented):
Thanks for the PR. Is it possible to have a test for this or would it be too complicated to set up because of the version dependency?

shingjan:
Thanks @sgugger and @BenjaminBossan for the reviews!

I added one test with `include_buffers=True` to make if testing with torch >= 2.0, we go with the device context manager. This PR however won't be fixing the issue I had before:

```
with tensor_mode():
  model = transformers.from_pretrained(...)
```
as `init_empty_weights(...)` uses `include_buffers=False` as default. I wonder if you have a better context on why we use `include_buffers=False` as default for `init_empty_weights`. If the ultimate goal is to move a nn.Module to meta device in order to save RAM, wouldn't buffer also be included in the state dict of a nn.Module and therefore better to move those to meta as well?

sgugger (approved):
Nice new test!

shingjan:
thanks @sgugger and @BenjaminBossan for getting this PR in!

PR #1829, opened 2023-08-09
Minor idiomatic change for fp8 check.

float-trip:
Tuple comparison in Python is a slightly cleaner way to check that the major/minor is >= 8.9.


sgugger (approved):
Thansk!

PR #1833, opened 2023-08-10
Change CUDA check

muellerzr:
While we need to check if a GPU is available to disable XLA, we need to use the NVML version of `torch.cuda.is_available()` otherwise any import will spawn a call to `torch.cuda` and break the `notebook_launcher`. Uses recommendation from the torch team of spawning a dummy process to verify it will run


sgugger (approved):
Thanks. If you find a way to check if CUDA is available without initializaing CUDA, that would be great.

src/accelerate/test_utils/scripts/test_notebook.py
```diff
@@ -1,10 +1,17 @@
...
 
-from accelerate import Accelerator  # noqa
 
-assert not torch.cuda.is_initialized(), "CUDA was initialized upon importing the `Accelerator` class."
+def function():
+    print(f"PartialState:\n{PartialState()}")
+
+
+if __name__ == "__main__":
+    notebook_launcher(function, num_processes=int(args.num_processes))
```
muellerzr, line 17: Want to make sure this looks fine to you @sgugger, I was able to run this without issue from the CLI with no cuda devices, one, and two

sgugger: Works for me!

src/accelerate/utils/imports.py
```diff
@@ -77,19 +77,28 @@ def is_fp8_available():
     return _is_package_available("transformer_engine")
 
 
+def is_cuda_available():
+    "Checks if `cuda` is available via an `nvml-based` check that won't trigger the drivers"
```
sgugger, line 81: And leave cuda uninitialized.

src/accelerate/utils/imports.py
```diff
@@ -77,19 +77,28 @@ def is_fp8_available():
...
+    os.environ["PYTORCH_NVML_BASED_CUDA_CHECK"] = str(1)
+    available = torch.cuda.is_available()
+    del os.environ["PYTORCH_NVML_BASED_CUDA_CHECK"]
```
sgugger, line 84: Let's use `try`/`finally` to make sure we cleanup the change in the environment in the (unlikely) case we get an error.

sgugger (commented):
Nice! Just left a couple of comments but it's way better this way.

src/accelerate/utils/imports.py
BenjaminBossan: Not sure why the use of `patch_environment` here was reverted, as it would take care of that. Also, why `str(1)` and not `"1"`?

muellerzr: Resulted in a cyclical import issue @BenjaminBossan

muellerzr:
Okay, ready for final review @sgugger. Changed the tactic slightly:

We can spawn a `PartialState` once in a dummy process to verify that CUDA can be initialized just fine, and then immediately end it. This will not cause issues with future calls as well, as it's its own process and won't lead to CUDA errors. This was the recommendation of the torch team to ensure that we can always verify this will work in the future.

src/accelerate/launchers.py
```diff
@@ -118,21 +123,28 @@ def train(*args):
...
-                    "function."
-                )
-
             # torch.distributed will expect a few environment variable to be here. We set the ones common to each
             # process here (the other ones will be set be the launcher).
             with patch_environment(
                 world_size=num_processes, master_addr="127.0.01", master_port=use_port, mixed_precision=mixed_precision
             ):
+                # First dummy launch
+                launcher = PrepareForLaunch(test_launch, distributed_type="MULTI_GPU")
```
sgugger, line 132: Let's only do this in debug mode.

muellerzr: Done, and added to the doc. Let me know if it looks alright 😄

src/accelerate/launchers.py
```diff
@@ -118,21 +128,29 @@ def train(*args):
...
-                    "function."
-                )
-
             # torch.distributed will expect a few environment variable to be here. We set the ones common to each
             # process here (the other ones will be set be the launcher).
             with patch_environment(
                 world_size=num_processes, master_addr="127.0.01", master_port=use_port, mixed_precision=mixed_precision
             ):
+                # First dummy launch
+                if debug or os.environ.get("ACCELERATE_DEBUG_MODE", "false").lower() == "true":
```
sgugger, line 137: Yes, just that here. Let's not add a new arg, no?

muellerzr: Done. I feel like this is common enough that a `debug` param directly would be better, but we can see how much use it gets (with the added docstring) during testing before fully commiting

sgugger (approved):
Thanks!

PR #1834, opened 2023-08-10
Ipex bug fix for device properties in modelling

abhilash1910:
For our arc intel gpus, there is an issue where if cpu is present in the device map (which it is by default from get_max_memory)  , it will flag an error in ipex (during device_map =balanced or auto) as ipex expects no 'cpu' type in its arc gpu memory initialization. This fix  is exclusively for our gpus .
@sgugger @muellerzr


abhilash1910:
@sgugger  can you check on this , this is an urgent fix for us . Thanks

sgugger (approved):
LGTM!

PR #1838, opened 2023-08-11
FIX: Bug with unwrap_model and keep_fp32_wrapper=False

BenjaminBossan:
Using `accelerator.unwrap_model(model, keep_fp32_wrapper=False)` results in a defective `forward` method. This bug was (probably) introduced in PR #872. I could reproduce it with current main, v0.21 and v0.20.

This PR provides a bugfix and tests for the regression.

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -15,6 +15,7 @@
 import os
 import socket
 from contextlib import contextmanager
+from types import MethodType
 
 import torch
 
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
Thanks! The `MethodType` logic was introduced afterwords, but wasn't propagated to here. Good find!

sgugger (approved):
Nice catch! Thanks for the fix!

PR #1842, opened 2023-08-13
Fix verify_device_map

Rexhaif:
Currently `verify_device_map`, which is called during `Accelerator.prepare` only checks if there is an attribute `hf_device_map` in any of the submodules. However, it does not check the contents of the device map itself. In case of trivial device map with only one element this will prevent user from training a model in distributed mode, while the model would in fact work perfectly fine as it is not spread across many devices.

I've made a small change to the function, so now it will **only** return True if any of the child modules has `hf_device_map` **and** this map has more than one entry.


sgugger (approved):
LGTM, thanks for your PR!
@SunMarc could you also have a look and merge if you're happy?

SunMarc (approved):
LGTM @Rexhaif  ! Please make the tests pass before I merge this PR.

Issue #1816, opened 2023-08-06
[Feature Request] support to pass custom slice function in accelerate data loader (useful for models like Flamingo)


thevasudevgupta:
@muellerzr & team, first of all, thanks for this great library!!

It would be great to be able to pass a custom slice function (instead of the using the default one):
https://github.com/huggingface/accelerate/blob/30eff605df5db9ba20289a0c7471e4d8547a5a8f/src/accelerate/data_loader.py#L596

use case: it would make it easy to train models like [flamingo](https://github.com/mlfoundations/open_flamingo) where text & images are interleaved. When text & image are interleaved, the data collator returns 2 tensors: `text` of size `(batch_size, seqlen)` & `images` of size `(num_images, H, W, C)`. Now we can't slice the image tensor using the original function in the library (because it slices based on the batch size). rather we want to slice the image tensor after finding the number of images required on each device. For eg: a batch of device-0 can get 2 images and a batch of device-1 can get 3 images.

Another workaround could also be to have slicing optional and let users control it after the same tensors are returned on all the devices.

Do you think accelerate library can support these kinda scenarios as well? I believe it will be super useful to train models like flamingo.

Let me know if my issue wasn't clear. Happy to discuss more. Thanks!

PR #1846, opened 2023-08-14
support custom slice function in DataLoaderDispatcher

thevasudevgupta:
@muellerzr I got a chance to work on this issue: #1816. Can you please review this PR? Happy to work on any suggestions.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1767,7 +1767,9 @@ class Accelerator:
                 result[i] = optimizer
         return tuple(result)
 
-    def prepare_data_loader(self, data_loader: torch.utils.data.DataLoader, device_placement=None, slice_fn=None):
+    def prepare_data_loader(
+        self, data_loader: torch.utils.data.DataLoader, device_placement=None, slice_fn_for_dispatch=None
+    ):
         """
         Prepares a PyTorch DataLoader for training in any distributed setup. It is recommended to use
         [`Accelerator.prepare`] instead.
@@ -1778,6 +1780,10 @@ class Accelerator:
             device_placement (`bool`, *optional*):
                 Whether or not to place the batches on the proper device in the prepared dataloader. Will default to
                 `self.device_placement`.
+            slice_fn_for_dispatch (`Callable`, *optional*`):
+                If passed, this function will be used to slice tensors across `num_processes`. Will default to
+                [`~utils.slice_tensors`]. This argument is used only when `dispatch_batches` is set to `True` and will
+                be ignored otherwise.
 
         Example:
 
@@ -1807,7 +1813,7 @@ class Accelerator:
             rng_types=self.rng_types.copy(),
             dispatch_batches=self.dispatch_batches,
             even_batches=self.even_batches,
-            slice_fn=slice_fn,
+            slice_fn_for_dispatch=slice_fn_for_dispatch,
         )
         self._dataloaders.append(prepared_data_loader)
         return prepared_data_loader
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -258,7 +258,6 @@ class Accelerator:
         step_scheduler_with_optimizer: bool = True,
         kwargs_handlers: list[KwargsHandler] | None = None,
         dynamo_backend: DynamoBackend | str | None = None,
-        slice_fn: Callable | None = None,
     ):
         if project_config is not None:
             self.project_configuration = project_config
@@ -462,8 +461,6 @@ class Accelerator:
         self.rng_types = rng_types
         if self.rng_types is None:
             self.rng_types = ["generator"]
-
-        self.slice_fn = slice_fn
 
     @property
     def use_distributed(self):
@@ -1770,7 +1767,7 @@ class Accelerator:
                 result[i] = optimizer
         return tuple(result)
 
-    def prepare_data_loader(self, data_loader: torch.utils.data.DataLoader, device_placement=None):
+    def prepare_data_loader(self, data_loader: torch.utils.data.DataLoader, device_placement=None, slice_fn=None):
         """
         Prepares a PyTorch DataLoader for training in any distributed setup. It is recommended to use
         [`Accelerator.prepare`] instead.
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -258,6 +258,7 @@ def __init__(
         step_scheduler_with_optimizer: bool = True,
         kwargs_handlers: list[KwargsHandler] | None = None,
         dynamo_backend: DynamoBackend | str | None = None,
+        slice_fn: Callable | None = None,
```
sgugger, line 261: No this can't be added as an argument of the main Accelerate class. This is way too specific a use case for an arg here. You can manually call the `prepare_data_loader` method for this use case.

src/accelerate/data_loader.py
```diff
@@ -503,6 +503,8 @@ def __init__(self, dataset, split_batches: bool = False, skip_batches=0, _drop_l
         self._drop_last = _drop_last
         self.skip_batches = skip_batches
 
+        self.slice_fn = slice_fn
```
sgugger, line 506: It would be better to have a default `slice_fn` that is set here when `slice_fn` is `None`.

src/accelerate/data_loader.py
```diff
@@ -567,7 +569,7 @@ def __iter__(self):
 
             if not self._drop_last and first_batch is None:
                 # We keep at least num processes elements of the first batch to be able to complete the last batch
-                first_batch = slice_tensors(batch, slice(0, self.state.num_processes))
+                first_batch = slice_tensors(batch, slice(0, self.state.num_processes)) if self.slice_fn is None else self.slice_fn(batch, self.state.process_index, self.state.num_processes)
```
sgugger, line 572: And here we can just call `self.slice_fn` all the time.

src/accelerate/data_loader.py
```diff
@@ -593,7 +595,7 @@ def __iter__(self):
                 batch_size += 1
 
             data_slice = slice(self.state.process_index * batch_size, (self.state.process_index + 1) * batch_size)
-            batch = slice_tensors(batch, data_slice)
+            batch = slice_tensors(batch, data_slice) if self.slice_fn is None else self.slice_fn(batch, self.state.process_index, self.state.num_processes)
```
sgugger, line 598: Same here.

src/accelerate/data_loader.py
```diff
@@ -633,6 +635,7 @@ def prepare_data_loader(
     rng_types: Optional[List[Union[str, RNGType]]] = None,
     dispatch_batches: Optional[bool] = None,
     even_batches: bool = True,
+    slice_fn: Optional[Callable] = None,
```
sgugger, line 638: The name should make it clear this is only for `dispatch` mode. This also needs to be properly documented.

sgugger (commented):
Thanks for your PR! I left a couple of comments.

src/accelerate/data_loader.py
thevasudevgupta: same comment as above: https://github.com/huggingface/accelerate/pull/1846#discussion_r1294093483

src/accelerate/data_loader.py
thevasudevgupta: shall I rename to `slice_fn_for_dispatch`? I will add it to the docstring.

src/accelerate/data_loader.py
sgugger: It takes `state.num_processes` and can just ignore `state.process_index`

src/accelerate/data_loader.py
sgugger: Yes, that's good for here! Thanks!

src/accelerate/data_loader.py
thevasudevgupta: following this: https://github.com/huggingface/accelerate/pull/1846#discussion_r1294226878, I am keeping it

src/accelerate/accelerator.py
```diff
@@ -1767,7 +1767,7 @@ def _prepare_ipex(self, *args):
                 result[i] = optimizer
         return tuple(result)
 
-    def prepare_data_loader(self, data_loader: torch.utils.data.DataLoader, device_placement=None):
+    def prepare_data_loader(self, data_loader: torch.utils.data.DataLoader, device_placement=None, slice_fn=None):
```
sgugger, line 1770: Here too, `slice_fn_for_dispatch` + your docstring would be great.

src/accelerate/data_loader.py
sgugger: That's not what I said. You can build a default slice function that ignores `state.process_index` is what I said ;-)

sgugger (commented):
Thanks for iterating!
Also cc @BenjaminBossan if you want to have a second look.

src/accelerate/data_loader.py
thevasudevgupta: So, it's not clear to me. Can you explain more on what is expected?

note: slice_tensors is used in 2 places inside `__iter__` and both places expect different arguments

```python
first_batch = (
    slice_tensors(batch, slice(0, self.state.num_processes))
    if self.slice_fn is None
    else self.slice_fn(batch, self.state.num_processes)
)
```

```python
batch = (
    slice_tensors(batch, data_slice)
    if self.slice_fn is None
    else self.slice_fn(batch, self.state.num_processes)
)
```

BenjaminBossan: I think the idea is the following: Change the signature of `slice_tensors` like this:

```python
def slice_tensors(data, tensor_slice, num_processes):
    # body is the same, ignores num_processes
```

Then you can just call `self.slice_fn(batch, data_slice, num_processes=self.state.num_processes)` without `if ... else`. Your custom slice function would ignore the `data_slice` argument if it doesn't need it.

thevasudevgupta: thanks, done.

sgugger (approved):
Thanks for iterating!

muellerzr (approved):
Thanks for the enhancement!

BenjaminBossan (approved):
LGTM, thanks for making the adjustments.

Issue #1848, opened 2023-08-16
the `gpu_ids` is not working


devymex:
My configurations:

```
compute_environment: LOCAL_MACHINE
distributed_type: MULTI_GPU
downcast_bf16: 'no'
machine_rank: 0
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: false
use_cpu: false
gpu_ids: 6,7
main_process_port: 29103
```

But `accelerate launch --config_file myconfig.yaml train.py` always use GPU 0 and 1. After some searching I found follwing bug:

https://github.com/huggingface/accelerate/blob/f67e11afd76f0dad3d4937f503fc1a2eccd5db5a/src/accelerate/utils/launch.py#L149

```
        if not is_xpu_available(): # <----------------- not???
            current_env["ZE_AFFINITY_MASK"] = gpu_ids
        elif is_npu_available():
            current_env["ASCEND_RT_VISIBLE_DEVICES"] = gpu_ids
        else:
            current_env["CUDA_VISIBLE_DEVICES"] = gpu_ids # this line is never be executed
```

should be

```
        if is_xpu_available(): # <----------------- remove the not
            current_env["ZE_AFFINITY_MASK"] = gpu_ids
        elif is_npu_available():
            current_env["ASCEND_RT_VISIBLE_DEVICES"] = gpu_ids
        else:
            current_env["CUDA_VISIBLE_DEVICES"] = gpu_ids
```

sgugger:
Yes, that sounds like the right fix. Would you like to open a PR?

muellerzr:
I think this can be closed now right @devymex? :)

PR #1850, opened 2023-08-16
Fix the noneffective parameter: gpu_ids (Rel. Issue #1848)

devymex:
[ the gpu_ids is not working #1848 ](https://github.com/huggingface/accelerate/issues/1848)


sgugger (approved):
Thanks for the fix!

PR #1852, opened 2023-08-16
Add env variable for init_on_device

shingjan:
Fixes [#25448](https://github.com/huggingface/transformers/issues/25448) from transformers

cc: @sgugger

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -39,6 +39,7 @@ from .utils import (
     is_torch_version,
     load_checkpoint_in_model,
     offload_state_dict,
+    parse_flag_from_env,
     retie_parameters,
 )
 
@@ -47,13 +48,13 @@ logger = logging.getLogger(__name__)
 
 
 @contextmanager
-def init_empty_weights(include_buffers: bool = False):
+def init_empty_weights(include_buffers: bool = None):
     """
     A context manager under which models are initialized with all parameters on the meta device, therefore creating an
     empty model. Useful when just initializing the model would blow the available RAM.
 
     Args:
-        include_buffers (`bool`, *optional*, defaults to `False`):
+        include_buffers (`bool`, *optional*):
             Whether or not to also put all buffers on the meta device while initializing.
 
     Example:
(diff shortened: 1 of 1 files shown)


src/accelerate/big_modeling.py
```diff
@@ -100,6 +101,7 @@ def init_on_device(device: torch.device, include_buffers: bool = False):
     ```
     """
     # TODO(shingjan): remove the torch version check once older versions are deprecated
+    include_buffers = parse_flag_from_env('INIT_INCLUDE_BUFFERS', False)
```
sgugger, line 104: Let's just set the default (at line 83) to `None` and then do this only if `include_buffers` is `None` (so the user can override the default from the env variable).

Also please leave the comment after this addition (before the torch version test).

sgugger (commented):
Thanks for your PR! Left a few comments.

src/accelerate/big_modeling.py
muellerzr: Let's also add `ACCELERATE_` as a prefix please so we can keep consistency with the rest of the library :)

shingjan: Thanks @sgugger and @muellerzr for the review! I also make the same change to default value of `include_buffers` for `init_empty_weights` to make the env variable useful and keep BC. Thanks!

src/accelerate/big_modeling.py
```diff
@@ -47,13 +48,13 @@
...
 @contextmanager
-def init_empty_weights(include_buffers: bool = False):
+def init_empty_weights(include_buffers: bool = None):
     """
     A context manager under which models are initialized with all parameters on the meta device, therefore creating an
     empty model. Useful when just initializing the model would blow the available RAM.
 
     Args:
-        include_buffers (`bool`, *optional*, defaults to `False`):
+        include_buffers (`bool`, *optional*, defaults to `None`):
```
sgugger, line 57:
```suggestion
        include_buffers (`bool`, *optional*):
```
Small nit: we don't put the default when it's None as it's implied by `*optional*`

src/accelerate/big_modeling.py
```diff
@@ -74,19 +75,21 @@ def init_empty_weights(include_buffers: bool = False):
...
-def init_on_device(device: torch.device, include_buffers: bool = False):
+def init_on_device(device: torch.device, include_buffers: bool = None):
     """
     A context manager under which models are initialized with all parameters on the specified device.
 
     Args:
         device (`torch.device`):
             Device to initialize all parameters on.
-        include_buffers (`bool`, *optional*, defaults to `False`):
+        include_buffers (`bool`, *optional*, defaults to `None`):
```
sgugger, line 92:
```suggestion
        include_buffers (`bool`, *optional*):
```

sgugger (approved):
Thanks for iterating! Two last small suggestions and we should be ready to go.

shingjan:
thanks @sgugger and @muellerzr for getting this in!

PR #1853, opened 2023-08-17
Loading logic safetensors

SunMarc:
# What does this PR do ? 
This PR adds the possibility to load the safetensors checkpoint with just passing the folder containing the checkpoint in `load_checkpoint_in_model`.


sgugger (approved):
Thanks for adding this!

PR #1855, opened 2023-08-17
fix dispatch for quantized model

SunMarc:
# What does this do ? 
This PR fixes the `dispatch` function for quantized model. Only bnb model needs to have hooks in a single gpu setup. For other quantization methods such as GPTQ, we don't need that.


younesbelkada (approved):
Makes sense, thanks Marc!

fxmarty (approved):

PR #1874, opened 2023-08-22
support logging with mlflow in case of mlflow-skinny installed

ghtaro:
This is PR following the below issue found in trl.
https://github.com/huggingface/trl/issues/658#issuecomment-1687284028

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -210,18 +210,15 @@ def is_tqdm_available():
 
 
 def is_mlflow_available():
-    package_exists = importlib.util.find_spec("mlflow") is not None
-    if package_exists:
+    if _is_package_available("mlflow"):
+        return True
+
+    if importlib.util.find_spec("mlflow") is not None:
         try:
-            _ = importlib.metadata.metadata("mlflow")
+            _ = importlib.metadata.metadata("mlflow-skinny")
             return True
         except importlib.metadata.PackageNotFoundError:
-            # support a case of mlflow installed via mlflow-skinny
-            try:
-                _ = importlib.metadata.metadata("mlflow-skinny")
-                return True
-            except importlib.metadata.PackageNotFoundError:
-                return False
+            return False
     return False
 
 
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -210,7 +210,19 @@ def is_tqdm_available():
 
 
 def is_mlflow_available():
-    return _is_package_available("mlflow")
+    package_exists = importlib.util.find_spec("mlflow") is not None
+    if package_exists:
+        try:
+            _ = importlib.metadata.metadata("mlflow")
+            return True
+        except importlib.metadata.PackageNotFoundError:
+            # support a case of mlflow installed via mlflow-skinny
+            try:
+                _ = importlib.metadata.metadata("mlflow-skinny")
+                return True
+            except importlib.metadata.PackageNotFoundError:
+                return False
+    return False
 
 
 def is_mps_available():


sgugger:
cc @muellerzr

src/accelerate/utils/imports.py
```diff
@@ -210,7 +210,19 @@ def is_tqdm_available():
 
 
 def is_mlflow_available():
-    return _is_package_available("mlflow")
+    package_exists = importlib.util.find_spec("mlflow") is not None
```
BenjaminBossan, line 213: I haven't tried it out, but could the code not be simplified like this:

```python
def is_mlflow_available():
    # <explanation + reference to issue>
    return _is_package_available("mlflow") or _is_package_available("mlflow-skinny")
```

muellerzr: Yes, this is indeed what we should do here.

src/accelerate/utils/imports.py
```diff
@@ -210,7 +210,19 @@ def is_tqdm_available():
...
+            _ = importlib.metadata.metadata("mlflow")
+            return True
+        except importlib.metadata.PackageNotFoundError:
+            # support a case of mlflow installed via mlflow-skinny
+            try:
+                _ = importlib.metadata.metadata("mlflow-skinny")
+                return True
+            except importlib.metadata.PackageNotFoundError:
+                return False
+    return False
```
muellerzr, line 225:
```suggestion
    return _is_package_available("mlflow") or _is_package_available("mlflow-skinny")
```

muellerzr (commented):
As with @BenjaminBossan's comment, let's simplify this to just use `_is_package_available` and check both `mlflow` and `mlflow-skinny`

ghtaro:
@BenjaminBossan @muellerzr Thank you for your comment.

I found the below behaviour and did not come up with any simple code...
image

src/accelerate/utils/imports.py
ghtaro: @BenjaminBossan @muellerzr Thank you for your comments.

I found the below behaviour.



I did not come with simple code.

BenjaminBossan: I see. Would this work:

```python
def is_mlflow_available():
    if _is_package_available("mlflow"):
        return True

    if importlib.util.find_spec("mlflow") is not None:
        try:
            _ = importlib.metadata.metadata("mlflow-skinny")
            return True
        except importlib.metadata.PackageNotFoundError:
            return False
    return False
```

BenjaminBossan:
@ghtaro Implementation-wise, this looks good.

Just to confirm: mlflow-skinny has the same API as standard mlflow, so logging through [`MLflowTracker`](https://github.com/huggingface/accelerate/blob/812719644d15576b2279164dd5f3cb020595d047/src/accelerate/tracking.py#L534) would just work? Ideally, we would have a test for this but this would create significant extra work for what seems to me a relatively niche use case, so your personal confirmation would be enough for me.

ghtaro:
@BenjaminBossan To be honest, I do not know if the fix works for all the cases where they use mlflow-skinny, but at least I have checked the below case.

- databricks, runtime 13.0ML (https://docs.databricks.com/en/release-notes/runtime/13.0ml.html) and 13.3LTS ML (mlflow-skinny preinstalled for both).
- run modified version of https://github.com/huggingface/trl/blob/main/examples/research_projects/stack_llama/scripts/rl_training.py

I usually use databricks with mlflow tracking, so the fix is fine with me.
I would like to wait for a reply in https://github.com/huggingface/accelerate/issues/1879.

BenjaminBossan:
I see, thanks for the update.

> I would like to wait for a reply in #1879.

Okay, let's wait for a reply. Otherwise, this LGTM.

muellerzr (approved):
Thanks for the PR! I think even though we're waiting this solution is sound to me, and we can go ahead and merge. WDYT @BenjaminBossan

PR #1875, opened 2023-08-22
Add test decorators for npu

statelesshz:
### What does this PR do?
This PR includes essential test decorators for npu, allowing us to reuse already existing test cases with slight adjustments.


sgugger:
Would be nice to introduce them when they are actually used in the code base though. Do you have test cases that need them?

PR #1876, opened 2023-08-22
Introduce model memory estimator

muellerzr:
# `accelerate estimate-memory`

## What does this add?

This PR introduces a new CLI command: `accelerate estimate-memory`, which will utilize the `meta` device to calculate the total vRAM needed for big model inference, the entire model, and training for a particular model and a variety of precision formats. This also does so *without needing to download model weights*.

## Who is it for?

Users of accelerate, and MLE's in general who want to know how big of a model they can fit on their system

## Why is it needed?

Many individuals have asked for a way to figure out just what size of a model they can perform big model inference on, the size of a model to fit on their system in general, and for training. As a result, this PR finally fulfills this, using much of the memory-calculating logic already existing in accelerate

Currently this supports `transformers` and `timm`, in the future it will support `diffusers`

### User-facing:

A new `accelerate estimate-memory` command

### Internal structure:

A few utils were added for convenience to `big_modeling` and `other` in relation to this PR

## Basic Usage Example(s):

All from the CLI:

Using `transformers`:
```bash
accelerate estimate-memory  togethercomputer/LLaMA-2-7B-32K --dtypes float32 float16
Loading pretrained config for `togethercomputer/LLaMA-2-7B-32K` from `transformers`...
┌────────────────────────────────────────────────────┐
│ Memory Usage for `togethercomputer/LLaMA-2-7B-32K` │
├───────┬─────────────┬──────────┬───────────────────┤
│ dtype │Largest Layer│Total Size│Training using Adam│
├───────┼─────────────┼──────────┼───────────────────┤
│float32│   500.0 MB  │ 25.61 GB │     102.46 GB     │
│float16│   250.0 MB  │ 12.81 GB │      51.23 GB     │
└───────┴─────────────┴──────────┴───────────────────┘
```

Using `timm`:
```bash
accelerate estimate-memory timm/resnet50.a1_in1k --dtypes float32 float16
Loading pretrained config for `timm/resnet50.a1_in1k` from `timm`...
┌────────────────────────────────────────────────────┐
│      Memory Usage for `timm/resnet50.a1_in1k`      │
├───────┬─────────────┬──────────┬───────────────────┤
│ dtype │Largest Layer│Total Size│Training using Adam│
├───────┼─────────────┼──────────┼───────────────────┤
│float32│    9.0 MB   │ 97.7 MB  │     390.78 MB     │
│float16│    4.5 MB   │ 48.85 MB │     195.39 MB     │
└───────┴─────────────┴──────────┴───────────────────┘
```

A user would want to use this feature if they wanted to know if a model will fit on their card or be inferenceable with big model inference before downloading the weights

### DIFF
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -239,7 +239,7 @@ Estimates the total vRAM a particular model hosted on the Hub needs to perform B
 **Usage**: 
 
 ```bash
-accelerate estimate-memory --model_name {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
+accelerate estimate-memory {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
 ```
 
 **Required Arguments**:
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -234,19 +234,23 @@ The following arguments are only useful when training in SageMaker
 
 `accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
 
-Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training.
+Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training. Requires that `huggingface_hub` be installed.
 
 **Usage**: 
 
 ```bash
-accelerate estimate-memory [arguments]
+accelerate estimate-memory --model_name {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
 ```
+
+**Required Arguments**:
+
+* `MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
 
 **Optional Arguments**:
 
-* `--model_name MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
-* `--library_name LIBRARY_NAME` (`str`)-- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
+* `--library_name {timm,transformers}` (`str`)-- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
 * `--dtypes {float32,float16,int8,int4}` (`[{float32,float16,int8,int4} ...]`) --The dtypes to use for the model, must be one (or many) of `float32`, `float16`, `int8`, and `int4`
+* `--trust_remote_code` (`bool`) -- Whether or not to allow for custom models defined on the Hub in their own modeling files. This option should only be passed for repositories you trust and in which you have read the code, as it will execute code present on the Hub on your local machine.
 
 ## accelerate tpu-config
 
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -228,6 +228,26 @@ The following arguments are only useful when training in SageMaker
 * `--aws_access_key_id AWS_ACCESS_KEY_ID` (`str`) -- The AWS_ACCESS_KEY_ID used to launch the Amazon SageMaker training job
 * `--aws_secret_access_key AWS_SECRET_ACCESS_KEY` (`str`) -- The AWS_SECRET_ACCESS_KEY used to launch the Amazon SageMaker training job
 
+## accelerate estimate-memory
+
+**Command**:
+
+`accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
+
+Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training.
+
+**Usage**: 
+
+```bash
+accelerate estimate-memory [arguments]
+```
+
+**Optional Arguments**:
+
+* `--model_name MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
+* `--library_name LIBRARY_NAME` (`str`)-- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
+* `--dtypes {float32,float16,int8,int4}` (`[{float32,float16,int8,int4} ...]`) --The dtypes to use for the model, must be one (or many) of `float32`, `float16`, `int8`, and `int4`
+
 ## accelerate tpu-config
 
 `accelerate tpu-config`
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -234,7 +234,7 @@ The following arguments are only useful when training in SageMaker
 
 `accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
 
-Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training. Requires that `huggingface_hub` be installed.
+Estimates the total vRAM a particular model hosted on the Hub needs to perform inference and training. Requires that `huggingface_hub` be installed.
 
 **Usage**: 
 
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -248,8 +248,8 @@ accelerate estimate-memory {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {
 
 **Optional Arguments**:
 
-* `--library_name {timm,transformers}` (`str`)-- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
-* `--dtypes {float32,float16,int8,int4}` (`[{float32,float16,int8,int4} ...]`) --The dtypes to use for the model, must be one (or many) of `float32`, `float16`, `int8`, and `int4`
+* `--library_name {timm,transformers}` (`str`) -- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
+* `--dtypes {float32,float16,int8,int4}` (`[{float32,float16,int8,int4} ...]`) -- The dtypes to use for the model, must be one (or many) of `float32`, `float16`, `int8`, and `int4`
 * `--trust_remote_code` (`bool`) -- Whether or not to allow for custom models defined on the Hub in their own modeling files. This option should only be passed for repositories you trust and in which you have read the code, as it will execute code present on the Hub on your local machine.
 
 ## accelerate tpu-config
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -234,7 +234,13 @@ The following arguments are only useful when training in SageMaker
 
 `accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
 
-Estimates the total vRAM a particular model hosted on the Hub needs to perform inference and training. Requires that `huggingface_hub` be installed.
+Estimates the total vRAM a particular model hosted on the Hub needs to be loaded in with an estimate for training. Requires that `huggingface_hub` be installed. 
+
+<Tip>
+
+    When performing inference, typically add ≤20% to the result as overall allocation [as referenced here](https://blog.eleuther.ai/transformer-math/). We will have more extensive estimations in the future that will automatically be included in the calculation.
+
+</Tip>
 
 **Usage**: 
 
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ extras = {}
 extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0", "urllib3 < 2.0.0"]
 extras["docs"] = []
 extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests", "parameterized"]
-extras["test_dev"] = ["datasets", "evaluate", "transformers", "scipy", "scikit-learn", "deepspeed", "tqdm", "bitsandbytes", "timm", "torchvision"]
+extras["test_dev"] = ["datasets", "evaluate", "transformers", "scipy", "scikit-learn", "deepspeed", "tqdm", "bitsandbytes", "timm", "torchvision", "einops"]
 extras["testing"] = extras["test_prod"] + extras["test_dev"]
 extras["rich"] = ["rich"]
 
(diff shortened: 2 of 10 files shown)


src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,206 @@
...
+    calculate_maximum_sizes,
+    convert_bytes,
+    is_huggingface_hub_available,
+    is_timm_available,
+    is_transformers_available,
+)
+
+
+if is_huggingface_hub_available():
+    from huggingface_hub import HfApi
```
sgugger, line 31: Let's just import `model_info` since you use `HfApi` without changing the defaults.

docs/source/package_reference/cli.md
```diff
@@ -228,6 +228,26 @@ The following arguments are only useful when training in SageMaker
...
+
+**Usage**: 
+
+```bash
+accelerate estimate-memory [arguments]
+```
+
+**Optional Arguments**:
+
+* `--model_name MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
```
sgugger, line 247: This is not optional at all, not passing this will result in an error. I think it's more natural to do `accelerate estimate-memory model_name` then have to type the `--model_name` on top of that.

docs/source/package_reference/cli.md
```diff
@@ -228,6 +228,26 @@ The following arguments are only useful when training in SageMaker
...
+**Command**:
+
+`accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
+
+Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training.
+
+**Usage**: 
+
+```bash
+accelerate estimate-memory [arguments]
```
sgugger, line 242: I'd show a real example here.

src/accelerate/utils/modeling.py
```diff
@@ -856,6 +856,19 @@ def get_balanced_memory(
...
 
+def calculate_maximum_sizes(model: torch.nn.Module):
+    "Computes the total size of the model and its largest layer"
+    sizes = compute_module_sizes(model)
+    modules_to_treat = (
+        list(model.named_parameters(recurse=False))
+        + list(model.named_children())
+        + list(model.named_buffers(recurse=False))
+    )
+    largest_layer = get_max_layer_size(modules_to_treat, sizes, [])
```
sgugger, line 867: The `no_split_module_classes` is what will determine the size of the largest layer very often, so this info should be passed along if we have it (from a Transformers model for instance).

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,206 @@
...
+    parser.add_argument(
+        "--library_name",
+        type=str,
+        help="The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub.",
+    )
+    parser.add_argument(
+        "--dtypes",
+        type=str,
+        nargs="+",
+        default=["float32"],
```
sgugger, line 151: Default should be the full list I think.

sgugger (commented):
Thanks for working on this!

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,206 @@
...
+            model = timm.create_model(model_name, pretrained=False)
+    else:
+        raise ValueError(
+            f"Library `{library_name}` is not supported yet, please open an issue on GitHub for us to add support."
+        )
+    return model
+
+
+def create_ascii_table(headers: list, rows: list, title: str):
+    "Creates a pretty table from a list of rows, minimal version of `tabulate`."
```
BenjaminBossan, line 107: I would have just used tabulate. It has extremely minimal requirements and has never once caused issues in years of using it.

BenjaminBossan (commented):
Super useful feature, thanks for adding it.

Besides fixing the failing test and what Sylvain mentioned, I only have minor comments, please take a look.

One minor concern that I have is that the tests re-use a lot of the same code as the functions that they're testing. Thus certain types of bugs would not be caught (say, a bug in `convert_bytes`). I wonder if we can add a test for a given model where we know the values for the vRAM usage and can just hard code the expected values. WDYT?

src/accelerate/commands/estimate.py
muellerzr: While it is, we tend to avoid adding any dependencies to accelerate whatsoever, and here it was a small enough code chunk to work fine. E.g. see the CLI menu selector

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,206 @@
...
+if is_transformers_available():
+    from transformers import AutoConfig, AutoModel
+
+if is_timm_available():
+    import timm
+
+
+def verify_on_hub(repo: str):
+    "Verifies that the model is on the hub and returns the model info."
+    api = HfApi()
```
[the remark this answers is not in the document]
muellerzr: Yes, I'm missing a check for if the hf hub is available here, I'll include it. Nice catch

docs/source/package_reference/cli.md
```diff
@@ -228,6 +228,26 @@ The following arguments are only useful when training in SageMaker
...
+```
+
+**Optional Arguments**:
+
+* `--model_name MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
+* `--library_name LIBRARY_NAME` (`str`)-- The library the model has an integration with, such as `transformers`, needed only if this information is not stored on the Hub
```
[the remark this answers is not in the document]
muellerzr: I'll definitely include it then :) (And diffusers will be a later PR)

muellerzr:
@BenjaminBossan re;

> One minor concern that I have is that the tests re-use a lot of the same code as the functions that they're testing. Thus certain types of bugs would not be caught (say, a bug in convert_bytes). I wonder if we can add a test for a given model where we know the values for the vRAM usage and can just hard code the expected values. WDYT?

I'll look at refactoring the code, as yes if we have to go that far to test with utils it smells of a design issue

docs/source/package_reference/cli.md
muellerzr: Fixed

docs/source/package_reference/cli.md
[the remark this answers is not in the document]
muellerzr: This is now included both here and when doing `--help`

muellerzr:
Also looking at why the tests are failing on the CI, as they pass locally for me so this is quite strange (and in a brand new testing env)

docs/source/package_reference/cli.md
```diff
@@ -234,19 +234,23 @@ The following arguments are only useful when training in SageMaker
...
 `accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
 
-Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training.
+Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training. Requires that `huggingface_hub` be installed.
 
 **Usage**: 
 
 ```bash
-accelerate estimate-memory [arguments]
+accelerate estimate-memory --model_name {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
```
sgugger:
```suggestion
accelerate estimate-memory {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
```

src/accelerate/utils/modeling.py
muellerzr: Done.

src/accelerate/utils/other.py
```diff
@@ -222,3 +222,13 @@ def is_port_in_use(port: int = None) -> bool:
...
+
+
+def convert_bytes(size):
+    "Converts `size` from bytes to the largest possible unit"
+    for x in ["bytes", "KB", "MB", "GB", "TB"]:
+        if size < 1024.0:
+            return f"{round(size, 2)} {x}"
+        size /= 1024.0
+
+    return size
```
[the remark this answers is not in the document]
muellerzr: That'll be some big models 😉

docs/source/package_reference/cli.md
muellerzr: Resolved in a latter commit :) )

BenjaminBossan:
> Also looking at why the tests are failing on the CI, as they pass locally for me so this is quite strange (and in a brand new testing env)

I think they're now failing because they still use `--model_name`, but IIRC there was another error before that.

docs/source/package_reference/cli.md
```diff
@@ -228,6 +228,30 @@ The following arguments are only useful when training in SageMaker
 * `--aws_access_key_id AWS_ACCESS_KEY_ID` (`str`) -- The AWS_ACCESS_KEY_ID used to launch the Amazon SageMaker training job
 * `--aws_secret_access_key AWS_SECRET_ACCESS_KEY` (`str`) -- The AWS_SECRET_ACCESS_KEY used to launch the Amazon SageMaker training job
 
+## accelerate estimate-memory
+
+**Command**:
+
+`accelerate estimate-memory` or `accelerate-estimate-memory` or `python -m accelerate.commands.estimate`
+
+Estimates the total vRAM a particular model hosted on the Hub needs to perform Big Model Inference and training. Requires that `huggingface_hub` be installed.
```
sgugger, line 237:
```suggestion
Estimates the total vRAM a particular model hosted on the Hub needs to perform inference and training. Requires that `huggingface_hub` be installed.
```

sgugger: Also, maybe we should add `huggingface_hub` as a core dep of Accelerate? It's meant to be the place of low-level utils shared across the ecosystem.

docs/source/package_reference/cli.md
```diff
@@ -228,6 +228,30 @@ The following arguments are only useful when training in SageMaker
...
+
+**Usage**: 
+
+```bash
+accelerate estimate-memory {MODEL_NAME} --library_name {LIBRARY_NAME} --dtypes {dtype_1} {dtype_2} ...
+```
+
+**Required Arguments**:
+
+* `MODEL_NAME` (`str`)-- The model name on the Hugging Face Hub
```
sgugger, line 247: Can also be the path to a local model since you use auto classes, I think.

setup.py
```diff
@@ -19,7 +19,7 @@
 extras["quality"] = ["black ~= 23.1", "ruff >= 0.0.241", "hf-doc-builder >= 0.3.0", "urllib3 < 2.0.0"]
 extras["docs"] = []
 extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests", "parameterized"]
-extras["test_dev"] = ["datasets", "evaluate", "transformers", "scipy", "scikit-learn", "deepspeed", "tqdm", "bitsandbytes"]
+extras["test_dev"] = ["datasets", "evaluate", "transformers", "scipy", "scikit-learn", "deepspeed", "tqdm", "bitsandbytes", "timm", "torchvision", "einops"]
```
sgugger, line 22: Not sure why einops is coming? Or torchvision?

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,270 @@
...
+        is_transformers_available()
+        and isinstance(error, OSError)
+        and "does not appear to have a file named" in error.args[0]
+    ):
+        return "transformers"
+    else:
+        return "unknown"
+
+
+def create_empty_model(model_name, library_name: str, trust_remote_code: bool = False):
```
sgugger, line 70: If doing type annotations, do them all ;-)
This misses the return type and the type of `model_name`.

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,270 @@
...
+        trust_remote_code (`bool`, `optional`, defaults to `False`):
+            Whether or not to allow for custom models defined on the Hub in their own modeling files. This option
+            should only be set to `True` for repositories you trust and in which you have read the code, as it will
+            execute code present on the Hub on your local machine.
+
+    """
+    model_info = verify_on_hub(model_name)
+    # Simplified errors
+    if model_info == "gated":
+        raise GatedRepoError(f"Repo for model `{model_name}` is gated. You must be authenticated to access it.")
```
sgugger, line 89: Tell the user how to be authenticated (probably running `huggingface-cli login` or something like that)

src/accelerate/commands/estimate.py
```diff
@@ -0,0 +1,270 @@
...
+            config = AutoConfig.from_pretrained(model_name, trust_remote_code=trust_remote_code)
+        except ValueError as e:
+            if "trust_remote_code" in e.args[0]:
+                e.args = (
+                    f"Loading {model_name} requires to execute some code in that repo, you can inspect the content of "
+                    f"the repository at https://hf.co/{model_name}. You can bypass this by passing "
+                    "`--trust_remote_code`",
+                )
+            raise e
+        with init_empty_weights():
```
sgugger, line 125: I'd let the error be raised by Transformers and just do the config instantiation here.

tests/test_cli.py
```diff
@@ -211,3 +223,178 @@ def test_accelerate_install_version(self):
...
+            args = self.parser.parse_args(["muellerzr/dummy"])
+            estimate_command(args)
+
+    def test_gated(self):
+        with self.assertRaises(GatedRepoError, msg="Repo for model `meta-llama/Llama-2-7b` is gated"):
+            args = self.parser.parse_args(["meta-llama/Llama-2-7b"])
+            with patch_environment(hf_hub_disable_implicit_token="1"):
+                estimate_command(args)
+
+    @require_einops
```
sgugger, line 271: Can we use a model that does not require einops for this? Also this model will soon not require trust_remote_code so it's not a good candidate for this test.

sgugger (commented):
Thanks for iterating on this. I added a couple more comments.

docs/source/package_reference/cli.md
muellerzr: In order not to mess with too much in a single PR, I'll add that in as a follow up

tests/test_cli.py
muellerzr: Using `test_dynamic_model`, which also let's us remove the auto test since that's what it will do automatically.

docs/source/package_reference/cli.md
muellerzr: There's a bit more logic involved with this here, I'll do it as a follow-up PR

sgugger (approved):
Thanks for iterating!

BenjaminBossan (approved):
Thanks a lot for the PR, super useful and well tested and implemented.

SunMarc (approved):
LGTM ! Thanks for the extensive testing suite !

PR #1877, opened 2023-08-23
Update release instructions

sgugger:
This updates the release instructions to the latest (just followed those for v0.22.0 this morning).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,21 +67,27 @@ setup(
 )
 
 # Release checklist
-# 1. Change the version in __init__.py and setup.py.
-# 2. Commit these changes with the message: "Release: VERSION"
-# 3. Add a tag in git to mark the release: "git tag VERSION -m 'Adds tag VERSION for pypi' "
-#    Push the tag to git: git push --tags origin main
-# 4. Run the following commands in the top-level directory:
+# 1. Checkout the release branch (for a patch the current release branch, for a new minor version, create one):
+#      git checkout -b vXX.xx-release
+#    The -b is only necessary for creation (so remove it when doing a patch)
+# 2. Change the version in __init__.py and setup.py to the proper value.
+# 3. Commit these changes with the message: "Release: v<VERSION>"
+# 4. Add a tag in git to mark the release:
+#      git tag v<VERSION> -m 'Adds tag v<VERSION> for pypi'
+#    Push the tag and release commit to git: git push --tags origin vXX.xx-release
+# 5. Run the following commands in the top-level directory:
+#      rm -rf dist
+#      rm -rf build
 #      python setup.py bdist_wheel
 #      python setup.py sdist
-# 5. Upload the package to the pypi test server first:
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
Thank you @sgugger for updating the release instructions, LGTM!

muellerzr (approved):
Thanks! As @BenjaminBossan I think we should definitely include a section on depreciation so it's easier to find them (as there's a few across the lib that haven't been addressed in a bit). I can make a PR today converting them all to TODO :)

sgugger:
@BenjaminBossan The instructions are mostly for @LysandreJik who will do the releases from now on. It's not necessarily super comprehensive (for instance we only work from the main fork, so the comment related to that does not apply) and I mostly want to make sure he is okay with the comments before adding :-)

setup.py
```diff
@@ -67,21 +67,27 @@
...
+#      rm -rf dist
+#      rm -rf build
 #      python setup.py bdist_wheel
 #      python setup.py sdist
-# 5. Upload the package to the pypi test server first:
-#      twine upload dist/* -r pypitest
-#      twine upload dist/* -r pypitest --repository-url=https://test.pypi.org/legacy/
-# 6. Check that you can install it in a virtualenv by running:
+# 6. Upload the package to the pypi test server first:
+#      twine upload dist/* -r testpypi
```
LysandreJik, line 84: For `transformers` we're still using the `legacy` repository as we don't have a `pypitest` section in the `.pypirc`. 

It means generally using the command you erased:
```
#      twine upload dist/* -r pypitest --repository-url=https://test.pypi.org/legacy/
```

Have you managed to push to pypitest without a legacy URL and without a `.pypirc`?

setup.py
```diff
@@ -67,21 +67,27 @@
...
+# 7. Check that you can install it in a virtualenv by running:
 #      pip install -i https://testpypi.python.org/pypi accelerate
```
LysandreJik, line 86: For clean environments, may be worth putting:
```suggestion
#      pip install accelerate
#      pip uninstall accelerate
#      pip install -i https://testpypi.python.org/pypi accelerate
```

to ensure all deps can be installed first

LysandreJik (approved):
Looks good! Very similar to `transformers`. Thanks @sgugger.

setup.py
LysandreJik: Understood! I'll just try it with this one again next time and if it doesn't work I'll add the line that works for me once again.

BenjaminBossan (approved):
> The instructions are mostly for LysandreJik who will do the releases from now on. It's not necessarily super comprehensive (for instance we only work from the main fork, so the comment related to that does not apply) and I mostly want to make sure he is okay with the comments before adding :-)

Got it thx. LGTM

setup.py
muellerzr: I'd probably do:

> pip uninstall accelerate -y

So we can bypass the prompt? :)

PR #1880, opened 2023-08-23
fix detach_hook

### LINKED ISSUE #1869: 【BUG Report】accelerate method remove_hook_from_module can't remove cpu_offload hook.
### System Info

First I wanna cpu offloaded model back to cpu.
so i first call the method `accelerate.cpu_offload(module, CUDA_0)` to offload module to cuda.
then I want my module back to cpu, as i called `remove_hook_from_module(module, recurse=True)`

and remove method tells me NoneType error.

if tensor_name not in module._parameters and tensor_name not in module._buffers:
AttributeError: 'NoneType' object has no attribute '_parameters'

so i check the code, and found that.

image
```Shell
the hook detach method should return changed module. Also I see the Base Hook definition `ModelHook`. 
````
image

so the only thing is, class `AlignDevicesHook` not support it well for not return module out.

image

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

1. call `accelerate.cpu_offload(module, CUDA_0)`  to offload some module to cuda
2. call `remove_hook_from_module(module, recurse=True)` to remove offload hook.

### Expected behavior

detach  hook success.

SunMarc:
# What does this PR do ? 
Fixes #1869. This PR fixes the `detach_hook` function from `AlignDevicesHook` class so that we can call `.detach_hook()` on a `SequentialHook` object where one of the hooks is an `AlignDevicesHook` object.

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -311,6 +311,7 @@ class AlignDevicesHook(ModelHook):
             for name, device in self.original_devices.items():
                 if device != torch.device("meta"):
                     set_module_tensor_to_device(module, name, device, value=self.weights_map.get(name, None))
+        return module
 
 
 def attach_execution_device_hook(


sgugger (approved):
Thanks for fixing!

PR #1884, opened 2023-08-24
Update with new url for explore

muellerzr:
Now that accelerate has an official hub org, updates the docs to include it.

Note: the old URL will always redirect to the new one, so prior doc versions will not break


SunMarc (approved):
Thanks for adding this !

PR #1885, opened 2023-08-24
Add hub as core dep

muellerzr:
As the hub is meant to be the foundational library for every lib in the hf ecosystem, adds `huggingface_hub` as a core dep on Accelerate since we have need for it now with the model estimation tool.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ setup(
         ]
     },
     python_requires=">=3.8.0",
-    install_requires=["numpy>=1.17", "packaging>=20.0", "psutil", "pyyaml", "torch>=1.10.0"],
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
To be clear, is the argument for adding Hub as a dependency that users will probably already have it installed for the other packages they use? I just want to understand the reasoning. Apart from that, Hub doesn't seem to come with many extra deps (with just the base install), so I think it can be safely added.

muellerzr:
@BenjaminBossan the hub is designed to be the foundational library in general for all huggingface packages essentially, this includes it finally as we have a reason for it to exist. 

See this note by @sgugger: https://github.com/huggingface/accelerate/pull/1876#discussion_r1303382436

BenjaminBossan (approved):
I see, makes sense, thanks for explaining. LGTM.

PR #1895, opened 2023-08-28
improve help info when run `accelerate config` on npu

statelesshz:
### What does this PR do?
as title, this PR make the prompt information more reasonable when running `accelerate config` on NPU.


muellerzr (approved):
Thanks! This makes sense and seems quite helpful :)

PR #1896, opened 2023-08-28
Final nits on model util

muellerzr:
Just some final nits and cleaning on the model util and doc, specifying that after testing the real vs estimated values, we're off by at most a few % (further tests to be done on this), showing `bert-base-cased` as an example of this.


SunMarc (approved):
LGTM !

PR #1894, opened 2023-08-28
update import of deepspeed integration from transformers

pacman100:
### What does this PR do?
1. To be in sync with changes in PR https://github.com/huggingface/transformers/pull/25599

PR #1898, opened 2023-08-28
Solve at least one failing test

muellerzr:
@pacman100 we have multiple failing tests, stemming from https://github.com/huggingface/accelerate/pull/1894 + related, especially when working off `pypi`. should we skip these tests, if (one) of the fixes in this PR is deemed unnecessary? We also fail on integration tests with pypi as well: https://github.com/huggingface/accelerate/actions/runs/6000691481/job/16273604914


pacman100 (approved):
Thank you Zach for the quick fix, LGTM.

PR #1899, opened 2023-08-28
Fix nb launcher test

muellerzr:
Skips the nb launcher multi-gpu test, until a better solution can be found. The problem arises from the fact that when doing `make test`, somewhere along the way a forking issue will rise, causing CUDA to not like the fact we're spawning another process it can't handle (my best guess). If ran alone this test will pass (or with just `pytest -sv tests/test_multigpu.py`)


SunMarc (approved):
LGTM !

Issue #1830, opened 2023-08-10
Issues with DeepSpeed autotuning and `gradient_accumulation_steps`


acforvs:
Hey, I’m curious if there’s a conventional way to use [deepspeed autotuning ](http://deepspeed.ai/tutorials/autotuning/) with `accelerate launch`? DeepSpeed’s autotuning is useful to improve the throughput of training and inference setups, it would be great for this to be compatible with Accelerate.

I have been trying to set up autotuning in many ways, but none of it seems to be working. Below are my somewhat long thoughts on what seems to be the problem.

#### 1. Use `accelerate launch` with multiple `zero_optimization.stage` values

Using `accelerate launch --config_file accelerate.yaml` with deepspeed_config.deepspeed_config_file section pointing to the file with the following content

```json
{
  "zero_optimization": {
      "stage": [2, 3],
      "offload_optimizer": {
          "device": "nvme",
          "nvme_path": "/mnt/vlad/param_offload"
      },
      "offload_param": {
          "device": "nvme",
          "nvme_path": "/mnt/vlad/param_offload"
      },
      "overlap_comm": true,
      "contiguous_gradients": true,
      "reduce_bucket_size": "auto",
      "stage3_prefetch_bucket_size": "auto",
      "stage3_param_persistence_threshold": "auto",
      "sub_group_size": 1e9,
      "stage3_max_live_parameters": 1e9,
      "stage3_max_reuse_distance": 1e9,
      "stage3_gather_16bit_weights_on_model_save": true
  },
  "gradient_accumulation_steps": "auto",
  "train_batch_size": "auto",
  "train_micro_batch_size_per_gpu": "auto",
  "autotuning": {
    "enabled": true,
    "overwrite": false,
    "fast": false,
    "arg_mappings": {
      "train_micro_batch_size_per_gpu": "--per_device_train_batch_size",
      "gradient_accumulation_steps ": "--gradient_accumulation_steps"
    }
  }
}
```

produces an error here https://github.com/huggingface/accelerate/blob/b30a349078d5301411e8af421720eaa60e984fa2/src/accelerate/accelerator.py#L1602

```
ValidationError: 1 validation error for DeepSpeedZeroConfig
stage
  value is not a valid enumeration member; permitted: 0, 1, 2, 3 (type=type_error.enum; enum_values=[<ZeroStageEnum.disabled: 
[...]

PR #1901, opened 2023-08-29
deepspeed grad_acc_steps fixes

pacman100:
### What does this PR do?
1. Fixes https://github.com/huggingface/accelerate/issues/1830

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1507,12 +1507,13 @@ class Accelerator:
             batch_size_per_device = deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"]
             result = [obj for obj in args]
 
-        if self.gradient_accumulation_steps != deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"]:
-            logger.info(
-                f"Updating DeepSpeed's gradient accumulation steps to {self.gradient_accumulation_steps} from "
-                f"{deepspeed_plugin.deepspeed_config['gradient_accumulation_steps']}."
-            )
-            deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"] = self.gradient_accumulation_steps
+        # handle `gradient_accumulation_steps` when the value is `auto`
+        deepspeed_plugin.fill_match(
+            "gradient_accumulation_steps",
+            must_match=False,
+            gradient_accumulation_steps=self.gradient_accumulation_steps,
+        )
+
         config_kwargs = {
             "train_micro_batch_size_per_gpu": batch_size_per_device,
             "train_batch_size": batch_size_per_device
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -1507,12 +1507,13 @@ def _prepare_deepspeed(self, *args):
             batch_size_per_device = deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"]
             result = [obj for obj in args]
 
-        if self.gradient_accumulation_steps != deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"]:
-            logger.info(
-                f"Updating DeepSpeed's gradient accumulation steps to {self.gradient_accumulation_steps} from "
-                f"{deepspeed_plugin.deepspeed_config['gradient_accumulation_steps']}."
-            )
-            deepspeed_plugin.deepspeed_config["gradient_accumulation_steps"] = self.gradient_accumulation_steps
+        # handle `gradient_accumulation_steps` when the value is `auto`
```
BenjaminBossan, line 1510: Could you please explain the comment? IIUC, this updates the deepspeed plugin to use `self.gradient_accumulation_steps`, how exactly does this handle 'auto'?

tests/deepspeed/test_deepspeed.py
```diff
@@ -632,6 +632,42 @@ def test_autofill_dsconfig_from_ds_plugin(self, dtype):
...
+            train_set = RegressionDataset(length=80)
+            eval_set = RegressionDataset(length=20)
+            train_dataloader = DataLoader(train_set, batch_size=16, shuffle=True)
+            eval_dataloader = DataLoader(eval_set, batch_size=32, shuffle=False)
+            model = AutoModelForCausalLM.from_pretrained("gpt2")
+            dummy_optimizer = DummyOptim(params=model.parameters(), lr=5e-5, weight_decay=1e-4)
+            dummy_lr_scheduler = DummyScheduler(dummy_optimizer, warmup_num_steps=10, total_num_steps=1000)
+            model, _, train_dataloader, eval_dataloader, _ = accelerator.prepare(
+                model, dummy_optimizer, train_dataloader, eval_dataloader, dummy_lr_scheduler
+            )
```
BenjaminBossan, line 667: Is all of this setup necessary to check the deepspeed plugin?

BenjaminBossan (commented):
Thanks for taking care of the issue.

Unfortunately, I don't know enough about deepspeed to provide an in-depth review. From what I understand, this change would allow `gradient_accumulation_step` to be set to `'auto'`, which it is implicitly if it is not set via env var. The docs still state that it should be an `int` -- should `'auto'` also be mentioned or is it not something we want users to set?

I have some question in the comments, which might help me understand better, but apart from that, I hope someone else can do a proper review.

src/accelerate/accelerator.py
pacman100: It will only update the deepspeed_plugin if the value is `auto`. See https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/dataclasses.py#L646-L649

muellerzr (approved):
Excellent! Looks great to me :)

Issue #1902, opened 2023-08-29
Accelerator#_prepare_fsdp turns AdamW 8-bit optimizer into AdamW 32-bit


scottlogic-alex:
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-5.19.0-45-generic-x86_64-with-glibc2.35
- Python version: 3.11.2
- Numpy version: 1.25.2
- PyTorch version (GPU?): 2.1.0.dev20230802+cu121 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 62.79 GB
- GPU type: NVIDIA A40
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: FSDP
        - mixed_precision: bf16
        - use_cpu: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - fsdp_config: {'fsdp_auto_wrap_policy': 'TRANSFORMER_BASED_WRAP', 'fsdp_backward_prefetch_policy': 'NO_PREFETCH', 'fsdp_forward_prefetch': False, 'fsdp_offload_params': False, 'fsdp_sharding_strategy': 1, 'fsdp_state_dict_type': 'SHARDED_STATE_DICT', 'fsdp_sync_module_states': True, 'fsdp_transformer_layer_cls_to_wrap': 'LlamaDecoderLayer', 'fsdp_use_orig_params': True}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Training script is a [fork of `qlora.py`](https://github.com/scottlogic-alex/qlora/blob/d5e564ad3affcaf62c22912ba36cab2d77c2d585/qlora.py).

But if you have a preferred script, go ahead and use that; the triggers are:

- FSDP enabled
- AdamW 8-bit optimizer selected

### Configure accelerate

Set up your accelerate config (`~/.cache/huggingface/accelerate/default_config.yaml`) like so:

```yaml
[...]

PR #1905, opened 2023-08-29
[WIP] _prepare_fsdp: keep optimizer's 8-bitness and pagedness when reconstructing optimizer

scottlogic-alex:
Fixes https://github.com/huggingface/accelerate/issues/1902

Mostly raising this draft so I have something to point at and so that the proposed solution doesn't get lost.

See the TODOs for why I wouldn't necessarily merge it as-is.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1110,9 +1110,44 @@ class Accelerator:
                         "FSDP Warning: When using FSDP, several parameter groups will be conflated into "
                         "a single one due to nested module wrapping and parameter flattening."
                     )
+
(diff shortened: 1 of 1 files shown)


pacman100:
Hello, as mentioned in https://github.com/huggingface/accelerate/issues/1902#issuecomment-1697194776, this shouldn't be required when using the latest releases of Transformers and Accelerate. The important point to note here is that the recommended way when using FSDP is to prepare model before preparing optimizer as mentioned in the docs here https://huggingface.co/docs/accelerate/usage_guides/fsdp#a-few-caveats-to-be-aware-of: 
Screenshot 2023-08-29 at 4 58 49 PM

PR #1910, opened 2023-08-30
Fix docker images

muellerzr:
# What does this PR do?

Docker image workflows have been broken for the last few months silently, it comes from the workflow we use for it. Instead now we just manually upload it using the repository secrets, and I've confirmed this works and runs without issue. 

cc @LysandreJik @BenjaminBossan

### DIFF
--- a/.github/workflows/build-docker-images-release.yml
+++ b/.github/workflows/build-docker-images-release.yml
@@ -35,11 +35,9 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Build and Push CPU
-        uses: docker/build-push-action@v2
-        with:
-          context: ./docker/accelerate-cpu
-          push: true
-          tags: huggingface/accelerate-cpu:${{needs.get-version.outputs.version}}
+        run: |
+          docker build -t huggingface/accelerate-cpu -f ./docker/accelerate-cpu/Dockerfile .
+          docker push huggingface/accelerate-cpu::${{needs.get-version.outputs.version}}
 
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
Just for context, newer versions of `docker/build-push-action` would not work or do you prefer to avoid it completely because of the observed issues?

muellerzr:
Prefer to avoid completely so we have control over it.

.github/workflows/build-docker-images-release.yml
```diff
@@ -35,11 +35,9 @@ jobs:
...
 
       - name: Build and Push CPU
-        uses: docker/build-push-action@v2
-        with:
-          context: ./docker/accelerate-cpu
-          push: true
-          tags: huggingface/accelerate-cpu:${{needs.get-version.outputs.version}}
+        run: |
+          docker build -t huggingface/accelerate-cpu -f ./docker/accelerate-cpu/Dockerfile .
+          docker push huggingface/accelerate-cpu::${{needs.get-version.outputs.version}}
```
BenjaminBossan, line 40: Sorry, my docker knowledge is not deep enough: What does the `::` do? `:` would just be tag, right?

muellerzr: Yes, small typo, probably from copy/paste

muellerzr: (Though didn't seem to cause any issues)

BenjaminBossan (approved):
Thanks for clarifying. LGTM and big thanks for figuring out the issue.

PR #1916, opened 2023-08-31
More CI fun - run all test parts always

muellerzr:
# What does this PR do?

Noticed as I was paying close attention to the slow CI runners, if an earlier section failed then subsection tests would not be ran. We want all tests to be ran, and show what failed. For example, in https://github.com/huggingface/accelerate/actions/runs/6038483640/job/16385348804 the job stopped after the integration tests on multi-gpu because they failed, and then reported that to the slack runner

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -39,6 +39,7 @@ jobs:
           make test
           
       - name: Run examples on GPUs
+        if: always()
         run: |
           source activate accelerate
           pip uninstall comet_ml -y
@@ -79,11 +80,13 @@ jobs:
           make test_cli
 
       - name: Run Integration tests on GPUs
+        if: always()
         run: |
           source activate accelerate
           make test_integrations
 
       - name: Run examples on GPUs
+        if: always()
         run: |
           source activate accelerate
           pip uninstall comet_ml -y


BenjaminBossan (approved):
Eh, that's annoying that this requires to be added everywhere. I checked if there is a blanket option to enable this, but [fail-fast: false](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) seems to be only for matrix tests.

There is also [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error), which seems to invert the control (put it on the step that is allowed to fail and not on the step that should run despite failure).

Anyway, I think we can also live with this solution. Although, from my understanding, it means that the steps even run if the GH action is cancelled manually, that probably happens rarely enough that we can ignore it.

muellerzr:
Yep, exactly 😄 There shouldn't be a real reason to cancel them early, but if a need arises then we'll want to do a restructure of these workflows to allow it

Issue #1889, opened 2023-08-25
ValueError: Expected _orig_mod to NOT be FullyShardedDataParallel if using an `auto_wrap_policy`


AndrewMead10:
### System Info

```Shell
- `Accelerate` version: 0.22.0
- Platform: Linux-5.4.0-121-generic-x86_64-with-glibc2.31
- Python version: 3.10.11
- Numpy version: 1.24.3
- PyTorch version (GPU?): 2.0.1 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 125.74 GB
- GPU type: NVIDIA GeForce RTX 3090
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: FSDP
        - mixed_precision: fp16        
        - use_cpu: False        
        - debug: True
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main        
        - fsdp_config: {'fsdp_auto_wrap_policy': 'TRANSFORMER_BASED_WRAP',
           'fsdp_backward_prefetch_policy': 'BACKWARD_PRE',
           'fsdp_forward_prefetch': False,
           'fsdp_offload_params': False,
           'fsdp_sharding_strategy': 1,
           'fsdp_state_dict_type': 'SHARDED_STATE_DICT', 
           'fsdp_sync_module_states': True, 
           'fsdp_transformer_layer_cls_to_wrap': 'LlamaDecoderLayer',
           'fsdp_use_orig_params': False}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {'dynamo_backend': 'INDUCTOR'}
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am trying to fine tune LLama-2 using FSDP and accelerate, but when i run `accelerate launch finetune_base_model.py` i get `ValueError: Expected _orig_mod to NOT be FullyShardedDataParallel if using an auto_wrap_policy`. Below is the code that should reproduce the issue

```python
[...]

PR #1919, opened 2023-09-01
fix torch compile with FSDP

pacman100:
# What does this PR do?
1. Fixes https://github.com/huggingface/accelerate/issues/1889

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1215,7 +1215,7 @@ class Accelerator:
             for obj in args:
                 if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = type(obj) == FSDP
+                    is_type_fsdp = (type(obj) == FSDP) or isinstance(getattr(obj, "_orig_mod", None), FSDP)
                 if isinstance(obj, torch.optim.Optimizer):
                     optimizer_present = True
             if model_count > 1 and optimizer_present:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -96,6 +96,7 @@ from .utils import (
     wait_for_everyone,
 )
 from .utils.constants import FSDP_PYTORCH_VERSION
+from .utils.other import is_compiled_module
 
 
 if is_deepspeed_available():
@@ -1215,7 +1216,9 @@ class Accelerator:
             for obj in args:
                 if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = (type(obj) == FSDP) or isinstance(getattr(obj, "_orig_mod", None), FSDP)
+                    if is_compiled_module(obj):
+                        unwrapped_model = obj._orig_mod
+                    is_type_fsdp = (type(obj) == FSDP) or isinstance(unwrapped_model, FSDP)
                 if isinstance(obj, torch.optim.Optimizer):
                     optimizer_present = True
             if model_count > 1 and optimizer_present:
@@ -1423,8 +1426,9 @@ class Accelerator:
 
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
-                is_torch_compiled = getattr(model, "_orig_mod", None)
-                is_type_fsdp = (type(model) == FSDP) or isinstance(is_torch_compiled, FSDP)
+                if is_compiled_module(model):
+                    unwrapped_model = model._orig_mod
+                is_type_fsdp = (type(model) == FSDP) or isinstance(unwrapped_model, FSDP)
 
                 if not is_type_fsdp:
                     self.state.fsdp_plugin.set_auto_wrap_policy(model)
@@ -1469,7 +1473,7 @@ class Accelerator:
             elif self.distributed_type == DistributedType.TPU and self.state.fork_launched:
                 model = xmp.MpModelWrapper(model).to(self.device)
         # torch.compile should be called last and only if the model isn't already compiled.
-        if self.state.dynamo_plugin.backend != DynamoBackend.NO and getattr(model, "_orig_mod", None) is None:
+        if self.state.dynamo_plugin.backend != DynamoBackend.NO and not is_compiled_module(model):
             if not is_torch_version(">=", "2.0"):
                 raise ValueError("Using `torch.compile` requires PyTorch 2.0 or higher.")
             model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1213,10 +1213,9 @@ class Accelerator:
             optimizer_present = False
             is_type_fsdp = False
             for obj in args:
-                is_torch_compiled = getattr(obj, "_orig_mod", None)
-                if isinstance(obj, torch.nn.Module) or is_torch_compiled is not None:
+                if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = (type(obj) == FSDP) or isinstance(is_torch_compiled, FSDP)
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -1423,7 +1423,10 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
 
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
-                if type(model) != FSDP:
+                is_torch_compiled = getattr(model, "_orig_mod", None)
```
muellerzr, line 1426: Does this work if we use `accelerate.utils.other.is_compiled_module`? I'd rather us update that function to check `_orig_mod` if it does not as well, since we have a util for doing so.

src/accelerate/accelerator.py
```diff
@@ -1456,15 +1459,17 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
                 self._models[-1] = model
             elif self.distributed_type == DistributedType.MULTI_CPU:
                 kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
                 model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
             elif self.distributed_type == DistributedType.TPU and self.state.fork_launched:
                 model = xmp.MpModelWrapper(model).to(self.device)
-        # torch.compile should be called last.
-        if self.state.dynamo_plugin.backend != DynamoBackend.NO:
+        # torch.compile should be called last and only if the model isn't already compiled.
+        if self.state.dynamo_plugin.backend != DynamoBackend.NO and getattr(model, "_orig_mod", None) is None:
```
muellerzr, line 1472: Same here

src/accelerate/accelerator.py
```diff
@@ -1215,7 +1215,7 @@ def prepare(self, *args, device_placement=None):
             for obj in args:
                 if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = type(obj) == FSDP
+                    is_type_fsdp = (type(obj) == FSDP) or isinstance(getattr(obj, "_orig_mod", None), FSDP)
```
muellerzr, line 1218: And here

muellerzr (commented):
Thanks! What this PR is doing makes sense with me, a general question though as to if we can already use an existing tool already, which checks if a module was compiled via `torch.compile`.

src/accelerate/accelerator.py

src/accelerate/accelerator.py

src/accelerate/accelerator.py

muellerzr (approved):
Thanks! LG2M, cc @BenjaminBossan so he can get a little familiar with this side of the codebase

src/accelerate/accelerator.py
```diff
@@ -1215,7 +1216,8 @@ def prepare(self, *args, device_placement=None):
             for obj in args:
                 if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = type(obj) == FSDP
+                    unwrapped_model = obj._orig_mod if is_compiled_module(obj) else obj
+                    is_type_fsdp = (type(obj) == FSDP) or isinstance(unwrapped_model, FSDP)
```
BenjaminBossan, line 1220: Is there a specific reason for using strict type check on `obj` and `isinstance` on `unwrapped_model`?

src/accelerate/accelerator.py
```diff
@@ -1423,7 +1425,10 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
 
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
-                if type(model) != FSDP:
+                unwrapped_model = model._orig_mod if is_compiled_module(model) else model
+                is_type_fsdp = (type(model) == FSDP) or isinstance(unwrapped_model, FSDP)
```
BenjaminBossan, line 1429: Same question as above.

src/accelerate/accelerator.py
```diff
@@ -1456,15 +1461,17 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                             ),
                             auto_wrap_policy=fsdp_plugin.auto_wrap_policy,
                         )
-
+                # if the previous and current models are same, delete the previous one
+                if len(self._models) > 1 and (self._models[-2] is self._models[-1]):
+                    del self._models[-2]
```
BenjaminBossan, line 1466: Instead of fixing this here, would it be possible to prevent adding the identical model twice?

BenjaminBossan (approved):
Thanks for fixing this, LGTM. I have a few comments but no blockers for merging.

src/accelerate/accelerator.py
pacman100: added comments in the code to explain the reason

src/accelerate/accelerator.py
pacman100: Currently, I don't see a way of avoiding it, will circle back to this in a few days

src/accelerate/accelerator.py
BenjaminBossan: Even with the comment, I still don't understand why the check has to be strict for `obj` and `isinstance` for `obj._orig_mod`. Also, when the model is not wrapped, we assign `unwrapped_model = obj`. Then, the next line basically becomes:

`is_type_fsdp = (type(obj) == FSDP) or isinstance(obj, FSDP)`

i.e. first a strict type test on `obj`, then an `isinstance` test. To be clear, I don't think this leads to any bug, it's just confusing.

muellerzr: At first I was confused and agreed, however thinking on it more since we have `isinstance` next, and `isinstance` checks via inheritance too (which is what we just did in the latter kind), then yes should we not just do `isinstance(obj, FSDP)`? 

There are also other cases scattered throughout the library where we do `type(obj) == FSDP` rather than `isinstance`. Do we want to ensure that the top-level class is `FSDP`? Or can inheritance be allowed. If the latter, it'd be good to replace them all @pacman100

pacman100: Done, resolved it in latest commit

src/accelerate/accelerator.py
```diff
@@ -1215,7 +1216,11 @@ def prepare(self, *args, device_placement=None):
             for obj in args:
                 if isinstance(obj, torch.nn.Module):
                     model_count += 1
-                    is_type_fsdp = type(obj) == FSDP
+                    # if the model is compiled using PyTorch 2.0,
+                    # check that the wrapped model is FSDP or not;
+                    # else check if it is FSDP or not;
+                    unwrapped_model = obj._orig_mod if is_compiled_module(obj) else obj
+                    is_type_fsdp = isinstance(obj, FSDP) or isinstance(unwrapped_model, FSDP)
```
BenjaminBossan, line 1223: Minor nit, but note that if the obj is not compiled, the `isinstance` check is performed on the object twice. It doesn't really matter for performance, it could just be confusing for readers of the code.

src/accelerate/accelerator.py
```diff
@@ -1423,7 +1428,12 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
 
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
-                if type(model) != FSDP:
+                # In case the model is already compiled using PyTorch 2.0 and the wrapped model in it
+                # is a FSDP model, don't wrap it again
+                unwrapped_model = model._orig_mod if is_compiled_module(model) else model
+                is_type_fsdp = isinstance(model, FSDP) or isinstance(unwrapped_model, FSDP)
```
BenjaminBossan, line 1434: Same nit as above.

BenjaminBossan (approved):
LGTM, thanks Sourab.

src/accelerate/accelerator.py
pacman100: Done, resolved it in latest commit.

PR #1926, opened 2023-09-04
reduce gradient first for XLA when unscaling the gradients in mixed precision training with AMP.

statelesshz:
# What does this PR do?

As the [comment](https://github.com/huggingface/transformers/pull/25702#discussion_r1313098895) says, we need to reduce the gradient first for XLA when unscaling the gradients in mixed precision training with AMP, and a better implementation is moving this part of the logic to Accelerate. That's why we have this PR :-)

background: https://github.com/huggingface/transformers/pull/25702

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1987,9 +1987,9 @@ class Accelerator:
                 while isinstance(opt, AcceleratedOptimizer):
                     opt = opt.optimizer
                 # Reduce gradients first for XLA
-                if is_tpu_available():
+                if self.distributed_type == DistributedType.TPU:
                     gradients = xm._fetch_gradients(opt)
-                    xm.all_reduce("sum", gradients, scale=1.0 / xm.xrt_world_size())
+                    self.reduce(gradients, scale=1.0 / self.num_processes)
                 self.scaler.unscale_(opt)
 
     def clip_grad_norm_(self, parameters, max_norm, norm_type=2):
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1986,6 +1986,10 @@ class Accelerator:
             for opt in optimizer:
                 while isinstance(opt, AcceleratedOptimizer):
                     opt = opt.optimizer
+                # Reduce gradients first for XLA
+                if is_tpu_available():
+                    gradients = xm._fetch_gradients(opt)
+                    xm.all_reduce("sum", gradients, scale=1.0 / xm.xrt_world_size())
                 self.scaler.unscale_(opt)
 
     def clip_grad_norm_(self, parameters, max_norm, norm_type=2):
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2141,7 +2141,7 @@ class Accelerator:
             # Dataset had no length or raised an error
             return tensor
 
-    def reduce(self, tensor, reduction="sum"):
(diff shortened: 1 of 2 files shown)


src/accelerate/accelerator.py
```diff
@@ -1986,6 +1986,10 @@ def unscale_gradients(self, optimizer=None):
             for opt in optimizer:
                 while isinstance(opt, AcceleratedOptimizer):
                     opt = opt.optimizer
+                # Reduce gradients first for XLA
+                if is_tpu_available():
+                    gradients = xm._fetch_gradients(opt)
+                    xm.all_reduce("sum", gradients, scale=1.0 / xm.xrt_world_size())
```
muellerzr, line 1992:
```suggestion
                # Reduce gradients first for XLA
                if self.distributed_type == DistributedType.TPU:
                    gradients = xm._fetch_gradients(opt)
                    self.reduce(gradients, scale=1.0 / self.num_processes)
```

We should then update `accelerator.reduce to take in a `scale`, as well as `accelerate.utils.operations.reduce`, noting that `scale` is only valid on XLA.

muellerzr (commented):
Thanks! This is a good start, but we need a bit more to have it fully be good in accelerate.

src/accelerate/utils/operations.py
```diff
@@ -557,25 +557,29 @@ def reduce(tensor, reduction="mean"):
...
 
-    def _reduce_across_processes(tensor, reduction="mean"):
+    def _reduce_across_processes(tensor, reduction="mean", scale=1.0):
         state = PartialState()
         cloned_tensor = tensor.clone()
         if state.distributed_type == DistributedType.NO:
             return cloned_tensor
         if state.distributed_type == DistributedType.TPU:
-            xm.all_reduce("sum", cloned_tensor)
+            xm.all_reduce("sum", cloned_tensor, scale)
```
statelesshz, line 573: I'm not sure if this implementation is appropriate, i.e., using the scale parameter only on XLA and ignoring it in other scenarios.

statelesshz:
I have made some changes to my commit. @muellerzr Could you please review this pull request again?

src/accelerate/utils/operations.py
muellerzr: In this case it is, as `scale` is only an arg on xla :)

muellerzr (approved):
Excellent! LG2M, cc @BenjaminBossan for a final review

BenjaminBossan (approved):
Implementation-wise, this LGTM, thanks for working on this. I agree with Zach's comments.

src/accelerate/utils/operations.py
BenjaminBossan: Oh, one thing I notice now: This would not work for `reduction="mean"` because it would apply scaling twice, right? Should we raise an error here in that case, or, below, inside of `if reduction == "mean":`, place a guard for `DistributedType.TPU`?

statelesshz:
@muellerzr Yes, I will open a follow-up PR to remove this logic in Transformer, thanks for the reminder :-)
How do you feel about finishing it tomorrow?

src/accelerate/utils/operations.py
statelesshz: Nice catch! Sorry for not noticing this in time, I will submit a PR to add the necessary warnings soon

PR #1935, opened 2023-09-06
Check for invalid keys

muellerzr:
# What does this PR do?

This PR introduces a better error when loading the YAML file in by letting the user know what errors were actually present. 

Fixes # (issue)

Helps relieve the confusion when getting errors such as:

```python
ClusterConfig.__init__() got an unexpected keyword argument 'myargument'
```

And instead results with:
```python
Unknown keys in the config file: ['myargument'], please try upgrading your `accelerate` version or remove them.
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/commands/config/config_args.py
+++ b/src/accelerate/commands/config/config_args.py
@@ -60,7 +60,7 @@ def load_config_from_file(config_file):
                 config_class = ClusterConfig
             else:
                 config_class = SageMakerConfig
-            return config_class.from_json_file(json_file=config_file)
+                return config_class.from_json_file(json_file=config_file)
         else:
             if (
                 yaml.safe_load(f).get("compute_environment", ComputeEnvironment.LOCAL_MACHINE)
@@ -69,6 +69,7 @@ def load_config_from_file(config_file):
                 config_class = ClusterConfig
             else:
                 config_class = SageMakerConfig
+
             return config_class.from_yaml_file(yaml_file=config_file)
 
 
@@ -109,6 +110,12 @@ class BaseConfig:
             config_dict["use_cpu"] = False
         if "debug" not in config_dict:
             config_dict["debug"] = False
+        extra_keys = set(config_dict.keys()) - set(cls.__dataclass_fields__.keys())
+        if len(extra_keys) > 0:
+            raise ValueError(
+                f"Unknown keys in the config file: {list(extra_keys)}, please try upgrading your `accelerate` version or remove them."
+            )
+
         return cls(**config_dict)
 
     def to_json_file(self, json_file):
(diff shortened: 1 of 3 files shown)


src/accelerate/commands/config/config_args.py
```diff
@@ -109,6 +110,12 @@ def from_json_file(cls, json_file=None):
             config_dict["use_cpu"] = False
         if "debug" not in config_dict:
             config_dict["debug"] = False
+        extra_keys = set(config_dict.keys()) - set(cls.__dataclass_fields__.keys())
+        if len(extra_keys) > 0:
+            raise ValueError(
+                f"Unknown keys in the config file: {list(extra_keys)}, please try upgrading your `accelerate` version or remove them."
```
BenjaminBossan, line 116:
```suggestion
                f"Unknown keys in the config file: {sorted(extra_keys)}, please try upgrading your `accelerate` version or fix the keys."
```

To explain: 

1. Maybe the wrong key is just a typo, so "remove" is too specific. 
2. Explicitly say "the keys", otherwise users might think "them" refers to accelerate ;-)
3. `sorted` instead of `list` because `set` has no order, so theoretically the tests could just start failing with different Python versions. Also, maybe nicer for users to see the keys in alphabetical order.

src/accelerate/commands/config/config_args.py
```diff
@@ -137,6 +143,11 @@ def from_yaml_file(cls, yaml_file=None):
...
```
BenjaminBossan, line 149: Same argument as above.

src/accelerate/commands/config/config_args.py
```diff
@@ -60,7 +60,7 @@ def load_config_from_file(config_file):
                 config_class = ClusterConfig
             else:
                 config_class = SageMakerConfig
-            return config_class.from_json_file(json_file=config_file)
+                return config_class.from_json_file(json_file=config_file)
```
BenjaminBossan, line 63: Hmm, can you explain this change? To me, it looks like we're in the `if config_file.endswith(".json")` case. If we don't return here, the `config_class` will be called with `.from_yaml_file` farther below.

BenjaminBossan (changes requested):
Big :+1: for improving error messages, thanks for that. I have some comments, please take a look.

src/accelerate/commands/config/config_args.py
muellerzr: That would be a typo when I was trying to catch at the right time, will revert.

SunMarc (approved):
LGTM with @BenjaminBossan comments !

src/accelerate/commands/config/config_args.py
muellerzr: Went with a better option :)

BenjaminBossan (approved):
Thanks, looks good to me.

Nano-nit, `if len(extra_keys) > 0` could be `if extra_keys`, but I'm totally fine as is.

Issue #1925, opened 2023-09-03
[Feature Request] gather non-tensors objects for metrics (remove duplicates)


ywangmy:
As discussed in #1383, in certain NLP tasks, it is convenient to retain sentences as strings in the dataloader instead of tokenizing them into tensors. I propose implementing a function that gathers non-tensor objects and removes duplicates within the last batch. This function would complement the existing `gather_for_metrics()` (used for tensors) and enhance the capabilities for distributed evaluation in such scenarios. 

Thank you.

Lorenzobattistela:
Hey, I'd like to pick up this issue.

So basically we would need a new function like `gather_for_metrics` for objects and that removes duplicates within the last batch, correct?
As `gather_object` is implemented here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/operations.py
maybe it makes sense do build this in the same file.

What do you think @muellerzr ?

muellerzr:
@Lorenzobattistela or to do a check if something is a tensor or not in `gather_for_metrics`, and call `gather_object` and de-dupe instead :) This way it will automatically work for anything we pass to it. 

Yes that is exactly it! And ideally add some tests to `accelerate/test_utils/scripts/test_metrics`

Lorenzobattistela:
Yeah this approach makes sense. I'll start now. Feel free to assign me :)

Lorenzobattistela:
I thought something like this:

```python
def gather_for_metrics(self, input_data):
        """
        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system. For objects, duplicates are removed. Should be used
        for gathering the inputs and targets for metric calculation.

        Args:
            input (`torch.Tensor`, a nested tuple/list/dictionary of `torch.Tensor`, `object` or a nested tuple/list/dictionary of `object`):
                The tensors / objects for calculating metrics across all processes

        Example:

        ```python
        >>> # Assuming two processes, with a batch size of 5 on a dataset with 9 samples
        >>> import torch
        >>> from accelerate import Accelerator

        >>> accelerator = Accelerator()
        >>> dataloader = torch.utils.data.DataLoader(range(9), batch_size=5)
        >>> dataloader = accelerator.prepare(dataloader)
        >>> batch = next(iter(dataloader))
        >>> gathered_items = accelerator.gather_for_metrics(batch)
        >>> len(gathered_items)
        9
        ```
        """
        def check_input(input_data):
[...]

PR #1937, opened 2023-09-06
[WIP] Implementing gather_for_metrics with dedup for non tensor objects

Lorenzobattistela:
# What does this PR do?

Fixes # (issue)
#1925 

## Before submitting

      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr 

## Other

So, I wrote it as simple as I could. This is my first contribution here, so some good points are:
should we extract `check_input` function to a more general function on utils?
should we extract the dedup part for a utils function? I did not found the call mentioned in the issue. Also I'm a little lost as to testing. Did not find unit tests for functions, maybe I missed that. Style passing, and tests passing (without adding new cases).

Looking forward for suggestions and improvements. Thanks in advance.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2130,8 +2130,9 @@ class Accelerator:
 
         if not all_tensors:
             data = gather_object(input_data)
+        else:
+            data = self.gather(input_data)
 
-        data = self.gather(input_data)
         try:
             if self.gradient_state.end_of_dataloader:
                 # at the end of a dataloader, `gather_for_metrics` regresses to
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2098,8 +2098,8 @@ class Accelerator:
 
     def gather_for_metrics(self, input_data):
         """
-        Gathers `input_data` and potentially drops duplicates in the last batch if on a distributed system.
-        Should be used for gathering the inputs and targets for metric calculation.
+        Gathers `input_data` and potentially drops duplicates in the last batch if on a distributed system. Should be
+        used for gathering the inputs and targets for metric calculation.
 
         Args:
             input (`torch.Tensor`, `object`, a nested tuple/list/dictionary of `torch.Tensor`, or a nested tuple/list/dictionary of `object`):
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2098,7 +2098,7 @@ class Accelerator:
 
     def gather_for_metrics(self, input_data):
         """
-        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system.
+        Gathers `input_data` and potentially drops duplicates in the last batch if on a distributed system.
         Should be used for gathering the inputs and targets for metric calculation.
 
         Args:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2103,7 +2103,7 @@ class Accelerator:
 
         Args:
             input (`torch.Tensor`, a nested tuple/list/dictionary of `torch.Tensor`, `object` or a nested tuple/list/dictionary of `object`):
-                The tensors / objects for calculating metrics across all processes
+                The tensors or objects for calculating metrics across all processes
 
         Example:
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -80,8 +80,8 @@ from .utils import (
     is_megatron_lm_available,
     is_npu_available,
     is_safetensors_available,
-    is_torch_version,
     is_torch_tensor,
+    is_torch_version,
     is_tpu_available,
     is_xpu_available,
     load_fsdp_model,
@@ -2099,8 +2099,8 @@ class Accelerator:
 
     def gather_for_metrics(self, input_data):
         """
-        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system. For objects, duplicates are removed. Should be used
-        for gathering the inputs and targets for metric calculation.
+        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system. For
+        objects, duplicates are removed. Should be used for gathering the inputs and targets for metric calculation.
 
         Args:
             input (`torch.Tensor`, a nested tuple/list/dictionary of `torch.Tensor`, `object` or a nested tuple/list/dictionary of `object`):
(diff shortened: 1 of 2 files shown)


src/accelerate/accelerator.py
```diff
@@ -2095,14 +2097,14 @@ def gather(self, tensor):
         """
         return gather(tensor)
 
-    def gather_for_metrics(self, tensor):
+    def gather_for_metrics(self, input_data):
         """
-        Gathers `tensor` and potentially drops duplicates in the last batch if on a distributed system. Should be used
-        for gathering the inputs and targets for metric calculation.
+        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system. For
+        objects, duplicates are removed. Should be used for gathering the inputs and targets for metric calculation.
```
muellerzr, line 2103:
```suggestion
        Gathers `tensor` or object and potentially drops duplicates in the last batch if on a distributed system. Should be used for gathering the inputs and targets for metric calculation.
```
Since the end duplicates are removed in general, don't think we need this extra sentence.

src/accelerate/accelerator.py
```diff
@@ -2095,14 +2097,14 @@ def gather(self, tensor):
...
 
         Args:
-            tensor (`torch.Tensor`, or a nested tuple/list/dictionary of `torch.Tensor`):
-                The tensors for calculating metrics across all processes.
+            input (`torch.Tensor`, a nested tuple/list/dictionary of `torch.Tensor`, `object` or a nested tuple/list/dictionary of `object`):
+                The tensors / objects for calculating metrics across all processes
```
muellerzr, line 2107:
```suggestion
            input (`torch.Tensor`, a nested tuple/list/dictionary of `torch.Tensor`, `object` or a nested tuple/list/dictionary of `object`):
                The tensors or objects for calculating metrics across all processes
```

src/accelerate/accelerator.py
```diff
@@ -2120,8 +2122,29 @@ def gather_for_metrics(self, tensor):
...
+                    for item in obj:
+                        if item not in seen_objects:
+                            deduplicated_objects.append(item)
+                            seen_objects.add(item)
+                return deduplicated_objects
+
+            except Exception:
+                return obj
+
+        tensor = self.gather(input_data)
```
muellerzr, line 2147: I think we can do a bit better here:

```suggestion
        try:
            recursively_apply(lambda x: x, input_data, error_on_other_type=True)
            all_tensors = True
        except TypeError:
            all_tensors = False

        if not all_tensors:
            data= gather_object(input_data)

        data= self.gather(input_data)
```

From here, correct me if I'm wrong but the following `try/except` should then be able to filter out the deuplicates.

Note that `gather_for_metrics` should only remove the duplicated items at the end, not *all duplicates in the batch*, hence why we need this indexing. Does this make sense?

muellerzr (commented):
This is a great start! But there's still some more to do :) 

Re; tests, a test should be added to `src/accelerate/test_utils/scripts/extrenal_deps/test_metrics.py`. A CI runner will make sure that all of them get called.

src/accelerate/accelerator.py
```diff
@@ -2120,8 +2121,17 @@ def gather_for_metrics(self, tensor):
...
+        try:
+            recursively_apply(lambda x: x, input_data, error_on_other_type=True)
+            all_tensors = True
+        except TypeError:
+            all_tensors = False
+
+        if not all_tensors:
+            data= gather_object(input_data)
+
+        data= self.gather(input_data)
```
muellerzr, line 2134: We need an else here, sorry probably did a poor suggestion :) 

```suggestion
        if not all_tensors:
            data = gather_object(input_data)
        else:
            data = self.gather(input_data)
```

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -149,6 +149,40 @@ def test_mrpc(dispatch_batches: bool = False, split_batches: bool = False):
...
+
+        def __iter__(self):
+            for element in self.data:
+                yield element
+
+    iterable_dataset = DummyIterableDataset([n for n in range(30)])
+    dataloader = DataLoader(iterable_dataset, batch_size=4)
+    accelerator = Accelerator()
+    prepared_dataloader = accelerator.prepare(dataloader)
+    assert isinstance(prepared_dataloader, DataLoaderDispatcher)
```
muellerzr, line 168:
```suggestion
```
Keep tests as just testing what we care about :)

src/accelerate/accelerator.py
```diff
@@ -2120,7 +2121,17 @@ def gather_for_metrics(self, tensor):
         9
         ```
         """
-        tensor = self.gather(tensor)
+
+        try:
+            recursively_apply(lambda x: x, input_data, error_on_other_type=True)
```
BenjaminBossan, line 2126: Hmm, this seems a bit of a abuse of `recursively_apply` to me (as in: using it for checking the type). It is also a bit wasteful because we recursively traverse the whole data structure here, only to the same again further below when calling `gather`. @muellerzr would it be possibly to to just do something like this instead:

```python
try:
    data = self.gather(input_data)
except TypeError:
    data = gather_object(input_data)
```

i.e. not checking for `all_tensors` at all?

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -149,6 +149,40 @@ def test_mrpc(dispatch_batches: bool = False, split_batches: bool = False):
         ), f"Baseline and Distributed are not the same for key {key}:\n\tBaseline: {baseline[key]}\n\tDistributed: {distributed[key]}\n"
 
 
+def test_gather_for_metrics_with_non_tensor_objects_iterable_dataset():
```
BenjaminBossan, line 152: I think the test is not being run at all, right? Similar to `test_gather_for_metrics_with_iterable_dataset`, it needs to be explicitly called in the `main` function, because the tests in this file are _not_ collected by pytest but rather called directly.

BenjaminBossan (changes requested):
With the comments from Zach, this looks mostly good to me, thanks for working in this feature. I have two comments, one is more of an efficiency question, the other is that it seems that the test is not run. Please take a look.

src/accelerate/accelerator.py
muellerzr: gather is a sync op, so doing recursively apply avoids double sync ops and a slowdown on DDP

Lorenzobattistela: I started with the same strategy as @BenjaminBossan , but then changed to follow other tests "style". So we keep it as it is?

BenjaminBossan (approved):
Thanks a lot, LGTM.

muellerzr (approved):
Thanks again for the improvement!

Lorenzobattistela:
Appreciate all the help!

PR #1940, opened 2023-09-07
Introduce breakpoint API

muellerzr:
# Introduce breakpoint API

## What does this add?

This PR adds two new functions to `Accelerator`: `set_breakpoint` and `check_breakpoint` based on https://discuss.pytorch.org/t/how-to-use-break-in-distributeddataparallel-training/88296/7

## Who is it for?

https://discuss.huggingface.co/t/early-stopping-for-eval-loss-causes-timeout/51349

## Why is it needed?

When doing early stopping in DDP, if each process has a specific conditional that it can check, where it may not be synchronized across all of them, a `break` can happen on process 0 but not on process 1. As a result this will cause the code to hang indefinitely until a timeout occurs. 

To address that this PR introduces a new `set` and `check` breakpoint API, which should be used in-tandem with such conditionals to ensure that the breakpoint will be reached. This API can be used on any distributed type, not just DDP/multi-gpu

### User-facing:

`Accelerator.set_breakpoint` and `Accelerator.check_breakpoint`

### Internal structure:

The `Accelerator` object now keeps track of a tensor at `self.flag_tensor`

## Basic Usage Example(s):

```python
# Assume `should_do_breakpoint` is a custom defined function that returns a conditional, 
# and that conditional might be true only on process 1
if should_do_breakpoint(loss):
    accelerator.set_breakpoint()

# Later in the training script when we need to check for the breakpoint
if accelerator.check_breakpoint():
    break
```

A prime example would be triggering early stopping, where stopping on process 0 should trigger stopping the process across all of them.

### DIFF
--- a/docs/source/concept_guides/deferring_execution.md
+++ b/docs/source/concept_guides/deferring_execution.md
@@ -111,7 +111,7 @@ with accelerator.main_process_first():
 
 ## Applying checks such as Early Stopping
 
-To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used. Useful examples
+To have a check that works with a flag set by a particular process, the `set_trigger` and `check_trigger` API should be used. Useful examples
 for doing so can include situations such as using early stopping and monitoring the loss (as each loss slightly differs on each process).
 
 Call [`Accelerator.set_trigger`] when your condition has been met, and [`Accelerator.check_trigger`] when checking if that condition has been met in any process:
--- a/docs/source/concept_guides/deferring_execution.md
+++ b/docs/source/concept_guides/deferring_execution.md
@@ -108,3 +108,19 @@ with accelerator.main_process_first():
         remove_columns=["idx", "sentence1", "sentence2"],
     )
 ```
+
+## Applying checks such as Early Stopping
+
+To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used.
+
+Call [`Accelerator.set_breakpoint`] when your condition has been met, and [`Accelerator.check_breakpoint`] when checking if that condition has been met in any process:
+
+```python
+# Assume `should_do_breakpoint` is a custom defined function that returns a conditional
+if should_do_breakpoint(loss):
+    accelerator.set_breakpoint()
+
+# Later in the training script when we need to check for the breakpoint
+if accelerator.check_breakpoint():
+    break
+```
--- a/docs/source/concept_guides/deferring_execution.md
+++ b/docs/source/concept_guides/deferring_execution.md
@@ -111,16 +111,17 @@ with accelerator.main_process_first():
 
 ## Applying checks such as Early Stopping
 
-To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used.
+To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used. Useful examples
+for doing so can include situations such as using early stopping and monitoring the loss (as each loss slightly differs on each process).
 
-Call [`Accelerator.set_breakpoint`] when your condition has been met, and [`Accelerator.check_breakpoint`] when checking if that condition has been met in any process:
+Call [`Accelerator.set_trigger`] when your condition has been met, and [`Accelerator.check_trigger`] when checking if that condition has been met in any process:
 
 ```python
-# Assume `should_do_breakpoint` is a custom defined function that returns a conditional
-if should_do_breakpoint(loss):
-    accelerator.set_breakpoint()
+# Assume `should_do_early_stopping` is a custom defined function that returns a conditional
+if should_do_early_stopping(loss):
+    accelerator.set_trigger()
 
 # Later in the training script when we need to check for the breakpoint
-if accelerator.check_breakpoint():
+if accelerator.check_trigger():
     break
 ```
--- a/docs/source/concept_guides/deferring_execution.md
+++ b/docs/source/concept_guides/deferring_execution.md
@@ -117,11 +117,14 @@ for doing so can include situations such as using early stopping and monitoring 
 Call [`Accelerator.set_trigger`] when your condition has been met, and [`Accelerator.check_trigger`] when checking if that condition has been met in any process:
 
 ```python
-# Assume `should_do_early_stopping` is a custom defined function that returns a conditional
-if should_do_early_stopping(loss):
-    accelerator.set_trigger()
+for (x,y) in data_loader:
+    logits = model(x)
+    loss = loss_func(logits, y)
+    # Assume `should_do_early_stopping` is a custom defined function that returns a conditional
+    if should_do_early_stopping(loss):
+        accelerator.set_trigger()
 
-# Later in the training script when we need to check for the breakpoint
-if accelerator.check_trigger():
-    break
+    # Later in the training script when we need to check for the breakpoint
+    if accelerator.check_trigger():
+        break
 ```
--- a/examples/by_feature/early_stopping.py
+++ b/examples/by_feature/early_stopping.py
@@ -197,6 +197,15 @@ def training_function(config, args):
                 lr_scheduler.step()
                 optimizer.zero_grad()
 
+            # New code
+            # Check if we should stop the training on any processes
+            if callback.check_early_stopping(loss.item()):
+                accelerator.set_trigger()
+
+            # If so, we break the loop
+            if accelerator.check_trigger():
+                break
+
         model.eval()
         for step, batch in enumerate(eval_dataloader):
             # We could avoid this line since we set the accelerator with `device_placement=True`.
@@ -211,15 +220,6 @@ def training_function(config, args):
             )
 
         eval_metric = metric.compute()
-
-        # New code
-        # Check if we should stop the training on any processes
-        if callback.check_early_stopping(outputs.loss.item()):
-            accelerator.set_breakpoint()
-
-        # If so, we break the loop
-        if accelerator.check_breakpoint():
-            break
 
         # Use accelerator.print to print only on the main process.
         accelerator.print(f"epoch {epoch}:", eval_metric)
--- a/examples/by_feature/early_stopping.py
+++ b/examples/by_feature/early_stopping.py
@@ -1,0 +1,246 @@
+# coding=utf-8
+# Copyright 2021 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+
+import evaluate
+import torch
+from datasets import load_dataset
+from torch.optim import AdamW
+from torch.utils.data import DataLoader
+from transformers import AutoModelForSequenceClassification, AutoTokenizer, get_linear_schedule_with_warmup, set_seed
+
+from accelerate import Accelerator, DistributedType
+
+
+########################################################################
+# This is a fully working simple example to use Accelerate
+# specifically showcasing how to perform early stopping,
+# and builds off the `nlp_example.py` script
+#
+# This example trains a Bert base model on GLUE MRPC
+# in any of the following settings (with the same script):
+#   - single CPU or single GPU
+#   - multi GPUS (using PyTorch distributed mode)
+#   - (multi) TPUs
+#   - fp16 (mixed-precision) or fp32 (normal precision)
+#
+# To run it in each of these various modes, follow the instructions
+# in the readme for examples:
+# https://github.com/huggingface/accelerate/tree/main/examples
+#
+########################################################################
+
+
+MAX_GPU_BATCH_SIZE = 16
+EVAL_BATCH_SIZE = 32
+
+
+def get_dataloaders(accelerator: Accelerator, batch_size: int = 16):
+    """
+    Creates a set of `DataLoader`s for the `glue` dataset,
+    using "bert-base-cased" as the tokenizer.
+
+    Args:
+        accelerator (`Accelerator`):
+            An `Accelerator` object
+        batch_size (`int`, *optional*):
+            The batch size for the train and validation DataLoaders.
+    """
+    tokenizer = AutoTokenizer.from_pretrained("bert-base-cased")
+    datasets = load_dataset("glue", "mrpc")
+
+    def tokenize_function(examples):
+        # max_length=None => use the model max length (it's actually the default)
+        outputs = tokenizer(examples["sentence1"], examples["sentence2"], truncation=True, max_length=None)
+        return outputs
+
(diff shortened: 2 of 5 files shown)


src/accelerate/accelerator.py
```diff
@@ -1963,6 +1966,65 @@ def backward(self, loss, **kwargs):
...
+        >>> # Assume later in the training script
+        >>> if accelerator.check_breakpoint():
+        ...     break
+        ```
+        """
+        # Now that we are outside `__init__`, we can initialize it if it is `None` on device
+        if self.flag_tensor is None:
+            self.flag_tensor = torch.tensor(0, device=self.device)
+        flag_tensor = self.reduce(self.flag_tensor)
+        if flag_tensor.item() == 1:
```
BenjaminBossan, line 2023: Is it possible that the reduction operation, which is `sum` by default, returns a value greater than 1? If so, the operator should be `and`, or the check for `>=1`.

docs/source/concept_guides/deferring_execution.md
```diff
@@ -108,3 +108,19 @@ with accelerator.main_process_first():
         remove_columns=["idx", "sentence1", "sentence2"],
     )
 ```
+
+## Applying checks such as Early Stopping
+
+To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used.
```
BenjaminBossan, line 114: IMO, as a reader, I would be a bit confused about when I have to use this and when not. For instance, IIUC, it is not necessarily true that early stopping requires this feature. E.g. if the user collects some metrics and correctly applies `gather_for_metrics`, all processes should be synced up and there is no danger of getting a stopping signal on one process but not the others. IIUC, the initial issue the user had was caused by directly checking the loss, which can differ between processes.

BenjaminBossan (commented):
Thanks for investigating that issue and coming up with a solution.

This is not an in depth review yet, but maybe these comments can already help a bit. Apart from those, I also wondered about the term "breakpoint". I mostly associate it to the context of debugging, which is not what is happening here, but maybe my view is too narrow. In reality, this feature is really a synchronized flag and could in theory be used as a basis for all kinds of features, right?

src/accelerate/accelerator.py
muellerzr: It'd be a rare condition, but certainly something that could happen.

muellerzr:
@BenjaminBossan I'd say it's a conditional trigger setup, more than synchronizing flags. Because also synchronizing flags aren't really true, it's checking if a particular condition should be met that affects all processes which occurred in a singular process. (Or more than one of these processes). 

Agreed the naming could use some work however, if anyone has some ideas lmk otherwise I'll think on it :)

BenjaminBossan:
> I'd say it's a conditional trigger setup, more than synchronizing flags

True, it's basically an `any` operator that works across processes.

> Agreed the naming could use some work however, if anyone has some ideas lmk otherwise I'll think on it :)

Always the hard questions. Since you used the word yourself, I wonder if it could be something with "trigger".

docs/source/concept_guides/deferring_execution.md
muellerzr: Expanded on this a bit more

docs/source/concept_guides/deferring_execution.md
```diff
@@ -108,3 +108,20 @@ with accelerator.main_process_first():
         remove_columns=["idx", "sentence1", "sentence2"],
     )
 ```
+
+## Applying checks such as Early Stopping
+
+To have a check that works with a flag set by a particular process, the `check` and `set` breakpoint API should be used. Useful examples
```
BenjaminBossan, line 114: How about:

```suggestion
To have a check that works with a flag set by a particular process, the `set_trigger` and `check_trigger` API should be used. Useful examples
```

(the suggested change may require a different line break)

docs/source/concept_guides/deferring_execution.md
```diff
@@ -108,3 +108,20 @@ with accelerator.main_process_first():
...
+
+Call [`Accelerator.set_trigger`] when your condition has been met, and [`Accelerator.check_trigger`] when checking if that condition has been met in any process:
+
+```python
+# Assume `should_do_early_stopping` is a custom defined function that returns a conditional
+if should_do_early_stopping(loss):
+    accelerator.set_trigger()
+
+# Later in the training script when we need to check for the breakpoint
+if accelerator.check_trigger():
```
BenjaminBossan, line 125: I wonder if the example would be more easily understood if the loop is made explicit (since `break` requires some kind of loop). E.g.:

```python
for batch in data_loader:
    if accelerator.check_trigger():
        break
    # training code
```

examples/by_feature/early_stopping.py
```diff
@@ -0,0 +1,246 @@
+# coding=utf-8
```
BenjaminBossan, line 1: AFAICT, this whole script is almost identical to `examples/nlp_example.py`, just with the addition of early stopping. I wonder: Would it make more sense to add the early stopping code to that script and enable/disable it via an additional command line argument?

src/accelerate/accelerator.py
```diff
@@ -1963,6 +1966,65 @@ def backward(self, loss, **kwargs):
...
+        >>> if accelerator.check_trigger():
+        ...     break
+        ```
+        """
+        # Now that we are outside `__init__`, we can initialize it if it is `None` on device
+        if self.flag_tensor is None:
+            self.flag_tensor = torch.tensor(0, device=self.device)
+        flag_tensor = self.reduce(self.flag_tensor)
+        if flag_tensor.item() >= 1:
+            self.flag_tensor = None
```
BenjaminBossan, line 2024: Could `self.flag_tensor` be set to `torch.tensor(0, device=self.device)` at this point?

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -541,6 +541,20 @@ def test_split_between_processes_tensor():
     state.wait_for_everyone()
 
 
+def test_breakpoint():
```
BenjaminBossan, line 544:
```suggestion
def test_trigger():
```

Or possible: `test_flag_tensor`. If renamed, also rename it further below.

BenjaminBossan (approved):
Thanks a lot, LGTM. I have a few suggestions, but no blockers for merging. Feel free to follow or ignore them.

examples/by_feature/early_stopping.py
muellerzr: These scripts are designed to be just that :)

BenjaminBossan: Sorry, I do not understand your comment, could you please elaborate? :)

If you mean that the script should stay simple: the `examples/nlp_example.py` already has a couple of command line arguments, so adding one more wouldn't be very costly.

muellerzr: The `by_feature` script Re meant to be simple additions to `nlp_example.py` showing just a single feature being added to them, so users aren't overwhelmed with a million lines trying to find and do one thing. The `nlp_example.py` remains core, and small additions as we add new features get added in as feature scripts

BenjaminBossan: Ah I see, that makes sense.

docs/source/concept_guides/deferring_execution.md
muellerzr: Achieved via:

```python
for (x,y) in data_loader:
    logits = model(x)
    loss = loss_func(logits, y)
    # Assume `should_do_early_stopping` is a custom defined function that returns a conditional
    if should_do_early_stopping(loss):
        accelerator.set_trigger()

    # Later in the training script when we need to check for the breakpoint
    if accelerator.check_trigger():
        break
```

src/accelerate/accelerator.py
```diff
@@ -1963,6 +1966,65 @@ def backward(self, loss, **kwargs):
...
+        >>> if accelerator.check_breakpoint():
+        ...     break
+        ```
+        """
+        self.flag_tensor = torch.tensor(1, device=self.device)
+
+    def check_trigger(self):
+        """
+        Checks if the internal trigger tensor has been set to 1 in any of the processes. If so, will return `True` and
+        reset the trigger tensor to 0.
```
SunMarc, line 1998: I don't know if it might be useful but we can let the user decide if they want to reset the trigger or not ? And by default we reset it

muellerzr: The intended behavior is not to, but we can do this later if users need that flexibility. (And to not overload the API with too many options :) )

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -541,6 +541,20 @@ def test_split_between_processes_tensor():
...
+    # should start with being false
+    assert accelerator.check_trigger() is False
+
+    # set a breakpoint on the main process
+    if accelerator.is_main_process:
+        accelerator.set_trigger()
+
+    # check it's been activated across all processes
+    # calls `all_reduce` and triggers a sync
+    assert accelerator.check_trigger() is True
```
SunMarc, line 555: maybe add an assert to check that the trigger has been reset

SunMarc (approved):
LGTM ! Left a few suggestions ;)

beyondguo:
Hi @muellerzr , is `set_breakpoint` diabled now? I got `AttributeError: 'Accelerator' object has no attribute 'set_breakpoint'` with the latest acclerate (v0.31.0).

Actually I am very excited to find this PR, since this seems quite relevant to my problem (similar to the early stopping problem), I'm using 2-gpu to train a roberta model, and I want to calculate the valication loss after `eval_steps`, here's the core code:
```python
...
model, optimizer, train_dataloader, val_dataloader = accelerator.prepare(model, optimizer, train_dataloader, val_dataloader)
...
best_eval_loss = float('inf')

# Training loop:
    for batch in train_dataloader:
        model.train()
        outputs = model(**batch)
        loss = outputs.loss 
        accelerator.backward(loss)
        optimizer.step()
        optimizer.zero_grad()
        progress_bar.update(1)

        # validation when reach 'eval_steps'
        if progress_bar.n > 10 and progress_bar.n % args.eval_steps == 0:
            model.eval()
            losses = []
            for val_batch in val_dataloader:
                # ----------------> stuck here!
                with torch.no_grad():
                    outputs = model(**val_batch)
                e_loss = outputs.loss
                losses.append(accelerator.gather_for_metrics(e_loss.repeat(val_bs)))
            losses = torch.cat(losses)
            eval_loss = torch.mean(losses)
            print("eval_loss", eval_loss.item())
            
            # save the current best model
            if eval_loss.item() < best_eval_loss:
                print("Current best model!, Steps:", progress_bar.n, "Eval loss:", eval_loss.item())
                # save
                ...
```
The process just got stuck when computing the first validation batch.  Two GPUs are both 100% utilization, so I guess the process is hanging. 

After a long time, an error will occur:
```
[E ProcessGroupNCCL.cpp:828] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2315, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1808543 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:828] [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2315, OpType=ALLGATHER, Timeout(ms)=1800000) ran for 1808546 milliseconds before timing out.
...
  File "rolling_model_train.py", line 289, in <module>
    train()
  File "rolling_model_train.py", line 195, in train
    losses.append(accelerator.gather_for_metrics(loss.repeat(val_bs)))
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/accelerate/accelerator.py", line 2217, in gather_for_metrics
[E ProcessGroupNCCL.cpp:455] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:460] To avoid data inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2315, OpType=ALLGATHER, Timeout(ms)=1800000) ran for 1808546 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:455] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:460] To avoid data inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2315, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1808543 milliseconds before timing out.
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: -6) local_rank: 0 (pid: 89866) of binary: /home/guoby/app/Anaconda3-2021.05/envs/news/bin/python
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/bin/accelerate", line 8, in <module>
    sys.exit(main())
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/accelerate/commands/accelerate_cli.py", line 47, in main
    args.func(args)
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/accelerate/commands/launch.py", line 977, in launch_command
    multi_gpu_launcher(args)
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/accelerate/commands/launch.py", line 646, in multi_gpu_launcher
    distrib_run.run(args)
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/torch/distributed/run.py", line 785, in run
    elastic_launch(
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/torch/distributed/launcher/api.py", line 134, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
  File "/home/guoby/app/Anaconda3-2021.05/envs/news/lib/python3.8/site-packages/torch/distributed/launcher/api.py", line 250, in launch_agent
    raise ChildFailedError(
torch.distributed.elastic.multiprocessing.errors.ChildFailedError: 
======================================================
rolling_model_train.py FAILED
------------------------------------------------------
Failures:
[1]:
  time      : 2024-06-20_15:02:22
  rank      : 1 (local_rank: 1)
  exitcode  : -6 (pid: 89867)
  error_file: <N/A>
  traceback : Signal 6 (SIGABRT) received by PID 89867
------------------------------------------------------
Root Cause (first observed failure):
[0]:
  time      : 2024-06-20_15:02:22
  rank      : 0 (local_rank: 0)
  exitcode  : -6 (pid: 89866)
  error_file: <N/A>
  traceback : Signal 6 (SIGABRT) received by PID 89866
======================================================
```

Note that, if I move the validation code after each epoch, instead during one epoch, the program runs fine! Which is like this:
```python
# Training loop:
    for batch in train_dataloader:
        model.train()
        outputs = model(**batch)
        loss = outputs.loss 
        accelerator.backward(loss)
        optimizer.step()
        optimizer.zero_grad()
        progress_bar.update(1)

    # validation:
    model.eval()
    losses = []
    for val_batch in val_dataloader:
        with torch.no_grad():
            outputs = model(**val_batch)
        e_loss = outputs.loss
        losses.append(accelerator.gather_for_metrics(e_loss.repeat(val_bs)))
    losses = torch.cat(losses)
    eval_loss = torch.mean(losses)
    print("eval_loss", eval_loss.item())

```

Could you help me on this? Thanks a lot!

PR #1941, opened 2023-09-07
Fix non persistant buffer dispatch

SunMarc:
# What does this PR do ? 
This PR fixes the offload of buffers when using `dispatch_model` or `cpu_offload` (related to hooks). 
Currently, `dispatch_model` will offload all buffers if `offload_buffers` is set to `True`. This is problematic when we have non persistant buffers that are, by definition, not saved in the state_dict of the model. When we use `dispatch_model` alone, we will store the `state_dict` of some modules on the disk. Hence, we won't be able to retrieve them. To fix that, we simply remove the non persistant buffers from the list of buffers to offload. 

PS: diffusers team actually use offload buffers


BenjaminBossan (commented):
Thanks for addressing that issue. In general, the implementation LGTM, so no comments from me on that.

One thing I would like to see are some tests to check for excluding non-persistent buffers. There are some tests for `named_module_tensors`, which could be extended to check for the new option. This doesn't encapsulate every addition of this PR, so ideally the device hooks could also be tested, not sure how easy that is.

Another more general comment is that now, the new parameter `remove_non_persistant` has to be passed around all over the place. Are there use cases that _do_ require to include non-persistent buffers? If not, always removing them would simplify things.

Finally, the functions `big_modeling.load_checkpoint_and_dispatch` and `utils.bnb.load_and_quantize_model` call `dispatch_model`, should their signatures be extended to include `remove_non_persistant` too?

SunMarc:
> Another more general comment is that now, the new parameter `remove_non_persistant` has to be passed around all over the place. Are there use cases that do require to include non-persistent buffers? If not, always removing them would simplify things.

I don't think. If a user needs to offload non-persistant buffers, he will need to create a mapping using directly `named_buffers`. Since we didn't see any issue on that, mabye it means that nobody attempted to offload non persistant the buffers ? If so, it is would be easier to just remove them. Let me know that you think !

> One thing I would like to see are some tests to check for excluding non-persistent buffers. There are some tests for named_module_tensors, which could be extended to check for the new option. This doesn't encapsulate every addition of this PR, so ideally the device hooks could also be tested, not sure how easy that is.
In any case, I will add a few test for named_modules_tensors and try to come up with something for the hooks. 

> Finally, the functions big_modeling.load_checkpoint_and_dispatch and utils.bnb.load_and_quantize_model call dispatch_model, should their signatures be extended to include remove_non_persistant too?
Yes if we decide to keep `remove_non_persistant`

BenjaminBossan:
> I don't think. If a user needs to offload non-persistant buffers, he will need to create a mapping using directly `named_buffers`. Since we didn't see any issue on that, mabye it means that nobody attempted to offload non persistant the buffers ? If so, it is would be easier to just remove them. Let me know that you think !

Maybe @muellerzr can comment on that? If we decide to make the change, we should ensure to mention it somewhere (like making a note to include it in the next release text).

SunMarc:
I've added a few tests ! Let me know if you want to make the change. If we make the change @muellerzr, I will keep `remove_non_persistent` in the `get_named_tensors`. But I will remove it for the other functions by setting it to `True` when offload_buffers is `True`.

muellerzr:
@SunMarc let's go with that, and if users request that we expand the param/trickle it down to lower levels, we can.

SunMarc:
still having an issue with falcon 180B gptq offload. I will put it in draft for now

SunMarc:
After digging a little bit into it, I think we should keep `remove_non_persistent` arg for named_module_tensors. To keep BC, I've set it to `False`. Then, I only set it to `True` in offloaded modules. Also, I don't expand the param to `dispatch_model` or `cpu_offload` as I don't think anyone will change the behavior. cc @muellerzr @BenjaminBossan

sayakpaul:
Thanks for tagging me in. From the `diffusers` side the following piece of code was breaking:

```python
from diffusers import PixArtAlphaPipeline
import torch

def bytes_to_giga_bytes(bytes):
    return bytes / 1024 / 1024 / 1024

pipe = PixArtAlphaPipeline.from_pretrained(
    "PixArt-alpha/PixArt-XL-2-1024-MS", 
    torch_dtype=torch.float16
)
pipe.enable_sequential_cpu_offload()

images = pipe("hey", num_images_per_prompt=4).images[0]
print(
    f"Max memory allocated: {bytes_to_giga_bytes(torch.cuda.max_memory_allocated())} GB"
)
```

BenjaminBossan (approved):
Thanks for fixing this issue and adding a couple of tests to check it.

I'm a bit concerned that we have to rely on a private attribute `_non_persistent_buffers_set` but AFAICT, it's the only place that information is stored, so there is no way around it, and it seems to exist since many PyTorch releases, which hopefully means that it's safe.

src/accelerate/utils/modeling.py
```diff
@@ -377,13 +379,40 @@ def named_module_tensors(module: nn.Module, include_buffers: bool = True, recurs
...
+    Gather all non persistent buffers of a given modules into a set
+
+    Args:
+        module (`nn.Module`):
+            The module we want the non persistent buffers on.
+        recurse (`bool`, *optional*, defaults to `False`):
+            Whether or not to go look in every submodule or just return the direct non persistent buffers.
+    """
+
+    non_persistent_buffers_set = module._non_persistent_buffers_set
```
muellerzr, line 410: We have an issue here. This capability was only added in torch 1.13, and accelerate requires torch 1.10/the last 2 years of torch releases. We should do a guard here for the time being + modify the logic as such. https://github.com/pytorch/pytorch/commit/817eb94ff494e07e94a9effb626ce2121f226a54

muellerzr (commented):
This is a great fix! However one big issue with the logic here we need to address :)

src/accelerate/utils/modeling.py
SunMarc: Good catch ! Thanks for looking into =) I will add a guard.

SunMarc: Looks like in this commit, the arg was already present. From the torch 1.10 `[module.py](https://github.com/pytorch/pytorch/blob/v1.10.0/torch/nn/modules/module.py)` file, the arg is also there. So I think we are good. LMK if there is something I missed. Thanks for looking into it, I forgot to do it beforehand!

muellerzr: Good catch! LG2M then!

muellerzr (approved):

Issue #1943, opened 2023-09-08
fsdp example has an error


hamelsmu:
### System Info

When running [this example](https://huggingface.co/docs/accelerate/usage_guides/fsdp) got  Exception("Could not find the transformer layer class to wrap in the model.")

Running 
```
accelerate launch examples/nlp_example.py
```

With this config

```
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: FSDP
downcast_bf16: 'no'
fsdp_config:
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: FULL_STATE_DICT
  fsdp_transformer_layer_cls_to_wrap: GPT2Block
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 2
use_cpu: false
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Run the code in the docs as-is

### Expected behavior

Shouldn't get an error when running code in the docs as-is

muellerzr:
CC @pacman100

PR #1947, opened 2023-09-08
fix the fsdp docs

pacman100:
# What does this PR do?
1. Fixes https://github.com/huggingface/accelerate/issues/1943

### DIFF
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -49,7 +49,7 @@ fsdp_config:
   fsdp_offload_params: false
   fsdp_sharding_strategy: 1
   fsdp_state_dict_type: FULL_STATE_DICT
-  fsdp_transformer_layer_cls_to_wrap: GPT2Block
+  fsdp_transformer_layer_cls_to_wrap: BertLayer
 machine_rank: 0
 main_process_ip: null
 main_process_port: null


muellerzr (approved):
Thanks for the quick fix!

hamelsmu:
How do you figure out the right value for this config?  Let's say for other models.  Sorry for the dumb question!  Any tips appreciated (maybe put that in the docs too?)

BenjaminBossan (approved):
Thanks for the fix, LGTM

pacman100:
Hello @hamelsmu ,

> How do you figure out the right value for this config? Let's say for other models. Sorry for the dumb question! Any tips appreciated (maybe put that in the docs too?)

It is mentioned in the docs for transformers here https://huggingface.co/docs/transformers/main_classes/trainer#pytorch-fully-sharded-data-parallel, will raise a PR to have those in Accelerate. Specifically the below point:

> For transformer based auto wrap policy, it is recommended to specify `fsdp_transformer_layer_cls_to_wrap` in the config file. If not specified, the default value is `model._no_split_modules` when available. This specifies the list of transformer layer class name (case-sensitive) to wrap ,e.g, BertLayer, GPTJBlock, T5Block … This is important because submodules that share weights (e.g., embedding layer) should not end up in different FSDP wrapped units. Using this policy, wrapping happens for each block containing Multi-Head Attention followed by couple of MLP layers. Remaining layers including the shared embeddings are conveniently wrapped in same outermost FSDP unit. Therefore, use this for transformer based models.

Issue #1950, opened 2023-09-11
save_state multi-process/node with shared filesystem


zeyugao:
### System Info

```Shell
- `Accelerate` version: 0.22.0
- Platform: Linux-3.10.0-1160.83.1.el7.x86_64-x86_64-with-glibc2.35
- Python version: 3.11.4
- Numpy version: 1.25.2
- PyTorch version (GPU?): 2.0.1 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 1007.34 GB
- GPU type: NVIDIA A800-SXM4-80GB
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [X] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am doing a distributed training on multiple-node with kubeflow, setting the output_dir to a shared filesystem (pfs) across all nodes.

But when trying to call `save_state` in all the processes in all nodes, sometimes it will yield this error

<img width="899" alt="image" src="https://github.com/huggingface/accelerate/assets/6374697/5cf2659e-17c2-4d31-a305-fe0881de8649">

That is the following lines errored, which means there are some race in the codes

https://github.com/huggingface/accelerate/blob/cdb001ca5f9be875034ddd0aa86a542c182782fe/src/accelerate/accelerator.py#L2705-L2706

But some issues mentioned before, like https://github.com/huggingface/accelerate/issues/1171 . It says that there is no need to check for `is_main_process` when using `save_state`.

But after checking the code from accelerate, it seems that it doesn't handle the multi-process/node issue correctly?

In model saving, if we use deepspeed, it will handle the multi-node issue correctly and requires all process calling this function.

https://github.com/microsoft/DeepSpeed/blob/542dc0d5cb58d9aba78eceb4510296eddbaa2e1f/deepspeed/runtime/engine.py#L2965-L2980

But when it processes the other parameters, it seems that there is no check of the current process index at all, resulting possible race condition.  The weight should be save only at main process or the file should be different.

Like previous `save_custom_state`, no check is performed

https://github.com/huggingface/accelerate/blob/cdb001ca5f9be875034ddd0aa86a542c182782fe/src/accelerate/checkpointing.py#L200-L207

### Expected behavior

[...]

PR #1953, opened 2023-09-11
Enable shared file system with `save` and `save_state` via ProjectConfiguration

muellerzr:
# What does this PR do?

Lets `save` and `save_state` rely on a similar system to deepspeed, and to ensure we're not overloading the `Accelerator` with arguments delegating it to the `ProjectConfiguration`

Fixes # (issue)

Solves https://github.com/huggingface/accelerate/issues/1950

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @pacman100

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2731,10 +2731,10 @@ class Accelerator:
             schedulers,
             self.state.process_index,
             self.scaler,
-            self.project_configuration.save_on_each_node,
+            save_on_each_node=self.project_configuration.save_on_each_node,
         )
         for i, obj in enumerate(self._custom_objects):
-            save_custom_state(obj, output_dir, i, self.project_configuration.save_on_each_node)
+            save_custom_state(obj, output_dir, i, save_on_each_node=self.project_configuration.save_on_each_node)
         self.project_configuration.iteration += 1
         return save_location
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2418,8 +2418,8 @@ class Accelerator:
             f (`str` or `os.PathLike`): Where to save the content of `obj`.
 
         Note:
-            If `use_node_global_storage` was passed in as a `ProjectConfiguration`, will only save on the global main
-            process and assume a shared file system.
+            If `save_on_each_node` was passed in as a `ProjectConfiguration`, will save the object once per node,
+            rather than only once on the main node.
 
         Example:
 
@@ -2431,7 +2431,7 @@ class Accelerator:
         >>> accelerator.save(arr, "array.pkl")
         ```
         """
-        save(obj, f, self.project_configuration.use_node_global_storage)
+        save(obj, f, self.project_configuration.save_on_each_node)
 
     def save_model(
         self,
@@ -2710,10 +2710,10 @@ class Accelerator:
             schedulers,
             self.state.process_index,
             self.scaler,
-            self.project_configuration.use_node_global_storage,
+            self.project_configuration.save_on_each_node,
         )
         for i, obj in enumerate(self._custom_objects):
-            save_custom_state(obj, output_dir, i)
+            save_custom_state(obj, output_dir, i, self.project_configuration.save_on_each_node)
         self.project_configuration.iteration += 1
         return save_location
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2447,7 +2447,12 @@ class Accelerator:
         >>> accelerator.save(arr, "array.pkl")
         ```
         """
-        save(obj, f, save_on_each_node=self.project_configuration.save_on_each_node, safe_serialization=safe_serialization)
+        save(
+            obj,
+            f,
+            save_on_each_node=self.project_configuration.save_on_each_node,
+            safe_serialization=safe_serialization,
+        )
 
     def save_model(
         self,
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2418,8 +2418,8 @@ class Accelerator:
             f (`str` or `os.PathLike`): Where to save the content of `obj`.
 
         Note:
-            If `use_node_local_storage` was passed in as a `ProjectConfiguration`, will only save on the main process
-            and assume a shared file system.
+            If `use_node_global_storage` was passed in as a `ProjectConfiguration`, will only save on the global main
+            process and assume a shared file system.
 
         Example:
 
@@ -2431,7 +2431,7 @@ class Accelerator:
         >>> accelerator.save(arr, "array.pkl")
         ```
         """
-        save(obj, f, self.project_configuration.use_node_local_storage)
+        save(obj, f, self.project_configuration.use_node_global_storage)
 
     def save_model(
         self,
@@ -2710,7 +2710,7 @@ class Accelerator:
             schedulers,
             self.state.process_index,
             self.scaler,
-            self.project_configuration.use_node_local_storage,
+            self.project_configuration.use_node_global_storage,
         )
         for i, obj in enumerate(self._custom_objects):
             save_custom_state(obj, output_dir, i)
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2417,6 +2417,10 @@ class Accelerator:
             obj (`object`): The object to save.
             f (`str` or `os.PathLike`): Where to save the content of `obj`.
 
+        Note:
+            If `use_node_local_storage` was passed in as a `ProjectConfiguration`, will only save on the main process
+            and assume a shared file system.
+
         Example:
 
         ```python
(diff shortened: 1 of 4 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -137,14 +137,13 @@ def save_fsdp_optimizer(fsdp_plugin, accelerator, optimizer, model, output_dir,
...
-                output_optimizer_file = os.path.join(output_dir, optim_state_name)
-                logger.info(f"Saving Optimizer state to {output_optimizer_file}")
-                torch.save(optim_state, output_optimizer_file)
-                logger.info(f"Optimizer state saved in {output_optimizer_file}")
+            optim_state_name = (
+                f"{OPTIMIZER_NAME}.bin" if optimizer_index == 0 else f"{OPTIMIZER_NAME}_{optimizer_index}.bin"
+            )
+            output_optimizer_file = os.path.join(output_dir, optim_state_name)
+            logger.info(f"Saving Optimizer state to {output_optimizer_file}")
+            save(optim_state, output_optimizer_file)
```
BenjaminBossan, line 145: Same comment as above.

BenjaminBossan (commented):
Thanks for taking on this issue. I did a grep on the code base (as probably did you) and this seems to catch all relevant uses of `torch.save`.

While in general I agree that the PR looks "more correct" than the existing code, I do wonder if this change is sufficient. With the PR merged, when a user calls `save_accelerator_state`, they will only save the RNG state of process index 0. However, when loading, we still try to load the RNG state of each index, resulting in a crash. Do I see that correctly?

A similar argument applies to `load_custom_state`.

If I did not misunderstand, it would mean that we need to fix the loading functions as well to load only from index 0. But that also seems to be problematic, because it would mean that if a user upgrades accelerate and try to load a checkpoint from the previous version, they would get different results.

Please LMK if I understood the issue correctly.

zeyugao:
I think that there are more needed for consideration, possibly with a option like `use_node_local_storage` similar to deepspeed (https://github.com/microsoft/DeepSpeed/blob/542dc0d5cb58d9aba78eceb4510296eddbaa2e1f/deepspeed/runtime/engine.py#L2985), cause not all users doing distributed training have a shared filesystem. This matters cause we need to take care of loading checkpoint.

The `save` function only check for `local_process_index`. But for example, in `save_custom_state` I mentioned in the original issue, it should check for the `process_index` instead of `local_process_index` only.

src/accelerate/checkpointing.py
```diff
@@ -204,7 +204,7 @@ def save_custom_state(obj, path, index: int = 0):
     # Should this be the right way to get a qual_name type value from `obj`?
     save_location = Path(path) / f"custom_checkpoint_{index}.pkl"
     logger.info(f"Saving the state of {get_pretty_name(obj)} to {save_location}")
-    torch.save(obj.state_dict(), save_location)
+    save(obj.state_dict(), save_location)
```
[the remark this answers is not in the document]
zeyugao: The `index` here means the index of the `self._custom_objects`, not the process_index

muellerzr:
@zeyugao a better solution then is to have it as part of the `ProjectConfiguration`. I'll do a hard reset on the changes here (as there's a few issues with it as mentioned by @BenjaminBossan) and we'll start small with that change + clear docs

muellerzr:
@zeyugao I think that's more what you're looking for? :)

src/accelerate/utils/other.py
```diff
@@ -107,17 +107,20 @@ def wait_for_everyone():
...
         obj: The data to save
         f: The file (or file-like object) to use to save the data
+        global_only: Whether to only save on the global main process
     """
     if PartialState().distributed_type == DistributedType.TPU:
         xm.save(obj, f)
-    elif PartialState().local_process_index == 0:
+    elif PartialState().is_main_process and global_only:
+        torch.save(obj, f)
+    elif PartialState().is_local_main_process:
```
zeyugao, line 123: It should be `is_local_main_process and not global_only`?

This will equal to true if `is_local_main_process = True` even `global_only = False`.

muellerzr: Yes. Pushed

src/accelerate/accelerator.py
```diff
@@ -2700,7 +2704,13 @@ def _inner(folder):
...
+            output_dir,
+            weights,
+            optimizers,
+            schedulers,
+            self.state.process_index,
+            self.scaler,
+            self.project_configuration.use_node_global_storage,
         )
         for i, obj in enumerate(self._custom_objects):
             save_custom_state(obj, output_dir, i)
```
zeyugao, line 2716: This `save_custom_state` is missed to take care of :)

src/accelerate/utils/dataclasses.py
```diff
@@ -415,6 +415,13 @@ class ProjectConfiguration:
         metadata={"help": "The current save iteration."},
     )
 
+    use_node_global_storage: bool = field(
+        default=False,
```
zeyugao, line 419: One more point, the behavior of deepspeed is that it expects that a multi-node environment uses a shared storage (https://www.deepspeed.ai/docs/config-json/#checkpoint-options & https://huggingface.co/docs/transformers/main_classes/deepspeed#use-of-nonshared-filesystem). We can take this behavior as default.

muellerzr: Agreed, we'll also change this to be `save_on_each_node` to align with `TrainingArguments`: https://github.com/huggingface/transformers/blob/v4.33.0/src/transformers/training_args.py#L805-L813

src/accelerate/utils/other.py
```diff
@@ -109,22 +110,27 @@ def wait_for_everyone():
...
+        obj:
+            The data to save
+        f:
+            The file (or file-like object) to use to save the data
+        save_on_each_node (`bool`, *optional*, defaults to `False`):
+            Whether to only save on the global main process
+        safe_serialization (`bool`, *optional*, defaults to `False`):
+            Whether to save `obj` using `safetensors`
+    """
+    save_func = torch.save if not safe_serialization else partial(safe_save_file, metadata={"format": "pt"})
```
BenjaminBossan, line 127: Elegant.

src/accelerate/accelerator.py
```diff
@@ -2716,10 +2725,16 @@ def _inner(folder):
...
+            weights,
+            optimizers,
+            schedulers,
+            self.state.process_index,
+            self.scaler,
+            self.project_configuration.save_on_each_node,
         )
         for i, obj in enumerate(self._custom_objects):
-            save_custom_state(obj, output_dir, i)
+            save_custom_state(obj, output_dir, i, self.project_configuration.save_on_each_node)
```
BenjaminBossan, line 2737: Here and below, I'd personally prefer to pass `save_on_each_node` as a keyword argument.

src/accelerate/utils/dataclasses.py
```diff
@@ -415,6 +415,16 @@ class ProjectConfiguration:
         metadata={"help": "The current save iteration."},
     )
 
+    save_on_each_node: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                "When doing multi-node distributed training, whether to save models and checkpoints on each node, or"
+                " only on the main one"
```
BenjaminBossan, line 423: If this is something that can be left as default for every setting except for DeepSpeed, maybe it's worth mentioning here. If such a blanket statement cannot be made, ignore my comment.

BenjaminBossan (approved):
Thanks a lot, Zach, for working on this and doing a whole fresh start with the implementation. Overall, this LGTM, my comments are just nits.

Regarding the initial problem with persistence on a shared file system, I guess it's going to be pretty hard to write a test for that :) However, if @zeyugao can confirm that this fix solves the issue, that is enough in my opinion.

SunMarc (approved):
LGTM ! maybe pass the `save_on_each_node` as keyword arg when calling `save()`

zeyugao:
In my test, it doesn't lead to race over a 12 hours training. So it seems that it is correctly handled. Lots of thanks!

src/accelerate/utils/other.py
```diff
@@ -109,22 +110,27 @@ def wait_for_everyone():
...
     if PartialState().distributed_type == DistributedType.TPU:
         xm.save(obj, f)
-    elif PartialState().local_process_index == 0:
-        if safe_serialization:
-            safe_save_file(obj, f, metadata={"format": "pt"})
-        else:
-            torch.save(obj, f)
+    elif PartialState().is_main_process and save_on_each_node:
```
pacman100, line 130:
```suggestion
    elif PartialState().is_main_process and not save_on_each_node:
```

src/accelerate/utils/other.py
```diff
@@ -109,22 +110,27 @@ def wait_for_everyone():
...
+        save_func(obj, f)
+    elif PartialState().is_local_main_process and not save_on_each_node:
```
pacman100, line 132:
```suggestion
    elif PartialState().is_local_main_process and save_on_each_node:
```

pacman100 (commented):
Thank you @muellerzr for fixing the issues with saving state, let couple of comments.

src/accelerate/accelerator.py
muellerzr: Done

PR #1954, opened 2023-09-11
fix safetensor saving

SunMarc:
# What does this PR do ?
This PR fixes the `save_model` function for safetensors. It was never tested so it had a couple of problems.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -141,9 +141,6 @@ try:
 except ImportError:
     from torch.optim.lr_scheduler import _LRScheduler as LRScheduler
 
-if is_safetensors_available():
-    from safetensors.torch import save_file as safe_save_file
-
 logger = get_logger(__name__)
 
 
@@ -2412,13 +2409,14 @@ class Accelerator:
         for tracker in self.trackers:
             tracker.finish()
 
-    def save(self, obj, f):
+    def save(self, obj, f, safe_serialization=False):
         """
         Save the object passed to disk once per machine. Use in place of `torch.save`.
 
         Args:
             obj (`object`): The object to save.
             f (`str` or `os.PathLike`): Where to save the content of `obj`.
+            safe_serialization (`bool`, *optional*, defaults to `False`): Whether to save `obj` using `safetensors`
 
         Example:
 
@@ -2430,7 +2428,7 @@ class Accelerator:
         >>> accelerator.save(arr, "array.pkl")
         ```
         """
-        save(obj, f)
+        save(obj, f, safe_serialization)
 
     def save_model(
         self,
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -2538,7 +2541,10 @@ def save_model(
 
         # Save the model
         for shard_file, shard in shards.items():
-            self.save(shard, os.path.join(save_directory, shard_file))
+            if safe_serialization:
+                safe_save_file(shard, os.path.join(save_directory, shard_file), metadata={"format": "pt"})
```
muellerzr, line 2545: Do we know for certain this only happens in `process_index==0`? Otherwise I'd rather see `self.save` be modified (which is really `utils.other.save`) to use `safe_serialize_file` instead.

muellerzr (commented):
Thanks! This seems like a good improvement, but we need to be careful on process indexes. Gave a recommendation, or you can directly check `self.is_main_process`

src/accelerate/accelerator.py
SunMarc: I guess it is safer to modify `self.save` instead. Modified !

src/accelerate/accelerator.py
```diff
@@ -2427,7 +2428,7 @@ def save(self, obj, f):
         >>> accelerator.save(arr, "array.pkl")
         ```
         """
-        save(obj, f)
+        save(obj, f, safe_serialization)
```
BenjaminBossan, line 2431: Let's pass `safe_serialization` as keyword arg.

src/accelerate/accelerator.py
```diff
@@ -2507,7 +2508,7 @@ def save_model(
                             del state_dict[name]
                             warn_names.add(name)
             if len(warn_names) > 0:
-                logger.warning_once(
+                logger.warning(
```
BenjaminBossan, line 2511: Why was this changed?

src/accelerate/accelerator.py
```diff
@@ -2538,7 +2539,7 @@ def save_model(
 
         # Save the model
         for shard_file, shard in shards.items():
-            self.save(shard, os.path.join(save_directory, shard_file))
+            self.save(shard, os.path.join(save_directory, shard_file), safe_serialization)
```
BenjaminBossan, line 2542: Let's pass `safe_serialization` as keyword arg.

BenjaminBossan (approved):
Looks pretty good, thanks for the addition. I would prefer the new argument to passed as keyword argument, as more arguments could be added in the future and we shouldn't rely on order (see #1953). Feel free to merge once this is changed, or to disagree :)

Apart from that, only a small question.

src/accelerate/accelerator.py
SunMarc: it was copy pasted from transformers but logger in accelerate don't have the `warning_once` function

src/accelerate/accelerator.py
SunMarc: sounds good !

src/accelerate/accelerator.py
SunMarc: yes !

muellerzr (approved):
Great work! :)

BenjaminBossan:
Thanks for implementing the suggestions, LGTM.

PR #1957, opened 2023-09-12
[`Tests`] Finish all todos

younesbelkada:
# What does this PR do?

As per title ! 
Some TODOs were left for the next transformers release, this PR fixes them

note for the

```bash
remove the positional arg on the next `transformers` release
```

task it is irrelevant, we still need to use positional arguments for `replace_with_bnb_linear` method, check the signature of that method: https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/bitsandbytes.py#L184

cc @muellerzr

### DIFF
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -638,22 +638,16 @@ class BigModelingTester(unittest.TestCase):
         with init_empty_weights():
             model = AutoModel.from_config(AutoConfig.from_pretrained("bigscience/bloom-560m"))
 
-        # TODO: @younesbelkada remove the positional arg on the next `transformers` release
         quantization_config = BitsAndBytesConfig(load_in_8bit=True)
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks for the cleanup!

BenjaminBossan (approved):
Thanks for working on this, well done marking all relevant spots beforehand.

I only have a question, I don't understand the comment:

> we still need to use positional arguments for replace_with_bnb_linear method

I don't understand how that follows from the signature.

younesbelkada:
Thanks for your reviews! 
I think my goal was to be able to call `replace_with_bnb_linear` without having to pass explicit pos arguments. Indeed you can still call `replace_with_bnb_linear` with no explicit pos arguments but I'll have to pass a dummy argument for `current_key_name`. So I just preferred to keep it as it is currently!

PR #1959, opened 2023-09-12
Protect tensorflow dependency

SunMarc:
# What does this PR do ? 
This PR protect the tensorflow dependency, so that we import tensorflow only when it is needed ! This should speed-up the import by ~2s. See similar PR for [transformers](https://github.com/huggingface/transformers/pull/26106#issuecomment-1715318226). 

To get the dashboard : 
```py
pip install tuna
python -X importtime -c "import accelerate" 2> accelerate-import-profile.log
tuna accelerate-import-profile.log
```

cc @younesbelkada

Screenshot 2023-09-12 at 11 38 46 AM

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -39,11 +39,6 @@ from .utils import (
 _available_trackers = []
 
 if is_tensorboard_available():
-    try:
-        from torch.utils import tensorboard
-    except ModuleNotFoundError:
-        import tensorboardX as tensorboard
-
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
> This should speed-up the import by ~2s

Do you have a command that you used to test this? Just `time python -c "from accelerate import Accelerator"`?

SunMarc:
> > This should speed-up the import by ~2s
> 
> Do you have a command that you used to test this? Just `time python -c "from accelerate import Accelerator"`?

I've updated the description with the commands to get the dashboard !

PR #1963, opened 2023-09-12
Better guards for slow imports

muellerzr:
# What does this PR do?

This PR guards slower imports better, namely the trackers and especially `deepspeed` across the framework.

Alternative to https://github.com/huggingface/accelerate/pull/1961, unless we decide that it's getting too repetitive and we want a wrapper 

Fixes # (issue)

Resolves https://github.com/huggingface/accelerate/issues/1952

## Speedup

Before, with DeepSpeed installed too:

image

After: 
image

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@SunMarc @BenjaminBossan

PR #1964, opened 2023-09-12
Rm strtobool

muellerzr:
# What does this PR do?

As `strtobool` will be removed in the next python version (3.12), this PR re-implements it for what we care about

Tags along with https://github.com/huggingface/accelerate/pull/1963 to get us to ~.9s for maximum efficiency too 🚀 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/utils/environment.py
+++ b/src/accelerate/utils/environment.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 import os
-from distutils.util import strtobool
+
+from .other import str_to_bool
 
 
 def get_int_from_env(env_keys, default):
@@ -28,7 +29,7 @@ def get_int_from_env(env_keys, default):
 def parse_flag_from_env(key, default=False):
     """Returns truthy value for `key` from the env if available else the default."""
     value = os.environ.get(key, str(default))
-    return strtobool(value) == 1  # As its name indicates `strtobool` actually returns an int...
+    return str_to_bool(value) == 1  # As its name indicates `str_to_bool` actually returns an int...
 
 
 def parse_choice_from_env(key, default="no"):
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -20,7 +20,6 @@ import sys
 import tempfile
 import unittest
 from contextlib import contextmanager
-from distutils.util import strtobool
 from functools import partial
 from pathlib import Path
 from typing import List, Union
@@ -44,6 +43,7 @@ from ..utils import (
     is_transformers_available,
     is_wandb_available,
     is_xpu_available,
+    str_to_bool,
 )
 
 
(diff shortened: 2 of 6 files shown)


src/accelerate/utils/environment.py
```diff
@@ -13,7 +13,21 @@
 # limitations under the License.
 
 import os
-from distutils.util import strtobool
+
+
+def str_to_bool(value) -> bool:
```
BenjaminBossan, line 18: Just curious why the same implementation is not used as for `strtobool`? The latter is a bit faster, but it shouldn't matter in practice.

src/accelerate/utils/environment.py
```diff
@@ -13,7 +13,21 @@
...
+
+    True values are `y`, `yes`, `t`, `true`, `on`, and `1`; False value are `n`, `no`, `f`, `false`, `off`, and `0`;
+    """
+    truth = ["y", "yes", "t", "true", "on", "1"]
+    false = ["n", "no", "f", "false", "off", "0"]
+
+    if value.lower() not in truth + false:
+        raise ValueError(f"{value} is not a valid boolean value")
+
+    return value.lower() in truth
```
BenjaminBossan, line 30: Is there a specific reason to change the signature from int to bool? `str_to_bool(value) == 1` still works since bool is a subclass of int, but it can be confusing.

src/accelerate/utils/environment.py
```diff
@@ -28,7 +42,7 @@ def get_int_from_env(env_keys, default):
 def parse_flag_from_env(key, default=False):
     """Returns truthy value for `key` from the env if available else the default."""
     value = os.environ.get(key, str(default))
-    return strtobool(value) == 1  # As its name indicates `strtobool` actually returns an int...
+    return str_to_bool(value) == 1  # As its name indicates `str_to_bool` actually returns an int...
```
BenjaminBossan, line 45: This comment, which I assume is meant ironically, could be removed if we actually return a bool. But technically, a bool is an int, so not sure if it's really meant to be ironic :D

BenjaminBossan (commented):
Oh wow, nice catch, I didn't know that `strtobool` was deprecated. There doesn't seem to be any warning when importing it, who knows how many upgrades will fail when people switch to Python 3.12.

I like the decision to change the name, that way there can be no accidental usage of the previous function.

I only have one question, which is about the decision to return a bool instead of int now. I agree it makes more sense (maybe `strtobool` predates the existence of bools in Python), but it makes all the `if str_to_bool(...) == 0` calls a bit awkward.

src/accelerate/utils/environment.py
muellerzr: Reverted

BenjaminBossan (approved):
LGTM, thanks. Well done catching this deprecation.

PR #1971, opened 2023-09-13
Fix model copy after `dispatch_model`

### LINKED ISSUE #1835: Can't copy model after dispatch_model called
### System Info

```Shell
- `Accelerate` version: 0.20.3
- Platform: Linux-5.15.0-70-generic-x86_64-with-glibc2.35
- Python version: 3.10.6
- Numpy version: 1.22.2
- PyTorch version (GPU?): 2.1.0a0+4136153 (True)
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

Reproduce issue with script:
```python
import torch
from torch import nn
import accelerate
import copy

class MyModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc1 = nn.Linear(1,10)
     
    def forward(self, x):
        x = self.fc1(x)

        print(f'In forward pass of model with id: {self.id}')

        return x
    

inp = torch.rand((1,1))
model = MyModel()
model.id = 1
model = accelerate.dispatch_model(model, device_map={'fc1': 'cpu'})

print(f'Output from first model:')
model(inp) # prints id 1 as expected

model_copy = copy.deepcopy(model)
model_copy.id = 2
print(f'Output from copied model:')
model_copy(inp) # incorrectly prints id of 1, when id of 2 was expected
```

The output of this script:
```
Output from first model:
In forward pass of model with id: 1
Output from copied model:
In forward pass of model with id: 1
```

### Expected behavior

Expected output:
```
Output from first model:
In forward pass of model with id: 1
Output from copied model:
In forward pass of model with id: 2
```

When a model is copied, the Accelerate hooks copied over to the new model incorrectly reference the original model. The source of the issue is in `add_hook_to_module` in `hooks.py`
https://github.com/huggingface/accelerate/blob/058a3546ea588abf57de3705409c250cc52423e7/src/accelerate/hooks.py#L158-L168
The issue is that the hooks overwrite the forward method of the model, but do not bind that forward method to the specific class instance. Instead, the `new_forward` method references `module` which is defined outside of the new function. Because of this, when calling the copied version of the model, it references the version of `new_forward` that calls the forward method of the original model (rather than the copied model as expected), leading to the incorrect ID being printed out.

One proposed solution is to the replace the above code block with:
```python
    @functools.wraps(old_forward)
    def new_forward(self, *args, **kwargs):
        args, kwargs = self._hf_hook.pre_forward(self, *args, **kwargs)
        if self._hf_hook.no_grad:
            with torch.no_grad():
                output = self._old_forward(*args, **kwargs)
        else:
            output = self._old_forward(*args, **kwargs)
        return self._hf_hook.post_forward(self, output)

    module.forward = new_forward.__get__(module) # second param to __get__ is optional
```

In this modified version, we bind the forward function to the specific class instance, which means that when we call the forward on the copied model, it now properly calls the forward method of the copied model and outputs the correct ID.

New output (matching expected):
```
Output from first model:
In forward pass of model with id: 1
Output from copied model:
In forward pass of model with id: 2
```

The general motivation for this issue is that `torch.compile` functionality relies on `copy.deepcopy` to copy a model with fake tensor model enabled:
https://github.com/pytorch/pytorch/blob/bb6b157458a34d8b3499932035381fdb12683703/torch/_dynamo/utils.py#L915-L917

Currently, due to the model forward not operating correctly when copied, the `torch.compile` Dynamo tracing fails to properly trace the model. The tracing goes through the original model's forward function with real tensor values from the original model and not the model that is copied with fake tensors, leading to an exception.

Addressing this issue would help step towards being able to use Accelerate to distribute a model and then apply `torch.compile` functionality to that distributed model. I'm also eager to hear if there are alternative solutions to this issue!

austinapatel:
# What does this PR do?

Fixes #1835 where a model dispatched with `dispatch_model` was unable to be properly copied with `copy.deepcopy`. Specifically, the copied model would incorrectly still reference the align device hook forward function of the original model, rather than its own forward function. This PR includes a fix to this issue as well as a test to validate the fix.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @sgugger

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -156,16 +156,17 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
     module._hf_hook = hook
 
     @functools.wraps(old_forward)
-    def new_forward(*args, **kwargs):
-        args, kwargs = module._hf_hook.pre_forward(module, *args, **kwargs)
-        if module._hf_hook.no_grad:
+    def new_forward(self, *args, **kwargs):
+        args, kwargs = self._hf_hook.pre_forward(self, *args, **kwargs)
+        if self._hf_hook.no_grad:
             with torch.no_grad():
-                output = old_forward(*args, **kwargs)
+                output = self._old_forward(*args, **kwargs)
         else:
-            output = old_forward(*args, **kwargs)
-        return module._hf_hook.post_forward(module, output)
+            output = self._old_forward(*args, **kwargs)
+        return self._hf_hook.post_forward(self, output)
 
-    module.forward = new_forward
+    module.forward = new_forward.__get__(module)
+
     return module
 
 
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -155,7 +155,6 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
     module = hook.init_hook(module)
     module._hf_hook = hook
 
-    @functools.wraps(old_forward)
     def new_forward(self, *args, **kwargs):
         args, kwargs = self._hf_hook.pre_forward(self, *args, **kwargs)
         if self._hf_hook.no_grad:
@@ -165,7 +164,7 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
             output = self._old_forward(*args, **kwargs)
         return self._hf_hook.post_forward(self, output)
 
-    module.forward = new_forward.__get__(module)
+    module.forward = functools.update_wrapper(functools.partial(new_forward, module), old_forward)
 
     return module
 
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -155,14 +155,14 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
     module = hook.init_hook(module)
     module._hf_hook = hook
 
-    def new_forward(self, *args, **kwargs):
-        args, kwargs = self._hf_hook.pre_forward(self, *args, **kwargs)
-        if self._hf_hook.no_grad:
+    def new_forward(module, *args, **kwargs):
+        args, kwargs = module._hf_hook.pre_forward(module, *args, **kwargs)
+        if module._hf_hook.no_grad:
             with torch.no_grad():
-                output = self._old_forward(*args, **kwargs)
+                output = module._old_forward(*args, **kwargs)
         else:
-            output = self._old_forward(*args, **kwargs)
-        return self._hf_hook.post_forward(self, output)
+            output = module._old_forward(*args, **kwargs)
+        return module._hf_hook.post_forward(module, output)
 
     module.forward = functools.update_wrapper(functools.partial(new_forward, module), old_forward)
 
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import copy
 import os
 import unittest
 from tempfile import TemporaryDirectory
@@ -43,6 +43,18 @@ class ModelForTest(nn.Module):
 
     def forward(self, x):
         return self.linear2(self.batchnorm(self.linear1(x)))
+
+
+class ModelForTestCopy(nn.Module):
+    def __init__(self, id: int):
+        super().__init__()
+        self.id = id
+        self.linear1 = nn.Linear(3, 4)
+        self.batchnorm = nn.BatchNorm1d(4)
+        self.linear2 = nn.Linear(4, 5)
+
+    def forward(self, x):
+        return self.linear2(self.batchnorm(self.linear1(x))), self.id
 
 
 class ModelForTestTiedWeights(nn.Module):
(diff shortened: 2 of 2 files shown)


src/accelerate/hooks.py
```diff
@@ -155,17 +155,17 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
     module = hook.init_hook(module)
     module._hf_hook = hook
 
-    @functools.wraps(old_forward)
-    def new_forward(*args, **kwargs):
-        args, kwargs = module._hf_hook.pre_forward(module, *args, **kwargs)
-        if module._hf_hook.no_grad:
+    def new_forward(self, *args, **kwargs):
```
SunMarc, line 158: Maybe replace `self` by `module` to not use the syntax of a class.

tests/test_big_modeling.py
```diff
@@ -325,6 +337,24 @@ def test_dispatch_model_multi_gpu(self):
...
+        expected, original_output_id = original_model(x)
+
+        dispatch_model(original_model, device_map)
+
+        copied_model = copy.deepcopy(original_model)
+        copied_model.id = 2
+        output, copied_output_id = copied_model(x)
+
+        self.assertEqual(original_model.id, original_output_id)
+        self.assertEqual(copied_model.id, copied_output_id)
```
SunMarc, line 355: Maybe we can add a test that check if the forward is not the same 
```suggestion
        self.assertEqual(copied_model.id, copied_output_id)
        self.assertFalse(copied_model.linear1.forward is original_model.linear1.forward)
```

src/accelerate/hooks.py
```diff
@@ -155,17 +155,17 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
...
             with torch.no_grad():
-                output = old_forward(*args, **kwargs)
+                output = self._old_forward(*args, **kwargs)
         else:
-            output = old_forward(*args, **kwargs)
-        return module._hf_hook.post_forward(module, output)
+            output = self._old_forward(*args, **kwargs)
+        return self._hf_hook.post_forward(self, output)
+
+    module.forward = functools.update_wrapper(functools.partial(new_forward, module), old_forward)
```
SunMarc, line 167: we can use still use the decorator `@wraps(old_forward)` instead of `update_wrapper` but both are okay to me.

SunMarc (approved):
Hi @austinapatel, thanks a lot for fixing this ! Just a few nits :) Can you have a second look @BenjaminBossan ?

src/accelerate/hooks.py
austinapatel: Updated!

tests/test_big_modeling.py
austinapatel: Added!

src/accelerate/hooks.py
austinapatel: If I change to using `@wraps(old_forward)` instead, I end up failing these checks here: https://github.com/huggingface/accelerate/blob/a87c95da9e3b416fb10a0e7dac7d397c015c3ed5/tests/test_hooks.py#L64-L65
This happens since `functools.partial` will happen after `@wraps(forward_function)` occurs, so the signature of the forward function ends up being incorrect, which is why I went with `update_wrapper`.

austinapatel:
Thanks for the review @SunMarc! I have pushed changes per your feedback!

src/accelerate/hooks.py
```diff
@@ -155,17 +155,17 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
...
             with torch.no_grad():
-                output = old_forward(*args, **kwargs)
+                output = module._old_forward(*args, **kwargs)
         else:
-            output = old_forward(*args, **kwargs)
+            output = module._old_forward(*args, **kwargs)
         return module._hf_hook.post_forward(module, output)
 
-    module.forward = new_forward
+    module.forward = functools.update_wrapper(functools.partial(new_forward, module), old_forward)
```
BenjaminBossan, line 167: I wonder about this line versus:

```python
module.forward = MethodType(functools.update_wrapper(new_forward, old_forward), module)
```

@austinapatel Do you think that would be preferable? I thought it might affect pickling, but unfortunately, both variants don't allow it, but the error messages are different.

BenjaminBossan (commented):
In general, this LGTM. I only have one question, apart from that this can be merged.

src/accelerate/hooks.py
austinapatel: Thanks for the suggestion and review! I just tried your code snippet, and I run into the same issue as this comment: https://github.com/huggingface/accelerate/pull/1971#discussion_r1326765316 since `functools.update_wrapper` needs to be called last (`test_add_and_remove_hooks` in `test_hooks.py` fails). If I try `module.forward = functools.update_wrapper(MethodType(new_forward, module), old_forward)` I get the following error: `AttributeError: 'method' object has no attribute '__module__'` when calling `update_wrapper`. My understanding is that `update_wrapper` is now being called on a method rather than on a function, which doesn't play as nicely with `update_wrapper`.

BenjaminBossan: Thanks for explaining and testing, this makes sense. I'm a bit wary about binding the module via `partial` to basically imitate a bound method, but it seems that all other solutions don't work, so I guess it's the best we can do.

BenjaminBossan (approved):
Thanks so much @austinapatel for all the work you put into this.

austinapatel:
Thanks @SunMarc and @BenjaminBossan for the reviews and feedback!

PR #1972, opened 2023-09-14
Flex fix patch for accelerate

abhilash1910:
Fix for flex: https://github.com/huggingface/transformers/pull/26135

cc @muellerzr


muellerzr (approved):
Thanks for the check and enable!

SunMarc (approved):
LGTM !

PR #1974, opened 2023-09-14
Remove checkpoints only on main process

Kepnu4:
shutil.rmtree might throw errors if called on multiply processes. Remove checkpoints only on main process


muellerzr:
Hi @Kepnu4, can you do `make style; make quality` real quick? We're making a release today and it'd be great to have this in there!

Kepnu4:
> Hi @Kepnu4, can you do `make style; make quality` real quick? We're making a release today and it'd be great to have this in there!

Sure, done

muellerzr (approved):
This makes sense to me, thanks for the fix!

SunMarc (approved):
LGTM !

PR #1994, opened 2023-09-20
Safely end training even if trackers weren't initialized

Ben-Epstein:
# What does this PR do?

If you create an instance of `Accelerator()` without logging to a tracker, and then call `end_training` (which would be common in development/debugging), the accelerator shouldn't crash if there aren't trackers. Since `self.trackers` isn't initialized in the `__init__` of the class, this is possible

## Before submitting
      Pull Request section?
      to it if that's the case.

Tested by running the following
```
>>> a = Accelerator()
>>> a.end_training()
```

Made sure this wasn't thrown 
```
  File "<stdin>", line 1, in <module>
  File ".venv/lib/python3.10/site-packages/accelerate/accelerator.py", line 615, in _inner
    return PartialState().on_main_process(function)(*args, **kwargs)
  File ".venv/lib/python3.10/site-packages/accelerate/accelerator.py", line 2499, in end_training
    for tracker in self.trackers:
AttributeError: 'Accelerator' object has no attribute 'trackers'
```

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2496,6 +2496,8 @@ class Accelerator:
         >>> accelerator.end_training()
         ```
         """
+        if not hasattr(self, "trackers"):
+            return
         for tracker in self.trackers:
             tracker.finish()
 


BenjaminBossan:
I wonder if the better solution wouldn't be to set `self.trackers = []` inside of `Accelerator.__init__` instead of `init_trackers`. Within `init_trackers`, we could call `self.trackers.clear()` for consistency.

Ben-Epstein:
> I wonder if the better solution wouldn't be to set `self.trackers = []` inside of `Accelerator.__init__` instead of `init_trackers`. Within `init_trackers`, we could call `self.trackers.clear()` for consistency.

@BenjaminBossan i totally agree, and in fact wrote that first, but I thought there was likely a good reason why that wasn't done in the first place. It's uncommon _not_ to initialize all member variables on `__init__` (in fact i'm pretty sure mypy will complain), so I assumed it was purposeful.

Do you think I should go ahead with that?

BenjaminBossan:
You make a good point that there could be a non-obvious reason. If it's okay to you, we could wait for Zach to comment on this (which may take a week or so).

Ben-Epstein:
Hi @muellerzr updated!

muellerzr (approved):
Thanks for the fix! :)

BenjaminBossan (commented):
Thanks for updating the PR.

I think now that `trackers` are initialized in `__init__`, we can remove the `self.trackers = []` inside of `init_trackers` ([here](https://github.com/huggingface/accelerate/blob/956114ac92cfbdfe0874ca73aa37ac815326f040/src/accelerate/accelerator.py#L2402)).

Also, we don't need to check `getattr(self, "trackers", [])` in `get_tracker` anymore because `self.trackers` always exists ([here](https://github.com/huggingface/accelerate/blob/956114ac92cfbdfe0874ca73aa37ac815326f040/src/accelerate/accelerator.py#L2444C16-L2444C45)).

Ben-Epstein:
Hey thanks @BenjaminBossan  --  the first (removing self.trackers in `init_trackers`) was already removed, but i've cleaned up the `getattr` and just referenced `self.trackers`

BenjaminBossan (approved):
Great, LGTM, thanks for addressing the remaining points.

PR #2001, opened 2023-09-22
Fix skip first batch for deepspeed example

SumanthRH:
# What does this PR do?

Fixes skipping first batches for the DeepSpeed example. Same changes as #1466 

cc @muellerzr @pacman100


pacman100 (approved):
Thank you @SumanthRH, LGTM! 🤗

PR #1777, opened 2023-07-26
support for ram efficient loading of model with FSDP

pacman100:
### What does this PR do?
1. Enables this PR https://github.com/huggingface/transformers/pull/25107. All explanation is in that PR.

PR #2033, opened 2023-10-05
Ignore torch.autocast for mixed precision when FSDP is used

brcps12:
# What does this PR do?

FSDP supports mixed precision using [MixedPrecision](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.MixedPrecision) class, it does not need to wrap forward function with `torch.autocast`.

The code statement of ignoring this wrapping was added at accelerate v0.22.0, but now removed at v0.23.0

Related PRs are:
- #1777 (Added)
- #1919 (Removed)

I can't find any information about why it is added or removed.

In fact, mixed precision works well even without `torch.autocast`, and even if it is needed, it does not work properly in the current version.

So, I think it need to apply one of the following two options:

1. Add `self.distributed_type != DistributedType.FSDP` in [condition](https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L1385) not to use `torch.autocast`
2. Add `DistributedType.FSDP` in [this file](https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/modeling.py#L1455-L1461)

The reason for 2 is that when FSDP is used, the `distributed_type` field is replaced with `DistribytedType.FSDP` in [this line](https://github.com/huggingface/accelerate/blob/main/src/accelerate/state.py#L767), so I think it needs to be added to support FSDP as well.

As a related issue, the MPT posted on Huggingface Hub uses the [LPNorm class](https://huggingface.co/mosaicml/mpt-7b/blob/main/norm.py#L14), but when learning with FSDP + bfloat16, the dtype changes before and after norm. It is occurred in version v0.23.0.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1382,7 +1382,7 @@ class Accelerator:
         elif device_placement and not self.verify_device_map(model):
             model = model.to(self.device)
 
-        if self.native_amp:
+        if self.native_amp and self.distributed_type != DistributedType.FSDP:
             model._original_forward = model.forward
             model_forward_func = model.forward.__func__ if hasattr(model.forward, "__func__") else model.forward
             autocast_context = get_mixed_precision_context_manager(self.native_amp, self.autocast_handler)


pacman100:
> FSDP supports mixed precision using [MixedPrecision](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.MixedPrecision) class, it does not need to wrap forward function with torch.autocast.

It is required else Mixed Precision FP16 fails with error `RuntimeError: expected scalar type Half but found Float`. See the failing tests because of which this was required: https://github.com/huggingface/accelerate/actions/runs/6079386828/job/16491799916

pacman100:
How about going with the option 2 you mentioned? Does that solve the issue with MPT?

brcps12:
> It is required else Mixed Precision FP16 fails with error RuntimeError: expected scalar type Half but found Float. See the failing tests because of which this was required: https://github.com/huggingface/accelerate/actions/runs/6079386828/job/16491799916

Looking at the problem, it seems that FSDP's MixedPrecision only supports between FSDP modules, not torch's operator (softmax -> matmul). Thank you for sharing!

> How about going with the option 2 you mentioned? Does that solve the issue with MPT?

Yes. The issue has fixed. So, I'm gonna working with option 2 and changes PR title.

pacman100 (approved):
Thank you @brcps12 for fixing the bug wrt bf16 autocasting when using FSDP, LGTM! 🤗

Issue #1948, opened 2023-09-08
Error downloading wav2vec2 model when using accelerate launch with FSDP


markjosims:
### System Info

```Shell
- `Accelerate` version: 0.22.0
- Platform: Linux-5.4.0-155-generic-x86_64-with-glibc2.29
- Python version: 3.8.10
- Numpy version: 1.23.5
- PyTorch version (GPU?): 2.0.1+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 125.55 GB
- GPU type: NVIDIA TITAN V
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: FSDP
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - fsdp_config: {'fsdp_auto_wrap_policy': 'NO_WRAP', 'fsdp_backward_prefetch_policy': 'NO_PREFETCH', 'fsdp_forward_prefetch': False, 'fsdp_offload_params': True, 'fsdp_sharding_strategy': 1, 'fsdp_state_dict_type': 'SHARDED_STATE_DICT', 'fsdp_sync_module_states': True, 'fsdp_use_orig_params': False}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

download_wav2vec2.py
```python
from transformers import Wav2Vec2Model

model_url = 'facebook/wav2vec2-base-960h'
model = Wav2Vec2Model.from_pretrained(model_url)
```
If I run `python download_wav2vec2.py` in CLI it downloads the model as expected, but if I run `accelerate launch download_wav2vec2.py` I get a `NotImplentedError`. If I disable FSDP it works fine. This is the stack trace I get with FSDP enabled:
```bash
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
[...]

Issue #2031, opened 2023-10-04
NotImplementedError: Cannot copy out of meta tensor; no data! - while using FSDP [works with DDP and DeepSpeed]


ari9dam:
### System Info

```Shell
transformers==4.34.0.dev0
accelerate==0.23.0
torch==2.0.1
cuda==11.7
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts
- Training on A100 with 4 nodes

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

import transformers
model = transformers.MistralForCausalLM.from_pretrained(model_path)

Error:
Traceback (most recent call last):
File "./trainer.py", line 198, in
train()
File "./trainer.py", line 152, in train
model = transformers.MistralForCausalLM.from_pretrained(
File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/modeling_utils.py", line 3301, in from_pretrained
) = cls._load_pretrained_model(
File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/modeling_utils.py", line 3689, in _load_pretrained_model
new_error_msgs, offload_index, state_dict_index = _load_state_dict_into_meta_model(
File "/opt/conda/envs/ptca/lib/python3.8/site-packages/transformers/modeling_utils.py", line 741, in _load_state_dict_into_meta_model
set_module_tensor_to_device(model, param_name, param_device, **set_module_kwargs)
File "/opt/conda/envs/ptca/lib/python3.8/site-packages/accelerate/utils/modeling.py", line 317, in set_module_tensor_to_device
new_value = value.to(device)
NotImplementedError: Cannot copy out of meta tensor; no data!

### Expected behavior

model loads successfully

PR #2037, opened 2023-10-06
Make fsdp ram efficient loading optional

pacman100:
# What does this PR do?
1. Make fsdp ram efficient loading optional. Certain models are having issues when handling `meta` devices during pre-trained model loading. Fixes https://github.com/huggingface/accelerate/issues/1948 and https://github.com/huggingface/accelerate/issues/2031 by making the ram efficient loading optional.

### DIFF
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -96,7 +96,7 @@ all-gather while executing in the forward pass. only use with Static graphs.
 Useful in cases such as parameter-efficient fine-tuning. 
 Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019)
 
-`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers models. This should be set to False if experience errors when loading the pretrained model via `from_pretrained`. When using this, `Sync Module States` needs to be True else all the processes expect the main process would have random empty weights leading to unexpected behaviour during training.
+`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers models. This should be set to False if you experience errors when loading the pretrained 🤗 Transformers model via `from_pretrained` method. When using this, `Sync Module States` needs to be True else all the processes expect the main process would have random empty weights leading to unexpected behaviour during training.
 
 `Sync Module States`: If True, each individually wrapped FSDP unit will broadcast module parameters from rank 0
 ```
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -96,6 +96,8 @@ all-gather while executing in the forward pass. only use with Static graphs.
 Useful in cases such as parameter-efficient fine-tuning. 
 Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019)
 
+`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers. When using this, `Sync Module States` needs to True. 
+
 `Sync Module States`: If True, each individually wrapped FSDP unit will broadcast module parameters from rank 0
 ```
 
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -96,7 +96,7 @@ all-gather while executing in the forward pass. only use with Static graphs.
 Useful in cases such as parameter-efficient fine-tuning. 
 Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019)
 
-`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers. When using this, `Sync Module States` needs to True. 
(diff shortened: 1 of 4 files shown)


src/accelerate/commands/config/cluster.py
```diff
@@ -386,12 +386,21 @@ def get_cluster_input():
                 default=False,
                 error_message="Please enter yes or no.",
             )
-            fsdp_config["fsdp_sync_module_states"] = _ask_field(
-                "Do you want each individually wrapped FSDP unit to broadcast module parameters from rank 0 at the start? [YES/no]: ",
+            fsdp_config["fsdp_cpu_ram_efficient_loading"] = _ask_field(
+                "Do you want to enable CPU RAM efficient model loading? Only applicable for 🤗 Transformers. [YES/no]: ",
```
muellerzr, line 390: Should we be a bit more clear? Like Transformer models? (I say this because the problem model in general was downloaded from `transformers`, so we should find a better way to specify this I think)

src/accelerate/commands/launch.py
```diff
@@ -524,6 +524,14 @@ def launch_command_parser(subparsers=None):
         help="If True, allows non-uniform `requires_grad` during init, which means support for interspersed frozen and trainable paramteres."
         " (useful only when `use_fsdp` flag is passed).",
     )
+    fsdp_args.add_argument(
+        "--fsdp_cpu_ram_efficient_loading",
+        default="true",
+        type=str,
+        help="If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. "
+        "Only applicable for 🤗 Transformers. When using this, `--fsdp_sync_module_states` needs to True. "
+        "(useful only when `use_fsdp` flag is passed).",
```
muellerzr, line 533: Since we need `sync_module_states`, I think we should add a check for this and raise an error if it's anything but `"true"`

muellerzr (commented):
Thanks for the fix! I added a few suggestions to try and make a few points more clear, let's see if we can improve it some :)

src/accelerate/commands/launch.py
pacman100: Done

src/accelerate/commands/config/cluster.py
pacman100: Done

muellerzr (approved):
Thanks! cc @BenjaminBossan for a secondary :)

src/accelerate/utils/launch.py
```diff
@@ -187,7 +188,12 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
             current_env["FSDP_STATE_DICT_TYPE"] = str(args.fsdp_state_dict_type)
         current_env["FSDP_FORWARD_PREFETCH"] = str(args.fsdp_forward_prefetch).lower()
         current_env["FSDP_USE_ORIG_PARAMS"] = str(args.fsdp_use_orig_params).lower()
+        current_env["FSDP_CPU_RAM_EFFICIENT_LOADING"] = str(args.fsdp_cpu_ram_efficient_loading).lower()
         current_env["FSDP_SYNC_MODULE_STATES"] = str(args.fsdp_sync_module_states).lower()
+        if str_to_bool(current_env["FSDP_CPU_RAM_EFFICIENT_LOADING"]) and not str_to_bool(
```
BenjaminBossan, line 193: Is there a specific reason to check this on `current_env` instead of checking `args` directly, which seems easier? Also, I prefer to run all checks as early as possible, so I wonder if this shouldn't be moved up in the function body.

docs/source/usage_guides/fsdp.md
```diff
@@ -96,6 +96,8 @@ all-gather while executing in the forward pass. only use with Static graphs.
 Useful in cases such as parameter-efficient fine-tuning. 
 Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019)
 
+`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers models. When using this, `Sync Module States` needs to True.
```
BenjaminBossan, line 99: Do we know under what circumstances exactly this may cause problems? Maybe it would be worth mentioning, so that those users who currently encounter the problem know that they need to change this argument.

BenjaminBossan (approved):
Looks good, thanks Sourab. I only have two small nits, up to you if you want to address them.

src/accelerate/utils/launch.py
pacman100: Done

docs/source/usage_guides/fsdp.md
pacman100: Done

docs/source/usage_guides/fsdp.md
```diff
@@ -96,7 +96,7 @@ all-gather while executing in the forward pass. only use with Static graphs.
 Useful in cases such as parameter-efficient fine-tuning. 
 Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019)
 
-`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers models. When using this, `Sync Module States` needs to True. 
+`CPU RAM Efficient Model loading`: If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. Only applicable for 🤗 Transformers models. When using this, `Sync Module States` needs to True else all the processes expect the main process would have random empty weights leading to unexpected behaviour during training.
```
BenjaminBossan, line 99: "needs to True" => "needs to **be** True"

I think there was a misunderstanding here, maybe I didn't explain myself well. What I meant earlier with adding an explanation was not related to why `fsdp_sync_module_states` must be `True` if `fsdp_cpu_ram_efficient_loading` is set to `True`. What I meant is adding an explanation under what circumstances users should set `fsdp_cpu_ram_efficient_loading=False`.

pacman100: Done

BenjaminBossan (approved):
Great, thanks for enhancing the doc of the parameter.

PR #2041, opened 2023-10-09
Do not attempt to pad nested tensors

### LINKED ISSUE #2040: pad_across_processes not compatible with PyTorch's nested_tensor
Buddy issue for https://github.com/huggingface/transformers/issues/26687 . The issue is basically identical in accelerate to transformers, so this issue can just be used for tracking and not discussion.

frankier:
# What does this PR do?

Fixes pad_across_processes erroring on nested tensors when reaching `len(tensor.shape)`

Fixes #2040

Looking for a bit of guidance/direction on whether other things are needed.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@pacman100


BenjaminBossan:
@frankier Could you please merge with/rebase on main, that should fix the issue with CI.

frankier:
> AFAICT, accelerate doesn't really take them into account for now, I wouldn't be surprised if more errors can occur. 

There are not more issues when I'm using accelerate from `transformers` Trainer, but there are issues in `transformers` if they are not converted in a custom `preprocess_logits_for_metrics`. Once this PR is merged, I will write another PR to transformers to improve the error messages in `transformers`.

> It would probably require some unit tests to check this.

The test I know I how to write now at the moment is just a smoke/unit test that checks that nested_tensor is returned as-is roundtripped when passed to `pad_across_processes`. I think this test would be a bit vacuous, but I can easily add it to the PR if you like. Could you please direct me which file/class would be the best place for it?

> As for padding, I don't understand yet why [to_padded_tensor](https://pytorch.org/docs/stable/nested.html#torch.nested.to_padded_tensor) cannot be used

The problem is that it uses in-band signalling. Whichever padding element you choose, may actually occur in the array. Adding this as default behavior could cause confusing bugs for users. It's better to force them to choose how to deal with the nested_tensor.

> what is the relation to numpy?

This might be from and old, possibly miscorrect memory I have about the default behavior of `preprocess_logits_for_metrics` converting everything to numpy arrays.

BenjaminBossan:
> There are not more issues when I'm using accelerate from `transformers` Trainer, but there are issues in `transformers` if they are not converted in a custom `preprocess_logits_for_metrics`.

There could still be edge cases that you don't encounter but that other users may run into. E.g. the simple fact that `tensor.shape` may fail could easily lead to problems. Maybe accelerate dodges all of them, but we cannot know until we really test it. As is, I'd say we don't officially support nested tensors -- the docs mention they're at prototype stage, so it makes sense to wait.

That being said, there is no harm in accelerate getting out of the way if users happen to choose nested tensors, which is why I think this PR still makes sense.

> The test I know I how to write now at the moment is just a smoke/unit test that checks that nested_tensor is returned as-is roundtripped when passed to `pad_across_processes`. I think this test would be a bit vacuous, but I can easily add it to the PR if you like.

I don't think it would be vacuous. You need to consider that someone else might come in the future and remove the check inadvertently. Having a test would alert us to this.

> Could you please direct me which file/class would be the best place for it?

Not completely sure, [`test_utils.py`](https://github.com/huggingface/accelerate/blob/main/tests/test_utils.py) seems most appropriate, @muellerzr WDYT?

> The problem is that it uses in-band signalling. Whichever padding element you choose, may actually occur in the array.

I see, that makes sense. However, I do wonder if we should give a warning in that situation, so that the user is aware that padding did not take place.

> This might be from and old, possibly miscorrect memory I have about the default behavior of `preprocess_logits_for_metrics` converting everything to numpy arrays.

AFAICT, there is nothing in accelerate that requires this. A quick search didn't reveal anything in transformers `Trainer` either, but I don't have experience with it, so might be wrong.

muellerzr:
Agree with all of @BenjaminBossan's points

> Not completely sure, [test_utils.py](https://github.com/huggingface/accelerate/blob/main/tests/test_utils.py) seems most appropriate, @muellerzr WDYT?

Yes indeed!

frankier:
Please excuse the delay. I think I've addressed all this now.

BenjaminBossan:
@frankier Could you please run `make style` and `make quality`?

frankier:
Whoops. Thought I had(!) Done now.

BenjaminBossan (approved):
Thanks, looks good to me.

muellerzr (approved):
Thanks!

Issue #2007, opened 2023-09-25
`gather_for_metrics` does not respect dataloader `drop_last=True` setting


xmagpie:
### System Info

```Shell
- `Accelerate` version: 0.23.0
- Platform: Linux-5.15.0-76-generic-x86_64-with-glibc2.35
- Python version: 3.9.16
- Numpy version: 1.24.2
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

This is the minimal snippet adapted directly from the docstring:
```python
import torch
from accelerate import Accelerator

accelerator = Accelerator()
dataloader = torch.utils.data.DataLoader(range(9), batch_size=5, drop_last=True)
dataloader = accelerator.prepare(dataloader)

batch = next(iter(dataloader))
gathered_items = accelerator.gather_for_metrics(batch)

print(gathered_items)
```

This yields `tensor([0, 1, 2, 3], device='cuda:0')`, while the expected is `tensor([0, 1, 2, 3, 4], device='cuda:0')`.

The behavior is more visually erroneous in the following case

```python
dataloader = torch.utils.data.DataLoader(range(11), batch_size=5, drop_last=True)
dataloader = accelerator.prepare(dataloader)

iterator = iter(dataloader)
next(iterator) # Skip first batch tensor([0, 1, 2, 3, 4], device='cuda:0')
batch = next(iterator)
gathered_items = accelerator.gather_for_metrics(batch)
print(batch, gathered_items)
```

yields `tensor([5, 6, 7, 8, 9], device='cuda:0'), tensor([5], device='cuda:0')`.

The error seems to be related to this [check](https://github.com/huggingface/accelerate/blob/80da9cfb09bb3cc9f1b385cb55d6b90d025a5fd9/src/accelerate/accelerator.py#L2228), which does not correctly handle drop_last case, which results in the incorrect usage of remainder.

Also I have noticed that `drop_last` is not correctly initialized on prepared dataloader, in my case `DataLoaderShard`, e.g. it shows `False`. Dunno whether this is intended or not.

[...]

PR #2048, opened 2023-10-11
Let drop_last modify `gather_for_metrics`

muellerzr:
# What does this PR do?

Ensures that if `drop_last=True`, the default remainder for the `GradientState` is utilized since we are just dropping the non-fitting batch anyways instead.

Fixes # (issue)

closes https://github.com/huggingface/accelerate/issues/2007

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @LysandreJik

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -324,8 +324,9 @@ class DataLoaderStateMixin:
         "Prepares the gradient state for the current dataloader"
         self.reset()
         with suppress(Exception):
-            length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
-            self.remainder = length % self.total_batch_size
+            if not self._drop_last:
+                length = getattr(self.dataset, "total_dataset_length", len(self.dataset))
+                self.remainder = length % self.total_batch_size
         self.gradient_state._add_dataloader(self)
 
     def end(self):
@@ -366,13 +367,23 @@ class DataLoaderShard(DataLoader, DataLoaderStateMixin):
         - **total_dataset_length** (`int`) -- Total length of the inner dataset across all processes.
     """
 
-    def __init__(self, dataset, device=None, rng_types=None, synchronized_generator=None, skip_batches=0, **kwargs):
+    def __init__(
+        self,
+        dataset,
+        device=None,
+        rng_types=None,
+        synchronized_generator=None,
+        skip_batches=0,
+        _drop_last: bool = False,
+        **kwargs,
+    ):
         super().__init__(dataset, **kwargs)
         self.device = device
         self.rng_types = rng_types
         self.synchronized_generator = synchronized_generator
         self.skip_batches = skip_batches
         self.gradient_state = GradientState()
+        self._drop_last = _drop_last
 
     def __iter__(self):
         if self.rng_types is not None:
@@ -815,6 +826,7 @@ def prepare_data_loader(
             sampler=new_batch_sampler,
             batch_size=dataloader.batch_size,
             rng_types=rng_types,
+            _drop_last=dataloader.drop_last,
             synchronized_generator=synchronized_generator,
             **kwargs,
         )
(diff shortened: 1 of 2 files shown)


src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -219,6 +219,20 @@ def __iter__(self):
...
+    accelerator = Accelerator()
+    dataloader = DataLoader(range((10 * accelerator.num_processes) + 1), batch_size=5, drop_last=True)
+    dataloader = accelerator.prepare(dataloader)
+
+    iterator = iter(dataloader)
+    next(iterator)  # Skip first batch tensor([0, 1, 2, 3, 4], device='cuda:0')
+    batch = next(iterator)
+    gathered_items = accelerator.gather_for_metrics(batch)
+    assert gathered_items.size(0) == (
+        5 * accelerator.num_processes
```
BenjaminBossan, line 232: To be clear, the 5 comes from the fact that batch size is 5, right? Maybe, to make it more transparent, we could define `bs = 5` at the top of the function, initialize the loader with `dataloader = DataLoader(range((2 * bs * accelerator.num_processes) + 1), batch_size=bs, drop_last=True)`, and replace 5 here and below by `bs`.

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -219,6 +219,20 @@ def __iter__(self):
...
+    ), f"Expected number of items: {(5*accelerator.num_processes)}, Actual: {gathered_items.size(0)}"
```
BenjaminBossan, line 233:
```suggestion
    ), f"Expected number of items: {5*accelerator.num_processes}, Actual: {gathered_items.size(0)}"
```

src/accelerate/data_loader.py
```diff
@@ -366,13 +367,23 @@ class DataLoaderShard(DataLoader, DataLoaderStateMixin):
...
 
-    def __init__(self, dataset, device=None, rng_types=None, synchronized_generator=None, skip_batches=0, **kwargs):
+    def __init__(
+        self,
+        dataset,
+        device=None,
+        rng_types=None,
+        synchronized_generator=None,
+        skip_batches=0,
+        _drop_last: bool = False,
```
BenjaminBossan, line 377: Why use an underscore for the argument name? That's quite unusual. `drop_last` on the `dataloader` doesn't use an underscore. If the attribute should be marked as "private", it can still use an underscore, but for the argument name, I'd go without (i.e. `self._drop_last = drop_last`).

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -255,6 +269,9 @@ def main():
     accelerator = Accelerator()
     test_torch_metrics(accelerator, 512)
     accelerator.state._reset_state()
+    if accelerator.is_local_main_process:
+        print("**Test that `drop_last` is taken into account**")
```
BenjaminBossan, line 273: Could use `accelerator.print` instead of `if ...`, right? :)

BenjaminBossan (approved):
Thanks for fixing the bug. Only some nits, feel free to ignore.

src/accelerate/data_loader.py
muellerzr: Because it is indeed private, and to be consistent with the other dataloader arg which has it as private too. Because we control it, the user shouldn't ever pass in `_drop_last` (but we need to on the backend)

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
muellerzr: We *can*, I can do all of those in a follow-up PR 😉

src/accelerate/data_loader.py
BenjaminBossan: I see. It's a bit strange to the eye, but as you said, it should not be visible to the user.

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
BenjaminBossan: Really not necessary, was more in jest :-p

BenjaminBossan (approved):

Issue #2016, opened 2023-09-30
Dataloader yields incorrect sequence after resuming from checkpointing


antoche:
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-4.15.0-213-generic-x86_64-with-glibc2.27
- Python version: 3.9.10
- Numpy version: 1.22.3
- PyTorch version (GPU?): 1.13.0a0+git49444c3 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 125.39 GB
- GPU type: Quadro RTX 8000
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

Consider the following script, which compares the behaviour of a training loop with and without checkpointing:

```python
import os
import torch
import accelerate


def make_data(n: int, seed: int = 1) -> torch.utils.data.Dataset:
    """
    Create a dummy datatset
    """
    generator = torch.Generator(device="cpu").manual_seed(seed)
    data = torch.randint(0, 10, (n,), generator=generator)
    print(f"data is {data}")
    training_dataset = torch.utils.data.TensorDataset(data)
    return training_dataset


def prepare_training(
    accelerator: accelerate.Accelerator, training_dataset: torch.utils.data.Dataset, shuffle: bool, seed: int
) -> torch.utils.data.DataLoader:
    """
    Prepare a dataloader from the given dataste
    """
    generator = torch.Generator().manual_seed(seed)
    dataloader = accelerator.prepare(
        torch.utils.data.DataLoader(training_dataset, batch_size=1, shuffle=shuffle, generator=generator)
    )
    return dataloader


def train_loop(
    accelerator: accelerate.Accelerator,
    training_dataset: torch.utils.data.Dataset,
    epochs: int,
    shuffle: bool,
    checkpoint_path: str = None,
    restart_epoch: int = None,
    restart_step: int = None,
    stop_epoch: int = None,
    stop_step: int = None,
):
    """
    Mimic a basic training loop supporting checkpointing.

    Returns the items retrieved form the dataloader over all epochs as a list.
    """
    result = []
    dataloader = prepare_training(accelerator, training_dataset, shuffle, 0)
[...]

PR #2057, opened 2023-10-13
Allow for samplers to be seedable and reproducable

muellerzr:
# What does this PR do?

This PR ensures that `DataLoaders` made through `Accelerate` are reproducible on distributed systems by allowing `set_epoch` to work on `RandomSamplers` (what gets selected if `shuffle=True` on a `DataLoader`).

Fixes # (issue)

fixes https://github.com/huggingface/transformers/issues/26541
fixes https://github.com/huggingface/accelerate/issues/2016

To test this works, I ran the following script:

```python

set_seed(42)
accelerator = Accelerator()
dataset = range(24)
dataloader = torch.utils.data.DataLoader(dataset, batch_size=8, shuffle=True)

dataloader.set_epoch(0)
original_outputs = []

for x in dataloader:
    original_outputs += x

original_outputs = torch.stack(original_outputs)

dataloader.set_epoch(0)
new_outputs = []

for x in dataloader:
    new_outputs += x
    
new_outputs = torch.stack(new_outputs)

print(f'Original: {original_outputs}\n')
print(f'New: {new_outputs}\n')
```

When comparing the outputs, these two line up perfectly, ensuring we can in-fact recreate what was originally setup.

For testing we will be relying on the transformers integration CI (see the test runners on main that include the failing test). Linked with https://github.com/huggingface/transformers/pull/26850

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @LysandreJik


muellerzr:
Note from @BenjaminBossan:

Eventually we may want to consider whether a user passes in their own `generator` into the sampler init of their own `RandomSampler`, and if we should use that as part of the `__iter__`, such as:

```python
seed = self.epoch + getattr(self.generator, "initial_seed", 0)
```

src/accelerate/data_loader.py
```diff
@@ -64,6 +64,40 @@
...
+        self.epoch = 0
+
+    def __iter__(self):
+        g = torch.Generator()
+        if self.generator is not None:
+            seed = self.epoch + self.generator.initial_seed()
+        else:
+            seed = self.epoch
+        g.manual_seed(seed)
+        n = len(self.data_source)
```
BenjaminBossan, line 89: The implementation starting from here is quite different from the original `RandomSampler.__iter__`. Why is that?

src/accelerate/data_loader.py
```diff
@@ -412,8 +460,19 @@ def __iter__(self):
...
         self.end()
 
+    def set_epoch(self, epoch: int):
+        # In case it is manually passed in, the user can set it to what they like
+        if self.iteration != epoch:
+            self.iteration = epoch
+        if hasattr(self.batch_sampler, "sampler") and hasattr(self.batch_sampler.sampler, "set_epoch"):
+            self.batch_sampler.sampler.set_epoch(epoch)
+        elif hasattr(self.dataset, "set_epoch"):
+            self.dataset.set_epoch(epoch)
```
BenjaminBossan, line 474: Can we get a comment here about what types of `Dataset`s would have that method?

BenjaminBossan (commented):
Wow, this was a tough nut to crack. This looks good and should hopefully enable future runs to be completely reproducible.

I have some comments, nothing big, please take a look. Also, there are several typos concerning "reproduc**a**bility" => "reproduc**i**bility"

src/accelerate/data_loader.py
BenjaminBossan: Oh I see, I haven't checked different versions. It looks like the change to using `yield from` is already in PyTorch since v1.7.0 though:

https://github.com/pytorch/pytorch/blame/v1.7.0/torch/utils/data/sampler.py

I checked the PR where this was changed:

https://github.com/pytorch/pytorch/pull/40026

Honestly, not sure if we care so much about it. However, I think the changes from this PR would be worthwhile having here as well:

https://github.com/pytorch/pytorch/pull/103339

which would be an argument in favor of copying the current implementation. But I'm fine with either way.

(Too bad that PyTorch has not moved the seeding to a separate method, which would allow us to override it and leave `__iter__` as is. With the given implementation, I don't think we can seed and call `super().__iter__()`.)

BenjaminBossan (approved):
I dug a bit deeper on the `__iter__` implementation of `RandomSampler`. Please take a look. Regardless of that, I'm fine with the PR, feel free to merge if you disagree with my comment.

src/accelerate/data_loader.py
muellerzr: Fair point, brought it back in. We sadly can't just call `super().__iter__()`, so copy/pasted with a reference of what is going on/where

LysandreJik (approved):
Thank you!

src/accelerate/data_loader.py
BenjaminBossan: Note that this still does not contain the change from https://github.com/pytorch/pytorch/pull/103339, which uses `map(int, ...)` instead of `tolist()`. This is to avoid eagerly creating a Python list, which is very inefficient for big `n`.

PR #868, opened 2022-11-18
fix failing deepspeed test

pacman100:
### What does this PR do?
1. Fixes a failing deepspeed test, thanks to DeepSpeed team to bring this to our notice.

PR #2058, opened 2023-10-15
Change assignment during sub-module device alignment

blbadger:
# What does this PR do?

Fixes`peft` issue #868, https://github.com/huggingface/peft/issues/868

This PR addresses what appears to be a bug in the loading and merging/unloading of PeftModels.  

Background: After loading the necessary libraries,

```
!pip install -q -U bitsandbytes
!pip install -q -U git+https://github.com/huggingface/transformers.git
!pip install -q -U git+https://github.com/huggingface/peft.git
!pip install -q -U git+https://github.com/huggingface/accelerate.git
!pip install -q datasets==2.13.1 pyarrow==8.0.0 tokenizers==0.14.0 sentencepiece==0.1.99
```

attempting to call `PeftModel.from_pretrained()` on a model loaded using `device_map='auto'` as follows

```python
model = AutoModelForCausalLM.from_pretrained("emozilla/LLongMA-2-7b-flash", device_map='auto') 
lora_checkpoint = '/path/to/lora' # add path
model = PeftModel.from_pretrained(model, lora_checkpoint) 
```

results in many parameters being loaded as meta tensors with no data.  

Attempting to perform merging and unloading escapes explicit errors, but attempting to then load the merged model results in the error message "Cannot copy out of meta tensor; no data!" (`peft` issue #868).  This does not occur if `device_map='cpu'` is specified.

The root of the problem appears to be the choice to send module tensors to the "meta" device when enforcing a device alignment via `AlignDevicesHook` which would explain why it is present during LoRA application but typically not during the loading (and dispatching) of merged pretrained models.  

This PR allows for the loading and merging of LoRA parameters using the "auto" device map by assigning offloaded tensors to the CPU rather than the "meta" device. Doing so sends all sub-module tensors to one device (CPU) and thus aligns them, while saving the GPU memory that was specified by choosing the "auto" device map. 

Note that there may be good reason to recast offloaded parameters as meta tensors during device alignment, and if so I would be happy to work on an alternative approach to the problem.

## Before submitting
- [ x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
      to it if that's the case.

## Who can review?

Anyone!
@pacman100

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -266,7 +266,7 @@ class AlignDevicesHook(ModelHook):
             for name, _ in named_module_tensors(
                 module, include_buffers=self.offload_buffers, recurse=self.place_submodules
             ):
-                set_module_tensor_to_device(module, name, "meta")
+                set_module_tensor_to_device(module, name, "cpu")
             if not self.offload_buffers and self.execution_device is not None:
                 for name, _ in module.named_buffers(recurse=self.place_submodules):
                     set_module_tensor_to_device(module, name, self.execution_device)


SunMarc (commented):
Hi @blbadger , as you suspected, we need to put the offloaded parameters to the `meta` device because we also need to consider for parameters that are offloaded on the disk. If we put all offloaded params on the cpu, we may get OOM. The problem our design is that we don't make the difference between cpu offload and disk offload when putting the parameters on `meta` device.

blbadger:
Thanks for clarifying @SunMarc! 

I have simplified the working tree and will work on preserving the cpu/disk distinction while sending parameters to the meta device.

muellerzr (changes requested):
This still does not address what Marc mentioned. What would be a closer solution here is adding a param to `init_hook` (and the other hooks that set `"cpu"`) to include a default device to use, or explicit arg that is `cpu` which we can call at the higher levels. Because as a baseline the *init on meta should never change as a base case*.

blbadger:
Understood @muellerzr, I can see why the default "meta" assignment should not change. 

After more thought it seems to me that the `peft` issue 868 is most elegantly addressed in that library itself rather than this one. The problem in that issue is really that the offloaded parameters are not re-loaded before merging and unloading, so I am working on a solution where each module calls the pre-forward device alignment hook to load parameters and then offloads again after merging to best maintain the original device alignment.

Therefore I would be happy to close this PR if that sounds good to everyone.

muellerzr:
Sounds good to me @blbadger!

PR #2073, opened 2023-10-23
DOC: Fix broken link to designing a device map

BenjaminBossan:
# What does this PR do?

There is a typo in the link, which is fixed by this PR.

## Before submitting

## Who can review?

@muellerzr


muellerzr (approved):
Thanks!

Issue #2074, opened 2023-10-23
Question/Feature Request - Model Pipeline Parallelism on multiple GPUs without offloading?


chiragjn:
### System Info

```Shell
accelerate==0.23.0
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

This started from https://github.com/huggingface/peft/issues/868 issue where `device_map="auto"` can leave some of the weights on the meta device because of insufficient resources when calculating the device map but that can cause issues when saving the checkpoint as weights are not being moved to a real device before saving.

While device_map="auto" is very useful, how can I optionally,
- Disable any offloading to CPU and Disk but still use multiple GPUs while inferring the device map automatically
- Provide a list of devices explicitly and enforce that nothing stays on meta device - raise an error if the model will not fit on given devices.

### Expected behavior

NA

SunMarc:
Hi @chiragjn, thanks for raising this issue. We are looking into how to better handle this. 
To answer some of your questions: 

> Disable any offloading to CPU and Disk but still use multiple GPUs while inferring the device map automatically

If device_map = "auto" uses offloading, it means that you don't have enough space on your GPUs to load your your model. 

> Provide a list of devices explicitly and enforce that nothing stays on meta device - raise an error if the model will not fit on given devices.

Currently, there is a `memory_map` arg that can be used for this purpose but you need to map each device to a memory size.

PR #2075, opened 2023-10-23
Add logs offloading

SunMarc:
# What does this PR do ? 
This PR adds a warning if there is any parameters on the meta device because of the offloading.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -405,7 +405,7 @@ def dispatch_model(
         # Attaching the hook may break tied weights, so we retie them
         retie_parameters(model, tied_params)
 
-        # warn if there is any param on the meta device
+        # warn if there is any params on the meta device
         if len(offload) > 0:
             logging.warn("Some parameters are on the meta device device because they were offloaded to the cpu/disk.")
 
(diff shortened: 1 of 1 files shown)


src/accelerate/big_modeling.py
```diff
@@ -405,6 +405,10 @@ def dispatch_model(
         # Attaching the hook may break tied weights, so we retie them
         retie_parameters(model, tied_params)
 
+        # warn if there is any params on the meta device
+        if len(offload) > 0:
+            logging.warn("Some parameters are on the meta device device because they were offloaded to the cpu/disk.")
```
muellerzr, line 410:
```suggestion
            logging.warning("Some parameters are on the meta device device because they were offloaded to the cpu and/or disk.")
```

Warn has been deprecated since py 3.3, let's use warning :)

muellerzr (approved):
LG2M! Just a small nit. (Also would be nice if we could perhaps specify explicitly CPU or disk if `offload` let's us find that out :) )

src/accelerate/big_modeling.py
SunMarc: sound good

SunMarc:
> Apart from Zach's comment, LGTM, thanks. I assume this is in response to https://github.com/huggingface/accelerate/issues/2074?

Yes it is

> Only question for me is if the warning can be raised earlier already, e.g. line 386, or if it has to be at this point.
The model is dispatched after `attach_align_device_hook_on_blocks`, so we can raise it after this function. I will do the modification.

Issue #1929, opened 2023-09-06
wait_for_everyone() and other features of Accelerate do not function without shared memory


bghira:
### System Info

```Shell
- `Accelerate` version: 0.21.0
- Platform: Linux-5.15.0-76-generic-x86_64-with-glibc2.31
- Python version: 3.10.11
- Numpy version: 1.25.2
- PyTorch version (GPU?): 2.0.1+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 393.40 GB
- GPU type: NVIDIA RTX A6000
- `Accelerate` default config:
        Not found
```

### Information

- [x] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

The above system works, but an equivalent system and environment with the only difference being the kernel version, meant that Accelerate will hang.

I'm not skilled enough in this space to trace down what happens, but really the lack of shared memory in the 5.4 version kernel seems to bring this on.

Using simple multi gpu setup with 2 A6000s and Stable Diffusion 2.1.

### Expected behavior

Accelerate should properly have the `main_process_first()` move on to the child threads.

bghira:
I should add, it freezes as well with `prepare()` when you get rid of all the other state tracking/blocking decorators.

bghira:
yes, unsolved.

BenjaminBossan:
Could you please provide the code leads to the hang?

bghira:
https://github.com/bghira/SimpleTuner/blob/main/train_sdxl.py < this is where it occurs for SDXL

https://github.com/bghira/SimpleTuner/blob/main/train_sd21.py < this is the SD 2.x trainer that caused the problem

it's actually really frustrating. i found some other issues that caused the hang, where the dataset is improperly partitioned between processes - eg. the dataset ends up imbalanced with a few entries short on one process that caused it to move to the next epoch rather than do a backward pass. meanwhile, the other processes that had more (the "correct" amount of data) did not move to the next epoch.

it's hard to reproduce, even with the same codebase and kernel version. some of the 5.4 systems work correctly. but they are just, very slow. other times, it just waits forever - and then NCCL timeout.

BenjaminBossan:
[...]

PR #2077, opened 2023-10-24
Warn when kernel version is too low on Linux

BenjaminBossan:
# What does this PR do?

See #1929

On Linux with kernel version < 5.5, issues with hanging processes have been reported. It is not clear how to fix the issue, so instead we warn the user that they may encounter problems.

### Notes

As logging requires an initialized `PartialState`, the actual check happens at the end of `Accelerator.__init__`.

In a similar vein, the docstring of `get_logger` has been adjusted to first initialize the `Accelerator`, as it is not working as currently shown.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/src/accelerate/logging.py
+++ b/src/accelerate/logging.py
@@ -85,6 +85,9 @@ def get_logger(name: str, log_level: str = None):
 
     ```python
     >>> from accelerate.logging import get_logger
+    >>> from accelerate import Accelerator
+
+    >>> accelerator = Accelerator()
 
     >>> logger = get_logger(__name__)
 
@@ -95,9 +98,6 @@ def get_logger(name: str, log_level: str = None):
     >>> logger.info("My log")
     >>> logger.debug("My second log")
 
-    >>> from accelerate import Accelerator
-
-    >>> accelerator = Accelerator()
     >>> array = ["a", "b", "c", "d"]
     >>> letter_at_rank = array[accelerator.process_index]
     >>> logger.info(letter_at_rank, in_order=True)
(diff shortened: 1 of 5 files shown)


src/accelerate/logging.py
```diff
@@ -85,6 +85,9 @@ def get_logger(name: str, log_level: str = None):
 
     ```python
     >>> from accelerate.logging import get_logger
+    >>> from accelerate import Accelerator
+
+    >>> accelerator = Accelerator()
```
muellerzr, line 90: This is incorrect. We only need to instantiate before we log. Just making the logger should not require this, and the current example docstring is accurate. If it’s wrong in some way, we need to fix that.

muellerzr (commented):
Will give it a thorough review later today, but one initial comment

src/accelerate/logging.py
BenjaminBossan: You are right, it is possible to call `get_logger` without initializing the `Accelerator` first, but not to use it, like calling `logger.info("My log")`. Therefore, the example as is does not work. Of course, we could move the new lines below the call to `get_logger`, but I wanted to keep the imports grouped for better readability.

muellerzr: Import grouped together is fine :)

BenjaminBossan: So you mean the import can stay there but `accelerator = Accelerator()` should be moved below `logger = get_logger(__name__)`?

BenjaminBossan: Done

muellerzr (approved):
Thanks!

PR #2078, opened 2023-10-24
check port availability only in main deepspeed/torchrun launcher

Jingru:
# What does this PR do?

Currently, `main_process_port` availability is checked on all the deepspeed launchers (for multi_node senario), but this check is only necessary for main launcher whose machine rank is 0.

## Who can review?

- DeepSpeed: @pacman100

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -272,7 +272,8 @@ def prepare_deepspeed_cmd_env(args: argparse.Namespace) -> Tuple[List[str], Dict
     if main_process_port is None:
         main_process_port = 29500
 
-    if is_port_in_use(main_process_port):
+    need_port_check = num_machines <= 1 or int(args.machine_rank) == 0
+    if need_port_check and is_port_in_use(main_process_port):
         raise ConnectionError(
             f"Tried to launch distributed communication on port `{main_process_port}`, but another process is utilizing it. "
             "Please specify a different port (such as using the `----main_process_port` flag or specifying a different `main_process_port` in your config file)"
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -128,6 +128,8 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
     if main_process_port is None:
         main_process_port = 29500
 
+    # only need to check port availability in main process, in case we have to start multiple launchers on the same machine
+    # for some reasons like splitting log files.
     need_port_check = num_machines <= 1 or int(args.machine_rank) == 0
     if need_port_check and is_port_in_use(main_process_port):
         raise ConnectionError(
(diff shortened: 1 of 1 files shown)


muellerzr:
This is great @Jingru, can we also do this for the `torchrun`/`multi-gpu` launcher since it would be good to have the same thing there? Thanks!

Jingru:
> This is great @Jingru, can we also do this for the `torchrun`/`multi-gpu` launcher since it would be good to have the same thing there? Thanks!

Sure thing. Already updated.

muellerzr (approved):
Thanks! This looks great to me.

src/accelerate/utils/launch.py
```diff
@@ -128,7 +128,8 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
     if main_process_port is None:
         main_process_port = 29500
 
-    if is_port_in_use(main_process_port):
+    need_port_check = num_machines <= 1 or int(args.machine_rank) == 0
+    if need_port_check and is_port_in_use(main_process_port):
```
muellerzr, line 132: Just a small nit though thinking on this more, can we add a comment here just mentioning how/why we should only check on the main node for this, in case others want to learn why as well :)

BenjaminBossan (approved):
Apart from Zach's comment, LGTM, thanks.

src/accelerate/utils/launch.py
muellerzr: cc @Jingru :)

Jingru: > cc @Jingru :)

already add some comments

Jingru:
comments added

src/accelerate/utils/launch.py
```diff
@@ -128,7 +128,10 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
     if main_process_port is None:
         main_process_port = 29500
 
-    if is_port_in_use(main_process_port):
+    # only need to check port availability in main process, in case we have to start multiple launchers on the same machine
+    # for some reasons like splitting log files.
+    need_port_check = num_machines <= 1 or int(args.machine_rank) == 0
```
pacman100, line 133: I don't think the check `num_machines <= 1` is needed, wdyt?

pacman100 (commented):
Thank you @Jingru for fixing this issue, left a comment.

src/accelerate/utils/launch.py
Jingru: Cause we don't have a default value for machine_rank and if num_machines <=1, it is highly possible that machine_rank is None.

pacman100: Oh, okay, just checked again and because the `machine_rank` defaults to 0, we need the check on `num_machines`, got it.

pacman100 (approved):

PR #2080, opened 2023-10-24
Fix failing test on main

muellerzr:
# What does this PR do?

We should only save/load the sampler if it's an iterable dataset shard, and if we use our custom sampler. Otherwise pulling from the batch sampler and loading it is more nuanced. For the sake of the release, let's focus on just the new sampler. If users report that they can't reproduce trainings on batch samplers (or if we test and find we can't), then we can pivot after.

Fixes # (issue)

Failing test on main that caught this

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LysandreJik @BenjaminBossan

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -20,7 +20,6 @@ from typing import List
 import numpy as np
 import torch
 from torch.cuda.amp import GradScaler
-from torch.utils.data import BatchSampler
 
 from .utils import (
     MODEL_NAME,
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
LGTM, thanks for fixing the issue.

From my understanding, this is a -- potentially temporary -- fix that makes the saving and loading of the sampler more specific because the current code does not cover some cases. If that's right, could you elaborate where it goes wrong so that we have it written down for the future?

muellerzr:
Where it goes askew is we need a better way to get the users batch sampler from our custom batch sampler. Such as `dataloader.batch_sampler.batch_sampler`. (Since our sampler stores that in the `batch_sampler`.

One potential fix: add `dataloader.get_sampler()` which will know to return either the `batch_sampler` or `sampler` of the *original dataloader*

PR #2087, opened 2023-10-26
New CI Runners

muellerzr:
Migrates CI runners to use AWS

Some notes:

* Use `[self-hosted, intel-cpu, 8-cpu, ci]` to utilize a CPU runner. These aren't stored in our local runners and are spun up by demand, hence why they do not show up in settings
* Use `[self-hosted, docker-gpu]` for GPU runners. (We can also fine-tune these using `push-ci` and `nightly-ci` for whether it's a nightly run or pushing to main)
* In general now these CI runners assume no persistent storage to make things easier


.github/workflows/build-docker-images-release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
   version-cuda:
     name: "Latest Accelerate GPU [version]"
-    runs-on: [self-hosted, docker-gpu, multi-gpu, gcp]
+    runs-on: [self-hosted, intel-cpu, 8-cpu, ci]
```
ydshieh, line 44: a bit strange that here we use cpu machine while the job name is GPU

.github/workflows/self_hosted_integration_tests.yml
```diff
@@ -74,6 +74,10 @@ jobs:
         run: |
           source activate accelerate;
           pytest -sv tests/deepspeed
+      - name: Clean local files
+        if: always()
+        run: |
+          rm -rf accelerate transformers
 
       - name: Run transformers examples tests
         working-directory: transformers/
```
ydshieh, line 83: quick look: this seems a bit odd, where we delete `transformers` but having `working-directory: transformers` ..?

ydshieh (approved):
overall lgtm, but having 2 questions that confuse me ..?

.github/workflows/self_hosted_integration_tests.yml
ydshieh: my 2 comments may apply to multiple places (if they make sense)

.github/workflows/build-docker-images-release.yml
muellerzr: Fixed!

.github/workflows/self_hosted_integration_tests.yml
muellerzr: Yes we now have non-persistent workers so this shouldn't be necessary.

.github/workflows/build-docker-images-release.yml
glegendre01: i agree, but it was docker-cpu...

glegendre01: it's for building image. so maybe no need GPU ?

.github/workflows/build-docker-images-release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
   version-cuda:
     name: "Latest Accelerate GPU [version]"
-    runs-on: [self-hosted, docker-gpu, multi-gpu, gcp]
+    runs-on: [self-hosted, docker-gpu]
```
glegendre01, line 44: do you need GPU here ? if yes also need to use [self-hosted, single-gpu, nvidia-gpu, t4, daily-ci]

.github/workflows/build_docker_images.yml
```diff
@@ -41,8 +31,7 @@ jobs:
 
   latest-cuda:
     name: "Latest Accelerate GPU [dev]"
-    runs-on: [self-hosted, docker-gpu, multi-gpu, gcp]
-    needs: clean-storage
+    runs-on: [self-hosted, docker-gpu]
     steps:
```
glegendre01, line 35: same here

PR #2088, opened 2023-10-26
CRITICAL: fix failing ci

muellerzr:
CI is red when merging because new runners contain the same tags, but are not setup for our CI yet (and as a result, will completely fail). This PR keeps us on the old CI system for now until we are ready to migrate


SunMarc (approved):
LGTM ! Let's revert this commit when we are ready to switch to the new runners

Issue #1430, opened 2023-05-13
FP8 training causes OOM


overvalidated:
When FP8 is utilized model that is loaded in fp16 (llama) OOMs during training. 
Model works perfectly in fp16 mode.
My guess is that autocast of model to TE layers changes dtype to float32.

overvalidated:
In this code new TransfomerEngine Linear module is created but original weights are cloned and new module has param_dtype=torch.float32. It causes OOM when model is casted.
https://github.com/huggingface/accelerate/blob/dcde1e93d09abea02a8e7f4a07a2c5734b87b60e/src/accelerate/utils/transformer_engine.py#L31-L41C19

sgugger:
Indeed, the linear layer needs to be created with the same dtype as the original one. Would you like to suggest a PR with a fix?

overvalidated:
Honestly, I lack knowledge to fix it myself and it still uses the same amount of memory.

avisinghal6:
@sgugger  I would like to work on this PR. I am new to open source contribution, so if I you can guide me on how to work on this PR, I would like to take it up.

overvalidated:
I've tried remove old tensors and avoiding to create new tensors inside of Transfomer Engine (skip_weight_param_allocation), but may be some extra tensors are created inside of transformer engine. Not sure whether it causes OOM or I miss some tensors that should be cleaned.

avisinghal6:
I was just planning to add this to the existing code because we just want to r
<img width="869" alt="Screenshot 2023-05-20 at 11 58 45 AM" src="https://github.com/huggingface/accelerate/assets/97785770/40316e46-08c9-4682-bbb3-d29cb1b61569">
etain the datatype

overvalidated:
After this LLaMA still OOMs during loading. Probably the problem is deeper and I'm not sure now what causes it.

avisinghal6:
Yes, you are right, the parameter is params_dtype (i just wanted to show the idea in the snippet). The OOM error might be due to other causes as you said, the above change is to fix the problem that converted layer should have same dtype as original. @sgugger  kindly advise if I should create a PR for this change.

sgugger:
Yes, you can definitely open a PR with this fix.

avisinghal6:
added PR #1467.

overvalidated:
Still the issue is not resolved. It simply will create tensors in fp16 and replace them with model's ones.

sgugger:
[...]

PR #2089, opened 2023-10-26
Fix memory leak in fp8 causing OOM (and potentially 3x vRAM usage)

muellerzr:
# What does this PR do?

Doing `.copy()` like we were before leads to a *huge* memory leak during `fp8`, and also putting the model on CUDA *after* converting the layers can reduce some memory.

Example benchmark:
Model: `"meta-llama/Llama-2-7b-hf"`
Note: on FP8 weights are in bf16 beforehand
**Before fix**:
Memory in FP32: 25.23 GB
Memory on BF16: 12.61 GB
Memory on FP8: **37.23 GB** <- Warning sign that something is amiss!

**After fix**:
Memory on FP8:
1. Doing `.cuda()` before (current implementation in `.prepare()`): **12.87 GB**
2. Doing `.cuda()` after: **12.61 GB**

Fixes # (issue)

(Finally) fixes https://github.com/huggingface/accelerate/issues/1430

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @LysandreJik

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1372,7 +1372,7 @@ class Accelerator:
                 with torch.no_grad():
                     convert_model(model)
                 model._converted_to_transformer_engine = True
-            model._original_forward = model.forward
+            #model._original_forward = model.forward
 
             kwargs = self.fp8_recipe_handler.to_kwargs() if self.fp8_recipe_handler is not None else {}
             if "fp8_format" in kwargs:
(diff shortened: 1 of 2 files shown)


BenjaminBossan (approved):
Very good find, thanks for investigating and fixing this.

The movement of the code block in `accelerator.py` is presumably due to the order in which `convert_model` needs to be called?

(note for later: the 2nd error message in that block is messed up)

overvalidated:
So that's how it should've been done... Thank you, it's amazing to see this is finally resolved.
Even though I have already traded my 4090 for 3090s xD.

LysandreJik (approved):
Thanks for investigating this @muellerzr!

PR #2090, opened 2023-10-27
Fix flag typo

kuza55:
# What does this PR do?

Doc fix for flag typo

## Before submitting
- [ X ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

@pacman100 @muellerzr


muellerzr (approved):
Thanks for the PR! Can confirm this is the only place it exists.

PR #2104, opened 2023-10-31
Remove explicit copy of `model.forward()` and instead check/store the `id` to unwrap

muellerzr:
# What does this PR do?

This PR solves another bug with fp8 where having a link to the original forward in the model would cause yet another memory spike:

(From the NVIDIA team):
> we still experience OOM in fp8, but in model function call. In fp8, the forward step takes 15 GB of memory, while in bf16, it is 3GB.

The solution is to get rid of `_original_forward_func` (which was shown to be the issue) and to instead rely on the `id` of the function and don't make copies of it/copies of references. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1357,8 +1357,8 @@ class Accelerator:
                 " Please rerun your script specifying `--num_processes=1` or by launching with `python {{myscript.py}}`."
             )
 
+        model._accelerate_original_forward_id = id(model.forward)
         if self.native_amp:
-            model._original_forward = model.forward
             model_forward_func = model.forward.__func__ if hasattr(model.forward, "__func__") else model.forward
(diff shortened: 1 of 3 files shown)


BenjaminBossan (approved):
Nice work identifying and fixing the issue.

Regarding the solution, it isn't really pretty, but there is probably no pretty solution.

What confuses me is why the previous code would lead to a memory spike but not this one. Clearly, it can't be the case that, by holding a reference to the original module's `forward` method, accelerate somehow inhibits garbage collection. Since we can still retrieve the original method through unwrapping, a reference to the original module has to exist.

It could be some kind of circular reference issue, but that should be detected by the GC. I wonder if this points to some kind of more fundamental bug. Also, it's really unclear to me how it related to fp8.

Regarding `id`, I very rarely make us of it. Can we be sure that it will do the right thing? I think it should work, but we should consider all possible edge cases.

PR #2105, opened 2023-10-31
fix warning when offload

SunMarc:
# What does this PR do ? 

This PR fixes the issue related to the offload warning that was introduced in this [PR](https://github.com/huggingface/accelerate/pull/2075).


muellerzr (approved):
Thanks for the fix!

PR #2116, opened 2023-11-02
Fix dtype bug when `offload_state_dict=True` and `dtype` is specified

fxmarty:
As per title.

 I don't have time to provide a full repro & add a test - could do next week.

But for general context, the issue is the following:

When loading a state dict with `offload_state_dict=True`, we go into
https://github.com/huggingface/accelerate/blob/bd72a5f1a80d5146554458823f8aeda0a9db5297/src/accelerate/utils/modeling.py#L1403C18-L1403C18

and we are missing a check similar to https://github.com/huggingface/accelerate/blob/bd72a5f1a80d5146554458823f8aeda0a9db5297/src/accelerate/utils/modeling.py#L292 when the argument `dtype` is specified.

The byproduct is that all parameters are casted to `dtype`, although they should not.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -309,8 +309,13 @@ def set_module_tensor_to_device(
             device = "cpu"
         if value is None:
             new_value = old_value.to(device)
-            if dtype is not None and device in ["meta", torch.device("meta")]:
-                new_value = new_value.to(dtype)
+            if device in ["meta", torch.device("meta")]:
+                if dtype is None:
+                    # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
+                    new_value = new_value.to(old_value.dtype)
+                elif not str(old_value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
+                    new_value = new_value.to(dtype)
+
                 if not is_buffer:
                     module._parameters[tensor_name] = param_cls(new_value, requires_grad=old_value.requires_grad)
         elif isinstance(value, torch.Tensor):


src/accelerate/utils/modeling.py
```diff
@@ -309,8 +309,13 @@ def set_module_tensor_to_device(
             device = "cpu"
         if value is None:
             new_value = old_value.to(device)
-            if dtype is not None and device in ["meta", torch.device("meta")]:
-                new_value = new_value.to(dtype)
+            if device in ["meta", torch.device("meta")]:
+                if dtype is None:
+                    # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
+                    new_value = new_value.to(old_value.dtype)
```
fxmarty, line 315: not sure if this one is needed - it's the `elif` below that solves the issue I have

SunMarc: Indeed you can remove that line as it does nothing. new value is just old_value moved to `device`.

SunMarc, line 315:
```suggestion
```

src/accelerate/utils/modeling.py
```diff
@@ -309,8 +309,13 @@ def set_module_tensor_to_device(
             device = "cpu"
         if value is None:
             new_value = old_value.to(device)
-            if dtype is not None and device in ["meta", torch.device("meta")]:
-                new_value = new_value.to(dtype)
+            if device in ["meta", torch.device("meta")]:
```
SunMarc, line 312:
```suggestion
            if dtype is not None and device in ["meta", torch.device("meta")]:
```

src/accelerate/utils/modeling.py
```diff
@@ -309,8 +309,13 @@ def set_module_tensor_to_device(
...
+                if dtype is None:
+                    # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
+                    new_value = new_value.to(old_value.dtype)
+                elif not str(old_value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
```
SunMarc, line 316:
```suggestion
                if not str(old_value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
```

SunMarc (approved):
LGTM ! Left a few comments. I will run a few tests about bnb cpu offload to see if nothing is broken then merge.

src/accelerate/utils/modeling.py
fxmarty: sound good - will update tomorrow!

src/accelerate/utils/modeling.py
```diff
@@ -1311,7 +1313,7 @@ def get_state_dict_offloaded_model(model: nn.Module):
 def load_checkpoint_in_model(
     model: nn.Module,
     checkpoint: Union[str, os.PathLike],
-    device_map: Optional[Dict[str, Union[int, str, torch.device]]] = None,
```
fxmarty, line 1314: This typing is wrong as `None` is not an accepted value in this function (raises an error at https://github.com/huggingface/accelerate/blob/bd72a5f1a80d5146554458823f8aeda0a9db5297/src/accelerate/utils/modeling.py#L436)

fxmarty: I leave it as is because changing to `     device_map: Dict[str, Union[int, str, torch.device]]` raises errors in the CI.

muellerzr (approved):
Thanks for iterating! Overall looks great to me bar one suggestion. cc @SunMarc for a final look through?

tests/test_modeling_utils.py
```diff
@@ -425,6 +435,21 @@ def test_load_checkpoint_in_model_two_gpu(self):
         self.assertEqual(model.batchnorm.weight.device, torch.device("cpu"))
         self.assertEqual(model.linear2.weight.device, torch.device(1))
 
+    def test_load_checkpoint_in_model_dtype(self):
+        with tempfile.NamedTemporaryFile(suffix=".pt") as tmpfile:
+            print(tmpfile.name)
```
fxmarty, line 440:
```suggestion
```

fxmarty:
Done thank you!

SunMarc (approved):
LGTM ! Thanks for adding a test.

src/accelerate/utils/modeling.py
SunMarc: Thx for noticing this. This should be fixed [here](https://github.com/huggingface/accelerate/pull/2218#issue-2026649653).

PR #2117, opened 2023-11-02
Make SeedableRandomSampler the default always

muellerzr:
# What does this PR do?

Properly sets `SeedableRandomSampler` as the main sampler to be replaced if the user passed in `RandomSampler`. We need this always/should use it always.

Also fixes a missing `set_epoch` at the end of the `__iter__`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -93,6 +93,7 @@ class SeedableRandomSampler(RandomSampler):
                 yield from torch.randint(high=n, size=(32,), dtype=torch.int64, generator=g).tolist()
         else:
             yield from torch.randperm(n, generator=g).tolist()
+        self.set_epoch(self.epoch + 1)
 
     def set_epoch(self, epoch: int):
         "Sets the current iteration of the sampler."
@@ -836,7 +837,7 @@ def prepare_data_loader(
         sampler = getattr(dataloader.sampler, "sampler", None)
     else:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
-    if isinstance(sampler, RandomSampler) and num_processes > 1:
+    if isinstance(sampler, RandomSampler):
         # When iterating through the dataloader during distributed processes
         # we want to ensure that on each process we are iterating through the same
         # samples in the same order if a seed is set. This requires a tweak
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -80,19 +80,12 @@ class SeedableRandomSampler(RandomSampler):
         self.epoch = 0
 
     def __iter__(self):
-        g = torch.Generator()
-        if self.generator is not None:
-            seed = self.epoch + self.generator.initial_seed()
-        else:
-            seed = self.epoch
-        g.manual_seed(seed)
-        n = len(self.data_source)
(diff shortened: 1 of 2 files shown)


src/accelerate/data_loader.py
```diff
@@ -80,19 +80,13 @@ def __init__(self, *args, **kwargs):
...
-                yield from torch.randint(high=n, size=(32,), dtype=torch.int64, generator=g).tolist()
-        else:
-            yield from torch.randperm(n, generator=g).tolist()
+        if self.generator is None:
+            self.generator = torch.Generator()
+        # Allow `self.epoch` to modify the seed of the generator
+        seed = self.epoch + self.generator.initial_seed()
+        self.generator.manual_seed(seed)
+        yield from super().__iter__()
+        self.set_epoch(self.epoch + 1)
```
muellerzr, line 89: Here is what stemmed the original bug/issue, we needed to call `set_epoch` at the end of `__iter__`

BenjaminBossan (commented):
Thanks for being so persistent with this case.

By setting `self.generator`, it is now possible to call `super().__iter__()` and it should do the right thing, which is good.

Just to be completely sure, we now call `set_epoch` inside of `__iter__` because there is no external caller that would do that? For PyTorch's `DistributedSampler`, my understanding is that it's up to the user to call `set_epoch` correctly in each epoch. But here, we consider `SeedableRandomSampler` to only be used internally and no user code should ever call `set_epoch`, is this understanding correct? Does transformers call the method anywhere?

muellerzr:
@BenjaminBossan transformers does internally, and we do when resuming from a checkpoint in general. However if the user wants to avoid doing so, it is done for them here (apart from resuming). In general users *should* call `set_epoch`, however the native pytorch `DataLoader` does not have this capability, thus we still needed to do it manually during `__iter__` (and why the tests were failing)

muellerzr:
For code, the trainer calls it here: https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L1810-L1813

BenjaminBossan:
Okay, so to summarize for myself:

1. Users _should_ call `set_epoch` on the sampler when they resume training to make the run reproducible.
2. When they use their own training loop, they should _not_ call `set_epoch` because the sampler already increments it internally (although I guess calling it twice does not hurt, as long as the correct `epoch` argument is passed).
3. When using transformers, the user also does not need to do anything special during training, because accelerate handles this internally.

muellerzr:
Close!

> Users should call set_epoch on the sampler when they resume training to make the run reproducible. 
Yes, as a next stage I'll look at seeing if we can do so automatically via `load_state` and `save_state`
> When they use their own training loop, they should not call set_epoch because the sampler already increments it internally (although I guess calling it twice does not hurt, as long as the correct epoch argument is passed).
Correct
> When using transformers, the user also does not need to do anything special during training, because accelerate handles this internally.
The `Trainer` does, but also correct.

BenjaminBossan (approved):
Thanks for your patience in explaining the change. LGTM.

LysandreJik (approved):
Thanks for your work @muellerzr!

PR #2120, opened 2023-11-03
Make safetensors the default

muellerzr:
# What does this PR do?

Similar to transformers, makes `safetensors` the default and a library requirement 🤗 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LysandreJik @SunMarc @BenjaminBossan


LysandreJik (approved):
Looks great! Something that we're doing in transformers as well that you might want to do here, is to continue testing for `safe_serialization=False` in the common pathways.

By removing testing of the serialization of `pytorch_model.bin` (which we're doing as the default isn't explicitly defined in the tests and the default just changed), we're risking breaking that code without noticing.

You implemented another test, do you think maybe others would need implementing? That's what we did on `transformers` side (not merged yet): https://github.com/huggingface/transformers/pull/27242

src/accelerate/checkpointing.py
```diff
@@ -174,31 +179,39 @@ def load_accelerator_state(
...
     elif map_location == "on_device":
         map_location = PartialState().device
+
+    input_dir = Path(input_dir)
     # Model states
     for i, model in enumerate(models):
-        weights_name = f"{MODEL_NAME}.bin" if i == 0 else f"{MODEL_NAME}_{i}.bin"
-        input_model_file = os.path.join(input_dir, weights_name)
-        models[i].load_state_dict(torch.load(input_model_file, map_location=map_location), **load_model_func_kwargs)
+        weights_name = f"{MODEL_NAME}.safetensors" if i == 0 else f"{MODEL_NAME}_{i}.safetensors"
```
BenjaminBossan, line 186: Just to be sure I understand correctly, here we are "guessing" that the checkpoint uses safetensors and if it doesn't, we assume it uses the PyTorch format.

src/accelerate/accelerator.py
```diff
@@ -2536,7 +2535,7 @@ def save_model(
         model: torch.nn.Module,
         save_directory: Union[str, os.PathLike],
         max_shard_size: Union[int, str] = "10GB",
-        safe_serialization: bool = False,
+        safe_serialization: bool = True,
```
BenjaminBossan, line 2538: Would it be useful to have a deprecation cycle or can we assume that there is no risk for users when we change the default? Since this is intended for checkpointing, I assume it should be fine.

src/accelerate/checkpointing.py
```diff
@@ -75,45 +76,49 @@ def save_accelerator_state(
...
             if isinstance(sampler, SeedableRandomSampler):
-                save(sampler, output_sampler_file, save_on_each_node=save_on_each_node)
+                save(sampler, output_sampler_file, save_on_each_node=save_on_each_node, safe_serialization=False)
         logger.info(f"Sampler state for dataloader {i} saved in {output_sampler_file}")
 
     # GradScaler state
     if scaler is not None:
         state = scaler.state_dict()
-        output_scaler_file = os.path.join(output_dir, SCALER_NAME)
+        output_scaler_file = output_dir.joinpath(SCALER_NAME)
```
BenjaminBossan, line 121: Is it okay to keep `torch.save` for the grad scaler? Same question for the random state further below.

BenjaminBossan (commented):
In general, this LGTM. I have a few questions for my understanding, the implementation looks clean though.

The whole of `fsdp_utils.py` still relies on `torch.save`, should this also be adjusted?

src/accelerate/checkpointing.py
muellerzr: Correct!

src/accelerate/checkpointing.py
muellerzr: Yes. `safetensors` is only meant for tensors, so thus we should only use this for the model weights.

BenjaminBossan:
> By removing testing of the serialization of `pytorch_model.bin` (which we're doing as the default isn't explicitly defined in the tests and the default just changed), we're risking breaking that code without noticing.

Good point.

> You implemented another test, do you think maybe others would need implementing?

I think this is something where test coverage could be helpful, as it would reveal if we have code paths that are no longer taken but which used to be taken.

src/accelerate/checkpointing.py
BenjaminBossan: Maybe it's worth adding this caveat to the docstring?

src/accelerate/accelerator.py
muellerzr: There shouldn't be a risk, especially because of our fallback in case of different saving techniques

src/accelerate/checkpointing.py
muellerzr: The docstring already states that `safe_serialization` is strictly for the model, unless we want more? 

```
        safe_serialization (`bool`, *optional*, defaults to `True`):
            Whether to save the model using `safetensors` or the traditional PyTorch way (that uses `pickle`).
```

muellerzr:
@LysandreJik @BenjaminBossan I've updated all of our tests that do anything with `save`/saving the model to use both `.bin` and `.safetensors` :)

src/accelerate/checkpointing.py
```diff
@@ -75,45 +76,49 @@ def save_accelerator_state(
...
+        safe_serialization (`bool`, *optional*, defaults to `True`):
+            Whether to save the model using `safetensors` or the traditional PyTorch way (that uses `pickle`).
     """
+    output_dir = Path(output_dir)
     # Model states
     for i, state in enumerate(model_states):
-        weights_name = f"{MODEL_NAME}.bin" if i == 0 else f"{MODEL_NAME}_{i}.bin"
-        output_model_file = os.path.join(output_dir, weights_name)
-        save(state, output_model_file, save_on_each_node=save_on_each_node)
+        weights_name = f"{MODEL_NAME}" if i == 0 else f"{MODEL_NAME}_{i}"
```
SunMarc, line 85: In transformers, the model is named `pytorch_model` for bin format  and `model` for safetensors format. Is this something we want to do in accelerate ? In the `save_model` method, I use `WEIGHTS_NAME` and `SAFE_WEIGHTS_NAME` .

src/accelerate/checkpointing.py
```diff
@@ -75,45 +76,49 @@ def save_accelerator_state(
...
+        save(state, output_model_file, save_on_each_node=save_on_each_node, safe_serialization=safe_serialization)
         logger.info(f"Model weights saved in {output_model_file}")
     # Optimizer states
     for i, opt in enumerate(optimizers):
         state = opt.state_dict()
         optimizer_name = f"{OPTIMIZER_NAME}.bin" if i == 0 else f"{OPTIMIZER_NAME}_{i}.bin"
-        output_optimizer_file = os.path.join(output_dir, optimizer_name)
-        save(state, output_optimizer_file, save_on_each_node=save_on_each_node)
+        output_optimizer_file = output_dir.joinpath(optimizer_name)
+        save(state, output_optimizer_file, save_on_each_node=save_on_each_node, safe_serialization=False)
```
SunMarc, line 95: maybe worth adding a comment saying why we don't use `safetensors` here

src/accelerate/checkpointing.py
SunMarc: +1 for adding this in the docstring

SunMarc (approved):
LGTM ! I've left a few comments.

src/accelerate/checkpointing.py
BenjaminBossan: I see, it is implied by the docstring, I think it wouldn't hurt to make it explicit.

src/accelerate/test_utils/testing.py
```diff
@@ -65,6 +65,18 @@ def parse_flag_from_env(key, default=False):
...
+def parameterized_custom_name_func(func, param_num, param):
+    # customize the test name generator function as we want both params to appear in the sub-test
+    # name, as by default it shows only the first param
+    param_based_name = parameterized.to_safe_name("_".join(str(x) for x in param.args))
+    return f"{func.__name__}_{param_based_name}"
+
+
+SAFETENSORS = True
+PYTORCH = False
+SAVE_TYPES = (SAFETENSORS, PYTORCH)
```
BenjaminBossan, line 77: Why use the generic name `SAVE_TYPES`? I think it would be easier to understand the code if this is just deleted and the test changed like this:

    @parameterized.expand([True, False], name_func=parameterized_custom_name_func)
    def test_save_load_model(self, use_safetensors):

It would make sense to have a `SAVE_TYPES` if there were more than two options, but it's unlikely that `use_safetensors` will ever take values other than `True` or `False`.

BenjaminBossan (approved):
LGTM, thanks Zach. I have a few comments but none are blockers.

src/accelerate/checkpointing.py
muellerzr: We can change the default, and then have a warning for 3 issues if we find others that are using the old name and raise a deprecation

src/accelerate/checkpointing.py
muellerzr: Adding an explicit `<Tip>`

src/accelerate/checkpointing.py
muellerzr: Actually, I think we'll be okay doing this a bit cleaner without deprecation... :)

src/accelerate/checkpointing.py
muellerzr: We'd still have to guess, but we can simplify the logic some :) (As during `load` we don't know which to use)

src/accelerate/test_utils/testing.py
muellerzr: `SAVE_TYPES`/this method of doing it lets the tests be reported as `test_save_load_model_safetensors` and `test_save_load_model_pytorch`. Since we use it in more than one file, brought it out here.

Otherwise it'd be reported as `test_save_load_model_False`

muellerzr: Although that *is* what is actually happening here too, so this abstraction is unnecessary in the end, I'll remove it.

src/accelerate/checkpointing.py
```diff
@@ -54,10 +57,18 @@ def save_accelerator_state(
...
     save_on_each_node: bool = False,
+    safe_serialization: bool = True,
 ):
     """
     Saves the current states of the models, optimizers, scaler, and RNG generators to a given directory.
 
+    <Tip>
+
+    If `safe_serialization` is `True`, models will be saved with `safetensors` while the rest are saved using native
+    `pickle`.
```
BenjaminBossan, line 68: :+1:

tests/test_state_checkpointing.py
```diff
@@ -80,6 +81,14 @@ def forward(self, x):
...
 
 
+def parameterized_custom_name_func(func, param_num, param):
+    # customize the test name generator function as we want both params to appear in the sub-test
+    # name, as by default it shows only the first param
+    param_based_name = "use_safetensors" if param["use_safetensors"] is True else "use_pytorch"
+    return f"{func.__name__}_{param_based_name}"
+
+
+@parameterized_class(("use_safetensors",), [[True], [False]], class_name_func=parameterized_custom_name_func)
```
BenjaminBossan, line 91: TIL about `parameterized_class`.

BenjaminBossan (approved):
Great, LGTM, thanks.

LysandreJik (approved):
Thanks for iterating @muellerzr! LGTM

SunMarc (approved):
Thanks for iterating !

PR #2123, opened 2023-11-05
device agnostic testing

statelesshz:
# What does this PR do?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


statelesshz:
Verified with Ascend NPU:
- System info
  ```
  (mem) [root@localhost accelerate]# accelerate env
  
  - `Accelerate` version: 0.25.0.dev0
  - Platform: Linux-4.14.0-115.el7a.0.1.aarch64-aarch64-with-glibc2.26
  - Python version: 3.8.18
  - Numpy version: 1.24.4
  - PyTorch version (GPU?): 2.1.0 (False)
  - PyTorch NPU available: True
  - System RAM: 2045.09 GB
          Not found
  ```

 - the `spec.py` I used:
    ```
    
    import torch
    import torch_npu
    # User can add additional imports here
    
    # Specify the device name (eg. 'cuda', 'cpu')
    DEVICE_NAME = 'npu'
    
    # Specify device-specific backends to dispatch to.
    # If not specified (i.e., `None`) will fallback to 'default' in 'test_utils/other.py`
    DEVICE_COUNT_FN = torch.npu.device_count
    IS_AVAILABLE_FN = torch.npu.is_available
  
    ```

- the output log:
  ```
  (mem) [root@localhost accelerate]# RUN_SLOW=1 ACCELERATE_TEST_BACKEND="torch_npu" ACCELERATE_TEST_DEVICE="npu" ACCELERATE_TEST_DEVICE_SPEC="spec.py" python -m pytest -v tests/test_metrics.py
  ============================================================================================= test session starts =============================================================================================
  platform linux -- Python 3.8.18, pytest-7.4.3, pluggy-1.3.0 -- /root/miniconda3/envs/mem/bin/python
  cachedir: .pytest_cache
  rootdir: /home/w00668292/accelerate
  collected 4 items
  
  tests/test_metrics.py::MetricTester::test_metric_accelerator SKIPPED (test requires an accelerator)                                                                                                     [ 25%]
  tests/test_metrics.py::MetricTester::test_metric_accelerator_multi PASSED                                                                                                                               [ 50%]
  tests/test_metrics.py::MetricTester::test_metric_cpu_multi SKIPPED (test requires only a CPU)                                                                                                           [ 75%]
  tests/test_metrics.py::MetricTester::test_metric_cpu_noop SKIPPED (test requires only a CPU)                                                                                                            [100%]
  
  =================================================================================== 1 passed, 3 skipped in 99.08s (0:01:39) ===================================================================================
  ```

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,128 @@
...
+
+import torch
+
+from ..logging import get_logger
+from ..state import PartialState
+
+
+# We must initialize the accelerate state before using the logging utility.
+state = PartialState()
+logger = get_logger(__name__)
```
muellerzr, line 28: This is not allowed, any import to `test_utils` will init the PartialState. Get the logger, and then init the state in the test

src/accelerate/test_utils/testing.py
```diff
@@ -85,7 +86,15 @@ def require_cpu(test_case):
     """
     Decorator marking a test that must be only ran on the CPU. These tests are skipped when a GPU is available.
     """
-    return unittest.skipUnless(not torch.cuda.is_available(), "test requires only a CPU")(test_case)
+    return unittest.skipUnless(torch_device == "cpu", "test requires only a CPU")(test_case)
+
+
+def require_accelerator(test_case):
```
muellerzr, line 92:
```suggestion
def require_non_cpu(test_case):
```

src/accelerate/test_utils/testing.py
```diff
@@ -147,6 +156,14 @@ def require_tpu(test_case):
     return unittest.skipUnless(is_tpu_available(), "test requires TPU")(test_case)
 
 
+def require_single_accelerator(test_case):
```
muellerzr, line 159:
```suggestion
def require_single_device(test_case):
```

src/accelerate/test_utils/testing.py
```diff
@@ -163,6 +180,16 @@ def require_single_xpu(test_case):
     return unittest.skipUnless(torch.xpu.device_count() == 1, "test requires a XPU")(test_case)
 
 
+def require_multi_accelerator(test_case):
```
muellerzr, line 183:
```suggestion
def require_distributed(test_case):
```

muellerzr (changes requested):
Thanks, I feel like we're overcomplicating things immensely however. In general all accelerate tests should work on any device setup without needing to modify the code. If we want to enable certain tests to be more generic because we've proven they can be (if they are not, it is simply because we cannot *run them*), then they need to be enabled to be ran generically based on whatever environment we have available. 

Along with this, we should not use `accelerator` here as that would imply `Accelerator()`, which is not the case. I have added some suggestions.

src/accelerate/test_utils/testing.py
statelesshz: In order to correspond to `require_single_device`, I tend to rename `rquire_multi_accelerator` to `require_multi_device`

statelesshz:
> In general all accelerate tests should work on any device setup without needing to modify the code.

Hi @muellerzr, thanks for your advice. I agree with your above point of view. Most test cases can be executed on any device without modification. However, these test cases are usually decorated with cuda-related decorators and cannot be directly run on third-party accelerators such as NPU and XPU. In addition, some test cases use cuda-related interfaces, such as `torch.is_cuda_available`, which may cause the test cases to be unavailable on other devices.

statelesshz:
@muellerzr Good days. This PR has been rebased to the main branch. Could you please re-trigger CI and check this PR again?

src/accelerate/test_utils/__init__.py
```diff
@@ -7,8 +12,12 @@
     require_cuda,
     require_huggingface_suite,
     require_mps,
+    require_multi_device,
     require_multi_gpu,
     require_multi_xpu,
+    require_non_cpu,
+    require_safetensors,
```
muellerzr, line 19: This is no longer a thing
```suggestion
```

statelesshz: done :-)

statelesshz:
All checks have passed🤗

wangshuai09:
Great job! @statelesshz @muellerzr, this job make it easy for me to work and looking forward to the completion of this job.

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,118 @@
...
+
+# Utils for custom and alternative accelerator devices
+
+# Mappings from device names to callable functions to support device agnostic testing.
+BACKEND_DEVICE_COUNT = {"cuda": torch.cuda.device_count, "cpu": lambda: 0, "default": lambda: 1}
+BACKEND_IS_AVAILABLE = {"cuda": torch.cuda.is_available, "cpu": lambda: False, "default": torch.cuda.is_available}
+
+
+# This dispatches a defined function according to the accelerator from the function definitions.
+def _device_agnostic_dispatch(device: str, dispatch_table: Dict[str, Callable], *args, **kwargs):
```
muellerzr, line 55:
```suggestion
def device_agnostic_dispatch(device: str, dispatch_table: Dict[str, Callable], *args, **kwargs):
```

We tend not to have any private functions in accelerate, so let's keep this public even if not used :)

muellerzr (approved):
Thanks! Over all this is looking really, really great! Just one final comment

src/accelerate/test_utils/other.py
statelesshz: done!

statelesshz:
Hi @muellerzr Thanks for your reviewing, all necessary modifications have been done. Can you help trigger the CI? 😄

src/accelerate/test_utils/testing.py
```diff
@@ -146,6 +155,14 @@ def require_tpu(test_case):
     return unittest.skipUnless(is_tpu_available(), "test requires TPU")(test_case)
 
 
+def require_single_device(test_case):
+    """
+    Decorator marking a test that requires a single device. These tests are skipped when there are no accelerator
```
BenjaminBossan, line 160:
```suggestion
    Decorator marking a test that requires a single device. These tests are skipped when there are no accelerators
```

Maybe "hardware accelerators" to disambiguate?

src/accelerate/test_utils/testing.py
```diff
@@ -162,6 +179,16 @@ def require_single_xpu(test_case):
     return unittest.skipUnless(torch.xpu.device_count() == 1, "test requires a XPU")(test_case)
 
 
+def require_multi_device(test_case):
+    """
+    Decorator marking a test that requires a multi-device setup. These tests are skipped on a machine without multiple
+    devices.
+    """
+    return unittest.skipUnless(backend_device_count(torch_device) > 1, "test requires multiple accelerators")(
```
BenjaminBossan, line 187: Again, let's disambiguate "accelerators" here.

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,118 @@
...
+BACKEND_DEVICE_COUNT = {"cuda": torch.cuda.device_count, "cpu": lambda: 0, "default": lambda: 1}
+BACKEND_IS_AVAILABLE = {"cuda": torch.cuda.is_available, "cpu": lambda: False, "default": torch.cuda.is_available}
+
+
+# This dispatches a defined function according to the accelerator from the function definitions.
+def device_agnostic_dispatch(device: str, dispatch_table: Dict[str, Callable], *args, **kwargs):
+    if device not in dispatch_table:
+        return dispatch_table["default"](*args, **kwargs)
+
+    fn = dispatch_table[device]
```
BenjaminBossan, line 59:
```suggestion
    fn = dispatch_table.get(device, dispatch_table["default"])
```

Would this work?

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Some device agnostic functions return values. Need to guard against 'None' instead at
```
BenjaminBossan, line 61: I don't understand that. What would be a practical use case?

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,118 @@
...
+
+import importlib
+import os
+from pathlib import Path
+from typing import Callable, Dict
+
+import torch
+
+
+# Set a backend environment variable for any extra module import required for a custom accelerator
```
BenjaminBossan, line 23: As this (and the other new) env var is not currently being used (AFAICT), could you give a concrete example? Also, should they be documented in the list of env vars @muellerzr? (`package_references/utilities.md`)

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,118 @@
...
+def backend_device_count(device: str):
+    return device_agnostic_dispatch(device, BACKEND_DEVICE_COUNT)
+
+
+def backend_is_available(device: str):
+    return device_agnostic_dispatch(device, BACKEND_IS_AVAILABLE)
+
+
+# Update device function dict mapping
+def update_mapping_from_spec(device_fn_dict: Dict[str, Callable], attribute_name: str):
```
BenjaminBossan, line 79: Would it make sense to use `dispatch_table` instead of `device_fn_dict` as a variable name for consistency?

BenjaminBossan (commented):
I think the general goal is laudable, but I can't say that I understand what's really going on in `test_utils/other.py`. Maybe if I saw how it works in practice, it would be different. To me, this feels like a very specific solution that is overgeneralized, and could maybe be achieved by some monkey patching which doesn't need to live inside of the accelerate repo.

If this is not affecting existing tests, has proven to work in transformers, and Zach is fine with the changes, then please go ahead and merge despite my not understanding.

src/accelerate/test_utils/testing.py
statelesshz: `hardware accelerators` sounds more reasonable 😄

src/accelerate/test_utils/other.py
muellerzr: They shouldn't, only public env variables that users should know and use exist there. This is just for tests so users shouldn't know about this.

src/accelerate/test_utils/other.py
```diff
@@ -0,0 +1,108 @@
...
+    if "ACCELERATE_TEST_DEVICE" in os.environ and torch_device != device_name:
+        msg = f"Mismatch between environment variable `ACCELERATE_TEST_DEVICE` '{torch_device}' and device found in spec '{device_name}'\n"
+        msg += "Either unset `ACCELERATE_TEST_DEVICE` or ensure it matches device spec name."
+        raise ValueError(msg)
+
+    torch_device = device_name
+
+    # Add one entry here for each `BACKEND_*` dictionary.
+    update_mapping_from_spec(BACKEND_DEVICE_COUNT, "DEVICE_COUNT_FN")
+    update_mapping_from_spec(BACKEND_IS_AVAILABLE, "IS_AVAILABLE_FN")
```
muellerzr, line 108: This is never actually used. We don't have an example anywhere in the accelerate repo this is being used. Without this and clear example of what `device_spec_path.py` is meant to be, this feels exceedingly over-engineered.

muellerzr (commented):
One majorly critical comment below.

And another one:

Can we not just do `PartialState().num_processes`? Instead of doing this huge mapping/spec/testing and just let `test` run itself without needing to have this giant configuration that's being added? 

I get we need to bring in the right device type to be used and check the number of them available, however there has to be an easier way than this that we can just pick up from the environment, in a similar vein to what the `PartialState` does. 

As it currently stands there are only 2 patches we check for, and that is `xpu` and `npu`. If tests require specific libraries to use, they should check the `PartialState.distributed_type` instead of just `is_xyz` available as they have been, which is an issue with the current tests, yes.

However adding this complexity for two device types seems like extreme overkill, when we can just natively rely on what already exists in the Accelerate library and just add one util.

For instance, we can replace all of this with one function:

```python
def get_backend():
    if is_mps_available():
        return "mps", 1
        return "cuda", torch.cuda.device_count()
    elif is_xpu_available():
        return "xpu", torch.xpu.device_count()
    else:
        return "cpu", 1
```
And suddenly now all of this complexity is removed. We can also then extract this from the `PartialState` into a separate util and utilize it there.

Can you tell me any disadvantages you see with this? 

It mimics what we already do in accelerate, we *assume from the start* that all tests should be ran on any device (which is how this framework is supposed to be), and we remove any added complexities and configs and rely on just what the environment has.

statelesshz:
Hi, @muellerzr I simplified the code implementation in this PR and removed over-engineered code :)

BenjaminBossan (approved):
Great simplification, thanks for coming up with this version.

muellerzr (commented):
Agreed this is a great simplification, but concerned that the tests are handing. Let's rerun them to make sure something isn't up. Do CPU tests pass locally for you when you do `make test_core`?

muellerzr (approved):
Thanks!!

PR #2125, opened 2023-11-06
fix warmup lr when using deepspeed

### LINKED ISSUE #2124: The warmup lr schedule are not working as expected when using Deepspeed.
Hi, I was finetuning a model with hf trainer and Deepspeed and I found that the warmup lr schedule is not working as expected when using Deepspeed. 

I set the scheduler in Deepspeed config as follows:
```json    
"scheduler": {
      "type": "WarmupLR",
      "params": {
          "warmup_min_lr": 0.0,
          "warmup_max_lr": "auto",
          "warmup_num_steps": "auto",
          "warmup_type": "linear"
      }
    },
```
The accelerate.prepare function will construct a parameter config and return an optimizer with a maximum learning rate: https://github.com/huggingface/accelerate/blob/4f100318f499cff48b62555ca7d96ad97d7cb9be/src/accelerate/accelerator.py#L1665
The above kwargs would be like:
```json
"..."
"optimizer": {
    "type": "AdamW", 
    "params": {
        "lr": 0.0001, # initialized by maximum value 
        "betas": "[...]", 
        "eps": "1e-08", 
        "..."
    }
}
"..."
```
As the lr_scheduler.step() function is called behind by [optimizer.step()](https://github.com/microsoft/DeepSpeed/blob/de96e18ad5c6c58b21b906249296ca45ff46a24f/deepspeed/runtime/engine.py#L2026), the optimizer will use the maximum lr to update the model, cause a severe parameter shift in the first step, which will harm the final fine-tuning performance.

After the learning rate schedular step, the learning rate has been set to the correct value, so this behavior will **not** reflected in the learning rate log.

A workaround to fix this problem is initializing the optimizer with the warmup minimum learning rate.

Hope this problem will be fixed soon, thanks!

Eikor:
# What does this PR do?

Fixes #2124 
Current DeepSpeedEngineWrapper.backwards() calls [DeepSpeedEngine._take_model_step()](https://github.com/microsoft/DeepSpeed/blob/de96e18ad5c6c58b21b906249296ca45ff46a24f/deepspeed/runtime/engine.py#L2026), which performs optimizer step first, followed by lr_scheduler step. https://github.com/huggingface/accelerate/blob/4f100318f499cff48b62555ca7d96ad97d7cb9be/src/accelerate/utils/deepspeed.py#L176  

However, the DS optimizer will be initialized with the maximum learning rate in the prepare function, the optimizer will update model parameters with the maximum learning rate at the first step, which will cause an unexpected behavior. https://github.com/huggingface/accelerate/blob/4f100318f499cff48b62555ca7d96ad97d7cb9be/src/accelerate/accelerator.py#L1665 

This PR solves this problem by initializing the optimizer with warmup_min_lr if the warmup lr scheduler is activated.

@pacman100

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1646,6 +1646,9 @@ class Accelerator:
                     kwargs["model_parameters"] = optimizer.params
                     if isinstance(scheduler, (DummyScheduler)) and scheduler.lr_scheduler_callable is not None:
                         kwargs["lr_scheduler"] = scheduler.lr_scheduler_callable
+                    if isinstance(scheduler, (DummyScheduler)) and 'warmup' in kwargs['config_params']['scheduler']['type'].lower():
+                        # if using warmup, set inintial optimizer learning rate as min lr.
+                        kwargs['config_params']['optimizer']['params']['lr'] = kwargs['config_params']['scheduler']['params']['warmup_min_lr']
                 else:
                     if self.deepspeed_config["zero_optimization"].get("offload_optimizer", {}).get(
                         "device", "none"


muellerzr:
@Eikor can you please do `pip install -e .[quality]; make style; make quality`? This should fix the failing test, thanks!

pacman100:
I don't think this should be the case. Let me deep dive into this.

mjbommar:
Was there a reason this was never merged, @pacman100 and @muellerzr?

PR #2126, opened 2023-11-06
Allow for ACCELERATE_SEED env var

muellerzr:
# What does this PR do?

This PR introduces `ACCELERATE_SEED` as part of `set_seed`. Doing so will allow for the custom random samplers to rely on it for setting the seed, rather than the generator (as this can lead to imperfect issues). Solves related trainer test failures.

Fixes # (issue)

[Internal slack thread](https://huggingface.slack.com/archives/C01NE71C4F7/p1699294942410279)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@LysandreJik @BenjaminBossan

### DIFF
--- a/docs/source/package_reference/utilities.md
+++ b/docs/source/package_reference/utilities.md
@@ -54,7 +54,6 @@ These are basic dataclasses used throughout 🤗 Accelerate and they can be pass
 
 These are environmental variables that can be enabled for different use cases
 
-* `ACCELERATE_SEED` (`int`): The seed to use for different internal random states not covered through the base cases. Also set during [`utils.set_seed`]
 * `ACCELERATE_DEBUG_MODE` (`str`): Whether to run accelerate in debug mode. More info available [here](../usage_guides/debug.md).
 
 ## Plugins
--- a/docs/source/package_reference/utilities.md
+++ b/docs/source/package_reference/utilities.md
@@ -55,7 +55,7 @@ These are basic dataclasses used throughout 🤗 Accelerate and they can be pass
 These are environmental variables that can be enabled for different use cases
 
 * `ACCELERATE_SEED` (`int`): The seed to use for different internal random states not covered through the base cases. Also set during [`utils.set_seed`]
-* `ACCELERATE_DEBUG_MODE` (`str`, ["1","2"]): Whether to run accelerate in debug mode. More info available [here](../usage_guides/debug.md).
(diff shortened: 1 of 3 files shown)


docs/source/package_reference/utilities.md
```diff
@@ -50,6 +50,13 @@ These are basic dataclasses used throughout 🤗 Accelerate and they can be pass
 
 [[autodoc]] utils.ProjectConfiguration
 
+## Environmental Variables
+
+These are environmental variables that can be enabled for different use cases
+
+* `ACCELERATE_SEED` (`int`): The seed to use for different internal random states not covered through the base cases. Also set during [`utils.set_seed`]
```
BenjaminBossan, line 57: The description is a bit vague. As a user, what would interest me is in what circumstances I have to set this env var for reproducibility.

docs/source/package_reference/utilities.md
```diff
@@ -50,6 +50,13 @@ These are basic dataclasses used throughout 🤗 Accelerate and they can be pass
 
 [[autodoc]] utils.ProjectConfiguration
 
+## Environmental Variables
```
BenjaminBossan, line 53: Good idea to document those in a separate section.

src/accelerate/utils/random.py
```diff
@@ -30,7 +31,8 @@
 
 def set_seed(seed: int, device_specific: bool = False):
     """
-    Helper function for reproducible behavior to set the seed in `random`, `numpy`, `torch`.
+    Helper function for reproducible behavior to set the seed in `random`, `numpy`, `torch`. Also will set the random
+    seed for `SeedableRandomSampler` through the `ACCELERATE_SEED` environment variable.
```
BenjaminBossan, line 35: Similar comment to the one above: Users don't really care about `SeedableRandomSampler`, as this is an implementation detail of accelerate. They will care more about knowing when this function needs to be used.

BenjaminBossan (approved):
Implementation-wise, this looks good to me. I think the documentation could be improved, LMK if you agree.

docs/source/package_reference/utilities.md
muellerzr: Went with new implementation, irrelevent :)

BenjaminBossan (approved):
Nice and simple. So in the end, users should call `set_seed` to ensure reproducibility and that's enough, right?

muellerzr:
@BenjaminBossan yep! set_seed, and in the Trainer just having it set the torch seed will be enough for general users

PR #2131, opened 2023-11-07
Add examples to tests

muellerzr:
# What does this PR do?

Adds trainer integration tests (the examples) to our test suite since they test both the `Trainer` and the `Accelerator`

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -69,4 +69,5 @@ jobs:
       run: |
         cd ../transformers/examples/pytorch
         pip install -r _tests_requirements.txt
-        pytest -sv test_accelerate_examples.py test_pytorch_examples.py
+        cd ../../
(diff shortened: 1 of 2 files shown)


pacman100:
A general query I had. Would this mean the same tests are being run by both Transformers CI as well as Accelerate CI? That would be redundant compute spent which seems suboptimal.

pacman100 (commented):
A general query I had. Would this mean the same tests are being run by both Transformers CI as well as Accelerate CI? That would be redundant compute spent which seems suboptimal.

muellerzr:
Yes, but they are in our main CI logs so we can pick them up early,we don’t have to rely on transformers. It’s why we have the integration test CI

BenjaminBossan (approved):
Thanks. I agree that as long as those are not hugely expensive tests, it's a worthwhile addition to gain visibility.

muellerzr:
They add ~10 min to the PR CI, hence why they are just in merge and nightlies (as it should take less time than that).

PR #2132, opened 2023-11-07
Raise error when saving with param on meta device

SunMarc:
# what does this PR do ?
This PR fixes the `save_model` method when a user tries to use it on a model with some parameters on the `meta` device. I didn't check the `hf_device_map` since it is not the only case where a user might move the model to the meta device. However, the warning that we have put in the dispatch function should be enough to let the user know that some parameters were offloaded.


src/accelerate/accelerator.py
```diff
@@ -2580,6 +2580,10 @@ def save_model(
 
         os.makedirs(save_directory, exist_ok=True)
 
+        for param in model.parameters():
+            if param.device == torch.device("meta"):
+                raise RuntimeError("You can't save the model since some parameters are on the meta device.")
```
muellerzr, line 2585:
```suggestion
        if any(param.device == torch.device("meta") for param in model.parameters()):
            raise RuntimeError("You can't save the model since some parameters are on the meta device.")
```

muellerzr (approved):
LG2M! Small suggestion for efficiency :) Thanks for the blocker

src/accelerate/accelerator.py
SunMarc: better indeed =)

pacman100 (approved):
Thank you @SunMarc for adding this!

src/accelerate/accelerator.py
```diff
@@ -2580,6 +2580,9 @@ def save_model(
 
         os.makedirs(save_directory, exist_ok=True)
 
+        if any(param.device == torch.device("meta") for param in model.parameters()):
```
BenjaminBossan, line 2583: Just a nit, how about moving this check before creating the directory, that way it's not created when saving fails.

BenjaminBossan (approved):
Thanks for this update, it should hopefully help many users who are currently confused. I only have a nit, up to you if you want to fix it.

PR #2136, opened 2023-11-09
Deal with shared memory scenarios

muellerzr:
# What does this PR do?

We need to deal with shared memory scenarios when saving with `safetensors` and a model. Since the `safetensor` version relies on passing the *entire model* instead of the state dict directly, [see here](https://github.com/huggingface/safetensors/blob/main/bindings/python/py_src/safetensors/torch.py#L130), to deal with this, for now a raw copy/paste of the code is done instead to remove duplicate names.

(All `state_dict` in torch are `OrderedDict`, hence the check)

Fixes # (issue)
Two failures on the transformers CI

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -14,7 +14,6 @@
 
 from __future__ import annotations
 
-import collections
 import contextlib
 import functools
 import json
@@ -64,6 +63,7 @@ from .utils import (
     RNGType,
     TorchDynamoPlugin,
     check_os_kernel,
+    clean_state_dict_for_safetensors,
     compare_versions,
     convert_model,
     convert_outputs_to_fp32,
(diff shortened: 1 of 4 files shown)


SunMarc (approved):
LGTM ! I know that in the `save_model` in accelerate and in `save_pretrained` in transformers, we remove the shared tensors in a different [way](https://github.com/huggingface/accelerate/blob/9674b40580bcfb598e4d79950bad9bf9a3f83252/src/accelerate/accelerator.py#L2585). I don't know how comparable it is to this method in safetensors but my guess it that it is pretty similar.

muellerzr:
Good call, I’ll refactor this into a function so that it only warns, doesn’t actually raise an err :)

I think logically it should work the same, bar the bnb explicit part

SunMarc (approved):
Thx for iterating ! LGTM !

PR #2137, opened 2023-11-09
fix retie_parameters

SunMarc:
# What does this PR do ?

This PR fixes the `retie_parameters` function. The issue is that we were making the assumption that the first elem of the tied_params was the param to tie. This should solves the BNB tests that are failing on the CI because we switched to safetensors.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -535,15 +535,22 @@ def retie_parameters(model, tied_params):
     """
     for tied_group in tied_params:
         param_to_tie = None
-        # First iteration of the loop will set param_to_tie, next ones will tie it to the others
+        # two loops : the first one to set param_to_tie , the second one to change the values of tied_group
         for param_name in tied_group:
             module = model
             splits = param_name.split(".")
             for split in splits[:-1]:
                 module = getattr(module, split)
-            if param_to_tie is None:
-                param_to_tie = getattr(module, splits[-1])
-            else:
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
Thanks!

BenjaminBossan (approved):
Looks good, thanks.

A question: Something that can happen now is that no parameters are being tied, right? (when `param_to_tie is None` after finishing the first loop.) Can this be problematic?

Also, just a tip, code like this can be simplified with the `attrgetter` function from the `operator` module. E.g. in the first loop:

```python
        for param_name in tied_group:
            param = attrgetter(param_name)(module)
            if param_to_tie is ...
```

should work.

In the second loop, a bit of extra work is required:

```python
            for param_name in tied_group:
                mod, _, param = param_name.rpartition(".")
                module = attrgetter(mod)(model)
                setattr(module, param, param_to_tie)
```

No need to change here, just something I find neat.

SunMarc:
> A question: Something that can happen now is that no parameters are being tied, right? (when param_to_tie is None after finishing the first loop.) Can this be problematic?

If param_to_tie is None, it means that all the tensors are on the meta device and they should also have the same shape/dtype since they were tied before (here it is a retie function). So I think it should be fine. 

> Also, just a tip, code like this can be simplified with the attrgetter function from the operator module. E.g. in the first loop:

Indeed! Thx for sharing, I will use that more often from now =)

PR #2139, opened 2023-11-09
Adds dvclive tracker

dberenbaum:
# What does this PR do?

Adds a [DVCLive](https://dvc.org/doc/dvclive) tracker. DVCLive is a logger for [DVC](https://dvc.org), which version data, models, metrics, etc. and keeps them connected to your Git repo. They can then be visualized in several different interfaces.

A [PR](https://github.com/huggingface/transformers/pull/27352) was recently merged to add a logger to `transformers`, and this PR adds support for `accelerate`.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr From what I can tell, it looks like you would be the relevant reviewer. Please feel free to redirect me otherwise.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -30,6 +30,7 @@ from .utils import (
     is_aim_available,
     is_clearml_available,
     is_comet_ml_available,
+    is_dvclive_available,
     is_mlflow_available,
     is_tensorboard_available,
     is_wandb_available,
@@ -56,6 +57,9 @@ if is_mlflow_available():
 
 if is_clearml_available():
     _available_trackers.append(LoggerType.CLEARML)
+
+if is_dvclive_available():
+    _available_trackers.append(LoggerType.DVCLIVE)
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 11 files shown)


muellerzr (approved):
Excellent! Very beautiful PR, thanks for this! One last little thing, can you add this to the `test_trackers` section of the `setup.py` so the tests can be ran?

dberenbaum:
> Excellent! Very beautiful PR, thanks for this! One last little thing, can you add this to the `test_trackers` section of the `setup.py` so the tests can be ran?

Thanks for catching that! Added it in the latest commit.

dberenbaum:
Will take a look at the test failures

src/accelerate/tracking.py
```diff
@@ -837,13 +841,78 @@ def _get_title_series(name):
...
+
+        Args:
+            values (Dictionary `str` to `str`, `float`, or `int`):
+                Values to be logged as key-value pairs. The values need to have type `str`, `float`, or `int`.
+            step (`int`, *optional*):
+                The run step. If included, the log will be affiliated with this step.
+            kwargs:
+                Additional key word arguments passed along to `dvclive.Live.log_metric()`.
+        """
+        if step:
```
BenjaminBossan, line 895: What if `step == 0`. Should it not still be set? I.e., would it be safer to check `if step is not None`?

BenjaminBossan (commented):
Very clean implementation, thanks for that. I have only some small comments about usage and setting the step, the rest looks good.

dberenbaum:
@BenjaminBossan Thanks for the catching those issues and the quick review! All should be resolved now.

BenjaminBossan (approved):
Thanks a lot for addressing my comments, LGTM.

muellerzr (approved):
Thanks for all your hard work and iterating with us over this!

tests/test_tracking.py
```diff
@@ -473,3 +484,41 @@ def test_log(self):
                     "some_string": "",
                 }
                 self.assertDictEqual(data, truth)
+
+
+@require_dvclive
+@mock.patch("dvclive.live.get_dvc_repo", return_value=None)
```
muellerzr, line 490: @dberenbaum FYI we're seeing failures on the transformers side in the example scripts because of this. Is there a better way to ignore the git repo than patching for cases where we're running a full training script? (Like via an env variable or something). Otherwise I'll have to disable dvc for all of the example script tests in `transformers` which wouldn't be ideal. https://github.com/huggingface/accelerate/actions/runs/7023901505/job/19111477258

dberenbaum: I need to take a closer look when I have a minute, but you can use `DVCLIVE_TEST = true` to disable the git-related functionality. For example:

https://github.com/huggingface/accelerate/blob/31c723febe6d37f2c318c90bb6bebec64c8df035/tests/test_examples.py#L208

muellerzr: Ah I missed that one, fantastic! That should do the job just fine I think. I’ll get to that tommorow and cc you. Thanks!

dberenbaum: Is https://github.com/huggingface/accelerate/pull/2196 related?

muellerzr: It’s not, we saw dependency errors in an isolated env related to the release

dberenbaum: Thanks. Added a comment there since it should be fixed now. Let me know if it would be helpful to submit a PR for either or if it's quicker to take care of it yourself.

PR #2140, opened 2023-11-10
specify config file path on README

kwonmha:
I had hard time figuring out how to run on multi-node configuration with the given instructions.
It looked like no configuration file argument is required and accelerate will find the generated configuration automatically.
Just few more phrases are needed.

@muellerzr , @pacman100

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -149,8 +149,8 @@ To run it in each of these various modes, use the following commands:
 - multi GPUs (using PyTorch distributed mode)
     * With Accelerate config and launcher
         ```bash
-        accelerate config  # This will create a config file on your server
-        accelerate launch --config_file generated_config_file ./cv_example.py --data_dir path_to_data  # This will run the script on your server
+        accelerate config --config_file config.yaml  # This will create a config file on your server to `config.yaml`
+        accelerate launch --config_file config.yaml ./cv_example.py --data_dir path_to_data  # This will run the script on your server
         ```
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -150,7 +150,7 @@ To run it in each of these various modes, use the following commands:
     * With Accelerate config and launcher
         ```bash
         accelerate config  # This will create a config file on your server
```
muellerzr, line 152:
```suggestion
        accelerate config  # This will create a config file on your server at `~/.cache/huggingface/accelerate/default_config.yaml`
```

muellerzr, line 152:
```suggestion
        accelerate config --config_file config.yaml  # This will create a config file on your server to `config.yaml`
```

muellerzr (changes requested):
Good idea, just needs some tweaks. Let's change this to have `accelerate config` explicitly save it to a location if we want to do so. Otherwise the latter command should be modified to show *where* the config coming from. These commands need to be ran quickly via copy/paste for new users. So let's go with my suggestion here and propagate it below :)

kwonmha:
Good suggestion. I applied your suggestion to all proper positions. So the commands would look like as I modified.

muellerzr:
@kwonmha the `data_dir` path is fine, we care about everything up-to the arguments being passed to file

muellerzr (approved):
Thanks!

Issue #2148, opened 2023-11-13
notebook_launcher fails when loading model from_pretrained 


nikrruun:
### System Info

```Shell
- `Accelerate` version: 0.25.0.dev0
- Platform: Linux-6.2.0-36-generic-x86_64-with-glibc2.35
- Python version: 3.9.18
- Numpy version: 1.25.0
- PyTorch version (GPU?): 2.0.1+cu117 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 2015.52 GB
- GPU type: NVIDIA A40
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: 1,2
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Run this snippet in a Jupyter Notebook on a multi-gpu machine:

```python
from accelerate import notebook_launcher, Accelerator
from transformers import AutoModelForCausalLM

def a():
    state = Accelerator()
    print(state.process_index, state.device)

model = AutoModelForCausalLM.from_pretrained("gpt2")

notebook_launcher(a, args=(), num_processes=2)
```

**Output**
```python
RuntimeError: CUDA has been initialized before the `notebook_launcher` could create a forked subprocess. This likely stems from an outside import causing issues once the `notebook_launcher()` is called. Please review your imports and test them when running the `notebook_launcher()` to identify which one is problematic and causing CUDA to be initialized.
```

### Expected behavior

The code above without ```model = AutoModelForCausalLM.from_pretrained("gpt2")``` gives:
```terminal
Launching training on 2 GPUs.
0 cuda:0
1 cuda:1

```

Any suggestions?

muellerzr:
Pretty sure AutoModel will move the model to CUDA. I would check that, and keep that under your a() function if so.

[...]

PR #2151, opened 2023-11-14
Add warning for problematic libraries

muellerzr:
# What does this PR do?

Since this is a **very** common issue, adds a check in `notebook_launcher` for specific libraries that are *known* to cause issues (such as starting CUDA forks *on import*). 

Fixes https://github.com/huggingface/accelerate/issues/2148

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/launchers.py
+++ b/src/accelerate/launchers.py
@@ -19,7 +19,7 @@ import tempfile
 import torch
 
 from .state import AcceleratorState, PartialState
-from .utils import PrecisionType, PrepareForLaunch, is_mps_available, patch_environment
+from .utils import PrecisionType, PrepareForLaunch, are_libraries_initialized, is_mps_available, patch_environment
 
 
 def test_launch():
@@ -142,6 +142,18 @@ def notebook_launcher(
                     "inside your training function. Restart your notebook and make sure no cells initializes an "
                     "`Accelerator`."
                 )
+            # Check for specific libraries known to initialize CUDA that users constantly use
+            already_initialized = are_libraries_initialized("bitsandbytes")
+            if any(already_initialized.values()):
+                err = (
+                    "Could not start distributed process. Libraries known to initialize CUDA upon import have been "
+                    "imported already. Please keep these imports inside your `function` to try and help with this:"
+                )
+                for lib_name, initialized in already_initialized.items():
+                    if initialized:
+                        err += f"\n\t* `{lib_name}`"
+                raise AssertionError(err)
+
             # torch.distributed will expect a few environment variable to be here. We set the ones common to each
             # process here (the other ones will be set be the launcher).
             with patch_environment(
--- a/src/accelerate/launchers.py
+++ b/src/accelerate/launchers.py
@@ -143,16 +143,15 @@ def notebook_launcher(
                     "`Accelerator`."
                 )
             # Check for specific libraries known to initialize CUDA that users constantly use
-            already_initialized = are_libraries_initialized("bitsandbytes")
-            if any(already_initialized.values()):
+            problematic_imports = are_libraries_initialized("bitsandbytes")
+            if len(problematic_imports) > 1:
                 err = (
                     "Could not start distributed process. Libraries known to initialize CUDA upon import have been "
-                    "imported already. Please keep these imports inside your `function` to try and help with this:"
+                    "imported already. Please keep these imports inside your training function to try and help with this:"
                 )
-                for lib_name, initialized in already_initialized.items():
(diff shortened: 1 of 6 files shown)


tests/test_multigpu.py
```diff
@@ -66,23 +70,16 @@ def test_distributed_data_loop(self):
         with patch_environment(omp_num_threads=1, cuda_visible_devices="0,1"):
             execute_subprocess_async(cmd, env=os.environ.copy())
 
-    # Need to see why this test raises forking issues when ran as a suite
-    @skip
```
muellerzr, line 70: Essentially the `notebook_launcher` doesn't like using async subprocesses when doing things, so executing it fully as a non-async subprocess makes this fully work reliably.

src/accelerate/test_utils/scripts/test_notebook.py
```diff
@@ -1,17 +1,34 @@
...
+    with raises(AssertionError, match="Please keep these imports"):
+        notebook_launcher(basic_function, (), num_processes=NUM_PROCESSES)
 
 
-if __name__ == "__main__":
-    notebook_launcher(function, num_processes=int(args.num_processes))
+def main():
+    print("Test basic notebook can be ran")
+    test_can_initialize()
+    if is_bnb_available():
```
BenjaminBossan, line 32: This is so that bnb is getting imported?

src/accelerate/launchers.py
```diff
@@ -142,6 +142,18 @@ def train(*args):
...
+            already_initialized = are_libraries_initialized("bitsandbytes")
+            if any(already_initialized.values()):
+                err = (
+                    "Could not start distributed process. Libraries known to initialize CUDA upon import have been "
+                    "imported already. Please keep these imports inside your `function` to try and help with this:"
+                )
+                for lib_name, initialized in already_initialized.items():
+                    if initialized:
+                        err += f"\n\t* `{lib_name}`"
+                raise AssertionError(err)
```
BenjaminBossan, line 155: Not sure about `AssertionError`, I thought they are basically reserved for `assert`s. Maybe use `ImportError` or `RuntimeError`?

src/accelerate/utils/environment.py
```diff
@@ -48,3 +50,12 @@ def parse_flag_from_env(key, default=False):
 def parse_choice_from_env(key, default="no"):
     value = os.environ.get(key, str(default))
     return value
+
+
+def are_libraries_initialized(library_names: list = []) -> Dict[str, bool]:
```
BenjaminBossan, line 55:
```suggestion
def are_libraries_initialized(library_names: list[str]) -> Dict[str, bool]:
```

Alternatively, to avoid the `isinstance` check below, just do `*library_names`, then we know it's always a tuple.

src/accelerate/utils/environment.py
```diff
@@ -48,3 +50,12 @@ def parse_flag_from_env(key, default=False):
...
+    """
+    Checks if any of `library_names` are imported in the environment. Will return results as a `key:bool` pair.
+    """
+    if isinstance(library_names, str):
+        library_names = [library_names]
+    return {lib_name: lib_name in sys.modules for lib_name in library_names}
```
BenjaminBossan, line 61: The code here and at the call site could be simplified if only the list of loaded libraries were returned, instead of returning all libs in a dict with their status as value. But maybe you plan on using this for other purposes later which requires all the results, so not sure.

src/accelerate/launchers.py
```diff
@@ -142,6 +142,18 @@ def train(*args):
                     "inside your training function. Restart your notebook and make sure no cells initializes an "
                     "`Accelerator`."
                 )
+            # Check for specific libraries known to initialize CUDA that users constantly use
+            already_initialized = are_libraries_initialized("bitsandbytes")
+            if any(already_initialized.values()):
+                err = (
+                    "Could not start distributed process. Libraries known to initialize CUDA upon import have been "
+                    "imported already. Please keep these imports inside your `function` to try and help with this:"
```
BenjaminBossan, line 150: Do you think it's clear what "`function`" means in this context?

BenjaminBossan (commented):
Good idea to make this an explicit error, given how many users stumble upon this. I only have some smaller comments.

src/accelerate/launchers.py
muellerzr: Fixed to keep align with the rest of the err's in `notebook_launcher`

PR #2156, opened 2023-11-16
Assemble state dictionary for offloaded models

blbadger:
# What does this PR do?

Fixes `transformers` issue [#20072](https://github.com/huggingface/transformers/issues/20072), as discussed on PR [#27412](https://github.com/huggingface/transformers/pull/27412) for that library. 

To allow for saving of models with offloaded parameters, all offloaded modules of a model are iteratively onloaded in order to assemble an appropriate state dictionary, with placeholder parameters ignored.

## Before submitting
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [x ] Did you make sure to update the documentation with your changes? Here are the
- [ x] Did you write any new necessary tests?

## Who can review?
@SunMarc
@muellerzr


src/accelerate/accelerator.py
```diff
@@ -2579,7 +2581,14 @@ def save_model(
         os.makedirs(save_directory, exist_ok=True)
 
         # get the state_dict of the model
-        state_dict = self.get_state_dict(model)
+        if (
+            hasattr(model, "_hf_hook")
+            and isinstance(model._hf_hook, AlignDevicesHook)
+            and True in {module._hf_hook.offload for _, module in model.named_modules() if hasattr(module, "_hf_hook")}
+        ):
```
SunMarc, line 2588:
```suggestion
            and any(module._hf_hook.offload for module in model.modules() if hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook))
        ):
```
I've simplified a bit. Also only `AlignDevicesHook` have the offload attribute.

src/accelerate/accelerator.py
```diff
@@ -2579,7 +2581,14 @@ def save_model(
         os.makedirs(save_directory, exist_ok=True)
 
         # get the state_dict of the model
-        state_dict = self.get_state_dict(model)
+        if (
+            hasattr(model, "_hf_hook")
+            and isinstance(model._hf_hook, AlignDevicesHook)
```
SunMarc, line 2586: I don't think we need this since it only tells us if the model has a hook or not.

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1233,50 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            continue
+        if hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload:
+            original_device = module._hf_hook.execution_device
+            # assign hook execution device to cpu
+            module._hf_hook.execution_device = "cpu"
+            # onload meta tensors to execution device
+            try:
+                module._hf_hook.pre_forward(module)
+            except MemoryError:
+                print("Model must fit in CPU memory to call save_pretrained!")
```
SunMarc, line 1257:
```suggestion
                raise MemoryError("Offloaded module must fit in CPU memory to call save_model!")
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1233,50 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+        for key in module_state_dict:
+            # ignore placeholder parameters that are still on the meta device
+            if str(module_state_dict[key].device) == "meta":
+                placeholders.append(key)
+                continue
+            params = module_state_dict[key]
+            state_dict[name + f".{key}"] = params
+
+    if placeholders:
+        warnings.warn(f"The following modules contain unsaved placeholder tensors: {placeholders}")
```
SunMarc, line 1275: We prefer to use logger.warning to let the user control their log better. Also, I've modified slightly the message.
```suggestion
    if placeholders:
        logger.warning(f"The following tensors were not saved because they were still on meta device: {placeholders}")
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1233,50 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            module._hf_hook.post_forward(module, torch.tensor([]))
+            # re-assign hook to original execution device
+            module._hf_hook.execution_device = original_device
+        else:
+            module_state_dict = module.state_dict()
+
+        for key in module_state_dict:
+            # ignore placeholder parameters that are still on the meta device
+            if str(module_state_dict[key].device) == "meta":
+                placeholders.append(key)
```
SunMarc, line 1269: Easier to find the problematic tensor for debug purposes. 
```suggestion
                placeholders.append(name + f".{key}")
```

tests/test_utils.py
```diff
@@ -226,3 +228,30 @@ def forward(self, x):
                 save(model.state_dict(), save_path, safe_serialization=True)
                 self.assertEqual(len(log.records), 1)
                 self.assertIn("Removed shared tensor", log.output[0])
+
+    @require_cuda
+    def test_offload_save_state_dict(self):
+        device_map = {
```
SunMarc, line 234: You don't need to add this test. You can simply change the following test https://github.com/huggingface/accelerate/blob/cd515812483aedd070933c3367366a0b5ef43daa/tests/test_accelerator.py#L160 by removing the assertRaises since this PR will enable saving offloaded model !

src/accelerate/accelerator.py
```diff
@@ -2579,7 +2581,14 @@ def save_model(
         os.makedirs(save_directory, exist_ok=True)
```
SunMarc, line 2582: Can you move L2578-L2580 (checking if any param is on meta device) in the else block ?

tests/test_utils.py
```diff
@@ -22,6 +22,7 @@
 
 import torch
 from torch import nn
+from transformers import AutoModelForCausalLM
```
SunMarc, line 25: to remove

tests/test_utils.py
```diff
@@ -31,6 +32,7 @@
     convert_outputs_to_fp32,
     extract_model_from_parallel,
     find_device,
+    get_state_dict_offloaded_model,
```
SunMarc, line 35: to remove

SunMarc (commented):
We are practically there! Thanks again @blbadger for working on this. I've left a couple of comments. It should be good once the test pass.

src/accelerate/accelerator.py
blbadger: Perfect, that simplifies the condition nicely. Removed

src/accelerate/utils/modeling.py
blbadger: That is a better message, committed

tests/test_utils.py
blbadger: removed

tests/test_utils.py
blbadger: removed

tests/test_utils.py
blbadger: Must have missed this test, that works very well. The state dict test is removed and the assertRaises has been removed too.

src/accelerate/accelerator.py
blbadger: Oops I missed that, thanks for catching! Moved to the else block before the `self.get_state_dict` call

blbadger:
Thanks very much for your review and suggestions @SunMarc ! They were very helpful, and the changes have been committed.  Let me know if there is anything else we need and I would be happy to add.

blbadger:
@muellerzr circular import has been addressed, although the solution is a bit hacky so I am open to suggestions for improvement. The approach was to remove the `AlignDevicesHook` import entirely and check each module's hook's string representation rather than the instance itself.

We were getting the error in `test_save_model_offload` where the loaded model does not have a device alignment hook, so does not trigger `get_state_dict_offloaded_model`.  This was apparently because the model needs to be dispatched and the test has been modified accordingly.

One final test is failing: `test_pre_forward_hook_is_executed` fails a `torch.allclose`, although this is apparently a precision problem as the test passes after changing `atol=1e-5` to `atol=1e-4`.  I would be happy to make the change so this PR passes all tests if that is appropriate!

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,54 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            # offload meta tensors from cpu
+            module._hf_hook.post_forward(module, torch.tensor([]))
+            # re-assign hook to original execution device
+            module._hf_hook.execution_device = original_device
+        else:
+            module_state_dict = module.state_dict()
+
+        for key in module_state_dict:
+            # ignore placeholder parameters that are still on the meta device
+            if str(module_state_dict[key].device) == "meta":
```
SunMarc, line 1270: For consistency 
```suggestion
            if module_state_dict[key].device == torch.device("meta"):
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,54 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            The offloaded model we want to save
+    """
+    state_dict = {}
+    placeholders = []
+    for name, module in model.named_modules():
+        if name == "":
+            continue
+        if (
+            hasattr(module, "_hf_hook")
+            and str(module._hf_hook).startswith("AlignDevicesHook")
```
SunMarc, line 1249: Let's try to fix the import issue =) 
```suggestion
            and isinstance(module._hf_hook,AlignDevicesHook)
```

SunMarc (commented):
Thanks for iterating on this ! Let's just fix this import issue. I think the easier solution would be to import `AlignDevicesHook` in the function itself instead of the beginning of the file.

blbadger:
> Thanks for iterating on this ! Let's just fix this import issue. I think the easier solution would be to import `AlignDevicesHook` in the function itself instead of the beginning of the file.

Works very nicely, thanks!

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,52 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            continue
+        if hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload:
+            original_device = module._hf_hook.execution_device
+            # assign hook execution device to cpu
+            module._hf_hook.execution_device = "cpu"
+            # onload meta tensors to execution device
+            try:
+                module._hf_hook.pre_forward(module)
+            except MemoryError:
+                raise MemoryError("Offloaded module must fit in CPU memory to call save_model!")
```
muellerzr, line 1257:
```suggestion
            except MemoryError:
                raise MemoryError("Offloaded module must fit in CPU memory to call save_model!") from None
```

To help avoid the double error message here :)

muellerzr (approved):
Thanks for iterating! One small comment so the stack trace can be a bit cleaner :)

SunMarc (approved):
Thanks for all your work !

blbadger:
My pleasure, thanks for all your help @SunMarc and @muellerzr !

tests/test_accelerator.py
```diff
@@ -156,9 +156,8 @@ def test_save_model_offload(self, use_safetensors):
         model = ModelForTest()
```
SunMarc, line 156: Let's compare the output to have a better test 

```suggestion
        inputs = torch.randn(3, 3)
        model = ModelForTest()
        expected = model(inputs)
```

tests/test_accelerator.py
```diff
@@ -156,9 +156,8 @@ def test_save_model_offload(self, use_safetensors):
...
         with tempfile.TemporaryDirectory() as tmp_dir:
             accelerator.save_model(model, tmp_dir, safe_serialization=use_safetensors)
-            load_checkpoint_in_model(model, tmp_dir, device_map=device_map, offload_folder=tmp_dir)
-            with self.assertRaises(RuntimeError):
-                accelerator.save_model(model, tmp_dir, safe_serialization=use_safetensors)
+            load_checkpoint_and_dispatch(model, tmp_dir, device_map=device_map, offload_folder=tmp_dir)
+            accelerator.save_model(model, tmp_dir, safe_serialization=use_safetensors)
```
SunMarc, line 160:
```suggestion
            # load and save offloaded model 
            load_checkpoint_and_dispatch(model, tmp_dir, device_map=device_map, offload_folder=tmp_dir)
            accelerator.save_model(model, tmp_dir, safe_serialization=use_safetensors)
            
            # load weights that were saved from the offloaded model
            load_checkpoint_and_dispatch(model, tmp_dir)
            output = model(inputs)
        self.assertTrue(torch.allclose(expected, output, atol=1e-5))
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,52 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+    Returns the state dictionary for an offloaded model via iterative onloading
+
+    Args:
+        model (`torch.nn.Module`):
+            The offloaded model we want to save
+    """
+    from ..hooks import AlignDevicesHook
+
+    state_dict = {}
+    placeholders = []
```
SunMarc, line 1245: Use a `set` to not have duplicates
```suggestion
    placeholders = set()
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,52 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+            module._hf_hook.post_forward(module, torch.tensor([]))
+            # re-assign hook to original execution device
+            module._hf_hook.execution_device = original_device
+        else:
+            module_state_dict = module.state_dict()
+
+        for key in module_state_dict:
+            # ignore placeholder parameters that are still on the meta device
+            if module_state_dict[key].device == torch.device("meta"):
+                placeholders.append(name + f".{key}")
```
SunMarc, line 1269:
```suggestion
                placeholders.add(name + f".{key}")
```

src/accelerate/utils/modeling.py
```diff
@@ -1231,6 +1231,52 @@ def load_state_dict(checkpoint_file, device_map=None):
...
+                continue
+            params = module_state_dict[key]
+            state_dict[name + f".{key}"] = params
+
```
SunMarc, line 1273: We need to remove some key that are not on meta device. This happens because we iterate though the modules and we have sub modules in the model. 
```suggestion
    for key in placeholders.copy():
        if key in state_dict:
            placeholders.remove(key)
```

SunMarc (approved):
I've tested the PR and everything looks good. I suggested some modifications to have a better test and fix the placeholder issue (saying that some parameters were not saved when it is the case). This happens because we iterate through the modules and we can have a situation where the parent module is not offloaded but some of the children modules are. Thus, we need to do a final check and remove the keys that are indeed in the `state_dict`. I will merge this PR when the suggestions will be addressed. I've tested with a big model:
```py
from transformers import AutoTokenizer, AutoModelForCausalLM

accelerator = Accelerator()
device_map = {'model.embed_tokens': 0, 'model.layers.0': 0, 'model.layers.1': 0, 'model.layers.2': 0, 'model.layers.3': 0, 'model.layers.4': 0, 'model.layers.5': 0, 'model.layers.6': 1, 'model.layers.7': 1, 'model.layers.8': 1, 'model.layers.9': 1, 'model.layers.10': 1, 'model.layers.11': 1, 'model.layers.12': 1, 'model.layers.13': 2, 'model.layers.14': 2, 'model.layers.15': 2, 'model.layers.16': 2, 'model.layers.17': 2, 'model.layers.18': 2, 'model.layers.19': 2, 'model.layers.20': 3, 'model.layers.21': 3, 'model.layers.22': 3, 'model.layers.23': 3, 'model.layers.24': 3, 'model.layers.25': 3, 'model.layers.26': 3, 'model.layers.27': 4, 'model.layers.28': 4, 'model.layers.29': "disk", 'model.layers.30': "disk", 'model.layers.31': 'cpu', 'model.norm': 'cpu', 'lm_head': 'cpu'}
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-hf")
model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-hf",device_map=device_map,offload_folder="llama")
prompts = ["I would like to"]
token_dict = tokenizer(prompts, return_tensors="pt").to(1)
output_ids = model.generate(**token_dict, max_new_tokens=20)
print(tokenizer.batch_decode(output_ids))

accelerator.save_model(model, "llama", safe_serialization=True)
model.config.save_pretrained("llama")
model_offload = AutoModelForCausalLM.from_pretrained("llama",device_map=device_map,offload_folder="llama")
output_ids = model_offload.generate(**token_dict, max_new_tokens=20)     
print(tokenizer.batch_decode(output_ids))
```

blbadger:
@SunMarc thanks for the review and for testing on a big model! All suggestions have been committed.  It makes sense that a parent module would not need to be onloaded if the child modules are, thanks for clearing that up.

PR #2158, opened 2023-11-16
[Docs] add distributed dataset mapping usage

kashif:
# What does this PR do?

This PR adds a usage guide on disitributed mapping a function on datasets

### DIFF
--- a/docs/source/usage_guides/datasets.md
+++ b/docs/source/usage_guides/datasets.md
@@ -1,0 +1,59 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Distributed datasets mapping with 🤗 Accelerate
+
+Distributed mapping a function which to run over a dataset is a common use case, especially with natural language processing (NLP) models and datasets.  Typically one needs to augment a dataset with extra columns coming from some expensive operation.
+
+## The Problem
+
+Sometimes one needs to augment a dataset with extra columns coming from some expensive operation.  For example, one might want to add a column to a dataset with logits or embeddings from a model.  In such a usecase one might want to use a distributed mapping function to speed up the process.
+
+## The Solution
+
+🤗 Accelerate provides a way to distribute the dataset and mapping model via the `prepare` method.  This method will wrap the dataset and model with the appropriate distributed wrappers.  
+
+```python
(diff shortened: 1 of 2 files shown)


docs/source/usage_guides/datasets.md
```diff
@@ -0,0 +1,59 @@
...
+from datasets import load_dataset
+
+def map_to_dataset(batch):
+    # do some expensive operation using the model
+    value = model(batch["input"])
+    return value
+
+dataset = load_dataset("my example")
+dataloader = DataLoader(dataset, batch_size=8)
+model = MyModel()
```
muellerzr, line 40: Not entirely sure this will work? This makes the user think you don't need to tensorize your dataset, which will lead to many, many issues during DDP.

muellerzr (commented):
Thanks, overall this looks good. A few nits.

docs/source/usage_guides/datasets.md
```diff
@@ -0,0 +1,59 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Distributed datasets mapping with 🤗 Accelerate
+
+Distributed mapping a function which to run over a dataset is a common use case, especially with natural language processing (NLP) models and datasets.  Typically one needs to augment a dataset with extra columns coming from some expensive operation.
```
MKhalusova, line 18: I think we can clarify the first sentence here, e.g. "Distributed mapping involves running a function over a dataset, a common scenario, especially in the context of natural language processing (NLP) models and datasets."

docs/source/usage_guides/datasets.md
```diff
@@ -0,0 +1,59 @@
...
+
+## The Problem
+
+Sometimes one needs to augment a dataset with extra columns coming from some expensive operation.  For example, one might want to add a column to a dataset with logits or embeddings from a model.  In such a usecase one might want to use a distributed mapping function to speed up the process.
+
+## The Solution
+
+🤗 Accelerate provides a way to distribute the dataset and mapping model via the `prepare` method.  This method will wrap the dataset and model with the appropriate distributed wrappers.  
+
+```python
```
MKhalusova, line 28: The example is very abstract. It may be helpful to illustrate this use case with a specific example like the one you mention earlier - add a column with embeddings to a dataset.

docs/source/usage_guides/datasets.md
```diff
@@ -0,0 +1,59 @@
...
+rendered properly in your Markdown viewer.
+-->
+
+# Distributed datasets mapping with 🤗 Accelerate
+
+Distributed mapping a function which to run over a dataset is a common use case, especially with natural language processing (NLP) models and datasets.  Typically one needs to augment a dataset with extra columns coming from some expensive operation.
+
+## The Problem
+
+Sometimes one needs to augment a dataset with extra columns coming from some expensive operation.  For example, one might want to add a column to a dataset with logits or embeddings from a model.  In such a usecase one might want to use a distributed mapping function to speed up the process.
```
MKhalusova, line 22:
```suggestion
Sometimes one needs to augment a dataset with extra columns coming from some expensive operation.  For example, one might want to add a column to a dataset with logits or embeddings from a model.  In such a use case, one might want to use a distributed mapping function to speed up the process.
```

MKhalusova (commented):
Overall it's a great use case to highlight, but I think this doc can be improved a little bit :)

Issue #2175, opened 2023-11-21
Using a leaked loop index when saving FSDP optimizers with accelerator.save_state()?


xhan77:
Hello, it seems that the [code](https://github.com/huggingface/accelerate/blob/0e51680994f5b9085de615604661d9022439bd0b/src/accelerate/accelerator.py#L2758C103-L2758C103) referenced below that saves the FSDP optimizers uses an optimizer index `i` leaked from a previous loop (which saves the models). Shouldn't this be problematic if we have multiple FSDP optimizers/models? (@pacman100)

https://github.com/huggingface/accelerate/blob/0e51680994f5b9085de615604661d9022439bd0b/src/accelerate/accelerator.py#L2758C103-L2758C103

xhan77:
Thanks for the prompt reply and fix!

PR #2177, opened 2023-11-21
fsdp refactoring

pacman100:
# What does this PR do?
FSDP refactogin based on:
1. Torch 2.1 official docs: https://pytorch.org/docs/stable/fsdp.html
2. with `use_orig_params=True`, we no longer require preparing `model` before creating optimizer object. Earlier, we needed to prepare model, i.e., wrap the model with FSDP before creating optimizer object because of below warning from PyTorch official docs:
```
The optimizer must be initialized after the module has been wrapped with FSDP since FSDP will shard and transform the module’s parameters in a way that may not preserve the original parameter variables. Thus, the previously initialized optimizer may have stale references to the parameters.
```
Now, with `use_orig_params=True`, it is no longer the case. This makes the Accelerate training APi consistent, i.e., users using single GPU, DDP, FSDP, DeepSpeed now need to follow the same logic as below:
```
model, optimizer, lr_scheduler, train_dataloader, eval_dataloader = accelerator.prepare(model, optimizer, lr_scheduler, train_dataloader, eval_dataloader)
```

Earlier, for FSDP, the recommended practice was shown as below. Else we used to receate the optimizer post preparing the model and it didn't preserve optimizer groups. Now, all that is resolved. Now, optimizer groups are also supported.
```

optim = torch.optim.AdamW(model.parameters(), lr=lr)
scheduler = ...
optimizer, lr_scheduler, train_dataloader, eval_dataloader = accelerator.prepare(optimizer, lr_scheduler, train_dataloader, eval_dataloader)
```
As such, `use_orig_params=True` is now the default.

3. https://github.com/facebookresearch/llama-recipes: Using this as best practices guide for FSDP, we are inline with it for all the features and usage of FSDP APIs. For checkpointing, they support `FULL_STATE_DICT` and `SHARDED_STATE_DICT`. We are also supporting both of these and already have tests for it. They don't show how to save and load for `LOCAL_STATE_DICT` state dict type.
4. Regression: A test for `LOCAL_STATE_DICT` checkpointing feature of FSDP is now failing. Couldn't find anything about it in llama recipes, FSDP documentation, torch FSDP codebase https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp as well as on the internet. Will raise an issue with PyTorch team regarding it.
5. Ran all the slow tests for FSDP and all green! Updated documentation and examples.


src/accelerate/accelerator.py
```diff
@@ -2753,7 +2670,7 @@ def _inner(folder):
         # Save the optimizers taking care of FSDP and DeepSpeed nuances
         optimizers = []
         if self.distributed_type == DistributedType.FSDP:
-            for opt in self._optimizers:
+            for i, opt in enumerate(self._optimizers):
```
BenjaminBossan, line 2673: Was this a bug in the old code?

src/accelerate/accelerator.py
```diff
@@ -3068,6 +2985,18 @@ def get_state_dict(self, model, unwrap=True):
                 from deepspeed.checkpoint.utils import clone_tensors_for_torch_save
 
                 state_dict = clone_tensors_for_torch_save(self.unwrap_model(model).state_dict())
+        elif self.distributed_type == DistributedType.FSDP:
+            from torch.distributed.fsdp import (
```
BenjaminBossan, line 2989: These imports should fit on one line.

BenjaminBossan (commented):
This looks like a great change, love to see so many lines deleted.

I don't have experience with FSDP, so a few questions:

1. Does this still work as expected when using PyTorch < 2.1?
2. `use_orig_params` default was changed to True. Is there any disadvantage to that, e.g. more memory usage?

src/accelerate/accelerator.py

src/accelerate/accelerator.py
pacman100: Yes, this was brought to our notice via https://github.com/huggingface/accelerate/issues/2175

pacman100:
Hello Benjamin,

> Does this still work as expected when using PyTorch < 2.1?
Accelerate will throw an error when FSDP integration is used with PyTorch < 2.1 now because of the lines below. 

https://github.com/huggingface/accelerate/blob/0e51680994f5b9085de615604661d9022439bd0b/src/accelerate/accelerator.py#L306-L307

BenjaminBossan:
> Accelerate will throw an error when FSDP integration is used with PyTorch < 2.1

Ah okay, I missed the version bump, thanks for pointing me to it.

> It is expected to become default as per the above dev blogpost

Thanks for providing more context. My question arose because `use_orig_params=True` seems to be strictly better and I wondered why it wasn't the default. I wondered if there is any disadvantage, but it seems to be mainly just for backwards compatibility in PyTorch.

muellerzr (approved):
_**Nicely**_ done @pacman100! Excellent refactor and loving that diff. Keeping the simplistic API all around is a phenomenal win!

BenjaminBossan (approved):
Great work, thanks Sourab.

PR #2180, opened 2023-11-22
Better error when device mismatches when calling gather() on CUDA

muellerzr:
# What does this PR do?

This PR adds a new explicit error when a user tries to call `.gather()` in a GPU scenario and the device of the passed in tensor != the device in `PartialState` (aka CUDA). Avoids users getting err:

```python
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 2897, in all_gather_into_tensor
    return honor_type(
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/accelerate/utils/operations.py", line 83, in honor_type
    return type(obj)(generator)
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/accelerate/utils/operations.py", line 112, in <genexpr>
    recursively_apply(
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/accelerate/utils/operations.py", line 128, in recursively_apply
    return func(data, *args, **kwargs)
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/accelerate/utils/operations.py", line 307, in _gpu_gather_one
    gather_op(output_tensors, tensor)
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/torch/distributed/c10d_logger.py", line 47, in wrapper
    return func(*args, **kwargs)
  File "/home/student/Experiemnts/MultiLabelClassification_LLMs/.mlc/lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 2897, in all_gather_into_tensor
    work = group._allgather_base(output_tensor, input_tensor)
RuntimeError: Tensors must be CUDA and dense
    work = group._allgather_base(output_tensor, input_tensor)
RuntimeError: Tensors must be CUDA and dense
```

And instead gives them a much clearer err:

```python
"One or more of the tensors passed to `gather` was not on the GPU while the `Accelerator` is configured for CUDA. "
                "Please move it to the GPU before calling `gather`."
```

Fixes # (issue)

https://discuss.huggingface.co/t/problem-with-model-inference-using-accelerate/63078

## Before submitting
[...]

PR #2199, opened 2023-11-29
Add  allgather check for xpu

abhilash1910:
Related to #2180 . There can be an alternate design as (cc @muellerzr  Let me know your thoughts.) :
```python
if state.device.type != tensor.device.type :
     if state.device.type=="cuda:"
           <print error>
     
     elif ......<logic for other devices>
            <print error>
```

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -298,7 +298,6 @@ def _gpu_gather(tensor):
         if not tensor.is_contiguous():
             tensor = tensor.contiguous()
 
-
         if state.backend is not None and state.backend != "gloo":
             # We use `empty` as `all_gather_into_tensor` slightly
             # differs from `all_gather` for better efficiency,
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -298,6 +298,13 @@ def _gpu_gather(tensor):
         if not tensor.is_contiguous():
             tensor = tensor.contiguous()
 
+        # Check if `tensor` is not on XPU
+        if state.device.type == "xpu" and tensor.device.type != "xpu":
+            raise RuntimeError(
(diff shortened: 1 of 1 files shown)


abhilash1910:
Makes sense, we can directly modify verify_operation by adding additional common check  for any device ?

abhilash1910:
@muellerzr could you please re-trigger CI and review? As of yet I placed it inside verify_operation as the wrapper is already called in gather method.  Thanks

src/accelerate/utils/operations.py
```diff
@@ -351,6 +344,12 @@ def wrapper(*args, **kwargs):
             tensor = kwargs["tensor"]
         else:
             tensor = args[0]
+        state = PartialState()
```
muellerzr, line 347: No, we already instantiate the state here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/operations.py#L287C27-L287C27

muellerzr: (And probably why the tests are failing)

abhilash1910: Yes makes sense, I modified it to use the existing PartialState() instead. Thanks

muellerzr (approved):
This is looking much better to me! Nice job

src/accelerate/utils/operations.py
```diff
@@ -351,6 +344,11 @@ def wrapper(*args, **kwargs):
             tensor = kwargs["tensor"]
         else:
             tensor = args[0]
+        if PartialState().device.type != tensor.device.type:
```
muellerzr, line 347: `tensor` isn't guaranteed to be a tensor at this point, hence our failures. (e.g. could be a dict, list, etc). Let's use `find_device` instead so that it should work fine.

```suggestion
        if PartialState().device.type != find_device(tensor).type:
```

muellerzr (commented):
One note for our failures here

PR #2204, opened 2023-12-01
Allow deepspeed without distributed launcher

pacman100:
# What does this PR do?
1. Allow deepspeed without distributed launcher. Fixes https://github.com/huggingface/transformers/issues/27770


muellerzr (approved):
LG2M as long as the user reports back it works too.

PR #2207, opened 2023-12-02
[deepspeed plugin] fails when DL uses batch_sampler

stas00:
currently if the DL doesn't have `batch_size` set, because it uses `batch_sampler` the deepspeed plugin code fails to work.

The problem comes from here:

https://github.com/huggingface/accelerate/blob/77628ec6033b47e489054b383953378288526391/src/accelerate/accelerator.py#L1418

so even if the user acted on the exception and set `AcceleratorState().deepspeed_plugin.deepspeed_config['train_micro_batch_size_per_gpu']` this code still fails. 

Here is a possible quick fix that overcomes this problem.

But I suspect that the original logic is a bit off. Wouldn't it be better if it did this check instead:

```
if deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"] == "auto":
    if is_dataloader_present:
        current logic
    else:
        raise ValueError("can't derive train_micro_batch_size_per_gpu - either use a DL with batch_size attribute returning an int or set train_micro_batch_size_per_gpu manually")
```

@pacman100, @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1421,16 +1421,19 @@ class Accelerator:
                 for obj in args
             ]
 
-            batch_sizes = [obj.batch_size for obj in args if hasattr(obj, "batch_size")]
-            if self.split_batches:
-                batch_sizes = [batch_size // self.num_processes for batch_size in batch_sizes]
+            if deepspeed_plugin.deepspeed_config["train_micro_batch_size_per_gpu"] == "auto":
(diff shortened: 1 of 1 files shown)


stas00:
@pacman100, what needs to be done to get this PR merged? Thank you!

pacman100 (commented):
Hello Stas, I agree with your suggestion and have raised the above PR with those changes and made you the co-author of it. Please let us know if that resolves the concerns of this PR.

PR #2212, opened 2023-12-04
Check notebook launcher for 3090+

muellerzr:
# What does this PR do?

Similar to prior work done in https://github.com/huggingface/accelerate/pull/2195, I noticed when using the `notebook_launcher` this would need to be done here as well. Rather than having the user manually set these values themselves, modifies the check to include a torch-save version of checking GPU names (by just calling `nvidia-smi` under the hood).

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -182,14 +182,6 @@ class PartialState:
                         self.backend = "nccl"
                     dist.init_distributed(dist_backend=self.backend, auto_mpi_discovery=False, **kwargs)
 
-                if not check_cuda_p2p_ib_support():
-                    if "NCCL_P2P_DISABLE" not in os.environ or "NCCL_IB_DISABLE" not in os.environ:
-                        raise NotImplementedError(
-                            "Using RTX 3090 or 4000 series doesn't support faster communication broadband via P2P or IB. "
-                            'Please set `NCCL_P2P_DISABLE="1"` and `NCCL_IB_DISABLE="1" or use `accelerate launch` which '
-                            "will do this automatically."
-                        )
-
                 self.num_processes = torch.distributed.get_world_size()
                 self.process_index = torch.distributed.get_rank()
                 self.local_process_index = int(os.environ.get("LOCAL_RANK", -1))
@@ -206,6 +198,13 @@ class PartialState:
                         self.device = torch.device("cuda", self.local_process_index)
                         if self.device is not None:
                             torch.cuda.set_device(self.device)
+                if self.device.type == "cuda" and not check_cuda_p2p_ib_support():
+                    if "NCCL_P2P_DISABLE" not in os.environ or "NCCL_IB_DISABLE" not in os.environ:
+                        raise NotImplementedError(
+                            "Using RTX 3090 or 4000 series doesn't support faster communication broadband via P2P or IB. "
+                            'Please set `NCCL_P2P_DISABLE="1"` and `NCCL_IB_DISABLE="1" or use `accelerate launch` which '
+                            "will do this automatically."
+                        )
                 self._mixed_precision = "no"  # deepspeed handles mixed_precision using deepspeed_config
             elif int(os.environ.get("LOCAL_RANK", -1)) != -1 and not cpu and torch.cuda.is_available():
                 self.distributed_type = DistributedType.MULTI_GPU
--- a/src/accelerate/utils/environment.py
+++ b/src/accelerate/utils/environment.py
@@ -97,9 +97,13 @@ def check_cuda_p2p_ib_support():
     """
     try:
         device_names, device_count = get_gpu_info()
-        unsupported_devices = ["RTX 3090", "RTX 40"]
+        unsupported_devices = {"RTX 3090", "RTX 40"}
         if device_count > 1:
-            if any(device in device_names for device in unsupported_devices):
+            if any(
+                unsupported_device in device_name
+                for device_name in device_names
+                for unsupported_device in unsupported_devices
+            ):
                 return False
     except Exception:
         pass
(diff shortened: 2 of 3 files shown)


src/accelerate/state.py
```diff
@@ -206,6 +198,13 @@ def __init__(self, cpu: bool = False, **kwargs):
                         self.device = torch.device("cuda", self.local_process_index)
                         if self.device is not None:
                             torch.cuda.set_device(self.device)
+                if self.device.type == "cuda" and not check_cuda_p2p_ib_support():
```
BenjaminBossan, line 201: Good idea to check for cuda

src/accelerate/utils/environment.py
```diff
@@ -61,17 +62,41 @@ def are_libraries_initialized(*library_names: str) -> Dict[str, bool]:
...
     Checks if the devices being used have issues with P2P and IB communications, namely any consumer GPU hardware after
     the 3090.
+
+    Noteably uses `nvidia-smi` instead of torch to not initialize CUDA.
     """
-    if torch.cuda.is_available():
-        # Get the first device/default
-        device_name = torch.cuda.get_device_name()
-        device_count = torch.cuda.device_count()
+    try:
```
BenjaminBossan, line 94: Should cuda availability still be checked? Sure, the `try ... except` takes care, but do we ever need to run this if cuda is not available?

BenjaminBossan (commented):
Thanks for expanding on this issue.

> modifies the check to include a torch-save version of checking GPU names (by just calling nvidia-smi under the hood)

Could you explain this a bit more? At first glance, the old method looks more robust.

src/accelerate/utils/environment.py
muellerzr: The issue is we can’t check if CUDA is available. If we do, we’ll hit the CUDA was initialized issue because any call to the `torch.cuda` module will init it

muellerzr: And thus leading to an error about CUDA being initialized before the notebook launcher could be ran

src/accelerate/utils/environment.py
```diff
@@ -61,17 +62,41 @@ def are_libraries_initialized(*library_names: str) -> Dict[str, bool]:
...
+        command = spawn.find_executable("nvidia-smi")
+        if command is None:
+            command = "%s\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe" % os.environ["systemdrive"]
+    else:
+        command = "nvidia-smi"
+    output = subprocess.check_output([command, "-L"], universal_newlines=True)
+    lines = output.split(os.linesep)
+    gpu_count = len(lines) - 1
+    # Get name from first GPU
+    gpu_name = lines[0].split("(")[0].strip()
```
BenjaminBossan, line 83: Here we check only the first GPU, but theoretically, the user could have multiple different devices. Since we cannot know at this point which one will actually be used, shouldn't we check all of them?

src/accelerate/utils/environment.py
muellerzr: Nice! This is a much, *much* cleaner implementation!

src/accelerate/utils/environment.py
muellerzr: Went ahead and did so with your new version

src/accelerate/utils/environment.py
```diff
@@ -61,17 +62,44 @@ def are_libraries_initialized(*library_names: str) -> Dict[str, bool]:
...
+    else:
+        command = "nvidia-smi"
+    # Returns as list of `n` GPUs and their names
+    output = subprocess.check_output(
+        [command, "--query-gpu=count,name", "--format=csv,noheader"], universal_newlines=True
+    )
+    gpus = output.split(os.linesep)
+    # Get names from output
+    gpu_count = len(gpus)
+    gpu_names = [gpu.split(",")[1].strip() for gpu in gpus]
```
BenjaminBossan, line 86: When running locally, this line failed because `output` had a trailing `\n`, resulting in `gpus == ['1, NVIDIA GeForce RTX 2060', '']`. Maybe strip the output before splitting?

src/accelerate/utils/environment.py
```diff
@@ -61,17 +62,44 @@ def are_libraries_initialized(*library_names: str) -> Dict[str, bool]:
...
-    if torch.cuda.is_available():
-        # Get the first device/default
-        device_name = torch.cuda.get_device_name()
-        device_count = torch.cuda.device_count()
+    try:
+        device_names, device_count = get_gpu_info()
         unsupported_devices = ["RTX 3090", "RTX 40"]
         if device_count > 1:
-            if any(device in device_name for device in unsupported_devices):
+            if any(device in device_names for device in unsupported_devices):
```
BenjaminBossan, line 101: Since `device_names` is a list of str, this is no longer a substring match, right? So for instance "RTX 40" would never match here.

BenjaminBossan (changes requested):
Thanks for updating the PR. I had some issues running the new code on my machine, please check the comments.

src/accelerate/utils/environment.py
muellerzr: Done!

src/accelerate/utils/environment.py
muellerzr: Correct. UPdated with better set logic

BenjaminBossan (approved):
Thanks, LGTM.

SunMarc (approved):
Thanks for fixing this !

PR #2214, opened 2023-12-04
Enable cpu offload with weights inside the module

SunMarc:
# What does this PR do ? 
This PR adds the possibility to perform cpu offload with the weights stored inside the module. You just need to pass `cpu_offload = True` in the `dispatch_model`. (not sure about the naming of the arg as it can be confusing) 

Before this PR, all offloaded modules were placed on the `meta` device and the weights were either stored in a dict (cpu offload) or a mmap (disk offload). We would then move the modules to the execution device with their respective value taken from the dict/mmap during the `forward`. 

For the user, this seems a little counter intuitive to put weights in a `dict` in the cpu offload case. Moreover, letting these weights on the modules should not degrade the performance during inference at all. Offloading created a number of issues about the parameters being on the `meta` device. While this does not completely solves issues related to `meta` device, this should cover most cases users don't use disk offload that much. 

For now, the default value is `False` but I would like to make it the default behavior + extend it to Transformers. LMK if this make sense. 

cc @mfuntowicz since you had an issue with offloaded model + quantization
cc @LysandreJik for visibility
Solves partially [1190](https://github.com/huggingface/peft/pull/1190)

TODO:
- Fix a memory leak somewhere

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -412,7 +412,7 @@ def dispatch_model(
 
         # warn if there is any params on the meta device
         offloaded_devices_str = " and ".join(
-            [device for device in set(device_map.values()) if device in ("cpu", "disk")]
+            [device for device in set(device_map.values()) if device in offloaded_devices]
         )
         if len(offloaded_devices_str) > 0:
             logging.warning(
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -302,6 +302,7 @@ def dispatch_model(
     offload_dir: Optional[Union[str, os.PathLike]] = None,
     offload_index: Optional[Dict[str, str]] = None,
     offload_buffers: bool = False,
+    cpu_offload: bool = False,
     skip_keys: Optional[Union[str, List[str]]] = None,
     preload_module_classes: Optional[List[str]] = None,
     force_hooks: bool = False,
@@ -321,6 +322,8 @@ def dispatch_model(
             `"disk"`.
         state_dict (`Dict[str, torch.Tensor]`, *optional*):
             The state dict of the part of the model that will be kept on CPU.
+        cpu_offload (`bool`, *optional*, defaults to `False`):
+            Whether the weights offloaded on the cpu should be kept in the module or not.
         offload_dir (`str` or `os.PathLike`):
             The folder in which to offload the model weights (or where the model weights are already offloaded).
         offload_index (`Dict`, *optional*):
@@ -358,7 +361,7 @@ def dispatch_model(
             else:
                 main_device = [d for d in device_map.values() if d not in ["cpu", "disk"]][0]
 
-        if main_device != "cpu":
+        if main_device != "cpu" and not cpu_offload:
             cpu_modules = [name for name, device in device_map.items() if device == "cpu"]
             if state_dict is None and len(cpu_modules) > 0:
                 state_dict = extract_submodules_state_dict(model.state_dict(), cpu_modules)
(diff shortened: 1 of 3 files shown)


tests/test_big_modeling.py
```diff
@@ -338,6 +338,24 @@ def test_dispatch_model_with_non_persistent_buffers(self):
...
+        device_map = {"linear1": "disk", "batchnorm": "cpu", "linear2": 0}
+
+        x = torch.randn(2, 3)
+        expected = model(x)
+
+        with TemporaryDirectory() as tmp_dir:
+            dispatch_model(model, device_map, offload_dir=tmp_dir, cpu_offload=True)
+
+            self.assertEqual(model.linear1.weight.device, torch.device("meta"))
+            self.assertEqual(model.batchnorm.weight.device, torch.device("cpu"))
```
BenjaminBossan, line 356: The new behavior of getting "cpu" here instead of "meta" looks more intuitive to me.

src/accelerate/big_modeling.py
```diff
@@ -302,6 +302,7 @@ def dispatch_model(
     offload_dir: Optional[Union[str, os.PathLike]] = None,
     offload_index: Optional[Dict[str, str]] = None,
     offload_buffers: bool = False,
+    cpu_offload: bool = False,
```
BenjaminBossan, line 305: In general, should newly added parameters be placed last in case someone calls this function with purely positional arguments?

src/accelerate/hooks.py
```diff
@@ -230,12 +233,12 @@ def __init__(
     ):
         self.execution_device = execution_device
         self.offload = offload
+        self.cpu_offload = cpu_offload
```
BenjaminBossan, line 236: Would it make sense to pass both `offload` and `cpu_offload`? It seems the former would take precedent over the latter. Maybe this could be checked or documented?

src/accelerate/hooks.py
```diff
@@ -310,7 +317,9 @@ def post_forward(self, module, output):
                 if type(module).__name__ == "Linear8bitLt":
                     module.state.SCB = None
                     module.state.CxB = None
-
+        elif self.cpu_offload:
+            for name, _ in named_module_tensors(module, recurse=self.place_submodules):
+                set_module_tensor_to_device(module, name, "cpu")
```
BenjaminBossan, line 322: Is special handling for `Linear8bitLt` required, similar to above?

BenjaminBossan (commented):
Thanks for tackling this annoying issue. Overall, I understand too little about the mechanism used by accelerate to control this, so I'd leave that part of the review to others.

> This PR adds the possibility to perform cpu offload with the weights stored inside the module.

Sorry for being dense, but where exactly is that happening?

> not sure about the naming of the arg as it can be confusing

Yes, I'd definitely rename the argument, especially since already have a `cpu_offload` function in the same file.

tests/test_big_modeling.py
SunMarc: Yes this is what we are aiming in this PR ! We want to let the module on `cpu` and not on `meta` device.

src/accelerate/big_modeling.py
fxmarty, line 305: I think it can be confusing as CPU offloading is already indicated in the `device_map`.

IMO ideally there should not be any argument added, and by default the weights of modules offloaded on RAM should be on `cpu` device, not `meta`. it this is kind of a breaking change in case anybody is assuming that by default attached weights are on `meta` and `weights_map` holds the true weights.

PR #2218, opened 2023-12-05
Add condition before using `check_tied_parameters_on_same_device`

SunMarc:
# What does this do ?
As pointed by @fxmarty [here](https://github.com/huggingface/accelerate/pull/2116#discussion_r1415338368), we need to check if the `device_map` is `None` or not before using `check_tied_parameters_on_same_device`.


muellerzr (approved):
Thanks!

PR #2223, opened 2023-12-06
fix tqdm wrapper to print when process id ==0

kashif:
# What does this PR do?

Fix the condition on when to disable the tqdm output


muellerzr (approved):
Thanks for the fix!

PR #2224, opened 2023-12-06
Integrate MS-AMP Support for FP8 Precision

muellerzr:
# Integrate MS-AMP to the `Accelerator`

## What does this add?

This PR introduces an additional backend for FP8 support through [MS-AMP](https://github.com/Azure/MS-AMP/tree/main) which has shown to decrease memory and increase throughput when using FP8 precision

## Who is it for?

Individuals training with FP8 (H100/4090's, etc)

## Issues linked to

https://github.com/Azure/MS-AMP/issues/128

### User-facing:

Two new arguments were added to the `FP8RecipeKwargs`:
* `enable_ms_amp` (`bool`): Whether a user should use MS-AMP. True by default if it's available in the environment
* `optimization_level` (`str`), should be one of `"O1"` or `"O2"`. `"O3"` is for DeepSpeed and we need to wait for them to update to v0.9.3 of deepspeed to match what Accelerate supports

General guideline to optimization levels:
  * O1: Weight gradients and `all_reduce` communications are done in fp8, reducing GPU
      memory usage and communication bandwidth
  * O2: First-order optimizer states are in 8-bit, and second order states are in FP16.
      Only available when using Adam or AdamW. This maintains accuracy and can potentially save the highest
      memory.
  * 03: Specifically for DeepSpeed, implements capabilities so weights and master weights of models
      are stored in FP8. If `fp8` is selected and deepspeed is enabled, will be used by default.
      (Not available currently).
      
As a result, `"O2"` is the default. Here is an overview of each optimization level and what it does, taken from their [docs](https://azure.github.io/MS-AMP/docs/user-tutorial/optimization-level):

Optimization Level | Computation(GEMM) | Comm | Weight | Master Weight | Weight Gradient | Optimizer States
-- | -- | -- | -- | -- | -- | --
FP16 AMP | FP16 | FP32 | FP32 | N/A | FP32 | FP32+FP32
Nvidia TE | FP8 | FP32 | FP32 | N/A | FP32 | FP32+FP32
MS-AMP O1 | FP8 | FP8 | FP16 | N/A | FP8 | FP32+FP32
MS-AMP O2 | FP8 | FP8 | FP16 | N/A | FP8 | FP8+FP16
MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16

### Internal structure:

With how `fp8` optimization works with MS-AMP, we get the best bang for our buck if we combine both MS-AMP and transformers engine. As a result when preparing the model and optimizer we run through the same fix that exists for TPU optimizers so that we can replace the new `te.Linear` layers with the equivalent `ms-amp` ones that increase throughput without decreasing performance.

## Basic Usage Example(s):

A user can either do:

```python
accelerator = Accelerator(mixed_precision="fp8")
```
Or use the `FP8RecipeKwargs`:

```python
# To disable MS-AMP if available in the environment
kwarg_handlers = [FP8RecipeKwargs(enable_ms_amp=False)]

# To change the optimization level
kwarg_handlers = [FP8RecipeKwargs(optimization_level="O1")]

accelerator = Accelerator(
    mixed_precision="fp8",
    kwargs_handlers=kwarg_handlers,
)
```

### Benchmarks

When running on `bloomz-560m` I saw the following speedups on the first 100 batches:

Batch size: 8
Max seq length: 256
GPU used: single 4090

| Model Configuration   | Time per Batch | Peak Memory   |
|-----------------------|----------------|---------------|
| FP16                  | 0.183s          | 20.62 GB      |
| BF16                  | 0.139s          | 15.41 GB      |
| Raw TransformersEngine| 0.129s          | 12.02 GB      |
| TE + MS-AMP           | 0.108s          | 10.66 GB      |

I also verified the loss curves for the experiment were identical


casper-hansen:
> - 03: Specifically for DeepSpeed, implements capabilities so weights and master weights of models
are stored in FP8. If fp8 is selected and deepspeed is enabled, will be used by default.
(Not available currently).

Can we not use DeepSpeed with O3 FP8? Storing gradients in FP8 is already a great step, but storing weights in FP8 would be an even better step, especially when we also enable zero3 to split the model.

> `bloomz-560m`

Small model which already sees improvements is really nice! Do you by chance have the time+compute to do a Llama/Mistral 7B run to check the difference there?

src/accelerate/utils/dataclasses.py
```diff
@@ -184,21 +184,59 @@ class FP8RecipeKwargs(KwargsHandler):
...
+    optimization_level: str = "O2"
 
     def __post_init__(self):
         self.fp8_format = self.fp8_format.upper()
         if self.fp8_format not in ["E4M3", "HYBRID"]:
             raise ValueError("`fp8_format` must be 'E4M3' or 'HYBRID'.")
         if self.amax_compute_algo not in ["max", "most_recent"]:
             raise ValueError("`amax_compute_algo` must be 'max' or 'most_recent'")
+        if self.enable_ms_amp and not is_msamp_available():
+            self.enable_ms_amp = False
```
pacman100, line 237: raise warning mentioning that `msamp` is not available but the flag is True

src/accelerate/accelerator.py
```diff
@@ -1286,13 +1292,16 @@ def prepare(self, *args, device_placement=None):
             result = tuple(
                 self._prepare_one(obj, first_pass=True, device_placement=d) for obj, d in zip(args, device_placement)
             )
+            if self.mixed_precision == "fp8" and self.fp8_recipe_handler.enable_ms_amp:
+                # MS-AMP needs both model and optimizer
+                args = self._prepare_ms_amp(*result)
```
pacman100, line 1297:
```suggestion
                result = self._prepare_ms_amp(*result)
```

pacman100: given that the second pass takes the result

muellerzr:
@casper-hansen as mentioned in the PR, we cannot until they update the deepspeed version they require

pacman100 (commented):
Hello Zach, thank you for working on adding MS AMP FP8 support 🔥🚀✨! The experiments you performed with bloom-560M are already showing nice improvements. Looking forward to experiments on larger scales.

Left a couple suggestions/comment.

pacman100:
Also, what happens if we run this with latest DeepSpeed?

SunMarc (commented):
Thanks for adding ms_amp support ! This looks very good overall. I left a couple of comments. One thing that can be added is some docs about mixed-precision training in the How-to Guide and Concept guide. Correct me if i'm wrong but I wasn't able to find much doc in it. This can be done in a follow-up PR. LMK what you think. The table you linked in this PR is a very good summary.

Issue #2226, opened 2023-12-07
BFloat16 is not supported on MPS


jxysoft:
### System Info

```Shell
- `Accelerate` version: 0.25.0
- Platform: macOS-14.1.2-arm64-arm-64bit
- Python version: 3.9.16
- Numpy version: 1.23.0
- PyTorch version (GPU?): 2.1.1 (False)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 8.00 GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: fp16
        - use_cpu: True
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: False
        - main_training_function: main
        - downcast_bf16: False
        - tpu_use_cluster: False
        - tpu_use_sudo: False
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. my real case
- code = 

from transformers import AutoModelForCausalLM, AutoTokenizer
import torch
model_name = "Qwen/Qwen-1_8B-Chat"
tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
model = AutoModelForCausalLM.from_pretrained(model_name, device_map="auto", trust_remote_code=True, 
        offload_folder='offload', max_memory={"mps": "2GB"}, torch_dtype=torch.float16).eval()
response, history = model.chat(tokenizer, "hello", history=None)
print(response)

- error = 

File [~/miniconda3/envs/pytorch/lib/python3.9/site-packages/accelerate/utils/offload.py:169](https://file+.vscode-resource.vscode-cdn.net/Users/jxy/Projects/person/ai/pytorch/models/qwen/~/miniconda3/envs/pytorch/lib/python3.9/site-packages/accelerate/utils/offload.py:169), in OffloadedWeightsLoader.__getitem__(self, key)
[...]

PR #2227, opened 2023-12-07
fix BFloat16 is not supported on MPS (#2226)

jxysoft:
Fixes BFloat16 is not supported on MPS # (2226)

### DIFF
--- a/src/accelerate/utils/offload.py
+++ b/src/accelerate/utils/offload.py
@@ -169,13 +169,13 @@ class OffloadedWeightsLoader(Mapping):
             try:
                 with safe_open(weight_info["safetensors_file"], framework="pt", device=device) as f:
                     tensor = f.get_tensor(weight_info.get("weight_name", key))
-            except TypeError:  
+            except TypeError:
                 with safe_open(weight_info["safetensors_file"], framework="pt", device="cpu") as f:
                     tensor = f.get_tensor(weight_info.get("weight_name", key))
-                    
+
             if "dtype" in weight_info:
                 tensor = tensor.to(getattr(torch, weight_info["dtype"]))
-            
+
             if tensor.device != torch.device(device):
                 tensor = tensor.to(device)
             return tensor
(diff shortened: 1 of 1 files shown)


pacman100 (approved):
Thank you @jxysoft for the fix, makes sense. cc @SunMarc for a second look

SunMarc (approved):
LGTM ! Thanks for fixing this @jxysoft. Can you do `make style ` to fix the CI ?

jxysoft:
> LGTM ! Thanks for fixing this @jxysoft. Can you do `make style ` to fix the CI ?

ok, I will fix it today

BenjaminBossan:
Nothing much to add from my side. But how about adding some comment to indicate why these steps are necessary. Just from reading the code, it's not obvious that this is a workaround for bf16 on MPS.

src/accelerate/utils/offload.py
```diff
@@ -165,13 +165,21 @@ def __getitem__(self, key: str):
         weight_info = self.index[key]
         if weight_info.get("safetensors_file") is not None:
             device = "cpu" if self.device is None else self.device
-            with safe_open(weight_info["safetensors_file"], framework="pt", device=device) as f:
-                tensor = f.get_tensor(weight_info.get("weight_name", key))
+            tensor = None
+            try:
+                with safe_open(weight_info["safetensors_file"], framework="pt", device=device) as f:
+                    tensor = f.get_tensor(weight_info.get("weight_name", key))
```
muellerzr, line 171: @jxysoft question: does the error happen during `safe_open` or during `f.get_tensor`. 

Aka, will we face an error if we do:
```python
handle = safe_open(weight_info["safetensors_file"], framework="pt", device=device)
```

If so, there's another area where a case like this is needed, and it may be better to refactor out `safe_open` as a `partial`

jxysoft: the error happened during f.get_tensor

muellerzr (approved):

jxysoft:
who can help me to merge the pr?

Issue #2129, opened 2023-11-07
accelerator.disk_offload(): NotImplementedError: Cannot copy out of meta tensor; no data!


blap:
### System Info

```Shell
Hi,

I just tried to disk_offload a model and I got this error:

Like: disk_offload(model=model, offload_dir=save_dir)

File "/home/blap/AutoGPTQ_teste/update_models.py", line 29, in <module>
    disk_offload(model=model, offload_dir=save_dir)
  File "/home/blap/AutoGPTQ-env/lib/python3.10/site-packages/accelerate/big_modeling.py", line 280, in disk_offload
    offload_state_dict(offload_dir, model.state_dict())
  File "/home/blap/AutoGPTQ-env/lib/python3.10/site-packages/accelerate/utils/offload.py", line 99, in offload_state_dict
    index = offload_weight(parameter, name, save_dir, index=index)
  File "/home/blap/AutoGPTQ-env/lib/python3.10/site-packages/accelerate/utils/offload.py", line 33, in offload_weight
    array = weight.cpu().numpy()
NotImplementedError: Cannot copy out of meta tensor; no data!

Maybe the docs lack something or I don't know...


- `Accelerate` version: 0.25.0.dev0
- Platform: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.37
- Python version: 3.10.13
- Numpy version: 1.26.1
- PyTorch version (GPU?): 2.2.0.dev20231106+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 5.71 GB
- GPU type: NVIDIA GeForce MX330
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: fp16
        - use_cpu: True
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - ipex_config: {'ipex': True}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
        - dynamo_config: {'dynamo_backend': 'INDUCTOR', 'dynamo_mode': 'reduce-overhead', 'dynamo_use_dynamic': True, 'dynamo_use_fullgraph': False}
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

[...]

PR #2231, opened 2023-12-07
[Docs] Add doc for cpu/disk offload

SunMarc:
# What does this PR do ? 
As per title.
Solves #2129

### DIFF
--- a/docs/source/concept_guides/big_model_inference.md
+++ b/docs/source/concept_guides/big_model_inference.md
@@ -303,7 +303,7 @@ If you want to offload your model on CPU, you can use [`cpu_offload`]. As a resu
 cpu_offload(model, execution_device)
 ```
 
-You can also use [`cpu_offload_with_hook`]. This function will offloads a model on the CPU and puts it back to an execution device when executed. The difference with [`cpu_offload`] is that the model stays on the execution device after the forward and is only offloaded again when the `offload` method of the returned `hook` is called. Futhermore, ['cpu_offload_with_hook'] is more performant but less memory saving. It is useful for pipelines running a model in a loop:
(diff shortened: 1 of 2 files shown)


docs/source/concept_guides/big_model_inference.md
```diff
@@ -295,15 +295,15 @@ device_map = {"block1": 0, "block2.linear1": 1, "block2.linear2": 1}
...
 
 ```python
 cpu_offload(model, execution_device)
 ```
 
-You can also use `cpu_offload_with_hook`. This function will offloads a model on the CPU and puts it back to an execution device when executed. The difference with [`cpu_offload`] is that the model stays on the execution device after the forward and is only offloaded again when the `offload` method of the returned `hook` is called. Futhermore, ['cpu_offload_with_hook'] is more performant but less memory saving. It is useful for pipelines running a model in a loop:
+You can also use [`cpu_offload_with_hook`]. This function will offloads a model on the CPU and puts it back to an execution device when executed. The difference with [`cpu_offload`] is that the model stays on the execution device after the forward and is only offloaded again when the `offload` method of the returned `hook` is called. Futhermore, ['cpu_offload_with_hook'] is more performant but less memory saving. It is useful for pipelines running a model in a loop:
```
muellerzr, line 306:
```suggestion
You can also use [`cpu_offload_with_hook`]. This function will offloads a model on the CPU and puts it back to an execution device when executed. The difference with [`cpu_offload`] is that the model stays on the execution device after the forward and is only offloaded again when the `offload` method of the returned `hook` is called. Furthermore, [`cpu_offload_with_hook`] is more performant but less memory saving. It is useful for pipelines running a model in a loop:
```

muellerzr (approved):
LG2M :)

PR #2232, opened 2023-12-08
Integrate MS-AMP Support for FP8 as a seperate backend

muellerzr:
# Integrate MS-AMP to the `Accelerator` (round 2)

## What does this add?

This PR introduces an additional backend for FP8 support through [MS-AMP](https://github.com/Azure/MS-AMP/tree/main) which has shown to decrease memory when using FP8 precision while maintaining accuracy

## Who is it for?

Individuals training with FP8 (H100/4090's, etc)

## Issues linked to

https://github.com/Azure/MS-AMP/issues/128

### User-facing:

Two new arguments were added to the `FP8RecipeKwargs`:
* `backend` (`str`): Whether a user should use MS-AMP or TE (transformerengine). Uses `MS-AMP` by default.
* `optimization_level` (`str`), should be one of `"O1"` or `"O2"`. `"O3"` is for DeepSpeed and we need to wait for them to update to v0.9.3 of deepspeed to match what Accelerate supports

General guideline to optimization levels:
  * O1: Weight gradients and `all_reduce` communications are done in fp8, reducing GPU
      memory usage and communication bandwidth
  * O2: First-order optimizer states are in 8-bit, and second order states are in FP16.
      Only available when using Adam or AdamW. This maintains accuracy and can potentially save the highest
      memory.
  * 03: Specifically for DeepSpeed, implements capabilities so weights and master weights of models
      are stored in FP8. If `fp8` is selected and deepspeed is enabled, will be used by default.
      (Not available currently).
      
As a result, `"O2"` is the default. Here is an overview of each optimization level and what it does, taken from their [docs](https://azure.github.io/MS-AMP/docs/user-tutorial/optimization-level):

Optimization Level | Computation(GEMM) | Comm | Weight | Master Weight | Weight Gradient | Optimizer States
-- | -- | -- | -- | -- | -- | --
FP16 AMP | FP16 | FP32 | FP32 | N/A | FP32 | FP32+FP32
Nvidia TE | FP8 | FP32 | FP32 | N/A | FP32 | FP32+FP32
MS-AMP O1 | FP8 | FP8 | FP16 | N/A | FP8 | FP32+FP32
MS-AMP O2 | FP8 | FP8 | FP16 | N/A | FP8 | FP8+FP16
MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16

## Basic Usage Example(s):

A user can either do:

```python
accelerator = Accelerator(mixed_precision="fp8")
```
Or use the `FP8RecipeKwargs`:

```python
# To use TransformerEngine instead
kwarg_handlers = [FP8RecipeKwargs(backend="TE")]

# To change the optimization level
kwarg_handlers = [FP8RecipeKwargs(optimization_level="O1")]

accelerator = Accelerator(
    mixed_precision="fp8",
    kwargs_handlers=kwarg_handlers,
)
```

### Benchmarks

When running on bloomz-560m I saw a memory decrease of ~1/3. 

More experiments need to be conducted on the behavior between TE x MS-AMP wrt performance. For instance, when running my sample script I use here for speed ([here](https://github.com/muellerzr/h100-performance-tests/blob/main/bloomz-560m/accelerate_script.py)) running on the first 100 batches, I saw a stark contrast in the ending training loss between TE and MS-AMP:

BF16 (baseline): 2.4867
TE: *11.3125*
MS-AMP: **2.89**

I also found overall there wasn't much of a time save with MS-AMP, as it actually added time instead (BF16 was ~0.139s/batch, while MS-AMP was 0.169s/batch). I want to run some more tests to verify but these were some local results. 

This performance difference isn't much in the case of BF16 vs MS-AMP, but it is *starkly contrast* when compared to TE. More work is needed to investigate why, so as a result I've opted to make MS-AMP just an entirely separate backend to use, rather than combine the two. 

### What went wrong in the last PR

While the training speed results looked *very good*, I quickly noticed issues with the losses that didn't make sense. Models just simply weren't training or converging, which was surprising. For now taking this more staged approach to the integration while we discover behaviors (both with TE and MS-AMP) through longer training runs.

### DIFF
--- a/docs/source/concept_guides/low_precision_training.md
+++ b/docs/source/concept_guides/low_precision_training.md
@@ -15,7 +15,7 @@ rendered properly in your Markdown viewer.
 
 # Low Precision Training Methods
 
-As new hardware has been released, new training paradigms to better utilize them ahve come out as a result. Currently this is in the form of training
+As new hardware has been released, new training paradigms to better utilize them have come out as a result. Currently this is in the form of training
 in 8-bit precision using packages such as [TranformersEngine](https://github.com/NVIDIA/TransformerEngine) (TE) or [MS-AMP](https://github.com/Azure/MS-AMP/tree/main).
 
 For an introduction to the topics discussed today, it's recommended to review the low precision usage documentation available [here](../usage_guides/low_precision_training.md) as this documentation will reference it regularly. 
@@ -43,7 +43,7 @@ Specifically, 🤗 Accelerate will find and replace the following layers with th
 
 As a result we wind up with a model that has most of its layers in BF16, while some layers are in FP8 reducing some of the memory. 
 
-Anecdotally, we have noticied that profmance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
+Anecdotally, we have noticed that performance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
 in the model are made up of those two layers to replace. As a result only larger models have been seen to show performance improvements when the number of parameters is around and upwards of a few billion. 
 
 There are many different arguments that can be passed to the `TransformerEngine` that customize how FP8 calculations are performed and what they do. A full list of the arguments is available below:
@@ -57,7 +57,7 @@ There are many different arguments that can be passed to the `TransformerEngine`
 
 Each of these can be customized as part of the [`utils.FP8RecipeKwargs`] and tweaked to help optimize performance of your models.
 
-If we notice in the chart mentioned earlier, TE simply casts the computation layers into FP8, while everything else is in FP32. As a result this winds up utilizing the most memory but does so with the benefit of guarenting the least amount of loss in end accuracy during training. 
+If we notice in the chart mentioned earlier, TE simply casts the computation layers into FP8, while everything else is in FP32. As a result this winds up utilizing the most memory but does so with the benefit of guaranteeing the least amount of loss in end accuracy during training. 
 
 ## `MS-AMP`
 
--- a/docs/source/concept_guides/low_precision_training.md
+++ b/docs/source/concept_guides/low_precision_training.md
@@ -1,0 +1,74 @@
+<!--Copyright 2023 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Low Precision Training Methods
+
+As new hardware has been released, new training paradigms to better utilize them ahve come out as a result. Currently this is in the form of training
+in 8-bit precision using packages such as [TranformersEngine](https://github.com/NVIDIA/TransformerEngine) (TE) or [MS-AMP](https://github.com/Azure/MS-AMP/tree/main).
+
+For an introduction to the topics discussed today, it's recommended to review the low precision usage documentation available [here](../usage_guides/low_precision_training.md) as this documentation will reference it regularly. 
+
+## A Quick Chart
+
+Below is a quick chart from the MS-AMP documentation showing the different bit-precisions for each solution during training:
+
+Optimization Level | Computation(GEMM) | Comm | Weight | Master Weight | Weight Gradient | Optimizer States
+-- | -- | -- | -- | -- | -- | --
+FP16 AMP | FP16 | FP32 | FP32 | N/A | FP32 | FP32+FP32
+Nvidia TE | FP8 | FP32 | FP32 | N/A | FP32 | FP32+FP32
+MS-AMP O1 | FP8 | FP8 | FP16 | N/A | FP8 | FP32+FP32
+MS-AMP O2 | FP8 | FP8 | FP16 | N/A | FP8 | FP8+FP16
+MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16
+
+## `TransformersEngine`
+
+`TranformersEngine` is the first solution to trying to train in 8-bit floating point. It operates by changing layers inside the model to ones that utilize 8-bit accelerated ones that should not degrade the final accuracy of the model upon convergence. 
+
+Specifically, 🤗 Accelerate will find and replace the following layers with their own versions:
+
+* `nn.LayerNorm` for `te.LayerNorm`
+* `nn.Linear` for `te.Linear`
+
+As a result we wind up with a model that has most of its layers in BF16, while some layers are in FP8 reducing some of the memory. 
+
+Anecdotally, we have noticied that profmance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
+in the model are made up of those two layers to replace. As a result only larger models have been seen to show performance improvements when the number of parameters is around and upwards of a few billion. 
+
+There are many different arguments that can be passed to the `TransformerEngine` that customize how FP8 calculations are performed and what they do. A full list of the arguments is available below:
+
+* `margin`: The margin to use for the gradient scaling.
+* `interval`: The interval to use for how often the scaling factor is recomputed.
+* `fp8_format``: The format to use for the FP8 recipe. Must be one of `E4M3` or `HYBRID`.
+* `amax_history_len`: The length of the history to use for the scaling factor computation
+* `amax_compute_algo`: The algorithm to use for the scaling factor computation. Must be one of `max` or `most_recent`.
+* `override_linear_precision`: Whether or not to execute `fprop`, `dgrad`, and `wgrad` GEMMS in higher precision.
+
+Each of these can be customized as part of the [`utils.FP8RecipeKwargs`] and tweaked to help optimize performance of your models.
+
+If we notice in the chart mentioned earlier, TE simply casts the computation layers into FP8, while everything else is in FP32. As a result this winds up utilizing the most memory but does so with the benefit of guarenting the least amount of loss in end accuracy during training. 
+
+## `MS-AMP`
+
+MS-AMP takes a different approach to `TransformersEngine` by providing three different optimization levels to convert more operations in FP8 or FP16.
+
+* The base optimization level (`O1`), will also pass communications of the weights (such as in DDP) in FP8, stores the weights of the model in FP16, and leaves the optimizer states in FP32. The main benefit of this optimization level is we can reduce the communication bandwidth by essentially half and more GPU memory is saved due to 1/2 of everything being cast in FP8, and the weights are cast to FP16. Notably both the optimizer states remain in FP32.
+
+* The second optimization level (`O2`) improves upon this by also reducing the precision of the optimizer states. One is in FP8 while the other is in FP16. Generally it's been shown that this will only provide a net-gain of no degredated end accuracy, increased training speed, and reduced memory as now every state is either in FP16 or FP8. 
+
+* Finally, MS-AMP has a third optimization level (`O3`) which helps during DDP scenarios such as DeepSpeed. The weights of the model in memory is fully cast to FP8 and the master weights are now stored in FP16. This fully reduces memory by the highest factor as now not only is almost everything in FP8, only two states are left in FP16. Currently only DeepSpeed versions up through 0.9.2 are supported, so this capability is not included in the 🤗 Accelerate integration
+
+## Combining the two
+
+More experiments need to be performed but it's been noted that combining both MS-AMP and TransformersEngine can lead to the highest throughput by relying on NVIDIA's optimized FP8 operators and utilizing how MS-AMP reduces the memory overhead.
--- a/docs/source/concept_guides/low_precision_training.md
+++ b/docs/source/concept_guides/low_precision_training.md
@@ -15,10 +15,10 @@ rendered properly in your Markdown viewer.
 
 # Low Precision Training Methods
 
-As new hardware has been released, new training paradigms to better utilize them have come out as a result. Currently this is in the form of training
+The release of new kinds of hardware led to the emergence of new training paradigms that better utilize them. Currently, this is in the form of training
 in 8-bit precision using packages such as [TranformersEngine](https://github.com/NVIDIA/TransformerEngine) (TE) or [MS-AMP](https://github.com/Azure/MS-AMP/tree/main).
 
-For an introduction to the topics discussed today, it's recommended to review the low precision usage documentation available [here](../usage_guides/low_precision_training.md) as this documentation will reference it regularly. 
+For an introduction to the topics discussed today, we recommend reviewing the [low-precision usage guide](../usage_guides/low_precision_training.md) as this documentation will reference it regularly. 
 
 ## A Quick Chart
 
@@ -36,7 +36,7 @@ MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16
 
 `TranformersEngine` is the first solution to trying to train in 8-bit floating point. It operates by changing layers inside the model to ones that utilize 8-bit accelerated ones that should not degrade the final accuracy of the model upon convergence. 
 
-Specifically, 🤗 Accelerate will find and replace the following layers with their own versions:
+Specifically, 🤗 Accelerate will find and replace the following layers with `TranformersEngine` versions:
 
 * `nn.LayerNorm` for `te.LayerNorm`
 * `nn.Linear` for `te.Linear`
@@ -44,7 +44,7 @@ Specifically, 🤗 Accelerate will find and replace the following layers with th
 As a result we wind up with a model that has most of its layers in BF16, while some layers are in FP8 reducing some of the memory. 
 
 Anecdotally, we have noticed that performance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
-in the model are made up of those two layers to replace. As a result only larger models have been seen to show performance improvements when the number of parameters is around and upwards of a few billion. 
+in the model are made up of those two layers to replace. As a result, only larger models have shown performance improvements when the number of parameters is around and upwards of a few billion. 
 
 There are many different arguments that can be passed to the `TransformerEngine` that customize how FP8 calculations are performed and what they do. A full list of the arguments is available below:
 
@@ -63,11 +63,11 @@ If we notice in the chart mentioned earlier, TE simply casts the computation lay
 
 MS-AMP takes a different approach to `TransformersEngine` by providing three different optimization levels to convert more operations in FP8 or FP16.
 
-* The base optimization level (`O1`), will also pass communications of the weights (such as in DDP) in FP8, stores the weights of the model in FP16, and leaves the optimizer states in FP32. The main benefit of this optimization level is we can reduce the communication bandwidth by essentially half and more GPU memory is saved due to 1/2 of everything being cast in FP8, and the weights are cast to FP16. Notably both the optimizer states remain in FP32.
+* The base optimization level (`O1`), passes communications of the weights (such as in DDP) in FP8, stores the weights of the model in FP16, and leaves the optimizer states in FP32. The main benefit of this optimization level is that we can reduce the communication bandwidth by essentially half. Additionally, more GPU memory is saved due to 1/2 of everything being cast in FP8, and the weights being cast to FP16. Notably, both the optimizer states remain in FP32.
 
 * The second optimization level (`O2`) improves upon this by also reducing the precision of the optimizer states. One is in FP8 while the other is in FP16. Generally it's been shown that this will only provide a net-gain of no degredated end accuracy, increased training speed, and reduced memory as now every state is either in FP16 or FP8. 
 
-* Finally, MS-AMP has a third optimization level (`O3`) which helps during DDP scenarios such as DeepSpeed. The weights of the model in memory is fully cast to FP8 and the master weights are now stored in FP16. This fully reduces memory by the highest factor as now not only is almost everything in FP8, only two states are left in FP16. Currently only DeepSpeed versions up through 0.9.2 are supported, so this capability is not included in the 🤗 Accelerate integration
+* Finally, MS-AMP has a third optimization level (`O3`) which helps during DDP scenarios such as DeepSpeed. The weights of the model in memory are fully cast to FP8, and the master weights are now stored in FP16. This fully reduces memory by the highest factor as now not only is almost everything in FP8, only two states are left in FP16. Currently, only DeepSpeed versions up through 0.9.2 are supported, so this capability is not included in the 🤗 Accelerate integration
 
 ## Combining the two
 
--- a/docs/source/concept_guides/low_precision_training.md
+++ b/docs/source/concept_guides/low_precision_training.md
@@ -34,7 +34,7 @@ MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16
 
 ## `TransformersEngine`
 
-`TranformersEngine` is the first solution to trying to train in 8-bit floating point. It operates by changing layers inside the model to ones that utilize 8-bit accelerated ones that should not degrade the final accuracy of the model upon convergence. 
+`TranformersEngine` is the first solution to trying to train in 8-bit floating point. It works by using drop-in replacement layers for certain ones in a model that utilize their FP8-engine to reduce the number of bits (such as 32 to 8) without degrading the final accuracy of the model. 
 
 Specifically, 🤗 Accelerate will find and replace the following layers with `TranformersEngine` versions:
 
@@ -46,7 +46,7 @@ As a result we wind up with a model that has most of its layers in BF16, while s
 Anecdotally, we have noticed that performance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
 in the model are made up of those two layers to replace. As a result, only larger models have shown performance improvements when the number of parameters is around and upwards of a few billion. 
 
-There are many different arguments that can be passed to the `TransformerEngine` that customize how FP8 calculations are performed and what they do. A full list of the arguments is available below:
+The `TransformerEngine` can receive many different arguments that customize how it performs FP8 calculations and what they do. A full list of the arguments is available below:
 
 * `margin`: The margin to use for the gradient scaling.
 * `interval`: The interval to use for how often the scaling factor is recomputed.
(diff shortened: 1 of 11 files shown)


muellerzr:
TODO: write some doc guides on MS-AMP

src/accelerate/utils/dataclasses.py
```diff
@@ -173,19 +173,53 @@ class InitProcessGroupKwargs(KwargsHandler):
 class FP8RecipeKwargs(KwargsHandler):
     """
     Use this object in your [`Accelerator`] to customize the initialization of the recipe for FP8 mixed precision
-    training. Please refer to the documentation of this
+    training with `transformers-engine`. Please refer to the documentation of this
```
BenjaminBossan, line 176: This makes it sound that TE is the only option, but MS-AMP is the default.

src/accelerate/utils/dataclasses.py
```diff
@@ -173,19 +173,53 @@ class InitProcessGroupKwargs(KwargsHandler):
...
+                    memory usage and communication bandwidth
+                * O2: First-order optimizer states are in 8-bit, and second order states are in FP16.
+                    Only available when using Adam or AdamW. This maintains accuracy and can potentially save the
+                    highest memory.
+                * 03: Specifically for DeepSpeed, implements capabilities so weights and master weights of models
+                    are stored in FP8. If `fp8` is selected and deepspeed is enabled, will be used by default. (Not
+                    available currently).
     """
 
+    backend: str = "msamp"
```
BenjaminBossan, line 221: You could type limited options as `Literatal["msamp", "te"]` instead of `str`, which would tell users immediately what is possible when they tab-complete. Same for `opt_level`, `fp8_format`, `amax_compute_algo`.

src/accelerate/utils/dataclasses.py
```diff
@@ -194,6 +228,11 @@ class FP8RecipeKwargs(KwargsHandler):
     override_linear_precision: Tuple[bool, bool, bool] = (False, False, False)
 
     def __post_init__(self):
+        self.backend = self.backend.upper()
+        if self.backend not in ["MSAMP", "TE"]:
+            raise ValueError("`backend` must be 'MSAMP' or 'TE' (TransformerEngine).")
+        if self.opt_level not in ["O1", "O2"]:
+            raise NotImplementedError("MS-AMP with Accelerate is only supported for `optimization_level` '01' or '02'")
```
BenjaminBossan, line 235: This error could be confusing when users use TE. Should the check include that the chosen backend is indeed MS-AMP?

src/accelerate/utils/imports.py
```diff
@@ -214,6 +214,18 @@ def is_pandas_available():
     return _is_package_available("pandas")
 
 
+def is_msamp_available():
+    package_exists = importlib.util.find_spec("msamp") is not None
+    if package_exists:
+        try:
+            # MS-AMP has a different metadata name
```
BenjaminBossan, line 221: ugh

src/accelerate/accelerator.py
```diff
@@ -1749,6 +1756,31 @@ def _prepare_ipex(self, *args):
                 result[i] = optimizer
         return tuple(result)
 
+    def _prepare_msamp(self, *args):
+        if not is_msamp_available():
+            raise ImportError(
+                "MS-AMP was not found on your system. Please ensure that MS-AMP is available or choose `'te'` as the backend for FP8 mixed precision training."
```
BenjaminBossan, line 1762: Break the string into 2 lines?

src/accelerate/accelerator.py
```diff
@@ -1749,6 +1756,31 @@ def _prepare_ipex(self, *args):
...
+            )
+        else:
+            import msamp
+
+        model = None
+        optimizer = None
+        result = [obj for obj in args]
+        for obj in result:
```
BenjaminBossan, line 1770: I wonder if there is a flaw in this logic: When `*args` contain more than one module or optimizer, this loop will choose the last one. Thus only the last module and optimizer are initialized by MS-AMP. Then in the loop below, each instance of model and optimizer are replaced by the one prepared model and optimizer.

Also, when there is no module or optimizer, nothing happens.

From checking the [MS-AMP code](https://github.com/Azure/MS-AMP/blob/ca2d8d5dc22070899ba7e9714b8a6eab6cf20064/msamp/__init__.py#L16), it seems that (for the moment), MS-AMP requires exactly one module and 0 or 1 optimizers. I would suggest, instead of silently ignoring when users pass != 1 module or != 0 or 1 optimizers, an error should be raised. Also, when there is no module, `msamp.initialize` can be called with `optimizer=None`, which defaults to `AdamW`.

It is a bit annoying to have to tie the code so closely to MS-AMP implementation details that could change at any moment, but I think it's better than silently ignoring invalid arguments.

(If there is already a previous check on `args` to ensure that these conditions are met, ignore my comment. In that case, it would be nice to add a comment here about what assumptions hold for `args`)

BenjaminBossan (commented):
Thanks for enabling MS-AMP FP8 in accelerate. Overall this looks good, but I had a couple of comments, please have a look.

In addition, I have these more general comments:

1. IIUC, when using FP8 with TE, when it is detected that the device does not support it, there is an automatic fallback to fp16. Is there a similar mechanism for MS-AMP?
2. Implementation-wise, the arguments for FP8 with MS-AMP vs TE are completely disjoint, right? This is a bit unfortunate, as users could e.g. set MS-AMP as backend and change `amax_history_len` and wonder why it has no effect. To be super user-friendly, we would have to add checks and docs that only arguments are changed that are valid for the given backend. A cleaner solution would be to use a completely separate `dataclass` for MS-AMP, although that might clash with the accelerate philosophy of abstracting away such implementation details.
3. I guess no way to run CI tests for this :(

muellerzr:
@BenjaminBossan re;

> A cleaner solution would be to use a completely separate dataclass for MS-AMP, although that might clash with the accelerate philosophy of abstracting away such implementation details.

Eventually we will support a "mixed" backend that combines both, as MS-AMP has support for converting `te.Linear` layers to MS-AMP ones to get the best of both worlds. I'm not enabling this though until tests I run show that combining them does not degrade final accuracy, so for now they are disjoint and part of the same `Plugin`

src/accelerate/utils/imports.py
muellerzr: Yes, very annoying 😢

src/accelerate/utils/dataclasses.py
muellerzr: Done. Will also do a quick runthrough of the dataclasses today to document them all similar to this

src/accelerate/utils/dataclasses.py
muellerzr: Adjusted

muellerzr:
@BenjaminBossan re; 1, we straight up don't allow it. In `AcceleratorState` we have:

```python
            if mixed_precision == "fp8" and not is_fp8_available():
                raise ValueError("Using `fp8` precision requires `transformer_engine` to be installed.")
```

muellerzr:
We should probably refactor this then as part of `is_fp8_available`. Thanks for the smell :)

src/accelerate/accelerator.py
```diff
@@ -1749,6 +1748,40 @@ def _prepare_ipex(self, *args):
...
+            elif isinstance(obj, (torch.optim.Optimizer)):
+                optimizer = obj
+                num_optimizers += 1
+        if optimizer is None and model is None:
+            raise ValueError("You must pass a model and an optimizer to `accelerate.prepare()` when using MS-AMP.")
+        elif num_models > 1 or num_optimizers > 1:
+            raise ValueError(
+                f"You can't use multiple models ({num_models}) or optimizers {num_optimizers} with MS-AMP."
+            )
+        else:
```
BenjaminBossan, line 1776: There is also the case of `num_models==0` and `num_optimizers==1`, which is invalid. I guess MS-AMP will error in that case, but could be worth to check explicitly and give a nice error message to the user (probably they try preparing the components separately if that happens).

src/accelerate/utils/dataclasses.py
```diff
@@ -173,32 +173,84 @@ class InitProcessGroupKwargs(KwargsHandler):
...
     override_linear_precision: Tuple[bool, bool, bool] = (False, False, False)
 
     def __post_init__(self):
-        self.fp8_format = self.fp8_format.upper()
-        if self.fp8_format not in ["E4M3", "HYBRID"]:
-            raise ValueError("`fp8_format` must be 'E4M3' or 'HYBRID'.")
-        if self.amax_compute_algo not in ["max", "most_recent"]:
-            raise ValueError("`amax_compute_algo` must be 'max' or 'most_recent'")
+        self.backend = self.backend.upper()
+        if self.backend not in ["MSAMP", "TE"]:
```
BenjaminBossan, line 240: Not a huge issue, just a tip: You can get the values from the `Literal`, which can help avoid duplication:

```python
from typing import Literal, get_args
FP8_Opt_Level = Literal["O1", "O2"]

...
    opt_level: FP8_Opt_Level = "O2"
...
            if self.opt_level not in get_args(FP8_Opt_Level):
...
```

BenjaminBossan (approved):
LGTM, thanks! Unfortunately, I can't test it :(

src/accelerate/utils/dataclasses.py
muellerzr: There are other cases where this could be good in the library to do, I'll change it here then a follow up to help other areas.

src/accelerate/accelerator.py
muellerzr: Resolved, we now raise the err if either `optimizer` or `model` is `None`, since while msamp can magically create `Adam` we do not want to advertise Accelerate as doing so

SunMarc (approved):
LGTM ! Just a few nits

pacman100 (approved):
Thank you @muellerzr  for working on the MS-AMP FP8 support! ✨ Overall looks good wrt integration and the memory savings of 33% (1/3) for 560M param model. However, It would be great to see more experiments at scale at intermediate model scales of ~10B model sizes as the paper claims:

> Experiment results show that, during the training of GPT-175B model
on H100 GPU platform, our FP8 mixed-precision training framework not only achieved a
remarkable 42% reduction in real memory usage but also ran 64% faster than the widely
adopted BF16 framework (i.e., Megatron-LM), surpassing the speed of Nvidia Transformer Engine by 17%.

Here, it is odd that we see no savings in time.

muellerzr:
@pacman100 I'd expect that likely to be the case, I'll check FLOPS mainly on the scaled up training to see what they can result in :)

muellerzr:
cc @MKhalusova for docs and then we can merge 🤗

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Low Precision Training Methods
+
+As new hardware has been released, new training paradigms to better utilize them have come out as a result. Currently this is in the form of training
```
MKhalusova, line 18:
```suggestion
The release of new kinds of hardware led to the emergence of new training paradigms that better utilize them. Currently, this is in the form of training
```

MKhalusova: Style suggestion, totally up to you.

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+in 8-bit precision using packages such as [TranformersEngine](https://github.com/NVIDIA/TransformerEngine) (TE) or [MS-AMP](https://github.com/Azure/MS-AMP/tree/main).
+
+For an introduction to the topics discussed today, it's recommended to review the low precision usage documentation available [here](../usage_guides/low_precision_training.md) as this documentation will reference it regularly.
```
MKhalusova, line 21:
```suggestion
For an introduction to the topics discussed today, we recommend reviewing the [low-precision usage guide](../usage_guides/low_precision_training.md) as this documentation will reference it regularly. 
```

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+-- | -- | -- | -- | -- | -- | --
+FP16 AMP | FP16 | FP32 | FP32 | N/A | FP32 | FP32+FP32
+Nvidia TE | FP8 | FP32 | FP32 | N/A | FP32 | FP32+FP32
+MS-AMP O1 | FP8 | FP8 | FP16 | N/A | FP8 | FP32+FP32
+MS-AMP O2 | FP8 | FP8 | FP16 | N/A | FP8 | FP8+FP16
+MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16
+
+## `TransformersEngine`
+
+`TranformersEngine` is the first solution to trying to train in 8-bit floating point. It operates by changing layers inside the model to ones that utilize 8-bit accelerated ones that should not degrade the final accuracy of the model upon convergence.
```
MKhalusova, line 37: I find this a little confusing - "changing layers [...] to the **ones** that utilize accelerated **ones** that should [...]". What does the "accelerated ones" refer to?  Consider simplifying this sentence.

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+Nvidia TE | FP8 | FP32 | FP32 | N/A | FP32 | FP32+FP32
+MS-AMP O1 | FP8 | FP8 | FP16 | N/A | FP8 | FP32+FP32
+MS-AMP O2 | FP8 | FP8 | FP16 | N/A | FP8 | FP8+FP16
+MS-AMP O3 | FP8 | FP8 | FP8 | FP16 | FP8 | FP8+FP16
+
+## `TransformersEngine`
+
+`TranformersEngine` is the first solution to trying to train in 8-bit floating point. It operates by changing layers inside the model to ones that utilize 8-bit accelerated ones that should not degrade the final accuracy of the model upon convergence. 
+
+Specifically, 🤗 Accelerate will find and replace the following layers with their own versions:
```
MKhalusova, line 39:
```suggestion
Specifically, 🤗 Accelerate will find and replace the following layers with `TranformersEngine` versions:
```

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+
+* `nn.LayerNorm` for `te.LayerNorm`
+* `nn.Linear` for `te.Linear`
+
+As a result we wind up with a model that has most of its layers in BF16, while some layers are in FP8 reducing some of the memory. 
+
+Anecdotally, we have noticed that performance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
+in the model are made up of those two layers to replace. As a result only larger models have been seen to show performance improvements when the number of parameters is around and upwards of a few billion.
```
MKhalusova, line 47:
```suggestion
in the model are made up of those two layers to replace. As a result, only larger models have shown performance improvements when the number of parameters is around and upwards of a few billion. 
```

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+
+* `nn.LayerNorm` for `te.LayerNorm`
+* `nn.Linear` for `te.Linear`
+
+As a result we wind up with a model that has most of its layers in BF16, while some layers are in FP8 reducing some of the memory. 
+
+Anecdotally, we have noticed that performance gains don't really start showing when using `TransformerEngine` until a large majority of the layers
+in the model are made up of those two layers to replace. As a result only larger models have been seen to show performance improvements when the number of parameters is around and upwards of a few billion. 
+
+There are many different arguments that can be passed to the `TransformerEngine` that customize how FP8 calculations are performed and what they do. A full list of the arguments is available below:
```
MKhalusova, line 49: Suggestion:
The `TransformerEngine` can receive many different arguments that customize how it performs FP8 calculations and what they do.

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+
+* `margin`: The margin to use for the gradient scaling.
+* `interval`: The interval to use for how often the scaling factor is recomputed.
+* `fp8_format``: The format to use for the FP8 recipe. Must be one of `E4M3` or `HYBRID`.
+* `amax_history_len`: The length of the history to use for the scaling factor computation
+* `amax_compute_algo`: The algorithm to use for the scaling factor computation. Must be one of `max` or `most_recent`.
+* `override_linear_precision`: Whether or not to execute `fprop`, `dgrad`, and `wgrad` GEMMS in higher precision.
+
+Each of these can be customized as part of the [`utils.FP8RecipeKwargs`] and tweaked to help optimize performance of your models.
```
MKhalusova, line 58: Suggestion: 
You can customize each of these as part of the [`utils.FP8RecipeKwargs`] to optimize the performance of your models.

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+
+If we notice in the chart mentioned earlier, TE simply casts the computation layers into FP8, while everything else is in FP32. As a result this winds up utilizing the most memory but does so with the benefit of guaranteeing the least amount of loss in end accuracy during training. 
+
+## `MS-AMP`
+
+MS-AMP takes a different approach to `TransformersEngine` by providing three different optimization levels to convert more operations in FP8 or FP16.
+
+* The base optimization level (`O1`), will also pass communications of the weights (such as in DDP) in FP8, stores the weights of the model in FP16, and leaves the optimizer states in FP32. The main benefit of this optimization level is we can reduce the communication bandwidth by essentially half and more GPU memory is saved due to 1/2 of everything being cast in FP8, and the weights are cast to FP16. Notably both the optimizer states remain in FP32.
```
MKhalusova, line 66:
```suggestion
* The base optimization level (`O1`), passes communications of the weights (such as in DDP) in FP8, stores the weights of the model in FP16, and leaves the optimizer states in FP32. The main benefit of this optimization level is that we can reduce the communication bandwidth by essentially half. Additionally, more GPU memory is saved due to 1/2 of everything being cast in FP8, and the weights being cast to FP16. Notably, both the optimizer states remain in FP32.
```

docs/source/concept_guides/low_precision_training.md
```diff
@@ -0,0 +1,74 @@
...
+
+* The second optimization level (`O2`) improves upon this by also reducing the precision of the optimizer states. One is in FP8 while the other is in FP16. Generally it's been shown that this will only provide a net-gain of no degredated end accuracy, increased training speed, and reduced memory as now every state is either in FP16 or FP8. 
+
+* Finally, MS-AMP has a third optimization level (`O3`) which helps during DDP scenarios such as DeepSpeed. The weights of the model in memory is fully cast to FP8 and the master weights are now stored in FP16. This fully reduces memory by the highest factor as now not only is almost everything in FP8, only two states are left in FP16. Currently only DeepSpeed versions up through 0.9.2 are supported, so this capability is not included in the 🤗 Accelerate integration
```
MKhalusova, line 70:
```suggestion
* Finally, MS-AMP has a third optimization level (`O3`) which helps during DDP scenarios such as DeepSpeed. The weights of the model in memory are fully cast to FP8, and the master weights are now stored in FP16. This fully reduces memory by the highest factor as now not only is almost everything in FP8, only two states are left in FP16. Currently, only DeepSpeed versions up through 0.9.2 are supported, so this capability is not included in the 🤗 Accelerate integration
```

docs/source/usage_guides/low_precision_training.md
```diff
@@ -0,0 +1,92 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Low Precision Training Methods
+
+🤗 Accelerate provides integrations to train on lower precision methods using specified supported hardware through the `TransformersEngine` and `MS-AMP` packages. This documentation will help guide you through what hardware is supported, how to configure your [`Accelerator`] to do so, and what you can expect when training.
```
MKhalusova, line 18:
```suggestion
🤗 Accelerate provides integrations to train on lower precision methods using specified supported hardware through the `TransformersEngine` and `MS-AMP` packages. This documentation will help guide you through what hardware is supported, how to configure your [`Accelerator`] to leverage the low precision methods, and what you can expect when training. 
```

docs/source/usage_guides/low_precision_training.md
```diff
@@ -0,0 +1,92 @@
...
+Currently two different backends for FP8 are supported (`TransformersEngine` and `MS-AMP`), each with different capabilities and configurations. 
+
+To use either, the same core API is used. Just pass `mixed_precision="fp8"` to either the [`Accelerator`], during `accelerate config` when prompted about mixed precision, or as part of your `config.yaml` file in the `mixed_precision` key:
+
+```{python}
+from accelerate import Accelerator
+accelerator = Accelerator(mixed_precision="fp8")
+```
+
+By default if `MS-AMP` is available in your environment, 🤗 Accelerate will automatically utilize it as a backend. To specify it yourself however (and customize other parts of the FP8 mixed precision setup), you can utilize the [`utils.FP8RecipeKwargs`] to configure it:
```
MKhalusova, line 42:
```suggestion
By default, if `MS-AMP` is available in your environment, 🤗 Accelerate will automatically utilize it as a backend. To specify it yourself (and customize other parts of the FP8 mixed precision setup), you can utilize the [`utils.FP8RecipeKwargs`]:
```

skyshine102:
> Thank you @muellerzr for working on the MS-AMP FP8 support! ✨ Overall looks good wrt integration and the memory savings of 33% (1/3) for 560M param model. However, It would be great to see more experiments at scale at intermediate model scales of ~10B model sizes as the paper claims:
> 
> > Experiment results show that, during the training of GPT-175B model
> > on H100 GPU platform, our FP8 mixed-precision training framework not only achieved a
> > remarkable 42% reduction in real memory usage but also ran 64% faster than the widely
> > adopted BF16 framework (i.e., Megatron-LM), surpassing the speed of Nvidia Transformer Engine by 17%.
> 
> Here, it is odd that we see no savings in time. Maybe is it the case that with reduced memory, they fit larger batches leading to faster training?

Original paper shows that TE>MS-AMP>BF16 under same batch size setting (paper table 5). It is interesting to see MS-AMP slower than BF16.  
Thank you are all for making this feature integrated to accelerate!

docs/source/usage_guides/low_precision_training.md
```diff
@@ -0,0 +1,92 @@
...
+# Or to specify the backend as `TransformersEngine` even if MS-AMP is installed
+# kwargs = [FP8RecipeKwargs(backend="te")]
+accelerator = Accelerator(mixed_precision="fp8", kwarg_handlers=kwargs)
+```
+
+## Configuring MS-AMP
+
+Of the two, `MS-AMP` is traditionally the easier one to configure as there is only a single argument: the optimization level. 
+
+Currently we support one of two different levels, `"O1"` and `"O2"` (using the letter 'o', not zero).
```
MKhalusova, line 57: "...we support one of two different levels.." - does it mean we support two levels but not at the same time? or does it mean we support one of them, in which case which one?

MKhalusova (approved):
Great work on the docs! I left a few suggestions :)

muellerzr:
@skyshine102:

> Original paper shows that TE>MS-AMP>BF16 under same batch size setting (paper table 5). It is interesting to see MS-AMP slower than BF16.

I have an inkling model size plays a *huge* factor in this. With TE I've pretty much always seen it to be slower unless our model size is > 3B

skyshine102:
> @skyshine102:
> 
> > Original paper shows that TE>MS-AMP>BF16 under same batch size setting (paper table 5). It is interesting to see MS-AMP slower than BF16.
> 
> I have an inkling model size plays a _huge_ factor in this. With TE I've pretty much always seen it to be slower unless our model size is > 3B

I see!

wangpengfei1013:
I have tried this work，It's  amazing. But  some questions:
1、TE and MSAMP didn't have  any difference？
2、BF16 is a little slower than the two，and  comsumed a little more GPU memory comparation to fp8

GPU ：L20
code：
```
compute_dtype = (
        torch.float16
        if training_args.fp16
        else (torch.bfloat16 if training_args.bf16 else torch.float32)
    )

    model = transformers.AutoModelForCausalLM.from_pretrained(
        model_args.model_name_or_path,
        cache_dir=training_args.cache_dir,
        device_map=device_map,
        trust_remote_code=True,
        torch_dtype=compute_dtype,

    )

    #kwargs = FP8RecipeKwargs(backend="te", fp8_format="HYBRID")
    kwargs = FP8RecipeKwargs(backend="msamp", opt_level="O1")

    accelerator = Accelerator(mixed_precision='fp8',  kwargs_handlers=[kwargs])
    #accelerator = Accelerator()
    #device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
    device = accelerator.device
    model.to(device)


    tokenizer = transformers.AutoTokenizer.from_pretrained(
        model_args.model_name_or_path,
        cache_dir=training_args.cache_dir,
        model_max_length=training_args.model_max_length,
        padding_side="right",
        use_fast=False,
        trust_remote_code=True
    )
    tokenizer.pad_token = tokenizer.unk_token
    tokenizer.add_tokens(
        ['<eod>', '<sep>', '<pad>', '<mask>', '<predict>', '<FIM_SUFFIX>', '<FIM_PREFIX>', '<FIM_MIDDLE>',
         '<commit_before>',
         '<commit_msg>', '<commit_after>', '<jupyter_start>', '<jupyter_text>', '<jupyter_code>',
         '<jupyter_output>', '<empty_output>'], special_tokens=True)

    data_module = make_supervised_data_module(tokenizer=tokenizer, data_args=data_args)
    train_dataloader = DataLoader(data_module['train_dataset'], shuffle=True, batch_size=8,)  # 通过这里的dataloader，每个batch的seq_len可能不同
    eval_dataloader = DataLoader(data_module['eval_dataset'], batch_size=8, )
    optimizer = torch.optim.Adam(model.parameters(), lr=5e-5)

    num_epochs = 3
    num_training_steps = num_epochs * len(train_dataloader)  # num of batches * num of epochs
    lr_scheduler=torch.optim.lr_scheduler.StepLR(optimizer, num_training_steps, gamma=0.1, last_epoch=-1, verbose=False)

    model, optimizer, train_dataloader,eval_dataloader,lr_scheduler = accelerator.prepare(model, optimizer, train_dataloader,eval_dataloader,lr_scheduler)

    i=0
        for batch in tqdm(train_dataloader):
            # 要在GPU上训练，需要把数据集都移动到GPU上：
            batch = {k: v.to(device) for k, v in batch.items()}
            loss = model(**batch).loss
            print("step: ",i,"loss: ",loss.item())
            #loss.backward()
            accelerator.backward(loss)
            optimizer.step()
            lr_scheduler.step()
            optimizer.zero_grad()
            if (i+1) % 500==0:
                accelerator.save_state("wpf_test_abcd")
            i=i+1


```

PR #2253, opened 2023-12-14
auto_wrap_policy for PEFT with FSDP

fs4r:
# What does this PR do?

While doing finetuning with a PEFT model and FSDP I noticed excessive memory usage caused by the parameter use_orig_params which leads to additional memory usage for frozen parameters. When trying to disable use_orig_params the following error was thrown by FSDP

> ValueError: Must flatten tensors with uniform `requires_grad` when `use_orig_params=False`

The reason for this is documented in the torch library.

> FSDP has some constraints on freezing parameters (i.e. setting param.requires_grad=False). For use_orig_params=False, each FSDP instance must manage parameters that are all frozen or all non-frozen. For use_orig_params=True, FSDP supports mixing frozen and non-frozen, but we recommend not doing so since then the gradient memory usage will be higher than expected (namely, equivalent to not freezing those parameters). This means that ideally, frozen parameters should be isolated into their own nn.Module s and wrapped separately with FSDP.

I implemented a check in the auto_wrap_policy function of the FullyShardedDataParallelPlugin and wrapped the unfrozen parameters individually if necessary.

## Before submitting
      Pull Request section?
      to it if that's the case.
I wrote an additional test called `test_auto_wrap_policy_peft`which tests if FSDP fails when using PEFT and use_orig_params=False.

## Who can review?

@pacman100
@muellerzr

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -968,7 +968,12 @@ class FullyShardedDataParallelPlugin:
 
     def set_auto_wrap_policy(self, model):
         from peft import PeftModel
-        from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy, transformer_auto_wrap_policy
+        from torch.distributed.fsdp.wrap import (
+            _or_policy,
+            lambda_auto_wrap_policy,
+            size_based_auto_wrap_policy,
+            transformer_auto_wrap_policy,
+        )
 
         default_transformer_cls_names_to_wrap = (
             ",".join(model._no_split_modules) if getattr(model, "_no_split_modules", None) is not None else ""
@@ -996,18 +1001,10 @@ class FullyShardedDataParallelPlugin:
                 # In an FSDP setting PEFT models require individually wrapping unfrozen parameters
 
                 if isinstance(model, PeftModel):
-                    print("PEFT wrapping")
-                    from torch.distributed.fsdp.wrap import (
-                        _or_policy,
-                        lambda_auto_wrap_policy,
-                        transformer_auto_wrap_policy,
-                    )
 
                     def lambda_policy_fn(module):
-                        if (
-                            len(list(module.named_children())) == 0
-                            and getattr(module, "weight", None) is not None
-                            and module.weight.requires_grad
+                        if len(list(module.named_children())) == 0 and any(
+                            p.requires_grad for p in module.parameters()
                         ):
                             return True
                         return False
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -967,6 +967,7 @@ class FullyShardedDataParallelPlugin:
                     return module_class
 
     def set_auto_wrap_policy(self, model):
+        from peft import PeftModel
         from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy, transformer_auto_wrap_policy
 
         default_transformer_cls_names_to_wrap = (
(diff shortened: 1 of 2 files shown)


fs4r:
I also added a test called `test_auto_wrap_policy_peft`in the fsdp tests

src/accelerate/utils/dataclasses.py
```diff
@@ -986,11 +987,36 @@ def set_auto_wrap_policy(self, model):
...
                 )
+
+                # In an FSDP setting PEFT models require individually wrapping unfrozen parameters
+
+                if isinstance(model, PeftModel):
+                    print("PEFT wrapping")
+                    from torch.distributed.fsdp.wrap import (
+                        _or_policy,
+                        lambda_auto_wrap_policy,
+                        transformer_auto_wrap_policy,
```
BenjaminBossan, line 1003: This was already imported at the start of the method, right?

src/accelerate/utils/dataclasses.py
```diff
@@ -986,11 +987,36 @@ def set_auto_wrap_policy(self, model):
...
+                    )
+
+                    def lambda_policy_fn(module):
+                        if (
+                            len(list(module.named_children())) == 0
```
BenjaminBossan, line 1008: I assume that this is called recursively on all sub-modules of the model, and you're checking here that they are "leaf nodes". One thing I wonder: Is it enough to only check for `weight`? What if the learnable parameter has a different name, would this fail? Maybe it's better to check `and any(p.requires_grad for p in module.parameters()`?

src/accelerate/utils/dataclasses.py
```diff
@@ -986,11 +987,36 @@ def set_auto_wrap_policy(self, model):
...
+                auto_wrap_policy = functools.partial(
                     transformer_auto_wrap_policy,
                     # Transformer layer class to wrap
                     transformer_layer_cls=transformer_cls_to_wrap,
                 )
+
+                # In an FSDP setting PEFT models require individually wrapping unfrozen parameters
+
+                if isinstance(model, PeftModel):
+                    print("PEFT wrapping")
```
BenjaminBossan, line 999: Please remove -- or use logging with a more meaningful message.

BenjaminBossan (commented):
Thanks a lot for improving the interop of accelerate with PEFT in an FSDP setting. I have a few small comments, please check.

Regarding the change in general, I have a question. I don't have experience with FSDP, so maybe it's obvious, but what is the _special_ thing about PEFT that requires this extra policy? Is it just that PEFT results in a mixture of params with and without `requires_grad` or is there something else that's special for PEFT?

fs4r:
> Thanks a lot for improving the interop of accelerate with PEFT in an FSDP setting. I have a few small comments, please check.
> 
> Regarding the change in general, I have a question. I don't have experience with FSDP, so maybe it's obvious, but what is the _special_ thing about PEFT that requires this extra policy? Is it just that PEFT results in a mixture of params with and without `requires_grad` or is there something else that's special for PEFT?

Yes the reason is a mixture of params with and without grad. And PEFT does not make sense with use_orig_params=True configuration. That's why this is necessary.

src/accelerate/utils/dataclasses.py
fs4r: Will remove it.

src/accelerate/utils/dataclasses.py
fs4r: Good point. I just ran some tests with different kind of models and your suggestion works flawlessly - i therefore changed it.

fs4r:
@BenjaminBossan Implemented your suggestions.

BenjaminBossan:
> Yes the reason is a mixture of params with and without grad. And PEFT does not make sense with use_orig_params=True configuration. That's why this is necessary.

Thanks for explaining. I wonder if this could be detached from PEFT, WDYT? This would also allow to avoid the `PeftModel` import, which needs to be guarded otherwise, as PEFT is not a strict dependency of accelerate.

fs4r:
> > Yes the reason is a mixture of params with and without grad. And PEFT does not make sense with use_orig_params=True configuration. That's why this is necessary.
> 
> Thanks for explaining. I wonder if this could be detached from PEFT, WDYT? This would also allow to avoid the `PeftModel` import, which needs to be guarded otherwise, as PEFT is not a strict dependency of accelerate.

What we could do is check if there exist parameters with requires_grad=False in the model and if so run the wrapping. I don't know if there exist cases where you do not want the extra wrapping (using use_orig_params=True). @pacman100 what do you think?

pacman100 (commented):
Hello, `use_orig_params=True` is recommended and required if one wants to prepare model and optimizer together in a single `accelerator.prepare` call else without it one needs to prepare/wrap model in FSDp before creating the optimizer. See the PR https://github.com/huggingface/accelerate/pull/2177 for more information.

With respect to auto wrap policy for PEFT when using FSDP, it is already present in PEFT here: https://github.com/huggingface/peft/blob/482a2a6d9aaa01d534b1240e8c1ab6d346eb278f/src/peft/utils/other.py#L354 and used in the example here: https://github.com/huggingface/peft/blob/main/examples/conditional_generation/peft_lora_seq2seq_accelerate_fsdp.py

Can you please provide a minimal reproducer for us to deep dive into the memory increase?

PR #2254, opened 2023-12-14
[`Big-Modeling`] Harmonize device check to handle corner cases

younesbelkada:
# What does this PR do?

This PR addresses the final issue raised by @poedator here: https://github.com/huggingface/transformers/pull/26037#issuecomment-1855421796 

In fact, there is a corner case when users pass `device_map="cuda"` since the device will be initialized without the index argument, leading the check `elif value is not None or torch.device(device) != module._parameters[tensor_name].device:` to pass whereas it should fail. 

I propose a simple approach where we define a new utils function `check_device_same` that effectively checks whether two devices are the same and deals with the corner case if one of the device that we want to check do not have the attribute `index`. 

cc @SunMarc @muellerzr

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -32,6 +32,7 @@ from .constants import SAFE_WEIGHTS_NAME, WEIGHTS_NAME
 from .dataclasses import AutocastKwargs, CustomDtype, DistributedType
 from .imports import is_mps_available, is_npu_available, is_xpu_available
 from .offload import load_offloaded_weight, offload_weight, save_offload_index
+from .other import check_device_same
 from .tqdm import is_tqdm_available, tqdm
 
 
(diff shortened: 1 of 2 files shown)


younesbelkada:
I also confirm 4-bit bnb tests on transformers pass with this patch, which makes this PR totally BC

SunMarc (approved):
LGTM ! Thanks for fixing this

muellerzr (approved):
This is great! Iirc we have a similar check for the gather operation which checks the `PartialState` device. Eventually would be good to think about if we can use that functionality instead but for now this is fine

poedator:
well, I just wanted to propose a 2 line solution instead of 20:
```
    def normalize_device(device):
        return torch.device(f"{device.type}:{device.index if device.index is not None else 0}")
```
Usage: `if normalize_device(first_device) == normalize_device(second_device) ...`

@SunMarc, what do you think?

SunMarc:
Thanks for your suggestion @poedator ! After some discussions, we prefer to keep the longer solution as it is clearer.

PR #2257, opened 2023-12-14
Support `log_images` for aim tracker

Justin900429:
This PR aims to add `log_images` function to aim tracker for better integration between accelerate and aim.

Below is a simple file to check the integration result:

```python
import accelerate
import numpy as np
from PIL import Image


def main():
    accelerator = accelerate.Accelerator(
        project_dir="runs/test_aim_images",
        log_with=["aim"],
    )

    accelerator.init_trackers(project_name="test aim images")
    for cur_iter in range(0, 10):
        image_grid = Image.fromarray(np.random.randint(0, 255, size=(256, 256, 3), dtype=np.uint8))
        for tracker in accelerator.trackers:
            if tracker.name == "aim":
                tracker.log_images(
                    {"test_image": (image_grid, f"iter: {cur_iter}")}, step=cur_iter + 1
                )
    accelerator.end_training()


if __name__ == "__main__":
    main()
```

After running the sample code, users can open aim with `aim up --repo runs/test_aim_images` to check the result.

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -495,9 +495,8 @@ class AimTracker(GeneralTracker):
     def __init__(self, run_name: str, logging_dir: Optional[Union[str, os.PathLike]] = ".", **kwargs):
         self.run_name = run_name
 
-        from aim import Image, Run
+        from aim import Run
 
-        self.aim_image_callable = Image
         self.writer = Run(repo=logging_dir, **kwargs)
         self.writer.name = self.run_name
         logger.debug(f"Initialized Aim project {self.run_name}")
@@ -552,6 +551,7 @@ class AimTracker(GeneralTracker):
                 Additional key word arguments passed along to the `Run.Image` and `Run.track` method specified by the
                 keys `aim_image` and `track`, respectively.
         """
+        from aim import Image
 
         aim_image_kw = {}
         track_kw = {}
@@ -565,7 +565,7 @@ class AimTracker(GeneralTracker):
                 img, caption = value
             else:
                 img, caption = value, ""
-            aim_image = self.aim_image_callable(img, caption=caption, **aim_image_kw)
+            aim_image = Image(img, caption=caption, **aim_image_kw)
             self.writer.track(aim_image, name=key, step=step, **track_kw)
 
     @on_main_process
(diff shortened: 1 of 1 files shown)


src/accelerate/tracking.py
```diff
@@ -536,6 +537,31 @@ def log(self, values: dict, step: Optional[int], **kwargs):
         for key, value in values.items():
             self.writer.track(value, name=key, step=step, **kwargs)
 
+    @on_main_process
+    def log_images(
+        self, values: dict, step: Optional[int] = None, kwargs: Dict[str, dict] = {"aim_image": {}, "track": {}}
```
BenjaminBossan, line 542: Let's not have a mutable default argument, even if right now, nothing incorrect is happening. Instead, let's set it to `None` and in the function body, add those lines:

```python
aim_image_kw = {} if kwargs is None else kwargs["aim_image"]
track_kw = {} if kwargs is None else kwargs["track"]
```

BenjaminBossan (commented):
Thanks for this update. I have no experience with the aim tracker, so I can't comment on that specifically. I just have a small comment regarding the implementation.

src/accelerate/tracking.py
Justin900429: Thanks for the suggestion! I made a slightly change by using `get` to avoid `KeyError`.

src/accelerate/tracking.py
```diff
@@ -495,8 +495,9 @@ class AimTracker(GeneralTracker):
     def __init__(self, run_name: str, logging_dir: Optional[Union[str, os.PathLike]] = ".", **kwargs):
         self.run_name = run_name
 
-        from aim import Run
+        from aim import Image, Run
 
+        self.aim_image_callable = Image
```
muellerzr, line 500: Is there a reason to do `aim_image_callable` and not just do `Image`?

muellerzr (approved):
Thanks! Overall looks great to me, I think considering we've had enough users doing this it may be time to expose auto-handling (or something) as part of the `Accelerator` API as I think this is the 3rd or 4th integration including it. Will look into this at the start of Q1 next year.  

Just left one nit

src/accelerate/tracking.py
Justin900429: I used `aim_image_callable` to remove the need for `import` statement in `log_images`. However, other trackers do `import` also in `log_images`, so I follow the current style and change it also. Thanks for pointing out that!

BenjaminBossan (approved):
Thanks for updating, LGTM.

Issue #2247, opened 2023-12-13
`test_trigger()` fails when running `accelerate test` on TPU v3-8


vanbasten23:
### System Info

```Shell
accelerate version: HEAD
OS: debian
python version: py 3.8
torch: nightly (11/13/2023)
accelerate config:

compute_environment: LOCAL_MACHINE
distributed_type: TPU
downcast_bf16: 'no'
machine_rank: 0
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 4
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

In short, when I run `accelerate test` on my TPU v3-8, it fails with error: [full output]( https://gist.github.com/vanbasten23/efda86202a2e8ff6da83bf5ce1c9eac7)
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in <listcomp>
    return [fn(*args) for args in chunk]
  File "/ansible/.venv/lib/python3.8/site-packages/torch_xla/runtime.py", line 87, in wrapper
    return fn(*args, **kwargs)
  File "/ansible/.venv/lib/python3.8/site-packages/torch_xla/_internal/pjrt.py", line 77, in _run_thread_per_device
    replica_results = list(
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
    yield fs.pop().result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 437, in result
    return self.__get_result()
  File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
    raise self._exception
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
[...]

PR #2263, opened 2023-12-15
Fix breakpoint API in test_script.py on TPU.

vanbasten23:
# What does this PR do?

breakpoint API in test_script.py currently fails on TPU. This PR is intended to fix it:
- `xm.all_reduce("sum", cloned_tensor, scale)` doesn't do an in-place update so a return value should be used to get the value after the all_reduce.
- Run https://github.com/huggingface/accelerate/blob/b0528392c83c1d7afb823d87abb94a924f7f69b1/src/accelerate/test_utils/scripts/test_script.py#L605-L606 not just on the main process but on all processes.

Fixes [# (issue)](https://github.com/huggingface/accelerate/issues/2247)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -615,6 +615,10 @@ def reduce(tensor, reduction="mean", scale=1.0):
         if state.distributed_type == DistributedType.NO:
             return cloned_tensor
         if state.distributed_type == DistributedType.TPU:
+            # Some processes may have different HLO graphs than other
+            # processes, for example in the breakpoint API
+            # accelerator.set_trigger(). Use mark_step to make HLOs
+            # the same on all processes.
             xm.mark_step()
             xm.all_reduce(xm.REDUCE_SUM, [cloned_tensor], scale)
         elif state.distributed_type.value in TORCH_DISTRIBUTED_OPERATION_TYPES:
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -615,7 +615,8 @@ def reduce(tensor, reduction="mean", scale=1.0):
         if state.distributed_type == DistributedType.NO:
             return cloned_tensor
         if state.distributed_type == DistributedType.TPU:
-            cloned_tensor = xm.all_reduce("sum", cloned_tensor, scale)
+            xm.mark_step()
+            xm.all_reduce(xm.REDUCE_SUM, [cloned_tensor], scale)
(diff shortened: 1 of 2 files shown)


muellerzr (changes requested):
No, this is fully changing the intention of the `set_trigger` API here. A different solution must be needed as the setting of the breakpoint must be done on its own process. (The tensor itself has been made on every device during Accelerator's `__init__`)

This change needs to happen in `check_trigger` only if a logic needs to be changed for XLA.

vanbasten23:
I see your point in https://discuss.pytorch.org/t/how-to-use-break-in-distributeddataparallel-training/88296/5. Let me see if there's something else we can do.

src/accelerate/utils/operations.py
```diff
@@ -615,7 +615,12 @@ def _reduce_across_processes(tensor, reduction="mean", scale=1.0):
         if state.distributed_type == DistributedType.NO:
             return cloned_tensor
         if state.distributed_type == DistributedType.TPU:
-            xm.all_reduce("sum", cloned_tensor, scale)
+            # Some processes may have different HLO graphs than other
+            # processes, for example in the breakpoint API
+            # accelerator.set_trigger(). Use mark_step to make HLOs
+            # the same on all processes.
+            xm.mark_step()
```
will-cromar, line 622: Good find! From [this comment](https://github.com/huggingface/accelerate/issues/2247#issuecomment-1874583136), the difference in HLO comes from `set_trigger`, right? That's just an assignment:

https://github.com/huggingface/accelerate/blob/403c0714d1dd019a481022afce5df75a9963ecd9/src/accelerate/accelerator.py#L1966

(I actually don't understand why assignment of a constant changes the HLO. Maybe @jackcaog can explain)

In general, I'd rather make sure each operation that we want to run immediately clears the graph by calling `mark_step` at the end. But I don't think there's a clean way to do that in the case of `set_trigger` unless we actually just check the device type there. So calling `xm.mark_step` at the beginning here makes sense.

Can you also call `xm.mark_step` after the `xm.all_reduce` in this function to clear the graph before the next operation? This will make the XLA case immediately run the collective like the `torch.distributed` case below, rather than waiting until the tensor is referenced again.

vanbasten23: Thanks. It's done.

vanbasten23:
hi @muellerzr , could you take another look at the fix?

will-cromar (approved):

muellerzr (approved):
Very nice investigation, LG2M

Issue #2208, opened 2023-12-02
v0.25 breaks my test cases - totally different training losses than v0.24


danielhanchen:
### System Info

```Shell
- `Accelerate` version: 0.24.1
- Platform: Linux-5.15.120+-x86_64-with-glibc2.35
- Python version: 3.10.12
- Numpy version: 1.23.5
- PyTorch version (GPU?): 2.1.0+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 51.00 GB
- GPU type: Tesla T4
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Accelerate v0.24.1: https://colab.research.google.com/drive/1VD4PXDhcx1t5dM2hYewyFK891GNfcELI?usp=sharing

Accelerate v0.25.0: https://colab.research.google.com/drive/1-5MdIOp0cM0scC-CdRZhh8OYhnGHqct4?usp=sharing

### Expected behavior

During testing for https://github.com/unslothai/unsloth, the first 7 steps for training errors on v0.24.1 should be:

|Step|Loss|
|-----|-----|
1 | 1.051200
2 | 1.317400
3 | 1.874600
4 | 1.451000
5 | 1.611200
6 | 1.385900
7 | 1.443000

But then on v0.25:
|Step|Loss|
|-----|-----|
1 | 1.691600
2 | 1.662400
3 | 1.790300
4 | 1.853500
5 | 1.421400
6 | 1.478500
7 | 1.437300

The Colab notebook's seed is 3407. I tried it on Tesla T4, A100 and the different training losses persist.

I can't really theorize exactly what's going on, but I narrowed it down to accelerate's new version.

1. I checked if bitsandbytes, transformers, datasets, peft, trl and even tim dettmer's dataset whether they changed - all no
2. Accelerate got updated 20 hrs ago - and the Colab notebooks show if you install an old version, the errors are again equivalent.

muellerzr:
What if you used <0.24.0? We hit a bug where the data was never being shuffled so you were overfitting on the same samples being shown in the same order each time

danielhanchen:
@muellerzr Let me run it now!

[...]

PR #2264, opened 2023-12-15
Bring old seed technique back

muellerzr:
# What does this PR do?

This PR disables the seedable random sampler by default. While yes it can make user results *fully reproducible*, it has a questionable degradation in quality which outweighs its benefits. It can be enabled explicitly in the `Accelerator`. Needs to be paired with a PR in Transformers to configure it there as well. 

TODO: Fix tests

Fixes # (issue)
Fixes https://github.com/huggingface/accelerate/issues/2208

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -219,7 +219,8 @@ class Accelerator:
             all workers.
         use_seedable_sampler (`bool`, *optional*`):
             Whether or not use a fully seedable random sampler ([`~data_loader.SeedableRandomSampler`]). Comes at a
-            cost of potentially different performances due to different shuffling algorithms.
+            cost of potentially different performances due to different shuffling algorithms, but will ensure the
+            training results are fully reproducible.
         step_scheduler_with_optimizer (`bool`, *optional`, defaults to `True`):
             Set `True` if the learning rate scheduler is stepped at the same time as the optimizer, `False` if only
             done under certain circumstances (at the end of each epoch, for instance).
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -963,8 +963,8 @@ class Accelerator:
 
         Args:
             *models (list of `torch.nn.Module`):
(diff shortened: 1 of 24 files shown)


src/accelerate/accelerator.py
```diff
@@ -217,6 +217,9 @@ class Accelerator:
             If set to `True`, in cases where the total batch size across all processes does not exactly divide the
             dataset, samples at the start of the dataset will be duplicated so the batch can be divided equally among
             all workers.
+        use_seedable_sampler (`bool`, *optional*`):
+            Whether or not use a fully seedable random sampler ([`~data_loader.SeedableRandomSampler`]). Comes at a
```
SunMarc, line 221: Maybe add a quick note about making the user results fully reproducible

src/accelerate/accelerator.py
```diff
@@ -217,6 +217,9 @@ class Accelerator:
             If set to `True`, in cases where the total batch size across all processes does not exactly divide the
             dataset, samples at the start of the dataset will be duplicated so the batch can be divided equally among
             all workers.
+        use_seedable_sampler (`bool`, *optional*`):
```
SunMarc, line 220:
```suggestion
        use_seedable_sampler (`bool`, *optional*, defaults to `False`):
```

SunMarc (approved):
LGTM ! Just a few nits

BenjaminBossan (approved):
Makes sense to make this optional, too bad that this seems to cause issues for some users (but we'll probably never hear from users whose training has improved because of the seedable sampler).

Having a test for this would be nice, but could be added later.

pacman100 (approved):
Thank you @muellerzr for making this configurable!

muellerzr:
Will not merge for now as a few things need to be done (probably fresh in the new year):

1. Run train from scratch examples with known results to verify performance. On my end this is the `imagewoof` and `imagenette` datasets as I know how they behave on untrained models well and they train very quickly
2. Update `Trainer` to have this modification as well in the `TrainingArguments`

PR #2274, opened 2023-12-21
[doc] FSDP improvements

stas00:
fix the FSDP doc:

- use the real config entry names when describing their values
- add missing `fsdp_min_num_params` in the example
- set the `fsdp_forward_prefetch=true` - you of course want it to be enabled for a maximum performance


muellerzr (approved):
LG2M, it would be *really* nice if we can update this as part of the CLI directly and in `docs/source/package_reference/cli` as well since people look at those regularly

stas00:
the current state of `cli.md` is very lacking. it's only useful to know which keys to grep for, IMHO

```
**Fully Sharded Data Parallelism Arguments**:

The following arguments are only useful when `use_fsdp` is passed or Fully Sharded Data Parallelism is configured through `accelerate config`:

* `--fsdp_offload_params` (`str`) -- Decides Whether (true|false) to offload parameters and gradients to CPU.
* `--fsdp_min_num_params` (`int`) -- FSDP's minimum number of parameters for Default Auto Wrapping.
* `--fsdp_sharding_strategy` (`int`) -- FSDP's Sharding Strategy.
* `--fsdp_auto_wrap_policy` (`str`) -- FSDP's auto wrap policy.
* `--fsdp_transformer_layer_cls_to_wrap` (`str`) -- Transformer layer class name (case-sensitive) to wrap, e.g, `BertLayer`, `GPTJBlock`, `T5Block` ...
* `--fsdp_backward_prefetch_policy` (`str`) -- FSDP's backward prefetch policy.
* `--fsdp_state_dict_type` (`str`) -- FSDP's state dict type.
```

but let's do it one step at a time as they aren't dependent on each other wrt to improvements.

stas00:
More things to fix:

> Only use with Static graphs.

There is no reference anywhere in the doc what that means.

stas00:
@pacman100, did you set `fsdp_forward_prefetch: false` in the example because setting it to true wasn't working? it should be true by default since it's faster, but I see it works in pt-nightly and it breaks in pt-2.1.2 with:

```
output = self.model(
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 825, in forward
    args, kwargs = _pre_forward(
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 429, in _pre_forward
    unshard_fn(state, handle)
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 470, in _pre_forward_unshard
    _prefetch_handle(state, handle, _PrefetchMode.FORWARD)
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 1220, in _prefetch_handle
    _unshard(state, handle, state._unshard_stream, state._pre_unshard_stream)
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 336, in _unshard
    ran_pre_unshard = handle.pre_unshard()
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/flat_param.py", line 1194, in pre_unshard
    ret = self._writeback_orig_params()
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/flat_param.py", line 2194, in _writeback_orig_params
    self._writeback_tensor(
  File "/home/stas/anaconda3/envs/py39-pt21/lib/python3.9/site-packages/torch/distributed/fsdp/flat_param.py", line 2262, in _writeback_tensor
    raise RuntimeError(
RuntimeError: Cannot writeback when the gradient shape changes
Expects torch.Size([256]) but got torch.Size([16, 16])
```

pacman100 (approved):
Thank you @stas00 for improving the docs! ✨

pacman100:
> @pacman100, did you set fsdp_forward_prefetch: false in the example because setting it to true wasn't working? it should be true by default since it's faster, but I see it works in pt-nightly and it breaks in pt-2.1.2 with:

I'm trying to be in sync with the defaults of https://pytorch.org/docs/stable/fsdp.html unless the integration requires the non-default values.

Examples of defaults being different:
1. To enable RAM efficient pre-trained model loading similar to deepspeed zero init, `sync_module_states` needed to be `True` instead of the default `False`. So, the default now for `sync_module_states`  is `True` with Accelerate's integration of FSDP.
2. To enable Accelerate's API wherein the model and optimizer are prepared together as opposed to the recommendation of preparing the model before creating the optimizer for FSDP due to the following reason, `use_orig_params` needs to be `True` instead of the default `False`. So, the default now for `use_orig_params`  is `True` with Accelerate's integration of FSDP.
```
The optimizer must be initialized after the module has been wrapped with FSDP since FSDP will shard and transform the module’s parameters in a way that may not preserve the original parameter variables. Thus, the previously initialized optimizer may have stale references to the parameters.
```

stas00:
Thank you for clarifying that you were copying the defaults, Sourab!

I was asking specifically about `fsdp_forward_prefetch` - reading the original docs - it's absolutely unclear to what and why - I'm asking at the torch slack and will follow up once I gain clarity.

Issue #2275, opened 2023-12-21
[feature request] fsdp fwd/bwd prefetch config name consistency 


stas00:
fwd and bwd prefetch are "symmetrical" values, though currently the naming is strangely inconsistent:

```
  fsdp_forward_prefetch:
  fsdp_backward_prefetch_policy: 
```

should it be:
```
fsdp_forward_prefetch
fsdp_backward_prefetch
```
or:
```
fsdp_forward_prefetch_policy
fsdp_backward_prefetch_policy
```

thanks.

pacman100:
Hello, the above PR should address this issue by deprecating `fsdp_backward_prefetch_policy` and having it as `fsdp_backward_prefetch`

stas00:
super, thank you for the quick improvements, Sourab!

Issue #2273, opened 2023-12-21
[feature request] improve FSDP config usability 


stas00:
I have just tried out using FSDP and run into:

```
  fsdp_sharding_strategy: 1
```

how can a user know the meaning of this setting?

And the rest of the options aren't documented either.

After grepping I found

```
src/accelerate/utils/constants.py:FSDP_SHARDING_STRATEGY = ["FULL_SHARD", "SHARD_GRAD_OP", "NO_SHARD", "HYBRID_SHARD", "HYBRID_SHARD_ZERO2"]
```

I'm guessing then that `1` maps to `SHARD_GRAD_OP`? as in index 1 of the above list? I think not, since I tried setting it to `0` and got:

```
ValueError: 0 is not a valid ShardingStrategy
```

Can these strings be used as the values instead of numbers so that the user can see what they are configuring? The rest of the FSDP Accelerate options are implemented this way (meaningful strings), so I'm not sure why `fsdp_sharding_strategy` isn't.

and could these options be added to the docs?

Thank you!

stas00:
ok, so debugging this some more - this is enum so the values are starting from 1 and not 0 and they are:

```
1:ShardingStrategy.FULL_SHARD
2:ShardingStrategy.SHARD_GRAD_OP
3:ShardingStrategy.NO_SHARD
4:ShardingStrategy.HYBRID_SHARD
5:ShardingStrategy._HYBRID_SHARD_ZERO2
```

Allowing a user to say:

```
  fsdp_sharding_strategy: FULL_SHARD
```

would probably be the most consistent and self-documenting approach if I may suggest.

muellerzr:
Agreed with you there @stas00, I was going to look into enabling both options for the user :)

pacman100:
Yes, this is something I wanted to correct but didn't do it for backwards compatibility. Will work on it

pacman100:
Hello, the above PR should resolve this issue by being backwards compatible while making the string values the default now.

stas00:
Thanks a lot, this is much better now, Sourab.

AfrinDange:
Can you also add more details about the several options used in FSDP config? I am a beginner and struggling to find resources to use FSDP for my custom pytorch module. 
Currently, I am referring to [this page](https://huggingface.co/docs/accelerate/usage_guides/fsdp). It focuses on transformer models only.
Please point out any relevant documentation that I might be missing.

stas00:
[...]

PR #2288, opened 2023-12-27
Improve FSDP config usability

pacman100:
# What does this PR do?
1. Fixes https://github.com/huggingface/accelerate/issues/2275 and https://github.com/huggingface/accelerate/issues/2273 while being backwards compatible.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -503,7 +503,7 @@ def launch_command_parser(subparsers=None):
         "--fsdp_backward_prefetch_policy",
         default=None,
         type=str,
-        help="This argument is deprecated. Use `fsdp_backward_prefetch` instead.",
+        help="This argument is deprecated and will be removed in version 0.27.0 of 🤗 Accelerate. Use `fsdp_backward_prefetch` instead.",
     )
     fsdp_args.add_argument(
         "--fsdp_backward_prefetch",
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -482,7 +482,7 @@ def launch_command_parser(subparsers=None):
     )
     fsdp_args.add_argument(
         "--fsdp_sharding_strategy",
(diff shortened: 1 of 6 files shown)


src/accelerate/commands/launch.py
```diff
@@ -503,6 +503,12 @@ def launch_command_parser(subparsers=None):
         "--fsdp_backward_prefetch_policy",
         default=None,
         type=str,
+        help="This argument is deprecated. Use `fsdp_backward_prefetch` instead.",
```
muellerzr, line 506: Let’s add a part that it will be removed in two 0.x.0 versions please. (0.27.0 I think)

src/accelerate/utils/launch.py
```diff
@@ -188,7 +189,14 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
         if args.fsdp_transformer_layer_cls_to_wrap is not None:
             current_env["FSDP_TRANSFORMER_CLS_TO_WRAP"] = str(args.fsdp_transformer_layer_cls_to_wrap)
         if args.fsdp_backward_prefetch_policy is not None:
-            current_env["FSDP_BACKWARD_PREFETCH"] = str(args.fsdp_backward_prefetch_policy)
+            warnings.warn(
+                "`fsdp_backward_prefetch_policy` is deprecated and will be removed in version 0.26.0 of 🤗 Accelerate. Use"
```
muellerzr, line 193: 0.27.0 here as well please :)

muellerzr (approved):
Thanks for the improvement! Overall LG2M just some nits:)

src/accelerate/utils/launch.py
pacman100: Done.

src/accelerate/commands/launch.py
pacman100: Done.

PR #2295, opened 2023-12-30
[deepspeed] fix setting `auto` values for comm buffers

stas00:
Currently, if the model doesn't have `model.config` or `model.config.hidden` and auto values are used for comms buffers Accelerate silently ignores the issue and then subsequently fails with:

```
[:0]:    ) = self.accelerator.prepare(
[:0]:  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/accelerator.py", line 1212, in prepare
[:0]:    result = self._prepare_deepspeed(*args)
[:0]:  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/accelerator.py", line 1560, in _prepare_deepspeed
[:0]:    deepspeed_plugin.deepspeed_config_process(must_match=False, **config_kwargs)
[:0]:  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 754, in deepspeed_config_process
[:0]:    self.deepspeed_config_process(
[:0]:  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 758, in deepspeed_config_process
[:0]:    self.fill_match(prefix + key, mismatches, must_match=must_match, **kwargs)
[:0]:  File "/mnt/nvme0/code/huggingface/accelerate-master/src/accelerate/utils/dataclasses.py", line 731, in fill_match
[:0]:    raise ValueError(
[:0]:ValueError: `zero_optimization.reduce_bucket_size` not found in kwargs. Please specify `zero_optimization.reduce_bucket_size` without `auto` (set to correct value) in the DeepSpeed config file or pass it in kwargs.
```

This PR fixes this issue by issuing an error telling the user what's going on and how to fix it.

What this PR does is restoring the original code as it was in HF Transformers as it can be seen here:
https://github.com/huggingface/transformers/commit/3b7ed25da9ce8291b9dabd6943f5b928352f4a16
I'm not sure why it was "simplified" to remove proper edge case handling. I think it may have fallen between the cracks so to speak as this improvement was done when Accelerate already implemented its own version of the integration so it was never synced with HF Transformers later fixes before it took over HF Transformers' implementation.

Also why did Accelerate drop the use of `HfDeepSpeedConfig` methods? A lot of work went into creating easy to use query methods yet on porting Accelerate dropped that and switched to direct nested dict look ups? Not only this code is much more difficult to read, it's also not fool-proof, for example:

This code assumes the key is present, but what if it's not set? it'll fail to run
https://github.com/huggingface/accelerate/blob/403c0714d1dd019a481022afce5df75a9963ecd9/src/accelerate/accelerator.py#L1422

the original `dsconfig.is_auto()` carefully takes care of avoiding that and does all the work for you, you can see it here: https://github.com/huggingface/transformers/commit/3b7ed25da9ce8291b9dabd6943f5b928352f4a16

so in this PR I hacked a local `is_auto()` but ideally it should switch to an object lookup.

I'd advocate replacing all the manual lookups and code like this:
https://github.com/huggingface/transformers/commit/3b7ed25da9ce8291b9dabd6943f5b928352f4a16
with specially designed `HfDeepSpeedConfig` wrappers to make the code more readable and more fool-proof.

@pacman100

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1516,20 +1516,46 @@ class Accelerator:
                 )
 
         if model is not None:
-            if hasattr(model, "config"):
-                hidden_size = (
-                    max(model.config.hidden_sizes)
-                    if getattr(model.config, "hidden_sizes", None)
-                    else getattr(model.config, "hidden_size", None)
+            ds_config = deepspeed_plugin.deepspeed_config
+            # deal with config keys that use `auto` value and rely on model's hidden_size
+            hidden_size_based_keys = [
+                "zero_optimization.reduce_bucket_size",
+                "zero_optimization.stage3_prefetch_bucket_size",
+                "zero_optimization.stage3_param_persistence_threshold",
+            ]
+
+            def is_auto(ds_config, ds_key_long):
+                nodes = ds_key_long.split(".")
+                val = ds_config.get(nodes[0], {}).get(nodes[1], None)
+                return False if None else val == "auto"
+
+            hidden_size_auto_keys = [x for x in hidden_size_based_keys if is_auto(ds_config, x)]
+            if len(hidden_size_auto_keys) > 0:
+                reasoning = (
+                    "therefore it's not possible to automatically fill out the following `auto` entries "
+                    + f"in the DeepSpeed config file: {hidden_size_auto_keys}. You can fix that by replacing "
+                    + "`auto` values for these keys with an integer value of your choice."
                 )
-                if hidden_size is not None:
-                    config_kwargs.update(
-                        {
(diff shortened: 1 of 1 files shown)


pacman100 (commented):
Hello, thank you for the PR. As per your comments, we have looked into it and addressed them in https://github.com/huggingface/accelerate/pull/2313. Please take a look and let us know.

stas00:
I'm not following you, you copied my PR and added a few things - why not do it in the original PR?

pacman100:
Let's merge this and I can follow up a separate PR with other changes.

pacman100 (approved):
Thank you @stas00 for fixing these issues.

PR #2305, opened 2024-01-04
Fixes bug in swapping weights when replacing with Transformer-Engine layers

sudhakarsingh27:
# What does this PR do?
1. When `linear` layers in the model are replaced with Transformer-Engine's linear layers, the weights of the latter should be replaced with the weights of the former (currently, it's the other way around).
2. Adding TE's `TransformerLayer` to `has_transformer_engine_layers` function to allow the case when larger chunks of modules could be replaced (`TransformerLayer` could replace `LlamaDecoderLayer` in this case)

@muellerzr, could you pls take a look? Thanks!


muellerzr (approved):
Thank you for the fix @sudhakarsingh27 🤗

Issue #2250, opened 2023-12-14
incorrect reference to bnb 4bit param class name (FP4Params vs Params4bit)


poedator:
### System Info
version: Accelerate since April 2023 till now

When debugging 4bit serialization, I noticed that accelerate still uses previous class name for 4bit quantized parameters `FP4Params`. 
Example line: https://github.com/huggingface/accelerate/blob/ad3a5bc92084b9c5c6d5e141eb5ff0029f3fcc78/src/accelerate/utils/modeling.py#L330

In bnb it changed to `Params4bit` in April 2023 in this PR https://github.com/TimDettmers/bitsandbytes/commit/4ea489d3bfc119ab4ceb50f999ce611690dc21e2#diff-4d235c7e595546c6656c229dfa139298ce6602b356c2d0bafcb2352eb2cfae79R136

It appeared in Accelerate in June 2023, in this PR: https://github.com/huggingface/accelerate/pull/1626, after the class was renamed in bnb.

Not sure if this actually causes any errors, but the HF / BNB maintainers may want to look into this. Maybe it is not an issue and only aimed at users with older bnb versions??

### Expected behavior

coordination with bnb classes, incl Params4bit

cc @younesbelkada @SunMarc @Titus-von-Koeller

SunMarc:
thanks for the heads-up ! I will investigate what are the potential issues in accelerate ! In transformers, we use a [patched](https://github.com/huggingface/transformers/blob/bb1d0d0d9e7ca356cf5673031183e955cc160158/src/transformers/integrations/bitsandbytes.py#L24) version of this function. So this should not impact transformers at least

poedator:
> Thanks for the heads-up @poedator ! I will investigate what are the potential issues in accelerate ! In transformers, we use a [patched](https://github.com/huggingface/transformers/blob/bb1d0d0d9e7ca356cf5673031183e955cc160158/src/transformers/integrations/bitsandbytes.py#L24) version of this function. So this should not impact transformers at least

Thank you for the quick feedback. I see that you have a copy of this function in the integrations. Still it it does impact transformers because during dispatch phase it gets called inside accelerate. So it bit me when I tested the 4bit quantization (ref [huggingface/accelerate#2254](https://github.com/huggingface/accelerate/pull/2254). Perhaps the accelerate function copy could be simplified, with most functionality moved to transformers.

SunMarc:
i'm closing this since the PR was merged !

PR #2315, opened 2024-01-08
`Params4bit` added to bnb classes in set_module_tensor_to_device()

poedator:
adding `"Params4bit"` to the list of bnb classes in set_module_tensor_to_device(). 
It looks like this class was originally named "FP4Params", but later renamed to "Params4bit", and it was not reflected in accelerate.  See https://github.com/huggingface/accelerate/issues/2250 for discovery details.

This issue was not affecting anything until recently, but as of today (possibly due to some other changes) this breaks 4-bit bnb serialization tests in transformers.

I left `"FP4Params"` intact in case there are compatibility issues with some older bnb version.

tagging @younesbelkada, @SunMarc @TimDettmers @Titus-von-Koeller

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -328,7 +328,7 @@ def set_module_tensor_to_device(
             param is not None
             and param.device.type != "cuda"
             and torch.device(device).type == "cuda"
-            and param_cls.__name__ in ["Int8Params", "FP4Params"]
+            and param_cls.__name__ in ["Int8Params", "FP4Params", "Params4bit"]
         ):
             device_quantization = device
             device = "cpu"


SunMarc (approved):
LGTM ! Can you list which tests are failing because of this Is it related to the refactor that you are currently doing. If the tests are failing because of that, we will have to enforce a minimum version of accelerate for the 4-bit serialization

SunMarc:
Thanks for listing the tests @poedator. I launched these tests on the main branch of transformers and accelerate and they are not failing. On our daily CI, we did not notice anything also. Can you double check on your side the version of the libraries ?

younesbelkada (approved):
LGTM, I second what @SunMarc said, should be all good IMO !

poedator:
well,  it is good to know that the tests are not failing elsewhere. I will search for the other issues on my side.
Meanwhile this change in this PR still makes sense - hope to see it merged soon.

Issue #2268, opened 2023-12-20
DDP Dataloader shuffle order not synchronize between different ddp workers


Teoge:
### System Info

```Shell
accelerate version: 0.25.0
torch version: 1.12.1
accelerate's configuration:

compute_environment: LOCAL_MACHINE
debug: false
distributed_type: MULTI_GPU
downcast_bf16: 'no'
gpu_ids: all
machine_rank: 0
main_training_function: main
mixed_precision: fp16
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```
from accelerate import Accelerator
from torch.utils.data import DataLoader
import time

accelerator = Accelerator()

dataloader = DataLoader(list(range(24)), shuffle=True, batch_size=1)

dataloader = accelerator.prepare(dataloader)

for batch in dataloader:
    print(batch)
    time.sleep(1)
```

### Expected behavior

I expect the dataset to be distributed to all ddp workers without replacement. But results show that the sampling order of different ddp workers are not the same, leading to repeated sampling, like "9" is sampled three times and "3" is sampled two times.
```
tensor([1], device='cuda:7')
tensor([9], device='cuda:3')
tensor([15], device='cuda:5')
tensor([3], device='cuda:1')
tensor([23], device='cuda:6')
tensor([16], device='cuda:2')
tensor([3], device='cuda:0')
tensor([14], device='cuda:4')
tensor([4], device='cuda:7')
tensor([12], device='cuda:3')
tensor([10], device='cuda:5')
tensor([0], device='cuda:1')
tensor([20], device='cuda:6')
tensor([13], device='cuda:2')
tensor([4], device='cuda:0')
tensor([19], device='cuda:4')
tensor([14], device='cuda:7')
tensor([7], device='cuda:3')
tensor([9], device='cuda:5')
tensor([9], device='cuda:1')
tensor([10], device='cuda:6')
tensor([10], device='cuda:2')
tensor([5], device='cuda:0')
tensor([8], device='cuda:4')
```

[...]

PR #2319, opened 2024-01-09
Bring old seed technique back

muellerzr:
# What does this PR do?

This PR disables the seedable random sampler by default. While yes it can make user results *fully reproducible*, it has a questionable degradation in quality which outweighs its benefits. It can be enabled explicitly in the `Accelerator`. 

Fixes https://github.com/huggingface/accelerate/issues/2268
Fixes https://github.com/huggingface/accelerate/issues/2208

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -221,7 +221,7 @@ class Accelerator:
             Whether or not use a fully seedable random sampler ([`~data_loader.SeedableRandomSampler`]). Ensures
             training results are fully reproducable using a different sampling technique. While seed-to-seed
             results may differ, on average the differences are neglible when using multiple different seeds to 
-            compare. 
+            compare. Should also be ran with [`~utils.set_seed`] for the best results.
         step_scheduler_with_optimizer (`bool`, *optional`, defaults to `True`):
             Set `True` if the learning rate scheduler is stepped at the same time as the optimizer, `False` if only
             done under certain circumstances (at the end of each epoch, for instance).
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -219,9 +219,9 @@ class Accelerator:
             all workers.
         use_seedable_sampler (`bool`, *optional*, defaults to `False`):
             Whether or not use a fully seedable random sampler ([`~data_loader.SeedableRandomSampler`]). Ensures
-            training results are fully reproducable using a different sampling technique. While seed-to-seed
-            results may differ, on average the differences are neglible when using multiple different seeds to 
-            compare. Should also be ran with [`~utils.set_seed`] for the best results.
+            training results are fully reproducable using a different sampling technique. While seed-to-seed results
+            may differ, on average the differences are neglible when using multiple different seeds to compare. Should
+            also be ran with [`~utils.set_seed`] for the best results.
         step_scheduler_with_optimizer (`bool`, *optional`, defaults to `True`):
             Set `True` if the learning rate scheduler is stepped at the same time as the optimizer, `False` if only
             done under certain circumstances (at the end of each epoch, for instance).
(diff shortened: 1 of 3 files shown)


muellerzr:
re-done version of https://github.com/huggingface/accelerate/pull/2264, rebase issues 😓

src/accelerate/accelerator.py
```diff
@@ -217,6 +217,10 @@ class Accelerator:
             If set to `True`, in cases where the total batch size across all processes does not exactly divide the
             dataset, samples at the start of the dataset will be duplicated so the batch can be divided equally among
             all workers.
+        use_seedable_sampler (`bool`, *optional*, defaults to `False`):
+            Whether or not use a fully seedable random sampler ([`~data_loader.SeedableRandomSampler`]). Comes at a
+            cost of potentially different performances due to different shuffling algorithms, but will ensure the
+            training results are fully reproducible.
```
BenjaminBossan, line 223: Should it mention that users should manually `set_seed` if they enable this?

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -360,18 +363,28 @@ def mock_training(length, batch_size, generator):
...
+    train_set, old_model = mock_training(length, batch_size * state.num_processes, generator, use_seedable_sampler)
     assert are_the_same_tensors(old_model.a), "Did not obtain the same model on both processes."
     assert are_the_same_tensors(old_model.b), "Did not obtain the same model on both processes."
 
     accelerator = Accelerator()
-    train_dl = DataLoader(train_set, batch_size=batch_size, shuffle=True, generator=generator)
+    if use_seedable_sampler:
+        # The SeedableRandomSampler is needed during distributed setups
+        # for full reproducability across processes with the `DataLoader`
+        sampler = SeedableRandomSampler(
```
BenjaminBossan, line 380: I don't quite understand, shouldn't the `SeedableRandomSampler` be set automatically when `use_seedable_sampler=True`?

BenjaminBossan (approved):
In principle, this LGTM. I added some comments, but no blockers for merging.

Regarding the original issue(s), do we have confirmation that the usage of `SeedableRandomSampler` causes problems for the _validation/test scores_ when the _seed is fixed correctly_?

Even if not, I can see the argument for going back to the previous default for conserving previous behavior, but arguably it should be recommended to use this new setting then.

src/accelerate/accelerator.py
muellerzr: Yes that would definitely be good. Will do so

vanbasten23:
Hi @muellerzr , we observed that the test `accelerate/test_utils/scripts/test_script.py` started failing on cloud TPU v3 (but not v4) after this commit ([failure](https://gist.github.com/vanbasten23/4a1ddddf615b4e0d78767ae936e6fb25)). Per https://github.com/huggingface/transformers/pull/28421, it sounds to me the plan is to make Trainer take an accelerate configuration ( Need to set `seedable_sampler=True`), right?
cc: @will-cromar @JackCaoG

muellerzr:
Correct. I’ll be doing that this week/next

vanbasten23:
hi @muellerzr , just want to follow up and check if we can make Trainer take an accelerate configuration so to make our test pass on cloud TPU v3.

muellerzr:
@vanbasten23 almost. See the PR for it here: https://github.com/huggingface/transformers/pull/28664

vanbasten23:
Hi @muellerzr , could you check if https://github.com/huggingface/transformers/pull/28664 is ready to merge? We would like to build on top your PR and bring the TPU tests into a successful state. Thanks!

PR #2321, opened 2024-01-09
Add `adapter_only` option to `save_fsdp_model` and `load_fsdp_model` to only save/load PEFT weights

AjayP13:
# What does this PR do?

- See related PR in `transformers` for motivation: https://github.com/huggingface/transformers/pull/28297
- Adds the ability to only save/load PEFT weights for `save_fsdp_model` and `load_fsdp_model`
- The `adapter_only` parameter (default off) controls whether only the PEFT weights will be saved.

## Before submitting
      Pull Request section?
      to it if that's the case.

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -17,7 +17,7 @@ import torch
 
 from ..logging import get_logger
 from .constants import FSDP_MODEL_NAME, FSDP_PYTORCH_VERSION, OPTIMIZER_NAME
-from .imports import is_torch_distributed_available, is_peft_available
+from .imports import is_peft_available, is_torch_distributed_available
 from .other import extract_model_from_parallel
 from .versions import is_torch_version
 
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -35,14 +35,15 @@ logger = get_logger(__name__)
 def _is_peft_model(model):
     if is_peft_available():
         from peft import PeftModel
-    return is_peft_available() and isinstance(model.module, PeftModel)
+    unwrapped_model = getattr(model.module, "_orig_mod", model.module)
+    return is_peft_available() and isinstance(unwrapped_model, PeftModel)
 
 
 def _get_model_state_dict(model, adapter_only=False):
     if adapter_only and _is_peft_model(model):
         from peft import get_peft_model_state_dict
 
-        return get_peft_model_state_dict(model, adapter_name=model.active_adapter, unwrap_compiled=True)
+        return get_peft_model_state_dict(model, adapter_name=model.active_adapter)
     else:
         return model.state_dict()
 
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -18,6 +18,7 @@ import torch
 from ..logging import get_logger
 from .constants import FSDP_MODEL_NAME, FSDP_PYTORCH_VERSION, OPTIMIZER_NAME
 from .imports import is_torch_distributed_available, is_peft_available
+from .other import extract_model_from_parallel
 from .versions import is_torch_version
 
 
@@ -35,8 +36,7 @@ logger = get_logger(__name__)
 def _is_peft_model(model):
     if is_peft_available():
         from peft import PeftModel
-    unwrapped_model = getattr(model.module, "_orig_mod", model.module)
-    return is_peft_available() and isinstance(unwrapped_model, PeftModel)
+    return is_peft_available() and isinstance(extract_model_from_parallel(model), PeftModel)
 
 
 def _get_model_state_dict(model, adapter_only=False):
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -17,7 +17,7 @@ import torch
 
 from ..logging import get_logger
 from .constants import FSDP_MODEL_NAME, FSDP_PYTORCH_VERSION, OPTIMIZER_NAME
-from .imports import is_torch_distributed_available
+from .imports import is_torch_distributed_available, is_peft_available
 from .versions import is_torch_version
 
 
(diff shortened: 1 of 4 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -32,7 +32,32 @@
 logger = get_logger(__name__)
 
 
-def save_fsdp_model(fsdp_plugin, accelerator, model, output_dir, model_index=0):
+def _is_peft_model(model):
+    if is_peft_available():
+        from peft import PeftModel
+    unwrapped_model = getattr(model.module, "_orig_mod", model.module)
```
younesbelkada, line 38: do we have an unwrap model utility method in accelerate? If `model` is not a DDP `model.module` will fail here no?

younesbelkada (commented):
Thanks a lot ! I left one comment - I will let @pacman100 give his final opinion on the PR as he is more familiar with FSDP and accelerate

src/accelerate/utils/fsdp_utils.py
younesbelkada: ok perfect then! maybe worht mentioning in a comment that `_is_peft_model` is only meant to be used for FSDP models, or maybe change the method name to `_is_fsdp_peft_model`

muellerzr: I think `extract_model_from_parallel` in accelerate should cover this case if desired

muellerzr: But worth double checking/writing a test :)

AjayP13: Thanks @muellerzr, I've just updated this with the unwrapping utility and re-ran the test we have on the other branch on the multiple-GPU machine and it works.

pacman100 (commented):
Thank you @AjayP13, the changes in this PR make sense to me. However, they have been only tested with the `FULL_STATE_DICT` state dict type, would this work with `SHARDED_STATE_DICT` too? If not, then this logic should only be limited to the cases wherein `FULL_STATE_DICT` is used.

AjayP13:
> Thank you @AjayP13, the changes in this PR make sense to me. However, they have been only tested with the `FULL_STATE_DICT` state dict type, would this work with `SHARDED_STATE_DICT` too? If not, then this logic should only be limited to the cases wherein `FULL_STATE_DICT` is used.

Thanks for the review @pacman100 , I just tested this, and this works with `SHARDED_STATE_DICT` as well (similarly the loss drops, sharded state dicts per worker get saved to disk, resuming & load_best both work).

pacman100 (approved):
Thank you @AjayP13 for enabling only the storage of adapter weights when using PEFT+FSDP, very useful! ✨

src/accelerate/utils/fsdp_utils.py
```diff
@@ -32,7 +33,31 @@
...
+        return model.state_dict()
+
+
+def _set_model_state_dict(model, state_dict, adapter_only=False):
+    if adapter_only and _is_peft_model(model):
+        from peft import set_peft_model_state_dict
+
+        return set_peft_model_state_dict(model, state_dict, adapter_name=model.active_adapter)
+    else:
+        return model.load_state_dict(state_dict)
```
muellerzr, line 57: I'm not the biggest fan of this pattern we're working with here. 

These should live inside of one function, and currently from what I can tell we have 2 different ways of getting the state_dict inside accelerate now:
1. Here
2. `Accelerator.get_state_dict`

This hints at me that perhaps we need to chunk up `Accelerator`'s `get_state_dict` into something that can be called elsewhere (and stored in `utils.modeling` probably). 

Same with `_set_state_dict` as well. While yes it's just FSDP, then we can write a check for that if necessary. 

Secondly: we generally don't follow the practice of hidden function names in Accelerate, and everything should be made public unless for *extreme* circumstances, and I'm not convinced this is one of those. 

Can we rewrite this a bit to make it more extensible? 

Perhaps just do an `if/else` for calling `set_peft_model_state_dict` or not, and just importing them at the top of the file. This way we can avoid this entirely.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -32,7 +33,31 @@
 logger = get_logger(__name__)
 
 
-def save_fsdp_model(fsdp_plugin, accelerator, model, output_dir, model_index=0):
+def _is_peft_model(model):
+    if is_peft_available():
+        from peft import PeftModel
+    return is_peft_available() and isinstance(extract_model_from_parallel(model), PeftModel)
```
muellerzr, line 39: Let's move this entire function to `utils.modeling` please, as that makes the most logical sense for it

src/accelerate/utils/imports.py
```diff
@@ -175,6 +175,10 @@ def is_datasets_available():
     return _is_package_available("datasets")
 
 
+def is_peft_available():
```
muellerzr, line 178: We need to add this to `utils.__init__.py`

muellerzr (commented):
Thanks! Overall this seems very helpful, but let's work on the design a bit to adhear to more of our practices. Left some comments and suggestions 😄

src/accelerate/utils/imports.py
AjayP13: Done.

src/accelerate/utils/fsdp_utils.py
AjayP13: Done.

src/accelerate/utils/fsdp_utils.py
AjayP13: @muellerzr See the newest refactor, but I believe this is not possible.

`Accelerator.get_state_dict` already has logic for FSDP, but it handles a different use case for getting the FSDP state dict. `Accelerator.get_state_dict` always returns the `FULL_STATE_DICT` from FSDP. Meanwhile, this file, gets the state dict according to how the user configured they want it to be *saved* (could be a `SHARDED_STATE_DICT`). I believe that is why this file originally never called `Accelerator.get_state_dict`.

As for your suggestion of importing `get_peft_model_state_dict` at the top v.s. encapsulated in the hidden function, we can't import at the top because `accelerate` does not have a dependency on `peft` and importing at the top would throw an error if `peft` is not available on a users computer. Also, the hidden function `_get_model_state_dict` is used 3 times in this file. I could get rid of the hidden function and replace with an inline if-statement in those 3 places, but it would introduce 3 places of repeated code that would need to be updated together. Between the repeated code and the need for keeping the `peft` imports local v.s. top-of-file imports, I think it may make sense to keep the two hidden functions here.

AjayP13: @muellerzr Have you had a chance to look this over? Waiting on this PR to get https://github.com/huggingface/transformers/pull/28297 merged.

muellerzr (approved):
Thanks, your comments make sense!

PR #2313, opened 2024-01-08
DeepSpeed refactoring

pacman100:
# What does this PR do?
1. Addresses comments of https://github.com/huggingface/accelerate/pull/2295

PR #2323, opened 2024-01-09
Update test_deepspeed.py

pacman100:
# What does this PR do?
1. Fixing the DS test following up the PR https://github.com/huggingface/accelerate/pull/2313


muellerzr (approved):
Thanks for the quick fix Sourab :)

PR #2325, opened 2024-01-10
Raise error when using batches of different sizes with `dispatch_batches=True`

SunMarc:
# What does this PR do ? 

This PR raise an error when the user tries to use batches of different sizes with `dispatch_batches=True`. This arg is set to `True` when using an `IterableDataset`. 
Solves [26548](https://github.com/huggingface/transformers/issues/26548). See this [comment](https://github.com/huggingface/transformers/issues/26548#issuecomment-1885798533) for more details.

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -608,12 +608,11 @@ class DataLoaderDispatcher(DataLoader, DataLoaderStateMixin):
                         batches.append(next(iterator))
                     try:
                         batch = concatenate(batches, dim=0)
-                    except RuntimeError:
+                    except RuntimeError as e:
                         raise RuntimeError(
                             "You can't use batches of different size with `dispatch_batches=True` or when using an `IterableDataset`."
-                            "As an alternative, you can either: 1) Have batches of same size 2) Set `dispatch_batches=False`. "
-                            "This way, each process will get fetch their own batch. "
-                            "3) Set `split_batches=True`. By doing so, the main process will fetch a full batch and "
+                            "either pass `dispatch_batches=False` and have each process fetch its own batch "
+                            " or pass `split_batches=True`. By doing so, the main process will fetch a full batch and "
                             "slice it into `num_processes` batches for each process."
                         )
                 # In both cases, we need to get the structure of the batch that we will broadcast on other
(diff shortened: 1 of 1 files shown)


src/accelerate/data_loader.py
```diff
@@ -606,7 +606,16 @@ def _fetch_batches(self, iterator):
                     batches = []
                     for _ in range(self.state.num_processes):
                         batches.append(next(iterator))
-                    batch = concatenate(batches, dim=0)
+                    try:
+                        batch = concatenate(batches, dim=0)
+                    except RuntimeError:
+                        raise RuntimeError(
+                            "You can't use batches of different size with `dispatch_batches=True` or when using an `IterableDataset`."
```
muellerzr, line 613: Alternative: we can directly check if a user has an `IterableDataset` and raise the right part of the err

src/accelerate/data_loader.py
```diff
@@ -606,7 +606,16 @@ def _fetch_batches(self, iterator):
...
+                            "As an alternative, you can either: 1) Have batches of same size 2) Set `dispatch_batches=False`. "
+                            "This way, each process will get fetch their own batch. "
+                            "3) Set `split_batches=True`. By doing so, the main process will fetch a full batch and "
+                            "slice it into `num_processes` batches for each process."
```
muellerzr, line 617:
```suggestion
                            "either pass `dispatch_batches=False` and have each process fetch its own batch "
                            " or pass `split_batches=True`. By doing so, the main process will fetch a full batch and "
                            "slice it into `num_processes` batches for each process."
```

I don't think we need 1, since the user has gathered from the error that 1 won't happen if the batch == the batch size.

src/accelerate/data_loader.py
```diff
@@ -606,7 +606,16 @@ def _fetch_batches(self, iterator):
                     batches = []
                     for _ in range(self.state.num_processes):
                         batches.append(next(iterator))
-                    batch = concatenate(batches, dim=0)
+                    try:
+                        batch = concatenate(batches, dim=0)
+                    except RuntimeError:
+                        raise RuntimeError(
```
muellerzr, line 612:
```suggestion
                    except RuntimeError as e:
                        raise RuntimeError(
```

and have `from e` at the end.

muellerzr (approved):
Thanks! Overall this seems sensible. Made some suggestions for clarity in the error message

src/accelerate/data_loader.py
SunMarc: This makes the quality check red. You you can to print the error message somewhere ?

PR #2330, opened 2024-01-12
Avoid duplicating memory for tied weights in `dispatch_model`, and in forward with offloading 

fxmarty:
@Giuseppe5 noticed that `dispatch_model` does not preserve tied weights during its call, although `retie_parameters` as the end so the end model dispatched model preserves tied weights.

This results in possible large memory spike during `dispatch_model` in case many weights are tied. The core issue in `init_hook` and `set_module_tensor_to_device` is:

```python
import torch
import torch.nn as nn
from collections import OrderedDict

model = nn.Sequential(
    OrderedDict(
        [
            ("linear0", nn.Linear(4, 4)),
            ("linear1", nn.Linear(6, 6)),
            ("linear2", nn.Linear(4, 4)),
            ("linear3", nn.Linear(6, 6)),
        ]
    )
)
model.linear2.weight = model.linear0.weight
model.linear2.bias = model.linear0.bias

old_value0 = getattr(model.linear0, "weight")
old_value2 = getattr(model.linear2, "weight")
print("on cpu linear0:", old_value0.data_ptr())
print("on cpu linear2:", old_value2.data_ptr())

model[0]._parameters["weight"] = old_value0.to("cuda")
model[2]._parameters["weight"] = old_value2.to("cuda")

print("on cuda linear0:", model.linear0.weight.data_ptr())
print("on cuda linear2:", model.linear2.weight.data_ptr())

# Printing
# on cpu linear0: 94179276421568
# on cpu linear2: 94179276421568
# on cuda linear0: 140074192732672
# on cuda linear2: 140074192733184
```

This PR proposes to keep track of a dictionary `tied_params_map` that maps the model's data pointers to tied parameters to a dictionary that maps devices to dispatched data.

When a tied weight has already been dispatched, we can reuse the dispatched data for the other tied weights instead of allocating memory (as currently).

@muellerzr @SunMarc What do you think? Do you see a more elegant solution?

@SunMarc suggested
```python
model.linear0.weight.data = model.linear0.weight.to("cuda")
model.linear2.weight.data = model.linear2.weight.to("cuda")
```

but I did not manage to have this work, the `new_value` [here](https://github.com/huggingface/accelerate/blob/fce61a99eca8cef2a65726d7f4feac41889691cf/src/accelerate/utils/modeling.py#L339) always points to a different data pointer, despite the `old_value` being shared, see:

```
device_map OrderedDict([('linear0', 0), ('linear2', 0), ('linear1', 1), ('linear3', 1)])
---- call attach_align_device_hook_on_blocks linear0
old_value dataptr 115696640
new_value dataptr 140711905198080  <----------------- here
new_value dataptr after param_cls 140711905198080
module weight data_ptr after add hook 140711905198080
---- call attach_align_device_hook_on_blocks linear1
old_value dataptr 118073344
new_value dataptr 140711368327168
new_value dataptr after param_cls 140711368327168
module weight data_ptr after add hook 140711368327168
---- call attach_align_device_hook_on_blocks linear2
old_value dataptr 115696640
new_value dataptr 140711905198592  <----------------- here
new_value dataptr after param_cls 140711905198592
module weight data_ptr after add hook 140711905198592
---- call attach_align_device_hook_on_blocks linear3
old_value dataptr 114558400
new_value dataptr 140711368327680
new_value dataptr after param_cls 140711368327680
module weight data_ptr after add hook 140711368327680
```

It's not clear to me why this is not the same pointer, I could not reproduce locally. Maybe some PyTorch issue.

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -22,13 +22,13 @@ from .state import PartialState
 from .utils import (
     PrefixedDataset,
     find_device,
+    find_tied_parameters,
     named_module_tensors,
     send_to_device,
     set_module_tensor_to_device,
-    find_tied_parameters,
 )
-from .utils.other import recursive_getattr
 from .utils.modeling import get_non_persistent_buffers
+from .utils.other import recursive_getattr
 
 
 class ModelHook:
@@ -118,7 +118,9 @@ class SequentialHook(ModelHook):
         return module
 
 
-def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False, init_hook_kwargs: Optional[Dict] = None):
+def add_hook_to_module(
+    module: nn.Module, hook: ModelHook, append: bool = False, init_hook_kwargs: Optional[Dict] = None
+):
     """
     Adds a hook to a given module. This will rewrite the `forward` method of the module to include the hook, to remove
     this behavior and restore the original `forward` method, use `remove_hook_from_module`.
@@ -332,7 +334,12 @@ class AlignDevicesHook(ModelHook):
         return module
 
 
-def add_align_hook_to_module(module: nn.Module, hook: AlignDevicesHook, append: bool = False, tied_params_map: Optional[Dict[int, Dict[torch.device, torch.Tensor]]] = None):
+def add_align_hook_to_module(
+    module: nn.Module,
+    hook: AlignDevicesHook,
+    append: bool = False,
+    tied_params_map: Optional[Dict[int, Dict[torch.device, torch.Tensor]]] = None,
+):
     """
     Adds a AlignDevicesHook hook to a given module, supporting the `tied_params_map` argument. Please refer to the `add_hook_to_module` function documentation as well.
 
@@ -342,9 +349,13 @@ def add_align_hook_to_module(module: nn.Module, hook: AlignDevicesHook, append: 
             to reuse the first available pointer of a shared weight for all others, instead of duplicating memory.
     """
     if not isinstance(hook, AlignDevicesHook):
-        raise ValueError(f"The hook passed to add_align_hook_to_module should be a AlignDevicesHook, found a {hook.__class__.__name__}.")
+        raise ValueError(
+            f"The hook passed to add_align_hook_to_module should be a AlignDevicesHook, found a {hook.__class__.__name__}."
+        )
 
-    return add_hook_to_module(module=module, hook=hook, append=append, init_hook_kwargs={"tied_params_map": tied_params_map})
+    return add_hook_to_module(
+        module=module, hook=hook, append=append, init_hook_kwargs={"tied_params_map": tied_params_map}
+    )
 
 
 def attach_execution_device_hook(
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -25,7 +25,9 @@ from .utils import (
     named_module_tensors,
     send_to_device,
     set_module_tensor_to_device,
+    find_tied_parameters,
 )
+from .utils.other import recursive_getattr
 from .utils.modeling import get_non_persistent_buffers
 
 
@@ -116,7 +118,7 @@ class SequentialHook(ModelHook):
         return module
 
 
-def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False):
+def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False, init_hook_kwargs: Optional[Dict] = None):
     """
     Adds a hook to a given module. This will rewrite the `forward` method of the module to include the hook, to remove
     this behavior and restore the original `forward` method, use `remove_hook_from_module`.
@@ -135,6 +137,8 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
             The hook to attach.
         append (`bool`, *optional*, defaults to `False`):
             Whether the hook should be chained with an existing one (if module already contains a hook) or not.
+        init_hook_kwargs (Optional[Dict], *optional*, defaults to `None`):
+            Optional arguments to pass to the hook initialization.
 
     Returns:
         `torch.nn.Module`: The same module, with the hook attached (the module is modified in place, so the result can
@@ -153,8 +157,17 @@ def add_hook_to_module(module: nn.Module, hook: ModelHook, append: bool = False)
         old_forward = module.forward
         module._old_forward = old_forward
 
-    module = hook.init_hook(module)
+    #if isinstance(module, nn.Linear):
+    #    print("module weight data ptr bef init hook", module.weight.data_ptr())
+
+    if init_hook_kwargs is None:
+        init_hook_kwargs = {}
+
+    module = hook.init_hook(module, **init_hook_kwargs)
     module._hf_hook = hook
+
+    #if isinstance(module, nn.Linear):
+    #    print("module weight data ptr after init hook", module.weight.data_ptr())
 
     def new_forward(module, *args, **kwargs):
         args, kwargs = module._hf_hook.pre_forward(module, *args, **kwargs)
@@ -248,11 +261,12 @@ class AlignDevicesHook(ModelHook):
             f"place_submodules={self.place_submodules}, skip_keys={repr(self.skip_keys)})"
         )
 
-    def init_hook(self, module):
+    def init_hook(self, module, tied_params_map):
         if not self.offload and self.execution_device is not None:
             for name, _ in named_module_tensors(module, recurse=self.place_submodules):
-                set_module_tensor_to_device(module, name, self.execution_device)
+                set_module_tensor_to_device(module, name, self.execution_device, tied_params_map=tied_params_map)
         elif self.offload:
+            # TODO: validate that this tied param fix works correctly with offloading
             self.original_devices = {
                 name: param.device for name, param in named_module_tensors(module, recurse=self.place_submodules)
             }
@@ -266,13 +280,13 @@ class AlignDevicesHook(ModelHook):
             for name, _ in named_module_tensors(
                 module, include_buffers=self.offload_buffers, recurse=self.place_submodules, remove_non_persistent=True
             ):
-                set_module_tensor_to_device(module, name, "meta")
+                set_module_tensor_to_device(module, name, "meta", tied_params_map=tied_params_map)
             if not self.offload_buffers and self.execution_device is not None:
                 for name, _ in module.named_buffers(recurse=self.place_submodules):
-                    set_module_tensor_to_device(module, name, self.execution_device)
+                    set_module_tensor_to_device(module, name, self.execution_device, tied_params_map=tied_params_map)
             elif self.offload_buffers and self.execution_device is not None:
                 for name in get_non_persistent_buffers(module, recurse=self.place_submodules):
-                    set_module_tensor_to_device(module, name, self.execution_device)
+                    set_module_tensor_to_device(module, name, self.execution_device, tied_params_map=tied_params_map)
 
         return module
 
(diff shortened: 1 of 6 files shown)


src/accelerate/hooks.py
```diff
@@ -248,11 +257,12 @@ def __repr__(self):
...
         )
 
-    def init_hook(self, module):
+    def init_hook(self, module, tied_params_map: Optional[Dict[int, Dict[torch.device, torch.Tensor]]] = None):
         if not self.offload and self.execution_device is not None:
             for name, _ in named_module_tensors(module, recurse=self.place_submodules):
-                set_module_tensor_to_device(module, name, self.execution_device)
+                set_module_tensor_to_device(module, name, self.execution_device, tied_params_map=tied_params_map)
         elif self.offload:
+            # TODO: validate that this tied param fix works correctly with offloading
```
fxmarty, line 265: I am not familiar about the offloading logic so I should test more this part.

tests/test_big_modeling.py
```diff
@@ -363,6 +364,56 @@ def test_dispatch_model_tied_weights(self):
         dispatch_model(model, device_map)
         self.assertIs(model.linear2.weight, model.linear1.weight)
 
+    @require_multi_gpu
+    def test_dispatch_model_tied_weights_memory(self):
```
fxmarty, line 368: This test does not pass on main but passes with this fix.

SunMarc (approved):
Thanks for digging into this. I'm okay with the proposed solution of adding a `tied_params_map` arg to access the pointer since it will not impact the code too much. I'm wondering if we should keep `retie_parameters` or not since the weights should be tied with this PR. However, we can keep it for more security since the goal of this PR is to make sure that we don't have memory spike due to the tied weights. 

LMK how the test with the offload goes ! (normally, there shouldn't be any issues since we put tied weights on the same device + we store the weights when offloading before `attach_align_device_hook_on_blocks` in `dispatch_model` -> no impact from this PR)

src/accelerate/hooks.py
```diff
@@ -266,13 +276,13 @@ def init_hook(self, module):
             for name, _ in named_module_tensors(
                 module, include_buffers=self.offload_buffers, recurse=self.place_submodules, remove_non_persistent=True
             ):
-                set_module_tensor_to_device(module, name, "meta")
+                set_module_tensor_to_device(module, name, "meta", tied_params_map=tied_params_map)
```
SunMarc, line 279: We probably don't need the `tied_params_map` since we are sending them to the `meta` device.

src/accelerate/utils/modeling.py
```diff
@@ -831,7 +855,8 @@ def get_balanced_memory(
         model (`torch.nn.Module`):
             The model to analyze.
         max_memory (`Dict`, *optional*):
-            A dictionary device identifier to maximum memory. Will default to the maximum memory available if unset.
+            A dictionary device identifier to maximum memory (in bytes). Will default to the maximum memory available
+            if unset.
```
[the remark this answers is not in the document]
fxmarty: Oh okay, I will fix.

tests/test_big_modeling.py
```diff
@@ -363,6 +364,56 @@ def test_dispatch_model_tied_weights(self):
...
+        )
+        model.linear2.weight = model.linear0.weight
+        model.linear3.weight = model.linear0.weight
+        model.linear4.weight = model.linear0.weight
+
+        x = torch.randn(5, 5000)
+        with torch.no_grad():
+            expected = model(x)
+
+        # We should need only 5000 * 5000 * 32 // 8 * 1e-6 = 100 MB on the device 0 for the three linear weights.
```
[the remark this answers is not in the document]
fxmarty: Yes sorry the correct comment is: `We should need only 5000 * 5000 * 32 // 8 * 1e-6 = 100 MB on the device 0 for the four linear weights.`. Weights 0, 2, 3, 4 are tied and on device 0, taking 5000 * 5000 * 32 // 8 * 1e-6 = 100 MB.

fxmarty:
@SunMarc It appears that disk offloading with tied weights uses more memory than needed as well: the `new_value` data_ptr [here](https://github.com/huggingface/accelerate/blob/fce61a99eca8cef2a65726d7f4feac41889691cf/src/accelerate/utils/modeling.py#L370) is not the same as for the other tied weights.

SunMarc:
> An other (better) option could be not to offload at all tied weights which have an other shared weight already on device

This should already be the case no when we infer automatically the `device_map` no ?

src/accelerate/utils/modeling.py
```diff
@@ -302,6 +307,24 @@ def set_module_tensor_to_device(
...
+        module._parameters[tensor_name] = tied_params_map[value.data_ptr()][device]
+        return
+    elif (
+        tied_params_map is not None
+        and old_value.data_ptr() in tied_params_map
+        and device in tied_params_map[old_value.data_ptr()]
+    ):
+        module._parameters[tensor_name] = tied_params_map[old_value.data_ptr()][device]
+        return
+
```
fxmarty, line 327: @SunMarc WDYT?

fxmarty:
@Giuseppe5 @SunMarc The `cpu` offloading case is now working well. See below for the motivation of the changes in `set_module_tensor_to_device`, and the comments for the changes in `pre_forward`, `post_forward`.

src/accelerate/utils/modeling.py
```diff
@@ -302,6 +307,24 @@ def set_module_tensor_to_device(
...
+    # Treat the case where old_value (or a custom `value`, typically offloaded to RAM/disk) belongs to a tied group, and one of the weight
+    # in the tied group has already been dispatched to the device, by avoiding reallocating memory on the device and just copying the pointer.
+    if (
+        value is not None
+        and tied_params_map is not None
+        and value.data_ptr() in tied_params_map
+        and device in tied_params_map[value.data_ptr()]
+    ):
+        module._parameters[tensor_name] = tied_params_map[value.data_ptr()][device]
+        return
```
fxmarty, line 319: We need to treat the case `value is not None` separately. This is because it corresponds to the offloading case, where (at least for RAM offloading) `value` from `self.weights_map` is indeed shared for tied weights.

When we move `self.weights_map[name]` to device, we want to do it only once. Especially, given that we may have nested modules that refer to the same data (see `test_dispatch_model_tied_weights_memory_with_nested_offload` for an example), we may have several `pre_forward` calls that use the same data, so we need that here.

src/accelerate/utils/modeling.py
```diff
@@ -397,6 +420,22 @@ def set_module_tensor_to_device(
...
+        and device not in tied_params_map[old_value.data_ptr()]
+    ):
+        tied_params_map[old_value.data_ptr()][device] = new_value
+    elif (
+        value is not None
+        and tied_params_map is not None
+        and value.data_ptr() in tied_params_map
+        and device not in tied_params_map[value.data_ptr()]
+    ):
+        tied_params_map[value.data_ptr()][device] = new_value
```
fxmarty, line 437: Same reasoning. Note that we need to remove the data `new_value` added here in `tied_params_map` in the `post_forward`, as we are using offloading and want the garbage collector to clear the on-device data.

tests/test_big_modeling.py
```diff
@@ -363,6 +364,199 @@ def test_dispatch_model_tied_weights(self):
...
+        self.assertTrue((free_memory_bytes_after_dispatch - free_memory_bytes_before_dispatch) * 1e-6 < 130)
+
+        original_pointer = model.compute1._hf_hook.weights_map["weight"].data_ptr()
+
+        with torch.no_grad():
+            try:
+                output = model(x)
+            except torch.cuda.OutOfMemoryError as e:
+                raise torch.cuda.OutOfMemoryError(
+                    f"OOM error in dispatch_model. This is a bug and should not happen, see test_dispatch_model_tied_weights_memory. {e}"
```
SunMarc, line 544:
```suggestion
                    f"OOM error in dispatch_model. This is a bug and should not happen, see test_dispatch_model_tied_weights_memory_with_nested_offload. {e}"
```

tests/test_big_modeling.py
```diff
@@ -363,6 +364,199 @@ def test_dispatch_model_tied_weights(self):
...
+        n_vals = (free_memory_bytes - required_memory_bytes - int(50e6)) // (32 // 8)
+        foo = torch.rand(n_vals, device="cuda:0")  # noqa: F841
+
+        # If this does OOM: there is an issue in somewhere in dispatch_model, memory of tied weights is duplicated.
+        try:
+            dispatch_model(model, device_map)
+        except torch.cuda.OutOfMemoryError as e:
+            raise torch.cuda.OutOfMemoryError(
+                f"OOM error in dispatch_model. This is a bug and should not happen, see test_dispatch_model_tied_weights_memory. {e}"
+            )
```
SunMarc, line 464:
```suggestion
                f"OOM error in dispatch_model. This is a bug and should not happen, see test_dispatch_model_tied_weights_memory_with_offload. {e}"
            )
```

src/accelerate/hooks.py
```diff
@@ -481,14 +540,32 @@ def attach_align_device_hook_on_blocks(
...
+        tied_params = find_tied_parameters(module)
+
+        tied_params_map = {}
+        for group in tied_params:
+            for param_name in group:
+                # data_ptr() is enough here, as `find_tied_parameters` finds tied params simply by comparing `param1 is param2`, so we don't need
+                # to care about views of tensors through storage_offset.
+                data_ptr = recursive_getattr(module, param_name).data_ptr()
+                tied_params_map[data_ptr] = {}
+
```
SunMarc, line 561: Maybe we put this logic in `dispatch_model` since we only call it once + we already call `find_tied_parameters(module)` in `dispatch_model` just before this function.

SunMarc (approved):
Thanks for taking care of the nested offload logic ! I've left a few comments. Can you have a second look @muellerzr ?

tests/test_big_modeling.py
```diff
@@ -363,6 +364,199 @@ def test_dispatch_model_tied_weights(self):
...
+        model.linear3.weight = model.linear0.weight
+        model.linear4.weight = model.linear0.weight
+
+        x = torch.randn(5, 5000)
+        with torch.no_grad():
+            expected = model(x)
+
+        # We should need only 5000 * 5000 * 32 // 8 * 1e-6 = 100 MB on the device 0 for the four linear weights.
+        device_map = {"linear0": 0, "linear1": 1, "linear2": 0, "linear3": "cpu", "linear4": 0}
+
```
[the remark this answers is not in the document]
fxmarty: This does pass, but I will simply remove this test as it is not a real case with `infer_auto_device_map`.

src/accelerate/hooks.py
```diff
@@ -266,32 +279,67 @@ def init_hook(self, module):
             for name, _ in named_module_tensors(
                 module, include_buffers=self.offload_buffers, recurse=self.place_submodules, remove_non_persistent=True
             ):
+                # When using disk offloading, we can not rely on `weights_map[name].data_ptr()` as the reference pointer,
+                # as we have no guarantee that safetensors' `file.get_tensor()` will always give the same pointer.
+                # As we have no reliable way to track the shared data pointer of tied weights in this case, we use tied_params_names: List[str]
+                # to add on the fly pointers to `tied_params_map` in the pre_forward call.
+                if tied_params_map is not None and recursive_getattr(module, name).data_ptr() in tied_params_map:
+                    self.tied_params_names.add(name)
```
fxmarty, line 287: Workaround for disk offloading.

fxmarty:
@SunMarc The disk offloading case is a bit more tricky, as we have no reliable way to know the cpu pointer to tied weights. This is fixed in https://github.com/huggingface/accelerate/pull/2330/commits/30631a65be7bd7c35952a327693ff88fc1e76887, see comments there.

SunMarc (approved):
LGTM ! Thanks for adding comments and great tests. The pre_forward and post_forward are indeed a little bit bloated but it's fine. Please have a second look @muellerzr before we merge this

muellerzr (approved):
Thanks, overall this looks good to me. Thanks for adding in a few tests!

muellerzr:
Looks like this is causing a test to fail @fxmarty (cc @SunMarc): 
```bash
FAILED tests/test_big_modeling.py::BigModelingTester::test_dispatch_model_tied_weights_memory_with_nested_offload_disk
```

It fails on the single-GPU runner

fxmarty:
@muellerzr thanks will fix

PR #2342, opened 2024-01-16
Fix `block_size` picking in `megatron_lm_gpt_pretraining` example.

nilq:
# What does this PR do?

If no `block_size` is explicitly passed, and the tokenizer happens to have a `model_max_length` of less than 1024, the current example will produce a bad `block_size`. With this change, we only cap `block_size` to 1024 if `tokenizer.model_max_length` is actually greater than 1024.

## Who can review?

Tagging @pacman100 / @sgugger as the authors of the surrounding code.


pacman100 (approved):
Thank you @nilq for the fix, LGTM!

muellerzr (approved):
Thanks!

PR #2310, opened 2024-01-06
Fix batch_size sanity check in `prepare_data_loader`

izhx:
# What does this PR do?

Fixes #2285

The `dataloader.batch_size` is set to `None` in `DataLoader.__init__` if `batch_sampler` is given.

I made the `split_batches` sanity check depend on `dataloader.batch_sampler.batch_size` if `dataloader.batch_size` is `None`

I have pass passed `pytest tests/test_data_loader.py`, `make style` and `make quality` on my linux GPU server.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@muellerzr

PR #2344, opened 2024-01-16
Fix batch_size sanity check logic for `split_batches `

### LINKED ISSUE #2285: `accelerator.prepare(dataloader)` sanity check fails when `batch_sampler` is given and `split_batches` is `True`
### System Info

```Shell
Ubuntu 20.04, Python 3.10, accelerate 0.25.0, torch 2.1.2
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

I am attempting to manually pass a `BatchSampler` to `DataLoader` while enabling the `split_batches` option.

```
from torch.utils.data import Dataset, DataLoader, BatchSampler, RandomSampler


class SimpleDataset(Dataset):
    def __init__(self, data):
        self.data = data

    def __getitem__(self, index) -> dict:
        return self.data[index]

    def __len__(self):
        return len(self.data)

accelerator = Accelerator(split_batches=True)

dataset = SimpleDataset([0, 1, 2, 3, 4])
batch_sampler = BatchSampler(RandomSampler(dataset), 2, True)
loader = DataLoader(dataset, batch_sampler=batch_sampler)
accelerator.prepare(loader)
```

But it fails at the sanity check of https://github.com/huggingface/accelerate/blob/ceb7c699bc36bdb3bbf32cceaaca2d1ceaf62dae/src/accelerate/data_loader.py#L827

```
  File "test.py", line 20, in <module>
    accelerator.prepare(loader)
  File "/envs/flash/lib/python3.10/site-packages/accelerate/accelerator.py", line 1213, in prepare
    result = tuple(
  File "/envs/flash/lib/python3.10/site-packages/accelerate/accelerator.py", line 1214, in <genexpr>
  File "/envs/flash/lib/python3.10/site-packages/accelerate/accelerator.py", line 1092, in _prepare_one
    return self.prepare_data_loader(obj, device_placement=device_placement)
  File "/envs/flash/lib/python3.10/site-packages/accelerate/accelerator.py", line 1788, in prepare_data_loader
    prepared_data_loader = prepare_data_loader(
  File "/envs/flash/lib/python3.10/site-packages/accelerate/data_loader.py", line 820, in prepare_data_loader
    if split_batches and dataloader.batch_size > 1 and dataloader.batch_size % num_processes != 0:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

The `dataloader.batch_size` is set to `None` in `dataloader.__init__` if `batch_sampler` is given, see https://github.com/pytorch/pytorch/blob/f6dfbffb3bb46ada6fe66b5da4f989f9d4d69b3c/torch/utils/data/dataloader.py#L336
```
        if batch_sampler is not None:
            # auto_collation with custom batch_sampler
            if batch_size != 1 or shuffle or sampler is not None or drop_last:
                raise ValueError('batch_sampler option is mutually exclusive '
                                 'with batch_size, shuffle, sampler, and '
                                 'drop_last')
            batch_size = None
            drop_last = False
```

### Expected behavior

Pass the check.

izhx:
# What does this PR do?

Fixes #2285 and [PR/2310 comment: Unfortunately, this broke Megatron DL](https://github.com/huggingface/accelerate/pull/2310#issuecomment-1893072023)

Some libraries may not follow PyTorch's design of `BatchSampler`.
If the `batch_size` cannot be found, a warning is left to notify the user. 

Do anyone have suggestions for such situations?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr @BenjaminBossan


muellerzr (commented):
I think this solution makes sense, but I know this originates from Megatron so cc @pacman100 for guidance.

pacman100 (commented):
Thank you @izhx for the fixes in cases when the Batch Sampler doesn't conform to PyTorch. Makes Sense.

muellerzr (approved):
If Sourab is happy, I'm happy. But would *definitely* be happier if @stas00 could also give this an all-clear? :)

stas00:
While this PR makes the error user-friendlier - this approach swipes the problem under the carpet. If anything it should be an assert and not a warning. i.e. if the user wants `split_batches` this code path must not proceed.

But there is more to it - I have never set `split_batches=True` - so somewhere in the stack I reported at https://github.com/huggingface/accelerate/pull/2310#issuecomment-1893072023 Accelerate sets it to True by default w/o user's intent to do so.

I think you have this problem here as well:
https://github.com/huggingface/accelerate/blob/c0a94724223b34a0e2e93f1768c069584ba48ca0/src/accelerate/data_loader.py#L487C9-L492
it always calls `sampler.batch_size`. so that code also probably should check `hasattr` and assert if it is False.

src/accelerate/data_loader.py
```diff
@@ -845,12 +845,10 @@ def prepare_data_loader(
             if hasattr(dataloader.batch_sampler, "batch_size"):
                 batch_size_for_check = dataloader.batch_sampler.batch_size
             else:
-                logger.warning(
-                    f"Make sure that you are using a correct batch size for `split_batches`. "
-                    f"We can not check it since dataloader.batch_size is None and your batch_sampler "
-                    f"{type(dataloader.batch_sampler)} does not have `batch_size` attribute."
+                raise ValueError(
+                    "Can not check batch_size for `split_batches` mode since dataloader.batch_size is None and "
+                    f"your batch_sampler {type(dataloader.batch_sampler)} does not have `batch_size` attribute."
```
stas00, line 850: Excellent. A small mod proposal to define it more rigorously:

```suggestion
                raise ValueError(
                    "In order to use `split_batches==True` you must have a `batch_size` attribute either in the passed `dataloader` or `dataloader.batch_sampler` objects, and it has to return a natural number. Your batch sampler `{type(dataloader.batch_sampler)}` does not have the `batch_size` attribute set."
```

stas00: And as I wrote in the error you probably also want to check that it doesn't return 0, right?

so probably need to add that check as well?

izhx: If you are referring to `batch_size % num_process == 0`, we don't need to add it because it will be checked later by line 853.

stas00: It won't check for == 0 because it does:

`if batch_size_for_check > 1 and batch_size_for_check % num_processes != 0:`

But we can leave it as is for now, it's unlikely it'd be set to that

stas00 (approved):
let's land this one, @muellerzr?

muellerzr (approved):
Thanks for iterating!

PR #2345, opened 2024-01-16
torch-native pipeline parallelism for big models

muellerzr:
## Example use:

```python
import torch
from accelerate.inference import prepare_pippy
from transformers import T5ForConditionalGeneration, T5Config

set_seed(42)

config = T5Config()
model = T5ForConditionalGeneration(config)
model.eval()

# Create example inputs for the model
input = torch.randint(
    low=0,
    high=config.vocab_size,
    size=(2, 1024),  # bs x seq_len
    device="cpu",
    dtype=torch.int64,
    requires_grad=False,
)

example_inputs = {"input_ids": input, "decoder_input_ids": input}

model = prepare_pippy(model, example_kwargs=example_inputs)

args = (
    example_inputs["input_ids"].to("cuda:0"),
    example_inputs["decoder_input_ids"].to("cuda:0")
)
with torch.no_grad():
    output = model(*args)
```

## Speed up:

Using 2x4090's in full precision

### Bert

|  | Accelerate/Sequential | PiPPy + Accelerate |
|---|---|---|
| First batch | 0.2137s | 0.3119s |
| Average of 5 batches | 0.0099s | **0.0062s** |

### GPT2

|  | Accelerate/Sequential | PiPPy + Accelerate |
|---|---|---|
| First batch | 0.1959s | 0.4189s |
| Average of 5 batches | 0.0205s | **0.0126s** |

### T5

|  | Accelerate/Sequential | PiPPy + Accelerate |
|---|---|---|
| First batch | 0.2789s | 0.3809s |
| Average of 5 batches | 0.0198s | **0.0166s** |


src/accelerate/inference.py
```diff
@@ -0,0 +1,96 @@
...
+
+def prepare_pippy(model, device_map="auto", example_args=(), example_kwargs={}):
+    """
+    Wraps `model` for PipelineParallelism
+    """
+    example_args = send_to_device(example_args, "cpu")
+    example_kwargs = send_to_device(example_kwargs, "cpu")
+    if device_map == "auto":
+        device_map = generate_device_map(model, PartialState().num_processes)
+    stage = build_pipeline(model, device_map, example_args, example_kwargs)
```
SunMarc, line 83: Just a thought about how to handle the split points. 
- 1) We only expose `device_map` with predefined options ("sequential", "balanced_pippy")
- 2) We let the user use a custom `device_map`. For the custom case, it can be complicated since the user needs to be careful about the order (OrderedDict()) and he needs to attribute the gpu in a sequential manner because of that `split_points.append(next(k for k, v in device_map.items() if v == i))`. So that can be quite complicated. 
- 3) We let the user let his own split points `List[str]`. 
I think that 1) is a must. between 2) and 3), I prefer 3) since it is easier for the user.

src/accelerate/inference.py
```diff
@@ -0,0 +1,96 @@
...
+def build_pipeline(model, device_map, args, kwargs) -> PipelineStage:
+    """
+    Attaches the split points to the model based on `self.device_map` and generates a `PipelineStage`. Requires passing
+    in needed `args` and `kwargs` as the model needs on the CPU.
+    """
+    # We need to annotate the split points in the model for PiPPy
+    state = PartialState()
+    split_points = []
+    for i in range(1, state.num_processes):
+        split_points.append(next(k for k, v in device_map.items() if v == i))
```
SunMarc, line 55: Let's pass split_points in `build_pipeline` instead of `device_map`. WDYT ? It will be better if we decide to let the user pass split points.

src/accelerate/inference.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def prepare_pippy(model, device_map="auto", example_args=(), example_kwargs={}):
+    """
+    Wraps `model` for PipelineParallelism
+    """
+    example_args = send_to_device(example_args, "cpu")
+    example_kwargs = send_to_device(example_kwargs, "cpu")
+    if device_map == "auto":
+        device_map = generate_device_map(model, PartialState().num_processes)
```
SunMarc, line 82: For debug purpose, I'm thinking that we could save the `split_points` as `model.hf_split_points`. Not sure about the `device_map` as `model.hf_device_map `since if we let the user pass the `split_points`, we might don't have the `device_map`. However, I think that we should be able to recreate it using `split_points`.

SunMarc (commented):
Very cool API ! I like the design and how easily it is to use. I left a few comments around the `split_points` mainly.

src/accelerate/inference.py
```diff
@@ -0,0 +1,96 @@
...
+
+    model_forward = partial(pippy_forward, forward=model.pippy_stage.forward)
+
+    def forward(*args, **kwargs):
+        return model_forward(*args, **kwargs)
+
+    # To act like a decorator so that it can be popped when doing `extract_model_from_parallel`
+    forward.__wrapped__ = model_forward
+    model.forward = forward
+    return stage
```
SunMarc, line 96: You are returning the stage. I think i should be the model no ?

src/accelerate/inference.py
muellerzr: Not sure if I like this yet, it seems clunky in implementation but I'll keep at it.

src/accelerate/inference.py
muellerzr: Agreed to do 1 and 3

src/accelerate/inference.py
```diff
@@ -59,7 +59,9 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
 def pippy_forward(forward, *args, **kwargs):
     state = PartialState()
     output = None
-    if state.is_local_main_process:
+    if state.num_processes == 1:
+        output = forward(*args, **kwargs)
```
muellerzr, line 63: Nice!

src/accelerate/inference.py
```diff
@@ -0,0 +1,98 @@
...
+    # Add a chunk to deal with potential extra shared memory instances
+    memory = math.ceil(float(value)) * 1.1
+    memory = f"{memory} {ending}"
+    device_map = infer_auto_device_map(
+        model,
+        max_memory={i: memory for i in range(num_processes)},
+        no_split_module_classes=no_split_module_classes,
+        clean_result=False,
+    )
+    return device_map
```
SunMarc, line 42: Let's add the `max_memory` arg, so that the user can set it by himself if we want it. So if `max_memory` is not None and `num_processes != 1`, we would change its value with the above allocation. Then call `infer_auto_device_map` once the end of the function.

src/accelerate/inference.py
```diff
@@ -0,0 +1,98 @@
...
+    Wraps `model` for PipelineParallelism
+    """
+    state = PartialState()
+    example_args = send_to_device(example_args, "cpu")
+    example_kwargs = send_to_device(example_kwargs, "cpu")
+    if split_points == "auto":
+        device_map = generate_device_map(model, state.num_processes, no_split_module_classes=no_split_module_classes)
+        split_points = []
+        for i in range(1, state.num_processes):
+            split_points.append(next(k for k, v in device_map.items() if v == i))
```
SunMarc, line 84: it would be great to have a sanity check, to make sure that we indeed have `self.num_processes` split points when we are generating the `split_points` + when the user manually pass them

src/accelerate/inference.py
```diff
@@ -0,0 +1,98 @@
...
+    model._original_call = model.__call__
+    model.pippy_stage = stage
+    model.hf_split_points = split_points
+
+    def forward(*args, **kwargs):
+        return pippy_forward(stage.forward, *args, **kwargs)
+
+    # To act like a decorator so that it can be popped when doing `extract_model_from_parallel`
+    model_forward = MethodType(forward, model)
+    forward.__wrapped__ = model_forward
```
SunMarc, line 96: We can keep them for now but these lines create an infinite recursion loop with `generate`

src/accelerate/inference.py
```diff
@@ -0,0 +1,98 @@
...
+    return output
+
+
+def prepare_pippy(model, split_points="auto", no_split_module_classes=[], example_args=(), example_kwargs={}):
+    """
+    Wraps `model` for PipelineParallelism
+    """
+    state = PartialState()
+    example_args = send_to_device(example_args, "cpu")
+    example_kwargs = send_to_device(example_kwargs, "cpu")
```
SunMarc, line 79: The api from pippy may change but we should keep in mind that dealing with kwarg is a mess in pippy. Passed kwargs are expected but it applies only for tensor args. The other args are "burned".

SunMarc (approved):
The API is on good shape ! Let's document the main functions a bit and we can merge it. I left a few comments but nothing blocking.

src/accelerate/inference.py
```diff
@@ -0,0 +1,98 @@
...
+
+def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
+    """
+    Attaches the split points to the model based on `self.device_map` and generates a `PipelineStage`. Requires passing
+    in needed `args` and `kwargs` as the model needs on the CPU.
+    """
+    # We need to annotate the split points in the model for PiPPy
+    state = PartialState()
+    annotate_split_points(model, {split_point: PipeSplitWrapper.SplitPoint.BEGINNING for split_point in split_points})
+    pipe = Pipe.from_tracing(model, num_chunks=state.num_processes, example_args=args, example_kwargs=kwargs)
```
SunMarc, line 53: We are hardcoding `num_chunks` by `state.num_processes`. This is def a good since Ke that we would need at least `num_processes` for the pipeline to have good overlap. However, we should maybe let the user change this value and test it by themselves. I think this is a really important arg and the user should try to understand its effect.

muellerzr: Documented

src/accelerate/inference.py
muellerzr: Made a note about this

src/accelerate/inference.py
muellerzr: Made a note in the docstring warning users. We can do so more thoroughly in the true documetation

kwen2501 (approved):
Thanks a lot for the integration effort! 
LGTM!

src/accelerate/inference.py
```diff
@@ -0,0 +1,155 @@
...
+    elif state.is_last_process:
+        output = forward()
+    else:
+        forward()
+    return output
+
+
+def prepare_pippy(
+    model, split_points="auto", no_split_module_classes=None, example_args=(), example_kwargs={}, num_chunks=None
+):
```
SunMarc, line 104: Learned from @fxmarty that it is dangerous to set a default dictionary. https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python
Let's set it to `example_kwargs: Optional[Dict[str, Any]] = None` just like in `from_tracing` docstring.

src/accelerate/inference.py
```diff
@@ -0,0 +1,155 @@
...
+    if state.num_processes == 1:
+        output = forward(*args, **kwargs)
+    elif state.is_local_main_process:
+        found_batch_size = find_pippy_batch_size(args, kwargs)
+        if found_batch_size is None:
+            raise ValueError("Could not find batch size from args or kwargs")
+        else:
+            if found_batch_size != state.num_processes:
+                args = pad_input_tensors(args, found_batch_size, state.num_processes)
+                kwargs = pad_input_tensors(kwargs, found_batch_size, state.num_processes)
```
SunMarc, line 93: I think we should replace `state.num_processes` by `num_chunk`

tests/test_utils.py
```diff
@@ -237,3 +237,68 @@ def test_pad_across_processes(self):
...
+        result = pad_input_tensors(batch, batch_size, num_processes)
+        # We should expect there to be 6 items now
+        assert result.shape == torch.Size([6, 4, 4])
+
+        # Eighth base case: 6 processes, batch size of 61
+        num_processes = 6
+        batch_size = 61
+        batch = torch.rand(batch_size, 4, 4)
+        result = pad_input_tensors(batch, batch_size, num_processes)
+        # We should expect there to be 6 items now
```
SunMarc, line 303:
```suggestion
        # We should expect there to be 66 items now
```

src/accelerate/inference.py
```diff
@@ -0,0 +1,155 @@
...
+            A list of class names for layers we don't want to be split.
+        example_args (tuple of `torch.Tensor`):
+            The expected inputs for the model that uses order-based inputs. Recommended to use this method if possible.
+        example_kwargs (dict of `torch.Tensor`)
+            The expected inputs for the model that uses dictionary-based inputs. This is a *highly* limiting structure
+            that requires the same keys be present at *all* inference calls. Not recommended unless the prior condition
+            is true for all cases.
+        num_chunks (`int`):
+            The number of different stages the Pipeline will have. By default it will assign one chunk per GPU, but
+            this can be tuned and played with. In general one should have num_chunks > num_gpus.
```
SunMarc, line 124: Hi @kwen2501, could you help us to better understand what this arg do ? Is this the same arg as mentioned in this [doc](https://huggingface.co/docs/transformers/v4.15.0/parallelism#:~:text=PP%20introduces%20a,B0%2C2%2C%20B0%2C1%2C%20B0%2C0) ? Let's say we set a `num_chunks` of 4. It will split the data of `batch_size` n into 4 chunks of batch_size `n/4` ? And if we set `num_chunks=1`, we have the naive MP case. It that right ?

SunMarc (approved):
Thx for iterating ! LGTM

src/accelerate/inference.py
kwen2501: That's right. Same as the definition in the doc you linked, i.e. how many microbatches we cut the input batch into.

Note that `num_chunks` is not necessarily the number of pipeline stages. For example, you can still have `num_chunks=1` but 4 pipeline stages. That 1 chunk would pass through the 4 stages in series. But since there are no more data chunks, a stage (GPU) would become idle once the chunk passes through it.

kwen2501: That said, I agree with the comment that says:
> In general one should have num_chunks >= num_gpus

(maybe adding a "=" sign too)

SunMarc: Thanks for clarifying @kwen2501 !

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
-Distributed inference is a common use case, especially with natural language processing (NLP) models. Users often want to
-send a number of different prompts, each to a different GPU, and then get the results back. This also has other cases
-outside of just NLP, however for this tutorial we will focus on just this idea of each GPU receiving a different prompt,
-and then returning the results.
+Distributed inference can fall into three brackets:
 
-## The Problem
+1. Loading an entire model onto each GPU and sending chunks of a batch through each model at a time
+2. Load parts of a model onto each GPU and process a single input at one time
+3. Load parts of a model onto each GPU and used what is called scheduled Pipeline Parallelism to combine the two prior techniques.
```
kwen2501, line 22: used --> use

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
 
 On the first GPU, the prompts will be `["a dog", "a cat"]`, and on the second GPU it will be `["a chicken", "a chicken"]`.
 Make sure to drop the final sample, as it will be a duplicate of the previous one.
+
+## A more memory-efficient version (experimental)
+
+This next part will discuss using *pipeline parallelism*. This is an **experimental** API utilizing the [PiPPy library by PyTorch](https://github.com/pytorch/PiPPy/) as a native solution. 
+
+The general idea with pipeline parallism is say you have 4 GPUs, and a model big enough it can be *split* on four GPUs using `device_map="auto"`. What this version will do is you can send in 4 inputs at at time (for example here, any amount works) and each models chunk will work on an ainput, then recieve the next input after the prior chunk finished it making it *much* more efficient **and faster** than the prior version. Here's a visual taken from the PyTorch repository:
```
kwen2501, line 147: each models chunk --> each model chunk
an ainput --> an input

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+<Tip>
+    There are a variety of parameters you can pass through to `prepare_pippy`:
+    * `split_points` will let you determine where to split the model at. By default we use wherever `device_map="auto" declares
+    * `num_chunks` can be used to determine how the batch will be split and sent to the model itself (so `num_chunks=1` with four split points/four GPUs would have a naive MP where a single input gets passed between the four layer split points)
+</Tip>
+
+From here all that's left is to actually perform the distributed inference!
+
+<Tip warning={true}>
+    When passing in inputs, while using `kwargs` are supported currently those are even *more* experimental, so it's highly recommended to just simply pass inptus in as a tuple of arguments.
```
kwen2501, line 205: inptus --> inputs

kwen2501 (approved):
Thanks for writing the doc so quick! Looks good to me!

src/accelerate/inference.py
```diff
@@ -90,7 +91,7 @@ def pippy_forward(forward, *args, **kwargs):
         if found_batch_size is None:
             raise ValueError("Could not find batch size from args or kwargs")
         else:
-            if found_batch_size != state.num_processes:
+            if found_batch_size != num_chunks:
                 args = pad_input_tensors(args, found_batch_size, state.num_processes)
                 kwargs = pad_input_tensors(kwargs, found_batch_size, state.num_processes)
```
kwen2501: Do we want to pad with `num_chunks` or `state.num_processes`?

muellerzr: Indeed should be `num_chunks` all the way through, bad copy/paste caught me :)

muellerzr:
cc @MKhalusova for the docs!

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
 # Distributed Inference with 🤗 Accelerate
 
-Distributed inference is a common use case, especially with natural language processing (NLP) models. Users often want to
-send a number of different prompts, each to a different GPU, and then get the results back. This also has other cases
-outside of just NLP, however for this tutorial we will focus on just this idea of each GPU receiving a different prompt,
-and then returning the results.
+Distributed inference can fall into three brackets:
 
-## The Problem
+1. Loading an entire model onto each GPU and sending chunks of a batch through each model at a time
```
MKhalusova, line 20: Suggestion: "Loading an entire model onto each GPU and sending chunks of a batch through each GPU's model copy at a time"

I think it's a bit clearer this way.

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
+2. Load parts of a model onto each GPU and process a single input at one time
```
MKhalusova, line 21:
```suggestion
2. Loading parts of a model onto each GPU and processing a single input at one time
```

MKhalusova: For consistency with the first item in the list, let's say "loading" instead of "load".

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
+3. Load parts of a model onto each GPU and use what is called scheduled Pipeline Parallelism to combine the two prior techniques.
```
MKhalusova, line 22:
```suggestion
3. Loading parts of a model onto each GPU and using what is called scheduled Pipeline Parallelism to combine the two prior techniques. 
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
-outside of just NLP, however for this tutorial we will focus on just this idea of each GPU receiving a different prompt,
-and then returning the results.
+Distributed inference can fall into three brackets:
 
-## The Problem
+1. Loading an entire model onto each GPU and sending chunks of a batch through each model at a time
+2. Load parts of a model onto each GPU and process a single input at one time
+3. Load parts of a model onto each GPU and use what is called scheduled Pipeline Parallelism to combine the two prior techniques. 
+
+We're going to go through the first and the last, showcasing how to do each as they are more realistic scenarios.
```
MKhalusova, line 24:
```suggestion
We're going to go through the first and the last bracket, showcasing how to do each as they are more realistic scenarios.
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
+
+
+## Sending chunks of inputs automatically to each loaded model
```
MKhalusova, line 27:
```suggestion
## Sending chunks of a batch automatically to each loaded model
```

MKhalusova: To be consistent with the first item in the list

docs/source/usage_guides/distributed_inference.md
```diff
@@ -15,12 +15,18 @@ rendered properly in your Markdown viewer.
...
+
+This is the most memory-intensive solution, as it requires each GPU keeps a full copy of the model in memory at a given time.
```
MKhalusova, line 29:
```suggestion
This is the most memory-intensive solution, as it requires each GPU to keep a full copy of the model in memory at a given time. 
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
 
 On the first GPU, the prompts will be `["a dog", "a cat"]`, and on the second GPU it will be `["a chicken", "a chicken"]`.
 Make sure to drop the final sample, as it will be a duplicate of the previous one.
+
+## A more memory-efficient version (experimental)
```
MKhalusova, line 143: This is a vague title. Let's rename to something a bit more clear. 
E.g. "Memory-efficient pipeline parallelism (experimental)"

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+This next part will discuss using *pipeline parallelism*. This is an **experimental** API utilizing the [PiPPy library by PyTorch](https://github.com/pytorch/PiPPy/) as a native solution. 
+
+The general idea with pipeline parallism is say you have 4 GPUs, and a model big enough it can be *split* on four GPUs using `device_map="auto"`. What this version will do is you can send in 4 inputs at at time (for example here, any amount works) and each model chunk will work on an input, then recieve the next input after the prior chunk finished it making it *much* more efficient **and faster** than the prior version. Here's a visual taken from the PyTorch repository:
```
MKhalusova, line 147:
```suggestion
The general idea with pipeline parallelism is: say you have 4 GPUs and a model big enough it can be *split* on four GPUs using `device_map="auto"`. With this method you can send in 4 inputs at a time (for example here, any amount works) and each model chunk will work on an input, then receive the next input once the prior chunk finished, making it *much* more efficient **and faster** than the method described earlier. Here's a visual taken from the PyTorch repository:
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+![PiPPy example](https://camo.githubusercontent.com/681d7f415d6142face9dd1b837bdb2e340e5e01a58c3a4b119dea6c0d99e2ce0/68747470733a2f2f692e696d6775722e636f6d2f657955633934372e706e67)
+
+To use this with Accelerate, we have created a [model zoo](https://github.com/muellerzr/pippy-device-map-playground/) showcasing a number of different models and situations to do so. In this tutorial we'll take GPT2 however across two gpus.
```
MKhalusova, line 151:
```suggestion
To illustrate how you can use this with Accelerate, we have created a [model zoo example](https://github.com/muellerzr/pippy-device-map-playground/) showcasing a number of different models and situations. In this tutorial, we'll show this method for GPT2 across two GPUs.
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+Before anything, please make sure you have the latest pippy installed by performing:
```
MKhalusova, line 153:
```suggestion
Before you proceed, please make sure you have the latest pippy installed by running the following:
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+```bash
+pip install torchpippy
+```
+
+We require at least version 0.2.0, please perform `pip show torchpippy` to check this!
```
MKhalusova, line 159:
```suggestion
We require at least version 0.2.0. To confirm that you have the correct version, run `pip show torchpippy`.
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+First we need to create the model on the CPU:
```
MKhalusova, line 161:
```suggestion
Start by creating the model on the CPU:
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+```{python}
+from transformers import GPT2ForSequenceClassification, GPT2Config
+
+config = GPT2Config()
+model = GPT2ForSequenceClassification(config)
+model.eval()
+```
+
+Next we need to create some example inputs to use. These help PiPPy trace the model.
```
MKhalusova, line 171:
```suggestion
Next you'll need to create some example inputs to use. These help PiPPy trace the model.
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+<Tip warning={true}>
+    However you make this example will determine the relative batch size that will be used/passed
+    through the model at a given time, so make sure to remember them!
```
MKhalusova, line 175: I'm not sure I understand what "them" refers to in "make sure to remember them"

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+input = torch.randint(
+    low=0,
+    high=config.vocab_size,
+    size=(2, 1024),  # bs x seq_len
+    device="cpu",
+    dtype=torch.int64,
+    requires_grad=False,
+)
+```
+Next we need to actually perform the tracing and get the model ready. To do so you simply use the [`inference.prepare_pippy`] function and it will fully wrap the model for pipeline parallism automatically:
```
MKhalusova, line 188:
```suggestion
Next we need to actually perform the tracing and get the model ready. To do so, use the [`inference.prepare_pippy`] function and it will fully wrap the model for pipeline parallelism automatically:
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+```{python}
+from accelerate.inference import prepare_pippy
+example_inputs = {"input_ids": input}
+model = prepare_pippy(model, example_args=(input,))
+```
+
+<Tip>
+    There are a variety of parameters you can pass through to `prepare_pippy`:
+    * `split_points` will let you determine where to split the model at. By default we use wherever `device_map="auto" declares
```
MKhalusova, line 198:
```suggestion
    * `split_points` lets you determine where to split the model. By default, we use wherever `device_map="auto" declares
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+```
+Next we need to actually perform the tracing and get the model ready. To do so you simply use the [`inference.prepare_pippy`] function and it will fully wrap the model for pipeline parallism automatically:
+
+```{python}
+from accelerate.inference import prepare_pippy
+example_inputs = {"input_ids": input}
+model = prepare_pippy(model, example_args=(input,))
+```
+
+<Tip>
```
MKhalusova, line 196:
```suggestion
<Tip>

```

MKhalusova: Added a newline, so that the content inside the tip would render properly

docs/source/usage_guides/distributed_inference.md
MKhalusova: Declares what? Maybe give an example?

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+    There are a variety of parameters you can pass through to `prepare_pippy`:
+    * `split_points` will let you determine where to split the model at. By default we use wherever `device_map="auto" declares
+    * `num_chunks` can be used to determine how the batch will be split and sent to the model itself (so `num_chunks=1` with four split points/four GPUs would have a naive MP where a single input gets passed between the four layer split points)
```
MKhalusova, line 199:
```suggestion
    * `num_chunks` determines how the batch will be split and sent to the model itself (so `num_chunks=1` with four split points/four GPUs will have a naive MP where a single input gets passed between the four layer split points)
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+</Tip>
+
+From here all that's left is to actually perform the distributed inference!
```
MKhalusova, line 202:
```suggestion
From here, all that's left is to actually perform the distributed inference!
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+
+<Tip warning={true}>
```
MKhalusova, line 204:
```suggestion
<Tip warning={true}>

```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+    When passing in inputs, while using `kwargs` are supported currently those are even *more* experimental, so it's highly recommended to just simply pass inputs in as a tuple of arguments.
```
MKhalusova, line 205:
```suggestion
When passing inputs, we highly recommend to pass them in as a tuple of arguments. Using `kwargs` is supported, however, this approach is experimental.
```

docs/source/usage_guides/distributed_inference.md
```diff
@@ -134,3 +139,85 @@ with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"],
...
+</Tip>
+
+```{python}
+args = some_more_arguments
+with torch.no_grad():
+    output = model(*args)
+```
+
+Afterwards all the data will be on the last GPU, which you can use the [`PartialState`] to find and extract:
```
MKhalusova, line 214:
```suggestion
When finished, all the data will be on the last GPU, which you can use the [`PartialState`] to find and extract:
```

MKhalusova (commented):
Nice work! I left a few comments to polish things in the docs a bit.

docs/source/usage_guides/distributed_inference.md
muellerzr: Done!

muellerzr:
Final comment before merging, things that still need to be done in a latter PR at some point (but okay not being in the first iteration of this joint effort):

1. Specify `balanced_pippy` device map and allow a `sequential` device_map when making the pipeline via `prepare_pippy`
2. Look into supporting `model.generate()` through an alternative hook into the model forward if possible
3. Make sure all outputs end up on the CPU so users don't need to check at the end and we can call them via a .gather
4. Migrate the `pippy-device-map-playground` examples over to here as part of our `examples` folder

(I'll be doing 3& 4 this week as a follow-up prior to release)

PR #2352, opened 2024-01-18
Allow for dynamic batch padding

muellerzr:
# What does this PR do?

This PR allows for dynamic batch padding by finding and duplicating the last item in the batch before concating everything. 

Current issue:
* `PiPPy` has issues with batch inference on GPT-2, will talk to the pippy folks about this

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -71,16 +71,13 @@ def pippy_forward(forward, *args, **kwargs):
         return None
 
     def _pad_inputs(args):
-        core = slice_tensors(args, slice(0, state.num_processes), process_index=0, num_processes=state.num_processes)
-        # Do args first
+        # Slice and copy the last input
         extra = slice_tensors(
             args,
             slice(state.num_processes, state.num_processes + 1),
-            process_index=0,
-            num_processes=state.num_processes,
         )
-        extra = concatenate([extra] * ((found_batch_size % state.num_processes) + 1))
-        args = concatenate([core, extra])
+        # Concat it to `args`
+        args = concatenate([args, extra])
         return args
 
     if state.num_processes == 1:
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -71,14 +71,16 @@ def pippy_forward(forward, *args, **kwargs):
         return None
 
     def _pad_inputs(args):
-        # Slice and copy the last input
-        extra = slice_tensors(
+        # We keep at least `num_processes` elements of the first batch to add to the end
+        first_batch = slice_tensors(
             args,
-            slice(state.num_processes, state.num_processes + 1),
+            slice(0, state.num_processes)
         )
         # Concat it to `args`
-        args = concatenate([args, extra])
-        return args
+        args = concatenate([args, first_batch])
+        # Then slice it
+        data = slice(state.num_processes, state.num_processes + found_batch_size)
+        return data
 
     if state.num_processes == 1:
         output = forward(*args, **kwargs)
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -44,6 +44,18 @@ def generate_device_map(model, num_processes: int = 1, no_split_module_classes=N
     return device_map
 
 
+def find_pippy_batch_size(args, kwargs):
+    for arg in args:
+        found_batch_size = ignorant_find_batch_size(arg)
+        if found_batch_size is not None:
+            break
+    for kwarg in kwargs.values():
+        found_batch_size = ignorant_find_batch_size(kwarg)
+        if found_batch_size is not None:
+            break
+    return found_batch_size
+
+
 def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
     """
     Attaches the split points to the model based on `self.device_map` and generates a `PipelineStage`. Requires passing
@@ -77,8 +89,9 @@ def pippy_forward(forward, *args, **kwargs):
         if found_batch_size is None:
             raise ValueError("Could not find batch size from args or kwargs")
         else:
-            args = pad_input_tensors(args, found_batch_size, state.num_processes)
-            kwargs = pad_input_tensors(kwargs, found_batch_size, state.num_processes)
+            if found_batch_size != state.num_processes:
+                args = pad_input_tensors(args, found_batch_size, state.num_processes)
+                kwargs = pad_input_tensors(kwargs, found_batch_size, state.num_processes)
         forward(*args, **kwargs)
     elif state.is_last_process:
         output = forward()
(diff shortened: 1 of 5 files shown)


src/accelerate/inference.py
```diff
@@ -59,9 +62,42 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
 def pippy_forward(forward, *args, **kwargs):
     state = PartialState()
     output = None
+
+    def _find_batch_size(args):
```
muellerzr, line 66: I'll likely move then outside `pippy_forward`, but just here as a start

src/accelerate/inference.py
```diff
@@ -59,9 +62,42 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
...
+            found_batch_size = _find_batch_size(arg)
+        if found_batch_size is None:
+            for kwarg in kwargs.values():
+                found_batch_size = _find_batch_size(kwarg)
+        if found_batch_size is None:
+            raise ValueError("Could not find batch size from args or kwargs")
+        else:
+            if (found_batch_size % state.num_processes) != 0:
+                args = _pad_inputs(args)
+                kwargs = _pad_inputs(kwargs)
```
SunMarc, line 100: This will work if we only have nested list/dict/tuple of tensors. Let's see if this works for most models.

src/accelerate/test_utils/scripts/external_deps/test_pippy.py
```diff
@@ -91,7 +95,12 @@ def test_t5():
     if state.distributed_type == DistributedType.MULTI_GPU:
         state.print("Testing GPT2...")
         test_gpt2()
+        # Issue: When modifying the tokenizer for batch GPT2 inference, there's an issue
+        # due to references
+        # NameError: cannot access free variable 'chunk_args_list' where it is not associated with a value in enclosing scope
```
SunMarc, line 100: strange issue indeed

src/accelerate/inference.py
```diff
@@ -59,9 +62,42 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
...
+    def _find_batch_size(args):
+        try:
+            return find_batch_size(arg)
+        except (ValueError, TypeError):
+            pass
+        return None
+
+    def _pad_inputs(args):
+        core = slice_tensors(args, slice(0, state.num_processes), process_index=0, num_processes=state.num_processes)
+        # Do args first
```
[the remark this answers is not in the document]
muellerzr: In my head to remove the extra, but I suppose we can keep the core and just slice at the remaining one to copy. Fair point (and a bit more efficient)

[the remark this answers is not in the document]
muellerzr: Yeah, it's there for a custom slice function, but we don't need that here!

src/accelerate/inference.py
muellerzr: We can try it on a CV model maybe? (Since it just takes in BSxCxWxH tensors)

muellerzr: It will also work on just a raw tensor :)

src/accelerate/inference.py
```diff
@@ -59,9 +62,42 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
...
+        except (ValueError, TypeError):
+            pass
+        return None
+
+    def _pad_inputs(args):
+        # Slice and copy the last input
+        extra = slice_tensors(
+            args,
+            slice(state.num_processes, state.num_processes + 1),
+        )
```
SunMarc, line 78: Are we use that the last input will be in the index state.num_processes ? Maybe it is better to duplicate the first input ?

src/accelerate/inference.py
```diff
@@ -59,9 +62,42 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
...
+        # Concat it to `args`
+        args = concatenate([args, extra])
```
SunMarc, line 80: You are only concatenating one extra input. it should be `state.num_processes - (found_batch_size % state.num_processes` no ?

SunMarc (commented):
We are practically there. Left a few comments.

src/accelerate/inference.py
muellerzr: We do first input in the `DataLoader` class so would be good to do the same here

src/accelerate/inference.py
```diff
@@ -59,9 +61,26 @@ def build_pipeline(model, split_points, args, kwargs) -> PipelineStage:
...
+                break
+        for kwarg in kwargs.values():
+            found_batch_size = ignorant_find_batch_size(kwarg)
+            if found_batch_size is not None:
+                break
+        if found_batch_size is None:
+            raise ValueError("Could not find batch size from args or kwargs")
+        else:
+            if (found_batch_size % state.num_processes) != 0:
+                slice_to_cut = slice(0, found_batch_size % state.num_processes)
```
SunMarc, line 81: Still not sure about this one. If we have a `found_batch_size` of 4 with `state.num_processes` of 3, it will do `slice(0,1)` and we will only concatenate one extra input, so we will get in the end a input of batch_size 5 and this is still not % by 3.

tests/test_utils.py
```diff
@@ -237,3 +238,88 @@ def test_pad_across_processes(self):
         with self.assertWarns(CannotPadNestedTensorWarning):
             nt2 = pad_across_processes(nt)
         self.assertIs(nt, nt2)
+
+    def test_slice_and_concatenate(self):
+        # Should be equivalent to the slice func used in `pippy_forward`
```
SunMarc, line 243: Also it would be great to have a pippy example that was failing before this PR (2 gpus and a `batch_size` of 1 for example)

muellerzr: Done :)

src/accelerate/utils/operations.py
```diff
@@ -591,6 +608,46 @@ def _pad_across_processes(tensor, dim=0, pad_index=0, pad_first=False):
...
+        else:
+            to_pad = num_processes - (batch_size // num_processes)
+        # In the rare case that `to_pad` is negative,
+        # we need to pad the last inputs - the found `to_pad`
+        if last_inputs > to_pad & to_pad < 1:
+            to_pad = last_inputs - to_pad
+        old_size = tensor.shape
+        new_size = list(old_size)
+        new_size[0] = batch_size + to_pad
+        new_tensor = tensor.new_zeros(tuple(new_size))
```
muellerzr, line 636: It's okay to just pad `0`, we can drop these afterwards and the user won't know that padded inputs were event sent ideally

PR #2354, opened 2024-01-18
Fix dispatch_model with tied weights test on T4

fxmarty:
Running on T4 GPU now passes for me

```
CUDA_VISIBLE_DEVICES="0" pytest tests/test_big_modeling.py -k "test_dispatch" -s -vvvvv
```

Note: using torch 2.1.2 with CUDA 12+ and NOT running tests in parallel, maybe can still fail otherwise, let's see

### DIFF
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import copy
+import gc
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Nice! Thanks for the quick-fix

SunMarc (approved):
Thx for the fix !

PR #2366, opened 2024-01-22
Pin DeepSpeed until patch

muellerzr:
Yet another deepspeed release that broke things, this time though we can't install it on our main runners so we need to pin until a fix is released so PR CI can pass.

w.r.t: https://github.com/microsoft/DeepSpeed/issues/4984


pacman100 (approved):
Thank you @muellerzr for quick patch!

PR #2369, opened 2024-01-22
`add_hook_to_module` and `remove_hook_from_module` compatibility with fx.GraphModule

fxmarty:
It appears to be illegal to modify the forward method of a `GraphModuleImpl` instance, PyTorch rather favors modifying the class method (https://github.com/pytorch/pytorch/blob/c3780010a58a84920335296ee5f091a0db18259f/torch/fx/graph_module.py#L708-L723) though it is unclear why the behavior would be different.

Just modifying the class instance forward method results in a frozen forward call - despite modifying the graph (as in the test).

@Giuseppe5 and @nickfraser found this bug. cc @jamesr66a as this was quite a tricky one.

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -17,6 +17,7 @@ import unittest
 
 import torch
 import torch.nn as nn
+from torch.fx import symbolic_trace
 
 from accelerate.hooks import (
     AlignDevicesHook,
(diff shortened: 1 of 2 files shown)


tests/test_hooks.py
```diff
@@ -347,3 +348,48 @@ def test_attach_align_device_hook_as_cpu_offload_with_weight_map(self):
...
+            output_node = None
+            for node in graph_model.graph.nodes:
+                if node.name == "output":
+                    output_node = node
+            self.assertTrue(output_node is not None)
+
+            output_node.replace_input_with(linear2_node, new_node)
+
+            graph_model.graph.lint()
+            graph_model.recompile()
```
SunMarc, line 390: Could you describe a little bit what we are doing here ?

SunMarc (approved):
LGTM ! Just a nit.

tests/test_hooks.py
fxmarty: Added a comment, thank you!

muellerzr (commented):
Thanks!

PR #2383, opened 2024-01-28
Fix XPU inference

notsyncing:
# What does this PR do?

With `device_map = {"xpu": "16GB", "cpu": "32GB"}` on an Intel XPU, it will complain about 

```
Device xpu is not recognized, available devices are integers(for GPU/XPU), 'mps', 'cpu' and 'disk'"
```
but you cannot just put 0 as device, or it will treat 0 as CUDA device, then complains again that torch is not compiled with CUDA enabled.

You will need safetensors >= 0.4.2 if using safetensors files (https://github.com/huggingface/safetensors/pull/428).

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -14,6 +14,7 @@
 
 import contextlib
 import gc
+import importlib
 import inspect
 import json
 import logging
@@ -24,6 +25,7 @@ import tempfile
 from collections import OrderedDict, defaultdict
 from typing import Dict, List, Optional, Tuple, Union
 
+import packaging
 import torch
 import torch.nn as nn
 
@@ -33,6 +35,7 @@ from .dataclasses import AutocastKwargs, CustomDtype, DistributedType
 from .imports import is_mps_available, is_npu_available, is_peft_available, is_xpu_available
 from .offload import load_offloaded_weight, offload_weight, save_offload_index
 from .tqdm import is_tqdm_available, tqdm
+from .versions import compare_versions
 
 
 if is_npu_available(check_device=False):
@@ -367,6 +370,8 @@ def set_module_tensor_to_device(
         # `torch.Tensor.to(<int num>)` is not supported by `torch_npu` (see this [issue](https://github.com/Ascend/pytorch/issues/16)).
         if is_npu_available() and isinstance(device, int):
             device = f"npu:{device}"
+        if is_xpu_available() and isinstance(device, int):
+            device = f"xpu:{device}"
         if value is None:
             new_value = old_value.to(device)
             if dtype is not None and device in ["meta", torch.device("meta")]:
@@ -427,6 +432,8 @@ def set_module_tensor_to_device(
     # clean pre and post foward hook
     if is_npu_available():
         torch.npu.empty_cache()
+    elif is_xpu_available():
+        torch.xpu.empty_cache()
     else:
         torch.cuda.empty_cache()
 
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/operations.py
```diff
@@ -171,6 +171,9 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
             # `torch.Tensor.to("npu")` could not find context when called for the first time (see this [issue](https://gitee.com/ascend/pytorch/issues/I8KECW?from=project-issue)).
             elif device == torch.device("npu"):
                 device = "npu:0"
+        elif is_xpu_available():
+            if isinstance(device, int):
+                device = f"xpu:{device}"
```
SunMarc, line 176: maybe also add the case when `device == torch.device("xpu")` just like npu ?

src/accelerate/utils/modeling.py
```diff
@@ -1351,7 +1355,12 @@ def load_state_dict(checkpoint_file, device_map=None):
             else:
                 progress_bar = None
             for device in devices:
-                with safe_open(checkpoint_file, framework="pt", device=device) as f:
+                target_device = device
+
+                if is_xpu_available() and isinstance(device, int):
+                    target_device = f"xpu:{device}"
+
+                with safe_open(checkpoint_file, framework="pt", device=target_device) as f:
```
SunMarc, line 1363: Just to be sure, the model won't be loaded in the correct xpu device if we don't have the prefix "xpu:". Is that right ? cc @statelesshz since you added the support for npu and it might also be required.

SunMarc (approved):
Thanks for fixing the support of xpu for big model inference. Could you have a look @abhilash1910 since you did the initial integration with xpu ?

src/accelerate/utils/modeling.py
notsyncing: Yes, you can see it here: https://github.com/huggingface/safetensors/pull/428 file `bindings/python/src/lib.rs` line 263:

```rust
 name if name.starts_with("xpu:") => {
    let tokens: Vec<_> = name.split(':').collect();
    if tokens.len() == 2 {
        let device: usize = tokens[1].parse()?;
        Ok(Device::Xpu(device))
    } else {
        Err(SafetensorError::new_err(format!(
            "device {name} is invalid"
        )))
    }
}
```

If you supply only a number, it will treat as CUDA device (the same file, line 278):

```rust
} else if let Ok(number) = ob.extract::<usize>() {
            Ok(Device::Cuda(number))
}
```

src/accelerate/utils/operations.py
notsyncing: Added

src/accelerate/utils/modeling.py
abhilash1910: Yes I think the changes are correct . It requests a "xpu" device ; since the addition of safetensors.

abhilash1910 (approved):
LGTM ! Thanks @notsyncing for adding this; with the safetensors being added, this was something in the pipeline

src/accelerate/utils/modeling.py
SunMarc: Perfect ! Thanks for the explanation !

SunMarc (approved):
Thanks again @notsyncing and @abhilash1910 for reviewing. Before merging this, could you add a check to raise an error if the user don't have the right version of safetensors with xpu by using [`compare_version`](https://github.com/huggingface/accelerate/blob/7ba64e632c38426831614c270e66ea13e136499a/src/accelerate/utils/versions.py#L26) and `safentsors_version = version.parse(importlib.metadata.version("safentsors")) `

notsyncing:
@SunMarc I have added the check of safetensors version, and fixed style errors.

SunMarc (approved):
LGTM ! Thanks for iterating.

PR #2387, opened 2024-01-29
Prefer `is_torch_tensor` over `hasattr` for torch.compile.

PhilJd:
# What does this PR do?
`torch.compile` breaks when using `hasattr` but succeeds when using `isinstance(torch.Tensor)`.  This commit short-circuits the `hasattr` call for `torch.Tensor`s if possible.

Note: `is_npu_available` is also not torch.compile compatible due to (1) lru_cache and (2) importlib checks, so I've moved it into the try block, catching the AssertionError instead.

## Before submitting
      Pull Request section?
      to it if that's the case.
- [X] Did you write any new necessary tests?

## Who can review?

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -177,12 +177,14 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
             return tensor.to(device, non_blocking=non_blocking)
         except TypeError:  # .to() doesn't accept non_blocking as kwarg
             return tensor.to(device)
-        except AssertionError:
+        except AssertionError as error:
             # `torch.Tensor.to(<int num>)` is not supported by `torch_npu` (see this [issue](https://github.com/Ascend/pytorch/issues/16)).
             # This call is inside the try-block since is_npu_available is not supported by torch.compile.
             if is_npu_available():
                 if isinstance(device, int):
                     device = f"npu:{device}"
+            else:
+                raise error
         except Exception as error:
             if is_xpu_available():
                 if isinstance(device, int):
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -174,7 +174,7 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
         })
     elif is_torch_tensor(tensor) or hasattr(tensor, "to"):
         # `torch.Tensor.to("npu")` could not find context when called for the first time (see this [issue](https://gitee.com/ascend/pytorch/issues/I8KECW?from=project-issue)).
-        if device == torch.device("npu"):
+        if device == "npu":
             device = "npu:0"
         if device == "xpu":
             device = "xpu:0"
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """
 A set of basic tensor ops compatible with tpu, gpu, and multigpu
 """
@@ -34,10 +33,8 @@ from .imports import (
     is_xpu_available,
 )
 
-
 if is_tpu_available(check_device=False):
     import torch_xla.core.xla_model as xm
-
 
 if is_torch_distributed_available():
     from torch.distributed import ReduceOp
@@ -90,7 +87,12 @@ def honor_type(obj, generator):
         return type(obj)(generator)
 
 
-def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_other_type=False, **kwargs):
+def recursively_apply(func,
+                      data,
+                      *args,
+                      test_type=is_torch_tensor,
+                      error_on_other_type=False,
+                      **kwargs):
     """
     Recursively apply a function on a data structure that is a nested list/tuple/dictionary of a given base type.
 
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -237,3 +238,15 @@ def test_pad_across_processes(self):
...
+        compiled_send_to_device = torch.compile(operations.send_to_device, fullgraph=True)
+        compiled_send_to_device(torch.zeros([1], dtype=torch.bfloat16), "cpu")
+
+    def test_convert_to_fp32(self):
+        compiled_convert_to_fp32 = torch.compile(operations.convert_to_fp32, fullgraph=True)
+        compiled_convert_to_fp32(torch.zeros([1], dtype=torch.bfloat16))
+
+
+if __name__ == "__main__":
+    unittest.main()
```
muellerzr, line 252: We generally don't include these in the tests as we run everything via `pytest`

muellerzr (commented):
Thanks! Looks good to me, ready to merge bar one nit

PhilJd:
Thanks for the review! Sorry, unittest.main wasn't supposed to go in, I just added it for local testing - removed now.

PhilJd:
Friendly ping :)

muellerzr:
Thanks @PhilJd apologies for missing this. Can you resolve the conflicts and we can merge? Thanks!

PhilJd:
Updated; Note: I've put the xpu handling when xpu is available but only an int is provided to `to` under a general exception as I don't know what the XPU lib would raise.

tests/test_utils.py
```diff
@@ -32,6 +32,7 @@
     extract_model_from_parallel,
     find_device,
     listify,
+    operations,
```
muellerzr, line 35: Why are we importing the entire operations here? Just import `send_to_device` and `convert_to_fp32` please

muellerzr (approved):
Thanks for the fix! Left two comments to fix, otherwise we can merge :)

PhilJd:
I found only one comment? Sorry for the delay, commit to address it is now pushed.

tests/test_utils.py
```diff
@@ -39,6 +39,7 @@
     save,
     send_to_device,
 )
+from accelerate.utils.operations import convert_to_fp32
```
muellerzr, line 42: We're close. This needs to be moved into where `save` and `send_to_device` is imported from. E.g.

```
from accelerate.utils import (
    ...,
    ...,
    save,
    send_to_device,
    convert_to_fp32,
)
```
(And then run `make style; make quality` as the imports need to be in alphabetical order

PhilJd: Sorry, I missed that this is already retrieved in the init file, updated.

muellerzr (approved):
Thanks for your work on this!

src/accelerate/utils/operations.py
```diff
@@ -169,19 +167,28 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
...
+        # `torch.Tensor.to("npu")` could not find context when called for the first time (see this [issue](https://gitee.com/ascend/pytorch/issues/I8KECW?from=project-issue)).
+        if device == "npu":
+            device = "npu:0"
+        if device == "xpu":
+            device = "xpu:0"
+        try:
+            return tensor.to(device, non_blocking=non_blocking)
+        except TypeError:  # .to() doesn't accept non_blocking as kwarg
+            return tensor.to(device)
+        except AssertionError:
```
BenjaminBossan, line 180: Here, we implicitly make the assumption that `AssertionError` is always associated with NPU and all other errors with XPU, right? Otherwise, it looks like strange things would happen. Hopefully this is a robust assumption.

src/accelerate/utils/operations.py
```diff
@@ -169,19 +167,28 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
...
-                device = "npu:0"
-        elif is_xpu_available():
-            if isinstance(device, int):
-                device = f"xpu:{device}"
-            elif device == torch.device("xpu"):
-                device = "xpu:0"
+            # This call is inside the try-block since is_npu_available is not supported by torch.compile.
+            if is_npu_available():
+                if isinstance(device, int):
+                    device = f"npu:{device}"
```
BenjaminBossan, line 185: Should we add an `else` clause here, to explicitly deal with the case that `is_npu_available()` is `False`?

BenjaminBossan (commented):
Thanks for making these functions work with `torch.compile`.

I'm not super happy with the "contortions" we have to make in order to enable this, but I'm not sure if there is a better way.

src/accelerate/utils/operations.py
PhilJd: Unfortunately yes - when I sent the first version of this PR only the NPU check existed. I don't know where the XPU code lives that raises this error and I don't have a device to check, would you have a pointer?

src/accelerate/utils/operations.py
PhilJd: Done.

PhilJd (commented):
I've tried a few things before submitting this, another option would be to run the is_npu_available function during module import (i.e., just call it at the bottom) and cache its result, but I feel like this is also not what a user might expect.

I also totally understand if you'd rather forgo torch.compile support for readability - at least on my side I wouldn't be blocked as we've migrated away from accelerate/transformers due to issues with torch.compile :)

BenjaminBossan (approved):
Thanks for the adjustment, LGTM.

sheepymeh:
It seems like this issue still persists for me using the following with multiple GPUs:

```python
from transformers import AutoTokenizer, AutoModelForCausalLM, StaticCache
import torch
import os

os.environ["TOKENIZERS_PARALLELISM"] = "true"

tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-hf")
    "meta-llama/Llama-2-7b-hf",
    device_map='auto',
    torch_dtype=torch.float16
)

model._setup_cache(StaticCache, max_batch_size=1, max_cache_len=128)
model.compile(mode="reduce-overhead", fullgraph=True)
input_text = "A few facts about the universe: "
input_ids = tokenizer(input_text, return_tensors="pt").to("cuda").input_ids
model_outputs = model.generate(input_ids, max_new_tokens=100)
```

The error is:
```
from user code:
   File "/home/sheepymeh/transformers/venv/lib/python3.10/site-packages/accelerate/hooks.py", line 161, in new_forward
    args, kwargs = module._hf_hook.pre_forward(module, *args, **kwargs)
  File "/home/sheepymeh/transformers/venv/lib/python3.10/site-packages/accelerate/hooks.py", line 356, in pre_forward
    return send_to_device(args, self.execution_device), send_to_device(
  File "/home/sheepymeh/transformers/venv/lib/python3.10/site-packages/accelerate/utils/operations.py", line 149, in send_to_device
    if is_torch_tensor(tensor) or hasattr(tensor, "to"):
```

This issue does not occur when using only one GPU (doesn't seem to run the affected `send_to_device` function).

muellerzr:
@sheepymeh please open a new issue answering all the details it asks for in the bug report, including the *entire* stack trace please :)

PR #2394, opened 2024-01-30
Pin Torch version to <2.2.0

Rocketknight1:
See https://github.com/huggingface/transformers/pull/28785 - torch 2.2.0 was pushed to pypi and broke the CI. We pinned the version in Transformers, but because it installs accelerate from `main` at the end of the other installations, we need to pin the version in `accelerate` as well, or the `accelerate` installation upgrades to `torch 2.2.0` regardless.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
         ]
     },
     python_requires=">=3.8.0",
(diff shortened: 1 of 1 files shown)


Rocketknight1:
Merging this as an emergency measure so we can fix the transformers CI - it's a one-line change, and we can discuss/revert it later if it's an issue!

sayakpaul:
Why do we need torch to be less than 2.2.0?

Rocketknight1:
CI failures - we'll unpin it soon, I'm sure!

PR #2398, opened 2024-01-31
[RFC] Precisely exclude buffer size when inferring auto device map

notsyncing:
Hello, I'm trying using `accelerate` to offload a large model (https://huggingface.co/TheBloke/WizardCoder-33B-V1.1-GPTQ) to CPU, with following code (requires https://github.com/huggingface/accelerate/pull/2383 and https://github.com/huggingface/transformers/pull/28755):

```python
import datetime

import torch

import accelerate

# Without offload_buffers
pipe = pipeline("text-generation", model=model_id, model_kwargs={"torch_dtype": torch.bfloat16, "device_map": "auto", "max_memory": {0: "16GB", "cpu": "128GB"}})

# With offload_buffers (won't OOM)
#pipe = pipeline("text-generation", model=model_id, model_kwargs={"torch_dtype": torch.bfloat16, "device_map": "auto", "max_memory": {0: "16GB", "cpu": "128GB"}, "offload_buffers": True})

print(results)
``` 

After some digging, I found that this model `TheBloke/WizardCoder-33B-V1.1-GPTQ` contains a huge buffer (model is 17GB, but buffers are more than 16GB), so it will not fit into my GPU at all without `offload_buffers`.

So I created this PR, to calculate and exclude buffer size from device max memory before trying to put layers on GPU when not setting offload_buffers, and print a warning for such model with big buffers.
With this PR, the code above works well (fully on CPU without offload_buffers, and working with offload_buffers on GPU, ~97% VRAM usage).

It also contains another fix, but I'm not sure about this: the method in `utils/modeling.py`:

```python
def compute_module_sizes(
    # ... (other code) ...
            size = tensor.numel() * min(dtype_size, dtype_byte_size(tensor.dtype))
```

This will get wrong size (2) for that model `TheBloke/WizardCoder-33B-V1.1-GPTQ` , because it contains `int32` weights, which size should be 4, so I changed this `min` to `max`.
The `int32` weight won't be converted because of these lines in `utils/modeling.py`:

```python
  # ... (other code) ...
        elif not str(value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):  # This line will ignore int32 values
            value = value.to(dtype)
```

With this modification, the layer containing `int32` values can get correct size.

# What does this PR do?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -567,7 +567,8 @@ def load_checkpoint_and_dispatch(
                 low_zero=(device_map == "balanced_low_0"),
             )
         device_map = infer_auto_device_map(
-            model, max_memory=max_memory, no_split_module_classes=no_split_module_classes, dtype=dtype
+            model, max_memory=max_memory, no_split_module_classes=no_split_module_classes, dtype=dtype, 
+            exclude_buffers=not offload_buffers
         )
     if offload_state_dict is None and device_map is not None and "disk" in device_map.values():
         offload_state_dict = True
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -697,12 +697,39 @@ def compute_module_sizes(
         elif dtype is None:
             size = tensor.numel() * dtype_byte_size(tensor.dtype)
         else:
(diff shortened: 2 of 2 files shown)


notsyncing:
This is wrong for multiple card, will submit another PR later. Closing this.

PR #1046, opened 2023-02-07
Update quality tools to 2023

sgugger:
This PR updates the quality tools the same way as in Transformers by:
- updating black to the 23 version
- updating black target to python 3.7
- removing flake8 and isort to use ruff instead
- removing third parties entirely

PR #2400, opened 2024-01-31
Use Ruff for formatting too

akx:
# What does this PR do?

This PR replaces Black formatting with `ruff format`, following in the footsteps of https://github.com/huggingface/transformers/pull/27144 and https://github.com/huggingface/diffusers/pull/5841.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

Tagging @sgugger as the author of https://github.com/huggingface/accelerate/pull/1046 that last touched Ruff configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: quality style test docs utils
 
-check_dirs := tests src examples benchmarks utils
+check_dirs := .
 
 # Check that source code meets quality standards
 
@@ -12,14 +12,14 @@ extra_quality_checks:
 
 # this target runs checks on all files
 quality:
-	black --required-version 23 --check $(check_dirs)
 	ruff $(check_dirs)
+	ruff format --check $(check_dirs)
 	doc-builder style src/accelerate docs/source --max_len 119 --check_only
 
 # Format source code automatically and check is there are any problems left that need manual fixing
 style:
-	black --required-version 23 $(check_dirs)
 	ruff $(check_dirs) --fix
+	ruff format $(check_dirs)
 	doc-builder style src/accelerate docs/source --max_len 119
 	
 # Run tests for the library
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,3 @@
-[tool.black]
-line-length = 119
-target-version = ['py37']
-
 [tool.ruff]
 # Never enforce `E501` (line length violations).
 ignore = ["E501", "E741", "W605"]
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: quality style test docs utils
 
-check_dirs := tests src examples benchmarks utils
```
muellerzr, line 3: This should remain unchanged

Makefile
```diff
@@ -12,14 +12,14 @@ extra_quality_checks:
 
 # this target runs checks on all files
 quality:
-	black --required-version 23 --check $(check_dirs)
 	ruff $(check_dirs)
+	ruff format --check $(check_dirs)
```
muellerzr, line 16: If it's a truly 1:1 then this should come first, what's the reason it's after? :)

pyproject.toml
```diff
@@ -11,7 +7,25 @@ line-length = 119
...
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["E402", "F401", "F403", "F811"]
+"manim_animations/*" = ["ALL"]
 
 [tool.ruff.isort]
 lines-after-imports = 2
 known-first-party = ["accelerate"]
+
+[tool.ruff.format]
+# Like Black, use double quotes for strings.
```
muellerzr, line 17: Let's just have this all under "Configurations similar to black"

muellerzr (commented):
Thanks! Overall this looks good to me, just some questions in regards to ordering

pyproject.toml
akx: In that case, I'll just preferably remove the explicit configuration that's repeating the defaults; Ruff is by default Black-compatible. I borrowed this verbiage from https://github.com/huggingface/transformers/pull/27144.

Makefile
akx: For instance the Ruff `COM` rules (which are not enabled at present, but could be) may add trailing commas, and since Ruff format supports the "magic trailing comma" convention Black uses (i.e. if e.g. an arglist has a trailing comma, the whole shebang should be split into line-per-argument), that formatting needs to happen after a trailing comma is added.

Makefile
akx: Why? Not having this here makes it easier to run Ruff on the command line (for instance to enable more lint rules) without having to remember which directories to include or exclude.

muellerzr: Makes sense!

muellerzr (approved):
Thanks for doing this!

PR #2403, opened 2024-01-31
Fix fp8 things

muellerzr:
Fixes a few things with the fp8 integration since the merger of msamp:

* Couldn't just use the accelerate configuration and leave the defaults, as the mixed precision wasn't set yet for when we check to build a recipe handler default
* explicitly makes the backend te as the default (hit the dataclass not liking this w/o it being passed in)
* some fixes to allow users to pass in either `backend="te"` or `backend="TE"`


SunMarc (approved):
Thx for fixing this !

PR #2408, opened 2024-02-01
Fix CI due to pytest

muellerzr:
Pytest released a new version that broke our CI because:
>  Files and directories are now collected in alphabetical order jointly, unless changed by a plugin. Previously, files were collected before directories. See below for an example.

The failures were coming from deepspeed being the first one being ran instead of later on. This PR change `make test` to run in the following order:

```make
test:
	$(MAKE) test_core
	$(MAKE) test_cli
	$(MAKE) test_big_modeling
	$(MAKE) test_deepspeed
	$(MAKE) test_fsdp
```

Instead of the prior:
```make
python -m pytest -s -v ./tests/ --ignore=./tests/test_examples.py $(if $(IS_GITHUB_CI),--report-log "$(PYTORCH_VERSION)_all.log",)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,6 @@ style:
 	doc-builder style src/accelerate docs/source --max_len 119
 	
 # Run tests for the library
-test:
-	python -m pytest -s -v ./tests/ --ignore=./tests/test_examples.py $(if $(IS_GITHUB_CI),--report-log "$(PYTORCH_VERSION)_all.log",)
-
 test_big_modeling:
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
It's great that we manage to fix the issue without having to pin pytest ! LGMT !

akx:
Hold on, why is the order of tests significant for the test suite to pass? 😳

SunMarc:
Yeah, this is definitely strange, but the deepspeed tests seems to mess up with our env. We will investigate this further when we get the time !

PR #2411, opened 2024-02-02
Fix the size of int and bool type when computing module size

notsyncing:
# What does this PR do?

Hello, I'm trying using accelerate to offload a large model (https://huggingface.co/TheBloke/WizardCoder-33B-V1.1-GPTQ) to CPU, with following code (requires https://github.com/huggingface/accelerate/pull/2383 if using Intel GPU, and https://github.com/huggingface/transformers/pull/28755):

```python
import datetime

import torch

import accelerate

pipe = pipeline("text-generation", model=model_id, model_kwargs={"torch_dtype": torch.bfloat16, "device_map": "auto", "max_memory": {0: "16GB", "cpu": "128GB"}, "offload_buffers": True})

print(results)
```

After some digging, I found that in the utils/modeling.py:

```python
def compute_module_sizes(
    # ... (other code) ...
            size = tensor.numel() * min(dtype_size, dtype_byte_size(tensor.dtype))  # here, dtype_size = 2, dtype_byte_size = 4
```

This will get wrong size 2 for that model `TheBloke/WizardCoder-33B-V1.1-GPTQ`, because it contains `int32` weights, which size should be 4, so the layer size has been under-estimated, causing an OOM on real load.

The int32 weight won't be converted because of these lines in utils/modeling.py:

```python
  # ... (other code) ...
        elif not str(value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):  # This line will ignore int32 values
            value = value.to(dtype)
```

So I added a check in `compute_module_sizes`, matching the `set_module_tensor_to_device`.

With this PR, the layer containing int32 values can get correct size, and my code can finish correctly without OOM.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -696,6 +696,10 @@ def compute_module_sizes(
             size = tensor.numel() * special_dtypes_size[name]
         elif dtype is None:
             size = tensor.numel() * dtype_byte_size(tensor.dtype)
+        elif str(tensor.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
+            # According to the code in set_module_tensor_to_device, these types won't be converted
+            # so use their original size here
+            size = tensor.numel() * dtype_byte_size(tensor.dtype)
         else:
             size = tensor.numel() * min(dtype_size, dtype_byte_size(tensor.dtype))
         name_parts = name.split(".")


SunMarc (approved):
Great catch ! Thanks for fixing this !

PR #2412, opened 2024-02-02
Check if the buffers fit GPU memory after device map auto inferred

notsyncing:
# What does this PR do?

Hello, I'm trying using `accelerate` to offload a large model (https://huggingface.co/TheBloke/WizardCoder-33B-V1.1-GPTQ) to CPU, with following code (requires https://github.com/huggingface/accelerate/pull/2383 if using Intel GPU, and https://github.com/huggingface/transformers/pull/28755):

```python
import datetime

import torch

import accelerate

# Without offload_buffers (OOM)
pipe = pipeline("text-generation", model=model_id, model_kwargs={"torch_dtype": torch.bfloat16, "device_map": "auto", "max_memory": {0: "16GB", "cpu": "128GB"}})

# With offload_buffers (won't OOM)
#pipe = pipeline("text-generation", model=model_id, model_kwargs={"torch_dtype": torch.bfloat16, "device_map": "auto", "max_memory": {0: "16GB", "cpu": "128GB"}, "offload_buffers": True})

print(results)
``` 

After some digging, I found that this model `TheBloke/WizardCoder-33B-V1.1-GPTQ` contains a huge buffer (model is 17GB, but buffers are more than 16GB), so it will not fit into my GPU at all without `offload_buffers`.

So I created this PR, to check if the buffer can fit on the GPU, and raise an exception for such case.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -22,6 +22,7 @@ import os
 import re
 import shutil
 import tempfile
+import warnings
 from collections import OrderedDict, defaultdict
 from typing import Dict, List, Optional, Tuple, Union
 
@@ -690,6 +691,7 @@ def compute_module_sizes(
     model: nn.Module,
     dtype: Optional[Union[str, torch.device]] = None,
     special_dtypes: Optional[Dict[str, Union[str, torch.device]]] = None,
+    buffers_only: bool = False,
 ):
     """
     Compute the size of each submodule of a given model.
@@ -701,7 +703,15 @@ def compute_module_sizes(
         special_dtypes = {key: _get_proper_dtype(dtyp) for key, dtyp in special_dtypes.items()}
         special_dtypes_size = {key: dtype_byte_size(dtyp) for key, dtyp in special_dtypes.items()}
     module_sizes = defaultdict(int)
-    for name, tensor in named_module_tensors(model, recurse=True):
+
+    module_list = []
+
+    if not buffers_only:
+        module_list = named_module_tensors(model, recurse=True)
+    else:
+        module_list = model.named_buffers(recurse=True)
+
+    for name, tensor in module_list:
         if special_dtypes is not None and name in special_dtypes:
             size = tensor.numel() * special_dtypes_size[name]
         elif dtype is None:
@@ -717,6 +727,18 @@ def compute_module_sizes(
             module_sizes[".".join(name_parts[:idx])] += size
 
     return module_sizes
+
+
+def compute_module_total_buffer_size(
+    model: nn.Module,
+    dtype: Optional[Union[str, torch.device]] = None,
+    special_dtypes: Optional[Dict[str, Union[str, torch.device]]] = None,
+):
+    """
+    Compute the total size of buffers in each submodule of a given model.
+    """
+    module_sizes = compute_module_sizes(model, dtype=dtype, special_dtypes=special_dtypes, buffers_only=True)
+    return module_sizes.get("", 0)
 
 
 def get_max_layer_size(
@@ -1030,6 +1052,7 @@ def infer_auto_device_map(
     special_dtypes: Optional[Dict[str, Union[str, torch.dtype]]] = None,
     verbose: bool = False,
     clean_result: bool = True,
+    offload_buffers: bool = False,
 ):
     """
     Compute a device map for a given model giving priority to GPUs, then offload on CPU and finally offload to disk,
@@ -1066,6 +1089,9 @@ def infer_auto_device_map(
             Whether or not to provide debugging statements as the function builds the device_map.
         clean_result (`bool`, *optional*, defaults to `True`):
             Clean the resulting device_map by grouping all submodules that go on the same device together.
+        offload_buffers (`bool`, *optional*, defaults to `False`):
+            In the layers that are offloaded on the CPU or the hard drive, whether or not to offload the buffers as
+            well as the parameters.
     """
     # Get default / clean up max_memory
     max_memory = get_max_memory(max_memory)
@@ -1098,6 +1124,9 @@ def infer_auto_device_map(
     device_map = OrderedDict()
     current_device = 0
     current_memory_used = 0
+    device_memory_used = {}
+    device_memory_reserved = {}
+    device_buffer_sizes = {}
 
     # Direct submodules and parameters
     modules_to_treat = (
@@ -1150,6 +1179,7 @@ def infer_auto_device_map(
         # Reduce max size available by the largest layer.
         if devices[current_device] in main_devices:
             current_max_size = current_max_size - max_layer_size
+            device_memory_reserved[current_device] = max_layer_size
         # Case 1 -> We're too big!
         if current_max_size is not None and current_memory_used + module_size > current_max_size:
             # Split or not split?
@@ -1167,6 +1197,8 @@ def infer_auto_device_map(
                 # -> no split, we go to the next device
                 if verbose:
                     print("This module cannot be split, going to the next device.")
+
+                device_memory_used[device] = current_memory_used
                 current_device += 1
                 modules_to_treat = [(name, module)] + modules_to_treat
                 current_memory_used = 0
@@ -1218,6 +1250,12 @@ def infer_auto_device_map(
                         modules_to_treat.pop(tied_module_index)
                     device_map[tied_module_name] = devices[current_device]
 
+                if not offload_buffers and isinstance(module, nn.Module):
+                    current_buffer_size = compute_module_total_buffer_size(
+                        module, dtype=dtype, special_dtypes=special_dtypes
+                    )
+                    device_buffer_sizes[device] = device_buffer_sizes.get(device, 0) + current_buffer_size
+
             else:
                 # We don't fit with the tied modules. Next question is: can we split one of the tied modules to make it
                 # smaller or do we need to go on the next device?
@@ -1258,6 +1296,8 @@ def infer_auto_device_map(
                     # If the tied module is not split, we go to the next device
                     if verbose:
                         print("None of the tied module can be split, going to the next device.")
+
+                    device_memory_used[device] = current_memory_used
                     current_device += 1
                     modules_to_treat = [(name, module)] + modules_to_treat
                     current_memory_used = 0
@@ -1272,10 +1312,39 @@ def infer_auto_device_map(
                         f"(available={current_max_size - current_memory_used})."
                     )
             current_memory_used += module_size
+            device_memory_used[device] = current_memory_used
             device_map[name] = devices[current_device]
+
+            if not offload_buffers and isinstance(module, nn.Module):
+                current_buffer_size = compute_module_total_buffer_size(
+                    module, dtype=dtype, special_dtypes=special_dtypes
+                )
+                device_buffer_sizes[device] = device_buffer_sizes.get(device, 0) + current_buffer_size
 
     if clean_result:
         device_map = clean_device_map(device_map)
+
+    if not offload_buffers:
+        non_gpu_buffer_size = device_buffer_sizes.get("cpu", 0) + device_buffer_sizes.get("disk", 0)
+
+        is_buffer_fit_any_gpu = False
+        for gpu_device, gpu_max_memory in max_memory.items():
+            if gpu_device == "cpu" or gpu_device == "disk":
+                continue
+
+            if not is_buffer_fit_any_gpu:
+                gpu_memory_used = device_memory_used.get(gpu_device, 0) + device_memory_reserved.get(gpu_device, 0)
+
+                if gpu_max_memory >= non_gpu_buffer_size + gpu_memory_used:
+                    is_buffer_fit_any_gpu = True
+
+        if len(gpus) > 0 and not is_buffer_fit_any_gpu:
+            warnings.warn(
+                f"Current model requires {non_gpu_buffer_size} bytes of buffer for offloaded layers, which seems does "
+                f"not fit any GPU's remaining memory. If you are experiencing a OOM later, please consider using "
+                f"offload_buffers=True."
+            )
+
     return device_map
 
 
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1125,7 +1125,6 @@ def infer_auto_device_map(
     current_device = 0
     current_memory_used = 0
     device_memory_used = {}
-    device_memory_reserved = {}
     device_buffer_sizes = {}
 
     # Direct submodules and parameters
@@ -1176,10 +1175,11 @@ def infer_auto_device_map(
 
         device = devices[current_device]
         current_max_size = max_memory[device] if device != "disk" else None
+        current_memory_reserved = 0
         # Reduce max size available by the largest layer.
         if devices[current_device] in main_devices:
             current_max_size = current_max_size - max_layer_size
-            device_memory_reserved[current_device] = max_layer_size
+            current_memory_reserved = max_layer_size
         # Case 1 -> We're too big!
         if current_max_size is not None and current_memory_used + module_size > current_max_size:
             # Split or not split?
@@ -1198,7 +1198,7 @@ def infer_auto_device_map(
                 if verbose:
                     print("This module cannot be split, going to the next device.")
 
-                device_memory_used[device] = current_memory_used
+                device_memory_used[device] = current_memory_used + current_memory_reserved
                 current_device += 1
                 modules_to_treat = [(name, module)] + modules_to_treat
                 current_memory_used = 0
@@ -1297,7 +1297,7 @@ def infer_auto_device_map(
                     if verbose:
                         print("None of the tied module can be split, going to the next device.")
 
-                    device_memory_used[device] = current_memory_used
+                    device_memory_used[device] = current_memory_used + current_memory_reserved
                     current_device += 1
                     modules_to_treat = [(name, module)] + modules_to_treat
                     current_memory_used = 0
@@ -1312,7 +1312,7 @@ def infer_auto_device_map(
                         f"(available={current_max_size - current_memory_used})."
                     )
             current_memory_used += module_size
-            device_memory_used[device] = current_memory_used
+            device_memory_used[device] = current_memory_used + current_memory_reserved
             device_map[name] = devices[current_device]
 
             if not offload_buffers and isinstance(module, nn.Module):
@@ -1324,16 +1324,15 @@ def infer_auto_device_map(
     if clean_result:
         device_map = clean_device_map(device_map)
 
-    if not offload_buffers:
-        non_gpu_buffer_size = device_buffer_sizes.get("cpu", 0) + device_buffer_sizes.get("disk", 0)
-
+    non_gpu_buffer_size = device_buffer_sizes.get("cpu", 0) + device_buffer_sizes.get("disk", 0)
+    if non_gpu_buffer_size > 0 and not offload_buffers:
         is_buffer_fit_any_gpu = False
         for gpu_device, gpu_max_memory in max_memory.items():
             if gpu_device == "cpu" or gpu_device == "disk":
                 continue
 
             if not is_buffer_fit_any_gpu:
-                gpu_memory_used = device_memory_used.get(gpu_device, 0) + device_memory_reserved.get(gpu_device, 0)
+                gpu_memory_used = device_memory_used.get(gpu_device, 0)
 
                 if gpu_max_memory >= non_gpu_buffer_size + gpu_memory_used:
                     is_buffer_fit_any_gpu = True
--- a/tests/test_modeling_utils.py
+++ b/tests/test_modeling_utils.py
@@ -635,9 +635,10 @@ class ModelingUtilsTester(unittest.TestCase):
         # Only linear1 (144) fits on device 0, and remaining buffers (batchnorm's 160 + linear2's 120 = 280) won't fit
         # device 0, but with offload_buffers they won't be loaded to device 0 all at once, so it's ok now
         # Should NOT print a warning in such case
-        with warnings.catch_warnings():
+        with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
             device_map = infer_auto_device_map(model, max_memory={0: 400, "cpu": "1GB"}, offload_buffers=True)
+        assert len(w) == 0
         assert device_map == {"linear1": 0, "batchnorm": "cpu", "linear2": "cpu"}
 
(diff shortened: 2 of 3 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -1267,8 +1308,33 @@ def infer_auto_device_map(
...
+            if gpu_device == "cpu" or gpu_device == "disk":
+                continue
+
+            gpu_memory_used = device_memory_used.get(gpu_device, 0)
+            if gpu_max_memory < non_gpu_buffer_size + gpu_memory_used:
+                raise ValueError(
+                    f"Device {gpu_device} has {gpu_max_memory} bytes of memory, and has planned to use {gpu_memory_used} bytes, "
+                    f"so the remaining memory is not enough to store additional buffers ({non_gpu_buffer_size} bytes)! "
+                    f"Please consider using offload_buffers=True."
+                )
```
SunMarc, line 1336: This will trigger an error even when we have, for example, the first gpu being free but the second gpu is full. The buffers can potentially go on the first gpu. When we don't offload the buffers, they are on the execution device (can be any device). Maybe we can just put a warning to let the user know that the buffer that won't be offloaded and it won't probably fit on one of the gpus.

SunMarc (commented):
Thanks for digging into the buffer issue @notsyncing ! Could you add some tests to see that the buffer calculation is right and that we raise the error when the buffers don't fit the remaining space on the gpu. Can you have a second look @muellerzr ?

src/accelerate/utils/modeling.py
```diff
@@ -698,7 +699,16 @@ def compute_module_sizes(
...
         special_dtypes_size = {key: dtype_byte_size(dtyp) for key, dtyp in special_dtypes.items()}
     module_sizes = defaultdict(int)
-    for name, tensor in named_module_tensors(model, recurse=True):
+
+    module_list = []
+    
+    if not buffers_only:
+        module_list = named_module_tensors(model, recurse=True) 
+    else:
+        if hasattr(model, "named_buffers"):
```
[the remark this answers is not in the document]
notsyncing: But there is a test that won't not pass without this. I'm not sure if I should touch the code of that test.

src/accelerate/utils/modeling.py
notsyncing: Yes, I did not consider the multi-gpu scenario because I have only one card. I'll change this to a warning instead of an `Error`. I will also try to determine the execution device more precisely to see if we can make this warning be more useful.

notsyncing: btw, could you please give some hint on the code which implemented the "the first gpu being free but the second gpu is full. The buffers can potentially go on the first gpu"? The only thing I found is the `AlignDevicesHook` class in `src/accelerate/hooks.py`, which seems moving all buffers to all execution devices.

SunMarc: Basically, what I mean is if the user let a `device_map` that only most of the modules on the second gpus and leave of the space available on the first gpu, the buffer will most likely fit in the first gpu (if we suppose the first gpu is the execution device). For example, this is the case with `device_map="balanced_low_0"` strategy.

src/accelerate/utils/modeling.py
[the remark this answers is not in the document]
SunMarc: Could you link the test ?

SunMarc:
Hi @notsyncing, thanks for your work ! I've replied to your questions. This is not pressing but are you planning to finish this PR ? Otherwise, I can finish it and add you as co-author =)

notsyncing:
> Hi @notsyncing, thanks for your work ! I've replied to your questions. This is not pressing but are you planning to finish this PR ? Otherwise, I can finish it and add you as co-author =)

Yes, I'm planning to finish it. But for the multi-gpu scenario, I cannot test it because I have only one gpu. Would you mind helping testing this? Thanks!

muellerzr:
We can run the relevant tests for you when you are ready/the normal CI here is green and report back any failures, yes :)

src/accelerate/utils/modeling.py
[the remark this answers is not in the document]
notsyncing:
```
FAILED tests/test_modeling_utils.py::ModelingUtilsTester::test_infer_auto_device_map_with_tied_weights - AttributeError: 'Parameter' object has no attribute 'named_buffers'
``` 

detail log:

```
_________________________ ModelingUtilsTester.test_infer_auto_device_map_with_tied_weights _________________________

self = <test_modeling_utils.ModelingUtilsTester testMethod=test_infer_auto_device_map_with_tied_weights>

    def test_infer_auto_device_map_with_tied_weights(self):
        model = nn.Sequential(
            OrderedDict([("layer1", ModelForTest()), ("layer2", ModelForTest()), ("layer3", ModelForTest())])
        )
        model.layer3.linear2.weight = model.layer1.linear2.weight
        device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 500})
        expected = {"layer1": 0, "layer3.linear2": 0, "layer2": 1, "layer3.linear1": 1, "layer3.batchnorm": 1}
        assert device_map == expected
    
        # With three weights tied together
        model.layer2.linear2.weight = model.layer1.linear2.weight
        device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 500})
        expected = {
            "layer1": 0,
            "layer2.linear2": 0,
            "layer3.linear2": 0,
            "layer2.linear1": 1,
            "layer2.batchnorm": 1,
            "layer3.linear1": 1,
            "layer3.batchnorm": 1,
        }
        assert device_map == expected
    
        # With two groups of weights tied together
        model.layer2.linear1.weight = model.layer1.linear1.weight
        device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 500})
        expected = {
            "layer1": 0,
            "layer2.linear1": 0,
            "layer2.linear2": 0,
            "layer3.linear2": 0,
            "layer2.batchnorm": 1,
            "layer3.linear1": 1,
            "layer3.batchnorm": 1,
        }
        assert device_map == expected
    
        # With weights ties in the same module
        model = nn.Sequential(
            OrderedDict(
                [
                    ("linear1", nn.Linear(4, 4)),
                    ("linear2", nn.Linear(6, 6)),
                    ("linear3", nn.Linear(4, 4)),
                    ("linear4", nn.Linear(6, 6)),
                ]
            )
        )
        model.linear3.weight = model.linear1.weight
        model.linear3.bias = model.linear1.bias
        device_map = infer_auto_device_map(model, max_memory={0: 250, 1: 400})
        expected = {"linear1": 0, "linear2": 1, "linear3": 0, "linear4": 1}
        assert device_map == expected
    
        # With tied weights sharing a same prefix name (`compute.weight` vs `compute.weight_submodule.parameter`)
        class SubModule(torch.nn.Module):
            def __init__(self, ref_to_parameter):
                super().__init__()
                self.parameter = ref_to_parameter
    
            def forward(self, x):
                return self.x + torch.max(self.parameter)
    
        class LinearModuleAndSubModule(torch.nn.Linear):
            def __init__(self, in_features, out_features):
                super().__init__(in_features, out_features)
                self.weight_submodule = SubModule(self.weight)
    
            def forward(self, x):
                return torch.nn.functional.linear(self.weight_submodule(x), self.weight)
    
        class Model(torch.nn.Module):
            def __init__(self):
                super().__init__()
                self.compute = LinearModuleAndSubModule(3, 8)
    
            def forward(self, x):
                return self.compute(x)
    
        model = Model()
    
        device_memory = {0: 4, "cpu": 96000}  # Low memory device, just to force splitting and trigger the error
>       infer_auto_device_map(model, device_memory)

tests/test_modeling_utils.py:581: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/accelerate/utils/modeling.py:1251: in infer_auto_device_map
    current_buffer_size = compute_module_total_buffer_size(
src/accelerate/utils/modeling.py:740: in compute_module_total_buffer_size
    module_sizes = compute_module_sizes(model, dtype, special_dtypes, True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

model = Parameter containing:
tensor([[-0.1732, -0.2755,  0.5513],
        [ 0.3202, -0.3520, -0.2377],
        [ 0.1762,  0.4...0662, -0.2712, -0.3916],
        [-0.3685,  0.5621,  0.1996],
        [ 0.2687,  0.0723, -0.5561]], requires_grad=True)
dtype = None, special_dtypes = None, buffers_only = True

    def compute_module_sizes(
        model: nn.Module,
        dtype: Optional[Union[str, torch.device]] = None,
        special_dtypes: Optional[Dict[str, Union[str, torch.device]]] = None,
        buffers_only: bool = False,
    ):
        """
        Compute the size of each submodule of a given model.
        """
        if dtype is not None:
            dtype = _get_proper_dtype(dtype)
            dtype_size = dtype_byte_size(dtype)
        if special_dtypes is not None:
            special_dtypes = {key: _get_proper_dtype(dtyp) for key, dtyp in special_dtypes.items()}
            special_dtypes_size = {key: dtype_byte_size(dtyp) for key, dtyp in special_dtypes.items()}
        module_sizes = defaultdict(int)
    
        module_list = []
    
        if not buffers_only:
            module_list = named_module_tensors(model, recurse=True)
        else:
            #if hasattr(model, "named_buffers"):
>           module_list = model.named_buffers(recurse=True)
E           AttributeError: 'Parameter' object has no attribute 'named_buffers'

src/accelerate/utils/modeling.py:712: AttributeError
```

src/accelerate/utils/modeling.py
```diff
@@ -1211,6 +1241,15 @@ def infer_auto_device_map(
...
 
+                if not offload_buffers:
+                    if device not in device_buffer_sizes.keys():
+                        device_buffer_sizes[device] = 0
+
+                    current_buffer_size = compute_module_total_buffer_size(
+                        module, dtype=dtype, special_dtypes=special_dtypes
+                    )
+                    device_buffer_sizes[device] = device_buffer_sizes[device] + current_buffer_size
+
```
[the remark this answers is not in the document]
notsyncing: Fixed.

src/accelerate/utils/modeling.py
```diff
@@ -1211,6 +1241,15 @@ def infer_auto_device_map(
                         modules_to_treat.pop(tied_module_index)
                     device_map[tied_module_name] = devices[current_device]
 
+                if not offload_buffers:
+                    if device not in device_buffer_sizes.keys():
+                        device_buffer_sizes[device] = 0
```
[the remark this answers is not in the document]
notsyncing: Fixed

src/accelerate/utils/modeling.py
```diff
@@ -716,6 +726,18 @@ def compute_module_sizes(
...
 
+def compute_module_total_buffer_size(
+    model: nn.Module,
+    dtype: Optional[Union[str, torch.device]] = None,
+    special_dtypes: Optional[Dict[str, Union[str, torch.device]]] = None,
+):
+    """
+    Compute the total size of buffers in each submodule of a given model.
+    """
+    module_sizes = compute_module_sizes(model, dtype, special_dtypes, True)
```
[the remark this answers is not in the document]
notsyncing: Fixed

src/accelerate/utils/modeling.py
```diff
@@ -1267,8 +1308,33 @@ def infer_auto_device_map(
             current_memory_used += module_size
             device_map[name] = devices[current_device]
 
+            if not offload_buffers:
+                if device not in device_buffer_sizes.keys():
+                    device_buffer_sizes[device] = 0
```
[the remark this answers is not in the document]
notsyncing: Fixed

src/accelerate/utils/modeling.py
```diff
@@ -1267,8 +1308,33 @@ def infer_auto_device_map(
...
+
+                current_buffer_size = compute_module_total_buffer_size(
+                    module, dtype=dtype, special_dtypes=special_dtypes
+                )
+                device_buffer_sizes[device] = device_buffer_sizes[device] + current_buffer_size
+
```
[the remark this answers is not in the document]
notsyncing: Fixed

notsyncing:
> Basically, what I mean is if the user let a `device_map` that only most of the modules on the second gpus and leave of the space available on the first gpu, the buffer will most likely fit in the first gpu (if we suppose the first gpu is the execution device). For example, this is the case with `device_map="balanced_low_0"` strategy.

Now I added a check to see if the buffer can fit any GPU, and only print the warning when every GPU does not fit. Will this work in your scenario?

notsyncing:
> Could you add some tests to see that the buffer calculation is right and that we raise the error when the buffers don't fit the remaining space on the gpu. 

As for the tests, it seems difficult to write a test for this if it only prints a warning. I propose that we could use a environment variable (like `ACCELERATE_BUFFER_SIZE_CHECK = "raise"`) to make it raises an exception instead:

```python
if os.getenv("ACCELERATE_BUFFER_SIZE_CHECK") == "raise":
    raise ValueError("xxx")
else:
    logger.warn("yyy")
```

Would this be better?

muellerzr:
You can use `warnings.warning()` to do so, and unittest has a method to check for a warning: `with self.assertWarns(Warning):`

muellerzr:
Also there's no need to force push, we squash the commit history when merging

SunMarc:
> Now I added a check to see if the buffer can fit any GPU, and only print the warning when every GPU does not fit. Will this work in your scenario?

Thanks ! Let's do that first and see the feedback. We can always change this behavior if needed. Most probably, the user will receive this warning if all his gpus are full and some of the layers are offloaded to cpu/disk. Hence, the buffer won't fit any of the gpus.

src/accelerate/utils/modeling.py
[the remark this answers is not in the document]
SunMarc: I found the issue. This happens because you don't make sure that the `module` that you are passing in `infer_auto_device_map` is indeed a `module`. See comment below ! If the changes, it should be safe to remove the condition.

src/accelerate/utils/modeling.py
```diff
@@ -1214,6 +1244,12 @@ def infer_auto_device_map(
                         modules_to_treat.pop(tied_module_index)
                     device_map[tied_module_name] = devices[current_device]
 
+                if not offload_buffers:
```
SunMarc, line 1247: We need to add an extra condition to make sure that we are indeed passing a `nn.Module` in `compute_module_total_buffer_size` since `module` can also be a `nn.Parameter`. This is why the test was failing.

src/accelerate/utils/modeling.py
```diff
@@ -1270,8 +1308,34 @@ def infer_auto_device_map(
             current_memory_used += module_size
             device_map[name] = devices[current_device]
 
+            if not offload_buffers:
```
SunMarc, line 1311: same here !

SunMarc (commented):
Thanks for the iteration. I left a few comments. Could you also add tests that would trigger the warning for a single and multi gpu setup ? We can test the multi gpu test if needed !

src/accelerate/utils/modeling.py
notsyncing: fixed

src/accelerate/utils/modeling.py
notsyncing: fixed

notsyncing:
@SunMarc I have added two test cases in `test_modeling_utils.py`, and fixed a problem that my previous code did not consider the reserved memory for the largest layer. Looks like we don't need an actual multi-gpu setup to test this.

tests/test_modeling_utils.py
```diff
@@ -604,6 +618,59 @@ def test_infer_auto_device_map_on_t0pp(self):
...
+        # model has size 336(parameters) + 440(buffers): linear1 64 + 80, batchnorm 72 + 160, linear2 100 + 120,
+        # linear3 100 + 80
+
+        # Now we have two devices, linear1 will fit on device 0, batchnorm will fit on device 1, and the second device
+        # can hold all remaining buffers
+        # Should NOT print a warning in such case
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 400, "cpu": "1GB"})
+        assert device_map == {"linear1": 0, "batchnorm": 1, "linear2": "cpu", "linear3": "cpu"}
```
SunMarc, line 659: We need to use simplefilter("always") and we can test if we catched any warning 
```suggestion
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 400, "cpu": "1GB"})
        assert len(w) == 0
        assert device_map == {"linear1": 0, "batchnorm": 1, "linear2": "cpu", "linear3": "cpu"}
```

tests/test_modeling_utils.py
```diff
@@ -604,6 +618,59 @@ def test_infer_auto_device_map_on_t0pp(self):
...
+        # Should print a warning as intended in such case
+        with self.assertWarns(Warning):
+            device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 200, "cpu": "1GB"})
+        assert device_map == {"linear1": 0, "batchnorm": 1, "linear2": "cpu", "linear3": "cpu"}
+
+        # Now we have two devices, neither can hold all the buffers, but we are using the offload_buffers=True
+        # Should NOT print a warning in such case
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 200, "cpu": "1GB"}, offload_buffers=True)
```
SunMarc, line 671:
```suggestion
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            device_map = infer_auto_device_map(model, max_memory={0: 400, 1: 200, "cpu": "1GB"}, offload_buffers=True)
        assert len(w) == 0
```

tests/test_modeling_utils.py
```diff
@@ -604,6 +618,59 @@ def test_infer_auto_device_map_on_t0pp(self):
...
+        with self.assertWarns(Warning):
+            device_map = infer_auto_device_map(model, max_memory={0: 400, "cpu": "1GB"})
+        assert device_map == {"linear1": 0, "batchnorm": "cpu", "linear2": "cpu"}
+
+        # Only linear1 (144) fits on device 0, and remaining buffers (batchnorm's 160 + linear2's 120 = 280) won't fit
+        # device 0, but with offload_buffers they won't be loaded to device 0 all at once, so it's ok now
+        # Should NOT print a warning in such case
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            device_map = infer_auto_device_map(model, max_memory={0: 400, "cpu": "1GB"}, offload_buffers=True)
```
SunMarc, line 640:
```suggestion
        with warnings.catch_warnings(record=True) as w:
            warnings.simplefilter("always")
            device_map = infer_auto_device_map(model, max_memory={0: 400, "cpu": "1GB"}, offload_buffers=True)
        assert len(w) == 0
```

SunMarc (commented):
Thanks for working on this PR and adding the tests ! I ran them and they passed. I left a couple of minor comments/suggestions. We can merge this PR after fixing these =)

src/accelerate/utils/modeling.py
```diff
@@ -1098,6 +1124,9 @@ def infer_auto_device_map(
     device_map = OrderedDict()
     current_device = 0
     current_memory_used = 0
+    device_memory_used = {}
+    device_memory_reserved = {}
```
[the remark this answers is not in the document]
notsyncing: Removed `device_memory_reserved` and using a `current_memory_reserved` in each loop instead.

src/accelerate/utils/modeling.py
```diff
@@ -1272,10 +1312,39 @@ def infer_auto_device_map(
...
+                current_buffer_size = compute_module_total_buffer_size(
+                    module, dtype=dtype, special_dtypes=special_dtypes
+                )
+                device_buffer_sizes[device] = device_buffer_sizes.get(device, 0) + current_buffer_size
+
     if clean_result:
         device_map = clean_device_map(device_map)
+
+    if not offload_buffers:
+        non_gpu_buffer_size = device_buffer_sizes.get("cpu", 0) + device_buffer_sizes.get("disk", 0)
```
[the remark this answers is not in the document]
notsyncing: fixed.

tests/test_modeling_utils.py
notsyncing: fixed

tests/test_modeling_utils.py
notsyncing: fixed

tests/test_modeling_utils.py
notsyncing: fixed

src/accelerate/utils/modeling.py
```diff
@@ -1150,6 +1179,7 @@ def infer_auto_device_map(
         # Reduce max size available by the largest layer.
         if devices[current_device] in main_devices:
             current_max_size = current_max_size - max_layer_size
+            device_memory_reserved[current_device] = max_layer_size
```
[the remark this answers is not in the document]
notsyncing: fixed

notsyncing:
> Thanks for working on this PR and adding the tests ! I ran them and they passed. I left a couple of minor comments/suggestions. We can merge this PR after fixing these =)

@SunMarc all fixed.

tests/test_modeling_utils.py
SunMarc: Hi @notsyncing, is there a reason why you didn't to add the `check assert len(w) == 0` with `with warnings.catch_warnings(record=True) as w `: ?  We need that since we don't really pay attention to the warning that show up in the logs

SunMarc (approved):
Thanks for iterating ! LGTM. I re-opened a comment I made. Could you have a quick look ? Also, can you have a second look @muellerzr ?

tests/test_modeling_utils.py
notsyncing: Oh, sorry, I missed this line :joy:, will add it later

muellerzr (approved):
Thanks! Overall this looks sound to me :)

PR #2420, opened 2024-02-07
Make test assertions more idiomatic

akx:
# What does this PR do?

This PR makes the assertions used in the test suite [more idiomatic for Pytest](https://docs.pytest.org/en/stable/how-to/assert.html).

Idiomatic assertions are rewritten by Pytest, making debugging much easier.

Namely, I did a first pass using [codemod-unittest-to-pytest-asserts](https://github.com/akx/codemod-unittest-to-pytest-asserts), followed by some `ast-grep` powered fixes, followed up with some manual fixes to make things neater still.

Sibling of https://github.com/huggingface/trl/pull/1301.

I don't have easy access to hardware suitable for running all tests locally (Apple Silicon 😄), but the tests I could run seem to pass, so here's hoping CI agrees!

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc @muellerzr

### DIFF
--- a/tests/deepspeed/test_deepspeed.py
+++ b/tests/deepspeed/test_deepspeed.py
@@ -623,31 +623,22 @@ class DeepSpeedConfigIntegration(AccelerateTestCase):
             model, _, train_dataloader, eval_dataloader, _ = accelerator.prepare(
                 model, dummy_optimizer, train_dataloader, eval_dataloader, dummy_lr_scheduler
             )
-            self.assertEqual(accelerator.deepspeed_config["train_micro_batch_size_per_gpu"], 16)
-            self.assertEqual(accelerator.deepspeed_config["train_batch_size"], 16)
+            config = accelerator.deepspeed_config
+            assert config["train_micro_batch_size_per_gpu"] == 16
+            assert config["train_batch_size"] == 16
 
-            self.assertEqual(accelerator.deepspeed_config["optimizer"]["params"]["lr"], 5e-5)
-            self.assertEqual(accelerator.deepspeed_config["optimizer"]["params"]["weight_decay"], 1e-4)
+            assert config["optimizer"]["params"]["lr"] == 5e-05
+            assert config["optimizer"]["params"]["weight_decay"] == 1e-4
 
-            self.assertEqual(accelerator.deepspeed_config["scheduler"]["params"]["warmup_min_lr"], 0.0)
-            self.assertEqual(accelerator.deepspeed_config["scheduler"]["params"]["warmup_max_lr"], 5e-5)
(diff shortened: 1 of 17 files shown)


muellerzr (changes requested):
No, it is standard with Hugging Face that we use unittest for our testing suite centralization, not pytest (the pytest part is just for when we need to, such as scripts that need to be ran on multiple devices). to avoid things like this (and the other PRs you have opened that are very, very large changes) can we please open an issue first to discuss them before much time is spent on them? Thanks!

muellerzr (changes requested):
We've decided not to go ahead with this, as our mixture of some pytest and mostly unittest + our own features is a better long-term approach we've found that maintains usability and longevity without the nuances and issues that pytest can provide long-term. Thank you though for your thoughts on the matter!

muellerzr:
We may change our mind in the future (and you're welcome to keep this PR open as a result) but for now that's where we're at here :)

akx:
> as our mixture of some pytest and mostly unittest + our own features

TBH, the current tests are already mixing e.g. `self.assertRaises` and `pytest.raises` within the same file, so there's no way to run the tests without `pytest` anyway. (See e.g. `test_state_checkpointing.py`) 😅  

> is a better long-term approach we've found that maintains usability and longevity without the nuances and issues that pytest can provide long-term. 

What nuances and issues have you found that Pytest provides long-term?

Anyway, I rebased this and backed out the last commit that made things more Pytesty, so now all this does is codemod `unittest`-specific assertions into native Pythonic assertions.

muellerzr:
> TBH, the current tests are already mixing e.g. self.assertRaises and pytest.raises within the same file, so there's no way to run the tests without pytest anyway. (See e.g. test_state_checkpointing.py) 😅

yes, hence why pytest is a dep in the testing requirements install.

muellerzr (approved):
Sorry for the pushback on this, we had a lengthy discussion internally yet again at the start of the week and are okay with using *some* things more py-testy rather than using unittest explictly such as this. 

Thanks for your contribution!

akx:
@muellerzr 

> Edit: can you run `make style; make quality` however?

Rebased + done.

fxmarty:
I don't understand the point. Pytest works smoothly with unittest, why change something that is already working?

BenjaminBossan:
@fxmarty We had some internal discussion to allow using more pytest features going forward. For some of these features, we need to get rid of `unittest.TestCase`, which requires removing `self.assertFoo` type calls. Going from `self.assertFoo` to `assert` is also more readable and we're still getting nice error messages.

fxmarty:
@BenjaminBossan Thank you, which features would that be?

PR #2421, opened 2024-02-07
Drop `parameterized` dependency in favor of Pytest's parametrize

akx:
# What does this PR do?

This PR drops the `parameterized` test dependency in favor of Pytest's own [`pytest.mark.parametrize`](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) feature.

Sibling of #2420 and https://github.com/TimDettmers/bitsandbytes/pull/1001.

> [!WARNING]  
> This will apparently need #2420 to land first.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc @muellerzr @BenjaminBossan

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ extras["quality"] = [
     "ruff ~= 0.1.15",
 ]
 extras["docs"] = []
-extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests", "parameterized"]
+extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests"]
 extras["test_dev"] = [
     "datasets",
     "evaluate",
--- a/tests/deepspeed/test_deepspeed.py
+++ b/tests/deepspeed/test_deepspeed.py
@@ -21,8 +21,8 @@ import tempfile
 from copy import deepcopy
 from pathlib import Path
 
+import pytest
 import torch
-from parameterized import parameterized
 from torch.utils.data import BatchSampler, DataLoader, RandomSampler, SequentialSampler
 from transformers import AutoConfig, AutoModel, AutoModelForCausalLM, get_scheduler
 from transformers.testing_utils import mockenv_context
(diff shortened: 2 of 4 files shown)


BenjaminBossan (commented):
Thanks for providing this PR. I think, however, that this might not work correctly because `pytest.mark.parametrized` and `unittest` don't mesh well. To give an example, I tried this:

```python
import pytest

class TestFoobar1(unittest.TestCase):
    @pytest.mark.parametrize("foo", [1, 2, 3])
    def test_foo(self, foo):
        print(foo)

class TestFoobar2:
    @pytest.mark.parametrize("foo", [1, 2, 3])
    def test_foo(self, foo):
        print(foo)
```

As you can see, the two examples are identical, except that the first is a `unittest.TestCase`. When I try to run the former, it fails because

> TypeError: TestFoobar1.test_foo() missing 1 required positional argument: 'foo'

The latter works. We cannot drop the inheritance from `unittest.TestCase`, so I think this change won't work correctly. Or am I missing something?

Don't get me wrong, I would love to use `@pytest.mark.parametrize`, as I think it's superior to (and less obscure than) `parameterized`, but we have the standard in HF to not use pytest features, only to use it as a test runner.

akx:
@BenjaminBossan Yeah, hence #2420 would need to land first (as noted in the PR description and why this is a draft), so `self.assert*` is not required, and the test cases wouldn't require to derive from `unittest.TestCase`.

If #2420 won't land, this can be closed :)

muellerzr (changes requested):
As Benjamin mentioned, we don't plan on going through with this methodology and intend to keep using the `paramertized` library

akx:
> we have the standard in HF to not use pytest features, only to use it as a test runner.

Kind of. 😁 

https://github.com/huggingface/accelerate/blob/06b138d84537ffb2d1d404f2f198a0446e8d7ec3/tests/test_state_checkpointing.py#L418-L420

PR #2423, opened 2024-02-07
Make output end up on all GPUs at the end

muellerzr:
# What does this PR do?

This PR makes it so that after running pippy inference, the outputs are gathered and returned on each GPU at the end. This does incur a comms call so this can be disabled if users want to keep it on the last GPU only

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @kwen2501

### DIFF
--- a/docs/source/usage_guides/distributed_inference.md
+++ b/docs/source/usage_guides/distributed_inference.md
@@ -216,7 +216,7 @@ with torch.no_grad():
     output = model(*args)
 ```
 
-When finished all the data will be on the CPU on each process for convenience:
+When finished all the data will be on each GPU for convenience:
 
 ```{python}
 print(output)
--- a/docs/source/usage_guides/distributed_inference.md
+++ b/docs/source/usage_guides/distributed_inference.md
@@ -216,10 +216,20 @@ with torch.no_grad():
     output = model(*args)
 ```
 
-When finished all the data will be on each GPU for convenience:
+When finished all the data will be on the last process only:
 
 ```{python}
-print(output)
+from accelerate import PartialState
+if PartialState().is_last_process:
+    print(output)
 ```
+
+<Tip>
+
+    If you pass in `gather_output=True` to [`inference.prepare_pippy`], the output will be sent
+    across to all the GPUs afterwards without needing the `is_last_process` check. This is 
+    `False` by default as it incurs a communication call.
+    
+</Tip>
 
 And that's it! To explore more, please check out the examples in [this repository](https://github.com/muellerzr/pippy-device-map-playground/) and our documentation as we work to improving this integration. 
--- a/docs/source/usage_guides/distributed_inference.md
+++ b/docs/source/usage_guides/distributed_inference.md
@@ -216,13 +216,10 @@ with torch.no_grad():
     output = model(*args)
 ```
 
-When finished, all the data will be on the last GPU, which you can use the [`PartialState`] to find and extract:
+When finished all the data will be on the CPU on each process for convenience:
 
 ```{python}
-from accelerate import PartialState
-
-if PartialState().is_last_process:
-    print(output)
+print(output)
 ```
 
 And that's it! To explore more, please check out the examples in [this repository](https://github.com/muellerzr/pippy-device-map-playground/) and our documentation as we work to improving this integration. 
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -1,6 +1,6 @@
 import math
 from types import MethodType
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from .state import PartialState
 from .utils import (
@@ -110,12 +110,12 @@ def pippy_forward(forward, num_chunks, gather_output, *args, **kwargs):
 
 def prepare_pippy(
     model,
-    split_points="auto",
-    no_split_module_classes=None,
-    example_args=(),
+    split_points: Optional[Union[str, List[str]]] = "auto",
+    no_split_module_classes: Optional[List[str]] = None,
+    example_args: Optional[Tuple[Any]] = (),
     example_kwargs: Optional[Dict[str, Any]] = None,
-    num_chunks=None,
-    gather_output=False,
+    num_chunks: Optional[int] = None,
+    gather_output: Optional[bool] = False,
 ):
     """
     Wraps `model` for pipeline parallel inference.
@@ -123,14 +123,14 @@ def prepare_pippy(
     Args:
         model (`torch.nn.Module`):
             A model we want to split for pipeline-parallel inference
-        split_points (`str`, defaults to 'auto'):
+        split_points (`str` or `List[str]`, defaults to 'auto'):
             How to generate the split points and chunk the model across each GPU. 'auto' will find the best balanced
-            split given any model.
+            split given any model. Should be a list of layer names in the model to split by otherwise.
         no_split_module_classes (`List[str]`):
             A list of class names for layers we don't want to be split.
-        example_args (tuple of `torch.Tensor`):
+        example_args (tuple of model inputs):
             The expected inputs for the model that uses order-based inputs. Recommended to use this method if possible.
-        example_kwargs (dict of `torch.Tensor`)
+        example_kwargs (dict of model inputs)
             The expected inputs for the model that uses dictionary-based inputs. This is a *highly* limiting structure
             that requires the same keys be present at *all* inference calls. Not recommended unless the prior condition
             is true for all cases.
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -6,6 +6,7 @@ from .state import PartialState
 from .utils import (
     calculate_maximum_sizes,
     convert_bytes,
+    gather_object,
     ignorant_find_batch_size,
     infer_auto_device_map,
     is_pippy_available,
@@ -82,7 +83,7 @@ def build_pipeline(model, split_points, args, kwargs, num_chunks):
     return stage
 
 
-def pippy_forward(forward, num_chunks, *args, **kwargs):
+def pippy_forward(forward, num_chunks, send_output_to_cpu, *args, **kwargs):
     state = PartialState()
     output = None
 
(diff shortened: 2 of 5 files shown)


src/accelerate/inference.py
```diff
@@ -101,6 +102,10 @@ def pippy_forward(forward, num_chunks, *args, **kwargs):
         output = forward()
     else:
         forward()
+    if send_output_to_cpu:
+        output = gather_object([output if output is not None else object()])
```
kwen2501, line 106: I wonder if accelerate has a broadcast or a broadcast_object API? Just thinking that may fit the usage here?

src/accelerate/inference.py
```diff
@@ -132,6 +138,9 @@ def prepare_pippy(
         num_chunks (`int`):
             The number of different stages the Pipeline will have. By default it will assign one chunk per GPU, but
             this can be tuned and played with. In general one should have num_chunks >= num_gpus.
+        send_output_to_cpu (`bool`, defaults to `True`):
+            Whether to broadcast the model output across processes and send it to the CPU afterwards or to leave the
+            model output on the last device.
```
kwen2501, line 143: Is it the default case that Accelerate would move result from GPU to CPU for the user? (Regardless of parallelism)
I am just thinking that some commands like `print(output)` (if user needs it) would also move the tensor from GPU to CPU, but that keeps the move more "as needed".
I am just referring to the GPU->CPU part (not the broadcast part), although both of them may have performance implications.

src/accelerate/inference.py
muellerzr: We do, but we can't broadcast when `None` is an input so `gather_object` gets the job done

src/accelerate/inference.py
muellerzr: Yes, this is aligned with how we do things in transformers so we're keeping this behavior. cc @SunMarc

muellerzr: However to keep it barebones, we won't move to the CPU, just keep everything on the GPU for the output

src/accelerate/inference.py
```diff
@@ -111,9 +114,11 @@ def prepare_pippy(
     example_args=(),
     example_kwargs: Optional[Dict[str, Any]] = None,
     num_chunks=None,
+    send_output_to_cpu=True,
```
SunMarc, line 117: it should be false no according to the docstring. Let's add the description of this arg.

src/accelerate/inference.py
```diff
@@ -82,7 +83,7 @@ def build_pipeline(model, split_points, args, kwargs, num_chunks):
     return stage
 
 
-def pippy_forward(forward, num_chunks, *args, **kwargs):
+def pippy_forward(forward, num_chunks, send_output_to_cpu, *args, **kwargs):
```
SunMarc, line 86: `send_output_to_cpu` is not used in the function

docs/source/usage_guides/distributed_inference.md
```diff
@@ -216,13 +216,10 @@ with torch.no_grad():
     output = model(*args)
 ```
 
-When finished, all the data will be on the last GPU, which you can use the [`PartialState`] to find and extract:
+When finished all the data will be on the CPU on each process for convenience:
```
SunMarc, line 219: we need to update this and let the user know that they can change this behavior using `send_output_to_cpu`.

SunMarc (commented):
Thanks ! I left a few comments =)

src/accelerate/utils/operations.py
```diff
@@ -481,6 +481,60 @@ def _tpu_broadcast(tensor, src=0, name="broadcast tensor"):
...
+    Args:
+        tensor (`torch.tensor`):
+            The tensor that should be sent to all devices. Must only have it be defined on a single device, the rest
+            should be `None`.
+    """
+    state = PartialState()
+    shape, dtype = gather_tensor_shape(tensor)
+    if tensor is None:
+        tensor = torch.empty(shape, dtype=TENSOR_INT_TO_DTYPE[dtype]).to(state.device)
+    return reduce(tensor, reduction="sum")
```
kwen2501, line 535: This operation can now use broadcast since we already have the empty tensor on every rank?

src/accelerate/utils/operations.py
```diff
@@ -481,6 +481,60 @@ def _tpu_broadcast(tensor, src=0, name="broadcast tensor"):
...
+
+
+def gather_tensor_shape(tensor):
+    """
+    Grabs the shape of `tensor` only available on one process and returns a tensor of its shape
+    """
+    # Allocate 80 bytes to store the shape
+    max_tensor_dimension = 2**20
+    state = PartialState()
+    base_tensor = torch.empty(max_tensor_dimension, dtype=torch.int, device=state.device)
```
kwen2501, line 507: I wonder if it could be more performant to directly use `cpu` as the device for `base_tensor`?
Reason is that `tensor.shape` and `tensor.dtype` below are both in the cpu domain.
But I don't know if the `reduce` op support `cpu` or not.

src/accelerate/utils/operations.py
```diff
@@ -481,6 +481,60 @@ def _tpu_broadcast(tensor, src=0, name="broadcast tensor"):
...
+
+    if tensor is not None:
+        shape = tensor.shape
+        tensor_dtype = TENSOR_TYPE_TO_INT[tensor.dtype]
+        base_tensor[: len(shape) + 1] = torch.tensor(list(shape) + [tensor_dtype], dtype=int)
+    # Perform a reduction
+    base_tensor = reduce(base_tensor, reduction="sum")
```
kwen2501, line 514: On non-last ranks, `base_tensor = torch.empty(...)` does not guarantee that all elements are initialized to 0. Would there be risk if using `sum` here?

kwen2501 (commented):
Thanks so much for considering the `false` default value.
I left a couple comments based on my shallow understanding of the code. Please pardon me if they are wrong.

src/accelerate/utils/operations.py
muellerzr: Yep! :)

src/accelerate/utils/operations.py
muellerzr: It is not

src/accelerate/utils/operations.py
muellerzr: Yes, those should be `torch.zeros()` instead :)

src/accelerate/utils/operations.py
```diff
@@ -481,6 +481,60 @@ def _tpu_broadcast(tensor, src=0, name="broadcast tensor"):
...
+    base_tensor = base_tensor[base_tensor.nonzero()]
+    dtype = int(base_tensor[-1:][0])
+    base_tensor = base_tensor[:-1]
```
SunMarc, line 517: can you explain a little bit what happens here ?

src/accelerate/test_utils/scripts/test_ops.py
```diff
@@ -129,6 +130,17 @@ def test_op_checker(state):
...
 
+def test_copy_tensor_to_devices(state):
+    if state.distributed_type not in [DistributedType.MULTI_GPU, DistributedType.TPU]:
+        return
+    if state.is_main_process:
+        tensor = torch.tensor([1, 2, 3], dtype=torch.int).to(state.device)
+    else:
+        tensor = None
+    tensor = copy_tensor_to_devices(tensor)
+    assert torch.allclose(tensor, torch.tensor([1, 2, 3], dtype=torch.int, device="cuda"))
```
SunMarc, line 141: nice test !

src/accelerate/inference.py
```diff
@@ -132,6 +137,8 @@ def prepare_pippy(
         num_chunks (`int`):
             The number of different stages the Pipeline will have. By default it will assign one chunk per GPU, but
             this can be tuned and played with. In general one should have num_chunks >= num_gpus.
+        gather_output (`bool`):
+            If `True`, the output from the last GPU (which holds the true outputs) is sent across to all GPUs.
```
SunMarc, line 141: let's also add the defaults value in the docstring

src/accelerate/inference.py
```diff
@@ -101,6 +102,9 @@ def pippy_forward(forward, num_chunks, *args, **kwargs):
         output = forward()
     else:
         forward()
+    if gather_output is True:
```
SunMarc, line 105:
```suggestion
    if gather_output:
```

src/accelerate/inference.py
```diff
@@ -111,9 +115,10 @@ def prepare_pippy(
     example_args=(),
     example_kwargs: Optional[Dict[str, Any]] = None,
     num_chunks=None,
+    gather_output=False,
```
SunMarc, line 118: let's add the type of each arg since you did it in the docstring and for example_kwargs

SunMarc, line 118: Based on the documentation above : `When finished all the data will be on each GPU for convenience:`

```suggestion
    example_args=(),
    example_kwargs: Optional[Dict[str, Any]] = None,
    num_chunks=None,
    gather_output=True,
```

SunMarc (approved):
Thanks for adding this cool feature ! I left a few comments but nothing blocking.

src/accelerate/inference.py
muellerzr: Updating docs to reflect `False`, as per discussion with Ke we want to make this not the default.

src/accelerate/utils/operations.py
muellerzr: Done, tl;dr pytorch doesn't have a way to send just a tensor from A -> B without  B knowing what's going on. So we store the size and dtype the tensor is and perform a `reduce` on an allocated tensor to get that information across to all devices

src/accelerate/utils/operations.py
muellerzr: Ah yes it's just `reduce` not broadcast oops 😓 

Not terrible of a miss though, I can put that fix in as a followup

kwen2501: Agree. Numerically `reduce` may end up correct too.

The other place may need a real fix:
https://github.com/huggingface/accelerate/pull/2423/files#r1483578592

PR #2427, opened 2024-02-08
[Fix] make all tests pass on XPU  

faaany:
## What does this PR do?
When running `make test` on XPUs, I got the following test results: 
```bash 
FAILED tests/test_kwargs_handlers.py::KwargsHandlerTester::test_grad_scaler_kwargs - AttributeError: 'GradScaler' object has no attribute '_init_scale'
FAILED tests/test_optimizer.py::OptimizerTester::test_accelerated_optimizer_step_was_skipped - AssertionError: False is not true
```
The failed reason for these 2 tests is the same: XPU hasn't implemented `torch.xpu.amp.GradScaler` yet and in the current logic, `self.scaler` will be set to `torch.cuda.amp.GradScaler` as shown [here](https://github.com/huggingface/accelerate/blob/06b138d84537ffb2d1d404f2f198a0446e8d7ec3/src/accelerate/accelerator.py#L449). 
To make the tests pass on XPU devices, we should skip these 2 tests on XPUs. 

In addition to that, I noticed that there is no `require_npu` function in the `testing.py`. So I added it to the patch. 

I am not sure whether it would be better to just use `require_cuda` instead of the combi of `require_non_cpu` and `require_non_xpu`. My goal is just to make all these tests work on XPU. Please let me know your thoughts. And if this patch is too small, just let me know as well. I can think about more tests related to XPU. Thanks a lot!

@muellerzr


muellerzr (approved):
Thanks, this skip format is indeed the right way!

Issue #2220, opened 2023-12-05
split_between_processes support for Datasets


conceptofmind:
### System Info

```Shell
python 3.10
pytorch 2.1
cuda 12.1
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

This does not work:
```python
dataset = load_dataset(
    "wikitext",
    "wikitext-2-v1",
    split="train",
)

print(len(dataset)) # 36718

# Assume 8 processes

with accelerator.split_between_processes(dataset) as data:
    print(len(data)) # 36718
    
   for out in pipe(
        KeyDataset(data, "text"), 
    ):
        continue
```
This works:
```python
dataset = load_dataset(
    "wikitext",
    "wikitext-2-v1",
    split="train",
)
print(len(dataset))

# Assume 8 processes

with accelerator.split_between_processes(dataset["text"]) as data:
    print(len(data))
```
Moving from: https://github.com/huggingface/transformers/issues/27838

### Expected behavior

I am unsure if this is intended behavior or not. A dataset consisting of the lists or input id tensors is split correctly. The dictionary provided from datasets will not be split.

Maybe this is more of a feature request?

Where the dataset could be split across all 8 processes to be used with the transformers feature extraction pipeline for inference.

muellerzr:
This would indeed be a feature request. Simply because a `Dataset` is not actually a dict:

```python
from datasets import load_dataset

dataset = load_dataset(
    "wikitext",
    "wikitext-2-v1",
    split="train",
)
isinstance(dataset, dict)
# Returns False
```

conceptofmind:
> This would indeed be a feature request. Simply because a `Dataset` is not actually a dict:
> 
> ```python
> from datasets import load_dataset
> 
> dataset = load_dataset(
>     "wikitext",
>     "wikitext-2-v1",
>     split="train",
> )
> isinstance(dataset, dict)
> # Returns False
> ```

Thanks for the response.

I should have been more explicit about what I had meant in relation to the returned dictionaries from datasets. I do understand that they are not dictionaries directly.

muellerzr:
Agreed, I'm saying this is why it's failing, we look for this :)

conceptofmind:
[...]

PR #2433, opened 2024-02-12
split_between_processes for Dataset

geronimi73:
# What does this PR do?
make `split_between_processes` handle a `datasets.Dataset`

Fixes https://github.com/huggingface/accelerate/issues/2220

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -22,6 +22,7 @@ import warnings
 from contextlib import contextmanager
 from functools import partial
 from typing import Any, Callable, Optional
+from datasets import Dataset
 
 import torch
 
@@ -426,7 +427,7 @@ class PartialState:
             self.wait_for_everyone()
 
     @contextmanager
-    def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor, apply_padding: bool = False):
+    def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor | Dataset, apply_padding: bool = False):
         """
         Splits `input` between `self.num_processes` quickly and can be then used on that process. Useful when doing
         distributed inference, such as with different prompts.
@@ -499,6 +500,12 @@ class PartialState:
                 for key in inputs.keys():
                     inputs[key] = _split_values(inputs[key], start_index, end_index)
                 return inputs
+            elif isinstance(inputs, Dataset):
+                if start_index >= len(inputs):
+                    start_index=len(inputs)-1
+                if end_index > len(inputs):
+                    end_index=len(inputs)
+                return inputs.select(range(start_index, end_index))
             else:
                 return inputs
 
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -509,7 +509,10 @@ class PartialState:
                             start_index = len(inputs) - 1
                         if end_index > len(inputs):
                             end_index = len(inputs)
-                        return inputs.select(range(start_index, end_index))
+                        result_idcs = list(range(start_index, end_index))
+                        if apply_padding:
+                            result_idcs += [end_index - 1] * (num_samples_per_process - len(result_idcs))
+                        return inputs.select(result_idcs) 
                 return inputs
 
         yield _split_values(inputs, start_index, end_index)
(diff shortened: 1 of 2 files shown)


src/accelerate/state.py
```diff
@@ -22,6 +22,7 @@
 from contextlib import contextmanager
 from functools import partial
 from typing import Any, Callable, Optional
+from datasets import Dataset
```
muellerzr, line 25: We cannot have this. Let's do a check for `is_datasets_available()` and potentially add it into the import utils file.

src/accelerate/state.py
```diff
@@ -426,7 +427,7 @@ def _goes_first(self, is_main: bool):
             self.wait_for_everyone()
 
     @contextmanager
-    def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor, apply_padding: bool = False):
+    def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor | Dataset, apply_padding: bool = False):
```
muellerzr, line 430: Use `"Dataset"` instead. Or `"datasets.Dataset"`. 

If that doesn't work let's just add a snippet in the doc of "can also split `datasets.Dataset`s"

muellerzr (changes requested):

geronimi73:
added the `is_datasets_available ` check and datasets `import` at the very end: import only when it's not a tensor or dict or list, only import and check instance if input is none of those things. not sure if that's what you had in mind

src/accelerate/state.py
```diff
@@ -500,7 +501,17 @@ def _split_values(inputs, start_index, end_index):
                     inputs[key] = _split_values(inputs[key], start_index, end_index)
                 return inputs
             else:
-                return inputs
+                if is_datasets_available():
+                    from datasets import Dataset
+
+                if isinstance(inputs, Dataset):
```
shagunsodhani, line 507: probably this should also be indented to be within the scope of if-condition in line 504

geronimi73: right! thanks

muellerzr (approved):
Great work on this! One last thing, can we please add a test in `test_utils/scripts/test_script.py` to include a test for `Datasets`? (See the `split_between_processes` group)

geronimi73:
one thing to note is that `apply_padding` is ignored in the current implementation. not sure if that's needed or okay to ignore and add to function doc?

src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -540,6 +541,19 @@ def training_check(use_seedable_sampler=False):
...
 
+def test_split_between_processes_dataset():
+    assert is_datasets_available(), "test requires Hugging Face datasets"
+    from datasets import Dataset
+
+    state = AcceleratorState()
+    data = Dataset.from_list([dict(k=v) for v in range(2 * state.num_processes)])
+    with state.split_between_processes(data) as results:
+        assert (
+            len(results) == 2
```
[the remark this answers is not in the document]
geronimi73: > We only test for if it can be evenly split. In Accelerate we need to make sure we cover bases when evenly split is not an option. Can you write a test for an odd number of dataset items?

done

>If that test fails, you will need to do the padding logic similar to what we do.

done

> Also this needs to be added down under if __name__ == "__main__" where the other test_split_x is to actually be called.

ok

> If we really wanted to get nitty/gritty, down there is where we can also include is_datasets_available() before calling the test function instead of having the assert here.

also done but it's ugly. because of scope, an import in `main()` is not accessible in `test_split_between_processes_dataset()` that's why we now pass the imported `datasets.Dataset` as a parameter to  `test_split_between_processes_dataset()` 

**one more thing**, with small datasets very counter-intuitive things happen because of the `math.ceil` in `num_samples_per_process = math.ceil(length / self.num_processes)`.

For example
```
Num. of GPUs: 5,
Num. of Samples: 6 ([1, 2, 3, 4, 5, 6])
Apply padding: False
GPU0: (['1', '2'])	GPU1: (['3', '4'])	GPU2: (['5', '6'])	GPU3: (['6'])	GPU4: (['6'])
```

but the same happens with `split_between_processes` for a `list` so I guess that's an accepted behaviour.

muellerzr (approved):
Thanks! (Sorry this took a bit)

Issue #2405, opened 2024-02-01
Doesn't work with new PyTorch TensorDict


andersonbcdefg:
### System Info

```Shell
`Accelerate` version: 0.26.1
- Platform: Linux-5.19.0-45-generic-x86_64-with-glibc2.31
- Python version: 3.9.16
- Numpy version: 1.23.4
- PyTorch version (GPU?): 2.2.0+cu121 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 88.38 GB
- GPU type: NVIDIA A100-SXM4-80GB
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. Use a TensorDict as the dataset passed into a PyTorch dataloader. For example, see the PyTorch example here: https://pytorch.org/tensordict/tutorials/data_fashion.html
2. Use accelerate to `accelerator.prepare` the dataloader.
3. Try to load batches from the prepared dataloader.

When I did this, I get the following error: 
```
ValueError: batch size was not specified when creating the TensorDict instance and it could not be retrieved from source.
```

This happens when Accelerate is putting the tensor on the right device. It seems to lose track of its batch_size attribute in the process. When I got rid of accelerator.prepare on my dataset (which is fine for now cause I'm training on 1 GPU) this error went away.

### Expected behavior

Expected behavior: it successfully returns batches from the dataloader.

vmoens:
Hi, thanks for flagging this.
I'll have a look at it on the tensordict side to see if there's anything we can do.

We were initially reluctant to have the empty batch-size assumed by default but it wouldn't be a big deal to allow it if that unblocks things on accelerate end.

andersonbcdefg:
@vmoens Wow how did you even find this issue 😄 Really appreciate this! I just started using TensorDicts and it's such a cool and useful abstraction, I was able to delete a lot of code. Excited to see it get battle-hardened and integrated into more things like Accelerate!

vmoens:
[...]

PR #2437, opened 2024-02-12
[WIP] Using pytree for send_to_device

vmoens:
# What does this PR do?

This PR is a prototype to use pytree for the `send_to_device` as discussed in #2405

The core idea is to allow for any pytree structure to work with send_to_device, allowing a wide range of datastructure to be processed when they are returned by a dataloader. For instance, it solves #2405 because it relies on the fact that tensordict is a pytree-structure, thereby pytree takes care of passing the metadata of the tensordict during reconstruction and the output tensordict has the desired batch-size. Any tuple-augmented structure or dict-augmented structure will work too if they're properly registered.

The function `tree_map_with_path` that I use isn't part of the current release of PyTorch, so we may have to either wait or find a less straightforward solution to this issue.

cc @zou3519

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

<!-- Your PR will be replied to more quickly if you can figure out the right person to tag with @

 If you know how to use git blame, that is the easiest way, otherwise, here is a rough guide of **who to tag**.

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -150,44 +150,48 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
             The data to send to a given device.
         device (`torch.device`):
             The device to send the data to.
+        non_blocking (bool): `non_blocking` for `torch.to()` calls.
+        skip_keys (str/list of str): an optional list of strings to skip when
+            a dictionary is encountered.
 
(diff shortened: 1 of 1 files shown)


vmoens:
#2438 will solve #2405 so I think we can safely drop this PR for now until pytree is fully integrated in torch

PR #2438, opened 2024-02-13
[Refactor] Refactor send_to_device to treat tensor-like first

### LINKED ISSUE #2405: Doesn't work with new PyTorch TensorDict

vmoens:
# What does this PR do?

I propose to reorder `send_to_device` to directly call `to` if it's available, which will make it possible to use `TensorDict` OOB.
The function should be slightly faster too because you presumably have more tensors than containers in the structures send to `send_to_device` (so the new implementation will result in fewer checks which are known to introduce overhead in python).

Fixes #2405 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


muellerzr (approved):
Thanks for the refactor! Indeed not having to check and just go should make things quite a bit faster :)

muellerzr:
@vmoens can you solve the conflicts and then we can merge this?

vmoens:
This should work!

muellerzr:
Great! Thanks for this. Let's keep an eye on `main` to make sure nothing fails on our single and multi-GPU CI's after merging (and these tests pass here)

PR #2439, opened 2024-02-13
Fix tied_pointers_to_remove type

fxmarty:
Keep this variable as a set, since some people may call `post_forward` multiple times.

Having it reset as `None` is breaking in the sense that `post_forward` may be called multiple times before, while it can not anymore on main.

Context: https://github.com/huggingface/optimum-amd/pull/62


fxmarty:
cc @Giuseppe5

SunMarc (approved):
LGTM !

PR #2441, opened 2024-02-13
[Big deprecation] Introduces a `DataLoaderConfig`

muellerzr:
# What does this PR do?

As we prepare for 1.0.0, this PR introduces a large deprecation to some of the arguments in the Accelerator, wherein all arguments related to the configuration of the `DataLoader` are now sent through the `DataLoaderConfig`. After this PR is merged, we will aim for a 2-release gap between this and 1.0.0 (by May at the latest), as this is a very large (and potentially scary) change for users. 

The reasoning is to make it easier for integrations like `transformers` (and in the future accelerate in general) to configure seperate entities related to more config-heavy items in the accelerator more manageable, and leave the base params in the Accelerator (without config) to be true core important items.

Mapping of old to new:

```diff
-accelerator = Accelerator(split_batches=True, dispatch_batches=True)
+dl_config = DataLoaderConfig(split_batches=True, dispatch_batches=True)
+accelerator = Accelerator(dataloader_config=dl_config)
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @BenjaminBossan


src/accelerate/accelerator.py
```diff
@@ -421,10 +427,30 @@ def __init__(
...
-        self.even_batches = even_batches
-        self.use_seedable_sampler = use_seedable_sampler
+        if dataloader_config is None:
+            self.dataloader_config = DataLoaderConfiguration()
+        # Deal with deprecated args
+        deprecated_dl_args = {}
+        if dispatch_batches is not None:
+            deprecated_dl_args["dispatch_batches"] = dispatch_batches
+            self.dataloader_config.dispatch_batches = dispatch_batches
+        if split_batches is not True:
```
shagunsodhani, line 437: shouldnt we check if this value is not `None` (similar to lie 434) ?

muellerzr: No, only for ones whose value changes from the default. While dispatch_batches is None by default, split_batches (and more) are not

src/accelerate/accelerator.py
```diff
@@ -421,10 +427,30 @@ def __init__(
...
-        self.split_batches = split_batches
-        self.dispatch_batches = dispatch_batches
-        self.even_batches = even_batches
-        self.use_seedable_sampler = use_seedable_sampler
+        if dataloader_config is None:
+            self.dataloader_config = DataLoaderConfiguration()
+        # Deal with deprecated args
+        deprecated_dl_args = {}
+        if dispatch_batches is not None:
+            deprecated_dl_args["dispatch_batches"] = dispatch_batches
```
BenjaminBossan, line 435: There is a bit of a danger that if a user passes `Accelerator(split_batches=True)` (i.e. passes the default), they will not see the deprecation message and their code will eventually fail. Two solutions come to mind:

- Remove the deprecated arguments from the `__init__` signature. Catch them with `kwargs` and check if they were passed explicitly.
- Use a sentinel value instead of the default and check if that value is being used:

```python
_split_batches_default = object()
...
def __init__(
    self,
    ...,
    split_batches=_split_batches_default,
):
    if split_batches is _split_batches_default:
        split_batches = True
    else:
        deprecated_dl_args["dispatch_batches"] = dispatch_batches
```

Both not very nice unfortunately.

src/accelerate/accelerator.py
```diff
@@ -421,10 +427,30 @@ def __init__(
...
+            deprecated_dl_args["even_batches"] = even_batches
+            self.dataloader_config.even_batches = even_batches
+        if use_seedable_sampler:
+            deprecated_dl_args["use_seedable_sampler"] = use_seedable_sampler
+            self.dataloader_config.use_seedable_sampler = use_seedable_sampler
+        if len(deprecated_dl_args) > 0:
+            values = ", ".join([f"{k}={v}" for k, v in deprecated_dl_args.items()])
+            warnings.warn(
+                f"Passing the following arguments to `Accelerator` is deprecated and will be removed in version 1.0 of Accelerate: {deprecated_dl_args.keys()}. "
+                "Please use `DataLoaderConfiguration` instead: \n"
```
BenjaminBossan, line 450:
```suggestion
                "Please pass `accelerate.utils.DataLoaderConfiguration` instead: \n"
```

Maybe even elevate this to `accelerate.DataLoaderConfiguration`?

src/accelerate/accelerator.py
BenjaminBossan: `split_batches` is `False` by default though.

tests/test_accelerator.py
```diff
@@ -55,6 +55,25 @@ def parameterized_custom_name_func(func, param_num, param):
 
 
 class AcceleratorTester(AccelerateTestCase):
+    # Should be removed after 1.0.0 release
+    def test_deprecated_values(self):
```
BenjaminBossan, line 59: How about adding these tests:

- Don't pass any of these deprecated arguments
- Pass only some of these arguments
- Pass some of these arguments but using their defaults

BenjaminBossan (changes requested):
In general, this looks good. I have some minor comments and then a discussion about detecting the deprecated arguments, where I'm unsure what the best course of action is. Please take a look.

src/accelerate/accelerator.py
muellerzr: The first one is better I think here. Not great, yes, but an easier transition especially when dealing with defaults.

muellerzr: Actually, the Accelerator doesn't have the ability to do `kwargs`. So we will have to do the first.

src/accelerate/accelerator.py
muellerzr: Changed with recommendations to be a sentinel value

docs/source/quicktour.md
```diff
@@ -83,7 +83,7 @@ is shuffled the same way (if you decided to use `shuffle=True` or any kind of ra
     your script. For instance, training on 4 GPUs with a batch size of 16 set when creating the training dataloader will
     train at an actual batch size of 64 (4 * 16).
     If you want the batch size remain the same regardless of how many GPUs the script is run on, you can use the 
-    option `split_batches=True` when creating and initializing [`Accelerator`].
+    option `split_batches=True` when creating and initializing [`Accelerator`] by passing in a [`utils.DataLoaderConfig`].
```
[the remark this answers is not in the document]
BenjaminBossan: Still open, no?

src/accelerate/accelerator.py
```diff
@@ -149,6 +150,12 @@
 
 logger = get_logger(__name__)
 
+# Sentinel values for defaults
+_split_batches = object()
+_dispatch_batches = object()
+_even_batches = object()
+_use_seedable_sampler = object()
```
BenjaminBossan, line 157: Ah, one disadvantage of this is that the type annotation is not correct, which might lead to some issues with IDEs, not sure. But since it's only transitory, I guess we can live with that.

src/accelerate/accelerator.py
```diff
@@ -511,6 +524,22 @@ def local_process_index(self):
...
+    @property
+    def split_batches(self):
+        return self.dataloader_config.split_batches
+
+    @property
+    def dispatch_batches(self):
+        return self.dataloader_config.dispatch_batches
+
+    @property
+    def even_batches(self):
```
BenjaminBossan, line 536: This suggestion would currently fail:

https://github.com/huggingface/accelerate/blob/613ad7089a4343341c1d89ab96998703b4dbf65c/src/accelerate/data_loader.py#L158

So either the suggestion could be adjusted, or we should have setters for these properties (which would be better for BC overall).

BenjaminBossan (changes requested):
I found another potential issue with backwards compatibility, please check.

src/accelerate/accelerator.py
muellerzr: They also have the type-hint in the signature which should help.

tests/test_accelerator.py
```diff
@@ -55,6 +55,43 @@ def parameterized_custom_name_func(func, param_num, param):
...
+            assert accelerator.dispatch_batches is True, "dispatch_batches should be True"
+            assert accelerator.even_batches is True, "even_batches should be True by default"
+            assert accelerator.use_seedable_sampler is False, "use_seedable_sampler should be False by default"
+            assert "dispatch_batches" in deprecation_warning
+            assert "split_batches" in deprecation_warning
+            assert "even_batches" not in deprecation_warning
+            assert "use_seedable_sampler" not in deprecation_warning
+
+        # Pass in some arguments, but with their defaults
+        with self.assertWarns(FutureWarning) as cm:
```
BenjaminBossan, line 84: If we want to stick with pytest features, we could `with pytest.warns` here.

BenjaminBossan (approved):
Looks good now from my side, thanks for working on this refactor.

src/accelerate/accelerator.py
```diff
@@ -422,10 +414,32 @@ def __init__(
...
+            self.dataloader_config.split_batches = split_batches
+        if even_batches is not _even_batches:
+            deprecated_dl_args["even_batches"] = even_batches
+            self.dataloader_config.even_batches = even_batches
+        if use_seedable_sampler is not _use_seedable_sampler:
+            deprecated_dl_args["use_seedable_sampler"] = use_seedable_sampler
+            self.dataloader_config.use_seedable_sampler = use_seedable_sampler
+        if len(deprecated_dl_args) > 0:
+            values = ", ".join([f"{k}={v}" for k, v in deprecated_dl_args.items()])
+            warnings.warn(
```
SunMarc, line 437: We could maybe use `logger.warning` so that users have more control of their logs.

tests/test_accelerator.py
```diff
@@ -55,6 +56,43 @@ def parameterized_custom_name_func(func, param_num, param):
 
 
 class AcceleratorTester(AccelerateTestCase):
+    # Should be removed after 1.0.0 release
+    def test_deprecated_values(self):
```
SunMarc, line 60: nice !

SunMarc (approved):
Thx for your work !

src/accelerate/accelerator.py
muellerzr: In this case we don't want them to mute that as these deprecations aren't ones that will exist for many, many months. They'll exist for a very short (relative) time. We have this for the other 1.0 warnings as well :) 

(Just checked, same with transformers)

PR #2442, opened 2024-02-13
Fix warning when dispatching model 

SunMarc:
# What does this PR do ?
This PR fixes the wraper around .to() method when we dispatch the model. We were displaying this warning even when the user was not moving the model but casting the model to a specific dtype for example. cc @sayakpaul 
Fixes one of the issues in this PR https://github.com/huggingface/diffusers/pull/6857


muellerzr (approved):
Interesting fix, nice catch!

sayakpaul (approved):
Thanks for the quick resolve!

PR #2445, opened 2024-02-14
Fix type hints for PEP 484 compliance, enable UP007

akx:
# What does this PR do?

This PR runs [`no_implicit_optional`](https://github.com/hauntsaninja/no_implicit_optional) to make the type hinting compliant with [PEP 484](https://peps.python.org/pep-0484/) (see the readme for `no_implicit_optional` for context), IOW more accurate.

It then enables the previously-disabled `UP007` Ruff rule to make the type annotations shorter (`X | Y`) where possible. (Since not all modules are using `from __typing__ import annotations`, and `accelerate` targets Python 3.8+, that's not everywhere. Targeting Python 3.10 will enable `X | Y` everywhere, see [PEP 604](https://peps.python.org/pep-0604/).)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc @muellerzr @BenjaminBossan

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ ignore = [
     "E501", # Line length (handled by ruff-format)
     "E741", # Ambiguous variable name
     "W605", # Invalid escape sequence
-    "UP007", # X | Y type annotations
 ]
 
 [tool.ruff.lint.per-file-ignores]
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -27,7 +27,7 @@ from collections import OrderedDict
 from contextlib import contextmanager
 from functools import partial
 from types import MethodType
-from typing import Any, Callable, Union
+from typing import Any, Callable
 
 import torch
 import torch.utils.hooks as hooks
@@ -625,7 +625,7 @@ class Accelerator:
         with PartialState().split_between_processes(inputs, apply_padding=apply_padding) as inputs:
             yield inputs
 
-    def on_main_process(self, function: Callable[..., Any] = None):
+    def on_main_process(self, function: Callable[..., Any] | None = None):
         """
         A decorator that will run the decorated function on the main process only. Can also be called using the
         `PartialState` class.
@@ -664,7 +664,7 @@ class Accelerator:
 
         return _inner
 
-    def on_local_main_process(self, function: Callable[..., Any] = None):
+    def on_local_main_process(self, function: Callable[..., Any] | None = None):
         """
         A decorator that will run the decorated function on the local main process only. Can also be called using the
         `PartialState` class.
@@ -745,7 +745,7 @@ class Accelerator:
 
         return _inner
 
-    def on_process(self, function: Callable[..., Any] = None, process_index: int = None):
+    def on_process(self, function: Callable[..., Any] | None = None, process_index: int | None = None):
         """
         A decorator that will run the decorated function on a given process index only. Can also be called using the
         `PartialState` class.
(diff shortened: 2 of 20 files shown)


BenjaminBossan:
We have a mix of `Optional[Foo]` and `Foo | None`, I wonder if we should favor one way over the other, at least for newly added code.

akx:
> We have a mix of `Optional[Foo]` and `Foo | None`, I wonder if we should favor one way over the other, at least for newly added code.

Unless you know there's code that needs introspection of annotation objects, I would recommend every file to have `from __typing__ import annotations` (which can be automated with Ruff).

BenjaminBossan:
> Unless you know there's code that needs introspection of annotations, I would recommend every file to have `from __typing__ import annotations` (which can be automated with Ruff).

I've been adding this import to PEFT and haven't encountered any issues. We use types very loosely anyway, I think that should be fine. I assume that the implication is that if we add `from __future__ import annotations` everywhere, we can use the `|` pattern instead of `Optional`?

akx:
> I assume that the implication is that if we add `from __future__ import annotations` everywhere, we can use the `|` pattern instead of `Optional`?

Yes. Enabling the future makes all annotations not actually evaluated at runtime, IOW they can use syntax that the concrete types won't support in the target interpreter version. (What a convoluted sentence!)

In other words, even if for Python 3.8 the concrete Type type doesn't support the `|` operator to turn it to an `Union` of itself and another type, type checkers such as Mypy and Pyright are smart enough to support the syntax as they're just reading the AST.

The flipside is that libraries that do require evaluating annotations (such as Pydantic) will require additional smarts to evaluate the deferred annotations, but the mainstream libraries (such as Pydantic) already support this in recent versions.

muellerzr:
For newly added code I agree we can just stick with `|`, the import will help fix anyone on a lower python version that doesn't support it "natively" yet. And at least getting started with this is good enough, as in general while we don't strictly type everything this is enough of a starting point

muellerzr:
@akx do you want to go ahead and include `from __typing__ import annotations` + the ruff rule for it? :)

akx:
@muellerzr I can do that in a follow-up PR to keep things clean and easier to review?

muellerzr:
Although, since we have the type annotations here being used, let's just add it here as our CI runs python 3.8 iirc so it'll just immediately break

akx:
@muellerzr Are you sure you want it here? The diff looks to be `121 files changed, 374 insertions(+), 234 deletions(-)`...

muellerzr:
Perhaps not...!

BenjaminBossan (approved):
Apart from the merge conflict, LGTM.

BenjaminBossan:
I just saw that we have other incorrectly typed optionals, outside of function signatures, e.g. here:

https://github.com/huggingface/accelerate/blob/21b225e8d556cf979ac60ba017af9a5cdfd6c8db/src/accelerate/commands/config/config_args.py#L183-L201

Those don't seem to be caught by `no_implicit_optional`. Not a blocker for this PR, just something to keep in mind.

BenjaminBossan:
> Unless you know there's code that needs introspection of annotations

There does indeed appear to be a problem with `HfArgumentParser`: https://github.com/huggingface/peft/pull/1537

akx:
> There does indeed appear to be a problem with `HfArgumentParser`: [huggingface/peft#1537](https://github.com/huggingface/peft/pull/1537)

... but only for `dataclasses`, which this PR doesn't touch 😄

muellerzr:
But for downstream libraries that rely on accelerate and will want to integrate with the HfArgumentParser, this could matter.

akx:
@muellerzr As I see it, HfArgumentParser only tries to parse annotations for class fields, not function arguments.

And anyway, ` | ` style annotations have been in `main` since at least April 2023, see e.g. https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L249-L266

BenjaminBossan:
Let's revive this PR.

PR #2449, opened 2024-02-15
Remove unnecessary `env=os.environ.copy()`s

akx:
# What does this PR do?

Subprocess execution calls had a copy-pasted `env=os.environ.copy()` tacked onto them.

When `env=None`, quoth the docs for `subprocess.run`:

> If env is not `None`, it must be a mapping that defines the environment variables for the new process; these are used **instead of the default behavior of inheriting the current process’ environment.**

, emphasis mine.

Unless someone expects `execute_subprocess_async()` to modify the `env` passed in, these are all unnecessary copy-paste. If `execute_subprocess_async()` is expected to modify it, then we can add ```if env is None: env = os.environ.copy()` within it.

Sibling of #2446.

## Before submitting
      Pull Request section?
      to it if that's the case.
  - No user-facing changes.
  - Shouldn't be necessary.

## Who can review?

cc @muellerzr @BenjaminBossan

### DIFF
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 import argparse
-import os
 
 from accelerate.test_utils import execute_subprocess_async, path_in_accelerate_package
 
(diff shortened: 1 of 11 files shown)


muellerzr (approved):
Thanks for the TIL, if the tests pass just fine then I see no issue with the change :) (I'll pull it down and run on a distributed system to be sure)

akx:
@muellerzr Rebased.

BenjaminBossan (approved):
Thanks, looks good.

PR #2450, opened 2024-02-15
Context manager fixes

akx:
# What does this PR do?

* Configures Ruff to prevent `os.getenv` and `os.setenv`; there should be one way to manage the env, and `os.environ` was used more.
* `clear_environment()` didn't actually clear environment variables, it just assigned a new `os.environ` proxy dictionary.
* The environment management context managers wouldn't clean up after themselves if the inner block raised.
  * There was another suspect place where cleanup code after `yield` is not wrapped; marked that as a TODO (@muellerzr seems to have touched that function last))

## Before submitting
      Pull Request section?
      to it if that's the case.
  -  No user-facing changes other than the CMs being more "do-what-I-mean"
  - Sure did!

## Who can review?

cc @muellerzr @BenjaminBossan

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,8 +10,9 @@ extend-select = [
     "E", # PEP8 errors
     "F", # PEP8 formatting
     "I", # Import sorting
-    "W", # PEP8 warnings
+    "TID251", # Banned API
     "UP", # Pyupgrade
+    "W", # PEP8 warnings
 ]
 ignore = [
     "E501", # Line length (handled by ruff-format)
@@ -34,3 +35,8 @@ known-first-party = ["accelerate"]
 exclude = [
     "manim_animations/*"
 ]
+
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"os.getenv".msg = "Use os.environ instead"
+"os.putenv".msg = "Use os.environ instead"
+"os.unsetenv".msg = "Use os.environ instead"
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -19,6 +19,7 @@ import warnings
 from collections import UserDict, namedtuple
 from unittest.mock import Mock, patch
 
+import pytest
 import torch
 from torch import nn
 
@@ -28,6 +29,7 @@ from accelerate.test_utils.training import RegressionModel
 from accelerate.utils import (
     CannotPadNestedTensorWarning,
     check_os_kernel,
+    clear_environment,
     convert_outputs_to_fp32,
     convert_to_fp32,
     extract_model_from_parallel,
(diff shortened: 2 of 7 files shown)


akx:
The test failures are interesting:
> torch._dynamo.exc.InvalidBackend: Invalid backend: 'aot_ts_nvfuser', see `torch._dynamo.list_backends()` for available backends.

That string is set in an unrelated `test_torch_dynamo_plugin` test (which admonishes about this exact thing)...

tests/test_utils.py
```diff
@@ -134,6 +136,26 @@ def test_patch_environment_key_exists(self):
         assert "BB" not in os.environ
         assert "CC" not in os.environ
 
+    def test_patch_environment_restores_on_error(self):
+        # we need to find an upper-case envvar
+        # because `patch_environment upper-cases all keys...
+        key, orig_value = next(kv for kv in os.environ.items() if kv[0].isupper())
```
BenjaminBossan, line 142: If there is no upper case env var, this will raise an error. Very unlikely, just mentioning it.

tests/test_utils.py
```diff
@@ -134,6 +136,26 @@ def test_patch_environment_key_exists(self):
...
+        new_value = orig_value[::-1]
```
BenjaminBossan, line 143: Maybe: `new_value = orig_value + "foo"` to be 100% sure that the values differ?

tests/test_utils.py
```diff
@@ -134,6 +136,26 @@ def test_patch_environment_key_exists(self):
...
+        with pytest.raises(RuntimeError), patch_environment(**{key: new_value}):
+            assert os.environ[key] == os.getenv(key) == new_value  # noqa: TID251
+            raise RuntimeError("Oopsy daisy!")
+        assert os.environ[key] == os.getenv(key) == orig_value  # noqa: TID251
```
BenjaminBossan, line 147: Why test both `os.environ[key]` and `os.getenv(key)`, can they differ?

tests/test_utils.py
```diff
@@ -134,6 +136,26 @@ def test_patch_environment_key_exists(self):
...
+
+    def test_clear_environment(self):
+        key, value = os.environ.copy().popitem()
+        with pytest.raises(RuntimeError), clear_environment():
+            assert key not in os.environ
+            assert not os.getenv(key)  # test the environment is actually cleared  # noqa: TID251
+            raise RuntimeError("Oopsy daisy!")
+        # Test values are restored
+        assert os.getenv(key) == value  # noqa: TID251
+        assert os.environ[key] == value
```
BenjaminBossan, line 157: Same question about `os.environ[key]` and `os.getenv(key)`

BenjaminBossan (commented):
Thanks for making these changes, the result is indeed more robust than what the code currently does. I have some comments regarding the tests, please take a look.

tests/test_utils.py
akx: I guess, but palindromes were more fun!

tests/test_utils.py
akx: They can't, but the issue with the preimage of `clear_environment()` was exactly that it wouldn't touch what `os.getenv()` would return, so it's belt-and-suspenders here :)

tests/test_utils.py
akx: Same answer!

pyproject.toml
```diff
@@ -34,3 +35,7 @@ known-first-party = ["accelerate"]
 exclude = [
     "manim_animations/*"
 ]
+
+[tool.ruff.lint.flake8-tidy-imports.banned-api]
+"os.setenv".msg = "Use os.environ instead"
+"os.getenv".msg = "Use os.environ instead"
```
BenjaminBossan, line 41: Is this because there are some issues with these functions or just to enforce a uniform way of getting env vars?

BenjaminBossan (approved):

pyproject.toml
muellerzr: Also I don't think `os.setenv` is a thing?

akx: Heh, yeah, `setenv` isn't a thing, it's `putenv`. Will fix.

But yeah, as the PR description says (I don't blame you, I tend to skim them too in general...), to prefer a single way.

akx:
Rebased. Previous test failures were due to Transformers being naughty:
```
transformers/testing_utils.py:129: in <module>
    from _pytest.doctest import (
E   ImportError: cannot import name 'import_path' from '_pytest.doctest' (/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/_pytest/doctest.py)
```

muellerzr (approved):
Thanks for the back and forth!

BenjaminBossan (approved):
Thanks, LGTM.

PR #2456, opened 2024-02-15
[docs] Quicktour

stevhliu:
This PR updates the Quicktour to focus more on the library's core offerings:

➕ organizes the Quicktour around the three main features: unified launcher, `Accelerator` class, and Big Model Inference
➕ new tutorials for Execution process and TPU training
➕ adds save Transformer models from Accelerator API page to same section with saving/loading models in Add Accelerate to your code tutorial
➕ adds `on_local_main_process`, `on_main_process`, `on_process`, and `on_local_process` from Accelerator API page to the Execution process docs
➖ removes the "Common modifications of the base case" section because all these different scenarios can be overwhelming for a new user and because it is not necessarily needed to help users start quickly. These sections will be integrated as a part of the tutorials, and I'll add links to redirect users there from the Quicktour. Only exception is the Launching distributed training from a notebook because there is already a tutorial for that.

- [x] move content from the "Common modifications of the base case" section into the Tutorials

### DIFF
--- a/docs/source/basic_tutorials/execution.md
+++ b/docs/source/basic_tutorials/execution.md
@@ -21,7 +21,12 @@ This tutorial will teach you how to execute a process on only one machine and ho
 
 ## Execute on one process
 
-Certain code only needs to be run once on a given machine, such as printing a log statement or only displaying one progress bar on the local main process. You should use `accelerator.is_local_main_process` to indicate code that should only be executed once.
+Certain code only needs to be run once on a given machine, such as printing a log statement or only displaying one progress bar on the local main process. 
+
+<hfoptions id="local-execution">
+<hfoption id="statements">
+
+You should use `accelerator.is_local_main_process` to indicate code that should only be executed once.
 
 ```py
 from tqdm.auto import tqdm
@@ -31,17 +36,86 @@ progress_bar = tqdm(range(args.max_train_steps), disable=not accelerator.is_loca
 
 You could also wrap a statement with `accelerator.is_local_main_process`.
 
+> [!TIP]
+> Replace `print` statements with Accelerate's [`~Accelerator.print`] method to only print once per process!
+
 ```py
 if accelerator.is_local_main_process:
     accelerator.print("Accelerate is the best")
 ```
 
-To only execute code once across *all processes* regardless of the number of machines, use `accelerator.is_main_process`. For example, this is useful if you're uploading a final model to the Hub.
+</hfoption>
+<hfoption id="function">
+
+For a function that should only be executed once, use [`~Accelerator.on_local_main_process`].
+
+```py
+@accelerator.on_local_main_process
+def do_my_thing():
+    "Something done once per server"
+    do_thing_once_per_server()
+```
+
+</hfoption>
+</hfoptions>
+
+You could also direct Accelerate to execute code once across *all processes* regardless of the number of machines. This is useful if you're uploading a final model to the Hub.
+
+<hfoptions id="main-execution">
+<hfoption id="statement">
+
+You should use `accelerator.is_main_process` to indicate code that should only be executed once across all processes.
 
 ```py
 if accelerator.is_main_process:
     repo.push_to_hub()
 ```
+
+</hfoption>
+<hfoption id="function">
+
+For a function that should only be executed once across all processes, use [`~Accelerator.on_main_process`].
+
+```py
+@accelerator.on_main_process
+def do_my_thing():
+    "Something done once per server"
+    do_thing_once()
+```
+
+</hfoption>
+</hfoptions>
+
+## Execute on a specific process
+
+Accelerate can also help you execute functions that should only be executed on a specific process or a local process index.
+
+<hfoptions id="specific-execution">
+<hfoption id="specific process">
+
+Use the [`~Accelerator.on_process`] method and specify the process index to execute a function on.
+
+```py
+@accelerator.on_process(process_index=0)
+def do_my_thing():
+    "Something done on process index 0"
+    do_thing_on_index_zero()
+```
+
+</hfoption>
+<hfoption id="local process">
+
+Use the [`~Accelerator.on_local_process`] method and specify the local process index to execute a function on.
+
+```py
+@accelerator.on_local_process(local_process_idx=0)
+def do_my_thing():
+    "Something done on process index 0 on each server"
+    do_thing_on_index_zero_on_each_server()
+```
+
+</hfoption>
+</hfoptions>
 
 ## Defer execution
 
--- a/docs/source/basic_tutorials/execution.md
+++ b/docs/source/basic_tutorials/execution.md
@@ -21,7 +21,7 @@ This tutorial will teach you how to execute a process on only one machine and ho
 
 ## Execute on one process
 
-Certain code only needs to be run once on a given machine, such as printing a log statement or only displaying one progress bar on the local main process. 
+Certain code only needs to be run once on a given machine, such as printing a log statement or only displaying one progress bar on the local main process.
 
 <hfoptions id="local-execution">
 <hfoption id="statements">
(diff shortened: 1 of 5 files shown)


docs/source/quicktour.md
```diff
@@ -13,429 +13,168 @@ specific language governing permissions and limitations under the License.
...
-## Getting started with 🤗 Accelerate
+* a unified command line launching interface for distributed training scripts
+* a training library for adapting PyTorch training code to run on different distributed setups
+* Big Model Inference
 
-### Enable distributed training in your script 
+## Unified launch interface
 
-To use 🤗 Accelerate in your own training script, you have to modify four things:
+Accelerate automatically detects the training environment and adapts it to the distributed training framework you're using. This way, you can use the same command to launch your training script regardless of framework or hardware. From the command line, run the following command to answer a few questions about your training system.
```
muellerzr, line 28: Given how users are utilizing accelerate, I'm not 100% we want to do this outright. 

I think something along the lines of:

"While you can pass in the values to the command line explicitly (this is helpful in situations like SLURM), accelerate can automatically pick up configuration values for deepspeed, fsdp, distributed training, and more through a unified configuration file. To generate one, run `accelerate config` and follow the prompts..."

docs/source/quicktour.md
```diff
@@ -13,429 +13,168 @@ specific language governing permissions and limitations under the License.
...
-To further customize where and how states saved through [`~Accelerator.save_state`] the [`~utils.ProjectConfiguration`] class can be used. For example 
-if `automatic_checkpoint_naming` is enabled each saved checkpoint will be located then at `Accelerator.project_dir/checkpoints/checkpoint_{checkpoint_number}`.
-
-If you have registered any other stateful items to be stored through [`~Accelerator.register_for_checkpointing`] they will also be saved and/or loaded.
-
-<Tip>
+## Big Model Inference
 
-    Every object passed to [`~Accelerator.register_for_checkpointing`] must have a `load_state_dict` and `state_dict` function to be stored
+Accelerate's Big Model Inference has two main features to load and run inference with large models that typically don't fit into memory.
```
muellerzr, line 139: Maybe as a list instead here? Or some sort of natural break between `load` and `run`

docs/source/quicktour.md
```diff
@@ -13,429 +13,168 @@ specific language governing permissions and limitations under the License.
...
 
+### Empty weights initialization
 
-### Use gradient clipping
+The [`~accelerate.init_empty_weights`] context manager initializes models of any size by creating a *model skeleton* and moving and placing parameters each time they're created to PyTorch's [**meta**](https://pytorch.org/docs/main/meta.html) device. This way, not all weights are immediately loaded and only a small part of the model is loaded into memory at a time.
 
-If you are using gradient clipping in your script, you should replace the calls to
-`torch.nn.utils.clip_grad_norm_` or `torch.nn.utils.clip_grad_value_` with [`~Accelerator.clip_grad_norm_`]
-and [`~Accelerator.clip_grad_value_`] respectively.
+For example, you can load an empty [Mixtral-8x7B](https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1) model like this:
```
muellerzr, line 148: + something like "and have it take up only a fraction of memory compared to the model and weights loaded on regular CPU"

muellerzr (commented):
First pass: quicktour only

docs/source/basic_tutorials/execution.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+You could also wrap a statement with `accelerator.is_local_main_process`.
+
+> [!TIP]
+> Replace `print` statements with Accelerate's [`~Accelerator.print`] method to only print once per process!
+
+```py
+if accelerator.is_local_main_process:
+    accelerator.print("Accelerate is the best")
```
muellerzr, line 44: Should this just be a regular print statement?

docs/source/basic_tutorials/execution.md
```diff
@@ -0,0 +1,128 @@
+<!--Copyright 2024 The HuggingFace Team. All rights reserved.
```
muellerzr, line 1: I feel like this whole doc is just a repeat of https://moon-ci-docs.huggingface.co/docs/accelerate/pr_2456/en/concept_guides/deferring_execution no? I don't think it makes sense to have it here in the base tutorials.

docs/source/basic_tutorials/migration.md
```diff
@@ -124,6 +104,122 @@ for batch in training_dataloader:
...
+### Mixed precision
+
+Mixed precision accelerates training by using a lower precision data type like fp16 (half-precision) to calculate the gradients. For the best performance with Accelerate, the loss should be computed inside your model (like in Transformers models) because computations outside of the model are computed in full precision.
+
+Set the mixed precision type to use in the [`Accelerator`], and then use the [`~Accelerator.autocast`] context manager to automatically cast the values to the specified data type.
+
+```diff
++ accelerator = Accelerator(mixed_precision="fp16")
++ with accelerator.autocast():
+      loss = complex_loss_function(outputs, target):
```
muellerzr, line 145: We should be careful here and mention that this (`autocast`) is only needed if there is some *other* thing in mixed precision that needs to be done outside of just what happens during `.backward()` on the loss, as accelerate already does the scaling during `.backward()`. 

As automatic mixed precision is one of our advertisements :)

docs/source/basic_tutorials/migration.md
```diff
@@ -124,6 +104,122 @@ for batch in training_dataloader:
...
+```
+
+With mixed precision training, it may skip a few gradient updates at the beginning and during training because of the dynamic loss scaling strategy. This strategy reduces the loss scaling factor when gradients overflow during training to avoid it happening again at the next step. This means the learning rate scheduler may be updated even though there was no update, which can impact the training process when you have very little training data or if the first learning rate values of your scheduler are very important. You can prevent the learning rate scheduler from updating if the optimizer wasn't updated with the [`~Accelerator.optimizer_step_was_skipped`] method.
+
+```py
+if not accelerator.optimizer_step_was_skipped:
+    lr_scheduler.step()
+```
```
muellerzr, line 153: I'm not sure this should be here at all, as we handle this internally and can be a bit confusing for users? (as you have to go against the grain and set `step_scheduler_with_optimizer=False` to even reach this point)

docs/source/basic_tutorials/migration.md
```diff
@@ -124,6 +104,122 @@ for batch in training_dataloader:
...
+
+## Save and load
+
+Accelerate can also save and load a *model* once training is complete or you can also save the model and optimizer *state* which could be useful for resuming training.
+
+### Model
+
+Once all processes are complete, unwrap the model with the [`~Accelerator.unwrap_model`] method before saving it because the [`~Accelerator.prepare`] method may have placed your model inside a larger model for distributed training. If you don't unwrap the model, saving the model state dictionary also saves any potential extra layers from the larger model and you won't be able to load the weights back into your base model.
```
muellerzr, line 161:
```suggestion
Once all processes are complete, unwrap the model with the [`~Accelerator.unwrap_model`] method before saving it because the [`~Accelerator.prepare`] method will have wrapped your model into the proper interface for distributed training. If you don't unwrap the model, saving the model state dictionary also saves any potential extra layers from the larger model and you won't be able to load the weights back into your base model.
```

docs/source/quicktour.md
stevhliu: Clarified that the two features are more for "loading" rather than "running" the models, so I don't think we need to add a break between `load` and `run` anymore :)

docs/source/basic_tutorials/execution.md
stevhliu: Is this because we're already using `is_local_main_process` which makes it redundant because `accelerator.print` already only prints once per process? (lifted from [here](https://huggingface.co/docs/accelerate/package_reference/accelerator#printing) 🤭)

docs/source/basic_tutorials/execution.md
stevhliu: I think the "Deferring execution" guide shows examples of when to use `main_process_first` or `wait_for_everyone` such as downloading a dataset or saving/loading a state dict. This doc has a broader scope and introduces the different _execution scenarios_ you should consider (ie, do you want to execute on one process, on a specific process).

I think introducing these types of scenarios in the tutorials gives users a nice basic understanding of how to apply it to the more specific examples, such as the ones showcased in the Deferring execution doc. Let me know what you think! 🙂

docs/source/basic_tutorials/execution.md
muellerzr: Yes! :)

docs/source/basic_tutorials/migration.md
stevhliu: Ah ok, I'll take it out then to avoid confusion!

muellerzr (approved):
Thanks for the rework!

PR #2461, opened 2024-02-19
Fix the pytest version to be less than 8.0.1

BenjaminBossan:
# What does this PR do?

Fixes the pytest version.

We're getting errors such as this one:

https://github.com/huggingface/accelerate/actions/runs/7958684877/job/21725397566?pr=2450

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ extras["quality"] = [
     "ruff ~= 0.2.1",
 ]
 extras["docs"] = []
-extras["test_prod"] = ["pytest>=7.2.0,<8.0.0", "pytest-xdist", "pytest-subtests", "parameterized"]
+extras["test_prod"] = ["pytest>=7.2.0,<=8.0.0", "pytest-xdist", "pytest-subtests", "parameterized"]
 extras["test_dev"] = [
     "datasets",
     "evaluate",
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ extras["quality"] = [
     "ruff ~= 0.2.1",
 ]
 extras["docs"] = []
-extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests", "parameterized"]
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Not sure whom to ping, as Zach is off this week, maybe @ydshieh with reference to https://github.com/huggingface/transformers/pull/28758.

fxmarty (approved):
LGTM

setup.py
```diff
@@ -22,7 +22,7 @@
     "ruff ~= 0.2.1",
 ]
 extras["docs"] = []
-extras["test_prod"] = ["pytest", "pytest-xdist", "pytest-subtests", "parameterized"]
+extras["test_prod"] = ["pytest>=7.2.0,<8.0.0", "pytest-xdist", "pytest-subtests", "parameterized"]
```
fxmarty, line 25:
```suggestion
extras["test_prod"] = ["pytest>=7.2.0,<=8.0.0", "pytest-xdist", "pytest-subtests", "parameterized"]
```

`8.0.1` is the one to avoid

SunMarc (approved):
LGTM with @fxmarty comments. We shouldn't have issues with 8.0.0. 8.0.1 is the problematic one. In [transformers](https://github.com/huggingface/transformers/pull/29097), they pinned the pytest version for the tests to <8.0.1

BenjaminBossan:
In the [linked PR](https://github.com/huggingface/transformers/pull/28758), the version was pinned to `<8.0.0`, not sure why it's done partly including and partly excluding `8.0.0`. I can change this to `<=8.0.0` if we're fine with that.

fxmarty:
@BenjaminBossan I don't think it really matters, let's leave it like is

SunMarc:
> In the https://github.com/huggingface/transformers/pull/28758, the version was pinned to <8.0.0, not sure why it's done partly including and partly excluding 8.0.0

In transformers, they had other issues with 8.0.0 for some tests. I will merge this since some checks are failing because without this specific PR.

PR #2466, opened 2024-02-20
[docs] Divide training and inference

stevhliu:
Separate training and inference docs for easier navigation.


muellerzr (commented):
Is there no way to have 2x-nested subsections in a TOC? 😢

muellerzr (approved):
Rockstar! Wow! 🤩

PR #2473, opened 2024-02-21
[FIX] allow `Accelerator` to detect distributed type from the "LOCAL_RANK" env variable for XPU 

faaany:
## What does this PR do?
Inspired by the newly added test `test_fsdp.py`, I found that Accelerator's default distributed type is NO on XPU with the following approach, while on NV GPU it is FSDP: 
```bash
$ export ACCELERATE_USE_FSDP="true"
$ export MASTER_ADDR="localhost"
$ export MASTER_PORT="10999"
$ export RANK="0"
$ export LOCAL_RANK="0"
$ export WORLD_SIZE="1"
$ python
>>> from accelerate.accelerator import Accelerator
>>> accelerator = Accelerator()
>>> accelerator.state.distributed_type
<DistributedType.NO: 'NO'>
## on NV GPU
>>> accelerator.state.distributed_type
<DistributedType.FSDP: 'FSDP'>
```
The reason lies at [this ](https://github.com/faaany/accelerate/blob/main/src/accelerate/state.py#L254) line: currently, MUTLI_XPU/MULTI_CPU is detected by `WORLD_SIZE`, while on NPU or GPU, it is detected by `LOCAL_RANK` as shown [here](https://github.com/faaany/accelerate/blob/main/src/accelerate/state.py#L220). 

To be safe, I add an "or" condition to fix this issue instead of doing refactoring for the entire code block. After the fix, the behavior on XPU aligns with that on GPU and the `test_mixed_precision` unit test in the `test_fsdp.py` can run through. 

Pls have a review, thx! @muellerzr

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -251,7 +251,10 @@ class PartialState:
                 if self.device is None:
                     self.device = torch.device("npu", self.local_process_index)
                 torch.npu.set_device(self.device)
-            elif get_int_from_env(["PMI_SIZE", "OMPI_COMM_WORLD_SIZE", "MV2_COMM_WORLD_SIZE", "WORLD_SIZE"], 1) > 1 or int(os.environ.get("LOCAL_RANK", -1)) != -1:
+            elif (
+                get_int_from_env(["PMI_SIZE", "OMPI_COMM_WORLD_SIZE", "MV2_COMM_WORLD_SIZE", "WORLD_SIZE"], 1) > 1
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
LGTM !

muellerzr:
@faaany can you rebase from main for us? This is to help ensure the tests are fine/passing :)

faaany:
@muellerzr  pls help merge, thx! (heart)

PR #2475, opened 2024-02-21
Fix wrong `is_namedtuple` implementation

fxmarty:
As per title.

The current implementation does not work in case of nested inheritance. Example:

```python
from torch import Tensor
import torch
from typing import NamedTuple, Optional
from collections import namedtuple


class QuantTensorBase(NamedTuple):
    value: Tensor
    scale: Optional[Tensor]
    zero_point: Optional[Tensor]
    bit_width: Optional[Tensor]
    signed_t: Optional[Tensor]
    training_t: Optional[Tensor]

class Second(QuantTensorBase):
    pass

a = QuantTensorBase(torch.tensor(1.), None, None, None, None, None)
b = Second(torch.tensor(1.), None, None, None, None, None)

point = namedtuple('Point', ['x', 'y'])
p = point(11, y=22)

def isnamedtupleinstance(x):
    t = type(x)
    b = t.__bases__
    print("b", b)
    if len(b) != 1 or b[0] != tuple:
        print("here")
        return False
    f = getattr(t, '_fields', None)
    if not isinstance(f, tuple):
        print("there")
        return False
    return all(type(n)==str for n in f)

print("-----")
print(isnamedtupleinstance(p))
print("-----")
print(isnamedtupleinstance(a))
print("-----")
print(isnamedtupleinstance(b))
```

giving

```
-----
b (<class 'tuple'>,)
True
-----
b (<class 'tuple'>,)
True
-----
b (<class '__main__.QuantTensorBase'>,)
here
False
```

Using instead https://stackoverflow.com/questions/2166818/how-to-check-if-an-object-is-an-instance-of-a-namedtuple/62692640#62692640

cc @Giuseppe5 @nickfraser


fxmarty:
pytest is bugged

BenjaminBossan (approved):
Thanks for fixing the function.

It looks like the old code corresponds to [this old SO answer](https://stackoverflow.com/a/2166841) whereas the new code corresponds to a [more recent SO answer](https://stackoverflow.com/a/62692640). It always pays to check more than the first SO answer :)

Since you provide code to check this, it could be added as a unit test, WDYT?

> pytest is bugged

Yes, it's unrelated, should be fixed by #2461.

fxmarty:
@BenjaminBossan I think it is a bit overly testing to test this, but I'll add a test.

fxmarty:
@BenjaminBossan added a test

SunMarc (approved):
Thanks for the fix !

PR #2483, opened 2024-02-23
Free mps memory

SunMarc:
# What does this PR do ?
This PR updates the `release_memory` function to also free mps memory if we have a mps backend. 

Thx to [Datamance](https://github.com/Datamance) for letting us know ! See related [issue](https://github.com/huggingface/transformers/issues/27181#issuecomment-1958902144).

### DIFF
--- a/src/accelerate/utils/memory.py
+++ b/src/accelerate/utils/memory.py
@@ -23,7 +23,7 @@ import inspect
 
 import torch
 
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Nice! This seems quite handy indeed :) 

Thanks for the improvement!

PR #2491, opened 2024-02-26
Fix CI tests due to pathlib issues

muellerzr:
# What does this PR do?

`subprocess` requires strings, it cannot use paths. Since the helper introduced in https://github.com/huggingface/accelerate/pull/2446 doesn't handle anything externally-facing and is just for testing, and a *vast majority of our tests* require `subprocess`, this PR makes it return a string instead and casts it back to a Path object if applicable. 

Fixes failures on main (and soon-to-be nightlies)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @akx

### DIFF
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -48,7 +48,7 @@ def test_command(args):
     if args.config_file is None:
         test_args = [script_name]
     else:
-        test_args = (f"--config_file={args.config_file} {script_name}").split()
+        test_args = f"--config_file={args.config_file} {script_name}".split()
 
     cmd = ["accelerate-launch"] + test_args
     result = execute_subprocess_async(cmd, env=os.environ.copy())
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -17,8 +17,7 @@
 import argparse
 import os
 
-from accelerate.test_utils import execute_subprocess_async
-from accelerate.utils.other import path_in_accelerate_package
+from accelerate.test_utils import execute_subprocess_async, path_in_accelerate_package
 
 
 def test_command_parser(subparsers=None):
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -46,7 +46,7 @@ def test_command(args):
     script_name = path_in_accelerate_package("test_utils", "scripts", "test_script.py")
 
     if args.config_file is None:
-        test_args = script_name
+        test_args = [script_name]
     else:
         test_args = (f"--config_file={args.config_file} {script_name}").split()
 
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -48,9 +48,9 @@ def test_command(args):
     if args.config_file is None:
         test_args = script_name
     else:
-        test_args = f"--config_file={args.config_file} {script_name}"
+        test_args = (f"--config_file={args.config_file} {script_name}").split()
 
-    cmd = ["accelerate-launch"] + test_args.split()
+    cmd = ["accelerate-launch"] + test_args
     result = execute_subprocess_async(cmd, env=os.environ.copy())
     if result.returncode == 0:
         print("Test is a success! You are ready for your distributed training!")
--- a/src/accelerate/commands/test.py
+++ b/src/accelerate/commands/test.py
@@ -47,7 +47,7 @@ def test_command(args):
     script_name = path_in_accelerate_package("test_utils", "scripts", "test_script.py")
 
     if args.config_file is None:
-        test_args = str(script_name)
+        test_args = script_name
     else:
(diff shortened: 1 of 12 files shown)


akx:
How did tests pass in #2446 to begin with then? 🤔

muellerzr:
@akx because these jobs require distributed workers. The PR CI does not run anything on GPUs because this can be quite expensive and costly.

akx:
That's unfortunate. Sounds like coverage checking should be enabled on test code so it'd be clear when changes to it are not actually being run...

muellerzr:
We have the `@require_{HARDWARE}` flags which state what needs accelerators and what does not. Not sure how coverage testing would help in this case? As these tests require certain acceleration backends to be available to even be ran (e.g. two GPUs, TPUs, NPUs, etc). A single coverage test could never be ran and actually account for all of these because you just don't have a system with TPUs and GPUs

muellerzr:
(Not saying coverage isn't something we should look into, we've discussed this quite a lot. However the nuances of it given all of our different testing hardware makes it difficult to know % just in a PR since they don't have access to the hardware to run those)

src/accelerate/utils/other.py
```diff
@@ -350,16 +350,16 @@ def _getattr(obj, attr):
     return reduce(_getattr, [obj] + attr.split("."))
 
 
-def path_in_accelerate_package(*components: str) -> pathlib.Path:
+def path_in_accelerate_package(*components: str) -> str:
```
akx, line 353: I do think this function should return a Path, and the call sites where a str is needed should cast there...

BenjaminBossan: I agree. It's a bit strange that a str is returned here and below we convert to `Path` again. I wonder if we could instead adjust `execute_subprocess_async` to convert everything to str first? Not sure if this would cover all bases.

muellerzr: Sounds good, I'll look at what we can do!

muellerzr: Reworked it to still return a `Path` and found a number of stragglers we can refactor

akx:
@muellerzr Just a heads-up: your commit identity is a bit messed up in some of these commits:

```
commit d260bd29c82158e8e0056eb9f2caab37a6fc47b9
Author: [[ -z $EMAIL ]] && read -e -p "Enter your email (for git configuration): " EMAIL <muellerzr@gmail.com>
Date:   Mon Feb 26 15:43:35 2024 -0500
```

src/accelerate/test_utils/testing.py
```diff
@@ -488,6 +491,10 @@ def tee(line, sink, pipe, label=""):
 
 
 def execute_subprocess_async(cmd, env=None, stdin=None, timeout=180, quiet=False, echo=True) -> _RunOutput:
+    # Cast everything in `cmd` to a string
+    for i, c in enumerate(cmd):
+        if isinstance(c, Path):
+            cmd[i] = str(c)
```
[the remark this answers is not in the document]
muellerzr: It must always be mutable here, which I'm open to adding a type hint for but this is known and a repeated pattern. 

And while true, is a must to keep things clean like you would like, without a blanket convert across everything.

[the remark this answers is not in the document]
akx: I'm not quite sure why everything couldn't be cast though? (Also, I guess `run_command` should use the same pattern..?)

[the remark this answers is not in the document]
muellerzr: What if a user decided to try passing in an `Accelerator` object instead? :)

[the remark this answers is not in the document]
akx: Well, I'd probably go with a full-on

```python
def _check_subprocess_args(cmd) -> list:
   for arg in cmd:
       if not isinstance(arg, (str, pathlib.Path)):
           raise TypeError(f"Subprocess argument must be a string or path, not {arg!r}")
       yield str(arg)
```
to get a clear early `TypeError` from the testing utility code instead of `subprocess.*` raising on a non-string argument later on :)

src/accelerate/commands/test.py
```diff
@@ -47,11 +46,11 @@ def test_command(args):
     script_name = path_in_accelerate_package("test_utils", "scripts", "test_script.py")
 
     if args.config_file is None:
-        test_args = str(script_name)
+        test_args = [script_name]
     else:
-        test_args = f"--config_file={args.config_file} {script_name}"
+        test_args = (f"--config_file={args.config_file} {script_name}").split()
```
BenjaminBossan, line 51: Parentheses shouldn't be necessary here but also don't hurt.

BenjaminBossan (approved):
LGTM, thanks for fixing the issue. I haven't done a check if each and every instance of str path operation has been updated, but I don't think that is a must have.

PR #2493, opened 2024-02-26
Launch mpirun from accelerate launch for multi-CPU training

### LINKED ISSUE #2492: [Feature request]: Launch mpirun from accelerate launch for multi-CPU training
## Motivation
Currently, multi-CPU training can be launched through accelerate by going through `accelerate config` and `mpirun -f hostfile -n 16 -ppn 4 accelerate launch ...` as described in the [Intel Extension for PyTorch usage guide](https://github.com/huggingface/accelerate/blob/main/docs/source/usage_guides/ipex.md#how-it-works-for-training-optimization-in-cpu). This method results in a lot of redundant information being provided by the user. For example, `acceleate config` asks how many machines will be used, the IP of the primary machine, and the total number of processes to run. Then, in the next step to launch the job, the user needs to set the `MASTER_ADDR` environment variable with the IP address of the main machine, and provide arguments to `mpirun` with the number of processes and number of processes per node (all information that was already given through `accelerate config`).  The launch command ends up looking something like this:
```
export CCL_WORKER_COUNT=1
export MASTER_ADDR=xxx.xxx.xxx.xxx
mpirun -f hostfile -n 16 -ppn 4 accelerate launch examples/nlp_example.py
```

In contrast, when running multi-GPU multi-node jobs with torchrun through Accelerate, `accelerate launch` uses the information in the config file to do the torch distributed run, so the launch command is simply `accelerate launch ./nlp_example.py` (on every machine). 

## Feature Description
To bring the multi-CPU mpirun user experience up to par with the multi-GPU multi-node torch distributed launch, we can have `accelerate launch` use the parameters in the config file to construct and launch the `mpirun ...` command. In order to do this, there are a few pieces of information that would need to be added to the config:
* Enabling mpirun through accelerate launch (if we wanted to provide the user with a choice of if they want accelerate launch to run mpirun, or if they want to call mpirun themselves)
* hostfile path
* CCL worker count

The updated `acceleate config` prompts for multi-CPU would look something like:
<pre><code>$ accelerate config
-----------------------------------------------------------------------------------------------------------------------------------------------------------
In which compute environment are you running?
This machine
-----------------------------------------------------------------------------------------------------------------------------------------------------------
Which type of machine are you using?
multi-CPU
How many different machines will you use (use more than 1 for multi-node training)? [1]: 4
-----------------------------------------------------------------------------------------------------------------------------------------------------------
What is the rank of this machine?
0
What is the IP address of the machine that will host the main process? 36.112.23.24
What is the port you will use to communicate with the main process? 29500
Are all the machines on the same local network? Answer `no` if nodes are on the cloud and/or on different network hosts [YES/no]: yes
Do you want to use Intel PyTorch Extension (IPEX) to speed up training on CPU? [yes/NO]:yes
<b>Do you want accelerate to launch mpirun? [yes/NO]: yes ⬅️ new config question </b>
<b>Please enter the path to the hostfile to use with mpirun [~/hostfile]: ~/hostfile ⬅️ new config question </b>
<b>Enter the number of oneCCL worker threads [1]: 1 ⬅️ new config question</b>
Do you wish to optimize your script with torch dynamo?[yes/NO]:NO
How many processes should be used for distributed training? [1]:16
-----------------------------------------------------------------------------------------------------------------------------------------------------------
Do you wish to use FP16 or BF16 (mixed precision)?
bf16
</code></pre>

The resulting config file would look something like:
<pre><code>compute_environment: LOCAL_MACHINE
distributed_type: MULTI_CPU
downcast_bf16: 'no'
ipex_config:
  ipex: true
machine_rank: 0
main_process_ip: 36.112.23.24
main_process_port: 29500
main_training_function: main
mixed_precision: bf16
<b>mpirun_config: ⬅️ new config </b>
<b>  mpirun_ccl: '1' ⬅️ new config </b>
<b>  mpirun_hostfile: /home/user/hostfile ⬅️ new config </b>
num_machines: 4
num_processes: 16
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: true
</code></pre>

Finally, the launch command simply becomes a single command on the main machine:
```
```

A caveat to consider is that `mpirun` has many other options that are not represented by parameters in the accelerate config file. It would be unrealistic to try to abstract out of those parameters, so only the minimum/basic set of parameters were considered. The main intent of this feature is to help out users who are less familiar with `mpirun` and match the usability of the multi-node multi-GPU flow that uses `torch.distributed.run`. If someone needs a more complex `mpirun` usage, the previous method of calling `mpirun` themselves can still be available.

I will follow this up with a PR.

dmsuehir:
# What does this PR do?

This PR updates multi-CPU training from the Accelerate CLI to launch `mpirun` under-the-hood. In order to accommodate this, a few more config options were added to get the path to the user's hostfile and the number of CCL workers. The purpose of this update is to make the multi-CPU experience more consistent with the multi-node multi-GPU user experience (which uses `torch.distributed.run` under-the-hood) and make it easier for users who are less familiar with the `mpirun` parameters.

Fixes #2492

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr


sywangyi:
@yao-matrix @sywangyi

sywangyi (approved):

sywangyi:
@muellerzr please help review this PR, with the PR we could enable multi-node CPU training only "accelerate launch" in master node.

src/accelerate/commands/config/cluster.py
```diff
@@ -486,6 +502,8 @@ def get_cluster_input():
         machine_type = str(distributed_type).split(".")[1].replace("MULTI_", "")
         if machine_type == "TPU":
             machine_type += " cores"
+        elif machine_type == "CPU":
+            machine_type = "processes"
```
muellerzr, line 506:
```suggestion
            machine_type = " processes"
```

src/accelerate/utils/launch.py
```diff
@@ -46,13 +48,49 @@ def _filter_args(args, parser, default_args=[]):
...
+    """
+    Determines the executable and argument names for mpirun, based on the type of install. The supported MPI programs
+    are: OpenMPI, Intel MPI, or MVAPICH.
+
+    Returns: Program name and arg names for hostfile, num processes, and processes per node
+    """
+    # Find the MPI program name
+    mpi_apps = [x for x in ["mpirun", "mpiexec"] if which(x)]
+
+    if not mpi_apps:
```
muellerzr, line 61:
```suggestion
    if len(mpi_apps) == 0:
```

src/accelerate/utils/launch.py
```diff
@@ -46,13 +48,49 @@ def _filter_args(args, parser, default_args=[]):
...
     """
     Prepares and returns the command list and an environment with the correct simple launcher environment variables.
     """
     cmd = []
     if args.no_python and args.module:
         raise ValueError("--module and --no_python cannot be used together")
+
+    mpirun_hostfile = getattr(args, "mpirun_hostfile", None)
+
+    if mpirun_hostfile:
```
muellerzr, line 85: `args` will always have `mpirun_hostfile`, so let's just use it
```suggestion
    if args.mpirun_hostfile is not None:
```

src/accelerate/utils/launch.py
```diff
@@ -74,6 +112,9 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> Tuple[List[str]
     if args.num_machines > 1:
         current_env["MASTER_ADDR"] = args.main_process_ip
         current_env["MASTER_PORT"] = str(args.main_process_port)
+
+        if mpirun_hostfile:
```
muellerzr, line 116: Same here
```suggestion
        if args.mpirun_hostfile is not None:
```

tests/test_cli.py
```diff
@@ -67,7 +67,7 @@ def test_no_config(self):
 
     def test_config_compatibility(self):
         for config in sorted(self.test_config_path.glob("**/*.yaml")):
-            if "invalid" in str(config):
+            if "invalid" in str(config) or "mpi" in str(config):
```
muellerzr, line 70: This should not be skipped, it is expected that users can have old configs that do not have `mpi` and code will still run (just we will not use `mpi`). If this test fails right now without it we have issues and need to fix the defaults.

tests/test_configs/mpi_multicpu.yaml
```diff
@@ -0,0 +1,22 @@
+compute_environment: LOCAL_MACHINE
```
muellerzr, line 1: Let's tag a version name with this filename please (we can specify mpi as well)

E.g. 0_28_0_mpi.yaml

tests/test_cpu.py
```diff
@@ -12,16 +12,59 @@
...
+from accelerate.commands.launch import _validate_launch_command, launch_command_parser
 from accelerate.test_utils import require_cpu, test_ops, test_script
+from accelerate.utils.launch import prepare_simple_launcher_cmd_env
 
 
 @require_cpu
 class MultiCPUTester(unittest.TestCase):
+    mod_file = inspect.getfile(accelerate.test_utils)
+    test_file_path = os.path.sep.join(mod_file.split(os.path.sep)[:-1] + ["scripts", "test_cli.py"])
+    test_config_path = Path("tests/test_configs")
```
muellerzr, line 33: Please use `path_in_accelerate` here instead as seen here: https://github.com/huggingface/accelerate/blob/main/tests/test_multigpu.py#L39

tests/test_cpu.py
```diff
@@ -12,16 +12,59 @@
...
+
     def test_cpu(self):
         debug_launcher(test_script.main)
 
     def test_ops(self):
         debug_launcher(test_ops.main)
+
+    def test_mpi_multicpu_config_cmd(self):
```
muellerzr, line 41: This checks the command, not the actual run of it which is what `test_cpu` is meant to do. Please include a decorator and actually run the test script please here, or move this test to `test_cli.py`

muellerzr (commented):
Thanks! This is a good start, I've made my first pass on reviews here.

src/accelerate/commands/config/cluster.py
dmsuehir: If I make this change, there ends up being an extra space in the prompt like:
`How many  processes should be used for distributed training?`

tests/test_cli.py
dmsuehir: This is skipping testing of test configs that have "mpi" in the file name. The problem that I ran into with this was the GitHub actions run saying that `mpirun` is not installed.

muellerzr: Yes. You need to make a test skipper similar to require_GPU in test_utils/testing.py to decorate it with.

muellerzr: Makes sense.

dmsuehir:
@muellerzr Thanks for the review. I think I have addressed all of your comments.

tests/test_cli.py
```diff
@@ -98,6 +102,37 @@ def test_notebook_launcher(self):
...
+            with patch("accelerate.utils.launch.subprocess.check_output", return_value=b"Intel MPI"):
+                cmd, current_env = prepare_simple_launcher_cmd_env(args)
+
+        # Verify the mpirun command args
+        expected_mpirun_cmd = ["mpirun", "-f", "/home/user/hostfile", "-ppn", "4", "-n", "16"]
+        self.assertGreater(len(cmd), len(expected_mpirun_cmd))
+        generated_mpirun_cmd = cmd[0 : len(expected_mpirun_cmd)]
+        self.assertEqual(expected_mpirun_cmd, generated_mpirun_cmd)
+
+        # Verify that the python script and arg
```
BenjaminBossan, line 130: comment missing a verb?

tests/test_cli.py
```diff
@@ -98,6 +102,37 @@ def test_notebook_launcher(self):
...
             run_command(cmd, env=os.environ.copy())
 
+    def test_mpi_multicpu_config_cmd(self):
+        """
+        Parses a launch command with a test file and the 0_28_0_mpi.yaml config. Tests getting the command and
+        environment vars and verifies the mpirun command arg values.
+        """
+        mpi_config_path = str(self.test_config_path / "0_28_0_mpi.yaml")
+        test_file_arg = "--cpu"
+        sys.argv = ["accelerate", str(self.test_file_path), test_file_arg]
```
BenjaminBossan, line 112: Not sure if it is a good idea to overwrite `sys.argv`. Could this be mocked instead? My concern is that it could have side effects on other tests, as it seems that this is not cleaned up automatically:

```python
class TestFoo:
    def test_foo(self):
        assert sys.argv != ["foobar"]  # passes

    def test_bar(self):
        sys.argv = ["foobar"]
        assert True

    def test_baz(self):
        assert sys.argv != ["foobar"]  # fails
```

Also not sure if this could not have some weird interactions with the test runner, since the runner also reads from `sys.argv`. Probably only once when it launches, but who knows, it would be hard to debug if this caused an error.

src/accelerate/commands/launch.py
```diff
@@ -625,6 +625,21 @@ def launch_command_parser(subparsers=None):
         ),
     )
 
+    # MPI arguments
+    mpirun_args = parser.add_argument_group("MPI Arguments", "Arguments related to mpirun for Multi-CPU")
+    mpirun_args.add_argument(
+        "--mpirun_hostfile",
+        type=str,
+        default=None,
+        help="Location for a hostfile for using Accelerate to launch a multi-CPU training job with mpirun.",
```
BenjaminBossan, line 634: IIUC, depending on the MPI implementation being used, the arguments can have different names. Would it be worth documenting that?

BenjaminBossan (changes requested):
The PR looks great, well implemented, documented, and even tested. I don't know much about MPI itself but trust that this part is all correct.

Apart from some nits, I have a concern with how the test writes to `sys.argv` which I think should be addressed. Please take a look.

dmsuehir:
@BenjaminBossan I pushed a fix. Thanks!

BenjaminBossan (approved):
Thanks a lot for this addition, LGTM.

muellerzr (approved):
Thanks for your work on this!

PR #2505, opened 2024-02-28
add custom dtype INT2

SunMarc:
# What does this PR do
This PR adds the custom dtype INT2 what will be useful to calculate the memory required for 2bits model for quanto.


muellerzr (approved):
Nice! Soon we may need one-bit 👀

PR #2506, opened 2024-02-28
Check if hub down

muellerzr:
# What does this PR do?

Adds a decorator to the test utils that will skip tests if we find the Hub is down automatically to avoid red CI in the case of an outtage

### DIFF
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -15,18 +15,18 @@
 import asyncio
 import inspect
 import os
-import requests
 import shutil
 import subprocess
 import sys
 import tempfile
 import unittest
 from contextlib import contextmanager
-from functools import partial, lru_cache
+from functools import lru_cache, partial
 from pathlib import Path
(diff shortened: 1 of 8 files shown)


SunMarc (approved):
Good idea ! 💯 LGTM. Maybe we can even add a test to check if the hub is down 👀

src/accelerate/test_utils/testing.py
```diff
@@ -73,6 +74,21 @@ def get_backend():
 torch_device, device_count, memory_allocated_func = get_backend()
 
 
+@lru_cache
+def is_hub_online():
+    """
+    Checks if the huggingface_hub is online
+    """
+    response = requests.get("https://huggingface.co/api")
+    if response.status_code != 200:
```
BenjaminBossan, line 83: Technically, any 2** status code should be fine (though maybe 200 can be returned).

Also, I wonder if the URL is correct. I get 404 at this very moment but can still successfully download models from the hub. Is there maybe an official health check that could be called instead?

PR #2511, opened 2024-02-29
Fix edge case in infer_auto_device_map when dealing with buffers

SunMarc:
# What does this PR do ?
This PR fixes the case where we have buffers in the `modules_to_treat` and we are at the point where we need to choose between split/not split the buffer. 

Fixes an issue in our CI with llama because of the `causal_mask` buffer cc @ArthurZucker


muellerzr (approved):
Thanks! Good work :)

PR #2524, opened 2024-03-04
Don't manage `PYTORCH_NVML_BASED_CUDA_CHECK` when calling `accelerate.utils.imports.is_cuda_available()`

luiscape:
# What does this PR do?

`PYTORCH_NVML_BASED_CUDA_CHECK=1` will tell PyTorch to use an NVML-based check when determining how many devices are available. That's useful for preventing CUDA initialization when calling `torch.cuda.is_available()`. 

`accelerate.utils.imports.is_cuda_available()` will set that env var and delete it afterwards. This is bad because that env var may be available, so deleting it after the function is called can lead to bugs. We ([Modal](https://modal.com/)) set that env var in some workloads. Calling this function breaks some user programs.

We can restore that env var its previous state after calling this function but I think that it is better to, instead of manipulating that env var, call the torch utility `_device_count_nvml` directly. What do you think?

## Who can review?

@muellerzr tagging you because git blame suggests you were the function author.

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -98,7 +98,7 @@ def is_cuda_available():
     Checks if `cuda` is available via an `nvml-based` check which won't trigger the drivers and leave cuda
     uninitialized.
     """
-    pytorch_nvml_based_cuda_check_previous_value = os.getenv("PYTORCH_NVML_BASED_CUDA_CHECK")
+    pytorch_nvml_based_cuda_check_previous_value = os.environ.get("PYTORCH_NVML_BASED_CUDA_CHECK")
     try:
         os.environ["PYTORCH_NVML_BASED_CUDA_CHECK"] = str(1)
         available = torch.cuda.is_available()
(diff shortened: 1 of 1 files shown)


muellerzr (commented):
Thanks for pointing out the issue with it! That's certainly not good!

Hesitant to use a private torch function, but it's been there for 2 years or so so it seems to be in there permanently. 

I'd rather a solution like checking if the os var is set, if so use it otherwise then we set it and delete it after (if we override). 

Does that seem like a more reasonable path you're willing to implement? 🤗

luiscape:
> I'd rather a solution like checking if the os var is set, if so use it otherwise then we set it and delete it after (if we override).

Sounds good. Will implement now.

luiscape:
@muellerzr just implemented the env var-based solution. I'm restoring a previous value if it exits, otherwise preserves the existing logic.

muellerzr (approved):
Nice! Much better solution here imo, thanks for doing it!

muellerzr:
@luiscape to get the failing quality test to pass, please do:

```bash
pip install -e .[quality]
make style; make quality
```

Issue #2241, opened 2023-12-13
[feature request] accelerate launcher: add numa affiinities control


stas00:
### Feature request

As explained here https://github.com/pytorch/pytorch/issues/115305 when using 2-cpu nodes it's important to get the NUMA affinities right to avoid cross NUMA node talk

As torchrun currently doesn't support it a workaround was posted here https://github.com/pytorch/pytorch/issues/115305#issuecomment-1845957682 and it includes a torchrun flag `--no-python` which the accelerate launcher doesn't have.

So any suggestions to how I could use this script with `accelerate`?

For simplicity here is the solution for torchrun:

```
trampoline.sh

#!/usr/bin/bash

# Query the bus ID for device LOCAL_RANK
BUS_ID=$(nvidia-smi --query-gpu=pci.bus_id -i $LOCAL_RANK --format=csv,noheader)
BUS_ID=${BUS_ID,,}

# Find the numa node for device LOCAL_RANK
NODE=$(cat /sys/bus/pci/devices/${BUS_ID:4}/numa_node)

echo "Starting local rank $RANK on numa node $NODE"
numactl --cpunodebind=$NODE --membind=$NODE "$@"
```

```
torchrun --nproc_per_node=8 --monitor-interval=1 --no-python ./trampoline.sh python3 -c "print('hello')"
```

update: I shared @yifuwang's workaround at https://twitter.com/StasBekman/status/1734724979496018215

If pynvml dependency is OK someone posted a python solution:
https://github.com/NVIDIA/DeepLearningExamples/blob/9dd9fcb98f56187e49c5ee280cf8dbd530dde57b/TensorFlow2/LanguageModeling/BERT/gpu_affinity.py
so that would probably be easier to integrate into the launcher.

Thanks.

stas00:
keepalive

stas00:
keepalive

muellerzr:
Will be looking into adding this week finally :)

stas00:
that's exciting, Zach - thank you!

muellerzr:
@stas00 actually wait, can't we just do this?

```bash
torchrun --nproc_per_node=8 --monitor-interval=1 --no-python ./trampoline.sh python3 -c "print('hello')"
```
Into:
```bash
accelerate launch --no_python --multi_gpu --num_processes 8 --monitor_interval=1 ./trampoline.sh python3 -c "print('hello')"
```
(I just pulled all this from `accelerate launch -h`)

Or am I missing something?...

[...]

PR #2525, opened 2024-03-04
Enable using dash or underscore in CLI

muellerzr:
# What does this PR do?

This PR adds a pinch of magic for users that are used to doing `--a-b` with arguments that pass-through to things such as `torchrun`, accelerate can process those if we have them defined ourselves already.

E.g. a user can do:
```bash
accelerate launch --no_python --multi_gpu --num_processes 8 --monitor_interval=1 ./trampoline.sh python3 -c "print('hello')"
```
**OR**:
```bash
accelerate launch --no-python --multi-gpu --num-processes 8 --monitor-interval=1 ./trampoline.sh python3 -c "print('hello')"
```

Fixes https://github.com/huggingface/accelerate/issues/2241

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan, @stas00 (for your future sanity 😄 )

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -128,10 +128,13 @@ class _CustomHelpAction(argparse._HelpAction):
 
 
 def launch_command_parser(subparsers=None):
+    description = "Launch a python script in a distributed scenario. Arguments can be passed in with either hyphens (`a-b`) or underscores (`a_b`)"
     if subparsers is not None:
-        parser = subparsers.add_parser("launch", add_help=False, allow_abbrev=False)
+        parser = subparsers.add_parser("launch", description=description, add_help=False, allow_abbrev=False)
     else:
-        parser = ArgumentParserWithDashSupport("Accelerate launch command", add_help=False, allow_abbrev=False)
+        parser = ArgumentParserWithDashSupport(
+            "Accelerate launch command", description=description, add_help=False, allow_abbrev=False
+        )
 
     parser.register("action", "help", _CustomHelpAction)
     parser.add_argument("-h", "--help", action="help", help="Show this help message and exit.")
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -28,6 +28,7 @@ import torch
 from accelerate.commands.config import default_config_file, load_config_from_file
 from accelerate.commands.config.config_args import SageMakerConfig
 from accelerate.commands.config.config_utils import DYNAMO_BACKENDS
+from accelerate.commands.utils import ArgumentParserWithDashSupport
 from accelerate.state import get_int_from_env
 from accelerate.utils import (
     ComputeEnvironment,
@@ -130,7 +131,7 @@ def launch_command_parser(subparsers=None):
     if subparsers is not None:
         parser = subparsers.add_parser("launch", add_help=False, allow_abbrev=False)
     else:
-        parser = argparse.ArgumentParser("Accelerate launch command", add_help=False, allow_abbrev=False)
+        parser = ArgumentParserWithDashSupport("Accelerate launch command", add_help=False, allow_abbrev=False)
 
     parser.register("action", "help", _CustomHelpAction)
     parser.add_argument("-h", "--help", action="help", help="Show this help message and exit.")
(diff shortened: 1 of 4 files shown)


stas00:
Zach, in the OP, the 2 lines of code are identical - a copy-n-paste mistake?

muellerzr:
OP was posted before I did this 😅 I'll go update it!

muellerzr:
Actually, was weird phrasing on my end. Edited it to be more succinct :)

stas00:
I meant this:
```
accelerate launch --no_python --multi_gpu --num_processes 8 --monitor_interval=1 ./trampoline.sh python3 -c "print('hello')"
accelerate launch --no-python --multi-gpu --num-processes 8 --monitor-interval=1 ./trampoline.sh python3 -c "print('hello')"
```

stas00:
oh, wait, they aren't identical ;) it's the `multi_gpu` vs `multi-gpu`

src/accelerate/commands/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        if len(option_tuples) > 1:
+            options = ", ".join([option_string for _, option_string, _ in option_tuples])
+            self.error(f"ambiguous option: {arg_string} could match {options}")
+
+        # If exactly one match, return it
+        elif len(option_tuples) == 1:
+            (option_tuple,) = option_tuples
+            return option_tuple
+
+        # If not found, but looks like a negative number, probably posisional
```
BenjaminBossan, line 45:
```suggestion
        # If not found, but looks like a negative number, probably positional
```

tests/test_cli.py
```diff
@@ -99,6 +100,50 @@ def test_notebook_launcher(self):
             run_command(cmd, env=os.environ.copy())
 
 
+class LaunchArgTester(unittest.TestCase):
+    """
+    Test cases revolving around the CLI wrappers
+    """
+
+    def test_hiphen(self):
```
BenjaminBossan, line 108:
```suggestion
    def test_hyphen(self):
```

src/accelerate/commands/launch.py
```diff
@@ -127,10 +128,13 @@ def __call__(self, parser, namespace, values, option_string=None):
 
 
 def launch_command_parser(subparsers=None):
+    description = "Launch a python script in a distributed scenario. Arguments can be passed in with either hyphens (`a-b`) or underscores (`a_b`)"
```
BenjaminBossan, line 131: I think the doc would be clearer with a real example, for `num-processes` instead of `a-b`.

BenjaminBossan (commented):
I'm +/- in this PR. I understand the goal to enable both ways of writing this, but I'm uncertain if the extra complexity is worth it.

We override two private methods, which in turn use more private methods. Are we sure that they work in all edge cases with all Python versions? Is the new code 100% covered by tests? If there is an issue there, it has the potential to break our users' scripts.

I went looking for alternatives. `argparse` has the concept of custom `Action`s, which could be a solution, but the documentation wasn't very helpful IMO:

https://docs.python.org/3/library/argparse.html#action-classes

Also, at a different place, it says "New in version 3.9", so not sure if this works with 3.8.

Maybe another way would be to detect if there are missing arguments with `_` in the name that, if replaced by `-`, would match an existing argument, and then add a useful warning, but leave the arguments untouched otherwise? This could be less disruptive.

Also note:
- [click](https://github.com/pallets/click/issues/1123) made the change to only use `-`
- [typer](https://github.com/tiangolo/typer/issues/341) automatically uses `-` instead of `_`

src/accelerate/commands/utils.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+
+
+class ArgumentParserWithDashSupport(argparse.ArgumentParser):
```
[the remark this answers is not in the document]
muellerzr: The aim is to not bloat `--help`, these are passthrough arguments to downstream API hook-ins such as `torchrun` which uses `-` rather than `_`.

PR #2526, opened 2024-03-04
Quanto compatibility with QBitsTensor

SunMarc:
# What does this PR
For quanto compatibility with lower bit quantization, we need to check for QBitsTensor class too. 
Related [PR](https://github.com/huggingface/accelerate/pull/2481).


muellerzr (approved):
Thanks! If we get too many more, might be good to store in a const list

muellerzr:
@SunMarc is there anything new that makes this test flakey? Saw this in a pr earlier today

SunMarc:
> Thanks! If we get too many more, might be good to store in a const list

Yes definitely but I think that we won't add more, just like bnb ! 

> @SunMarc is there anything new that makes this test flakey? Saw this in a pr earlier today

This should not impact the failing test at all :/

PR #2530, opened 2024-03-07
[`fsdp`] Allow `set_auto_wrap_policy` without any split modules

tomaarsen:
# What does this PR do?

## Before submitting
      Pull Request section?
      to it if that's the case.

## Details
If `_no_split_modules` is `None` or an empty list, then `default_transformer_cls_names_to_wrap` becomes an empty string: `""`. Then, `transformer_cls_names_to_wrap` calls `.split(",")` on that potentially empty string, resulting in `[""]` as the list of transformer class names that should be wrapped. We loop over these class names, and obviously the empty string can't be found within the model. This results in `raise Exception("Could not find the transformer layer class to wrap in the model.")` when `model._no_split_modules = []` and the auto wrap policy is `TRANSFORMER_BASED_WRAP`.

This PR updates this surprising behaviour to ignore the "empty class name" case.

I encountered this issue when subclassing the `transformers` `Trainer` for Sentence Transformers. I realise that it's possible that in doing so, the bug may have actually been on my side, so apologies if that is the case.

## Who can review?

cc @pacman100

- Tom Aarsen

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -1144,11 +1144,12 @@ class FullyShardedDataParallelPlugin:
                 ).split(",")
                 transformer_cls_to_wrap = set()
                 for layer_class in transformer_cls_names_to_wrap:
-                    transformer_cls = FullyShardedDataParallelPlugin.get_module_class_from_name(model, layer_class)
-                    if transformer_cls is None:
-                        raise Exception("Could not find the transformer layer class to wrap in the model.")
-                    else:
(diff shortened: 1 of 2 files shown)


tomaarsen:
I see! Fair enough. So, if someone's architecture doesn't include `_no_split_modules`, e.g. with BERT or MPNet, then they should set `transformer_layer_cls_to_wrap` in the `fsdp_config` with e.g. `["BertLayer"]`? What kind of class names should be used here?

Or should users use `NO_WRAP` instead?

PR #2531, opened 2024-03-07
Allow Gradients to be Synced Each Data Batch While Performing Gradient Accumulation

fabianlim:
# What does this PR do?

Fixes https://github.com/huggingface/transformers/issues/29425

This PR adds in logic to allow `no_sync` to be disabled in `accelerate`, in particular, so that memory can be saved whilst performing FSDP training. See details in the above linked issue.
- Introduced new  argument `GradientAccumulationPlugin.sync_each_batch`, when set will cause gradients to be synced each data batch, regardless of what the setting of `gradient_accumulation_steps` is.
- Introduced a `force` flag into `_do_sync`. This flag is set to `True` during `accumulate`, if `GradientAccumulationPlugin.sync_each_batch` has been configured to `True. 
- Updated tests `tests/test_grad_sync.py::SyncScheduler::test_gradient_sync_gpu_mult` and `test_gradient_sync_gpu`.
   - update test utilities `test_gradient_accumulation` and `test_gradient_accumulation_with_opt_and_scheduler` in [tests/test_grad_sync.py](https://github.com/fabianlim/accelerate/blob/feature-disable-no-sync/src/accelerate/test_utils/scripts/test_sync.py) to take in `sync_each_batch` that will test each case where the flag is set `True` or `False`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/docs/source/concept_guides/gradient_synchronization.md
+++ b/docs/source/concept_guides/gradient_synchronization.md
@@ -180,4 +180,5 @@ See the example below where we fine-tune Mixtral (47B parameters) on 8 A100-80GB
 | :-------------: | :-----------------: | :-----------------: | :-----------------: 
 mixtral 8x7B      | 69G                 | OOM                 | 69G
 
-Finally, note that disabling `no_sync` means there _will be slowdown_ due the extra data syncs, as explained by the earlier sections of this guide.
+> [!WARNING] 
+> Disabling `no_sync` means there _will be slowdown_ due the extra data syncs, as explained by the earlier sections of this guide.
--- a/docs/source/concept_guides/gradient_synchronization.md
+++ b/docs/source/concept_guides/gradient_synchronization.md
@@ -170,9 +170,9 @@ If you are worried about making sure everything is done properly, we highly reco
 
 ### `no_sync` requires additional GPU memory when using FSDP
 
-Be aware that not syncing gradients can have adverse effects while performing FSDP training. As it has been warned in `torch`, the [`no_sync` context manager for FSDP](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync) will incur additional memory.
+Be aware that not syncing gradients can have adverse effects while performing FSDP training. As it has been warned in `torch`, the [`no_sync` context manager for FSDP](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync) will require additional memory.
 
-Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `False` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
+Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `True` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
 
 See the example below where we fine-tune Mixtral (47B parameters) on 8 A100-80GB GPUs. We see that even for a modest `gradient_accumulation_steps=2` we quickly go out-of-memory (OOM) if `no_sync` is enabled. Again, this is due to additional memory overheads due to FSDP's `no_sync`. However, if `no_sync` is disabled via `sync_each_batch=True`, then the memory consumption for `gradient_accumulation_steps=16` reverts to that of `gradient_accumulation_steps=1`.
 
--- a/docs/source/concept_guides/gradient_synchronization.md
+++ b/docs/source/concept_guides/gradient_synchronization.md
@@ -167,3 +167,17 @@ As you can see, if you are not careful about how you set up your gradient synchr
 
 If you are worried about making sure everything is done properly, we highly recommend utilizing the [`~Accelerator.accumulate`] function and passing in
 `gradient_accumulation_steps` or `gradient_accumulation_plugin` to the [`Accelerator`] object so Accelerate can handle this for you.
+
+### `no_sync` requires additional GPU memory when using FSDP
+
+Be aware that not syncing gradients can have adverse effects while performing FSDP training. As it has been warned in `torch`, the [`no_sync` context manager for FSDP](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync) will incur additional memory.
+
+Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `False` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
+
+See the example below where we fine-tune Mixtral (47B parameters) on 8 A100-80GB GPUs. We see that even for a modest `gradient_accumulation_steps=2` we quickly go out-of-memory (OOM) if `no_sync` is enabled. Again, this is due to additional memory overheads due to FSDP's `no_sync`. However, if `no_sync` is disabled via `sync_each_batch=True`, then the memory consumption for `gradient_accumulation_steps=16` reverts to that of `gradient_accumulation_steps=1`.
+
+| Model           | `no_sync` (accum=1) | `no_sync` (accum=2) | `no_sync` disabled (accum=16)
+| :-------------: | :-----------------: | :-----------------: | :-----------------: 
+mixtral 8x7B      | 69G                 | OOM                 | 69G
+
+Finally, note that disabling `no_sync` means there _will be slowdown_ due the extra data syncs, as explained by the earlier sections of this guide.
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1025,7 +1025,7 @@ class Accelerator:
         ...         optimizer.zero_grad()
         ```
         """
-        # sync_each_batch=True will gaurantee below that self.sync_gradients=True, therefore
+        # sync_each_batch=True will guarantee below that self.sync_gradients=True, therefore
         # resulting in the nullcontext always being selected.
         self._do_sync(force=self.gradient_state.plugin_kwargs.get("sync_each_batch", False))
         with contextlib.ExitStack() as cm_stack:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -972,7 +972,7 @@ class Accelerator:
             model.require_backward_grad_sync = old_require_backward_grad_sync
             model.require_forward_param_sync = old_require_forward_param_sync
 
-    def _do_sync(self, force: bool = True):
+    def _do_sync(self, force: bool = False):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
         if self.gradient_state.sync_with_dataloader and self.gradient_state.end_of_dataloader:
             self.step = 0
@@ -1025,6 +1025,8 @@ class Accelerator:
         ...         optimizer.zero_grad()
         ```
         """
+        # sync_each_batch=True will gaurantee below that self.sync_gradients=True, therefore
+        # resulting in the nullcontext always being selected.
         self._do_sync(force=self.gradient_state.plugin_kwargs.get("sync_each_batch", False))
         with contextlib.ExitStack() as cm_stack:
             for m in models:
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -979,9 +979,7 @@ class Accelerator:
             self.gradient_state._set_sync_gradients(True)
         else:
             self.step += 1
-            self.gradient_state._set_sync_gradients(
-                force or ((self.step % self.gradient_state.num_steps) == 0)
-            )
+            self.gradient_state._set_sync_gradients(force or ((self.step % self.gradient_state.num_steps) == 0))
 
     @property
     def sync_gradients(self):
@@ -1027,7 +1025,7 @@ class Accelerator:
         ...         optimizer.zero_grad()
         ```
         """
-        self._do_sync(force=self.gradient_state.plugin_kwargs.get('sync_each_batch', False))
+        self._do_sync(force=self.gradient_state.plugin_kwargs.get("sync_each_batch", False))
         with contextlib.ExitStack() as cm_stack:
             for m in models:
                 cm_stack.enter_context(contextlib.nullcontext() if self.sync_gradients else self.no_sync(m))
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -972,14 +972,16 @@ class Accelerator:
             model.require_backward_grad_sync = old_require_backward_grad_sync
             model.require_forward_param_sync = old_require_forward_param_sync
 
-    def _do_sync(self):
+    def _do_sync(self, force: bool = True):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
         if self.gradient_state.sync_with_dataloader and self.gradient_state.end_of_dataloader:
             self.step = 0
             self.gradient_state._set_sync_gradients(True)
         else:
             self.step += 1
-            self.gradient_state._set_sync_gradients((self.step % self.gradient_state.num_steps) == 0)
+            self.gradient_state._set_sync_gradients(
+                force or ((self.step % self.gradient_state.num_steps) == 0)
+            )
 
     @property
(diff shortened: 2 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -972,14 +972,14 @@ def trigger_sync_in_backward(model):
...
-    def _do_sync(self):
+    def _do_sync(self, force: bool = True):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
         if self.gradient_state.sync_with_dataloader and self.gradient_state.end_of_dataloader:
             self.step = 0
             self.gradient_state._set_sync_gradients(True)
         else:
             self.step += 1
-            self.gradient_state._set_sync_gradients((self.step % self.gradient_state.num_steps) == 0)
+            self.gradient_state._set_sync_gradients(force or ((self.step % self.gradient_state.num_steps) == 0))
```
muellerzr, line 982: I don't think there's a reason to have this be a passthrough, Accelerator will always have a gradient_state so we can pull from it

src/accelerate/accelerator.py
```diff
@@ -1025,7 +1025,7 @@ def accumulate(self, *models):
         ...         optimizer.zero_grad()
         ```
         """
-        self._do_sync()
+        self._do_sync(force=self.gradient_state.plugin_kwargs.get("sync_each_batch", False))
```
muellerzr, line 1028: Similarly no need for a `get`

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -204,9 +204,13 @@ def test_distributed_sync_multiple_fwd(accelerator):
                 ), f"Gradients not in sync when they should be:\nModel grad ({param.grad}) != DDP grad ({ddp_param.grad})"
 
 
-def test_gradient_accumulation(split_batches=False, dispatch_batches=False):
+def test_gradient_accumulation(split_batches=False, dispatch_batches=False, sync_each_batch=False):
+    from accelerate.accelerator import GradientAccumulationPlugin
```
muellerzr, line 208: Imports should be at the top

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -204,9 +204,13 @@ def test_distributed_sync_multiple_fwd(accelerator):
...
+
     accelerator = Accelerator(
-        split_batches=split_batches, dispatch_batches=dispatch_batches, gradient_accumulation_steps=2
+        split_batches=split_batches,
+        dispatch_batches=dispatch_batches,
+        gradient_accumulation_plugin=GradientAccumulationPlugin(num_steps=2, sync_each_batch=sync_each_batch),
```
muellerzr, line 213: Let's make the plugin defined on a separate line for readability

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -242,9 +246,15 @@ def test_gradient_accumulation(split_batches=False, dispatch_batches=False):
     GradientState._reset_state()
 
 
-def test_gradient_accumulation_with_opt_and_scheduler(split_batches=False, dispatch_batches=False):
+def test_gradient_accumulation_with_opt_and_scheduler(
+    split_batches=False, dispatch_batches=False, sync_each_batch=False
+):
+    from accelerate.accelerator import GradientAccumulationPlugin
```
muellerzr, line 252: Similar

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -242,9 +246,15 @@ def test_gradient_accumulation(split_batches=False, dispatch_batches=False):
...
+
     accelerator = Accelerator(
-        split_batches=split_batches, dispatch_batches=dispatch_batches, gradient_accumulation_steps=2
+        split_batches=split_batches,
+        dispatch_batches=dispatch_batches,
+        gradient_accumulation_plugin=GradientAccumulationPlugin(num_steps=2, sync_each_batch=sync_each_batch),
```
muellerzr, line 257: Similar

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -254,32 +264,42 @@ def test_gradient_accumulation_with_opt_and_scheduler(split_batches=False, dispa
         input, target = accelerator.gather((ddp_input, ddp_target))
         input, target = input.to(accelerator.device), target.to(accelerator.device)
         # Perform our initial ground truth step in non "DDP"
-        model.train()
+        model.train()  # is this needed?
```
muellerzr, line 267: Yes, it's needed for clarity :)

src/accelerate/test_utils/scripts/test_sync.py
```diff
@@ -254,32 +264,42 @@ def test_gradient_accumulation_with_opt_and_scheduler(split_batches=False, dispa
...
-            check_model_parameters(model, ddp_model, did_step, iteration)
+            check_model_parameters(
+                model,
+                ddp_model,
+                did_step,
+                iteration,
+                rtol=1e-3,  # somehow needs a relative tolerance
+            )
+
+        if ((iteration + 1) % 2 == 0) or ((iteration + 1) == len(dataloader)) or sync_each_batch:
```
muellerzr, line 299: Just use `did_step`?

Also why are we changing when zero_grad is performed?

muellerzr (commented):
Thanks! This is a great start. Have a few implementation questions and suggestions

src/accelerate/accelerator.py
fabianlim: @muellerzr Sorry I made a mistake on line 975 where the default value for `force` in `_do_sync(self, force: bool = False)`  should be `False`. Hence if `force=False` it will still check the `((self.step % self.gradient_state.num_steps) == 0` condition.

The rationale for introducing the `force` flag, is so that then `accumulate` might be more explicit. Because then it reads as below, where if `sync_each_batch=False`, then `self.sync_gradients` will be gauranteed to be true and thus the `nullcontext` will be selected. 

```python
@contextmanager
def accumulate(self, *models):

    # sync_each_batch=True will gaurantee below that self.sync_gradients=True, therefore 
    # resulting in the nullcontext always being selected.
    self._do_sync(force=self.gradient_state.plugin_kwargs["sync_each_batch"])
    with contextlib.ExitStack() as cm_stack:
        for m in models:
            cm_stack.enter_context(contextlib.nullcontext() if self.sync_gradients else self.no_sync(m))
        yield
```

However if you prefer, I can move the `self.gradient_state.plugin_kwargs["sync_each_batch"])` check inside `_do_sync` (and in turn removing `force` from the signature of `_do_sync`).  Please let me know which you think is better?

src/accelerate/test_utils/scripts/test_sync.py
fabianlim: removed the comment i added.

src/accelerate/accelerator.py
fabianlim: @muellerzr the `get` is for the kwarg key `sync_each_batch`. Turns out that [KwargsHandler only returns a key-value if the value is non-default](https://github.com/huggingface/accelerate/blob/10882eeddd841b0a1a8064421816ee0daef99920/src/accelerate/utils/dataclasses.py#L57). Hence if `sync_each_batch` was set to `False`, then it will be missing in `plugin_kwargs`.

[See here for how `plugin_kwargs` are populated in `GradientState`](https://github.com/fabianlim/accelerate/blob/329327b0507686f8cf0c0748849cd4858aa5b37b/src/accelerate/state.py#L1034)

src/accelerate/accelerator.py
muellerzr: Let's just go ahead with it in `_do_sync` then, thanks for the explanation!

muellerzr:
@fabianlim can you run `make style; make quality`? Thanks!

fabianlim:
@muellerzr i have run `make style; make quality`. However I noticed that it linted two files which I did not touch, namely `utils/memory.py` and `tests/deepspeed/deepspeed.py`. The linting suggestion for the first file didnt seem correct to me; it changed the meaning of the doc string it linted. 


```python
>>> from accelerate.utils import find_executable_batch_size


>>> @find_executable_batch_size(starting_batch_size=128)
... def train(batch_size, model, optimizer):
...     ...


>>> train(model, optimizer)
```

to 

```python
>>> from accelerate.utils import find_executable_batch_size


>>> @find_executable_batch_size(starting_batch_size=128)
... def train(batch_size, model, optimizer): ...


... train(model, optimizer)
```
So i did not commit these 2 files.

Also as mentioned I have not yet updated the documents, maybe I can take a first stab at updating [this page on grad accum](https://moon-ci-docs.huggingface.co/docs/accelerate/pr_2531/en/concept_guides/gradient_synchronization)
- will have a section that says "Caveat when using gradient accumulation with FSDP"
- highlight the warning from `torch` that the `no_sync` of FSDP will not leave the params and grads unsharded until the end of a gradient accum period, and this will incur additional GPU usage
- for larger models, one may consider setting `sync_each_batch` to True in `GradientAccumulationPlugin` to not use `no_sync`.
- port over some memory usage numbers in my original issue.

muellerzr:
If the linter changed things unrelated to your PR, then there are issues with the versions of what you are using. You can do `pip install -e .[quality]` to make sure the right versions are installed

fabianlim:
> If the linter changed things unrelated to your PR, then there are issues with the versions of what you are using. You can do pip install -e .[quality] to make sure the right versions are installed
@muellerzr thanks. As you said.. once I had one the above I do not get complaints on those files.

As mentioned I have [updated the docs](https://github.com/huggingface/accelerate/pull/2531#:~:text=this%20page%20on%20grad%20accum). 

Also I have raised the accompanying PR https://github.com/huggingface/transformers/pull/29589 that allows the newly introduced `sync_each_batch` flag in  `GradientAccumulationPlugin` to be set.
- I noticed that `Trainer` currently [hardcodes `GradientAccumulationPlugin`](https://github.com/huggingface/transformers/blob/dd1c9052159ae824c8acef7c2552f9fad5ca020a/src/transformers/trainer.py#L4097-L4099). 
- the accompanying PR allows  `GradientAccumulationPlugin` to be configured by `AcceleratorConfig`. As you [mentioned to me previously](https://github.com/huggingface/transformers/issues/29425), we are not allowed to add new `accelerate` arguments directly into `TrainingArguments`. 
   * hence my strategy was to add `gradient_accumulation_kwargs` into `AcceleratorConfig` and pass that to configure `GradientAccumulationPlugin`.
- I have done some tests and indeed it works as intended with the `accelerate` changes in this PR.

docs/source/concept_guides/gradient_synchronization.md
```diff
@@ -167,3 +167,17 @@ As you can see, if you are not careful about how you set up your gradient synchr
...
+
+Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `False` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
+
+See the example below where we fine-tune Mixtral (47B parameters) on 8 A100-80GB GPUs. We see that even for a modest `gradient_accumulation_steps=2` we quickly go out-of-memory (OOM) if `no_sync` is enabled. Again, this is due to additional memory overheads due to FSDP's `no_sync`. However, if `no_sync` is disabled via `sync_each_batch=True`, then the memory consumption for `gradient_accumulation_steps=16` reverts to that of `gradient_accumulation_steps=1`.
+
+| Model           | `no_sync` (accum=1) | `no_sync` (accum=2) | `no_sync` disabled (accum=16)
+| :-------------: | :-----------------: | :-----------------: | :-----------------: 
+mixtral 8x7B      | 69G                 | OOM                 | 69G
+
+Finally, note that disabling `no_sync` means there _will be slowdown_ due the extra data syncs, as explained by the earlier sections of this guide.
```
muellerzr, line 183:
```suggestion
> [!WARNING] 
> Disabling `no_sync` means there _will be slowdown_ due the extra data syncs, as explained by the earlier sections of this guide.
```

src/accelerate/utils/dataclasses.py
```diff
@@ -585,6 +585,9 @@ class GradientAccumulationPlugin(KwargsHandler):
             `True` if the used scheduler was not adjusted for gradient accumulation.
         sync_with_dataloader (`bool`, *optional*, defaults to `True`):
             Whether to synchronize setting the gradients when at the end of the dataloader.
+        sync_each_batch (`bool`, *optional*, defaults to `False`):
```
muellerzr, line 588: `False` defaults don't need to be mentioned :) 

```suggestion
        sync_each_batch (`bool`, *optional*):
```

src/accelerate/utils/dataclasses.py
```diff
@@ -609,6 +612,12 @@ class GradientAccumulationPlugin(KwargsHandler):
             "help": "Whether to synchronize setting the gradients when at the end of the dataloader. Should only be set to `False` if you know what you're doing."
         },
     )
+    sync_each_batch: bool = field(
+        default=False,
+        metadata={
+            "help": "Whether to synchronize setting the gradients at each data batch. Seting to `True` may reduce memory requirements when using gradient accumulation with distributed training, at expense of speed."
```
muellerzr, line 618:
```suggestion
            "help": "Whether to synchronize setting the gradients at each data batch. Setting to `True` may reduce memory requirements when using gradient accumulation with distributed training, at expense of speed."
```

muellerzr (approved):
Thanks for this! Overall looks good to me, cc @BenjaminBossan for a quick check as well

docs/source/concept_guides/gradient_synchronization.md
```diff
@@ -167,3 +167,17 @@ As you can see, if you are not careful about how you set up your gradient synchr
 
 If you are worried about making sure everything is done properly, we highly recommend utilizing the [`~Accelerator.accumulate`] function and passing in
 `gradient_accumulation_steps` or `gradient_accumulation_plugin` to the [`Accelerator`] object so Accelerate can handle this for you.
+
+### `no_sync` requires additional GPU memory when using FSDP
+
+Be aware that not syncing gradients can have adverse effects while performing FSDP training. As it has been warned in `torch`, the [`no_sync` context manager for FSDP](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync) will incur additional memory.
```
BenjaminBossan, line 173: Nit: incur => require

docs/source/concept_guides/gradient_synchronization.md
```diff
@@ -167,3 +167,17 @@ As you can see, if you are not careful about how you set up your gradient synchr
...
+
+Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `False` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
```
BenjaminBossan, line 175:
```suggestion
Therefore in memory intensive situations while using FSDP, we recommend to set `sync_each_batch` to `True` in the [`~utils.GradientAccumulationPlugin`] to disable `no_sync`.
```

Right?

src/accelerate/accelerator.py
```diff
@@ -1025,7 +1025,9 @@ def accumulate(self, *models):
         ...         optimizer.zero_grad()
         ```
         """
-        self._do_sync()
+        # sync_each_batch=True will gaurantee below that self.sync_gradients=True, therefore
```
BenjaminBossan, line 1028:
```suggestion
        # sync_each_batch=True will guarantee below that self.sync_gradients=True, therefore
```

BenjaminBossan (changes requested):
Nice investigation and fix, well documented and tested. Thanks a lot.

A just found some minor issues with the docs, the rest looks good.

docs/source/concept_guides/gradient_synchronization.md
fabianlim: oh my a blunder on my part.. thanks for the catch

BenjaminBossan (approved):
Thanks for the quick fixes.

muellerzr:
Will merge once tests pass here, @fabianlim also afterwards be sure to keep an eye on the CI with `main` as those will run the tests with a GPU. (I'll ping you here if any fail)

PR #2535, opened 2024-03-07
Add NUMA affinity control for NVIDIA GPUs

muellerzr:
# What does this PR do?

This PR adds the ability to launch using `numactl`, which acts as an inbetween stage when launching from `torchrun` etc.

Before:

```bash
trampoline.sh

#!/usr/bin/bash

# Query the bus ID for device LOCAL_RANK
BUS_ID=$(nvidia-smi --query-gpu=pci.bus_id -i $LOCAL_RANK --format=csv,noheader)
BUS_ID=${BUS_ID,,}

# Find the numa node for device LOCAL_RANK
NODE=$(cat /sys/bus/pci/devices/${BUS_ID:4}/numa_node)

echo "Starting local rank $RANK on numa node $NODE"
numactl --cpunodebind=$NODE --membind=$NODE "$@"

torchrun --nproc_per_node=8 --monitor-interval=1 --no-python ./trampoline.sh python3 -c "print('hello')"
```

After:

```bash
accelerate launch --num_processes 8 --monitor-interaval 1 --enable-numa-affinity python3 -c "print('hello')"
```
Or through the config:

```diff
compute_environment: LOCAL_MACHINE
deepspeed_config: {}
distributed_type: 'MULTI_GPU'
...
enable_numa_affinity: true
```

```
accelerate launch myfile.py
```

Fixes https://github.com/huggingface/accelerate/issues/2241

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @stas00


src/accelerate/utils/numactl.py
```diff
@@ -0,0 +1,67 @@
...
+
+def get_bus_id(device_id: int = None, backend: str = None):
+    """
+    Gets the bus ID for `device_id`. If not passed, will default to the local rank.
+    """
+    if device_id is None:
+        device_id = os.environ.get("LOCAL_RANK", 0)
+    command = get_query_command(device_id, backend)
+    output = subprocess.check_output(command, universal_newlines=True)
+    return output.strip()[4:]
```
BenjaminBossan, line 42: Just to be sure: On my machine, this returns `'0000:01:00.0'`, would that be correct? Without stripping, it's `'00000000:01:00.0'`.

src/accelerate/utils/numactl.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def get_numa_node_for_device(device_id: int = None):
+    """
+    Gets the numa node for `device_id`. If not passed, will default to the local rank.
+    """
+    bus_id = get_bus_id(device_id)
+    if bus_id is None:
+        bus_id = 4
```
BenjaminBossan, line 51: Could you please add a comment where `4` is coming from?

tests/test_cli.py
```diff
@@ -190,6 +191,73 @@ def test_duplicate_entities(self):
                     assert bad_arg not in help_return, f"Found {bad_arg} in `accelerate launch -h`"
 
 
+class ClusterConfigTester(unittest.TestCase):
```
BenjaminBossan, line 194: Just for my understanding, these are general tests that don't relate to numactl specifically, right?

BenjaminBossan (commented):
Thanks, Zach, for enabling NUMA affinity control to be easily enabled with accelerate. I have no experience whatsoever with numactl, so I skipped those specific parts of the review and focused on the rest.

src/accelerate/utils/numactl.py
muellerzr: Yep that is indeed correct

src/accelerate/utils/numactl.py
muellerzr: cc @stas00 for insight? 👀

stas00: yeah, the above is not what the bash code did. Let's look at the original bash:

```
# Query the bus ID for device LOCAL_RANK
BUS_ID=$(nvidia-smi --query-gpu=pci.bus_id -i $LOCAL_RANK --format=csv,noheader)
BUS_ID=${BUS_ID,,}

# Find the numa node for device LOCAL_RANK
NODE=$(cat /sys/bus/pci/devices/${BUS_ID:4}/numa_node)
```
4 is there to remove the first 4 characters - it's indeed a good idea to add a comment explaining it, perhaps copy this verbatim as it's self-explanatory:

```
$ LOCAL_RANK=5 BUS_ID=$(nvidia-smi --query-gpu=pci.bus_id -i $LOCAL_RANK --format=csv,noheader)
$ echo $BUS_ID
00000000:85:00.0
$ ls -l /sys/bus/pci/devices/${BUS_ID}
ls: cannot access '/sys/bus/pci/devices/00000000:85:00.0': No such file or directory
$ echo ${BUS_ID:4}
0000:85:00.0
$ ls -l /sys/bus/pci/devices/${BUS_ID:4}
lrwxrwxrwx 1 root root 0 Feb 28 19:38 /sys/bus/pci/devices/0000:85:00.0 -> ../../../devices/pci0000:81/0000:81:00.0/0000:82:00.0/0000:83:01.0/0000:85:00.0
```

`${var:pos}` is bash's adoption of ksh functionality - it'll truncate the string starting at `pos`

src/accelerate/utils/numactl.py
```diff
@@ -0,0 +1,67 @@
...
+    return subprocess.check_output(["cat", f"/sys/bus/pci/devices/{bus_id}/numa_node"]).strip().decode()
```
stas00, line 52: why not just open this as a normal file in python?

src/accelerate/utils/numactl.py
```diff
@@ -0,0 +1,67 @@
...
+
+def launch():
+    args = sys.argv
+    script, *script_args = args[1:]
+    node = get_numa_node_for_device()
+    cmd = ["numactl", f"--cpunodebind={node}", f"--membind={node}", "python3", script] + script_args
+    process = subprocess.Popen(cmd, env=os.environ.copy())
+    process.wait()
+    if process.returncode != 0:
+        raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
```
stas00, line 63: why do you need to do that when you can do it from python?

https://github.com/NVIDIA/DeepLearningExamples/blob/9dd9fcb98f56187e49c5ee280cf8dbd530dde57b/TensorFlow2/LanguageModeling/BERT/gpu_affinity.py#L55-L63

This is the pynvml way I shared in the OP of the issue, but we are just removing `pynvml`

stas00: I also didn't see where you check that `numactl` exists - it is not installed on linux by default. So doing it via python is probably much simpler.

muellerzr: @stas00 do you know if we would be okay doing this in the `AcceleratorState`?

Or better phrased as:

Is it paramount that this gets set before things like `torch.distributed` is initialized

stas00: yes, of course, you can change the affinity dynamically after the process started:

Quote from https://linux.die.net/man/2/sched_setaffinity
> If the process specified by pid is not currently running on one of the CPUs specified in mask, then that process is migrated to one of the CPUs specified in mask. 

i.e. it doesn't matter how the process is started.

muellerzr:
@stas00 I think this should fully enable you now. You can either set `enable_cpu_affinity` to `True` in your `config.yaml`, or set the env variable `ACCELERATE_CPU_AFFINITY=1`

src/accelerate/utils/environment.py
```diff
@@ -121,3 +129,41 @@ def check_fp8_capability():
...
+    # As it can be passed to `ls -l /sys/bus/pci/devices/${bus_id}`
+    return output.strip()[4:]
+
+
+def get_numa_node_for_device(device_id: int = None, backend: str = None):
+    """
+    Gets the numa node for `device_id`. If not passed, will default to the local rank.
+    """
+    bus_id = get_bus_id(device_id, backend)
+    if bus_id is None:
```
BenjaminBossan, line 166: `get_bus_id` will always return a string unless it errors, right? So this is dead code.

BenjaminBossan (approved):
Thanks a lot. The code LGTM. Regarding whether this solves the initial problem, I defer to Stas.

stas00 (approved):
This looks good code-wise, but I won't have a chance to try it out till next week as I'm on the road.

Did you get a chance to test it on the H100 nodes at HF? If so it's probably good to go. If not and you don't mind waiting I will test next week.

muellerzr:
Despite my best efforts couldn't get away from using pynvml 😭 However it seems to come auto-installed (probably with PyTorch?) so did a simple import check. As we get more libs/backend support we can look at appropriate interfaces rather than reinvent the wheel :)

muellerzr:
Confirmed with @stas00 the load balancing looks proper, once tests pass will merge ✅

Issue #2479, opened 2024-02-22
Not to use_seedable_sampler causes `accelerate test` failure on TPU v2/v3


vanbasten23:
### System Info

We have found that https://github.com/huggingface/accelerate/pull/2319 causes `accelerate test` fails on TPU v2/v3 but the v4 tests succeed.  The PR huggingface/transformers#28664 doesn't help because we are directly using `accelerate test` in accelerate instead of transformers. 

The failing tests are here https://github.com/huggingface/accelerate/blob/60bfdaa9349e3e1a373863a8c5f20baa89cfd000/src/accelerate/test_utils/scripts/test_script.py#L214-L237 when we use a DataLoader with `shuffle` enabled. The error message is:
```
stderr: """
stderr: Traceback (most recent call last):
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
stderr:     r = call_item.fn(*call_item.args, **call_item.kwargs)
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in _process_chunk
stderr:     return [fn(*args) for args in chunk]
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/process.py", line 198, in <listcomp>
stderr:     return [fn(*args) for args in chunk]
stderr:   File "/ansible/pytorch/xla/torch_xla/runtime.py", line 88, in wrapper
stderr:     return fn(*args, **kwargs)
stderr:   File "/ansible/pytorch/xla/torch_xla/_internal/pjrt.py", line 78, in _run_thread_per_device
stderr:     replica_results = list(
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 619, in result_iterator
stderr:     yield fs.pop().result()
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 444, in result
stderr:     return self.__get_result()
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/_base.py", line 389, in __get_result
stderr:     raise self._exception
stderr:   File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
stderr:     result = self.fn(*self.args, **self.kwargs)
stderr:   File "/ansible/pytorch/xla/torch_xla/_internal/pjrt.py", line 71, in _thread_fn
stderr:     return fn()
[...]

PR #2542, opened 2024-03-10
Fix test_script.py on TPU v2/v3

vanbasten23:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/2479

- This PR fix the dataloader test: The same test `accelerate test` succeeds on TPU v4 but fails on v2/v3. The reason why it fails on TPU v2/v3 is multithreading. TPU v2/v3 involves multithreading: there are 4 processes and 8 TPU devices on the TPU v3-8, and each process spawns 2 threads for each device respectively. However TPU v4 and further TPU generations doesn't use multithreading. On TPU v2/v3, due to multithreading, we need to explicitly set the seed on each thread. Otherwise, each thread will permute the dataset indices in a different way and therefore generate the duplicate indices as shown in the gh issue. This PR explicitly set the `generator` for each thread and fix the test failure.
- This PR also fix a `xm.set_replication` issue for TPU v2/v3.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -870,11 +870,6 @@ def prepare_data_loader(
         sampler = getattr(dataloader.sampler, "sampler", None)
     else:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
-    if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
-        # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
-        generator = torch.Generator().manual_seed(42)
-        dataloader.generator = generator
-        dataloader.sampler.generator = generator
     if isinstance(sampler, RandomSampler) and use_seedable_sampler:
         # When iterating through the dataloader during distributed processes
         # we want to ensure that on each process we are iterating through the same
@@ -887,6 +882,12 @@ def prepare_data_loader(
             generator=getattr(sampler, "generator", torch.Generator()),
         )
 
+    if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
+        # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
+        generator = torch.Generator().manual_seed(42)
+        dataloader.generator = generator
+        dataloader.sampler.generator = generator
+        dataloader.batch_sampler.sampler = dataloader.sampler
     # No change if no multiprocess
     if (num_processes != 1 or state.distributed_type == DistributedType.MEGATRON_LM) and not dispatch_batches:
         if isinstance(new_dataset, IterableDataset):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -872,9 +872,9 @@ def prepare_data_loader(
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
     if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
         # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
-        generator=torch.Generator().manual_seed(42)
-        dataloader.generator=generator
-        dataloader.sampler.generator=generator
+        generator = torch.Generator().manual_seed(42)
+        dataloader.generator = generator
+        dataloader.sampler.generator = generator
     if isinstance(sampler, RandomSampler) and use_seedable_sampler:
         # When iterating through the dataloader during distributed processes
         # we want to ensure that on each process we are iterating through the same
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -971,8 +971,6 @@ def prepare_data_loader(
         else:
             dataloader.batch_sampler.sampler = sampler
     if state.distributed_type == DistributedType.XLA:
-        # dataloader.generator=torch.Generator().manual_seed(42)
-        # dataloader.batch_sampler
         return MpDeviceLoaderWrapper(dataloader, device)
     return dataloader
 
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -871,9 +871,10 @@ def prepare_data_loader(
     else:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
     if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
-        seed=torch.Generator().manual_seed(42)
-        dataloader.generator=seed
-        dataloader.sampler.generator=seed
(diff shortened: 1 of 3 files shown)


vanbasten23:
cc @will-cromar @muellerzr @SunMarc

muellerzr (commented):
Thanks for the fantastic fix!

Looks like there's some style nits here, can you do `make style; make quality` after doing `pip install -e .[quality]`? Thanks!

src/accelerate/data_loader.py
```diff
@@ -870,6 +870,10 @@ def prepare_data_loader(
         sampler = getattr(dataloader.sampler, "sampler", None)
     else:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
+    if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
+        seed=torch.Generator().manual_seed(42)
```
will-cromar, line 874: nit with the naming: `manual_seed` doesn't return the seed, it returns a _seeded generator_: https://critique.corp.google.com/cl/577336396

So `seed` should instead be `seeded_generator`, `rng`, etc (whichever name matches the existing style the best).

will-cromar (approved):
Thanks @vanbasten23!

src/accelerate/data_loader.py
```diff
@@ -870,6 +870,11 @@ def prepare_data_loader(
         sampler = getattr(dataloader.sampler, "sampler", None)
     else:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
+    if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
+        # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
+        generator = torch.Generator().manual_seed(42)
+        dataloader.generator = generator
+        dataloader.sampler.generator = generator
```
muellerzr, line 877: Is this sort of logic going to have us run into issues when `use_seedable_sampler=True`? 

Also not sure setting `dataloader.generator` truly does anything, as we recreate a new dataloader anyways later?

We should be adjusting `dataloader.sampler.generator`, which *really* should be `sampler.generator`

muellerzr: I'm actually not sure how this fixes anything, as we again fully recreate the dataloader, so setting any attribute inside `dataloader` is moot

muellerzr: Same reasoning, that doesn't do anything. We need to adjust `sampler`, we should never be modifying the dataloader directly

src/accelerate/data_loader.py
```diff
@@ -887,6 +882,12 @@ def prepare_data_loader(
             generator=getattr(sampler, "generator", torch.Generator()),
         )
 
+    if isinstance(dataloader.sampler, RandomSampler) and state.distributed_type == DistributedType.XLA:
+        # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
+        generator = torch.Generator().manual_seed(42)
+        dataloader.generator = generator
+        dataloader.sampler.generator = generator
+        dataloader.batch_sampler.sampler = dataloader.sampler
```
muellerzr, line 890: Same reasoning, that doesn't do anything. We need to adjust `sampler`, we should never be modifying the dataloader directly

src/accelerate/data_loader.py
vanbasten23: > Also not sure setting dataloader.generator truly does anything, as we recreate a new dataloader anyways later?
> I'm actually not sure how this fixes anything, as we again fully recreate the dataloader, so setting any attribute inside dataloader is moot

Actually, setting dataloader.generator has some impact on how we recreate a new dataloader: here https://github.com/huggingface/accelerate/blob/60bfdaa9349e3e1a373863a8c5f20baa89cfd000/src/accelerate/data_loader.py#L921 we extract the `dataloader.generator` and then pass it in when we recreate a new dataloader here https://github.com/huggingface/accelerate/blob/60bfdaa9349e3e1a373863a8c5f20baa89cfd000/src/accelerate/data_loader.py#L961. I also set `dataloader.batch_sampler.sampler` here because `batch_sampler` is used to create the `new_batch_sampler` https://github.com/huggingface/accelerate/blob/60bfdaa9349e3e1a373863a8c5f20baa89cfd000/src/accelerate/data_loader.py#L901.

It is a tricky function to get it work correctly as it reused the dataloader that users pass in and reconstruct a new one. Initially we figured it out how to make it work by changing [line](https://github.com/huggingface/accelerate/blob/60bfdaa9349e3e1a373863a8c5f20baa89cfd000/src/accelerate/test_utils/scripts/test_script.py#L276) to `dl = DataLoader(range(length), batch_size=8, shuffle=True, generator=torch.Generator().manual_seed(42))`, by adding `generator`, as it is mandatory for TPU v2/v3 as in [here](https://github.com/pytorch/xla/blob/5c4ed89bc7bb3f7757ca403be887b25d1bdedfe1/test/pjrt/test_train_hf_transformer.py#L41). But I don't want the accelerate user to fail silently if they don't use `generator` when creating a dataloader with shuffling. That's why I decided to change the non-test code. I basically set things similar to what [DataLoader](https://github.com/pytorch/xla/blob/5c4ed89bc7bb3f7757ca403be887b25d1bdedfe1/test/pjrt/test_train_hf_transformer.py#L41) does when `shuffle` is True.

src/accelerate/data_loader.py
vanbasten23: Actually modifying the dataloader has some impact on the new dataloader: https://github.com/huggingface/accelerate/pull/2542#discussion_r1520488077

If I don't do `dataloader.generator = generator`, the test would fail a later test with an error `AssertionError: Did not obtain the same model on CPU or distributed training.`: https://gist.github.com/vanbasten23/ce6d416859b89c23cfd72e272d356504.

So I think it's needed.

src/accelerate/data_loader.py
muellerzr: Fantastic, thanks for the explanation!

muellerzr (approved):
Thanks for doing this! Overall it looks fine to me, I'd like if possible for you to verify that things will still work fine if you run the code with `seedable_sampler` set to `True` though before merging to know if we need to include it there as well. 

(I'd rather wish we kept things without needing to modify `dataloader` directly, but if that's not possible then it's okay)

vanbasten23:
> Thanks for doing this! Overall it looks fine to me, I'd like if possible for you to verify that things will still work fine if you run the code with `seedable_sampler` set to `True` though before merging to know if we need to include it there as well.
> 
> (I'd rather wish we kept things without needing to modify `dataloader` directly, but if that's not possible then it's okay)

Sure. I set `use_seedable_sampler=True` at https://github.com/huggingface/accelerate/blob/ee163b66fb7848892519e804688cb4ae981aacbe/src/accelerate/test_utils/scripts/test_script.py#L215 and also at line 230 and the test `accelerate test` passed. I also change the default value of `use_seeable_sampler` to True in `def prepare_data_loader` and  the test `accelerate test` passed too.

PR #2544, opened 2024-03-11
Update FSDP mixed precision setter to enable fsdp+qlora

pacman100:
# What does this PR do?
1. Update FSDP mixed precision setter to enable fsdp+qlora. When using quantization storage datatype as `float32`, the MixedPrecision policy should use `float32` but with autocasting enabled. This PR supports this. 
2. It adds `override` argument to force update the mixed precision policy.
3. It adds `buffer_autocast` argument to support casting buffers to `float32` which can be important for RoPE layer which gives incorrect results when cast to lower precision especially with longer context lengths.


muellerzr (approved):
Thanks!

PR #2546, opened 2024-03-11
Use uv instead of pip install for github CI

muellerzr:
# What does this PR do?

Opts for using `uv` instead of `pip install` for the github CI. It's so fast that we don't need to cache anymore, fun fact.

Total time: 

Before: **14m 15s** (~10min on a second commit), taken from https://github.com/huggingface/accelerate/actions/runs/8231938776

After: **6m 30s**, taken from https://github.com/huggingface/accelerate/actions/runs/8238343827

As you can see, *much* of our time is spent on caching (both download and upload of `n` workers, github doesn't like that bandwidth cost)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
       run: |
         if [[ ${{ matrix.test-kind }} = test_prod ]]; then uv pip install -e .[test_prod]; fi
         if [[ ${{ matrix.test-kind }} != test_prod ]]; then uv pip install -e .[testing,test_trackers]; fi
-        if [[ ${{ matrix.test-kind }} = test_rest ]]; then uv pip uninstall comet_ml -y; fi
+        if [[ ${{ matrix.test-kind }} = test_rest ]]; then uv pip uninstall comet_ml; fi
         if [[ ${{ matrix.test-kind }} = minimum ]]; then uv pip install torch==1.10.0; fi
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -44,22 +44,20 @@ jobs:
...
-    - name: Activate python cache
-      uses: actions/cache@v3
-      with:
-        path: |
-          ${{ env.pythonLocation }}
-          ${{ env.HF_HOME }}
-        key: ${{ env.pythonLocation }}-${{ matrix.pytorch-version }}-${{ matrix.test-kind }}-${{ hashFiles('setup.py') }}
+    - name: Setup uv
+      run: | 
+          curl -LsSf https://astral.sh/uv/install.sh | sh
```
BenjaminBossan, line 49: This script seems to always point to the latest release, so there is no way to control the version with this install method. Might not be an issue, we'll probably have to see how stable uv turns out to be.

BenjaminBossan (approved):
Looks nice. The main time saving comes from the "Post Activate python cache" stage. Do you have an idea how much time this cache saves us in the first place?

My main concern with uv is that it still seems to be in early stages of development, so I'm not sure how stable and robust it is.

muellerzr:
Caching saves us the 2-3min it takes to setup the env in 5s, but the issue is sometimes the cache can take a bit to download

PR #2557, opened 2024-03-14
Add log message for RTX 4000 series when performing multi-gpu inference with device_map

SunMarc:
# What does this PR do ?
This PR adds a message for users with RTX 4000 when using multi-gpu inference with device_map. 
Maybe we can do a better check by looking at the driver version but I need to dig into that a little bit. 
Needed [here](https://github.com/huggingface/diffusers/pull/6857) 
cc @sayakpaul

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -31,6 +31,7 @@ from .hooks import (
 )
 from .utils import (
     OffloadedWeightsLoader,
+    check_cuda_p2p_ib_support,
     check_device_map,
     extract_submodules_state_dict,
     find_tied_parameters,
@@ -463,6 +464,11 @@ def dispatch_model(
         else:
             model.cuda = add_warning(model.cuda, model)
 
+        # Check if we are using multi-gpus with RTX 4000 series
+        use_multi_gpu = len([device for device in set(device_map.values()) if device not in ("cpu", "disk")]) > 1
+        if use_multi_gpu and not check_cuda_p2p_ib_support():
+            logger.info("RTX 4000 series have issues with P2P. This can affect the multi-gpu inference when using accelerate device_map."
+                        "Please make sure to update your driver to the latest version")
     else:
         device = list(device_map.values())[0]
         # `torch.Tensor.to(<int num>)` is not supported by `torch_npu` (see this [issue](https://github.com/Ascend/pytorch/issues/16)).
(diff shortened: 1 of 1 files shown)


src/accelerate/big_modeling.py
```diff
@@ -463,6 +464,13 @@ def wrapper(*args, **kwargs):
         else:
             model.cuda = add_warning(model.cuda, model)
 
+        # Check if we are using multi-gpus with RTX 4000 series
+        use_multi_gpu = len([device for device in set(device_map.values()) if device not in ("cpu", "disk")]) > 1
+        if use_multi_gpu and not check_cuda_p2p_ib_support():
+            logger.info(
```
muellerzr, line 470: Let's do a `warn` just in case

muellerzr (approved):
Thanks!

sayakpaul (approved):
Thank you ❤️

So, I guess we don't have to do a version check from the `diffusers` codebase for this.

src/accelerate/big_modeling.py
```diff
@@ -463,6 +464,13 @@ def wrapper(*args, **kwargs):
         else:
             model.cuda = add_warning(model.cuda, model)
 
+        # Check if we are using multi-gpus with RTX 4000 series
+        use_multi_gpu = len([device for device in set(device_map.values()) if device not in ("cpu", "disk")]) > 1
+        if use_multi_gpu and not check_cuda_p2p_ib_support():
+            logger.warning(
+                "RTX 4000 series have issues with P2P. This can affect the multi-gpu inference when using accelerate device_map."
+                "Please make sure to update your driver to the latest version"
```
muellerzr, line 472: Given what https://github.com/huggingface/accelerate/pull/2558 introduces, we can clean this up a bit!
```suggestion
                "We've detected an older driver with an RTX 4000 series GPU. These drivers have issues with P2P. "
                "This can affect the multi-gpu inference when using accelerate device_map."
                "Please make sure to update your driver to the latest version"
```

SunMarc: awesome !

SunMarc:
Not sure why this test is failing . I'm pretty sure this is not related to this PR. I will let you decide if you merge it or not @muellerzr ! 
```
FAILED tests/trainer/test_trainer_seq2seq.py::Seq2seqTrainerTester::test_bad_generation_config_fail_early - ImportError: 
T5Tokenizer requires the SentencePiece library but it was not found in your environment. Checkout the instructions on the
installation page of its repo: https://github.com/google/sentencepiece#installation and follow the ones
that match your environment. Please note that you may need to restart your runtime after installation.
```

PR #2561, opened 2024-03-17
Allow notebook_launcher to launch to multiple GPUs from Colab

### LINKED ISSUE #2540: Can't utilize multiple GPUs on Colab environment
### System Info

```Shell
- `Accelerate` version: 0.27.2
- Platform: Linux-6.1.58+-x86_64-with-glibc2.35
- Python version: 3.10.12
- Numpy version: 1.23.5
- System RAM: 14.64 GB
- GPU type: Tesla T4
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

1. Deploy a GCE virtual machine with >1 GPUs.
2. Connect to said VM on Google Colab.
3. Create the accelerate config with the correct number of GPUs.
4. Restart the environment and run the code.

### Expected behavior

It doesn't seem to be possible to get accelerate to use multiple GPUs in a Google Colab environment, even if the VM Colab is connected to has multiple GPUs. The GPUs are available, with `torch.cuda.device_count()` returning the expected number of GPUs.

I believe the issue is in `launchers.py`, where we have the following problematic code:
```
elif in_colab:
    # No need for a distributed launch otherwise as it's either CPU or one GPU.
    if torch.cuda.is_available():
        print("Launching training on one GPU.")
    else:
        print("Launching training on one CPU.")
    function(*args)
```
There is very much a need for distributed launch capabilities, even in a Colab environment. I'm really unsure as to why this code was ever added in the first place... if the config specifies 4 GPUs then accelerate should try to use 4 GPUs, not just ignore this (silently no less, there is no indication of *why* it launches on one GPU so the user needs to delve into the code to figure it out).

StefanTodoran:
# What does this PR do?

Prior behavior for `notebook_launcher` was to ignore the `num_processes` parameter if `in_colab` and simply launch to one CPU/GPU. Since multiple GPUs can be accessed through Colab if the connected VM has multiple GPUs, this behavior was removed. If running in Colab and `num_processes` was unspecified, it will be assumed to be 1 to not break existing code, however if `num_processes` is specified, that value will no longer be ignored and accelerate will launch distributed training.

Fixes #2540 

## Before submitting
      Pull Request section?
      to it if that's the case.
- [ ] Did you write any new necessary tests? _No new test were necessary._

## Who can review?

@muellerzr


src/accelerate/launchers.py
```diff
@@ -124,14 +124,14 @@ def train(*args):
...
-        else:
-            print("Launching training on one CPU.")
-        function(*args)
     else:
+        if in_colab and num_processes is None:
+            # Previous Google Colab behavior assumed only 1 process could be available and ignored the num_processes
+            # parameter. To not break existing notebooks which have omitted this parameter, we can set it to 1 in
+            # this case.
+            num_processes = 1
+            print("Running in Colab and num_processes unspecified. Continuing assuming `num_processes=1`.")
```
muellerzr, line 133: Let's change this to keep the `elif`, and simply check if the available GPUs is > 2. Let's not be annoying to our users and print something every time they use this.

Something like:
(`get_gpu_info` comes from `accelerate.utils.env`)
```python
    elif in_colab and get_gpu_info()[1] < 2:
            # No need for a distributed launch otherwise as it's either CPU or one GPU.
            if torch.cuda.is_available():
                print("Launching training on one GPU.")
            else:
                print("Launching training on one CPU.")
            function(*args)
```

muellerzr (changes requested):
Good attempt! I think we can rework this though to align to existing API's.

Doing it the way currently suggested will allow the later `if num_processes is None` check to be caught if multiple-gpus are supported automatically.

(Note: doing it this way circumvents `torch.cuda` calls to check the number of devices, we use this in a few places when we want info like this)

src/accelerate/launchers.py
StefanTodoran: I see, that makes more sense. I've updated the PR now. Please let me know if there are any more issues.

muellerzr (approved):
Thanks! Much simpler change :)

muellerzr (approved):

PR #2569, opened 2024-03-21
Allow for setting deterministic algorithms

muellerzr:
# What does this PR do?

This PR expands `set_seed` to include `torch.use_deterministic_algorithms` for true reproducability. Comes at a caveat of can slow things down on the GPU dramatically, but improves reproducability. 

PyTorch docs: https://pytorch.org/docs/stable/notes/randomness.html#avoiding-nondeterministic-algorithms

This also is a more global approach rather than `torch.backends.cudnn.deterministic` and should be device agnostic. 

Requested in an internal slack thread: https://huggingface.slack.com/archives/C038TF8B7UZ/p1711016602277049

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/utils/random.py
+++ b/src/accelerate/utils/random.py
@@ -28,7 +28,7 @@ if is_torch_xla_available():
     import torch_xla.core.xla_model as xm
 
 
-def set_seed(seed: int, device_specific: bool = False):
+def set_seed(seed: int, device_specific: bool = False, determinsitic: bool = False):
     """
     Helper function for reproducible behavior to set the seed in `random`, `numpy`, `torch`.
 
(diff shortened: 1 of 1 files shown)


clefourrier:
I suspect `determinsitic` should be `deterministic` :wink:

BenjaminBossan (approved):
The change itself looks good, but the new functionality is a bit orthogonal to setting the seed IMO. IIRC, 100% determinism cannot be guaranteed when using concurrency, even with this setting, is that right?

I guess the main advantage I see in this PR is that users may be educated about the existence of this toggle.

Issue #2564, opened 2024-03-20
'AcceleratorState' object has no attribute 'distributed_type'


Zhuxing01:
### System Info

```Shell
accelerate == 0.28.0
python == 3.8.18
transformers == 4.37.2
torch == 2.2.1
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I was fine-tuning a bert-base model using codes from huggingface's example.I just copied the code from the website and then the bug occured. Code and bug are listed below. By the way,the pycharm told me that accelerator's arguments should be list but I give it Dataloader. 
Bug:  
AttributeError: 'AcceleratorState' object has no attribute 'distributed_type'

Code:   
```python
model = AutoModelForSequenceClassification.from_pretrained(checkpoint, num_labels=2) 
train_dataloader = DataLoader(datasets["train"], shuffle=True, batch_size=8, collate_fn=data_collator) 
eval_dataloader = DataLoader(datasets["validation"], batch_size=8, collate_fn=data_collator) 
optimizer = torch.optim.AdamW(model.parameters(), lr=3e-5)
model, optimizer, train_dataloader, eval_dataloader = accelerator.prepare(model, optimizer, train_dataloader, eval_dataloader)
```

### Expected behavior

the code works properly.

muellerzr:
Please give us the full stack trace

Zhuxing01:
full stack trace:
Traceback (most recent call last):
  File "C:\Users\Noah\PycharmProjects\pythonProject\venv\fine-tune.py", line 44, in <module>
    model, optimizer, train_dataloader, eval_dataloader = accelerator.prepare(model, optimizer, train_dataloader, eval_dataloader)
  File "E:\Anaconda\envs\pytorch\lib\site-packages\accelerate\accelerator.py", line 1219, in prepare
    if self.distributed_type == DistributedType.DEEPSPEED:
  File "E:\Anaconda\envs\pytorch\lib\site-packages\accelerate\accelerator.py", line 509, in distributed_type
    return self.state.distributed_type
AttributeError: 'AcceleratorState' object has no attribute 'distributed_type'

muellerzr:
What is your full, entire code please.

[...]

PR #2572, opened 2024-03-21
Guard stateful objects

muellerzr:
# What does this PR do?

This PR is the second half of solving https://github.com/huggingface/accelerate/issues/2564

The issue boils down to:

1. The user defines a `PartialState`/creates an `Accelerator`
2. The user then calls `_reset_state()` themselves
3. The user then calls subsequent functions in `Accelerator`.

That instance will still point to our reset/broken `PartialState`, and so a better error needs to be raised that contextualizes what's going on early:

```python
  File "/home/zach/test.py", line 11, in <module>
    _ = accelerator.prepare(model)
  File "/home/zach/accelerate/accelerate/src/accelerate/accelerator.py", line 1230, in prepare
  File "/home/zach/accelerate/accelerate/src/accelerate/accelerator.py", line 515, in distributed_type
    return self.state.distributed_type
  File "/home/zach/accelerate/accelerate/src/accelerate/state.py", line 1082, in __getattr__
    raise AttributeError(
AttributeError: `AcceleratorState` object has no attribute `distributed_type`. This can happen if the state was reset and using an old reference. If you are using an `Accelerator`, make sure to re-initialize it.
```

Fixes https://github.com/huggingface/accelerate/issues/2564

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -133,6 +133,18 @@ class PartialState:
     """
 
     _shared_state = SharedDict()
+    _known_attrs = [
+        "_cpu",
+        "_mixed_precision",
+        "backend",
+        "debug",
+        "device",
+        "distributed_type",
+        "fork_launched",
+        "local_process_index",
+        "num_processes",
+        "process_index",
+    ]
 
     def __init__(self, cpu: bool = False, **kwargs):
         self.__dict__ = self._shared_state
@@ -805,8 +817,8 @@ class PartialState:
             # If not found, give a more contextualized answer
             raise AttributeError(
                 f"`PartialState` object has no attribute `{name}`. "
-                "This can happen if the state was reset and using an old reference. "
-                "If you are using an `Accelerator`, make sure to re-initialize it."
+                "This can happen if `PartialState._reset_state()` was called and "
+                "an `Accelerator` or `PartialState` was not reinitialized."
             )
 
 
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -811,11 +811,12 @@ class PartialState:
             return torch.device("cpu")
 
     def __getattribute__(self, name: str):
-        if name not in self._known_attrs:
-            return object.__getattribute__(self, name)
+        obj_getattr = object.__getattribute__
+        if name not in obj_getattr(self, "_known_attrs"):
+            return obj_getattr(self, name)
         else:
             try:
-                return object.__getattribute__(self, name)
+                return obj_getattr(self, name)
             except AttributeError:
                 # If not found, give a more contextualized answer
                 raise AttributeError(
@@ -1095,11 +1096,12 @@ class AcceleratorState:
         PartialState().print(*args, **kwargs)
 
     def __getattribute__(self, name: str):
-        if name not in self._known_attrs:
-            return object.__getattribute__(self, name)
+        obj_getattr = object.__getattribute__
+        if name not in obj_getattr(self, "_known_attrs"):
+            return obj_getattr(self, name)
         else:
             try:
-                return object.__getattribute__(self, name)
+                return obj_getattr(self, name)
             except AttributeError:
                 # If not found, give a more contextualized answer
                 raise AttributeError(
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -797,6 +797,18 @@ class PartialState:
         else:
             return torch.device("cpu")
 
+    def __getattr__(self, name: str):
+        try:
+            # First try the normal way
+            return object.__getattribute__(self, name)
+        except AttributeError:
+            # If not found, give a more contextualized answer
+            raise AttributeError(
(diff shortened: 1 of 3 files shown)


muellerzr:
Where this breaks/is found is in the `Trainer` we call `_reset_state()` during `setup_devices`, which was the true root cause (but it is needed, so we just need to have a clearer err)

pacman100 (approved):
Thank you @muellerzr for raising better error message when the user uses broken `AcceleratorState` unintentionally!

src/accelerate/state.py
```diff
@@ -797,6 +797,18 @@ def default_device(self) -> torch.device:
         else:
             return torch.device("cpu")
 
+    def __getattr__(self, name: str):
+        try:
+            # First try the normal way
+            return object.__getattribute__(self, name)
```
BenjaminBossan, line 803: Trying to think this through: When we hit `__getattr__`, we already know the attribute wasn't found, right? So can we ever successfully `return object.__getattribute__(self, name)`? Or can we raise the below error right away?

BenjaminBossan (commented):
I guess someone complained about this and that's why we want to have this extra message? Honestly, I could live without it, not only is `_reset_state` a private method, it's docstring even explicitly warns not to use it. At some point, we have to stop hand holding :D

The disadvantage I could see in this is if user code does some other non-existing attribute access and now they get this new message, which would increase confusion, making the user think they need to re-initialize something when it's maybe just a typo or so. Also, messing with `__getattr__` is always a bit icky.

As an alternative, we could add this error message to the docstring of `_reset_state`, WDYT?

src/accelerate/state.py
muellerzr: Can confirm that yes we can, `getattr` does not know by this point, it's what is immediately hit when doing `getattr()`. This then is what does the actual getattr check.

BenjaminBossan: Hmm, are you sure? I modified the method like this:

```python
    def __getattr__(self, name: str):
        try:
            # First try the normal way
            attr = object.__getattribute__(self, name)
        except AttributeError as e:
            if name in self._known_attrs:
                # If not found, give a more contextualized answer
                raise AttributeError(
                    f"`PartialState` object has no attribute `{name}`. "
                    "This happens if `PartialState._reset_state()` was called and "
                    "an `Accelerator` or `PartialState` was not reinitialized."
                )
            else:
                raise e
        breakpoint()
        return attr
```

and never hit the breakpoint when running `test_state_after_reset` or any other test. Attribute access goes via `__getattribute__` and only calls `__getattr__` when the former fails with an `AttributeError`. Therefore, I don't think this `__getattribute__` can succeed.

I guess one reason to still leave it there is so that we can re-trigger the `AttributeError` and re-raise it instead of directly raising `AttributeError`, though I can't come up with a scenario where this is useful.

src/accelerate/state.py
```diff
@@ -797,6 +809,21 @@ def default_device(self) -> torch.device:
...
+        except AttributeError as e:
+            if name in self._known_attrs:
+                # If not found, give a more contextualized answer
+                raise AttributeError(
+                    f"`PartialState` object has no attribute `{name}`. "
+                    "This happens if `PartialState._reset_state()` was called and "
+                    "an `Accelerator` or `PartialState` was not reinitialized."
```
BenjaminBossan, line 822: Do we need "an" here?

src/accelerate/state.py
```diff
@@ -1060,6 +1087,19 @@ def local_main_process_first(self):
...
         PartialState().print(*args, **kwargs)
 
+    def __getattr__(self, name: str):
+        try:
+            # First try the normal way
+            return object.__getattribute__(self, name)
+        except AttributeError:
+            # If not found, give a more contextualized answer
+            # adjust err message
+            raise AttributeError(
```
BenjaminBossan, line 1097: Would it make sense to have the same `__getattr__` here as for `PartialState`?

BenjaminBossan (commented):
Thanks for the update, I still have some questions. Sorry for being a bit nit-picky here, but modifying attribute access can be a delicate matter and I want to ensure we get it 100% right.

src/accelerate/state.py
muellerzr: For grammar I'd say yes, but no option one way or the other

src/accelerate/state.py
muellerzr: So this is because the implementation needs to be under `__getattribute__` via `__getattr__`, as you said. Thanks :)

BenjaminBossan (approved):
Thanks, this now looks as clean as it gets to me.

My only small gripe is that there isn't really a way to ensure that `_known_attrs` is kept up-to-date, but I don't see a good solution and it wouldn't be the end of the world if an attribute is missing there.

PR #2576, opened 2024-03-21
Refactor PartialState and AcceleratorState

muellerzr:
# What does this PR do?

This PR does a *huge* refactor of the `PartialState` device setup to make it easier for new backends to get added, ease the maintenance burden, and centralizes the code based on topics

* When adding a new device, `set_device()` should be updated to include the new backend to be used and calling it
* `_prepare_backend` will handle setting `self.backend` and bringing in any imports that are needed for injecting/monkeypatching (such as `torch_ccl`)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @pacman100 

cc @faaany @huismiling if I could get you to verify on your respective devices this doesn't break anything, it'd be appreciated 🙏 

(As I can only test on CPU/multi-GPU)

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -151,6 +151,7 @@ class PartialState:
             self.device = torch.device(env_device) if env_device is not None else None
             self.debug = parse_flag_from_env("ACCELERATE_DEBUG_MODE")
             use_sagemaker_dp = kwargs.pop("_use_sagemaker_dp", None)
+            dist_information = None
             if use_sagemaker_dp is None:
                 use_sagemaker_dp = (
                     os.environ.get("ACCELERATE_USE_SAGEMAKER", "false") == "true"
@@ -198,8 +199,6 @@ class PartialState:
             if self.distributed_type in (DistributedType.MULTI_XPU, DistributedType.MULTI_CPU):
                 if not torch.distributed.is_initialized():
                     dist_information = get_cpu_distributed_information()
-                    global DISTRIBUTED_LOCAL_RANK
-                    DISTRIBUTED_LOCAL_RANK = dist_information["local_rank"]
                     os.environ.update(
                         {
                             "RANK": str(dist_information["rank"]),
@@ -262,7 +261,9 @@ class PartialState:
             else:
                 self.num_processes = torch.distributed.get_world_size()
                 self.process_index = torch.distributed.get_rank()
-                self.local_process_index = DISTRIBUTED_LOCAL_RANK
+                self.local_process_index = (
+                    DISTRIBUTED_LOCAL_RANK if dist_information is None else dist_information["local_rank"]
+                )
             self.set_device()
 
         self.fork_launched = parse_flag_from_env("FORK_LAUNCHED", 0)
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -186,48 +186,47 @@ class PartialState:
                             torch.distributed.init_process_group(backend=self.backend, **kwargs)
             # XPU and CPU require special env configs to be set
             if self.distributed_type in (DistributedType.MULTI_XPU, DistributedType.MULTI_CPU):
+                dist_information = get_cpu_distributed_information()
+                os.environ["RANK"] = str(dist_information["rank"])
+                os.environ["WORLD_SIZE"] = str(dist_information["world_size"])
+                os.environ["LOCAL_RANK"] = str(dist_information["local_rank"])
+                os.environ["LOCAL_WORLD_SIZE"] = str(dist_information["local_world_size"])
+                if self.backend == "ccl" and self.distributed_type == DistributedType.MULTI_XPU:
+                    os.environ["CCL_PROCESS_LAUNCHER"] = "none"
+                    os.environ["CCL_LOCAL_SIZE"] = os.environ["LOCAL_WORLD_SIZE"]
+                    os.environ["CCL_LOCAL_RANK"] = os.environ["LOCAL_RANK"]
+                if not os.environ.get("MASTER_PORT", None):
+                    os.environ["MASTER_PORT"] = "29500"
+                if (
+                    not os.environ.get("MASTER_ADDR", None)
+                    and dist_information["local_world_size"] != dist_information["world_size"]
+                    and self.backend != "mpi"
+                ):
+                    raise ValueError(
+                        "Tried to launch on distributed with multinode, but `MASTER_ADDR` env was not set, "
+                        "please try exporting rank 0's hostname as `MASTER_ADDR`"
+                    )
+                kwargs["rank"] = dist_information["rank"]
+                kwargs["world_size"] = dist_information["world_size"]
+
+                if (
+                    self.distributed_type == DistributedType.MULTI_CPU
+                    and get_int_from_env(["OMP_NUM_THREADS", "OMP_NUM_THREADS"], 0) > 0
+                ):
+                    import psutil
+
+                    num_cpu_threads_per_process = int(
+                        psutil.cpu_count(logical=False) / dist_information["local_world_size"]
+                    )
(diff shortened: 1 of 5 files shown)


src/accelerate/state.py
```diff
@@ -143,231 +144,118 @@ def __init__(self, cpu: bool = False, **kwargs):
...
-                            torch.xpu.set_device(self.device)
-                    elif is_mlu_available():
-                        self.device = torch.device("mlu", self.local_process_index)
-                        if self.device is not None:
-                            torch.mlu.set_device(self.device)
-                    elif is_npu_available():
-                        self.device = torch.device("npu", self.local_process_index)
-                        if self.device is not None:
-                            torch.npu.set_device(self.device)
+            # Setups up self.backend + imports
```
BenjaminBossan, line 154: Please rephrase this comment :D

src/accelerate/state.py
```diff
@@ -143,231 +144,118 @@ def __init__(self, cpu: bool = False, **kwargs):
...
+            self._prepare_backend(cpu, use_sagemaker_dp, kwargs.pop("backend", None))
```
BenjaminBossan, line 155: I like the idea of moving this to a sub-method with a limited concern. A suggestion: How about that method returns the `backend` and `distributed_type` instead of setting it on `self`? The code here would change to:

```python
backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, kwargs.pop("backend", None))
self.backend = backend
self.distributed_type = distributed_type
```

That would have the advantage that if I want to understand the code here, I don't need to check `self._prepare_backend` if I'm not interested in these two attributes, as it's immediately obvious that this method is only concerned with determining those.

src/accelerate/state.py
```diff
@@ -143,231 +144,118 @@ def __init__(self, cpu: bool = False, **kwargs):
...
-                    1,
-                )
-                self.local_process_index = local_rank
-                os.environ["RANK"] = str(rank)
-                os.environ["WORLD_SIZE"] = str(size)
-                os.environ["LOCAL_RANK"] = str(local_rank)
-                os.environ["LOCAL_WORLD_SIZE"] = str(local_size)
-
-                if backend == "ccl" and self.distributed_type == DistributedType.MULTI_XPU:
+                    if int(os.environ.get("LOCAL_RANK", -1)) != -1:
```
BenjaminBossan, line 169: Let's add a comment here similar to the one above `# Deal with XLA` to delineate the blocks.

src/accelerate/state.py
```diff
@@ -797,6 +685,74 @@ def default_device(self) -> torch.device:
...
+
+    def set_device(self):
+        """
+        Sets the device in `self.device` to the current distributed environment.
+        """
+        if self.device is None:
+            if self.num_processes == 1:
+                self.device = torch.device("cpu") if self._cpu else self.default_device
+            else:
+                device = str(self.distributed_type).split(".")[-1].replace("MULTI_", "").lower()
```
BenjaminBossan, line 738: Just a personal preference, feel free to ignore: I prefer early returns:

```python
        if self.device is not None:
            return

        if self.num_processes == 1:
            self.device = torch.device("cpu") if self._cpu else self.default_device
            return

        device = str(self.distributed_type).split(".")[-1].replace("MULTI_", "").lower()
        ...
```

src/accelerate/utils/environment.py
```diff
@@ -176,3 +176,22 @@ def check_fp8_capability():
     """
     cuda_device_capacity = torch.cuda.get_device_capability()
     return cuda_device_capacity >= (8, 9)
+
+
+def get_cpu_distributed_information():
+    """
+    Attempts to return various information about the environment in relation to CPU distributed training
```
BenjaminBossan, line 183: Why "Attempts"? This made me think that this function may raise an error if it can't find the info.

src/accelerate/utils/environment.py
```diff
@@ -176,3 +176,22 @@ def check_fp8_capability():
...
+        ["WORLD_SIZE", "PMI_SIZE", "OMPI_COMM_WORLD_SIZE", "MV2_COMM_WORLD_SIZE"], 1
+    )
+    information["local_rank"] = get_int_from_env(
+        ["LOCAL_RANK", "MPI_LOCALRANKID", "OMPI_COMM_WORLD_LOCAL_RANK", "MV2_COMM_WORLD_LOCAL_RANK"], 0
+    )
+    information["local_world_size"] = get_int_from_env(
+        ["LOCAL_WORLD_SIZE", "MPI_LOCALNRANKS", "OMPI_COMM_WORLD_LOCAL_SIZE", "MV2_COMM_WORLD_LOCAL_SIZE"],
+        1,
+    )
+    return information
```
BenjaminBossan, line 197: Maybe worthwhile to return a `dataclass` instead of a dict so that the caller knows better what it can expect?

src/accelerate/state.py
```diff
@@ -143,231 +144,118 @@ def __init__(self, cpu: bool = False, **kwargs):
...
+                                    "DeepSpeed is not available => install it using `pip3 install deepspeed` or build it from source"
+                                )
+                            from deepspeed import comm as dist
+
+                            if is_xpu_available() and is_ccl_available():
+                                os.environ["CCL_PROCESS_LAUNCHER"] = "none"
+                                os.environ["CCL_LOCAL_SIZE"] = os.environ.get("LOCAL_WORLD_SIZE", "1")
+                                os.environ["CCL_LOCAL_RANK"] = os.environ.get("LOCAL_RANK", "0")
+                            dist.init_distributed(dist_backend=self.backend, auto_mpi_discovery=False, **kwargs)
+                            self.set_device()
```
BenjaminBossan, line 182: Just want to ensure that this `set_device()` call is correct here, since it is called again further down below. If it's indeed needed here, maybe add a comment to explain?

muellerzr: It is, we need it to happen before the distributed type is set to deepspeed (and if it's called again after the first time, nothing happens)

src/accelerate/state.py
muellerzr: Done

src/accelerate/state.py
muellerzr: Makes sense to me, good suggestion.

src/accelerate/state.py
```diff
@@ -797,6 +691,79 @@ def default_device(self) -> torch.device:
...
+                else:
+                    import torch_ccl  # noqa: F401
+
+                backend = "ccl"
+            elif torch.distributed.is_mpi_available():
+                backend = "mpi"
+            else:
+                backend = "gloo"
+        else:
+            distributed_type = DistributedType.NO
```
huismiling, line 737: This makes MLU devices `distributed_type = DistributedType.NO`.
distributed_type may be initialized at function head.

huismiling (commented):
With distributed_type initialized at _prepare_backend function head, nlp_example.py with MLUs is passed !

src/accelerate/state.py
muellerzr: Thanks, I'll look into seeing what I did wrong here. Thought I maintained the behavior but it seems like I did not

muellerzr: Should be fixed now!

src/accelerate/utils/environment.py
muellerzr: Done, good idea.

I kept it in this class as it's not a useful dataclass outside of here (and not to mention circular imports)

faaany:
Hi @muellerzr , sorry for the late response! I was OOO in the last few days. I tested the patch on XPU for training and it works. Great job! The code looks much cleaner now.

muellerzr:
Final integration tests are being performed here: https://github.com/huggingface/accelerate/actions/runs/8470701075

PR #2580, opened 2024-03-22
[docs] Missing functions from API

stevhliu:
Adds missing functions like `gather_object` which is in a separate file `accelerate/utils/operations.py` rather than the main `accelerate/accelerator.py`. This results in `gather_object` not being documented because the `[[autodoc]]` tag couldn't find it in `accelerate/accelerator.py`.


muellerzr (approved):
Thanks!

PR #2588, opened 2024-03-26
Fix `load_checkpoint_in_model` behavior when unexpected keys are in the checkpoint

fxmarty:
Up to now, `load_checkpoint_in_model` would behave differently when the checkpoint to be loaded has unexpected keys compared to the model state_dict keys.

With `device_map=None` input, we are currently not erroring out, while with `device_map` not None, we are erroring out.

This PR aligns the behavior of both, defaulting to `strict=False`, and logs the unloaded keys.

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1585,7 +1585,8 @@ def load_checkpoint_in_model(
         offload_8bit_bnb (`bool`, *optional*):
             Whether or not to enable offload of 8-bit modules on cpu/disk.
         strict (`bool`, *optional*, defaults to `False`):
-            Whether to strictly enforce that the keys in the checkpoint state_dict match the keys of the model's state_dict.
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -1660,17 +1663,25 @@ def load_checkpoint_in_model(
         state_dict_folder = tempfile.mkdtemp()
         state_dict_index = {}
 
+    unexpected_keys = set()
+    model_keys = set(model.state_dict().keys())
     buffer_names = [name for name, _ in model.named_buffers()]
     for checkpoint_file in checkpoint_files:
-        checkpoint = load_state_dict(checkpoint_file, device_map=device_map)
+        loaded_checkpoint = load_state_dict(checkpoint_file, device_map=device_map)
```
fxmarty, line 1670: This variable is renamed because it overrides the function input `checkpoint`, which is bug prone.

SunMarc (approved):
LGTM ! Thanks for fixing this !

PR #2590, opened 2024-03-26
[docs] Fix kwarg docstring

stevhliu:
Fixes the format of the `kwargs` docstrings.

### DIFF
--- a/src/accelerate/utils/deepspeed.py
+++ b/src/accelerate/utils/deepspeed.py
@@ -234,7 +234,7 @@ class DummyOptim:
             parameter groups
         weight_decay (float):
             Weight decay.
-        **kwargs:
+        **kwargs (additional keyword arguments, *optional*):
             Other arguments.
     """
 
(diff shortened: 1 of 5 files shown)


src/accelerate/utils/deepspeed.py
```diff
@@ -234,7 +234,7 @@ class DummyOptim:
             parameter groups
         weight_decay (float):
             Weight decay.
-        **kwargs:
+        **kwargs (additional keyword arguments, *optional*):
```
muellerzr, line 237: Does the `**` make a difference?

muellerzr (approved):
Thanks for doing this! Just one question

src/accelerate/utils/deepspeed.py
stevhliu: Nope, not in terms of formatting! `**kwargs` still shows up as expected. We can remove the `**` if you want :)

Screenshot 2024-03-26 at 12 39 09 PM

stevhliu: I got you 👍

PR #2594, opened 2024-03-28
Fix warning log for unused checkpoint keys

fxmarty:
As per title, small typo in https://github.com/huggingface/accelerate/pull/2588


muellerzr (approved):
Ah, yeah, we definitely only want to warn when relevant

SunMarc (approved):
Thanks, I missed it ;)

PR #2595, opened 2024-03-28
Allow for force unwrapping

muellerzr:
# What does this PR do?

There are cases when we need to unwrap *far more* than just what is done natively in Accelerate, such as when we have TPUs/SPMD.

This PR expands `extract_model_from_parallel` to support situations such as using XLA's `FSDPv2`.

Alternative to https://github.com/huggingface/transformers/pull/29780, which we can upstream to transformers by just using `recursive=True`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @shub-kris 

We can't test it on TPUs since we don't have our runners, but @shub-kris tested this himself :)

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -90,7 +90,7 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive
         model = model.module
 
     if recursive:
-
+        # This is needed in cases such as using FSDPv2 on XLA
         def _recursive_unwrap(module):
             if hasattr(module, "module"):
                 unwrapped_module = _recursive_unwrap(module.module)
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -53,7 +53,7 @@ def is_compiled_module(module):
     return isinstance(module, torch._dynamo.eval_frame.OptimizedModule)
 
 
-def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
+def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive: bool = False):
     """
     Extract a model from its distributed containers.
 
@@ -62,6 +62,9 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
             The model to extract.
         keep_fp32_wrapper (`bool`, *optional*):
             Whether to remove mixed precision hooks from the model.
+        recursive (`bool`, *optional*, defaults to `False`):
+            Whether to recursively extract all cases of `module.module` from `model` as well as unwrap child sublayers
+            recursively, not just the top-level distributed containers.
 
     Returns:
         `torch.nn.Module`: The extracted model.
@@ -85,6 +88,21 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
 
     while isinstance(model, options):
         model = model.module
+
+    if recursive:
+
+        def _recursive_unwrap(module):
+            if hasattr(module, "module"):
+                unwrapped_module = _recursive_unwrap(module.module)
+            else:
+                unwrapped_module = module
+            # Next unwrap child sublayers recursively
+            for name, child in unwrapped_module.named_children():
+                setattr(unwrapped_module, name, _recursive_unwrap(child))
+            return unwrapped_module
+
+        # Start with top-level
+        model = _recursive_unwrap(model)
 
     if not keep_fp32_wrapper:
         forward = model.forward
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -193,6 +206,31 @@ def test_extract_model(self):
 
         assert model == model_unwrapped
 
+    @require_tpu
+    @require_huggingface_suite
+    def test_extract_model_recursive_fsdpv2(self):
```
muellerzr, line 211: While we don't have runners for this test, @shub-kris verified it works
```
One more thing that needs to be in the test which I missed is: One has to make sure that you set this two environment variables:
export PJRT_DEVICE=TPU XLA_USE_SPMD=1
I encountered the error when I didn't set XLA_USE_SPMD=1
 Please enable SPMD via `torch_xla.runtime.use_spmd()`
```

tests/test_utils.py
```diff
@@ -193,6 +206,31 @@ def test_extract_model(self):
...
+        # reported in https://github.com/huggingface/transformers/pull/29780
+        xr.use_spmd()
+        from transformers import AutoModelForCausalLM
+
+        model = AutoModelForCausalLM.from_pretrained("gpt2")
+        orig_state_dict_keys = list(model.state_dict().keys())
+        num_devices = xr.global_runtime_device_count()
+        xs.set_global_mesh(xs.Mesh(np.array(range(num_devices)), (num_devices, 1), axis_names=("fsdp", "tensor")))
+
+        def nested_wrap(model):
```
BenjaminBossan, line 222: Could you add a comment on why this is needed?

src/accelerate/utils/other.py
```diff
@@ -86,6 +89,21 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
     while isinstance(model, options):
         model = model.module
 
+    if recursive:
+        # This is needed in cases such as using FSDPv2 on XLA
+        def _recursive_unwrap(module):
+            if hasattr(module, "module"):
```
BenjaminBossan, line 95: Could you add a comment on where `module.module` is coming from?

BenjaminBossan (approved):
Overall LGTM to me thanks. I only think it could be a bit better commented for future clarity.

Regarding the testing, good to have confirmation that this fixes the issue. I guess we could come up with an artificial test case that has `module.module` and works without TPU. Not sure how useful that would be.

src/accelerate/utils/other.py
muellerzr: Done

tests/test_utils.py
muellerzr: Done, @shub-kris can you verify my reasoning behind it is correct?

src/accelerate/utils/other.py
```diff
@@ -86,6 +89,23 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True):
     while isinstance(model, options):
         model = model.module
 
+    if recursive:
+        # This is needed in cases such as using FSDPv2 on XLA
```
SunMarc, line 93: Is there a way to detect that we are using fsdpv2 or xla , so that we don't have to set recursive. cc @shub-kris

SunMarc (approved):
LGTM !

src/accelerate/utils/other.py
muellerzr: There are, but accelerate doesn’t support it yet and it’s experimental. So for now a generic approach, as it’s a single location in the Trainer

tests/test_utils.py
shub-kris: I will do a test-run on this https://github.com/huggingface/transformers/issues/29659 to make sure training and loading of model works properly.

tests/test_utils.py
shub-kris: Yes, it's correct and to the point.

shub-kris (commented):
Thanks for the nice PR @muellerzr. Great work. 
I will approve the PR once I do a small training to verify everything works end-to-end.

tests/test_utils.py
shub-kris: @muellerzr shouldn't we add that in the test function that it should be ran with 
```bash
export PJRT_DEVICE=TPU XLA_USE_SPMD=1
```

shub-kris: > I will do a test-run on this https://github.com/huggingface/transformers/issues/29659 to make sure training and loading of model works properly.

Ahh I can't do as the changes aren't there yet in the `transformers` library.

shub-kris (approved):

Issue #2592, opened 2024-03-27
`torch.compile` not working with `device_map` and multiple GPUs


sheepymeh:
### System Info

```Shell
- `Accelerate` version: 0.28.0
- Platform: Linux-5.15.0-101-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.2.1+cu121 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 440.57 GB
- GPU type: NVIDIA GeForce RTX 3090
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am unable to use `torch.compile` with `device_map` when using multiple GPUs. I believe this issue is related to #2387.

This can be reproduced with the following (the model does not seem to matter):

```python
from transformers import ViTImageProcessor, ViTForImageClassification
from PIL import Image
import requests

url = "http://images.cocodataset.org/val2017/000000039769.jpg"
image = Image.open(requests.get(url, stream=True).raw)

processor = ViTImageProcessor.from_pretrained("google/vit-base-patch16-224")
model = ViTForImageClassification.from_pretrained("google/vit-base-patch16-224", device_map="auto")
model.compile(fullgraph=True)

inputs = processor(images=image, return_tensors="pt").to(model.device)
outputs = model(**inputs)
logits = outputs.logits
# model predicts one of the 1000 ImageNet classes
predicted_class_idx = logits.argmax(-1).item()
print("Predicted class:", model.config.id2label[predicted_class_idx])
```

<details>
<summary>Stack trace</summary>

```
Traceback (most recent call last):
  File "/home/sheepymeh/transformers/test_vit.py", line 13, in <module>
    outputs = model(**inputs)
  File "/home/sheepymeh/transformers/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1509, in _wrapped_call_impl
    return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
  File "/home/sheepymeh/transformers/venv/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 489, in _fn
[...]

PR #2609, opened 2024-04-02
make big model inference compatible with torch.compile

SunMarc:
# What does this PR do ?
This PR makes big model inference compatible with `torch.compile`. To do that, we need to fix `send_to_device`.

In pytorch 2.2.1, `hasattr` is not implemented for TupleVariable, so we get an error in `is_namedtuple` function. I tried a few methods to not use `hasattr` but I didn't manage to make it work (e.g. try except block). If we bypass the `is_namedtuple` function, compile works fine. In pytorch nightly (future 2.4), this is implemented but we hit another error ( breaking changes ... ). So, I need to investigate more.

Fixes https://github.com/huggingface/accelerate/issues/2592
Draft for now to not forget

EDIT: Feel free to try this PR. It should work for the latest torch. However, it doesn't support namedtuple.


Aknifejackzhmolong:
Hi, I've hit the exact same error. Any news on this?

SunMarc:
Recent version of torch works out of the box with the existing code. Closing this for now

Issue #2575, opened 2024-03-21
TypeError when using prepare_pippy with split_points="auto" in LLaMA example


AlexSalinas99:
I'm encountering a TypeError when using the prepare_pippy function with split_points="auto" in the LLaMA inference example provided in the accelerate repository. The error occurs when I try to prepare the model for pipeline parallelism. I'm using the following versions of the libraries:

accelerate: 0.28.0
transformers: 4.38.2
torchpippy: 0.2.0
The code I'm running is exactly as provided in the [llama.py example](https://github.com/huggingface/accelerate/blob/main/examples/inference/llama.py) in the accelerate repository.

The error message I receive is as follows:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[2], line 9
      7 tokenizer.pad_token = tokenizer.eos_token
      8 inputs = tokenizer(prompts, return_tensors="pt", padding=True)
----> 9 model = prepare_pippy(model, split_points="auto", example_args=inputs)
     10 with torch.no_grad():
     11     output = model(**inputs)

File ~/venvs/audit_llms/lib/python3.11/site-packages/accelerate/inference.py:174, in prepare_pippy(model, split_points, no_split_module_classes, example_args, example_kwargs, num_chunks, gather_output)
    172         split_points.append(next(k for k, v in device_map.items() if v == i))
    173 model.hf_split_points = split_points
--> 174 stage = build_pipeline(model, split_points, example_args, example_kwargs, num_chunks)
    175 model._original_forward = model.forward
    176 model._original_call = model.__call__

File ~/venvs/audit_llms/lib/python3.11/site-packages/accelerate/inference.py:92, in build_pipeline(model, split_points, args, kwargs, num_chunks)
     90 if found_batch_size != num_chunks:
     91     if args is not None:
---> 92         args = pad_input_tensors(args, found_batch_size, num_chunks)
     93     if kwargs is not None:
     94         kwargs = pad_input_tensors(kwargs, found_batch_size, num_chunks)

File ~/venvs/audit_llms/lib/python3.11/site-packages/accelerate/utils/operations.py:716, in pad_input_tensors(tensor, batch_size, num_processes, dim)
    713     new_tensor[indices] = tensor
    714     return new_tensor
--> 716 return recursively_apply(
    717     _pad_input_tensors,
    718     tensor,
    719     error_on_other_type=True,
    720     batch_size=batch_size,
    721     num_processes=num_processes,
[...]

PR #2616, opened 2024-04-03
fix llama example for pippy

SunMarc:
# What does this PR do ?
This PR fixes the llama example for pippy. I removed the llama test since we are unable to download the gated model.

Fixes https://github.com/huggingface/accelerate/issues/2575


muellerzr (approved):
Nice :)

PR #2630, opened 2024-04-06
Support schedulefree optimizer

sdbds:
# What does this PR do?

Support schedulefree optimizer
https://github.com/facebookresearch/schedule_free

it used `optimizer.train()`


muellerzr:
Hi! Thanks for your PR, I already have https://github.com/huggingface/accelerate/pull/2631 which has a bit more of a robust solution and an example

sdbds:
> Hi! Thanks for your PR, I already have #2631 which has a bit more of a robust solution and an example

Thank you for great work~

PR #2631, opened 2024-04-06
Schedule free optimizer support

muellerzr:
# What does this PR do?

Facebook recently released an optimizer that doesn't require an LR scheduler. This PR adds support in by enabling optimizers to call `.train()` and `.eval()` as a passthrough.

See https://github.com/facebookresearch/schedule_free/tree/main

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -121,6 +121,18 @@ class AcceleratedOptimizer(torch.optim.Optimizer):
                     raise ValueError("`set_to_none` for Optimizer.zero_grad` is not supported by this optimizer.")
                 self.optimizer.zero_grad()
 
+    def train(self):
+        """
+        Sets the optimizer to "train" mode. Useful for optimizers like `schedule_free`
+        """
+        return self.optimizer.train()
+
+    def eval(self):
+        """
+        Sets the optimizer to "eval" mode. Useful for optimizers like `schedule_free`
+        """
+        return self.optimizer.eval()
+
     def step(self, closure=None):
         if (
             not self.gradient_state.is_xla_gradients_synced
--- a/examples/by_feature/schedule_free.py
+++ b/examples/by_feature/schedule_free.py
@@ -1,0 +1,211 @@
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


muellerzr:
Some general observations off the nlp_example:

Seems to converge faster, here is our baseline vs the new example, read as (accuracy, f1 score):
```
Basic nlp_example.py:
78.92%, 0.8327
84.8%, 0.8927
85.5%, 89.87

With schedule-free:
79.16%, 0.8622
85.29%, 0.8976
85.29%, 0.8972
```

pacman100 (approved):
Thank you @muellerzr for adding `train` and `eval` methods required for scheduler free optimizers! 🚀

muellerzr:
Nope, just don't pass a scheduler to `Accelerator.prepare()` :)

Not a very complicated implementation needed here

src/accelerate/optimizer.py
```diff
@@ -121,6 +121,18 @@ def zero_grad(self, set_to_none=None):
...
+        Sets the optimizer to "train" mode. Useful for optimizers like `schedule_free`
+        """
+        return self.optimizer.train()
+
+    def eval(self):
+        """
+        Sets the optimizer to "eval" mode. Useful for optimizers like `schedule_free`
+        """
+        return self.optimizer.eval()
+
```
BenjaminBossan, line 135: How about adding a `@property` for `self.optimizer.training` too? I don't think we also need a setter for this, as `train()` and `eval()` should be enough.

BenjaminBossan (approved):
LGTM, thanks for adding this feature and a script to test it. Btw. any idea what would happen if a user forgets to call `optimizer.eval()`?

I have a comment suggesting to add a `training` property too, but no blocker.

src/accelerate/optimizer.py
muellerzr: That gets to be optimizer-specific, so not a fan of it unless its downstreamed, as they currently don't have that: https://github.com/facebookresearch/schedule_free/blob/main/schedulefree/adamw_schedulefree.py#L86

(Otherwise I'd agree, yes that's a good idea)

BenjaminBossan: Ah I see, good point.

Issue #2640, opened 2024-04-09
allow `strict=True` in `load_checkpoint_and_dispatch`


yiyixuxu:
### System Info

```Shell
accelerate = 0.29.1
```

### Information

this PR https://github.com/huggingface/accelerate/pull/2588 breaks the loading behavior in diffusers when:
1. `accelerate.load_checkpoint_and_dispatch` is used
2. the checkpoint contains deprecated weight names for the attention class 

As you can see [here](https://github.com/huggingface/diffusers/blob/7e808e768a7305b3f8f9e981ad14f2de3598e9a6/src/diffusers/models/modeling_utils.py#L703), we rely on the error to rename the parameters' names on our end. Is there a better way to do this in diffusers? 
The alternative is to allow us a pass a `strict=True` from `load_checkpoint_and_dispatch`- let me know what you think :)

### Reproduction

this script fails currently but it should not 
```python
from diffusers import DiffusionPipeline
pipe = DiffusionPipeline.from_pretrained(
    "hf-internal-testing/tiny-stable-diffusion-pipe", device_map="sequential", safety_checker=None
)
```

yiyixuxu:
cc @SunMarc

SunMarc:
Let's just pass `strict=True` in `load_checkpoint_and_dispatch`. Sorry for introducing a breaking change ;)

PR #2641, opened 2024-04-09
add strict arg to load_checkpoint_and_dispatch

SunMarc:
# What does this PR do ?
This PR adds the strict arg in `load_checkpoint_and_dispatch`. We need this since this [PR](https://github.com/huggingface/accelerate/pull/2588) introduced a breaking change in the way to error out when the keys in the checkpoint state_dict do not match the keys of the model's state_dict. 
Fixes https://github.com/huggingface/accelerate/issues/2640#issuecomment-2045579806 cc @yiyixuxu 
maybe worth doing a patch @muellerzr ? There are other issues in diffusers so let's wait a bit if we indeed do a patch.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -508,6 +508,7 @@ def load_checkpoint_and_dispatch(
     skip_keys: Optional[Union[str, List[str]]] = None,
     preload_module_classes: Optional[List[str]] = None,
     force_hooks: bool = False,
+    strict: bool = False,
 ):
     """
     Loads a (potentially sharded) checkpoint inside a model, potentially sending weights to a given device as they are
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks I agree its worth doing a patch, we also should probably run some diffusers stuff in our integration tests that we can run as part of the CI to check for breaking. @sayakpaul do you know an easy subset/file we can run just for sanity testing on `main`? For transformers we run the `test_trainer.py` file

sayakpaul:
@muellerzr 

Do the tests and our [`examples` directory](https://github.com/huggingface/diffusers/tree/main/examples/) work?

muellerzr:
@sayakpaul if they broke due to what this PR is fixing, yes :)

PR #2652, opened 2024-04-11
fix backend check

jiqing-feng:
Hi @muellerzr . This PR fixed the backend assignment. It avoids confusing backend checks and distributed_type checks. We should assign distributed_type by env first and then decide whether to assign backend.

Do you know how to add a proper test to avoid this issue? Please let me know, and I will add it : )

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -738,7 +738,7 @@ class PartialState:
                 distributed_type = DistributedType.MULTI_XPU
             else:
                 distributed_type = DistributedType.MULTI_CPU
-            if backend is None:
+            if backend is None or backend == "ccl":
                 if is_ccl_available() and (
                     get_int_from_env(["CCL_WORKER_COUNT"], 0) > 0 or distributed_type == DistributedType.MULTI_XPU
                 ):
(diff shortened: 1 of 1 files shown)


muellerzr (commented):
Sure, I'm open to this, so long as the tests pass. The tests specifically are all of the tests, which are run on XPU and multi-CPU hardware. 

A simple check is `accelerate test` if you have the hardware available. If not let me know

jiqing-feng:
> Sure, I'm open to this, so long as the tests pass. The tests specifically are all of the tests, which are run on XPU and multi-CPU hardware.
> 
> A simple check is `accelerate test` if you have the hardware available. If not, let me know

Hi @muellerzr, The `accelerate test` runs well. We can merge this PR to fix the problem as soon as possible, but we'd better design a test in CI to check that the configs are correctly assigned considering the long term.

zhangsheng377:
@muellerzr 
This modification changes all MULTI_GPU and MULTI_NPU to MULTI_CPU, right?

muellerzr:
Yes it does, I didn’t have a chance to run it before merging on the CI, I can see many failures. Reverting

muellerzr:
I’ll look at trying to do an alternative today

jiqing-feng:
I know why it failed.... The best way is to return in each branch. I will update a another PR to fix it quickly. Sorry for the inconvinient.

muellerzr:
No worries @jiqing-feng :) Lots of moving parts. 

I'll make sure to test it locally next time as well to make sure non-multi-cpu's don't break

PR #2654, opened 2024-04-11
`tqdm`: `*args` should come ahead of `main_process_only`

rb-synth:
# What does this PR do?

This allows for more intuitive use of `accelerate.utils.tqdm`. Prior to this PR, despite this being the normal usage of tqdm, the following wouldn't work:

```python
from accelerate.utils import tqdm
import time
for i in tqdm(range(10)):
    time.sleep(0.1)
```

Instead, one of the following was needed:

```python
from accelerate.utils import tqdm
import time
for i in tqdm(True, range(10)):
    time.sleep(0.1)

for i in tqdm(main_process_only=True, iterable=range(10)):
    time.sleep(0.1)

for i in tqdm(iterable=range(10)):
    time.sleep(0.1)
```

I would argue that the most frequently used argument to `tqdm` is the iterable. It would nice to be able to use the first example so that it is compatible with vanilla tqdm.

This PR addresses that.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr @BenjaminBossan


muellerzr:
Nice! Can you verify setting it to False works? Honestly didn’t know if we could do this or not :)

rb-synth:
Like this? The following works:

```python
from accelerate.utils import tqdm
import time
for i in tqdm(range(10), main_process_only=False):
    time.sleep(0.1)
```

rb-synth:
What would be a good unit test? Just run the code and check no error?

muellerzr:
IIRC tqdm has a progress bar catcher. In my head we check that the caught progress bar after a few iterations has stepped on the right device, and hasn’t/remains disabled on the others. 

This would need to go in the test_utils.py file inside of src/accelerate/test_utils/scripts. 

I can also make it after this PR if it’s confusing :)

muellerzr:
Not quite my intention, see here: https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/scripts/test_ops.py

and it can be called in tests/test_utils.py

rb-synth:
Not quite as involved as your suggestion, but I checked that all the permutations worked as expected: https://github.com/huggingface/accelerate/pull/2654/commits/a688f5578281a56400e98ba74d6832a04670d78d

BenjaminBossan (commented):
I agree that the previous signature was not ideal and the new one is a better choice.

Before merging, a question to @muellerzr: Would this function be considered to be "public" accelerate API? If yes, this has the potential to break existing code. For example, if user code calls `tqdm(False, range(10))`, `args` would become the tuple `(False, range(10))`, and `main_process_only` would stay `True`. This can lead to unforeseen bugs. We could add some heuristic to check for this faulty call and warn, but I don't think we can auto-correct it.

If it's considered private, the change should be fine.

rb-synth:
I agree that this probably constitutes a breaking change. 

I suspect that we could use a deprecated decorator and switch the arguments if the first argument is a boolean (and display a deprecated warning). We'd have to account for both positional and keyword arguments. Seems like a bit of a headache for a simple function.

muellerzr:
> I suspect that we could use a deprecated decorator and switch the arguments if the first argument is a boolean (and display a deprecated warning). We'd have to account for both positional and keyword arguments. Seems like a bit of a headache for a simple function.

This is exactly the way to go, deprecating it out 3 versions. There's no way to accidentally do this, as `tqdm` expects an `iterable` on the first argument. 

It's a headache, but compatibility and smooth deprecation outweigh anything else :)

rb-synth:
I don't know the `accelerate` codebase well enough to know what you have in place for deprecation. I'll leave the rest up to you. Feel free to add to this PR (no idea if you'll have push access to my branch), or feel free to close this and open a modified PR. Thanks for the hard work!

muellerzr (approved):
Thanks for doing this! 🤗

PR #2655, opened 2024-04-11
[DO NOT REVIEW] More fixings for FP8

muellerzr:


raghukiran1224:
Do you have any early measurements? We have seen the float8 experimental from Meta is benchmarked at only 1.2x with compile. One of the bottlenecks was gradients are being exchanged in 16bit. Is that different?

muellerzr:
1.2ish isn't far off.

This is different, FP8 has already been part of Accelerate, this fix is just to make sure that the gradients are actually autocast with bf16 instead of just keeping everything in fp32 but the fp8 layers. (but upcasting back to FP32 to keep accuracy in check)

Toying around with lower precision I'd recommend via MS-AMP if you want to try the bleeding edge of doing things in lower precision (if you have MS-AMP installed, accelerate uses it automatically). 

E.g. here's some numbers off the nlp_example.py:

Memory:
```
Current implementation of FP8/FP32:  2447.78 MB
W/o fix to autocast: 2109.06 MB
```
Metrics:
```
BF16/Baseline:
epoch 0: {'accuracy': 0.8308823529411765, 'f1': 0.8715083798882681}
epoch 1: {'accuracy': 0.8578431372549019, 'f1': 0.8999999999999999}
epoch 2: {'accuracy': 0.8529411764705882, 'f1': 0.8954703832752613}

Current implementation/FP32: 
epoch 0: {'accuracy': 0.8046875, 'f1': 0.8478701825557811}
epoch 1: {'accuracy': 0.8567708333333334, 'f1': 0.8975791433891992}
epoch 2: {'accuracy': 0.8541666666666666, 'f1': 0.8974358974358975}

W/ fixes
epoch 0: {'accuracy': 0.78125, 'f1': 0.8205128205128206}
epoch 1: {'accuracy': 0.8541666666666666, 'f1': 0.8970588235294118}
epoch 2: {'accuracy': 0.859375, 'f1': 0.901098901098901}
```

So this will reduce some of the memory needed, but I found that raw weights in BF16 leads to a degredation in performance, so we're still keeping some weights in FP32 fully. Checkout this doc for more info on MS-AMP and TransformerEngine in regards to how each does lower precision: https://huggingface.co/docs/accelerate/concept_guides/low_precision_training#a-quick-chart

src/accelerate/accelerator.py
```diff
@@ -1456,6 +1460,15 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                 model = torch.nn.parallel.DistributedDataParallel(model, **kwargs)
             elif self.distributed_type == DistributedType.XLA and self.state.fork_launched:
                 model = xmp.MpModelWrapper(model).to(self.device)
+        # Now we can apply the FP8 autocast
+        if (
+            self.distributed_type in [DistributedType.MULTI_GPU, DistributedType.FSDP]
+            and self.mixed_precision == "fp8"
+            and self.fp8_recipe_handler.backend == "TE"
+        ):
```
BenjaminBossan, line 1468: I'd be a bit cautious of this conditional getting out of sync with the code above, which is 100 LOC away. How about we set a variable like `need_fp8_autocast = False`, and then after line 1363, we set `else: need_fp8_autocast = True`. Then, we only need to check that value here and don't run the risk of the checks diverging.

BenjaminBossan (commented):
I don't have any general comments, as I haven't worked with FP8 yet, but I added a small suggestion for making the code more robust.

BenjaminBossan (approved):
Thanks, LGTM.

PR #2667, opened 2024-04-14
Megatron plugin can support NPU

zhangsheng377:
We are using accelerate's megatron plugin on the NPU.

Specifically, there are two scenarios:
1. After we `import torch_npu`, `torch.cuda.is_available()` is False, which is the `MULTI_NPU` branch.
2. When we `import transfer_to_npu from torch_npu.contrib`, `torch.cuda.is_available()` is True, which is the `MULTI_GPU` branch.

Therefore, in fact, accelerate's megatron plugin supports both `MULTI_GPU` and `MULTI_NPU` scenarios.


zhangsheng377:
@muellerzr

muellerzr (commented):
Just to be sure, have we verified that it works on that device? (Training or some minimal script using megatron). (We don't have any we can run tests on)

muellerzr (approved):

PR #2670, opened 2024-04-15
fix backend check

jiqing-feng:
Hi @muellerzr 
Related to #2652 . I think we'd better have a test on it.

I think the best way is to have an end-to-end test in the CI; it will avoid most issues : )

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -179,7 +179,10 @@ class PartialState:
                 )
 
             # Sets up self.backend + imports
-            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, kwargs.pop("backend", None))
+            original_backend = kwargs.pop("backend", None)
+            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, original_backend)
+            if original_backend and backend != original_backend:
+                logger.warning(f"The assigned backend is {original_backend}, but the real backend is {backend}")
             self.backend = backend
             self.distributed_type = distributed_type
             use_deepspeed = False
@@ -730,7 +733,8 @@ class PartialState:
                 elif is_npu_available():
                     backend = "hccl"
                     distributed_type = DistributedType.MULTI_NPU
-        if backend is None and (
+
+        if distributed_type is None and (
             int(os.environ.get("LOCAL_RANK", -1)) != -1
             or get_int_from_env(["PMI_SIZE", "OMPI_COMM_WORLD_SIZE", "MV2_COMM_WORLD_SIZE", "WORLD_SIZE"], 1) > 1
         ):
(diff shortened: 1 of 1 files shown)


src/accelerate/state.py
```diff
@@ -718,41 +721,45 @@ def _prepare_backend(
...
         ):
             if not cpu and is_xpu_available():
                 distributed_type = DistributedType.MULTI_XPU
             else:
                 distributed_type = DistributedType.MULTI_CPU
-            if is_ccl_available() and (
-                get_int_from_env(["CCL_WORKER_COUNT"], 0) > 0 or distributed_type == DistributedType.MULTI_XPU
+
+            if (
+                (backend is None or backend == "ccl")
```
muellerzr, line 746:
```suggestion
                backend in (None, "ccl")
```

src/accelerate/state.py
```diff
@@ -718,41 +721,45 @@ def _prepare_backend(
...
+                and (get_int_from_env(["CCL_WORKER_COUNT"], 0) > 0 or distributed_type == DistributedType.MULTI_XPU)
             ):
                 if get_ccl_version() >= "1.12":
                     import oneccl_bindings_for_pytorch  # noqa: F401
                 else:
                     import torch_ccl  # noqa: F401
 
                 backend = "ccl"
-            elif torch.distributed.is_mpi_available():
+            elif (backend is None or backend == "mpi") and torch.distributed.is_mpi_available():
```
muellerzr, line 756:
```suggestion
            elif (backend in (None, "mpi")) and torch.distributed.is_mpi_available():
```

src/accelerate/state.py
```diff
@@ -179,7 +179,10 @@ def __init__(self, cpu: bool = False, **kwargs):
                 )
 
             # Sets up self.backend + imports
-            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, kwargs.pop("backend", None))
+            original_backend = kwargs.pop("backend", None)
+            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, original_backend)
+            if original_backend and backend != original_backend:
+                logger.warning(f"The assigned backend is {original_backend}, but the real backend is {backend}")
```
muellerzr, line 185: I don't think it makes sense to do a warning here. This should be a `raise` if this is truly a situation we can get into.

muellerzr (commented):
Thanks! This looks a bit better on paper at least, I'll pull it down and test locally before giving the ✅ 

Left a few logic and style nits

src/accelerate/state.py
jiqing-feng: Done

muellerzr: I don't see this change propagated?

jiqing-feng: Sorry, I forgot to push. It should be updated now, pls check. Thx.

src/accelerate/state.py
```diff
@@ -179,7 +179,10 @@ def __init__(self, cpu: bool = False, **kwargs):
                 )
 
             # Sets up self.backend + imports
-            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, kwargs.pop("backend", None))
+            original_backend = kwargs.pop("backend", None)
+            backend, distributed_type = self._prepare_backend(cpu, use_sagemaker_dp, original_backend)
+            if original_backend and backend != original_backend:
```
muellerzr, line 184: One last final check, sorry missed this the first time. Let's not trust python here 😉 
```suggestion
            if original_backend is not None and backend != original_backend:
```

muellerzr (approved):
Thanks so much! One last, last thing then we can merge and get a patch out ✅

src/accelerate/state.py
jiqing-feng: Done.

muellerzr (approved):
Thanks for fixing!

PR #2672, opened 2024-04-15
add distributed examples

SunMarc:
# What does this PR do ? 

This PR adds a more complete example on how to perform distributed inference in the case we load the entire model on each gpu. I've also added these scripts in our test suite.


examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+# limitations under the License.
+
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+from accelerate import PartialState
+from accelerate.utils import gather_object
+
+
+distributed_state = PartialState()
```
muellerzr, line 22: Let's add a comment here about what we're doing.

```suggestion
# Start up the distributed environment without needing the Accelerator.
distributed_state = PartialState()
```

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+
+# You can change the model to any LLM such as mistralai/Mistral-7B-v0.1 or meta-llama/Llama-2-7b-chat-hf
+model_name = "microsoft/phi-2"
+model = AutoModelForCausalLM.from_pretrained(
+    model_name, device_map=distributed_state.device, torch_dtype=torch.float16
+)
+
+tokenizer = AutoTokenizer.from_pretrained(model_name)
+# needed for geenration
```
muellerzr, line 31:
```suggestion
# Need to set the padding token to the eos token for generation
```

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+prompts = [
+    "I would like to",
+    "hello how are you",
+    "what is going on",
+    "roses are red and",
+    "welcome to the hotel",
+]
+
+batch_size = 2
+pad_to_multiple_of = 8
```
muellerzr, line 43: Let's add a comment as to how `batch_size` and `pad_to_multiples_of` are chosen please

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+
+# split into batch
```
muellerzr, line 45:
```suggestion
# Split into batches
```

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+formatted_prompts = [prompts[i : i + batch_size] for i in range(0, len(prompts), batch_size)]
+
+# do the padding on the left since we are doing generation
```
muellerzr, line 48:
```suggestion
# Apply padding on the left since we are doing generation
```

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+padding_side_default = tokenizer.padding_side
+tokenizer.padding_side = "left"
+# tokenize each batch
+tokenized_prompts = [
+    tokenizer(formatted_prompt, padding=True, pad_to_multiple_of=pad_to_multiple_of, return_tensors="pt")
+    for formatted_prompt in formatted_prompts
+]
+# put back padding
```
muellerzr, line 56:
```suggestion
# Put back the original padding behavior
```

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+tokenizer.padding_side = padding_side_default
+
+completions_per_process = []
+with distributed_state.split_between_processes(tokenized_prompts, apply_padding=True) as batched_prompts:
```
muellerzr, line 60: Comments as to what is going on here is very important

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+    for batch in batched_prompts:
+        # move the batch to the device
+        batch = batch.to(distributed_state.device)
+        outputs = model.generate(**batch, max_new_tokens=20)
+        generated_text = tokenizer.batch_decode(outputs, skip_special_tokens=True)
+        completions_per_process.extend(generated_text)
+
+completions_gather = gather_object(completions_per_process)
```
muellerzr, line 68: Same with here, why are we doing `gather_object` for instance.

Which also, ideally we should *not* be and we should use `gather`. Can you check if that's possible please?

examples/inference/distributed/phi2.py
```diff
@@ -0,0 +1,73 @@
...
+# Drop duplicates produced by apply_padding in  split_between_processes
+completions = completions_gather[: len(prompts)]
+
+if distributed_state.is_main_process:
```
muellerzr, line 72: FYI, `distributed_state` should have a `print()` 😉

examples/inference/distributed/stable_diffusion.py
```diff
@@ -0,0 +1,27 @@
...
+from accelerate import PartialState  # Can also be Accelerator or AcceleratorState
+
+
+pipe = DiffusionPipeline.from_pretrained("runwayml/stable-diffusion-v1-5", torch_dtype=torch.float16)
+distributed_state = PartialState()
+pipe.to(distributed_state.device)
+
+# Assume two processes
+with distributed_state.split_between_processes(["a dog", "a cat", "a chicken"], apply_padding=True) as prompt:
+    result = pipe(prompt).images
```
muellerzr, line 27: Comments plz :)

muellerzr (commented):
Thanks for doing this! A great start! Left some suggestions, let's make sure to document the examples as much as possible like we do the other `examples/`! :)

examples/inference/distributed/phi2.py
SunMarc: good to know !

examples/inference/distributed/phi2.py
SunMarc: done !

examples/inference/distributed/phi2.py
SunMarc: We need to stick with gather object since we are gathering string and not tensor. Left a comment about that !

examples/inference/distributed/stable_diffusion.py
SunMarc: done !

examples/inference/distributed/phi2.py
SunMarc: done !

examples/inference/distributed/phi2.py
muellerzr: What if we instead gathered first then decoded?

SunMarc: TLDR : gather won't work in our case here

I tried to use gather without much success and for users, it is easier to use gather_object.
The main downside of gather is that it requires the user to make sure that the tensor have the same size + same structure (gather applied on a list of tensors need to have the same number of tensor).
For example, in the phi2 script, we have (bs=2, apply_padding=True) :
# GPU 0: ["I would like to", "hello how are you"],  ["what is going on", "roses are red and"]
# GPU 1: ["welcome to the hotel"], ["welcome to the hotel"] -> this prompt is duplicated to ensure that all gpus have the same number of prompts
The GPU 0 is stuck since we are not able to gather the tensors that will go where "hello how are you" is (same for "roses are red and" ). We could of course fix this by making sure that in each batch, we have the same number of data, but that's not a very good behavior.
Plus, in a generation, the tensors do not necessarily have the same size so it is quite complicated to split the tensors afterwards + we might loose information during the gather.

SunMarc:
Merging as discussed with @muellerzr

PR #2673, opened 2024-04-15
Deprecate tqdm args + slight logic tweaks

muellerzr:
# What does this PR do?

This PR is a follow-up to https://github.com/huggingface/accelerate/pull/2654, specifically it:

1. Writes a new test to check that the deprecation is flagged
2. Changes the logic slightly to check if users are passing in `disable` explicitly as a kwarg. We should be respecting this if so

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc or @BenjaminBossan

### DIFF
--- a/src/accelerate/utils/tqdm.py
+++ b/src/accelerate/utils/tqdm.py
@@ -33,7 +33,7 @@ def tqdm(*args, main_process_only: bool = True, **kwargs):
     """
     if not is_tqdm_available():
         raise ImportError("Accelerate's `tqdm` module requires `tqdm` to be installed. Please run `pip install tqdm`.")
-    if isinstance(args[0], bool):
+    if len(args) > 0 and isinstance(args[0], bool):
         warnings.warn(
             f"Passing `{args[0]}` as the first argument to Accelerate's `tqdm` wrapper is deprecated "
             "and will be removed in v0.33.0. Please use the `main_process_only` keyword argument instead.",
--- a/src/accelerate/utils/tqdm.py
+++ b/src/accelerate/utils/tqdm.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
+
 from .imports import is_tqdm_available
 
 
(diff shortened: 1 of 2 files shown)


SunMarc (approved):
LGTM !

src/accelerate/utils/tqdm.py
```diff
@@ -31,7 +33,15 @@ def tqdm(*args, main_process_only: bool = True, **kwargs):
     """
     if not is_tqdm_available():
         raise ImportError("Accelerate's `tqdm` module requires `tqdm` to be installed. Please run `pip install tqdm`.")
-    disable = False
-    if main_process_only:
+    if len(args) > 0 and isinstance(args[0], bool):
```
rb-synth, line 36: nice 🚀

tests/test_utils.py
```diff
@@ -405,14 +405,8 @@ def test_convert_dict_to_env_variables(self):
...
-        assert list(tqdm(main_process_only=True, iterable=range(3))) == expected
-
-        assert list(tqdm(range(3), main_process_only=False)) == expected
-        assert list(tqdm(main_process_only=False, iterable=range(3))) == expected
-
-        assert list(tqdm(range(3))) == expected
+    def test_tqdm_deprecation(self):
+        with pytest.warns(FutureWarning) as cm:
+            tqdm(True, range(3), disable=True)
+        assert "Passing `True` as the first argument to" in cm.pop().message.args[0]
```
BenjaminBossan, line 411: `pytest.warns` should take a `match` argument, which can be used instead of doing this assert.

tests/test_utils.py
```diff
@@ -405,14 +405,8 @@ def test_convert_dict_to_env_variables(self):
...
+        tqdm(range(3), main_process_only=True, disable=True)
```
BenjaminBossan, line 412: Should this line also check that there is [no warning](https://docs.pytest.org/en/7.0.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests)?

```python
with warnings.catch_warnings():
    warnings.simplefilter("error")
```

BenjaminBossan (approved):
Thanks for taking care of this deprecation, LGTM.

Some nit comments. Also some suggestions:

- Would it make sense to make `main_process_only` a keyword-only argument?
- Do we want to have a `# TODO remove in accelerate v0.33.0` comment or something like this? Or how else do you grep for deprecations before a release?

muellerzr:
@BenjaminBossan re:

> Would it make sense to make main_process_only a keyword-only argument?

Probably, but I don't like shoving it in `**kwargs` where it can't be found on the hover etc. Happy medium here

> Do we want to have a # TODO remove in accelerate v0.33.0 comment or something like this? Or how else do you grep for deprecations before a release?

A bit easier to do it that way, yes. I tend to try and quickly grep for "deprecate" before a release

BenjaminBossan:
> Would it make sense to make main_process_only a keyword-only argument?

> `def tqdm(*args, main_process_only: bool = True, **kwargs)`

Oh I'm stupid, it's already kw only, can't pass it as a positional kw anymore :D

Issue #2624, opened 2024-04-04
FSDP requires larger LR to converge at similar rate as DeepSpeed


fabianlim:
### System Info

```Shell
accelerate: 0.28
deepspeed: 0.14
torch: 2.2.1

FSDP configs

fsdp_config:
  
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_backward_prefetch_policy: BACKWARD_PRE 
  fsdp_forward_prefetch: false
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1 
  fsdp_state_dict_type: FULL_STATE_DICT
  fsdp_cpu_ram_efficient_loading: true process
  fsdp_sync_module_states: true 
mixed_precision: 'no'

machine_rank: 0 # rank of the machine where accelerate is launched
num_machines: 1
num_processes: 1  # default, override with --num_processes
rdzv_backend: static
same_network: true
```

Deepspeed config
```json
{
    "train_micro_batch_size_per_gpu": "auto",
    "steps_per_print": 1,
    "zero_optimization": {
        "stage": 3,
        "offload_param": {"device": "none"},
        "offload_optimizer": {"device": "none"}
    },
    "bf16": {"enabled": true},
    "gradient_clipping": 1.0,
    "prescale_gradients": false,
    "wall_clock_breakdown": false
}
```

To reproduce use this script

```python
# tested on
# - transformers==4.39.0.dev0
# - accelerate==0.28.0
# - deepspeed==0.14
# - torch==2.2.1
# - trl==nightly
# - fire==0.6.0

# to run:
#
# accelerate launch --config_file ./accelerate_config.yaml learning_rate_repro.py

from datasets import load_dataset
from transformers import AutoTokenizer, AutoModelForCausalLM, HfArgumentParser
from transformers import TrainingArguments
from trl import SFTTrainer, DataCollatorForCompletionOnlyLM
import torch

# to run on 4 gpus on FSDP (for DS change the config file): 
# 
# accelerate launch \
#     --num_processes 4 \
#     --main_process_port 29502 \
#     --config_file accelerate.yaml \
#     learning_rate_repro.py  \
#       --num_train_epochs 10 \
#       --output_dir './results' \
#       --per_device_train_batch_size 10 \
#       --lr_scheduler_type "linear" \
#       --learning_rate 1e-6 \
#       --logging_steps 1 \
#       --save_strategy 'no' \
#       --bf16

def main(
[...]

PR #2674, opened 2024-04-16
Add Upcasting for FSDP in Mixed Precision. Add Concept Guide for FSPD and DeepSpeed. 

fabianlim:
# What does this PR do?

This PR address the issues identified in #2624, whereby differences were found between how DeepSpeed (DS) and FSDP handle the sharded parameters during mixed precision. To address this we have:
- added logic in `accelerate.prepare` to upcast the FSDP sharded paramaters, if it has been detected that:
   1. mixed precision has been activated, and,
   2. the sharded parameters are in low precision.
- added concept guide to clarify to how DS and FSDP handles mixed precision. Also to inform that 🤗  Accelerate will upcast FSDP sharded weights.
- added warning to inform user, that if the weights were upcasted, then this will affect the precision in which the checkpoint will be saved. 

In addition to the above, we also:
- int the above mentioned concept guide, clarified the equivalences between DS and FSDP configs, enabling users to better transition between DS/FSDP.
- updated the outdated [accelerate launch cli commands](https://huggingface.co/docs/accelerate/en/package_reference/cli#accelerate-launch).

Checklist:
- [x] test with low mem 
- [x] test with `SHARD_GRAD_OP`
- [x] test with CPU offload
- ~consider the [impacts of fp8 training](https://github.com/huggingface/accelerate/issues/2624#issuecomment-2050404125) suggested by @stas00.~ Update: this has to be done later 

To test and reproduce:
-  get the FSDP/ DS accelerate configurations, and the reproduction script (call it `learning_rate_repro.py`)  [from here](https://github.com/huggingface/accelerate/issues/2624#issue-2225498257).  
- run the script with `accelerate_fspd.yaml` config and without `--bf16` flag; this gets the `bf16` case in the below plot. The script will load the model in `bfloat16` (since `load_model_dtype` was default) and **turn off mixed precision**.
```
    accelerate launch \
        --num_processes 4 \
        --config_file accelerate_fsdp.yaml \
        learning_rate_repro.py  \
          --num_train_epochs 10 \
          --output_dir './results' \
          --per_device_train_batch_size 10 \
          --lr_scheduler_type "linear" \
          --learning_rate 1e-6 \
          --logging_steps 1 \
          --save_strategy 'no' 
```
- run the same adding `--bf16` (this gets `bf16-with-mp` in the plot below). This runs `bfloat16` model in FSDP **with mixed precision**.
- run the same benchmark with a DS `accelerate_ds.yaml` config (here `--bf16` is irrelevant, DS will handle both cases similarly). 

In the plot below, we can see that with the casting logic put in:
- if we detect that the model is in low precision `bfloat16` but mixed precision is turned on (`--bf16`), then we perform the upcasting. Hence we see that the loss curves of DS and FSDP will now be the same with the added casting logic.
- *If the logic was absent*, then we would have observed that FSDP would not converge, regardless if `--bf16` had been supplied or not

**Sample of New Warnings**

We have added some logic to reduce repetitive warnings: 

```
/data/flim/accelerate/src/accelerate/accelerator.py:1492: UserWarning: Upcasting low precision parameters in MistralForCausalLM, mixed precision is turned on: model.embed_tokens.weight,model.norm.weight,lm_head.weight.
  warnings.warn(
/data/flim/accelerate/src/accelerate/accelerator.py:1492: UserWarning: Upcasting low precision parameters in MistralDecoderLayer, mixed precision is turned on: self_attn.q_proj.weight,self_attn.k_proj.weight,self_attn.v_proj.weight,self_attn.o_proj.weight,mlp.gate_proj.weight,mlp.up_proj.weight,mlp.down_proj.weight,input_layernorm.weight,post_attention_layernorm.weight.
  warnings.warn(
```

**Performance of low precision models under FSDP and DS and mixed precision, with the proposed casting fix** 

This was plotted when comparing FSDP (Full Shard) and DeepSpeed (Zero3)
image

This was plotted when comparing FSDP in various modes, namely GradOp and CPUOffload (while full-sharding)
image

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

Tagging @stas00, @muellerzr and @pacman100  first. We can add more reviewers later.

### DIFF
--- a/docs/source/concept_guides/fsdp_and_deepspeed.md
+++ b/docs/source/concept_guides/fsdp_and_deepspeed.md
@@ -13,82 +13,126 @@ specific language governing permissions and limitations under the License.
 rendered properly in your Markdown viewer.
 -->
 
-# FSDP VS DeepSpeed [DRAFT]
+# Moving between FSDP And DeepSpeed [DRAFT]
 
-🤗 Accelerate integrates two extremely powerful tools for distributed training, namely [Pytorch FSDP](../usage_guides/fsdp.md) and [Microsoft DeepSpeed](../usage_guides/deepspeed.md). It is important to note similarities / differences in order to make an informed choice which framework that will better suit the desired use case.
+🤗 Accelerate offers flexibilty of training frameworks, by integrating two extremely powerful tools for distributed training, namely [Pytorch FSDP](../usage_guides/fsdp.md) and [Microsoft DeepSpeed](../usage_guides/deepspeed.md). The aim fo this article is to draw parallels, as well as to outline potential differences, to empower the user to switch seamlessly between these two frameworks.
 
 <Tip>
-  To switch betwen the frameworks, we recommend 🤗 `accelerate launch`.
+  To switch between the frameworks, we recommend 🤗 `accelerate launch`.
 
-  Simply pass [FSDP and DeepSpeed arguments](../package_reference/cli.md#accelerate-launch) directly to `accelerate launch` or pass `--config_file`. No need for any code instrumentation!
+  Simply pass [FSDP and DeepSpeed arguments](../package_reference/cli#accelerate-launch) directly to `accelerate launch` or pass `--config_file`. No need for any code instrumentation!
 
-  For examplar configurations see [DeepSpeed Config](../usage_guides/deepspeed#accelerate-deepspeed-plugin). and [FSDP Configuration](../usage_guides/fsdp.md#how-it-works-out-of-the-box). 
-  
-
+  Exemplar 🤗 Accelerate configurations can be found here for [DeepSpeed](../usage_guides/deepspeed#accelerate-deepspeed-plugin) and [FSDP](../usage_guides/fsdp#how-it-works-out-of-the-box). 
+ 
 </Tip>
 
 <!--
 The aim of this concept guide is to elucidate similarities/differences with empirical observations and code aspects.  We assume that 🤗 Accelerate is used and configured using the [FSDP and DeepSpeed arguments](../package_reference/cli.md#accelerate-launch). No TPU aspects are discussed. Also we focus only on single-node aspects.
 -->
 
-This article is written with 🤗 Accelerate in mind, for single-node GPU scenarios only. No TPU aspects will be considered.
+This article is written with 🤗 Accelerate in mind, for single-node, multi-GPU, scenarios only. No TPU aspects will be considered.
 
 ## Configuring Various Functionalities
 
-Training parameters are seperated into different GPUs to scale up for large models; this is termed *sharding* in FSDP, and *partioning* in DeepSpeed. Since the terminologies used in FSDP and DeepSpeed are disparate, there are various guides that [reconcile the mapping between FSDP sharding and DeepSpeed ZeRO](../usage_guides/fsdp.md#mapping-between-fsdp-sharding-strategies-and-deepspeed-zero-stages). This section aims to reconcile these differences for a 🤗 Accelerate user that desires to use both frameworks in an equivalent manner.
+Model tensors are split into different GPUs in attempt scale up model sizes; this is termed *sharding* in FSDP, and *partitioning* in DeepSpeed. The FSDP sharding strategy and the DeepSpeed ZeRO Stages, are configured by `--fsdp_sharding_strategy`, and `--zero_stage`, respectively.  For example FSDP `FULL_SHARD` maps to DeepSpeed ZeRO stage `3`, see this for a comprehensive [mapping between FSDP sharding and DeepSpeed ZeRO settings](../usage_guides/fsdp#mapping-between-fsdp-sharding-strategies-and-deepspeed-zero-stages). There exists a host of other settings besides these two, summarized by similarity in the below table.
 
-In general, DeepSpeed offers more fine-grained control, but may incur extra memory consumption as model and optimizer parameters are always upcasted to `float32`. 
-Below shows a mapping between FSDP and DeepSpeed configuration. 
-Note that to access advanced DeepSpeed configs via `accelerate launch` beyond
-[those exposed by accelerate](../package_reference/cli.md#accelerate-launch), one can link a full DeepSpeed config and pointing to it using the accelerate `deepspeed_config_file`. 
- 
-Configuration | FSDP | DeepSpeed
---|--|--
-sharding/partitioning | `--fsdp_sharding_strategy` | `--zero_stage`
-offload | `--fsdp_offload_params` | `--offload_optimizer_device, --offload_param_device`
-efficient weights loading | `--fsdp_cpu_ram_efficient_loading` | `--zero3_init_flag`
-checkpointing | `--fsdp_state_dict_type` | `--zero3_save_16bit_model`
-pipeline | `--fsdp_backward_prefetch, --fsdp_backward_prefetch` | 
-model | `--fsdp_auto_wrap_policy, --fsdp_transformer_layer_cls_to_wrap` | 
-parameters summoning | `--fsdp_use_orig_params` | 
-parameters syncing | `--fsdp_sync_module_states` | 
-training |  | `--gradient_accumulation_steps, --gradient_clipping`
+Group | Framework | Configuration | Example | Restrictions (if any)
+--|--|--|--|--
+sharding / partitioning | FSDP<br>DeepSpeed | `--fsdp_sharding_strategy`<br>`--zero_stage` | `1` (`FULL_SHARD`) <br>`3` | 
+offload | FSDP<br>DeepSpeed | `--fsdp_offload_params`<br>`--offload_param_device`<br>`--offload_optimizer_device` | `true`<br>`cpu`<br>`cpu` | all or nothing <br><br> 
+model loading | FSDP<br>DeepSpeed | <span style="white-space:nowrap;">`--fsdp_cpu_ram_efficient_loading`</span><br>`--zero3_init_flag` | `true`<br>`true` | <br>only ZeRO 3
+efficient checkpointing | FSDP<br>DeepSpeed | `--fsdp_state_dict_type`<br>`--zero3_save_16bit_model` |  `SHARDED_STATE_DICT`<br>`true` |  <br>only ZeRO 3
+pipeline | FSDP<br><br>DeepSpeed | `--fsdp_forward_prefetch`<br>`--fsdp_backward_prefetch`<br>None | `true`<br>`BACKWARD_PRE` | <br><br>?? check for DS
+model | FSDP<br><br>DeepSpeed |  `--fsdp_auto_wrap_policy`<br><span style="white-space:nowrap;">`--fsdp_transformer_layer_cls_to_wrap`</span><br>None | `TRANSFORMER_BASED_WRAP`<br><Layer Class> |<br>Usually not needed <br>Transparent to user.
+parameters summoning | FSDP<br>DeepSpeed | `--fsdp_use_orig_params`<br>None | `true` | required for `torch.compile`<br>Transparent to user
+parameters syncing | FSDP<br>DeepSpeed | `--fsdp_sync_module_states`<br>None | `true` | <br>?? Need to check
+training | FSDP<br>DeepSpeed | None<br>`--gradient_accumulation_steps`<br>`--gradient_clipping` | <br>`auto`<br>`auto` | Transparent to user
+
+We reiterate again that for all possible settings for the above, refer to [`Accelerate` launch documentation](../package_reference/cli#accelerate-launch).
 
 <Tip>
-    When training transformers, there are some FSDP configuration recommendations that should be followed.
 
-    - set `fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP`, and `fsdp_transformer_layer_cls_to_wrap` not needed for latest `transformers` versions.
-    - when setting `fsdp_cpu_ram_efficient_loading: True`, ensure `fsdp_sync_module_states: True` so that parameters are communicated from main process.
-    - is using `torch.compile` set `fsdp_use_orig_params: True`
+    To access other DeepSpeed configurations, such as mixed precision settings, 
+    one has to link a `deepspeed_config_file`, see [instructions here](../usage_guides/deepspeed#deepspeed-config-file).  
+    
+</Tip>
+
+<!--
+Since the terminologies used in FSDP and DeepSpeed are disparate, there are various guides that [reconcile the mapping between FSDP sharding and DeepSpeed ZeRO](../usage_guides/fsdp#mapping-between-fsdp-sharding-strategies-and-deepspeed-zero-stages). This section aims to reconcile these differences for a 🤗 Accelerate user that desires to use both frameworks in an equivalent manner.
+In general, DeepSpeed offers more fine-grained control, but may incur extra memory consumption as model and optimizer parameters are always upcasted to `float32`. 
+-->
+
+<Tip>
+    When using FSDP these are recommendations that should be followed.
+
+    - set `fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP`. Do not set `fsdp_transformer_layer_cls_to_wrap` if using the latest `transformers` versions.
+    - if `fsdp_cpu_ram_efficient_loading: True`, set `fsdp_sync_module_states: True` otherwise the model will not load properly.
+    - when using `torch.compile` set `fsdp_use_orig_params: True`.
 
 </Tip>
 
 <Tip>
-    These are also the DeepSpeed configuration recommendations that should be followed.
+    When using DeepSpeed these are recommendations that should be followed.
 
     - set `gradient_accumulation_steps: "auto"` and `gradient_clipping: "auto"` to automatically pick up values set in [`TrainingArguments`].
 
 </Tip>
 
-## Differences in Training Precisions
+TODO: Consider elaborating on some points ? Maybe a small subsection for each?
+- on checkpoint sharding (DS does not have checkpoint sharding?) Des FSDP save the model in 16 bit also?
+- zero3 init in DS will be automatic if model is above a certain size. More details on this?
+- do I need to talk about bucketing in DS? this is how DS takes care of partitioning in an automatic manner without the wrap policy. Do I need to explain FSDP wrapping causes graph breaks in torch.compile and how this is being resolved?
+- how does DS take care of pipelining? I need to check further.
+- do we need to discuss parameter summoning for DS? maybe not because this is an advanced usage
+- should we discusss activation checkpointing in the configs or is this obvious?
 
+
+## On Data Precision
+
+To discuss the how data precision is handled in both FSDP and Deepspeed, it is instructive to first give a flow of how model parameters are handled in these frameworks. Before the model / optimizer parameters are distributed across GPUs, parameter preparation is involved to first "flatten" them to  one-dimensional [`torch.Tensor`]'s. The implementation of FSDP / DeepSpeed varies in the respect of the `dtype` in which these "flattened" parameters are stored, and there are ramnifications with regards to how [`torch.Optimizer`]'s allocate their `dtypes`'s. The table below outlines the processes for both frameworks; the "Local" column indicates the process occurring at a per-gpu level, therefore any memory overheads by upcasting should be understood to be amortized by the number of gpus used.
+
+<!--
+TODO: for FSDP there are some mixed precision settings like `keep_low_precision_grads`, should we discuss them? NVM, because they way huggingface prepares the model, keep_low_precision_grads will never need to be used
+-->
+
+Process | Local | Framework | Details
+--|--|--|--
+Loading, i.e., [`AutoModel.from_pretrained(..., torch_dtype)`] |  
+Preparation, i.e., creation of "flat params" | ✅ | FSDP<br>DeepSpeed | created in `torch_dtype`.<br> disregards `torch_dtype`, created in `float32`.
+Optimizer initialization | ✅ | FSDP<br>DeepSpeed  | creates parameters in `torch_dtype`<br> creates parameters in `float32`
+Training Step, i.e, forward, backward, reduction | | FSDP<br>DeepSpeed  | follows [`MixedPrecision`](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.MixedPrecision)<br> follows `deepspeed_config_file` mixed precision settings.
+Optimizer (Pre-Step) | ✅ | FSDP<br>DeepSpeed | upcasting (if any) to `torch_dtype`<br>upcasted to `float32`
+Optimizer (Actual Step) | ✅ | FSDP<br>DeepSpeed  | occurs in `torch_dtype` <br> occurs in `float32`.
+
+
+<!--
 Both FSDP and DeepSpeed have logic for sharding/partitioning the model / optimizer parameters across GPUs; however there are some differences to be aware of. Both of these
 frameworks will create new [`torch.Tensor`]'s to hold "flattened" parameters, which will be used to instantiate the optimizers in each shard/partition. However since the [`torch.Optimizer`]'s allocate their `dtypes`'s based on the optimized parameters, it is important to note the following:
 
+For FSDP there are two options i) load and train/optimize the model in low precision, or ii) load in full precision and configure [`MixedPrecision`](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.MixedPrecision) to have activations / reduction performed in low precision. But for DeepSpeed always load the model in full precision. To summarize:
+-->
+
 <Tip warning={true}>
 
-    FSDP will retain the model's `dtype`; if a model is loaded in `float16`, then that will be the `dtype` of the flattened model parameters, and also that of the optimizer, etc. This will be discussed more clearly below. 
-    
-    DeepSpeed however will always upcast the parameters to `float32`, so even if the model was loaded in `float16`, the optimizer will see `float32` parameters. In other words, DeepSpeed only operates in mixed precision. There is no reason to load in low precision and yet train with DeepSpeed.
+    Therefore when using DeepSpeed, one should always load the model with `torch_dtype=torch.float32`.
 
-    This difference may result in different observations during training.
+    However if the number of GPUs are small, be aware of potentially significant memory overheads due to the "local" upcasting.
 
 </Tip>
 
-For FSDP there are two options i) load and train/optimize the model in low precision, or ii) load in full precision and configure [`MixedPrecision`](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.MixedPrecision) to have activations / reduction performed in low precision. But for DeepSpeed always load the model in full precision. To summarize:
+<Tip warning={true}>
 
-Framework | Load Model | Mixed Precision | Flat Params | Forward / Backward | Optimizer
+    With FSDP it is possible to operate the [`torch.Optimizer`] in low precision `torch_dtype`, which may be helpful when using small number of GPUs.
+
+    And if migrating to DeepSpeed from FSDP, be aware that if `torch_dtype` had been previously set to low precision (unnecessarily, see above), it will result in non-equivalent observations since then FSDP will optimize in low precision.
+
+</Tip>
+
+
+To clarify the above table consider the concrete examples below; the optimizer pre- and actual step combined for brevity. Thus to ensure that the same data precisions are used in both FSDP and DeepSpeed, always specify `torch_dtype=torch.float32` while calling [`AutoModel.from_pretrained`].
+
+Framework | Model Loading (`torch_dtype`) | Mixed Precision | Preparation (Local) | Training | Optimizer (Local)
 --|--|--|--|--|--
-FSDP | bf16 | None | bf16 | bf16 | bf16
-FSDP | bf16| fp32 | fp32 | bf16 | fp32
-DeepSpeed   | bf16 | bf16 | fp32 | bf16 | fp32
+FSDP | bf16 | default (none) | bf16 | bf16 | bf16
+FSDP | fp32 | bf16 | fp32 | bf16 | fp32
+DeepSpeed   | fp32 | bf16 | fp32 | bf16 | fp32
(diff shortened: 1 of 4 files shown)


pacman100 (commented):
Thank you @fabianlim for the concept guides 📄 for switching between FSDP and DS and adding logic to upcast the loaded model to FP32 when using AMP given that the loaded model is having lower precision for proper convergence📉! Left few comments and suggestions.

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,157 @@
...
+<Tip>
+
+    When using DeepSpeed, set `gradient_accumulation_steps: "auto"` and `gradient_clipping: "auto"` to automatically pick up values set in [`TrainingArguments`].
+
+</Tip>
+
+
+## On Differences in Data Precision Handling
+
+To discuss the how data precision is handled in both FSDP and Deepspeed, it is instructive to first give a flow of how model parameters are handled in these frameworks. Before the model / optimizer parameters are distributed across GPUs, parameter preparation is involved to first "flatten" them to  one-dimensional [`torch.Tensor`](https://pytorch.org/docs/stable/tensors.html#torch-tensor)'s. The implementation of FSDP / DeepSpeed varies in the respect of the `dtype` in which these "flattened" parameters are stored, and there are ramifications with regards to how [`torch.Optimizer`](https://pytorch.org/docs/stable/optim.html#module-torch.optim)'s allocate their `dtypes`'s. The table below outlines the processes for both frameworks; the "Local" column indicates the process occurring at a per-gpu level, therefore any memory overheads by upcasting should be understood to be amortized by the number of gpus used.
```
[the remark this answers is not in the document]
fabianlim: Done I am adding a tip

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,157 @@
...
+
+</Tip>
+
+**Model**
+
+FSDP requires an explicit `--fsdp_auto_wrap_policy` for the algorithm to decide how to schedule the all-gather and reduce-scatter operations. But for DeepSpeed this is transparent to the user.
+
+<Tip>
+
+    For FSDP, simply set `fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP`. Do not set `fsdp_transformer_layer_cls_to_wrap` if using the latest `transformers` versions.
```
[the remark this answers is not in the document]
fabianlim: added

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,157 @@
...
+
+Group | Framework | Configuration | Example | Restrictions (if any)
+--|--|--|--|--
+sharding / partitioning | FSDP<br>DeepSpeed | `--fsdp_sharding_strategy`<br>`--zero_stage` | `1` (`FULL_SHARD`) <br>`3` | 
+offload | FSDP<br>DeepSpeed | `--fsdp_offload_params`<br>`--offload_param_device`<br>`--offload_optimizer_device` | `true`<br>`cpu`<br>`cpu` | all or nothing <br><br> 
+model loading | FSDP<br>DeepSpeed | <span style="white-space:nowrap;">`--fsdp_cpu_ram_efficient_loading`</span><br>`--zero3_init_flag` | `true`<br>`true` | <br>only ZeRO 3
+efficient checkpointing | FSDP<br>DeepSpeed | `--fsdp_state_dict_type`<br>`--zero3_save_16bit_model` |  `SHARDED_STATE_DICT`<br>`true` |  <br>only ZeRO 3
```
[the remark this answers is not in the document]
fabianlim: I have done the following:
- updated the **Checkpointing** section to specify the equivalence of `zero3_save_16bit_model=True` and `FULL_STATE_DICT`.
- Added a warning in the data precision section, to mention that if FSDP has upcasted a low precision model, then the checkpoints will be saved in the upcasted precision.

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Moving between FSDP And DeepSpeed (DRAFT)
+
+🤗 Accelerate offers flexibilty of training frameworks, by integrating two extremely powerful tools for distributed training, namely [Pytorch FSDP](../usage_guides/fsdp.md) and [Microsoft DeepSpeed](../usage_guides/deepspeed.md). The aim fo this article is to draw parallels, as well as to outline potential differences, to empower the user to switch seamlessly between these two frameworks.
```
muellerzr, line 18:
```suggestion
🤗 Accelerate offers flexibilty of training frameworks, by integrating two extremely powerful tools for distributed training, namely [Pytorch FSDP](../usage_guides/fsdp.md) and [Microsoft DeepSpeed](../usage_guides/deepspeed.md). The aim of this tutorial is to draw parallels, as well as to outline potential differences, to empower the user to switch seamlessly between these two frameworks.
```

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+
+<Tip>
+  To switch between the frameworks, we recommend 🤗 `accelerate launch`.
+
+  Simply pass [FSDP and DeepSpeed arguments](../package_reference/cli#accelerate-launch) directly to `accelerate launch` or pass `--config_file`. No need for any code instrumentation!
+
+  Exemplar 🤗 Accelerate configurations can be found here for [DeepSpeed](../usage_guides/deepspeed#accelerate-deepspeed-plugin) and [FSDP](../usage_guides/fsdp#how-it-works-out-of-the-box). 
+ 
+</Tip>
```
muellerzr, line 27:
```suggestion
<Tip>

  To switch between the frameworks, we recommend launching code 🤗 `accelerate launch` passing in the correct config file with `--config_file`, or passing in the respective arguments directly for [FSDP and DeepSpeed](../package_reference/cli#accelerate-launch) .

  Example 🤗 Accelerate configurations can be found here for [DeepSpeed](../usage_guides/deepspeed#accelerate-deepspeed-plugin) and [FSDP](../usage_guides/fsdp#how-it-works-out-of-the-box), or in the [example zoo under "Launch Configurations"](../usage_guides/explore)
 
</Tip>
```

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+
+This article is written with 🤗 Accelerate in mind, for single-node, multi-GPU, scenarios only. No TPU aspects will be considered.
```
muellerzr, line 29: As it's a concept guide in our documentation, no need to state it's for accelerate

```suggestion
<Tip warning={true}>

This tutorial is for single-node, multi-GPU, scenarios only.

</Tip>
```

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+parameters summoning | FSDP<br>DeepSpeed | `--fsdp_use_orig_params`<br>None | `true` | required for `torch.compile`<br>Transparent to user
+parameters syncing | FSDP<br>DeepSpeed | `--fsdp_sync_module_states`<br>None | `true` | 
+training | FSDP<br>DeepSpeed | None<br>`--gradient_accumulation_steps`<br>`--gradient_clipping` | <br>`auto`<br>`auto` | Transparent to user
+
+For detailed descriptions of the above, refer to [🤗 `Accelerate` launch documentation](../package_reference/cli#accelerate-launch).
+
+<Tip>
+
+    To access other DeepSpeed configurations, such as mixed precision settings, 
+    you need to link a `deepspeed_config_file`, see the [documentation](../usage_guides/deepspeed#deepspeed-config-file).
```
muellerzr, line 52:
```suggestion
    you need to pass in a `--deepspeed_config_file`, see the [documentation](../usage_guides/deepspeed#deepspeed-config-file).  
```

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+For detailed descriptions of the above, refer to [🤗 `Accelerate` launch documentation](../package_reference/cli#accelerate-launch).
+
+<Tip>
+
+    To access other DeepSpeed configurations, such as mixed precision settings, 
+    you need to link a `deepspeed_config_file`, see the [documentation](../usage_guides/deepspeed#deepspeed-config-file).  
+    
+</Tip>
+
+**Checkpointing**
```
muellerzr, line 56: Let's do `###` blocks instead, so they'll show up in the sidebar
```suggestion
### Checkpointing
```

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,167 @@
...
+
+Do note that while FSDP can be configured via `--fsdp_state_dict_type` to save either full / sharded checkpoints.
+
+<Tip>
+
+    For DeepSpeed Zero3, it is recommended to also pass a `--zero3_save_16bit_model true` in the [`Accelerate` launch arguments](../package_reference/cli#accelerate-launch), which conviniently consolidates the model to a single rank and saves; this is the FSDP equivalent of `fsdp_state_dict_type: FULL_STATE_DICT`.
```
muellerzr, line 62:
```suggestion
    For DeepSpeed Zero3, it is recommended to also pass a `--zero3_save_16bit_model true`, which conveniently consolidates the model to a single rank and saves; this is the FSDP equivalent of `fsdp_state_dict_type: FULL_STATE_DICT`.
```

src/accelerate/accelerator.py
```diff
@@ -1447,6 +1447,60 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+                            continue  # skip if FSDP module not managing parameters
+                        param = module._flat_param
+                        if (
+                            param.dtype != torch.float32
+                            and param.device != torch.device("meta")
+                            and param.requires_grad
+                        ):
+                            # TODO: make the warning issue only once
+                            if self.is_main_process:
+                                warnings.warn(
```
muellerzr, line 1492: You can do `warnings.warn_once`

muellerzr (commented):
This is very good so far! A recommendation:

Let's please also show doing the same with `FullyShardedDataParallelPlugin` and `DeepSpeedPlugin` directly, rather than just with the config file, as many users don't know this can be done

docs/source/concept_guides/fsdp_and_deepspeed.md
fabianlim: thanks! this is done in https://github.com/huggingface/accelerate/pull/2674/commits/af71e9f7184fed1bd0696c68bc305e1015fa04b9

src/accelerate/accelerator.py
fabianlim: @muellerzr thanks for the suggestion. I tried doing this with `warnings.catch_warnings` and `warnings.simplefilter`, but the `once` filter seems to be broken for me. Im on py 3.10

So Im proposing a "manual" way to do it. It doesnt look all that bad, if you are ok with it.
https://github.com/huggingface/accelerate/pull/2674/commits/0b8e97c98ba22ebe6ed8eb157efaa4cb7d859bed

fabianlim:
> Let's please also show doing the same with FullyShardedDataParallelPlugin and DeepSpeedPlugin directly, rather than just with the config file, as many users don't know this can be done

@muellerzr thanks for your comments, I have addressed most of except for the one above which requires a bit more effort. Will work on in. In the meantime if you have any suggestions to my responses pls let me know.

shagunsodhani:
> added logic in accelerate.prepare to upcast the FSDP sharded paramaters, if it has been detected that:
> * mixed precision has been activated, and,
> * the sharded parameters are in low precision.

Is there a way to disable that behavior ? i.e. no upcast the parameters even when both the conditions are met ? And to confirm, the parameters would not be upcasted when using just bf16 training (i,e, no mixed precision training)

fabianlim:
@shagunsodhani before this PR, if model is in low precision, and regardless there is mixed precision or not, we will not upcast

This PR is poposing that when model is in low precision
- mixed_precision == False -> no upcast
- mixed_precision == True -> upcast (to gain parity with DS)

shagunsodhani:
> @shagunsodhani before this PR, if model is in low precision, and regardless there is mixed precision or not, we will not upcast
> 
> This PR is poposing that when model is in low precision
> 
> * mixed_precision == False -> no upcast
> * mixed_precision == True -> upcast (to gain parity with DS)

Thank you for the context. Let me try rephrasing the question:

If I want to train a model using bf16 and FSDP, this is what I do:

1. Create an Accelerator object with mixed_precision=bf16 (with relevant FSDP flags)
2. Creat model in fp32
3. Pass the model to acclerate.prepare
4. Train...

Following #2624 , it seems deepspeed upcasts the parameters in the optimizer and this PR is applying that behavior to FSDP as well.

So if a user wants to opt-out of this behaviour, they could do the following:

1. Create an Accelerator object with mixed_precision=no (with relevant FSDP flags)
2. Creat model in bf16
3. Pass the model to acclerate.prepare
4. Train...

Is that correct ?

fabianlim:
> If I want to train a model using bf16 and FSDP, this is what I do:

> Create an Accelerator object with mixed_precision=bf16 (with relevant FSDP flags)
> Creat model in fp32
> Pass the model to acclerate.prepare
> Train...

@shagunsodhani  it depends what you mean by *opt out*.  If you do the above before this PR, you would actually have
- flat params in 32
- optimizer params in 32
- fwd_bwd, reduce, etc in bf16

After this PR, you would have exactly the same behavior as above, in the following two scenarios
- create model in fp32, mixed precision = bf16
- create model in bf16, mixed precision = bf16

On the ohter hand, before and after this PR if you had done this

> So if a user wants to opt-out of this behaviour, they could do the following:

> Create an Accelerator object with mixed_precision=no (with relevant FSDP flags)
> Creat model in bf16
> Pass the model to acclerate.prepare
> Train...

then you would have the below, which is not equivalent to the first scenario presented by you.
- flat params in bf16
- optimizer in bf16
- fwd_bkwd, reduce in bf16

shagunsodhani:
Great - thank you for explaining this :)

fabianlim:
@muellerzr I addressed your last comment on the PLugins
- put two small notes, once for FSDP and DS, pointing to an equivalence between `FullyShardedDataParallelPlugin and `DeepSpeedPlugin` with the config flags. https://github.com/huggingface/accelerate/commit/681c697d843c3fdf3d8bf45d39de4884c03a46a6
- tested cpu_offload and grad_op, updated the results in [the main description](https://github.com/huggingface/accelerate/pull/2674#issue-2245033037). See the two new curves just above *"Before Submitting*"
- put one more warning that the automatic upcast may increase the precision of the saved checkpoints.

I think its almost there, we have covered most bases.

muellerzr:
Accelerate will still always wrap those outputs w/ the `ConvertOutputsToFp32`, the chart itself was taken from MSAMP

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,178 @@
...
+
+<Tip>
+
+    With FSDP, in the absence of mixed precision, it is possible to operate the [`torch.Optimizer`](https://pytorch.org/docs/stable/optim.html#module-torch.optim) in low precision `torch_dtype`, which may be helpful when using small number of GPUs. 
+
+</Tip>
+
+<Tip warning={true}>
+
+    With mixed precision, then FSDP and DeepSpeed will upcast in the model preperation step (c.f. table above). But do note that FSDP will then save checkpoints in the upcasted precision; Deepspeed may still save low precision checkpoints if `--zero3_save_16bit_model` is specified.
```
muellerzr, line 167:
```suggestion
    With mixed precision, FSDP and DeepSpeed will upcast in the model preparation step (c.f. table above). But do note that FSDP will then save checkpoints in the upcasted precision; Deepspeed may still save low precision checkpoints if `--zero3_save_16bit_model` is specified.
```

muellerzr (approved):
This looks **phenomenal**! Great work 🤗 

I left a few wording nits, but looks *great* to me! cc @pacman100 as well for a last review

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,178 @@
...
+
+<Tip>
+
+    For DeepSpeed Zero3, it is recommended to also pass a `--zero3_save_16bit_model true`, which conveniently consolidates the model to a single rank and saves; this is the FSDP equivalent of `fsdp_state_dict_type: FULL_STATE_DICT`.
+
+</Tip>
+
+### Offloading
+
+FSDP only allows *all-or-nothing* offload, but DeepSpeed can offload parameters and optimizer differently. Furthermore, DeepSpeed also supports [offloading to NVME](https://www.deepspeed.ai/docs/config-json/#parameter-offloading).
```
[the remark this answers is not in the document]
fabianlim: done

pacman100 (approved):
Thank you, @fabianlim, for fixing the mixed precision training with FSDP when the model is *not* loaded in full-precision and for improving the documentation by outlining the parallels between FSDP and DeepSpeed. Super helpful!

Left a few nits

muellerzr:
@fabianlim the answer is we just currently haven't enabled FP8 mixed precision for FSDP, it's something we're looking into as `TransformersEngine` only recently finished getting support for that going

muellerzr:
So for right now, happy to merge this in when you're ready just let me know 🤗

fabianlim:
@muellerzr got it.. ok lets merge this first then address the fp8 later!

muellerzr:
Thanks a *bunch* for all of this fantastic work @fabianlim, once CI is green we'll merge and this will get in before the next release 🚀 (End of this week)

stas00:
Awesome work, @fabianlim!

Some belated feedback:

- I think this could be a misleading entry:

snapshot_731

while both help to overcome the big model loading issue - they aren't the same functionality - since `zero.Init` requires special code to init tensors (need to manually gather those or else they will be silently skipped).

- not sure what `pipeline` means here, should it say - `weights prefetch` instead?

snapshot_732

- should the newly added https://huggingface.co/docs/accelerate/main/en/concept_guides/fsdp_and_deepspeed be cross-linked from the Deepspeed and FSDP accelerate docs? just a suggestion so that it's easier to discover.

- I'm not at all sure this should be "recommended" - this is slow and expensive and typically undesirable during large model training

snapshot_734

perhaps it'd be better to say - "one could use ..." 

It's trivial to get this unsharded checkpoint during the training with zero_to_fp32.py script, w/o slowing down the training. At least this is the approach that we use.

- this is not entirely true

snapshot_735

For example, `stage3_param_persistence_threshold` will affect the prefetch - if the tensors aren't sharded there will be no prefetch. Then you  have `stage3_max_reuse_distance` and `stage3_max_live_parameters` which too will have an impact.

perhaps the doc should say "when needed"? so that not to erroneously imply that all tensors are always pre-fetched?

-

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,178 @@
...
+
+    When using DeepSpeed, set `gradient_accumulation_steps: "auto"` and `gradient_clipping: "auto"` to automatically pick up values set in the [`Accelerator`] or [`TrainingArguments`] (if using `transformers`).
+
+</Tip>
+
+
+## On Differences in Data Precision Handling
+
+To discuss the how data precision is handled in both FSDP and Deepspeed, it is instructive to first give a flow of how model parameters are handled in these frameworks. Before the model / optimizer parameters are distributed across GPUs, parameter preparation is involved to first "flatten" them to  one-dimensional [`torch.Tensor`](https://pytorch.org/docs/stable/tensors.html#torch-tensor)'s. The implementation of FSDP / DeepSpeed varies in the respect of the `dtype` in which these "flattened" parameters are stored, and there are ramifications with regards to how [`torch.Optimizer`](https://pytorch.org/docs/stable/optim.html#module-torch.optim)'s allocate their `dtypes`'s. The table below outlines the processes for both frameworks; the "Local" column indicates the process occurring at a per-gpu level, therefore any memory overheads by upcasting should be understood to be amortized by the number of gpus used.
```
stas00, line 136: - I don't know what "give a flow" means - did you mean "give an overview"?
- there should be no `'s` but `s` in ` #torch-tensor)'s. ` above

docs/source/concept_guides/fsdp_and_deepspeed.md
```diff
@@ -0,0 +1,178 @@
...
+
+<Tip>
+
+    As a rule of thumb, for stable training with automatic mixed precision, all the trainable parameters have to be in `torch.float32`.
+
+</Tip>
+
+Process | Local | Framework | Details
+--|--|--|--
+Loading, i.e., [`AutoModel.from_pretrained(..., torch_dtype)`] |
```
stas00, line 146: probably would be more clear as:

```
`AutoModel.from_pretrained(..., torch_dtype=torch_dtype)
```

fabianlim:
@stas00 thank you for your great suggestions! Let me see how I can quickly address them. Is it ok if I open another PR @muellerzr, to address @stas00 comments?

muellerzr:
Yep feel free!

fabianlim:
@stas00 thank you again fro your above comments I have attempted to address them in #2725. In particular:
- resolved your comments on apostrophes and positional argument
- renamed `pipeline` -> `weight prefetching`
- be more careful on the recommendation of `--zero3_save_16bit_model: True`.
- be more careful with the statement regarding the prefetch logic of DeepSpeed.

For Discussion (copying @muellerzr, @pacman100 ):
- regarding the equivalence of `fsdp_cpu_ram_efficient_loading` and `zero3_init_flag`. Is my understanding correct, it sounds like what you saying that  `zero.Init` will do some weight gathering for other ranks. Am I understanding correctly?
-  If so, this will be similar to `fsdp_cpu_ram_efficient_loading` which also loads the weights in a single rank, and requires the other ranks to gather them.
- to this end, accelerate launch will raise if  `fsdp_cpu_ram_efficient_loading` is set but `fsdp_sync_module_states` is left unset.
- However, if a user choses not to pass options directly in `acclerate launch`, but reply on the YAML config, then a user can set `cpu_ram_efficient_loading: True` and leave `sync_module_states`. This will cause problems.
- Thus in https://github.com/huggingface/accelerate/pull/2725/commits/19cfab4bb2b4371f2b91d7fedf7ce287a5a1bd5a, I propose to remove the raise in `accelerate.launch`, and just set it them automatically in `FullyShardedDataParallelPlugin.__post_init__`. 
- Or, we could keep the raise, and have it work in tandem with the proposed  automatic setting in `FullyShardedDataParallelPlugin`. 

Not Done:
- [] put links in FSDP and DeepSpeed docs.

varadhbhatnagar:
Hi @fabianlim 

Can you summarise, what happens to 

```
flat params 
optimizer 
fwd_bkwd, reduce
``` 
when model is created in bf16 and mixed_precision='bf16', before and after this PR? (I found this format of explaination to be very nice https://github.com/huggingface/accelerate/pull/2674#issuecomment-2067663170 )

src/accelerate/accelerator.py
```diff
@@ -1447,6 +1447,73 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+                            # NOTE: resorted to this because warnings.simplefilter("once") is somehow not working
+                            name_param_log = (module.module.__class__.__name__, ", ".join(module._flat_param._fqns))
+                            if name_param_log not in upcasted_log:
+                                upcasted_log.append(name_param_log)
+
+                            # this works because of FSDP's _runtime_utils.lazy_init.
+                            # Have to be careful not to call anything before this that
+                            # triggers lazy_init (e.g., _is_fsdp_root).
+                            param.data = param.data.to(torch.float32)  # upcasting
+                            module._handle._orig_param_dtype = torch.float32  # update
```
stas00, line 1501: @fabianlim, it looks like this introduced a breakage in FSDP integration - one of our tests now fails because of the last 2 lines, but works fine prior to this PR being merged.

```
    output = self.model(**batch)
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/fully_sharded_data_parallel.py", line 843, in forward
    args, kwargs = _pre_forward(
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 380, in _pre_forward
    unshard_fn(state, handle)
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 415, in _pre_forward_unshard
    _unshard(state, handle, state._unshard_stream, state._pre_unshard_stream)
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_runtime_utils.py", line 288, in _unshard
    ran_pre_unshard = handle.pre_unshard()
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_flat_param.py", line 1252, in pre_unshard
    ret = self._writeback_orig_params()
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_flat_param.py", line 2249, in _writeback_orig_params
    self._writeback_tensor(
  File "/env/lib/conda/mgor-core-dev/lib/python3.10/site-packages/torch/distributed/fsdp/_flat_param.py", line 2348, in _writeback_tensor
    raise RuntimeError(
RuntimeError: Cannot writeback when the parameter shape changes
Expects torch.Size([48000]) but got torch.Size([3000, 16])
```

the total size (numel) is correct, but the shape is wrong

stas00: also cc: @muellerzr

stas00: the training step was just a trivial tiny llama2 model, like: https://huggingface.co/stas/tiny-random-llama-2

and the config was:
```
distributed_type: FSDP
fsdp_config:
  fsdp_forward_prefetch: false  # can flip to true once requiring torch>=2.2.0
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: SHARDED_STATE_DICT
  fsdp_cpu_ram_efficient_loading: true
  fsdp_sync_module_states: true
  fsdp_use_orig_params: true
```

fabianlim: @stas00 can you share the accelerate config ? This looks like something commonly seen if NOSHaRD is used.

stas00: full config:

```

distributed_type: FSDP
fsdp_config:
  fsdp_forward_prefetch: false  # can flip to true once requiring torch>=2.2.0
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: SHARDED_STATE_DICT
  fsdp_cpu_ram_efficient_loading: true
  fsdp_sync_module_states: true
  fsdp_use_orig_params: true

mixed_precision: bf16
machine_rank: 0
main_process_ip: null
main_process_port: null
main_training_function: main
num_machines: 1
num_processes: {num_processes}
use_cpu: false
```

stas00: if I change the test to load the model in fp32 explicitly and not bf16, this code is bypassed and then everything works again.

fabianlim: Ok I think it’s because of use-orig-params being set to true. This was a setting that I didn’t test

stas00: Is there a reason to do the upcasting so late in the game? Won't it be much simpler to do it before the model gets wrapped by FSDP?

Accelerate knows by this time everything it needs to know to make the right thing, no?

fabianlim: The reason is because we only wanted to cast the sharded params and not the whole model. Casting the whole model is not a good idea IMO

stas00: Didn't we say that this was a convenience workaround and that the model should have been loaded in fp32 in the first place?

Frankly, when do you not shard most of the model parameters? and even if selective typically it's the big params that are sharded so 2x of that would be close to all of model's params doubled in size.

fabianlim: @stas00 
Essentially what I am saying is that upcasting ahead is will in general incur a 2X overhead in CPU memory in most general usage patterns that overhead will be there regardless if we use low_cpu_mem mode or not
image

BTW im having a bit of trouble reproducing your error on torch 2.1.2 and 2.2.0. 
- I took the `learning_rate_repro.py` script [in the main issue](https://github.com/huggingface/accelerate/issues/2624#issue-2225498257)
- I ran the following command, where `dataset.json` is [prepared following this](https://github.com/foundation-model-stack/fms-hf-tuning?tab=readme-ov-file#pre-process-the-jsonjsonl-dataset)
- `accelerate_stas.yaml ` is the config that was shared above.
- using the same `stas/tiny-random-llama-2`
```
accelerate launch \
    --num_processes 4 \
    --main_process_port $MASTER_PORT \
    --config_file accelerate_stas.yaml \
    learning_rate_repro.py  \
      --model_name_or_path stas/tiny-random-llama-2 \
      --max_seq_len 4096 \
      --num_train_epochs 1 \
      --output_dir './results' \
      --per_device_train_batch_size 8 \
      --gradient_accumulation_steps 1 \
      --include_tokens_per_second \
      --gradient_checkpointing \
      --lr_scheduler_type "linear" \
      --learning_rate 1e-6 \
      --logging_steps 1 \
	  --packing True \
       --dataset_for_packing dataset.json \
      --dataset_text_field output \
      --bf16 \
      --max_steps 100 \
      --save_strategy 'no'
```
and the training runs. 
```
{'loss': 8.0075, 'grad_norm': 0.9171388149261475, 'learning_rate': 9.9e-06, 'epoch': 1.0}
{'loss': 7.9684, 'grad_norm': 1.1230570077896118, 'learning_rate': 9.800000000000001e-06, 'epoch': 2.0}
{'loss': 7.8988, 'grad_norm': 1.232378602027893, 'learning_rate': 9.7e-06, 'epoch': 3.0}
{'loss': 7.8119, 'grad_norm': 1.1301443576812744, 'learning_rate': 9.600000000000001e-06, 'epoch': 4.0}
{'loss': 7.7319, 'grad_norm': 1.0580438375473022, 'learning_rate': 9.5e-06, 'epoch': 5.0}
```

PR #2677, opened 2024-04-16
Fix deepspeed moe test with version check

pacman100:
# What does this PR do?
1. Fix deepspeed moe test with version check


muellerzr (approved):
Good solution :)

PR #2681, opened 2024-04-17
Update modeling.py by adding try-catch section to skip the unavailable devices

MeVeryHandsome:
# What does this PR do?
In the current implementation of the library, when deploying tensor operations across multiple GPU devices, the program attempts to initialize a tensor on each device and get maximum available memory. 
```
def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] = None):
        ...
                for i in range(torch.cuda.device_count()):
                    _ = torch.tensor([0], device=i)
                max_memory = {i: torch.cuda.mem_get_info(i)[0] for i in range(torch.cuda.device_count())}
```
However, this approach leads to an error and terminates the program if it encounters a GPU that is unavailable(memory fully occupied or device damaged). 

For example, this error will occur when one of the devices is fully occupied:
```
RuntimeError: CUDA error: out of memory
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
```
To address this issue, I propose a modification to the logic for determining the maximum available memory on GPUs. Specifically, I have introduced a try-except block around the tensor deployment operation. 
```
            for i in range(torch.cuda.device_count()):
                try:
                    _ = torch.tensor([0], device=i)
                    max_memory.append({i: torch.cuda.mem_get_info(i)[0]})
                except Exception:
                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                    continue
```
If an error occurs due to the unavailability of a GPU, the code will now catch the exception and continue checking the next available GPU. This change ensures that the program gracefully skips over unavailable GPUs and only utilizes those that are operational.

This adjustment is intended to improve the library's usability in multi-devices.  I am open to any suggestions or further improvements from the community.

Thank you for watching this : )

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

I apologize, as I am uncertain about whom to tag. Perhaps @muellerzr or @pacman100 can review this. thx

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -811,7 +811,7 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
                     _ = torch.tensor(0, device=torch.device("npu", i))
                     max_memory.append({i: torch.npu.mem_get_info(i)[0]})
                 except Exception:
-                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
+                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                     continue
         elif is_mlu_available():
             for i in range(torch.mlu.device_count()):
@@ -819,7 +819,7 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
                     _ = torch.tensor(0, device=torch.device("mlu", i))
                     max_memory.append({i: torch.mlu.mem_get_info(i)[0]})
                 except Exception:
-                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
+                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                     continue
         elif is_xpu_available():
             for i in range(torch.xpu.device_count()):
@@ -827,7 +827,7 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
                     _ = torch.tensor(0, device=torch.device("xpu", i))
                     max_memory.append({i: torch.xpu.max_memory_allocated(i)})
                 except Exception:
-                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
+                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                     continue
         else:
             for i in range(torch.cuda.device_count()):
@@ -835,7 +835,7 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
                     _ = torch.tensor([0], device=i)
                     max_memory.append({i: torch.cuda.mem_get_info(i)[0]})
                 except Exception:
-                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
+                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                     continue
         # allocate everything in the mps device as the RAM is shared
         if is_mps_available():
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -803,27 +803,40 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]] 
     import psutil
 
     if max_memory is None:
-        if not (torch.cuda.is_available() or is_npu_available() or is_mlu_available() or is_xpu_available()):
-            max_memory = {}
-
-        else:
-            # Make sure CUDA is initialized on each GPU to have the right memory info.
-            if is_npu_available():
-                for i in range(torch.npu.device_count()):
+        max_memory = {}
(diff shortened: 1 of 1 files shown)


MeVeryHandsome:
Adding a logging statement such as `logger.warning(f"Device {i} is not available, proceeding to check subsequent devices.")` could be better : )

MeVeryHandsome:
By the way, if the unavailable GPU is the first one among the visible devices, even when the above method is used to obtain the maximum available VRAM, an error will still occur in this section: 
```
    module: nn.Module,
    tensor_name: str,
    device: Union[int, str, torch.device],
    value: Optional[torch.Tensor] = None,
    dtype: Optional[Union[str, torch.dtype]] = None,
    fp16_statistics: Optional[torch.HalfTensor] = None,
    tied_params_map: Optional[Dict[int, Dict[torch.device, torch.Tensor]]] = None,
):
    ...
        torch.cuda.empty_cache()
```
This is because PyTorch's cache clearing method, by default, affects the first device when no parameters are passed. However, this is a feature of Torch, I personally believe it may not be necessary to optimize this aspect within this project. Therefore, one could manually implement a try-catch block to handle this situation.

MeVeryHandsome:
@muellerzr hi, sorry to bother, could u please give me some suggestion😆

src/accelerate/utils/modeling.py
```diff
@@ -803,27 +803,40 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+        if is_npu_available():
+            for i in range(torch.npu.device_count()):
+                try:
                     _ = torch.tensor(0, device=torch.device("npu", i))
-                max_memory = {i: torch.npu.mem_get_info(i)[0] for i in range(torch.npu.device_count())}
-            elif is_mlu_available():
-                for i in range(torch.mlu.device_count()):
+                    max_memory.append({i: torch.npu.mem_get_info(i)[0]})
+                except Exception:
+                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```
SunMarc, line 814:
```suggestion
                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```

src/accelerate/utils/modeling.py
```diff
@@ -803,27 +803,40 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+        elif is_mlu_available():
+            for i in range(torch.mlu.device_count()):
+                try:
                     _ = torch.tensor(0, device=torch.device("mlu", i))
-                max_memory = {i: torch.mlu.mem_get_info(i)[0] for i in range(torch.mlu.device_count())}
-            elif is_xpu_available():
-                for i in range(torch.xpu.device_count()):
+                    max_memory.append({i: torch.mlu.mem_get_info(i)[0]})
+                except Exception:
+                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```
SunMarc, line 822:
```suggestion
                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```

src/accelerate/utils/modeling.py
```diff
@@ -803,27 +803,40 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+        elif is_xpu_available():
+            for i in range(torch.xpu.device_count()):
+                try:
                     _ = torch.tensor(0, device=torch.device("xpu", i))
-                max_memory = {i: torch.xpu.max_memory_allocated(i) for i in range(torch.xpu.device_count())}
-            else:
-                for i in range(torch.cuda.device_count()):
+                    max_memory.append({i: torch.xpu.max_memory_allocated(i)})
+                except Exception:
+                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```
SunMarc, line 830:
```suggestion
                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```

src/accelerate/utils/modeling.py
```diff
@@ -803,27 +803,40 @@ def get_max_memory(max_memory: Optional[Dict[Union[int, str], Union[int, str]]]
...
+                    continue
+        else:
+            for i in range(torch.cuda.device_count()):
+                try:
                     _ = torch.tensor([0], device=i)
-                max_memory = {i: torch.cuda.mem_get_info(i)[0] for i in range(torch.cuda.device_count())}
+                    max_memory.append({i: torch.cuda.mem_get_info(i)[0]})
+                except Exception:
+                    logger.warning(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```
SunMarc, line 838:
```suggestion
                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
```

SunMarc (approved):
Thanks @MeVeryHandsome for the PR ! I think it is best to use `info` instead of `warning`. As for your question about `empty_cache`, we use it a lot in other parts of the library and it won't make sense to try catch these each time. This is probably something that pytorch should try to fix !

PR #2685, opened 2024-04-18
Change dataloader send_to_device calls to non-blocking

drhead:
# What does this PR do?

Currently, it seems that even with pinned memory enabled on the dataloader, iterating over the dataloader causes a cudaStreamSynchronize call:

image

This is a devastating hit to performance if you are trying to overlap any other asynchronous transfers with your computations (such as offloading optimizer/EMA states), since it appears to be waiting for all async transfers to complete before continuing.

Setting these operations to be non-blocking, as well as using pinned memory for the dataloader, gets rid of the offending call:
image
(the timescale is very different here, this one is a fraction of a millisecond whereas the previous screenshot highlights a period of roughly one entire second)

With this, the tensors should only get forcibly synced when they are about to be used, and other non-blocking transfers will be left alone.  When I made this change, I had a noticeable speedup in my training loop where I am offloading an EMA state (3s/it -> 2.5s/it).

This also shouldn't cause issues when not using pinned memory.  It will just do a normal synchronous transfer as before.  The `send_to_device` utility function also has handling built in for any case where `Tensor.to()` does not accept `non_blocking` as an argument for whatever reason.  There shouldn't be any downsides to this change, as a result, so I don't see any reason to leave it as anything but a fixed value of True.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -524,6 +524,14 @@ class DataLoaderConfiguration:
             "multiple different seeds to compare. Should also be ran with [`~utils.set_seed`] for the best results."
         },
     )
+    non_blocking: bool = field(
+        default=False,
+        metadata={
+            "help": "If set to `True`, the dataloader prepared by the Accelerator will utilize non-blocking host-to-device"
+            " transfers, allowing for better overlap between dataloader communication and computation.  Requires that the"
+            " prepared dataloader has `pin_memory` set to `True` to work properly."
+        },
+    )
 
 
 @dataclass
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -528,7 +528,7 @@ class DataLoaderConfiguration:
         default=False,
         metadata={
             "help": "If set to `True`, the dataloader prepared by the Accelerator will utilize non-blocking host-to-device"
-            " transfers, allowing for better overlap between dataloader communication and computation.  Requires that the"
+            " transfers, allowing for better overlap between dataloader communication and computation.  Recommended that the"
             " prepared dataloader has `pin_memory` set to `True` to work properly."
         },
     )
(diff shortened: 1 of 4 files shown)


muellerzr:
As stated back here, we have reason to keep it at pytorch's default of `False`. What we can do though is add `non_blocking` as an argument to `utils.dataclasses.DataLoaderConfiguration` and have the user set it there. https://github.com/huggingface/accelerate/pull/607

Does this seem reasonable for you? :)

drhead:
> As stated back here, we have reason to keep it at pytorch's default of `False`. What we can do though is add `non_blocking` as an argument to `utils.dataclasses.DataLoaderConfiguration` and have the user set it there.
> 
> Does this seem reasonable for you? :)

I do think that would work.  I will look into setting that up and create appropriate documentation for it, including a warning about CPU operations on non-blocking tensors.  I do think that the common use case for a dataloader isn't threatened by this since this will usually be a single HtoD transfer with the tensor destroyed on device.

drhead:
This implementation works and has been tested.  I did add documentation notes where appropriate on the classes I modified and also added a warning in the Quickstart section about how to enable non-blocking transfers when migrating, it seems appropriate to bring it up there since that is where non-blocking calls would be made in vanilla Pytorch code.

src/accelerate/utils/dataclasses.py
```diff
@@ -524,6 +524,14 @@ class DataLoaderConfiguration:
             "multiple different seeds to compare. Should also be ran with [`~utils.set_seed`] for the best results."
         },
     )
+    non_blocking: bool = field(
+        default=False,
+        metadata={
+            "help": "If set to `True`, the dataloader prepared by the Accelerator will utilize non-blocking host-to-device"
+            " transfers, allowing for better overlap between dataloader communication and computation.  Requires that the"
+            " prepared dataloader has `pin_memory` set to `True` to work properly."
```
muellerzr, line 532:
```suggestion
            " transfers, allowing for better overlap between dataloader communication and computation.  Recommended that the"
            " prepared dataloader has `pin_memory` set to `True` to work properly."
```

Otherwise, we should raise a relevant error :)

muellerzr (approved):
Very nice! Just one small nit and we're good to go.

Also, you can fix the quality issues by doing:

```bash
pip install -e .[quality]
make style; make quality
```

muellerzr (approved):
Thanks a bunch! Nice work 🤗

PR #2686, opened 2024-04-18
Fix access error for torch.mps when using torch==1.13.1 on macOS

ringohoffman:
# What does this PR do?

When I run `accelerate` using `torch==1.13.1` on my macbook, I get:

```
        elif is_mps_available():
>           torch.mps.empty_cache()
E           AttributeError: module 'torch' has no attribute 'mps'
```

You can see the `torch.mps` module and the functions backing `torch.mps.empty_cache()` and `torch.mps.current_allocated_memory()` were not added until torch 2.0.0, whereas MPS support was added in 1.12.1:

https://github.com/pytorch/pytorch/blob/v1.13.1/torch/mps/__init__.py
https://github.com/pytorch/pytorch/blob/v1.13.1/torch/_C/__init__.pyi.in

So it is not safe to just use `is_mps_available()` to know if you can access `torch.mps`, you actually have to check if it exists before accessing it. There are plenty of other places where `is_mps_available()` is used and `torch.mps` is not accessed, so I don't think we should change its definition.

Ideally, `accelerate` would commit to running at least one 1.13.1 mac CI build, although I know it may not be feasible. The conditions I added appear to cover every access of `torch.mps`, but without the build, it seems likely there will be additional regressions.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr
@amyeroberts

### DIFF
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -63,7 +63,7 @@ def get_backend():
     elif is_cuda_available():
         return "cuda", torch.cuda.device_count(), torch.cuda.memory_allocated
     elif is_mps_available():
-        return "mps", 1, torch.mps.current_allocated_memory() if hasattr(torch, "mps") else 0
+        return "mps", 1, (torch.mps.current_allocated_memory() if hasattr(torch, "mps") else 0)
     elif is_mlu_available():
         return "mlu", torch.mlu.device_count(), torch.mlu.memory_allocated
     elif is_npu_available():
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -63,7 +63,7 @@ def get_backend():
     elif is_cuda_available():
         return "cuda", torch.cuda.device_count(), torch.cuda.memory_allocated
     elif is_mps_available():
-        return "mps", 1, torch.mps.current_allocated_memory()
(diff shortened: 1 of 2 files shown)


SunMarc (approved):
SGTM ! Could you have a second look @muellerzr ?

src/accelerate/test_utils/testing.py
```diff
@@ -63,7 +63,7 @@ def get_backend():
     elif is_cuda_available():
         return "cuda", torch.cuda.device_count(), torch.cuda.memory_allocated
     elif is_mps_available():
-        return "mps", 1, torch.mps.current_allocated_memory()
+        return "mps", 1, (torch.mps.current_allocated_memory() if hasattr(torch, "mps") else 0)
```
muellerzr, line 66: Rather than doing this, can we add a new arg in `is_mps_version` for `min_version` and check instead? I'd much rather do that than a `hasattr`

You can see an equivalent one for

muellerzr (changes requested):

SunMarc:
Hi @ringohoffman, are you planning to finish the PR ? Otherwise, I can take over !

ringohoffman:
@SunMarc you are welcome to!

SunMarc:
Closing this PR since it has been solved in the PR above.

PR #2687, opened 2024-04-18
[NULL AND VOID] Attempt at fixing sagemaker config issue

nroggendorff:
Updates config_args.py to work with the latest version of amazon sagemaker

### DIFF
--- a/src/accelerate/commands/config/config_args.py
+++ b/src/accelerate/commands/config/config_args.py
@@ -109,8 +109,6 @@ class BaseConfig:
             config_dict["use_cpu"] = False
         if "debug" not in config_dict:
             config_dict["debug"] = False
-        if "enable_cpu_affinity" not in config_dict:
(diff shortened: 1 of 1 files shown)


src/accelerate/commands/config/config_args.py
```diff
@@ -109,8 +109,6 @@ def from_json_file(cls, json_file=None):
             config_dict["use_cpu"] = False
         if "debug" not in config_dict:
             config_dict["debug"] = False
-        if "enable_cpu_affinity" not in config_dict:
-            config_dict["enable_cpu_affinity"] = False
```
muellerzr, line 113: We certainly can't have this, this removes CPU affinity entirely :) Can you please state what the error is in sagemaker? Maybe we need to add it to the sagemaker config

nroggendorff: Sagemaker doesnt support this arg

PR #2695, opened 2024-04-22
FEAT: Add LOMO optimizer

younesbelkada:
WIP - on par with https://github.com/huggingface/transformers/pull/30178


src/accelerate/accelerator.py
```diff
@@ -340,6 +341,8 @@ def __init__(
         self.init_handler = None
         self.fp8_recipe_handler = None
         self.autocast_handler = None
+        self._has_lomo_optimizer = False
```
muellerzr, line 344: We don't do private, so let's make this public please!
```suggestion
        self.has_lomo_optimizer = False 
```

src/accelerate/accelerator.py
```diff
@@ -1952,6 +1955,11 @@ def prepare_optimizer(self, optimizer: torch.optim.Optimizer, device_placement=N
         >>> optimizer = accelerator.prepare_optimizer(optimizer, device_placement=True)
         ```
         """
+        if is_lomo_available():
+            from lomo_optim import AdaLomo, Lomo
```
muellerzr, line 1959: This should be done at the top of the file

src/accelerate/accelerator.py
```diff
@@ -2031,6 +2041,8 @@ def backward(self, loss, **kwargs):
             return
         elif self.scaler is not None:
             self.scaler.scale(loss).backward(**kwargs)
+        elif learning_rate is not None and self._has_lomo_optimizer:
+            self._lomo_backward(loss, learning_rate)
```
muellerzr, line 2045: Does this need to interact with `self.scaler` for AMP? That may be important!

src/accelerate/accelerator.py
```diff
@@ -3279,3 +3291,16 @@ def verify_device_map(self, model: torch.nn.Module) -> bool:
                 return True
 
         return False
+
+    def _lomo_backward(self, loss: torch.Tensor, learning_rate: float) -> None:
```
muellerzr, line 3295: This should be in `.backward()`, where we can check for `learning_rate` as part of `**kwargs` IMO

src/accelerate/optimizer.py
```diff
@@ -134,13 +134,22 @@ def eval(self):
         return self.optimizer.eval()
 
     def step(self, closure=None):
+        if is_lomo_available():
+            from lomo_optim import AdaLomo, Lomo
```
muellerzr, line 138: Same thing as earlier, but let's move the detection for Lomo to happen in here rather than in the Accelerator. Then we can make an attribute `self.is_lomo_optimizer`

muellerzr (commented):
This is a very good start, left some design comments that are more in-tune to what I'd expect for Accelerate :)

src/accelerate/accelerator.py
younesbelkada: Doing that leads to an error due to circular imports 😢 this is because `lomo_optim` imports stuff from transformers, that itself import stuff from accelerate. Added a comment there

src/accelerate/accelerator.py
```diff
@@ -2031,6 +2041,8 @@ def backward(self, loss, **kwargs):
             return
         elif self.scaler is not None:
             self.scaler.scale(loss).backward(**kwargs)
+        elif learning_rate is not None and self._has_lomo_optimizer:
```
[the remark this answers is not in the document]
younesbelkada: Yes makes sense!

src/accelerate/accelerator.py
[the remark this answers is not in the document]
younesbelkada: Makes totally sense, refactored a bit things in [3a518dc](https://github.com/huggingface/accelerate/pull/2695/commits/3a518dc0043dd36905f0d5963ab7513f4c0cc6ce) - LMK what do you think !

src/accelerate/accelerator.py
```diff
@@ -3279,3 +3291,16 @@ def verify_device_map(self, model: torch.nn.Module) -> bool:
...
         return False
+
+    def _lomo_backward(self, loss: torch.Tensor, learning_rate: float) -> None:
+        """
+        Runs backward pass on LOMO optimizers.
+        """
+        if is_lomo_available():
+            from lomo_optim import AdaLomo, Lomo
+        else:
+            raise ValueError("`lomo_optim` package is needed to call backward on LOMO optimizers")
```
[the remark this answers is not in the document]
younesbelkada: Makes sense, removed it !

src/accelerate/accelerator.py
younesbelkada: hmm not sure here actually 🤯 I need to investigate a bit ..

muellerzr (approved):
Nice! This is looking fantastic! Last bit I’d like is a quick example in examples/ using LOMO, but for the sake of the release today we can get this in.

younesbelkada:
Nice, thanks so much ! OK I will work on an example that use accelerate + LOMO and open a follow up PR !

src/accelerate/accelerator.py
```diff
@@ -2027,6 +2030,13 @@ def prepare_optimizer(self, optimizer: torch.optim.Optimizer, device_placement=N
         >>> optimizer = accelerator.prepare_optimizer(optimizer, device_placement=True)
         ```
         """
+        if is_lomo_available():
+            # We need to import locally to avoid circular imports since lomo imports stuff from
+            # transformers & accelerate
+            from lomo_optim import AdaLomo, Lomo
+
+            self.has_lomo_optimizer = isinstance(optimizer, (Lomo, AdaLomo))
```
BenjaminBossan, line 2038: For my understanding, we assume that the accelerator could have multiple `self._optimizers`, with some of them LOMO and others not. Would that not create the issue that `self.has_lomo_optimizer` takes the value based on whatever the last optimizer is? Would we not have to set: `self.has_lomo_optimizer |= isinstance(optimizer, (Lomo, AdaLomo))`

Issue #2693, opened 2024-04-20
Fail to pass the accelerate test with mps as the device


JINNMnm:
### System Info

```Shell
- `Accelerate` version: 0.29.3
- Platform: macOS-14.1.1-arm64-arm-64bit
- `accelerate` bash location: /Users/jin/anaconda3/envs/openlrm/bin/accelerate
- Python version: 3.11.9
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.2.2 (False)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- System RAM: 36.00 GB
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - enable_cpu_affinity: False
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. First i use 
```
accelerate config
```
with the choice below to generate the config file:
```
In which compute environment are you running?
This machine                                                                                                                                                                                       
Which type of machine are you using?                                                                                                                                                               
No distributed training                                                                                                                                                                            
Do you want to run your training on CPU only (even if a GPU / Apple Silicon / Ascend NPU device is available)? [yes/NO]:no                                                                         
Do you wish to optimize your script with torch dynamo?[yes/NO]:no                                                                                                                                  
Do you wish to use FP16 or BF16 (mixed precision)? 
no                                                                                                                                                                                                 
accelerate configuration saved at /Users/jin/.cache/huggingface/accelerate/default_config.yaml     
```
2. Then i try:
```
accelerate test
```
to test the environment but i got the error like this:
```
Running:  accelerate-launch /Users/jin/anaconda3/envs/openlrm/lib/python3.11/site-packages/accelerate/test_utils/scripts/test_script.py
stdout: **Initialization**
stdout: Testing, testing. 1, 2, 3.
[...]

PR #2699, opened 2024-04-23
Block AMP for MPS device 

SunMarc:
# What does this PR do ? 
This PR fixes the support of AMP with mps. It was never supported in the first place. When this [PR](https://github.com/pytorch/pytorch/pull/99272) from pytorch gets merged, we can revert this PR. However, one can safely load a model in bf16 or fp16 with mps device with the latest torch. 

Fixes https://github.com/huggingface/accelerate/issues/2693


muellerzr (approved):
Oops 👀

Issue #2713, opened 2024-04-25
Llama3 8B with QLoRA ends in: You can't train a model that has been loaded in 8-bit precision on a different device than the one you're training on


oroojlooy:
### System Info

```Shell
torch==2.1.2
transformers==4.37.2
accelerate==0.26.1
numpy==1.26.3
OS type and version: fedora, 7.9
Python version: 3.11.0

I am using 8*A100-40Gb. 

Related bug: [1515](https://github.com/huggingface/accelerate/issues/1515)

I also tried the same code with brand new env, with the latest versions of every packages. The same error happens there.
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am calling an extension of QLoRA with `meta-llama/Meta-Llama-3-8B-Instruct`. I am using an extended version of `https://github.com/jondurbin/qlora.git`. The main parameters to set are: 
```
--model_name_or_path  meta-llama/Meta-Llama-3-8B-Instruct   --do_train   --lora_modules all   --bf16   --bits 4   --double_quant   --quant_type nf4 
```

### Expected behavior

The model should start fine-tuning, but it looks like that accelerate reserves most memory of `cuda:0` for some future operation and as a result the remaining memory is not enough for loading any part of the model. As a result, when in `accelerate/accelerator.py:1329` it checks for 
```
# if on the first device (GPU 0) we don't care
if (self.device.index is not None) or (current_device_index != 0)`
```
it founds that `current_device_index != 0` and it throws the error:
```
raise ValueError(
                        "You can't train a model that has been loaded in 8-bit precision on a different device than the one "
                        "you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device() or device_map={'':torch.xpu.current_device()}"
                    )
```

To add more details, I set `verbose=True` in `accelerate/utils/modeling.py:infer_auto_device_map()` where at `while len(modules_to_treat) > 0:` which designs the `device_map`. It returns the following parameters allocation:

```
Not enough space on 0 to put model (space available 167116948.79999995, module size 7572035584).
Splitting model.

Treating module model.embed_tokens.
[...]

Issue #2429, opened 2024-02-08
Accelerate refuse to work on `balanced_low_0` when GPU 0 is not filled.


xkszltl:
### System Info

```Shell
- `Accelerate` version: 0.26.1
- Platform: Linux-3.10.0-1160.105.1.el7.x86_64-x86_64-with-glibc2.35
- Python version: 3.10.12
- Numpy version: 1.26.3
- PyTorch version (GPU?): 2.2.0+cu121 (True)
- GPU type: NVIDIA TITAN V
- `Accelerate` default config:
        Not found
```

https://github.com/huggingface/accelerate/blob/6f05bbd41a179cc9a86238c7c6f3f4eded70fbd8/src/accelerate/accelerator.py#L1324-L1333

This part would throw if model is loaded by `transformers.AutoModelForCausalLM.from_pretrained(..., device_map="balanced_low_0", ...)`, because GPU 0 may be left completely unused.
This doesn't seem to be a good behavior as there's no way to tell what's the "first device" without computing the device map first.

SunMarc:
Hi @xkszltl , could you tell me more about your specific use case and the issue that you are facing ? I don't understand the part about the device map since we do compute it before and we do `model_devices =  model_devices = set(model.hf_device_map.values())`
Note that the related [PR](https://github.com/huggingface/accelerate/pull/1155) to this section is this one.

xkszltl:
We are loading LLAMA2 7B and Mistral for finetuning in a single node 8 GPU setup.
It works with balanced but not with balanced_low_0 because of exception thrown here.
I assume that means model device is not GPU0 probably the make room for low_0 requirement?

xkszltl:
The part I don’t understand is what make GPU 0 so special and worth asserting here?

SunMarc:
Hi @xkszltl, this check might be outdated since we added the possibility to fine-tune BNB model with naive PP . I'll let @younesbelkada confirms this point !

xkszltl:
Not stale

xkszltl:
@SunMarc @younesbelkada 
Any update on this?

SunMarc:
Hi @xkszltl, sorry for the delay. Would you like to submit a PR to fix this and check that the tests are passing on transformers and accelerate ? Thanks !

xkszltl:
Not stale

SunMarc:
Closing this since this issue should be solved by this [PR](https://github.com/huggingface/accelerate/pull/2714)

PR #2714, opened 2024-04-26
fix bnb multi gpu training

SunMarc:
# What does this PR do ? 

This PR fixes a bug that users who want to perform multi-gpu training with bnb gets when the first gpu is not used. In this [PR](https://github.com/huggingface/accelerate/pull/1155), you see that we did not allow multi-gpu training before and [all the code after](https://github.com/huggingface/accelerate/pull/1155/files#:~:text=current_device_index%20%3D%20list,.device%3A) that assume that we are in a single gpu setup. Since then, we added the possibility to perform multi-gpu training using naive PP. However, the logic after the multi-gpu check stayed the same as you can see on [main](https://github.com/huggingface/accelerate/blob/6af157ea93dfbace1db88b0fdc7dfb568dfdd5a5/src/accelerate/accelerator.py#L1393) . A quick fix would be to just add an elif statement. 

Fixes https://github.com/huggingface/accelerate/issues/2713 https://github.com/huggingface/accelerate/issues/2429

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1387,16 +1387,18 @@ class Accelerator:
                     " In order to use 8-bit models that have been loaded across multiple GPUs the solution is to use Naive Pipeline Parallelism."
                     " Therefore you should not specify that you are under any distributed regime in your accelerate config."
                 )
-            current_device = list(model_devices)[0]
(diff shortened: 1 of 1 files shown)


younesbelkada (approved):
Makes sense thanks @SunMarc !

muellerzr (approved):
Good catch! 🚀

Issue #2715, opened 2024-04-26
RAM leakage 


markovalexander:
### System Info

```Shell
- `Accelerate` version: 0.29.3
- Platform: Linux-5.15.0-1042-oracle-x86_64-with-glibc2.31
- Python version: 3.10.14
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.2.2+cu121 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- System RAM: 2015.67 GB
- GPU type: NVIDIA A100-SXM4-80GB
- `Accelerate` default config:
	Not found

deepspeed==0.14.2
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

preparing model/optimiser for training with deepspeed 2 multiple time in one process results in RAM leakage
Simple script:

```python
import torch
from accelerate import Accelerator, DeepSpeedPlugin
from accelerate.utils.deepspeed import HfDeepSpeedConfig
import wandb


DS2_CONFIG = {
    "train_micro_batch_size_per_gpu": 1,
    "gradient_accumulation_steps": 16,
    "gradient_clipping": 1.0,
    "zero_allow_untested_optimizer": True,
    "bf16": {
        "enabled": True,
        "loss_scale": 0,
        "initial_scale_power": 16,
        "loss_scale_window": 1000,
        "hysteresis": 2,
        "min_loss_scale": 1
    },
    "zero_optimization": {
        "stage": 2,
        "allgather_partitions": True,
        "reduce_scatter": True,
        "reduce_bucket_size": 1e9,
        "allgather_bucket_size": 1e9,
        "overlap_comm": True,
        "contiguous_gradients": True
    }
}


class Model(torch.nn.Module):

    def __init__(self, *, ndim: int, hidden_dim: int, use_bias: bool):
        super().__init__()
        self.a = torch.nn.Linear(ndim, hidden_dim, bias=use_bias)
        self.b = torch.nn.Linear(hidden_dim, ndim, bias=use_bias)


wandb.init()
for i in range(1000):
    print("--" * 30 + "Attempt #{}".format(i) + "--" * 30)
[...]

PR #2716, opened 2024-04-26
Fixup `free_memory` to deal with garbage collection

muellerzr:
# What does this PR do?

Tweaks `Accelerator.free_memory` to take in `*objects` and return `*objects`. This is because we need to fully replace the values, doing just `del` doesn't *actually* help as much as it should. 

Solves https://github.com/huggingface/accelerate/issues/2715, graph:

Solves https://github.com/huggingface/transformers/issues/28178

image

(Weird read, I know, I think it's because they delay the reading of the computer for a few seconds). But I think because the graphs are identical that's **good**

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -3054,9 +3054,10 @@ class Accelerator:
         self._models = []
         self._dataloaders = []
         # Deepspeed needs a bit more prep
-        if self.deepspeed_engine_wrapped is not None:
-            self.deepspeed_engine_wrapped.destroy()
-        self.deepspeed_engine_wrapped = None
+        if hasattr(self, "deepspeed_engine_wrapped"):
+            if self.deepspeed_engine_wrapped is not None:
+                self.deepspeed_engine_wrapped.destroy()
+            self.deepspeed_engine_wrapped = None
         self.step = 0
         return objects
 
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -3053,6 +3053,9 @@ class Accelerator:
         self._optimizers = []
         self._models = []
         self._dataloaders = []
+        # Deepspeed needs a bit more prep
+        if self.deepspeed_engine_wrapped is not None:
+            self.deepspeed_engine_wrapped.destroy()
         self.deepspeed_engine_wrapped = None
         self.step = 0
         return objects
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -3057,7 +3057,7 @@ class Accelerator:
         if hasattr(self, "deepspeed_engine_wrapped"):
             if self.deepspeed_engine_wrapped is not None:
                 self.deepspeed_engine_wrapped.destroy()
-        self.deepspeed_engine_wrapped = None
+            self.deepspeed_engine_wrapped = None
         self.step = 0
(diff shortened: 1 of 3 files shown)


chiragjn:
I am running into a memory leak and noticed this PR

There is also a issue with deepspeed engine's destroy() not being called which is supposed to remove torch hooks that known to leak memory:
https://github.com/huggingface/transformers/issues/28178

What would be the best way to accomodate this ^ too

SunMarc (approved):
Nice catch ! Thanks for fixing this ! Note that the following snippet will return `[None]`. Not sure if we really care about this case. 
```
import torch

accelerator = Accelerator()
a = torch.ones(1000, 1000).cuda()
a = accelerator.free_memory(a)
```

muellerzr:
@SunMarc that is indeed expected (but no, we don't care about this use-case really).

I'll be sure to include `.destroy()` in deepspeed as part of this, noticed this bug first when I was starting to look into it

src/accelerate/accelerator.py
```diff
@@ -3045,19 +3045,23 @@ def free_memory(self):
...
         """
+        objects = release_memory(*objects)
         self._schedulers = []
         self._optimizers = []
         self._models = []
         self._dataloaders = []
-        self.deepspeed_engine_wrapped = None
+        # Deepspeed needs a bit more prep
+        if hasattr(self, "deepspeed_engine_wrapped"):
+            if self.deepspeed_engine_wrapped is not None:
```
pacman100, line 3058: `deepspeed_engine_wrapped` doesn't have destroy method, it should be `deepspeed_engine_wrapped.engine.destroy()`. Also, it would be better to handle DeepSpeed engine before the statement `objects = release_memory(*objects)`.

pacman100 (commented):
Thank you, @muellerzr, for fixing the `free_memory` utility and also handling the deepspeed nuances therein. I think `trainer` should have a method `free_memeory` which would call the accelerator's counterpart so that people can have access to this via Trainer too. What are your thoughts on this? 

Also left a comment.

pacman100:
> I am running into a memory leak and noticed this PR

> There is also a issue with deepspeed engine's destroy() not being called which is supposed to remove torch hooks that known to leak memory:
https://github.com/huggingface/transformers/issues/28178

> What would be the best way to accomodate this ^ too

Thank you @chiragjn for the pointers, helpful!

src/accelerate/accelerator.py
muellerzr: @pacman100 the trainer does already :) 

https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L1885

pacman100: It is called at the start of training, I meant that people would like to call this method as per their convenience after training.

muellerzr: Ahhh.

I'll think about how we can go about it, since the optimizer is baked into the trainer so the user really just needs to pass in the `model` I think

muellerzr: Done, please see https://github.com/huggingface/transformers/pull/30549

pacman100 (approved):
Thank you for iterating, LGTM!

Issue #2718, opened 2024-04-28
checkpointing error when using IterableDataset and dispatch_batches=False


Fadelis98:
### System Info

```Shell
- `Accelerate` version: 0.29.1
- Platform: Linux-5.10.134-13.1.al8.x86_64-x86_64-with-glibc2.35
- `accelerate` bash location: /mnt/data/ai4phys/hugengyuan/miniconda3/envs/qec/bin/accelerate
- Python version: 3.11.8
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.2.2 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- System RAM: 739.74 GB
- GPU type: NVIDIA GeForce RTX 3090
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI_GPU
        - mixed_precision: no
        - use_cpu: False
        - debug: True
        - num_processes: 4
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - enable_cpu_affinity: False
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

NOTE: The bug only happens for num_processes > 1
```python
from typing import Iterator
import torch
from torch.utils.data import IterableDataset,DataLoader

class FooSet(IterableDataset):
    def __iter__(self) -> Iterator:
        for i in range(100):
            yield i

from accelerate import Accelerator

def main():
    dset = FooSet()
    dloader = DataLoader(dset,batch_size=1)

    ac = Accelerator(
        dispatch_batches=False,
        split_batches=False,
    )

    dloader = ac.prepare(dloader)
    for data in dloader:
        ac.print(data)
        # anything else in a typical training script here
        ac.save_state('test_checkpoint')
        break
        
from accelerate import notebook_launcher

notebook_launcher(main,num_processes=2,)
```

It will raise the following error:
```
-- Process 0 terminated with the following error:
Traceback (most recent call last):
[...]

PR #2723, opened 2024-04-29
fix sampler serialization

SunMarc:
# What does this PR do ?
This PR fixes an issue with sampler serialization. The sampler do not necessarily have a sampler attribute. I think this is only the case when the sampler is a `SeedableRandomSampler`. To fix the issue, we create a getter and setter function to get the sample of the dataloader. 

### Example of failing script 
You need to use num_processes >= 2 :

```python 
from typing import Iterator
from torch.utils.data import IterableDataset,DataLoader

class FooSet(IterableDataset):
    def __iter__(self) -> Iterator:
        for i in range(100):
            yield i

dset = FooSet()
dloader = DataLoader(dset, batch_size=1)

dataloader_config = DataLoaderConfiguration(dispatch_batches=False, split_batches=False)

ac = Accelerator(
    dataloader_config = dataloader_config
)

dloader = ac.prepare(dloader)
# print(dloader)
for data in dloader:
    ac.print(data)
    # This will fail saying that dataloader.sampler.sampler do not exist
    ac.save_state('test_checkpoint')
```

Fixes https://github.com/huggingface/accelerate/issues/2718

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -502,6 +502,23 @@ class DataLoaderShard(DataLoader, DataLoaderStateMixin):
         else:
             return len(self.dataset)
 
+    def get_sampler(self):
+        sampler_is_batch_sampler = isinstance(self.sampler, BatchSampler)
+        if sampler_is_batch_sampler:
+            sampler = getattr(self.sampler, "sampler", None)
+        else:
+            sampler = getattr(self.batch_sampler, "sampler", None)
+        return sampler
+
+    def set_sampler(self, sampler):
+        sampler_is_batch_sampler = isinstance(self.sampler, BatchSampler)
+        if sampler_is_batch_sampler:
+            self.sampler.sampler = sampler
+        else:
+            self.batch_sampler.sampler = sampler
+            if hasattr(self.batch_sampler, "batch_sampler"):
+                self.batch_sampler.batch_sampler.sampler = sampler
+
 
 if is_torch_xla_available():
     import torch_xla.distributed.parallel_loader as xpl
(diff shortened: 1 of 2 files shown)


muellerzr:
Actually, unapproving for now. Let's implement this please (see the last message in the PR): https://github.com/huggingface/accelerate/pull/2080

src/accelerate/data_loader.py
```diff
@@ -989,12 +1023,7 @@ def prepare_data_loader(
         )
 
     if isinstance(sampler, SeedableRandomSampler) and use_seedable_sampler:
```
muellerzr, line 1025: Note that here this is an unprepared dataloader, so it'd be better to have this be a `util` function called `get_sampler` (though I'm open to having this also be attached to the DL as a passthrough, could be handy with the trainer)

SunMarc: it should be fine from here since the dataloader have been initialized above. I think you are thinking about [this part](https://github.com/huggingface/accelerate/pull/2723/files#diff-d9858283a2ced902233727f6fddde0a00831ad9a66a069e57231a5057d550bf6R918:~:text=if%20sampler_is_batch_sampler%3A-,if%20sampler_is_batch_sampler%3A,-sampler%20%3D%20getattr) where indeed this is unprepared

SunMarc: I created the util function and called it in the different class for now since there is no difference. Do you want me to put in this function a specific file ?

muellerzr: Ack sorry yeah I just noticed that line, my bad. Let's just do that here then it's fine!

muellerzr (approved):
Thanks for doing this!

PR #2730, opened 2024-05-01
Fix stacklevel in `logging` to log the actual user call site (instead of the call site inside the logger wrapper) of log functions

### LINKED ISSUE #2114: Logging does not respect the "%(lineno)s" and "%(filename)s".
### System Info

```Shell

- `Accelerate` version: 0.24.1
- Platform: Linux-6.2.0-35-generic-x86_64-with-glibc2.35
- Python version: 3.11.5
- Numpy version: 1.26.1
- System RAM: 125.48 GB
- GPU type: NVIDIA GeForce RTX 3090
        - distributed_type: NO
        - mixed_precision: bf16
        - use_cpu: False
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

When using `accelerate.logging.get_logger` to log, all of the logs says it came from `logging.py` at line 60.

Interestingly, it respects the current module's name correctly.

```python
import logging

from accelerate.logging import get_logger
from rich.logging import RichHandler

logger = get_logger(__name__)

if __name__ == "__main__":
    logging.basicConfig(
        format="%(name)s: %(lineno)s - %(message)s",
        datefmt="%m/%d %H:%M:%S",
        level=logging.INFO,
        handlers=[RichHandler(show_time=True, show_level=True, show_path=True)],
    )
    logger.info("Test")
```

The output looks something like this.

```
11/02 16:45:19 INFO     __main__: 60 - Test                                                logging.py:60
```

### Expected behavior

The line number and filename should respect where the `logger.info()` is called.

luowyang:
# Fix `stacklevel` in `logging`

Fixes #2114

Currently, `accelerate.logging` does not correctly log the *user* call site of log functions such as `.log()` and `.info()`. Consider the following example modified from issue #2114:

```python
import logging

from accelerate.logging import get_logger
from rich.logging import RichHandler

logger = get_logger(__name__)

if __name__ == "__main__":
    accelerator = Accelerator()
    logging.basicConfig(
        format="%(filename)s:%(name)s:%(lineno)s - %(message)s",
        datefmt="%m/%d %H:%M:%S",
        level=logging.INFO,
        handlers=[RichHandler(show_time=True, show_level=True, show_path=True)],
    )
    logger.info("Test")
```

In its current state, the above script shows something like

```
05/01 19:26:22 INFO     logging.py:__main__:61 - Test
```

The calling site of the logging function `.info()` is incorrectly shown as [line 61](https://github.com/huggingface/accelerate/blob/232ebd159a3ff37a1089fbd4e7fc85ee915508ee/src/accelerate/logging.py#L61) of `accelerate.logging.MultiProcessAdapter.log`, which is the calling site inside the logger wrapper `MultiProcessAdapter`. This happens because the `stacklevel` argument of the logging function is set to its default value `1`, which shows the calling site of the wrapped logger's `.log()`. See the [docs](https://docs.python.org/3/library/logging.html#logging.Logger.debug) of Python `logging` for more details of what `stacklevel` means.

There exists a simple fix to the above issue: modify `stacklevel` in `MultiProcessAdapter.log` to `2` (skip `MultiProcessAdapter.log` in the calling stack), unless the users explicitly provide their own `stacklevel`. This fix is summarized as one line of additional code:

```python
kwargs.setdefault("stacklevel", 2)
```

After applying the above fix, the script mentioned earlier prints the correct message (`test_logging.py` is the script's name):

```
05/01 19:27:00 INFO     test_logging.py:__main__:17 - Test
```

## Before submitting
      Pull Request section?
      to it if that's the case.
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification). **No. This PR is a bug fix and contains no API change.**
- [ ] Did you write any new necessary tests? **No. This PR is a simple fix that does not change any expected behavior.**

## Who can review?

@kiyoon @BenjaminBossan

### DIFF
--- a/tests/test_logging.py
+++ b/tests/test_logging.py
@@ -1,0 +1,91 @@
+# Copyright 2024 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import inspect
+import logging
+import os
+
+import pytest
+
+from accelerate import Accelerator
+from accelerate.logging import get_logger
+
+
+def current_lineno() -> int:
+    # A simple helper that returns the lineno of its call-site.
+    caller_frame = inspect.currentframe().f_back
+    caller_info = inspect.getframeinfo(caller_frame)
+    return caller_info.lineno
+
+
+class CustomLogger(logging.LoggerAdapter):
+    # Mocks a user-defined custom logger wrapper that sets `stacklevel=3`.
+    def log(self, level, msg, *args, **kwargs):
+        # E.g. the user wants to modify `stacklevel`, `accelerate.logging`
+        # should respect the user's `stacklevel`. For the specific value
+        # of `3`, calling `CustomLogger.log()`, etc., should log that callsite,
+        # rather than the callsite of the following `self.logger.log()`.
+        kwargs["stacklevel"] = 3
+        self.logger.log(level, msg, *args, **kwargs)
+
+
+@pytest.fixture(scope="module")
+def accelerator():
+    return Accelerator()
+
+
+@pytest.mark.usefixtures("accelerator")
+def test_log_stack(caplog):
+    logger = get_logger(__name__)
+    logging.basicConfig(
+        format="%(filename)s:%(name)s:%(lineno)s:%(funcName)s - %(message)s",
+        datefmt="%m/%d %H:%M:%S",
+    )
+
+    message = "Test"
+    lineno = current_lineno() + 1  # the next line is the actual callsite
+    logger.warning(message)
+
+    assert len(caplog.records) == 1
+    rec = caplog.records[0]
+    assert rec.levelname == logging.getLevelName(logging.WARNING)
+    assert rec.filename == os.path.basename(__file__)
+    assert rec.name == __name__
+    assert rec.lineno == lineno
+    assert rec.funcName == test_log_stack.__name__
+    assert rec.message == message
+
+
+@pytest.mark.usefixtures("accelerator")
+def test_custom_stacklevel(caplog):
+    wrapped_logger = get_logger(__name__)
+    logging.basicConfig(
+        format="%(filename)s:%(name)s:%(lineno)s:%(funcName)s - %(message)s",
+        datefmt="%m/%d %H:%M:%S",
+    )
+    logger = CustomLogger(wrapped_logger, {})
+
+    message = "Test"
+    lineno = current_lineno() + 1  # the next line is the actual callsite
+    logger.warning(message)
+
+    # `CustomLogger.log` set custom `stacklevel=3`, so `logger.warning` should
+    # log its callsite (rather than those of the `warpped_logger`).
+    assert len(caplog.records) == 1
+    rec = caplog.records[0]
+    assert rec.levelname == logging.getLevelName(logging.WARNING)
+    assert rec.filename == os.path.basename(__file__)
+    assert rec.name == __name__
+    assert rec.lineno == lineno
+    assert rec.funcName == test_custom_stacklevel.__name__
+    assert rec.message == message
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
Sounds good, thanks for the fix. Would you be so kind to add a test for this bug? Your example should be a good template for testing.

luowyang:
> Sounds good, thanks for the fix. Would you be so kind to add a test for this bug? Your example should be a good template for testing.
> 
> The failing doc CI should be fixed if you merge with/rebase on main.

During writing tests for `logging`, I notice another possible bug: the processes hang if one passes `main_process_only=True` and `in_order=True` to the logging functions. An example is the following script:

```python
import logging

from accelerate.logging import get_logger
from rich.logging import RichHandler

logger = get_logger(__name__)

if __name__ == "__main__":
    accelerator = Accelerator()
    logging.basicConfig(
        format="%(filename)s:%(name)s:%(lineno)s - %(message)s",
        datefmt="%m/%d %H:%M:%S",
        level=logging.INFO,
        handlers=[RichHandler(show_time=True, show_level=True, show_path=True)],
    )
    logger.info(f"process index: {accelerator.process_index}", in_order=True, main_process_only=True)
```

The reason is that all process other than the main process [wait for the main process](https://github.com/huggingface/accelerate/blob/232ebd159a3ff37a1089fbd4e7fc85ee915508ee/src/accelerate/logging.py#L69) in this case, but the main process goes into [another conditional branch](https://github.com/huggingface/accelerate/blob/232ebd159a3ff37a1089fbd4e7fc85ee915508ee/src/accelerate/logging.py#L59) that never waits the remaining processes.

Since this PR is specific to `stacklevel`, I will leave this issue now, and omit it in the tests. But I think it should be fixed some time later.

luowyang:
> Sounds good, thanks for the fix. Would you be so kind to add a test for this bug? Your example should be a good template for testing.
> 
> The failing doc CI should be fixed if you merge with/rebase on main.

@BenjaminBossan Tests are added for `logging`. However, note that some bugs exist with `in_order` as mentioned [above](https://github.com/huggingface/accelerate/pull/2730#issuecomment-2091234513), so I omit tests for it.

tests/test_logging.py
```diff
@@ -0,0 +1,323 @@
...
+        return cls(
+            levelname=levelname.upper(),
+            filename=filename,
+            lineno=lineno,
+            funcName=funcName,
+            processIndex=processIndex,
+        )
+
+
+def assert_equals(value, expected, prefix=""):
```
BenjaminBossan, line 108: We don't need this, pytest already generates readable error messages.

BenjaminBossan (changes requested):
Thanks for adding tests to this PR. In fact, you added quite a lot of tests, since there are no existing dedicated tests for the logger (at least that I could find, ping @muellerzr).

However, I feel like this extensive testing detracts from the core issue that we want to check in this PR and I would suggest to create a separate PR to add general tests for the logger (and potentially solve the bug you found).

For this specific PR, I think a very simple, focused test should be sufficient, for instance:

```python
import logging
import pytest

from accelerate.logging import MultiProcessAdapter, get_logger


@pytest.fixture(scope="module")
def accelerator():
    return Accelerator()


@pytest.mark.usefixtures("accelerator")
def test_log_stack(caplog):
    logger = get_logger(__name__)
    logging.basicConfig(
        format="%(filename)s:%(name)s:%(lineno)s - %(message)s",
        datefmt="%m/%d %H:%M:%S",
    )
    logger.warning("Test")
    assert len(caplog.records) == 1
    rec = caplog.records[0]
    assert rec.filename == <name of test file>
    # ... more attributes if so desired
```

Note that I use pytest to simplify some stuff and if you indeed want to create a separate PR to add general tests, I would recommend to take a similar approach. As is, I find the tests hard to understand, as they use a lot of custom classes and use the logger in a very different way from what an accelerate user would typically do.

tests/test_logging.py
luowyang: I'm not quite an expert in testing, so thanks for your comments. I have re-written the tests into two simple tests: 1) check the logged message is correct in a normal setting; and 2) check the logged message is correct when the user overrides `stacklevel` in a custom logger wrapper. These tests are most relevant to this PR, and hopefully they can serve as a starting point for future tests.

BenjaminBossan (approved):
Thanks for simplifying the tests, they are much easier to understand now.

Something I like to do with tests for bugfixes is to add a comment that references the original issue to quickly find out why a test exists, but it's not a must have.

@muellerzr please check if this can be merged.

muellerzr (approved):
Thanks for doing this and nicely done with the tests!

PR #2737, opened 2024-05-03
handle weight sharing with init_on_device

aws-rhsoln:
# What does this PR do?
This PR handles the case when we are using meta_device and have tied weights. During model init, when we have a case, where `weight2 = weight1` , this would result in register_param getting called again. This call will re-init the param instead of using weight1. This will result in relationship between the weights getting lost.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -128,7 +128,14 @@ def init_on_device(device: torch.device, include_buffers: bool = None):
             param_cls = type(module._parameters[name])
             kwargs = module._parameters[name].__dict__
             kwargs["requires_grad"] = param.requires_grad
-            module._parameters[name] = param_cls(module._parameters[name].to(device), **kwargs)
+            # When we have a case of tensor2 = tensor1, it would call the set_attr
+            # of param, which in turn would call the register_parameter API.
+            # In this case, the new param is already on meta-device, since it was moved
+            # previously when it was initialized. Hence, when resetting, you can
+            # directly assign that tensor instead of re-init. If you re-init you would
+            # lose the relationship.
+            module._parameters[name] = param if param.device == device else \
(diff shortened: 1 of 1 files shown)


SunMarc (commented):
Hi @aws-rhsoln, thanks for the PR ! Could you add a few tests ? Also, could you explain in which context you needed that ?

SunMarc:
Hi @aws-rhsoln, are you still planning to finish the PR ? Thanks !

aws-rhsoln:
I'll add some tests soon. Context for the change: For our pipeline parallel implementation [here](https://github.com/aws-neuron/neuronx-distributed/blob/main/src/neuronx_distributed/pipeline/model.py#L46), we first create the model and then trace it, use the traced model to partition for PP. Now we create the model using meta-device and then copy the weights into device after PP init. When copying weights, if we have shared weights such that `self.output_layer.weight = self.embedding.weight` , the init_on_device api would first initialize embedding weight and then when its about to initialize output.layer weight, since its the embedding weight, its already on device and hence would end up throwing an error. Hence to fix it, this PR was introduced.

SunMarc:
Awesome ! Thanks for the update !

SunMarc:
Any news about this PR @aws-rhsoln ? ;) If you can't make it, can you maybe show us the script didn't work, so that other can fix it ?

Ghost-LZW:
I’ve added a unit test based on my use case https://github.com/huggingface/accelerate/issues/3668 (transformers tie_weight). 

https://github.com/Ghost-LZW/accelerate/commit/892ba83647419a1fa8dc365ee5449385df20b25f

Can we cherry-pick this and apply the fix? @SunMarc @aws-rhsoln

SunMarc:
Feel free to reopen a fix to add this ! @Ghost-LZW

PR #2745, opened 2024-05-06
LOMO / FIX: Support multiple optimizers

younesbelkada:
# What does this PR do?

Addresses: https://github.com/huggingface/accelerate/pull/2695#discussion_r1589164607 so that multiple LOMO optimizers can be used !

cc @muellerzr @BenjaminBossan


BenjaminBossan (approved):
Thanks.

PR #2751, opened 2024-05-07
Fix max_memory assignment

SunMarc:
#What does this PR do ?
This PR fixes a bug that was introduced in this [PR](https://github.com/huggingface/accelerate/pull/2681). max_memory is a dict so that can't use append 🤦. I've checked that this should fix the failing CI.


SunMarc:
Merging asap as this might affect a lot of users because of big model inference

Issue #2744, opened 2024-05-05
SageMaker config stopped working in Accelerate version 0.29+


ivan-khvostishkov:
### System Info

```Shell
The `enable_cpu_affinity` property is missing in `SageMakerConfig` class:

https://github.com/huggingface/accelerate/blob/v0.30.0/src/accelerate/commands/config/config_args.py#L228-L243

But this property is expected by the config parser:

https://github.com/huggingface/accelerate/blob/v0.30.0/src/accelerate/commands/config/config_args.py#L148-L155

As a result, SageMaker config doesn't parse anymore producing the exception:


ValueError: The config file at .../.cache/huggingface/accelerate/default_config.yaml had unknown keys (['enable_cpu_affinity']), please try upgrading your `accelerate` version or fix (and potentially remove) these keys from your config file.
```
```


### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

### Reproduction

1. Create config with `compute_environment: AMAZON_SAGEMAKER` .
2. Run `accelerate config update`

### Expected behavior

Message `Sucessfully updated the configuration file`

muellerzr:
Thanks @ivan-khvostishkov, can you test it out via this PR? https://github.com/huggingface/accelerate/pull/2753

You can do `pip install git+https://github.com/huggingface/accelerate@fix-sagemaker`

PR #2753, opened 2024-05-07
Fix sagemaker config

muellerzr:
# What does this PR do?

Puts `enable_cpu_affinity` down a level so sagemaker can work.

We already have checks to verify if we should call cpu_affinity, so its safe to do (and will default to `False` if they didn't pass it in)

Fixes https://github.com/huggingface/accelerate/issues/2744

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/commands/config/config_args.py
+++ b/src/accelerate/commands/config/config_args.py
@@ -79,6 +79,7 @@ class BaseConfig:
     mixed_precision: str
     use_cpu: bool
     debug: bool
+    enable_cpu_affinity: bool
 
     def to_dict(self):
         result = self.__dict__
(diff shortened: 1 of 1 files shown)


muellerzr:
(Will be safe to merge after the user can verify :) )

BenjaminBossan (commented):
> Puts enable_cpu_affinity down a level so sagemaker can work.

Not quite sure what that means, but if the user can confirm that this fixes the issue, it should be good. However, it looks like the new test is failing because it requires sagemaker to be installed, can you verify?

muellerzr:
@BenjaminBossan updated. Will also re-look at the test to make sure it doesn't need to be ran per-se, just check for config validity

BenjaminBossan (approved):
Thanks, LGTM, tests seem to be running correctly.

nroggendorff:
thank you!

PR #2755, opened 2024-05-09
fix cpu omp num threads set

jiqing-feng:
Hi @muellerzr 

This PR fixed the cpu num threads set. We should set a default value if the env parameter `OMP_NUM_THREADS` is not set.
Besides, we should keep `OMP_NUM_THREADS` instead of replacing it in `launch.py`. Could you pls review it? Thx!


muellerzr (approved):
This makes sense to me from what I can tell. Thanks for the fine-tuning :)

muellerzr:
@jiqing-feng can you run `make style; make quality`?

jiqing-feng:
> @jiqing-feng can you run `make style; make quality`?

Done.

PR #2756, opened 2024-05-09
optimize get_module_leaves speed

BBuf:
### Background
When I try to inference deepseek-v2 with transformers:

```python
import torch
from transformers import AutoTokenizer, AutoModelForCausalLM, GenerationConfig

model_name = "deepseek-ai/DeepSeek-V2"
tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
# `max_memory` should be set based on your devices
max_memory = {i: "75GB" for i in range(8)}
# `device_map` cannot be set to `auto`
model = AutoModelForCausalLM.from_pretrained(model_name, trust_remote_code=True, device_map="sequential", torch_dtype=torch.bfloat16, max_memory=max_memory, attn_implementation="eager")
model.generation_config = GenerationConfig.from_pretrained(model_name)
model.generation_config.pad_token_id = model.generation_config.eos_token_id

text = "An attention function can be described as mapping a query and a set of key-value pairs to an output, where the query, keys, values, and output are all vectors. The output is"
inputs = tokenizer(text, return_tensors="pt")
outputs = model.generate(**inputs.to(model.device), max_new_tokens=100)

result = tokenizer.decode(outputs[0], skip_special_tokens=True)
print(result)

```

I found that program with stuck...

### Solve
In fact, the stuck appearing [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/modeling.py#L1041) . 

图片

For DeepSeek V2, the length of `module_sizes` is 68185, and the code here has a complexity of O(N^2) (where `N = module_sizes`), which requires a very long time to execute, giving the illusion of being stuck. This PR optimizes the code to have a complexity of O(N), allowing it to quickly reach the stage of loading the large model. The inference results on an 8xA800 machine are also normal after this optimization.

img_v3_02an_91e8f57c-fb20-415f-84e8-8effc882f8bg

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -928,6 +928,16 @@ def load_offloaded_weights(model, index, offload_folder):
         weight = load_offloaded_weight(tensor_file, metadata)
         set_module_tensor_to_device(model, param_name, "cpu", value=weight, fp16_statistics=fp16_statistics)
 
+def get_module_leaves(module_sizes):
+    module_children = {}
+    for module in module_sizes:
+        if module == "":
+            continue
+        parent = module.rsplit(".", 1)[0]
+        module_children[parent] = module_children.get(parent, 0) + 1
+    
+    leaves = [module for module in module_sizes if module_children.get(module, 0) == 0]
+    return leaves
 
 def get_balanced_memory(
     model: nn.Module,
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -929,6 +929,17 @@ def load_offloaded_weights(model, index, offload_folder):
...
 
 
+def get_module_leaves(module_sizes):
+    module_children = {}
+    for module in module_sizes:
+        if module == "":
+            continue
+        parent = module.rsplit(".", 1)[0]
+        module_children[parent] = module_children.get(parent, 0) + 1
+    leaves = [module for module in module_sizes if module_children.get(module, 0) == 0]
```
SunMarc, line 939: Small nit even though it won't impact the logic in `get_balanced_memory`. 
```suggestion
    leaves = [module for module in module_sizes if module_children.get(module, 0) == 0 and module != ""]
```

SunMarc (approved):
Thanks for this great PR @BBuf ! Just a small nit.

src/accelerate/utils/modeling.py
BBuf: updated

muellerzr (approved):
Thanks for making this much more efficient and fixing 🤗

Issue #2731, opened 2024-05-01
The logic of some methods is too complex


Iain-S:
The quantity of conditional logic in some methods/functions makes them difficult to follow. For example, [PartialState.\_\_init\_\_](https://github.com/huggingface/accelerate/blob/2c767338f29989e54ce93b3036b41e02013af7a7/src/accelerate/state.py#L111) has a cyclomatic complexity of 25 and [Accelerator.\_\_init\_\_](https://github.com/huggingface/accelerate/blob/2c767338f29989e54ce93b3036b41e02013af7a7/src/accelerate/accelerator.py#L246) has a CC of 54 (as measured by enabling the "C90" ruff rule, which has a default max CC of 10). 

It looks like PartialState and AcceleratorState are mostly trying to:

1. Dynamically import libraries such as ipex and deepspeed.
2. Set environment variables needed by those libraries and/or other parts of Accelerate.
3. Choose which launcher to use.
4. Perform some sanity checks.

...but those concerns have become entangled. I think they could be simplified and am happy to give it a shot if the PR would be welcome, though I would probably need to know a bit more about the motivation behind the Borg-esque state dict shared between the two state classes.

In addition to lowering the complexity of some methods for the reader, I would suggest checking for pre-existing env var values before overriding them. For example, it's plausible that a user could manually set CCL_PROCESS_LAUNCHER only to have it [silently overridden](https://github.com/huggingface/accelerate/blob/2c767338f29989e54ce93b3036b41e02013af7a7/src/accelerate/state.py#L222). It should be straightforward to add a convenience function that checks for an existing, conflicting value before writing a new one.

muellerzr:
You can try, certainly, however those methods are that complex/centralized for a reason. After the recent refactor as well they're done in a way to allow for easier leveraging of new backends. 

Can you explain with an example of where you see they have become "entangled"?

Iain-S:
> they're done in a way to allow for easier leveraging of new backends

Great, I shall bear that in mind.

> Can you explain with an example of where you see they have become "entangled"?

Absolutely, though I won't have a chance for a couple of days.

Iain-S:
OK, I think I can summarise the thing that I found confusing. There are effectively four ways that "config" is handled (where "config" means some kind of run-time option):

1. CLI options, such as `--cpu`.
2. Config file values, such as `use_cpu: true`, which are loaded into Base/Cluster/SageMaker Config objects.
[...]

PR #2757, opened 2024-05-09
Simplify CLI args validation and ensure CLI args take precedence over config file.

Iain-S:
# What does this PR do?

- We simplify the processsing of config file sub-sections (e.g. fsdp, megatron, ipex, deepspeed, etc). Previously, any time a dict was encountered in a config, all dicts in the config were re-processed. 
- CLI options such as --megatron_some_option will no longer be overridden by config options (there was previously no `if getattr(args, name, None)` check before overwriting the `arg.name`). It makes things more consistent but note that it does change existing behaviour.

## Before submitting

      Pull Request section?
      to it if that's the case.

## Issues

Somewhat related to #2731.

## Who can review?

@muellerzr

## Notes for Reviewers

I've made the logic in `_validate_launch_command()` more explicit as the proposed

```python
    if args.config_file is not None or (os.path.isfile(default_config_file) and not args.cpu):
```

is equivalent to the previous

```python
    if args.config_file is not None or os.path.isfile(default_config_file) and not args.cpu:
```

However, if what we really want is

```python
    if (args.config_file is not None or os.path.isfile(default_config_file)) and not args.cpu:
```

then I'll happily edit but note that it would be changing the existing behaviour.


muellerzr (approved):
Nice refactor and tests! Thanks for doing this 🤗

muellerzr:
FYI this broke FSDP as defaults for that are not always `None`. Making a follow-up

Iain-S:
Let me know if I need to resubmit this PR. If you could suggest a test case to catch the problem you found, that would be great.

cyr0930:
hmm priority looks very messy

PR #2765, opened 2024-05-11
Support saving and loading of step while saving and loading state

### LINKED ISSUE #2702: Is it possible to get `save_state` to save `step` as well?
Is it possible to get `save_state` to save `accelerator.step` as well? Or is there another way to save and load the current iteration/step?

bipinKrishnan:
# What does this PR do?

This PR adds support to checkpoint the `step`(internal step tracker) value while calling `Accelerator.save_state` and also makes appropriate changes to `Accelerator.load_state` to support the loading of the saved `step`.

Fixes #2702 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -55,6 +55,7 @@ def save_accelerator_state(
     schedulers: list,
     dataloaders: list,
     process_index: int,
+    step: int,
     scaler: GradScaler = None,
     save_on_each_node: bool = False,
     safe_serialization: bool = True,
@@ -82,6 +83,8 @@ def save_accelerator_state(
             A list of dataloader instances to save their sampler states
         process_index (`int`):
             The current process index in the Accelerator state
+        step (`int`):
+            The current step in the internal step tracker
         scaler (`torch.cuda.amp.GradScaler`, *optional*):
             An optional gradient scaler instance to save
         save_on_each_node (`bool`, *optional*):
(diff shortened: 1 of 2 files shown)


muellerzr (approved):
Sorry for the delay! Nice work, this looks great to me.

Let's modify the docstring for `load_accelerator_state` and mention that it may return, and can you rebase from `main`? This should make the failing tests pass

bipinKrishnan:
Hi @muellerzr, thanks for the feedback! Just pushed the updates, let me know if this is good to go :)

muellerzr (approved):
Very nice, thanks!

src/accelerate/checkpointing.py
```diff
@@ -240,6 +249,7 @@ def load_accelerator_state(
     # Random states
     try:
         states = torch.load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"))
+        override_attributes["step"] = states["step"]
```
jmracek, line 252: This is a breaking change for people upgrading accelerate.  If previously saved pickled states objects don't contain a "step" key, then random states fail to initialize correctly when you resume a training that was checkpointed from a prior version of accelerate.

muellerzr: Correct, we've fixed this in a patch release

jmracek: Ah I see [this](https://github.com/huggingface/accelerate/pull/2992) now.  Thanks!

PR #2770, opened 2024-05-13
Fix broken PiPPy IR and PipelineStage imports

yaxan:
# What does this PR do?
Updates the PiPPy imports to not use the private _IR and _PipelineStage modules

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/src/accelerate/inference.py
+++ b/src/accelerate/inference.py
@@ -28,8 +28,15 @@ from .utils import (
 )
 
 
-if is_pippy_available():
+pippy_version = is_pippy_available()
+
+if pippy_version == "0.1.1-0.2.0":
+    from pippy.IR import Pipe, PipeSplitWrapper, annotate_split_points
+    from pippy.PipelineStage import PipelineStage
+    split_point_class = PipeSplitWrapper.SplitPoint
+elif pippy_version == ">0.2.0":
     from pippy import Pipe, PipelineStage, SplitPoint, annotate_split_points
+    split_point_class = SplitPoint
 
 
 def generate_device_map(model, num_processes: int = 1, no_split_module_classes=None, max_memory: dict = None):
@@ -84,7 +91,7 @@ def build_pipeline(model, split_points, args, kwargs, num_chunks):
     """
     # We need to annotate the split points in the model for PiPPy
     state = PartialState()
-    annotate_split_points(model, {split_point: SplitPoint.BEGINNING for split_point in split_points})
+    annotate_split_points(model, {split_point: split_point_class.BEGINNING for split_point in split_points})
     found_batch_size = find_pippy_batch_size(args, kwargs)
     if found_batch_size != num_chunks:
         if args is not None:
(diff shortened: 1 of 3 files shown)


yaxan:
I realize now I'm running a later version of PiPPy, so these changes will be needed eventually I think - but not on the current stable release.

yaxan:
> I realize now I'm running a later version of PiPPy, so these changes will be needed eventually I think - but not on the current stable release.

Might be better to switch the imports to `torch.distributed.pipelining` and remove standalone PiPPy library when the time comes anyways.

src/accelerate/inference.py
```diff
@@ -29,8 +29,7 @@
 
 
 if is_pippy_available():
-    from pippy.IR import Pipe, PipeSplitWrapper, annotate_split_points
-    from pippy.PipelineStage import PipelineStage
+    from pippy import Pipe, PipelineStage, SplitPoint, annotate_split_points
```
muellerzr, line 32: Let's modify `is_pippy_available` to give a `min_version` flag, then do an if/else to bring in the right imports

muellerzr (commented):
Thanks for making sure we're on the latest and greatest with PiPPy! 

I don't think it's bad to future-proof with where they are now, let's make sure to do an `if/else` for the import for the time being. Once it's in PyTorch fully then we can change this to be a min torch version and swap those imports out :)

src/accelerate/utils/imports.py
```diff
@@ -177,7 +177,10 @@ def is_pippy_available():
     package_exists = _is_package_available("pippy", "torchpippy")
     if package_exists:
         pippy_version = version.parse(importlib.metadata.version("torchpippy"))
-        return compare_versions(pippy_version, ">", "0.1.1")
+        if compare_versions(pippy_version, ">", "0.1.1") and compare_versions(pippy_version, "<=", "0.2.0"):
+            return "0.1.1-0.2.0"
+        elif compare_versions(pippy_version, ">", "0.2.0"):
+            return ">0.2.0"
```
muellerzr, line 183: Rather than this, let's keep the current behavior of `true` or `false`. 

Let's do a different argument of `max_version` to be passed in the `is_pippy_available` function with a default of `None`

muellerzr (commented):
Close, but not quite.

src/accelerate/utils/imports.py
yaxan: Ah I see. I didn't fully understand what you meant before but I think I get it now.
Let me know!

yaxan:
@muellerzr I assume this should wait until pytorch 2.4?
Do you want to support all versions of PiPPy and torch.distributed.pipelining, just the latter, or some other combination?

muellerzr:
@yaxan yes indeed. 

The last version of `PiPPy`, for a few months, and then `torch.distributed.pipelining`

PR #2774, opened 2024-05-14
Fix small edge case in get_module_leaves

SunMarc:
# What does this PR do ?
This PR fixes a small edge case when we compute the leaves of a model introduced in this [PR](https://github.com/huggingface/accelerate/pull/2756) . This should solves the failing test in our CI.


muellerzr (approved):
Thanks for the quick fix!

PR #2776, opened 2024-05-14
Skip deepspeed test

SunMarc:
# What does this PR do ? 
This PR skips the deepspeed test since it started failing with deepspeed > 0.12.6. This is most probably an issue on deepspeed side cc @pacman100. Let's make sure to not forget about this test :/ .


muellerzr (approved):
Thanks for doing this 🤗

PR #2777, opened 2024-05-14
Use partial state for fsdp saving/loading

muellerzr:
# What does this PR do?

As I continue my dive, updates `save_fsdp_*` and `load_fsdp_*` to use the `PartialState` instead, getting rid of the need to couple it with an `Accelerator`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@pacman100 @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2903,7 +2903,7 @@ class Accelerator:
         for i, model in enumerate(self._models):
             if self.distributed_type == DistributedType.FSDP:
                 logger.info("Saving FSDP model")
-                save_fsdp_model(self.state.fsdp_plugin, self, model, output_dir, i)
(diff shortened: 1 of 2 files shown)


muellerzr:
Might need a deprecation cycle for passing in `accelerator`, at the very least probably need to update `transformers`

BenjaminBossan:
> Might need a deprecation cycle for passing in `accelerator`, at the very least probably need to update `transformers`

Yes, probably better to leave the signature but check if accelerator was passed and issue a deprecation warning. A problem here is that the arguments could be passed as positional, so this would require some checking of the arguments to understand if the user passed `accelerator` or not.

pacman100 (approved):
Thank you, @muellerzr, for refactoring this to simplify things! Please note that Trainer code would need to be updated inline with this PR.

PR #2782, opened 2024-05-15
Skip tied weights disk offload test

SunMarc:
# What does this do ?
This PR skips a flaky test. Not sure why it happens but the `data_ptr() `of `compute2.weight` is sometimes the same of `compute1.weight.` Hence it is using the same value as compute1 through self.tied_params_map and we get an assert error. I checked that the values are not the same but it gives the same address. This does not happen on my local machine and it passes sometimes on the CI. Maybe an issue with the hardware. We should have enough coverage with `test_dispatch_model_tied_weights_memory_with_nested_offload_cpu`, so I think it's fine if we skip it. 

cc @fxmarty

### DIFF
--- a/tests/test_big_modeling.py
+++ b/tests/test_big_modeling.py
@@ -487,7 +487,9 @@ class BigModelingTester(unittest.TestCase):
     # This test fails because sometimes data_ptr() of compute2.weight is  the same of compute1.weight.
     # I check that the values are not the same but it gives the same address. This does not happen on my local machine.
     @require_cuda
-    @unittest.skip("Flaky test, we should have enough coverage with test_dispatch_model_tied_weights_memory_with_nested_offload_cpu test")
+    @unittest.skip(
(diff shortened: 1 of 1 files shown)


tests/test_big_modeling.py
```diff
@@ -484,7 +484,12 @@ def forward(self, x):
         del model
         gc.collect()
 
+    # This test fails because sometimes data_ptr() of compute2.weight is  the same of compute1.weight.
+    # I check that the values are not the same but it gives the same address. This does not happen on my local machine.
```
muellerzr, line 488: If we have enough coverage that seems okay, however we can also do a `@flakey` decorator that retries the test some amount of times first (maybe 3?)

SunMarc: I thought about that but I guess it depends more on the hardware. I tried a lot of times to debug and it was failing each time.

muellerzr: Okay, we can just skip then :)

muellerzr (approved):
Thanks!

PR #2790, opened 2024-05-18
Fix Wrong use of sync_gradients used to implement sync_each_batch 

fabianlim:
# What does this PR do?

As pointed  @Nightmare-n [here](https://github.com/huggingface/transformers/issues/29425#issuecomment-2116574203), the fix of [#29425](https://github.com/huggingface/transformers/issues/29425) was incorrect all this while, due to a misunderstanding of how `sync_gradients` is used to signal also [optimizer step](https://github.com/huggingface/accelerate/blob/main/src/accelerate/optimizer.py#L153).

**Tested On mistralai/Mixtral-8x7B-Instruct-v0.1**:
1. `torch =  2.2.0` , regressed changes to `accelerate = 0.28` (because latest `accelerate` does not support old torch
2. `torch = 2.3` with this PR

In both cases we observed the similar effect of the grad accumulation working on this new fix; the results shown below is from case 1.

image

Fixes [#29425](https://github.com/huggingface/transformers/issues/29425), #2531

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr @Nightmare-n

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -996,14 +996,14 @@ class Accelerator:
             model.require_backward_grad_sync = old_require_backward_grad_sync
             model.require_forward_param_sync = old_require_forward_param_sync
 
-    def _do_sync(self, force: bool = False):
+    def _do_sync(self):
         "Sets the right `sync_gradients` context and either resets or increases `self.step`"
         if self.gradient_state.sync_with_dataloader and self.gradient_state.end_of_dataloader:
             self.step = 0
             self.gradient_state._set_sync_gradients(True)
         else:
             self.step += 1
-            self.gradient_state._set_sync_gradients(force or ((self.step % self.gradient_state.num_steps) == 0))
+            self.gradient_state._set_sync_gradients((self.step % self.gradient_state.num_steps) == 0)
 
     @property
     def sync_gradients(self):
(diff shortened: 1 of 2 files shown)


muellerzr:
No worries @fabianlim take your time. Patchfixes go out on Fridays usually so you’ve got time 🤗

muellerzr:
BTW @fabianlim we squash, so you don't need to force-push 🤗

fabianlim:
@muellerzr yes sorry ill stop force-pushing. 

I adjusted the unit tests to test the correct behavior. 
- if `sync_each_batch==False` , then the closeness check by `check_model_parameters` is only at the end of the interval (i.e., `did_step == True`, see below)
- otherwise, then the closeness check must be done at every batch.
- the closness check is against a baseline `model` that implements gradient accum (since it flushes its gradients every grad accum num steps)

However regarding why the `rtol` in the closeness check is required, the following is my (unverified) hypothesis:

If we let A_0, B_0 to be the gradients at batch 0 of replica A and B, and similarly A_1, B_1,
- batch 0: mean reduction means both replicas has gradients, say C_0 = (A_0 + B_0) / 2
- batch 1 (before reduction), replica has C_0 + A_1 and other replica has C _0+ B_1
                 (after reduction),  both replicas has  [(C_0 + A_1) + (C_0 + B_1) ] / 2    -------------- > (1)

whereas a straight gradient accum would have
- batch 0: same as above
- batch 1: both replicas has  C_0 + [A_1 + B_1 ] / 2     -------------> (2)

So the slight differences in the equations (1) and (2), though mathematically equivalent, could have given different numerical results.

```python

with accelerator.accumulate(ddp_model)
    ...

did_step = (((iteration + 1) % 2) == 0) or ((iteration + 1) == len(dataloader))
if accelerator.num_processes > 1:
    check_model_parameters(
        model,
        ddp_model,
        did_step or sync_each_batch,  # syncs at each grad_accum interval of if sync_each_batch==True
        iteration,
        rtol=1e-3,  # needs a relative tolerance due to roundoff errors
    )
if did_step:
     opt.zero_grad()  # flush gradients every accum step
ddp_opt.zero_grad()
```

muellerzr (approved):
Thanks for the improvement! Makes sense to me. Merging

PR #2794, opened 2024-05-21
FIX / FSDP : Guard fsdp utils for earlier PyTorch versions

younesbelkada:
# What does this PR do?

Fixes: https://github.com/huggingface/transformers/issues/30908

As per title - I propose a simple fix as a follow up of https://github.com/huggingface/accelerate/pull/2772 in order to make `accelerate` happy for users with earlier torch versions. Currently when trying to import anything from accelerate using torch < 2.3.0 it fails

```bash
    import torch.distributed.checkpoint.format_utils as dist_cp_format_utils
ModuleNotFoundError: No module named 'torch.distributed.checkpoint.format_utils'
```

cc @muellerzr

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -261,7 +261,7 @@ def merge_fsdp_weights(
     from accelerate.state import PartialState
 
     if not is_torch_version(">=", "2.3.0"):
-        raise ValueError("Make sure to have torch>=2.3.0 installed in order to use `merge_fsdp_weights`")
+        raise ValueError("`merge_fsdp_weights` requires PyTorch >= 2.3.0`")
 
     # To setup `save` to work
     state = PartialState()
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -27,11 +27,13 @@ from .versions import is_torch_version
 
 if is_torch_version(">=", FSDP_PYTORCH_VERSION) and is_torch_distributed_available():
     import torch.distributed.checkpoint as dist_cp
-    import torch.distributed.checkpoint.format_utils as dist_cp_format_utils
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -258,6 +260,9 @@ def merge_fsdp_weights(
     """
     from accelerate.state import PartialState
 
+    if not is_torch_version(">=", "2.3.0"):
+        raise ValueError("Make sure to have torch>=2.3.0 installed in order to use `merge_fsdp_weights`")
```
muellerzr, line 264: Let's be a bit more template-able with what we already have
```suggestion
        raise ValueError("`merge_fsdp_weights` requires PyTorch >= 2.3.0`")
```

muellerzr (approved):
LG2M! Just one small nit.

(Also at some point we should be consistent and change all the `ValueError`'s done with `is_torch_version` to be `ImportErrors` I think)

src/accelerate/utils/fsdp_utils.py
```diff
@@ -27,11 +27,13 @@
 
 if is_torch_version(">=", FSDP_PYTORCH_VERSION) and is_torch_distributed_available():
     import torch.distributed.checkpoint as dist_cp
-    import torch.distributed.checkpoint.format_utils as dist_cp_format_utils
     from torch.distributed.checkpoint.default_planner import DefaultLoadPlanner, DefaultSavePlanner
     from torch.distributed.checkpoint.optimizer import load_sharded_optimizer_state_dict
     from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
     from torch.distributed.fsdp.fully_sharded_data_parallel import StateDictType
+# `dist_cp_format_utils is only available from pt>=2.3.0
+if is_torch_version(">=", "2.3.0"):
```
younesbelkada, line 35:
```suggestion
if is_torch_version(">=", "2.3.0") and is_torch_distributed_available():
```

PR #2806, opened 2024-05-27
Fix access error for torch.mps when using torch==1.13.1 on macOS

SunMarc:
# What does this PR do ?
Supersedes https://github.com/huggingface/accelerate/pull/2686


muellerzr (approved):
Thanks for saving old mps users :)

PR #2619, opened 2024-04-03
State dictionary retrieval from offloaded modules

blbadger:
# What does this PR do?

Adds a function to retrieve a state dictionary from an offloaded module, loading the parameters into a specified device.  

See transformers PR [#27412](https://github.com/huggingface/transformers/pull/27412) for more information

Working on the test too:)

TODO: replace disk offload conditional check with `expand_device_map`

@SunMarc

PR #2820, opened 2024-06-03
add cuda dep for a test

SunMarc:
# What does this PR do ? 

This PR fixes a test that was missing a cuda dependency from this [PR](https://github.com/huggingface/accelerate/pull/2619).


muellerzr (approved):
Thanks! ✅

PR #2825, opened 2024-06-04
xpu: support xpu backend from stock pytorch (>=2.4)

dvrogozh:
Fixes: https://github.com/huggingface/transformers/issues/31237

XPU backend is available in the stock PyTorch starting from version 2.4 [1]. This commit extends huggingface accelerate to support XPU from both IPEX and the stock pytorch. IPEX is being tried first.

Raising this PR as WIP and Draft to facilitate further discussion around XPU backend enabling in huggingface and be able to communicate observed XPU issues back to PyTorch.

[1] https://github.com/pytorch/pytorch/issues/114842

@EikanWang, @fengyuan14, @guangyey, @jgong5, @kding1, @sywangyi

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -475,6 +475,8 @@ class Accelerator:
                 self.scaler = torch.mlu.amp.GradScaler(**kwargs)
             elif is_npu_available():
                 self.scaler = torch.npu.amp.GradScaler(**kwargs)
+            elif is_xpu_available():
+                self.scaler = torch.amp.GradScaler("xpu", **kwargs)
             else:
                 self.scaler = torch.cuda.amp.GradScaler(**kwargs)
 
@@ -1286,9 +1288,9 @@ class Accelerator:
 
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
             if self.device.type == "cpu" and self.state.use_ipex:
-                args = self._prepare_ipex(*args)
+                args = self._prepare_ipex_or_xpu(*args)
             elif self.device.type == "xpu" and is_xpu_available():
-                args = self._prepare_ipex(*args)
+                args = self._prepare_ipex_or_xpu(*args)
         if self.distributed_type == DistributedType.DEEPSPEED:
             result = self._prepare_deepspeed(*args)
         elif self.distributed_type == DistributedType.MEGATRON_LM:
@@ -1896,14 +1898,18 @@ class Accelerator:
 
         return tuple(result)
 
-    def _prepare_ipex(self, *args):
-        if not is_ipex_available():
-            raise ImportError(
-                "IPEX is not installed or IPEX's version does not match current PyTorch version. Please refer"
-                " to https://github.com/intel/intel-extension-for-pytorch."
-            )
-        else:
-            import intel_extension_for_pytorch as ipex
+    def _prepare_ipex_or_xpu(self, *args):
+        """
+        Prepares model and optimizer for training with IPEX or XPU acceleration. This covers 3 cases, IPEX compiled
+        with CPU only support, IPEX compiled with XPU support and training with XPU pytorch backend available in stock
+        pytorch starting from version 2.4.
+        """
+        if self.state.use_ipex:
+            if not is_ipex_available():
+                raise ImportError(
+                    "IPEX is not installed or IPEX's version does not match current PyTorch version. Please refer"
+                    " to https://github.com/intel/intel-extension-for-pytorch."
+                )
 
         model = None
         optimizer = None
@@ -1916,12 +1922,12 @@ class Accelerator:
                 optimizer = obj
         if optimizer is not None and model is not None:
             dtype = torch.bfloat16 if self.state.mixed_precision == "bf16" else None
-            if self.device.type == "xpu" and is_xpu_available():
+            if self.device.type == "xpu":
                 model = model.to(self.device)
-                model, optimizer = torch.xpu.optimize(
-                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
-                )
-            else:
+            # ipex.optimize() is available only for IPEX, both IPEX-CPU and IPEX-XPU
+            if is_ipex_available():
+                import intel_extension_for_pytorch as ipex
+
                 model, optimizer = ipex.optimize(model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1")
         for i in range(len(result)):
             if isinstance(result[i], torch.nn.Module):
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -377,19 +377,23 @@ def is_npu_available(check_device=False):
 
 @lru_cache
 def is_xpu_available(check_device=False):
+    """
+    Checks if XPU acceleration is available either via `intel_extension_for_pytorch` or via stock PyTorch (>=2.4) and
+    potentially if a XPU is in the environment
+    """
+
     "check if user disables it explicitly"
     if not parse_flag_from_env("ACCELERATE_USE_XPU", default=True):
         return False
-    "Checks if `intel_extension_for_pytorch` is installed and potentially if a XPU is in the environment"
-    if is_ipex_available():
-        import torch
 
+    if is_ipex_available():
         if is_torch_version("<=", "1.12"):
             return False
-    else:
-        return False
 
-    import intel_extension_for_pytorch  # noqa: F401
+        import intel_extension_for_pytorch  # noqa: F401
+    else:
+        if is_torch_version("<=", "2.3"):
+            return False
 
     if check_device:
         try:


src/accelerate/utils/imports.py
```diff
@@ -374,16 +374,19 @@ def is_xpu_available(check_device=False):
     "check if user disables it explicitly"
     if not parse_flag_from_env("ACCELERATE_USE_XPU", default=True):
         return False
-    "Checks if `intel_extension_for_pytorch` is installed and potentially if a XPU is in the environment"
+    r"""
```
muellerzr, line 377: No need for the `r` here

src/accelerate/accelerator.py
```diff
@@ -1909,10 +1908,13 @@ def _prepare_ipex(self, *args):
...
-                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
-                )
-            else:
+                # torch.xpu.optimize is available only for xpu via IPEX
+                if hasattr(torch.xpu, "optimize"):
+                    model, optimizer = torch.xpu.optimize(
+                        model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
+                    )
+            elif is_ipex_available():
+                import intel_extension_for_pytorch as ipex
```
muellerzr, line 1917: Question: wouldn't the earlier check when doing `is_xpu_available()` bring in the monkey-patches that this currently does?

muellerzr (commented):
Thanks a bunch for doing this! Great start, just one question 🤗

src/accelerate/accelerator.py
dvrogozh: I think answer is - yes. Good that you've asked since I believe there is a space for improvements here. So, at the moment I tried to preserve previous logic and saved import (it was at line 1897 before) and a call to `ipex.optimize()`. However, per my understanding import of ipex which is already being done on `is_xpu_available()` should register IPEX for PyTorch and expose a set of operations such as `torch.xpu.optimize()`. Thus, I think that the "else" path with a call to `ipex.optimize()` might never be executed. This makes me think that this is some unspotted leftover. I suggest to consider to completely drop the else path.

muellerzr: Agreed :)

jgong5: So, if it is a dead code, why not removing it?

dvrogozh: I am going to remove it in the next iteration. Working on it at the moment.

src/accelerate/accelerator.py
```diff
@@ -1284,9 +1285,9 @@ def prepare(self, *args, device_placement=None):
 
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
             if self.device.type == "cpu" and self.state.use_ipex:
-                args = self._prepare_ipex(*args)
+                args = self._prepare_xpu(*args)
```
guangyey, line 1288: `_prepare_ipex` can import ipex, but `_prepare_xpu` could not.

src/accelerate/utils/imports.py
dvrogozh: removed

src/accelerate/utils/imports.py
```diff
@@ -374,16 +374,22 @@ def is_xpu_available(check_device=False):
     "check if user disables it explicitly"
     if not parse_flag_from_env("ACCELERATE_USE_XPU", default=True):
         return False
-    "Checks if `intel_extension_for_pytorch` is installed and potentially if a XPU is in the environment"
-    if is_ipex_available():
-        import torch
+    """
+    Checks if XPU acceleration is available either via `intel_extension_for_pytorch` or
+    via stock PyTorch (>=2.4) and potentially if a XPU is in the environment
+    """
```
muellerzr, line 380: This docstring should be on L 374

muellerzr (approved):
Thanks! Looking much better. Just a nit

src/accelerate/accelerator.py
dvrogozh: done

src/accelerate/accelerator.py
```diff
@@ -1907,13 +1905,13 @@ def _prepare_ipex(self, *args):
...
+            if self.device.type == "xpu":
                 model = model.to(self.device)
-                model, optimizer = torch.xpu.optimize(
-                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
-                )
-            else:
-                model, optimizer = ipex.optimize(model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1")
+                # torch.xpu.optimize is available only for xpu via IPEX
+                if hasattr(torch.xpu, "optimize"):
+                    model, optimizer = torch.xpu.optimize(
```
dvrogozh, line 1912: @muellerzr : I think I figured it out why there are 2 different calls (`torch.xpu.optimize` and `ipex.optimize`) in the current code. It occurs that IPEX can be built in 2 distinct ways:
1. The first way is to build it with Intel GPUs support, i.e. IPEX-XPU. If it's built this way, we get monkey-patches applied and `torch.xpu.optimize()` is exposed which is a handy version of `ipex.optimize()` which is also available.
2. The second way to built IPEX is with CPU support. In this case there is no any GPUs support, no monkey-patches available and no XPU is available.  In this case the only thing which is available is `ipex.optimize()`

So, it seems that current Huggingface accelerate code is supporting both paths, IPEX-XPU and IPEX-CPU. That's if judging from the source code. Interesting to note that the https://github.com/huggingface/accelerate/blob/main/docs/source/usage_guides/ipex.md talks only about IPEX-CPU and does not mention IPEX-XPU... I wonder whether IPEX-XPU is fully enabled in Huggingface?

**Summarizing, I think I need to update my PR covering all 3 cases: 1) IPEX-XPU, 2) IPEX-CPU, 3) XPU in stock PyTorch. I will try to rewrite in a way that it will be clear which options are there on the plate, will add some comments to the code.**

FYI The easiest way to check behavior would be probably trying out IPEX containers from https://hub.docker.com/r/intel/intel-optimized-pytorch. Here are some printouts:
```
# IPEX CPU
$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.3.0-pip-base python3 -c 'import torch; import intel_extension_for_pytorch; print(torch.xpu.is_available())'
False

$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.3.0-pip-base python3 -c 'import torch; import intel_extension_for_pytorch as ipex; print(hasattr(ipex, "optimize"))'
True

$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.3.0-pip-base python3 -c 'import torch; import intel_extension_for_pytorch as ipex; print(hasattr(torch.xpu, "optimize"))'
False

# IPEX XPU
$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.1.30-xpu python3 -c 'import torch; import intel_extension_for_pytorch; print(torch.xpu.is_available())'
True

$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.1.30-xpu python3 -c 'import torch; import intel_extension_for_pytorch as ipex; print(hasattr(ipex, "optimize"))'
True

$ docker run -it --rm --privileged intel/intel-extension-for-pytorch:2.1.30-xpu python3 -c 'import torch; import intel_extension_for_pytorch as ipex; print(hasattr(torch.xpu, "optimize"))'
True
```

dvrogozh: @muellerzr : I reworked the PR according to above. Please, help to review again.

dvrogozh:
I tried this PR (+ https://github.com/huggingface/transformers/pull/31238) as much as I could in the IPEX-CPU, IPEX-XPU, Pytorch-XPU, Pytorch-CPU scenarios. Tried to run some tests from accelerate and transformers and some examples from transformers. All seem to work engaging with XPU when expected. I promote these PRs from drafts for the qualified review. Let me know if any concerns or any feedback needs to be addressed.

dvrogozh:
Applied `doc-builder style src/accelerate docs/source --max_len 119` to fix format issues identified by ci.

EikanWang (approved):

dvrogozh:
@muellerzr : can you, please, help to run ci again? Also, is there anything else I can help with fixing in this PR to get it merged?

dvrogozh:
I did not see such failure before on this PR. Can this be something random since I can't associate this failure with the changes made:
```
FAILED tests/test_accelerator.py::AcceleratorTester::test_save_load_model_with_hooks_use_pytorch - assert 0.0007739067077636719 > 0.001
 +  where 0.0007739067077636719 = abs((4.019573211669922 - 4.0203471183776855))
 +    where 4.0203471183776855 = get_signature(Linear(in_features=2, out_features=4, bias=True))
```

dvrogozh:
@SunMarc : thank you for retriggering failed ci. I see it's passing now. I guess my assumption that this was sporadic failure is true.

@SunMarc, @muellerzr : I have outlined current status of xpu backend in pytorch in https://github.com/huggingface/transformers/issues/31237. There are a number of issues in xpu backend which are being worked on right now. I believe however that this PR and PR in transformers (https://github.com/huggingface/transformers/issues/31237) are ready as the first step to enable xpu backend in huggingface on top of which we can gradually improve the support. Can you, please, outline acceptance requirements for these PRs on Huggingface side?

src/accelerate/utils/imports.py
```diff
@@ -377,19 +377,28 @@ def is_npu_available(check_device=False):
...
     if not parse_flag_from_env("ACCELERATE_USE_XPU", default=True):
         return False
-    "Checks if `intel_extension_for_pytorch` is installed and potentially if a XPU is in the environment"
-    if is_ipex_available():
-        import torch
 
+    if importlib.util.find_spec("torch") is None:
+        return False
+
+    import torch
```
muellerzr, line 392: This part can actually be removed, as accelerate always requires PyTorch :)

muellerzr (approved):
Thanks! This looks great to me, thank you for the improvement! 

cc @SunMarc for a second pair of eyes, else we can merge it after the nit has been addressed!

src/accelerate/utils/imports.py
dvrogozh: Indeed:
https://github.com/huggingface/accelerate/blob/91a2599f933fa4ee7eda89e6b4f405d897a30367/src/accelerate/utils/imports.py#L21

I copied this from `is_npu_available` above. Which by the way also re-imports `torch`. Do you want me to also fix `is_npu_available()` in this PR?
https://github.com/huggingface/accelerate/blob/91a2599f933fa4ee7eda89e6b4f405d897a30367/src/accelerate/utils/imports.py#L362-L367

src/accelerate/accelerator.py
```diff
@@ -1916,12 +1922,12 @@ def _prepare_ipex(self, *args):
...
             dtype = torch.bfloat16 if self.state.mixed_precision == "bf16" else None
-            if self.device.type == "xpu" and is_xpu_available():
+            if self.device.type == "xpu":
                 model = model.to(self.device)
-                model, optimizer = torch.xpu.optimize(
-                    model, optimizer=optimizer, dtype=dtype, inplace=True, level="O1"
-                )
-            else:
+            # ipex.optimize() is available only for IPEX, both IPEX-CPU and IPEX-XPU
+            if is_ipex_available():
```
SunMarc, line 1928: Maybe change that by `self.state.use_ipex` (or add at least)?

SunMarc (approved):
LGTM ! Just a small nit

src/accelerate/accelerator.py
dvrogozh: I afraid this might break IPEX-XPU path.

Note that `use_ipex` is currently used with CPU path to differentiate the case when IPEX-CPU optimization should or should not be used:
https://github.com/huggingface/accelerate/blob/c0faec766c02973edb5b75b97fee5fe4ee1a7c69/src/accelerate/accelerator.py#L1288
In case of IPEX-XPU this flag was not used and I am not sure whether it will be `=True`.

SunMarc: Sound good !

src/accelerate/utils/imports.py
dvrogozh: > This part can actually be removed, as accelerate always requires PyTorch :)

Fixed. I will submit npu/mlu in a separate cleanup PR unless you will tell me otherwise.

muellerzr (approved):
Thanks for doing this! Next step: transformers :) 

(Also if you want the version this will release on, `accelerate==0.32.0`)

src/accelerate/utils/imports.py
dvrogozh: > I will submit npu/mlu in a separate cleanup PR unless you will tell me otherwise.

Submitted #2856.

PR #2826, opened 2024-06-05
Remove out-dated xpu device check code in `get_balanced_memory`

faaany:
## What does this PR do?
This PR removes the outdated and unnecessary code of xpu device check in the following code snippet
```python
if (
                    d != "cpu"
                    and (torch.device(d).type == "xpu" or torch.xpu.get_device_properties(d).dev_type == "gpu")
                )
```
The reasons are 
- If torch.xpu.get_device_properties(d) works, torch.device(d).type will always be "xpu"
- If torch.device(d).type == "xpu", `d` will definitively not be "cpu"

`torch.device(d).type == "xpu"` is enough to check xpu device just like the case in MLU and  NPU.

## Who can review?
@SunMarc and @muellerzr

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -957,13 +957,14 @@ def get_balanced_memory(
     max_memory = get_max_memory(max_memory)
 
     if is_npu_available():
-        num_devices = len([d for d in max_memory if torch.device(d).type == "npu" and max_memory[d] > 0])
+        expected_device_type = "npu"
     elif is_mlu_available():
-        num_devices = len([d for d in max_memory if torch.device(d).type == "mlu" and max_memory[d] > 0])
+        expected_device_type = "mlu"
     elif is_xpu_available():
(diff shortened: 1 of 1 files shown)


faaany:
@abhilash1910

muellerzr (commented):
Thanks! Love the simplification 😍 

Could we actually take this a bit further now that the logic is aligned?

Perhaps getting the device type expected under the `is_xxx_available()` and then only calling the 
```python
num_devices = len([d for d in max_memory if torch.device(d).type == expected_device_type and max_memory[d] > 0])
```

section once? WDYT

abhilash1910:
Yes @muellerzr  looks like a good abstraction, We are internally validating the "gpu" predicate of the validation logic . 
If it is no longer required then new design LGTM.

SunMarc (approved):
LGTM !

faaany:
@muellerzr PR updated with `expected_device_type`, pls review, thx!

abhilash1910 (approved):
LGTM

muellerzr (approved):
Beautiful 😍 

Nice work!

PR #2842, opened 2024-06-10
fix(ci): remove unnecessary permissions

McPatate:
### What does this PR do?

Remove unnecessary permissions granted to the actions workflow.

Sorry for the mishap.


muellerzr:
Thanks for the follow up!

muellerzr (approved):

Issue #2840, opened 2024-06-08
Big Models, move model to CPU after dispatching to multiple devices


balaabhijit:
### System Info

```Shell
- `Accelerate` version: 0.27.2
- Platform: Linux-5.15.0-76-generic-x86_64-with-glibc2.35
- Python version: 3.10.13
- Numpy version: 1.26.3
- PyTorch version (GPU?): 2.2.0+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 503.74 GB
- GPU type: NVIDIA A100 80GB PCIe
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

This PR #1790 seems to add a warning when trying to move a model which is dispatched on multiple devices. Is there a way to move the model back to CPU once we dispatch to multiple devices? I even tried saving the model, deleting and reloading the model but, deleting the model in python with `del model` (Even with GC and cuda cache clear) doesn't seem to clear GPU memory. Is there anyway to achieve this?

### Expected behavior

Move model to CPU after dispatching to multiple GPUs

SunMarc:
Hi @balaabhijit, thanks for reporting ! What would be the use case for this ? Are you modifying the model ? TO answer your question, you can use remove the hooks by calling `remove_hook_from_module(model, recurse=True)`, then you should be able to move the model to your chosen device. As for the warning, I will create a PR, so that we remove it when calling `remove_hook_from_module` !  LMK if this works !

balaabhijit:
@SunMarc Thanks for your prompt reply. This works!

The use case is a very specific case where we are trying to  quantize the model as part of a pipeline. The previous step requires the model to be distributed and the quantization step (AWQ) happens layer by layer so as to save GPU memory.

SunMarc:
Sounds good ! I'm closing this issue since this is solved !

PR #2843, opened 2024-06-10
remove warning hook addede during dispatch_model

SunMarc:
# What does this PR do?
This PR updates the `remove_hook_from_module`  function to also remove the warning hooks that were added during the dispatch function. I've added a small test to see if we indeed do not get the warning anymore 
Fixes https://github.com/huggingface/accelerate/issues/2840


src/accelerate/hooks.py
```diff
@@ -202,6 +202,12 @@ def remove_hook_from_module(module: nn.Module, recurse=False):
             module.forward = module._old_forward
         delattr(module, "_old_forward")
 
+    # Remove accelerate added warning hooks
+    module.__dict__.pop('to', None)
+    module.__dict__.pop('cuda', None)
+    module.__dict__.pop('xpu', None)
+    module.__dict__.pop('npu', None)
```
muellerzr, line 209: Proposal: what if we saved these all as `_accelerate_added_attributes` and then this way we can just pop the lot of them without needing to worry about keeping everything aligned in the future

muellerzr (commented):
Thanks for doing this! This should also fix that pickling issue. Just one suggestion

src/accelerate/hooks.py
BenjaminBossan: I agree. This also makes it easier to find where these attributes come from, as a reader can search for `_accelerate_added_attributes` and quickly find the origin.

SunMarc: Make sense !

muellerzr (approved):
Nice! :)

src/accelerate/hooks.py
```diff
@@ -30,6 +30,9 @@
 from .utils.other import recursive_getattr
 
 
+_accelerate_added_attributes = ["to", "cuda", "npu", "xpu", "mlu"]
```
BenjaminBossan, line 33: I would have hoped we could reuse this constant when the hooks are added (I guess [here](https://github.com/huggingface/accelerate/blob/f62854a28167a8775e8808669e0ec5f4a1f28537/src/accelerate/big_modeling.py#L460-L468)) but that won't quite work because of the additional checks needed there. But we could still add a comment there that if a new warning hooks is added, `_accelerate_added_attributes` should be updated.

SunMarc: Good idea ! Thanks !

PR #2845, opened 2024-06-11
Speed up imports and add a CI

muellerzr:
# What does this PR do?

This PR introduces a CI which utilizes the `cProfile` timings and my `tuna-interpreter` libraries to perform time-based import tests.

It was reported by @stas00 that we were taking *far* too long to do basic things like `accelerate launch`, and `tuna` can help visualize why by creating import graphs directing us to what is taking too long:

image

I wrote a small library called [tuna-interpreter](https://github.com/muellerzr/tuna-interpreter) that aims to take the best parts of `tuna` and work it into something parse-able that lets us run CIs off of it. 

After using the tool:

image

**We can see a decrease of ~68%**

## How it works:

In its current form, we are going based off of a baseline `torch` import, since Accelerate relies on torch no matter what. **BUT** we should be no more than ~20% slower than the torch import overall. Anything more and we have some slip-up or timing problem. 

An example test looks like the following:
```python
    def test_base_import(self):
        output = run_import_time("import accelerate")
        with open(f"{self.tmpdir}/base_results.log", "w") as f:
            f.write(output)
        data = read_import_profile(f"{self.tmpdir}/base_results.log")
        total_time = calculate_total_time(data)
        pct_more = total_time / self.pytorch_time
        # Base import should never be more than 10% slower than raw torch import
        err_msg = f"Base import is more than 20% slower than raw torch import ({pct_more * 100:.2f}%), please check the attached `tuna` profile:\n"
        sorted_data = sort_nodes_by_total_time(data)
        paths_above_threshold = get_paths_above_threshold(sorted_data, 0.1, max_depth=7)
        err_msg += f"\n{convert_list_to_string(paths_above_threshold)}"
        self.assertLess(pct_more, 1.2, err_msg)
```

Where essentially we:
1. Get the `import_time` to run a particular python import called via `-c` in subprocess
2. We then read this profile generated
3. From here, we take all the nodes, sort them by time, and get all paths above an arbitrary threshold. *This should be tweaked to your own discretion, as threshold and max_depth changes from library to library*. The key with `max_depth` is it should be *enough* to get your imports out of the slowdown trace, and show what external libraries you are really calling.
4. Afterwards, we write a note stating that it was above a slowdown expected %, and state what modules were slowing it down.

An example failure is below, where we can clearly see what module chain had the slowdown:
```
E       AssertionError: 1.3515017627366224 not less than 1.2 : Base import is more than 20% slower than raw torch import (135.15%), please check the attached `tuna` profile:
E       
E       main 0.973s
E       main->accelerate 0.961s
E       main->accelerate->accelerate.accelerator 0.959s
E       main->accelerate->accelerate.accelerator->torch 0.758s
E       main->accelerate->accelerate.accelerator->torch->torch._C 0.355s
E       main->accelerate->accelerate.accelerator->torch->torch._meta_registrations 0.154s
E       main->accelerate->accelerate.accelerator->torch->torch._meta_registrations->torch._decomp 0.109s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing 0.126s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils 0.125s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils->accelerate.utils.megatron_lm 0.107s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils->accelerate.utils.megatron_lm->transformers.modeling_outputs 0.106s

tests/test_imports.py:64: AssertionError
```

Or:
```
E       AssertionError: 1.8292819779293377 not less than 1.2 : Base import is more than 20% slower than raw torch import (182.93%), please check the attached `tuna` profile:
E       
E       main 1.324s
E       main->accelerate 1.308s
E       main->accelerate->accelerate.accelerator 1.307s
E       main->accelerate->accelerate.accelerator->torch 0.706s
E       main->accelerate->accelerate.accelerator->torch->torch._C 0.327s
E       main->accelerate->accelerate.accelerator->torch->torch._meta_registrations 0.152s
E       main->accelerate->accelerate.accelerator->torch->torch._meta_registrations->torch._decomp 0.108s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing 0.527s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils 0.526s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils->accelerate.utils.fsdp_utils 0.488s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils->accelerate.utils.fsdp_utils->torch.distributed.fsdp.fully_sharded_data_parallel 0.488s
E       main->accelerate->accelerate.accelerator->accelerate.checkpointing->accelerate.utils->accelerate.utils.fsdp_utils->torch.distributed.fsdp.fully_sharded_data_parallel->torch.distributed.fsdp 0.488s

tests/test_imports.py:64: AssertionError
```

If there are specific issues with using `tuna-interpreter`, please let me know, it's a very quickly hacked-together-but-working library for what we are doing, and open to improving it further after we battle-test it

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @BenjaminBossan @sayakpaul @Titus-von-Koeller @ArthurZucker @ydshieh @LysandreJik


muellerzr:
Let me know how I can improve on this tool further so we can then get it going throughout anyone at HF that wants to use it 🤗

tests/test_imports.py
```diff
@@ -0,0 +1,79 @@
...
+
+@require_tuna
+class ImportSpeedTester(TempDirTestCase):
+    clear_on_setup = False
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        output = run_import_time("import torch")
+        with open(f"{cls.tmpdir}/pytorch_results.log", "w") as f:
```
BenjaminBossan, line 47: This probably won't run on Windows, right? Not sure if it really matters though.

tests/test_imports.py
```diff
@@ -0,0 +1,79 @@
...
+            f.write(output)
+        data = read_import_profile(f"{self.tmpdir}/base_results.log")
+        total_time = calculate_total_time(data)
+        pct_more = total_time / self.pytorch_time
+        # Base import should never be more than 10% slower than raw torch import
+        err_msg = f"Base import is more than 20% slower than raw torch import ({pct_more * 100:.2f}%), please check the attached `tuna` profile:\n"
+        sorted_data = sort_nodes_by_total_time(data)
+        paths_above_threshold = get_paths_above_threshold(sorted_data, 0.1, max_depth=7)
+        err_msg += f"\n{convert_list_to_string(paths_above_threshold)}"
+        self.assertLess(pct_more, 1.2, err_msg)
```
BenjaminBossan, line 65: So IIUC, we want to ensure that the import doesn't add more than 20% on top of the PyTorch import. Did you check how robust this is? Should this test perhaps do multiple imports and then average? If so, should we throw away the first timing, as it may require a warm up (disk cache?) and thus make the first import unduly slow (and thus make the test pass trivially)?

src/accelerate/utils/__init__.py
```diff
@@ -192,24 +193,27 @@
...
-    gather_across_data_parallel_groups,
-)
-from .megatron_lm import initialize as megatron_lm_initialize
-from .megatron_lm import prepare_data_loader as megatron_lm_prepare_data_loader
-from .megatron_lm import prepare_model_optimizer_scheduler as megatron_lm_prepare_model_optimizer_scheduler
-from .megatron_lm import prepare_optimizer as megatron_lm_prepare_optimizer
-from .megatron_lm import prepare_scheduler as megatron_lm_prepare_scheduler
+
+
+if is_megatron_lm_available():
```
BenjaminBossan, line 198: For my understanding, this change is unrelated, right?

BenjaminBossan (commented):
Thanks for adding this workflow to detect regressions in import time.

I have a a couple of questions and comments, please check.

Regarding tuna, I did a quick check and from my understanding, this is mostly a package to lightly process the profile data generated by Python's cProfile module and to add visualization similar to snakeviz. Also note that the last commit is already almost 1.5 years old and that `read_import_profile` comes from a private module. For this reason, I wonder if it wouldn't make more sense to vendor this function with your lib and get rid of the tuna dependency completely.

src/accelerate/utils/__init__.py
BenjaminBossan: Ah I missed that, had to scroll right to see the megatron part :D So it's not related to implementing the new workflow, but it is related to actually increasing the startup time.

tests/test_imports.py
muellerzr: Correct. 

We specifically care about the warmup, because otherwise it will lead to slowdowns via `accelerate launch`. 

I'll check multiple calls, but it shouldn't because everything is run via `subprocess` so it shouldn't rely on disk caching

tests/test_imports.py
muellerzr: Probably not, no. I don't believe toma supports windows

muellerzr:
@BenjaminBossan what's missing here is the initial tuna check, which is why it's required (see the workflow). 

We can eventually look at gutting it out, sure. However I think having both the visual option for debugging further + the condensed output here is valuable and is why we should still use `tuna` itself.

tests/test_imports.py
muellerzr: `cls.tmpdir` is a `Path` object, so will fix

muellerzr:
@BenjaminBossan fully removed the requirement for `tuna` there, however I still think it's useful for further debugging so left that in as part of the testcase class description :)

src/accelerate/utils/__init__.py
stas00: the check name is somewhat misleading, since a user may have megatron installed (i.e. available) but they aren't telling accelerate to use it.

perhaps `is_megatron_needed`?

for `available` - shouldn't it fail anyway long before if it's not available?

what I'm trying to communicate is for the sake of imports. `needed` and `available` are orthogonal to each other.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -58,8 +46,12 @@ def _set_model_state_dict(model, state_dict, adapter_only=False):
 
 
 def save_fsdp_model(fsdp_plugin, accelerator, model, output_dir, model_index=0, adapter_only=False):
-    os.makedirs(output_dir, exist_ok=True)
+    import torch.distributed.checkpoint as dist_cp
+    from torch.distributed.checkpoint.default_planner import DefaultSavePlanner
+    from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
+    from torch.distributed.fsdp.fully_sharded_data_parallel import StateDictType
```
stas00, line 52: this feels somewhat yucky - since it's going to run the imports many times over - in particular because you replicate this exact import code multiple times below.

Why not move these imports into its own file and call them imports only if they haven't been loaded yet?

regardless it should be safe to assume that if a user configured accelerate to use FSDP it'll need all these save/load calls - so from the performance point of view all of them can be moved into a single place and loaded once - if the user asked for FSDP.

muellerzr: Will need to think of an ideal folder structure, but I'll start with some small version of this in this PR, since we're touching those imports now (specifically sectioning out to a different file)

muellerzr: Also the difference with megatron is we check for a package being installed. So while we can do similar for deepspeed, we cannot do this for fsdp (as it doesn't actually check for configurations, only if a package is available in the env)

src/accelerate/utils/__init__.py
muellerzr: No, it is always `available` and has been `available`, else we'd call it `needed` or `configured` as you say. 

`requires_` is for tests, we know we are using it so thus we use it.

`_available` simply means bring it into the environment if available.

muellerzr: Its aggressive importing, yes. Which as noted in my last comment, will aim to fix when I can dedicate some time to refactoring our entire import schema in accelerate

stas00: but the first thing the `is_megatron_available` check does is whether it's configured to be used, or am I misreading it?

https://github.com/huggingface/accelerate/blob/67b294398f4e3ab40a57a767f935a49a74e7f5d7/src/accelerate/utils/imports.py#L224

so it's not available - it's `is_wanted_and_available`

src/accelerate/utils/fsdp_utils.py
stas00: but it should be enough to check whether FSDP is being configured to be used - configured? import, not configured? don't import

stas00: the dependency chain could be:

this_component_is_needed_to_continue -> do we have it installed -> if so import the needed bits.

i.e. the logic that decides to whether import something or not is driven by the configuration of accelerate and not by availability of a 3rd party package.

BenjaminBossan: Hmm, but imports are cached in Python, so the 2nd time this is called, they don't add any overhead. Or am I missing something?

stas00: sure, but this just feels weird. 

imports don't belong to methods/functions and surely not when you repeat the same imports multiple times in multiple functions. Nothing stops you from doing that, of course.

imports are sort of a singleton thing and certainly we occasionally see an odd import in some function, but it's usually an exception and this happens more often in a throw-away code.

BenjaminBossan (approved):
Overall this LGTM, thanks for working on this and the tuna-interpreter package.

I have some comments, but none are blockers. Regarding the local imports, personally I don't think it's a huge anti-pattern, so this can be left for later IMO.

One question I have is how `test_imports.py` is supposed to be run, just manually by devs or some automatic job? If the latter, I'd consider running the checks multiple times and averaging the results (perhaps also discarding outliers) to make the job more robust. If it's manual, then that's not necessary.

src/accelerate/test_utils/testing.py
```diff
@@ -377,6 +378,14 @@ def require_pippy(test_case):
     return unittest.skipUnless(is_pippy_available(), "test requires pippy")(test_case)
 
 
+def require_tuna(test_case):
```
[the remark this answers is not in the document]
SunMarc: +1

SunMarc (approved):
Thanks for making the import faster !

sayakpaul (approved):
This is a good one! Good news for the `diffusers` team is that we already follow this mechanism :) 

But we could definitely include a CI for this. @yiyixuxu @DN6 WDYT?

ydshieh:
🔥

PR #2857, opened 2024-06-14
Fix get_backend bug and add clear_device_cache function

NurmaU:
# What does this PR do?

This PR introduces 2 updates:
1. adds new `clear_device_cache` function in memory.py to replace repetitive code. This also adds check for mps devices in  function `find_executable_batch_size`. This feature was tested with tests/test_memory_utils.py, which had bug.
2. fixes bug in `get_backend` function where function's 3rd outputs must be callable object. This bug caused failures in tests/test_memory_utils.py on MacOS.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -66,7 +66,7 @@ def get_backend():
     elif is_mps_available(min_version="2.0"):
         return "mps", 1, torch.mps.current_allocated_memory
     elif is_mps_available():
-        return "mps", 1, 0
+        return "mps", 1, lambda: 0
     elif is_mlu_available():
         return "mlu", torch.mlu.device_count(), torch.mlu.memory_allocated
     elif is_npu_available():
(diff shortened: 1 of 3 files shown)


src/accelerate/test_utils/testing.py
```diff
@@ -64,7 +64,7 @@ def get_backend():
     elif is_cuda_available():
         return "cuda", torch.cuda.device_count(), torch.cuda.memory_allocated
     elif is_mps_available(min_version="2.0"):
-        return "mps", 1, torch.mps.current_allocated_memory()
+        return "mps", 1, torch.mps.current_allocated_memory
```
BenjaminBossan, line 67: Nice catch. But doesn't this mean that the return values below `"mps", 1, 0` (for old mps) and `"cpu", 1, 0` (for CPU) are also broken? Should they return `lambda: 0` instead?

muellerzr: Agreed, we should use a `lambda` for those

NurmaU: Yes, you're right. Added `lambda: 0`. Thanks for pointing that out!

muellerzr (approved):
Thanks for the refactor, and sorry for the delay in getting this in!

PR #2343, opened 2024-01-16
device agnostic test_accelerator/test_multigpu

wangshuai09:
# What does this PR do?

Part of #2122 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

@muellerzr

PR #2860, opened 2024-06-17
[tests] skip bnb-related tests instead of failing on xpu

faaany:
## What does this PR do?
When running `pytest tests/test_accelerator.py` on XPU, there are 4 tests failed:
```bash
FAILED tests/test_accelerator.py::AcceleratorTester::test_accelerator_bnb - RuntimeError: No GPU found. A GPU is needed for quantization.
FAILED tests/test_accelerator.py::AcceleratorTester::test_accelerator_bnb_cpu_error - RuntimeError: No GPU found. A GPU is needed for quantization.
FAILED tests/test_accelerator.py::AcceleratorTester::test_accelerator_bnb_multi_device - ValueError: xpu is not supported in test_accelerator_bnb_multi_device.
FAILED tests/test_accelerator.py::AcceleratorTester::test_accelerator_bnb_multi_device_no_distributed - RuntimeError: No GPU found. A GPU is needed for quantization.
```
The reason for all 4 tests is the same: all tests use `load_in_8bit=True` when loading the model with `AutoModelForCausalLM.from_pretrained`, which will lead to the `torch.cuda.is_available()` check in `hf_quantizer.validate_environment`, e.g. 

```bash
___________________________________________ AcceleratorTester.test_accelerator_bnb ____________________________________________

self = <test_accelerator.AcceleratorTester testMethod=test_accelerator_bnb>

    @require_non_torch_xla
    @slow
    @require_bnb
    def test_accelerator_bnb(self):
        """Tests that the accelerator can be used with the BNB library."""
    
>       model = AutoModelForCausalLM.from_pretrained(
            "EleutherAI/gpt-neo-125m",
            load_in_8bit=True,
            device_map={"": 0},
        )

tests/test_accelerator.py:439: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../miniforge3/envs/acc-ut-ww23/lib/python3.9/site-packages/transformers/models/auto/auto_factory.py:563: in from_pretrained
    return model_class.from_pretrained(
../../miniforge3/envs/acc-ut-ww23/lib/python3.9/site-packages/transformers/modeling_utils.py:3202: in from_pretrained
    hf_quantizer.validate_environment(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <transformers.quantizers.quantizer_bnb_8bit.Bnb8BitHfQuantizer object at 0x7f2cef87ceb0>, args = ()
kwargs = {'device_map': {'': 0}, 'from_flax': False, 'from_tf': False, 'torch_dtype': None}

    def validate_environment(self, *args, **kwargs):
        if not torch.cuda.is_available():
>           raise RuntimeError("No GPU found. A GPU is needed for quantization.")
E           RuntimeError: No GPU found. A GPU is needed for quantization.

../../miniforge3/envs/acc-ut-ww23/lib/python3.9/site-packages/transformers/quantizers/quantizer_bnb_8bit.py:62: RuntimeError
```
Since BNB doesn't support XPU and NPU yet, we should skip these tests rather than let them fail. I also checked the original PR #2343, which makes those tests device-agnostic. These 4 tests are overlooked because the author doesn't install bitsandbytes. But when users install the libraries for testing using `pip install -e ".[testing]"`, the bitsandbytes library will be installed. So we should change the test maker to `require_cuda`, instead of making them device-agnostic, at least before the time when XPU and NPU are supported in BNB. 

 @muellerzr @SunMarc


faaany:
add one more test: `tests/test_big_modeling.py::BigModelingTester::test_dipatch_model_fp4_simple`

muellerzr (approved):
Thanks! cc @SunMarc

tests/test_accelerator.py
```diff
@@ -472,21 +472,13 @@ def test_accelerator_bnb_cpu_error(self):
         with self.assertRaises(ValueError):
             model = accelerator.prepare(model)
 
-    @require_non_torch_xla
+    @require_cuda
```
SunMarc, line 475: We should put require_multi_gpu-gpu no ?

tests/test_accelerator.py
```diff
@@ -472,21 +472,13 @@ def test_accelerator_bnb_cpu_error(self):
...
     @slow
     @require_bnb
-    @require_multi_device
     def test_accelerator_bnb_multi_device(self):
         """Tests that the accelerator can be used with the BNB library."""
         from transformers import AutoModelForCausalLM
 
-        if torch_device == "cuda":
-            PartialState._shared_state = {"distributed_type": DistributedType.MULTI_GPU}
```
SunMarc, line 484: Don't remove this part, so that it is just like how it was [before](https://github.com/huggingface/accelerate/pull/2343/files#:~:text=transformers%20import%20AutoModelForCausalLM-,PartialState._shared_state%20%3D%20%7B%22distributed_type%22%3A%20DistributedType.MULTI_GPU%7D,-if%20torch_device%20%3D%3D)

SunMarc (approved):
LGTM ! Left a few comments

tests/test_accelerator.py
faaany: good catch! I thought `require_cuda` could replace `requre_non_torch_xla`. I will keep `require_non_torch_xla`. Code updated, thx!

tests/test_accelerator.py
faaany: good catch! I thought require_cuda could replace requre_non_torch_xla. I will keep require_non_torch_xla. Code updated, thx!

muellerzr (approved):
Thanks!

PR #2552, opened 2024-03-13
Add Cambricon MLU accelerator support

huismiling:
# What does this PR do?

If I want to use Cambricon MLUs to train 🤗 Transformers models, the support should be added in Accelerate first and then will come in the Trainer for free.
This PR will support Cambricon MLU accelerator：

1. Sample config after running the accelerate config command:
```compute_environment: LOCAL_MACHINE
debug: false
distributed_type: MULTI_MLU
downcast_bf16: 'no'
gpu_ids: all
machine_rank: 0
main_training_function: main
mixed_precision: 'no'
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

2. run nlp_example.py with MLUs.
`accelerate launch  nlp_example.py`
Below are the output logs:
```Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
Some weights of BertForSequenceClassification were not initialized from the model checkpoint at bert-base-cased and are newly initialized: ['classifier.bias', 'classifier.weight']
[...]

PR #2874, opened 2024-06-20
fix bug when getting the real accelerator's device number

faaany:
## What does this PR do?
This PR is a follow-up fix for PR #2826  and I want to correct my statement in that PR that torch.device(d).type == "xpu" is enough to check the xpu device just like the case in npu and mlu. This was my mistake. In fact, `torch.device(0).type` will always return "cuda" on XPU as can be seen from the [pytorch code](https://github.com/pytorch/pytorch/blob/main/c10/core/Device.cpp#L45) and from the [pytorch offical doc](https://pytorch.org/docs/stable/tensor_attributes.html#torch-device) at least for now. But we are working on a PR to support it in the future pytorch version. Also for NPU path, I think torch.device(0).type` will return `cuda` as can be seen [here](https://github.com/Ascend/pytorch/issues/16).

In addition, users might pass device id that exceeds the available device count. For this case, we will not count that incorrect id to `num_devices` when calculating the balanced memory. So this PR actually fixes 2 issues:
- `num_devices` for non-cuda devices will always be 0
- `num_devices` will include device index that is larger than the available device number

## Who can review?
@SunMarc and @muellerzr

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -990,14 +990,14 @@ def get_balanced_memory(
     max_memory = get_max_memory(max_memory)
 
     if is_npu_available():
-        expected_device_type = "npu"
+        expected_device_count = torch.npu.device_count()
     elif is_mlu_available():
-        expected_device_type = "mlu"
+        expected_device_count = torch.mlu.device_count()
     elif is_xpu_available():
-        expected_device_type = "xpu"
+        expected_device_count = torch.xpu.device_count()
     else:
-        expected_device_type = "cuda"
-    num_devices = len([d for d in max_memory if torch.device(d).type == expected_device_type and max_memory[d] > 0])
+        expected_device_count = torch.cuda.device_count()
+    num_devices = len([d for d in max_memory if d != "cpu" and int(d) < expected_device_count and max_memory[d] > 0])
 
     if num_devices == 0:
         return max_memory
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -990,20 +990,10 @@ def get_balanced_memory(
     max_memory = get_max_memory(max_memory)
 
     if is_npu_available():
-        expected_device_count = torch.npu.device_count()
-    elif is_mlu_available():
-        expected_device_count = torch.mlu.device_count()
-    elif is_xpu_available():
-        expected_device_count = torch.xpu.device_count()
+        expected_device_type = "npu"
     else:
-        expected_device_count = torch.cuda.device_count()
-    num_devices = len(
-        [
-            d
-            for d in max_memory
-            if d not in ["mps", "cpu", "disk"] and int(d) < expected_device_count and max_memory[d] > 0
-        ]
-    )
+        expected_device_type = "cuda"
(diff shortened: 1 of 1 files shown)


yao-matrix:
OK for me.

src/accelerate/utils/modeling.py
```diff
@@ -990,14 +990,20 @@ def get_balanced_memory(
...
-        expected_device_type = "cuda"
-    num_devices = len([d for d in max_memory if torch.device(d).type == expected_device_type and max_memory[d] > 0])
+        expected_device_count = torch.cuda.device_count()
+    num_devices = len(
+        [
+            d
+            for d in max_memory
+            if d not in ["mps", "cpu", "disk"] and int(d) < expected_device_count and max_memory[d] > 0
+        ]
+    )
```
SunMarc, line 1006: Can't we just set `expected_device_type="cuda" `for both `xlu` and `mlu` ? Also, I prefer to not take into account the available device count when computing the `num_devices`. Even if we calculate the right number of devices, it will still fail since `max_memory` still still have these keys. This check is already done in `get_max_memory` function !

SunMarc (commented):
Thanks for your work @faaany ! I left a comment

src/accelerate/utils/modeling.py
faaany: yes, I see your point. Thx for explaining! Code updated.

src/accelerate/utils/modeling.py
```diff
@@ -991,10 +991,6 @@ def get_balanced_memory(
 
     if is_npu_available():
         expected_device_type = "npu"
-    elif is_mlu_available():
-        expected_device_type = "mlu"
```
SunMarc, line 995: The issue you shared @faaany is about the `to() `methods for mlu. I'm not sure if `torch.device(d).type` will really return `cuda` with mlu. Do you have any insights @huismiling since you added the [support](https://github.com/huggingface/accelerate/pull/2552) to mlu ? To be safe, I would suggest reverting this change.

SunMarc (commented):
Sounds good ! I left a comment about mlu since i'm not sure this is safe to change the `expected_device_type`.

src/accelerate/utils/modeling.py
faaany: good idea, I will update it. But I am also very curious about the behavior on npu.

Hi @statelesshz , could you help us verify what `torch.device(0)` returns on NPU? It is cuda or npu? Thanks a lot!

huismiling: @faaany  Hi, MLU devices type is `mlu` . 

```
>>> torch.device(0).type
'mlu'
```

SunMarc: Thanks @huismiling! Could you update @faaany ?

faaany: @huismiling thanks for updating!

faaany:
Good news is that during this time our proposal to make torch.device(0) return 'xpu' on xpu got approved by stock pytorch (PR [link](https://github.com/pytorch/pytorch/pull/129119)). To avoid "reverse-engineering", let me close this PR. Thanks so much for the discussion! @SunMarc

muellerzr:
That's great @faaany !

SunMarc:
Nice :)

Issue #2155, opened 2023-11-15
Accelerate FSDP always removed {'model.norm.weight'} layer of model when saving them


gauss5930:
### System Info

```Shell
Accelerate: 0.24.1
OS: Linux-5.4.0-148-generic-x86_64-with-glibc2.35
Python version: 3.10.12
Numpy version: 1.24.1
PyTorch: 2.1.0+cu118

Accelerate Configuration

compute_environment: LOCAL_MACHINE                                                                                             
debug: false                                                                                                                   
distributed_type: FSDP
downcast_bf16: 'no'
fsdp_config:
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_forward_prefetch: true
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: SHARDED_STATE_DICT
  fsdp_sync_module_states: true
  fsdp_use_orig_params: true
machine_rank: 0
main_training_function: main
mixed_precision: 'bf16'
num_machines: 0
num_processes: 2
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Run my own fine-tuning code with the Accelerate FSDP config mentioned above on 2 * A100 80G GPUs. I also used `use_flash_attention_2=True` and `gradient_checkpointing=True`. The following command and code were used for fine-tuning. Actually, I did set the epoch to 3 and max_step to 'max', but I changed the value of the hyperparameter to check the error message early. 

```
!accelerate launch --config_file=accelerate_configs/fsdp_config.yaml --num_processes=2 finetuning/finetune.py \
    --model_path beomi/llama-2-koen-13b \
    --data_path Cartinoe5930/KoRAE_filtered_12k \
    --output_dir finetuning/result/llama2/ \
    --wandb_project example_1 \
    --wandb_run_name example_1 \
    --hf_hub_path HUGGINGFACE_HUB_PATH_TO_UPLOAD \
[...]

PR #2875, opened 2024-06-20
Correct loading of models with shared tensors when using accelerator.load_state()

jkuntzer:
# What does this PR do?

I would run into problems with PyTorch's _load_state_dict_ complaining about missing keys. These keys belonged to shared tensors. These shared keys are intentionally omitted by the safetensors library. To load a model correctly, one has to use safetensor's _load_model_ function instead of the default _load_state_dict_ function (described [here](https://huggingface.co/docs/safetensors/torch_shared_tensors)). This was previously not done when using the _load_state_ function of the _Accelerator_.

Fixes # (issue)
I think this issue might be relevant as they also report problems when loading with _accelerator.load_state_.
https://github.com/huggingface/accelerate/issues/2155

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -18,7 +18,7 @@ from typing import List
 
 import numpy as np
 import torch
-from safetensors.torch import load_file
+from safetensors.torch import load_file, load_model
 from torch.cuda.amp import GradScaler
 
 from .utils import (
@@ -196,12 +196,12 @@ def load_accelerator_state(
         ending = f"_{i}" if i > 0 else ""
         input_model_file = input_dir.joinpath(f"{SAFE_MODEL_NAME}{ending}.safetensors")
         if input_model_file.exists():
-            state_dict = load_file(input_model_file, device=str(map_location))
+            load_model(model, input_model_file, device=str(map_location), **load_model_func_kwargs)
         else:
             # Load with torch
             input_model_file = input_dir.joinpath(f"{MODEL_NAME}{ending}.bin")
             state_dict = torch.load(input_model_file, map_location=map_location)
-        models[i].load_state_dict(state_dict, **load_model_func_kwargs)
+            model.load_state_dict(state_dict, **load_model_func_kwargs)
     logger.info("All model weights loaded successfully")
 
     # Optimizer states
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -18,7 +18,7 @@ from typing import List
 
 import numpy as np
 import torch
-from safetensors.torch import load_file, load_model
+from safetensors.torch import load_model
 from torch.cuda.amp import GradScaler
 
 from .utils import (
--- a/tests/test_accelerator.py
+++ b/tests/test_accelerator.py
@@ -66,7 +66,7 @@ class ModelForTest(torch.nn.Module):
 
 
 def get_signature(model):
-    return sum((param.abs().sum().item() for param in model.parameters()))
+    return sum(param.abs().sum().item() for param in model.parameters())
 
 
 def load_random_weights(model):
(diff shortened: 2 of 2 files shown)


src/accelerate/checkpointing.py
```diff
@@ -196,12 +196,12 @@ def load_accelerator_state(
...
         if input_model_file.exists():
-            state_dict = load_file(input_model_file, device=str(map_location))
+            load_model(model, input_model_file, device=str(map_location), **load_model_func_kwargs)
         else:
             # Load with torch
             input_model_file = input_dir.joinpath(f"{MODEL_NAME}{ending}.bin")
             state_dict = torch.load(input_model_file, map_location=map_location)
-        models[i].load_state_dict(state_dict, **load_model_func_kwargs)
+            model.load_state_dict(state_dict, **load_model_func_kwargs)
     logger.info("All model weights loaded successfully")
```
muellerzr, line 205: Any particular reason for this change? I'd expect only the prior to be modified.

muellerzr (commented):
Thanks! Just one question

src/accelerate/checkpointing.py
SunMarc: In the if statement, he's loading the safetensors model directly whereas before, we were only getting the state dict.

SunMarc (commented):
Thanks for the change and spotting the issue ! Could you add a test with a model with tied weight ? You can use the following test for reference : [test_save_load_model](https://github.com/huggingface/accelerate/blob/8330b375d4c076568096485b2b8d012c597a4141/tests/test_accelerator.py#L234)

src/accelerate/checkpointing.py
jkuntzer: _load_model_ does both: it loads the file and uses it to populate the _state_dict_. Previously, each branch of the if-condition only loaded the file and after the if-condition, the model would load the state dict. Since _load_model_ does both, I indented the statement on line 204 to become part of the else-clause. This becomes clearer when you have a look at the complete surroundings of the changes instead of only the affected lines.

jkuntzer:
Yes, I'll have a look into it.

jkuntzer:
You can verify that the shared weights are implemented correctly by checking the output. _safetensors_ warns you about that fact.

SunMarc (approved):
Thanks for iterating and adding the tests @jkuntzer  ! Could you do a final check and see if the test that you added fails when you remove the changes you did ?

jkuntzer:
Just did. This is the expected error message I get when reverting my changes.
Screenshot from 2024-07-09 15-15-51

tests/test_accelerator.py
```diff
@@ -31,9 +31,23 @@
...
+    def __init__(self):
+        super().__init__()
+        self.linear1 = torch.nn.Linear(2, 4)
+        self.linear2 = torch.nn.Linear(4, 2)
+        self.linear2.weight = self.linear1.weight
+        self.linear2.bias = self.linear1.bias
+
+        # need to add this for compliance with other methods
+        self.weight = self.linear1.weight
+        self.bias = self.linear1.bias
```
SunMarc, line 44: Do we really need that ? where does it fail ?

SunMarc:
> Just did. This is the expected error message I get when reverting my changes.

I was only expecting linear2.weight and linear2.bias to be missing. Maybe this is due to 
```python 
self.weight = self.linear1.weight
self.bias = self.linear1.bias
```

tests/test_accelerator.py
jkuntzer: It used to fail previously. You're right. This part can be safely removed.

jkuntzer:
> > Just did. This is the expected error message I get when reverting my changes.
> 
> I was only expecting linear2.weight and linear2.bias to be missing. Maybe this is due to
> 
> ```python
> self.weight = self.linear1.weight
> self.bias = self.linear1.bias
> ```

After removing the unnecessary bits, it correctly only throws an error for the weights and bias of the 2nd linear layer.
Screenshot from 2024-07-09 16-41-56

SunMarc:
Nice ! Could you just fix the quality issue (`make style`) and we are good to merge !

muellerzr (approved):
Thanks!

PR #2878, opened 2024-06-21
add `require_triton` and enable `test_dynamo` work on xpu

faaany:
## What does this PR do?
`test_dynamo` works on xpu, but fails because it requires extra installation of the triton library. This PR adds the `require_triton` test marker and make `test_dynamo` device-agnostic.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -27,12 +27,12 @@ from torch import nn
 
 from accelerate.state import PartialState
 from accelerate.test_utils.testing import (
-    require_cuda,
     require_huggingface_suite,
     require_non_cpu,
     require_non_torch_xla,
     require_torch_min_version,
     require_tpu,
+    require_triton,
     torch_device,
 )
 from accelerate.test_utils.training import RegressionModel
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -190,15 +190,16 @@ def test_can_undo_fp16_conversion(self):
         model = extract_model_from_parallel(model, keep_fp32_wrapper=False)
         _ = pickle.dumps(model)
 
-    @require_cuda
+    @require_triton
```
SunMarc, line 193: I think this should be fine but I would love to have your opinion @BenjaminBossan. triton should be installed when we install torch.

SunMarc (approved):
Thanks for adding this ! Can you have a second look @BenjaminBossan ?

tests/test_utils.py
```diff
@@ -190,15 +190,16 @@ def test_can_undo_fp16_conversion(self):
...
+    @require_non_cpu
```
BenjaminBossan, line 194: So by adding this, we enable this test on all non-CPU devices. Do we know that it passes on all of them? We don't cover non-CUDA on CI so we would not notice if we break it. Would it be better to explicitly check for XPU?

BenjaminBossan (commented):
Hmm, I always thought that triton comes with torch. I just created a test env and ran `pip install torch` and indeed triton was installed. Under what circumstances can it be missing, is it dependent on the device?

faaany:
> Hmm, I always thought that triton comes with torch. I just created a test env and ran `pip install torch` and indeed triton was installed. Under what circumstances can it be missing, is it dependent on the device?

yes, it depends on the device. For the pytorch cuda distribution, triton is installed by default. But for XPU and CPU, triton will not be installed by default.

tests/test_utils.py
faaany: I see your point, but this might be an opportunity to let them know that they should support it. And I think the previous device-agnostic work in transformers (#[25654](https://github.com/huggingface/transformers/issues/25654)) doesn't take into account whether the test case is supported yet or not. For devices that are not supported, they just add a manual skip. But I am not sure which way is the best to go for accelerate. Just let me know, I will update my code. :)

faaany:
Just let me know how you decide; I will update my code accordingly. Thanks so much for the review! @BenjaminBossan @SunMarc

BenjaminBossan:
Thanks for providing further information on the triton dependency. In that case, I agree that adding an explicit check is fine.

As to whether we should potentially break the tests for other non-CUDA devices: It's hard to say if this would be a good thing or not, as we don't know how that will affect the corresponding maintainers. Probably it's best for Zach to judge when he's back.

muellerzr (approved):
Thanks, this solution seems fine. I'll make a note of it in the release notes today and we can keep an eye out for any issues/follow up PRs from folks that found this broke things :)

PR #2879, opened 2024-06-21
fix `load_state_dict` for xpu and refine xpu safetensor version check 

faaany:
## What does this PR do?

- enable `test_load_state_dict` on xpu, because `torch.device(0)` by default is cuda
- add xpu support in `load_state_dict` when `device_map`'s values are identical integers 
- refine the xpu safetensor version check to make code clean and compact


src/accelerate/utils/modeling.py
```diff
@@ -1486,17 +1493,12 @@ def load_state_dict(checkpoint_file, device_map=None):
...
-                if is_xpu_available():
-                    current_safetensors_version = packaging.version.parse(importlib.metadata.version("safetensors"))
-
-                    if compare_versions(current_safetensors_version, "<", "0.4.2"):
-                        raise ModuleNotFoundError(
-                            f"You need at least safetensors 0.4.2 for Intel GPU, while you have {current_safetensors_version}"
+                if isinstance(device, int) and is_xpu_available():
+                    if compare_versions("safetensors", "<", "0.4.2"):
+                        raise ImportError(
+                            "Safetensors version must be >= 0.4.2 for Intel GPU. Please update safetensors."
```
yao-matrix, line 1499: I suppose device int judgement and xpu judgement should be parallel, tell me the reason you put it w/ and.

src/accelerate/utils/modeling.py
```diff
@@ -1456,7 +1454,16 @@ def load_state_dict(checkpoint_file, device_map=None):
         else:
             # if we only have one device we can load everything directly
             if len(set(device_map.values())) == 1:
-                return safe_load_file(checkpoint_file, device=list(device_map.values())[0])
+                device = list(device_map.values())[0]
+                target_device = device
+                if is_xpu_available() and isinstance(device, int):
```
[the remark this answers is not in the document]
faaany: there are 3 data types for `device` available: int, str and torch.device. Among these cases, only the integer case is worth extra handling for xpu. As a reference: https://github.com/faaany/transformers/blob/main/src/transformers/pipelines/base.py#L866.

src/accelerate/utils/modeling.py
faaany: why parallel? as a reference: https://github.com/huggingface/accelerate/blob/1f7a79b428749f45187ec69485f2c966fe21926e/src/accelerate/big_modeling.py#L482  but I put the integer check first, because if it is not an integer, no meaning to check xpu device. But this might not be the conversion.

faaany:
@SunMarc this PR is ready for review. Could you help review it? Thx a lot!

SunMarc (approved):
LGTM ! Thanks for fixing @faaany !

muellerzr (approved):
Nice!

muellerzr:
@faaany small merge conflict (the joys of day-of-release-merging-post-OOO), if you can fix that I'll get this in and it'll be part of the release today 🚀

faaany:
> @faaany small merge conflict (the joys of day-of-release-merging-post-OOO), if you can fix that I'll get this in and it'll be part of the release today 🚀

awesome, thx! conflict resolved.

tests/test_modeling_utils.py
```diff
@@ -748,7 +748,10 @@ def test_load_state_dict(self):
 
             for param, device in device_map.items():
                 device = device if device != "disk" else "cpu"
-                assert loaded_state_dict[param].device == torch.device(device)
+                expected_device = (
+                    torch.device(f"{torch_device}:{device}") if isinstance(device, int) else torch.device(device)
+                )
+                assert loaded_state_dict[param].device == expected_device
```
muellerzr, line 754: FYI this breaks CUDA tests because on CUDA we end up with `cuda:0:0`

muellerzr:
Please check out the new failure on main @faaany, thanks! :) (Test is failing bc of the aforementioned note above)

tests/test_modeling_utils.py
SunMarc: +1

PR #2883, opened 2024-06-23
Add Profiler Support for Performance Analysis

yhna940:
# What does this PR do?

This PR introduces profiler support to the Accelerate library, enabling users to collect performance metrics during model training and inference. The profiler allows for detailed analysis of execution time and memory consumption of model operators and can generate profiling traces for visualization in Chrome's tracing tool. Additionally, it provides options to profile long-running jobs with customizable scheduling options.

profile_export

Key changes include:
- Addition of `ProfileKwargs` to customize profiling options.
- Implementation of profiling context manager in the `Accelerate` class.
- New documentation explaining how to use the profiler with examples.
- Updated examples to demonstrate profiling capabilities (`profiler.py`).
- Tests for profiling functionality with various scheduling options.

## Context and Motivation
Other frameworks like [MMEngine](https://mmengine.readthedocs.io/en/latest/api/generated/mmengine.hooks.ProfilerHook.html) and [PyTorch Lightning](https://pytorch-lightning.readthedocs.io/en/1.5.10/advanced/profiler.html) offer profiling techniques based on the [Torch Profiler](https://pytorch.org/docs/stable/profiler.html). Inspired by these tools, we aimed to bring similar profiling capabilities to Accelerate. This enhancement helps users optimize and improve model performance by providing insights into the computational and memory aspects of their models.

## Before submitting
      Pull Request section?
      to it if that's the case.


BenjaminBossan (commented):
Thanks for implementing this cool feature. At a first glance, the PR already looks super clean, really great job.

I checked the [rendered profiler docs](https://moon-ci-docs.huggingface.co/docs/accelerate/pr_2883/en/usage_guides/profiler) and there seems to be an issue with `</hfoption>`, not sure what's going on there. Could you please check? _Edit_: It seems you were faster than me :)

I also saw that the dates of the copyright headers you added are outdated, could you please update them to 2024?

Otherwise, this PR looks great to me. I don't have a ton of experience with the profiler -- except that I found that it can be super slow :D. In terms of the design of the feature, it looks quite nice to me, but let's wait for Zach's return to office for a full review.

yhna940:
> Thanks for implementing this cool feature. At a first glance, the PR already looks super clean, really great job.
> 
> I checked the [rendered profiler docs](https://moon-ci-docs.huggingface.co/docs/accelerate/pr_2883/en/usage_guides/profiler) and there seems to be an issue with `</hfoption>`, not sure what's going on there. Could you please check? _Edit_: It seems you were faster than me :)
> 
> I also saw that the dates of the copyright headers you added are outdated, could you please update them to 2024?
> 
> Otherwise, this PR looks great to me. I don't have a ton of experience with the profiler -- except that I found that it can be super slow :D. In terms of the design of the feature, it looks quite nice to me, but let's wait for Zach's return to office for a full review.


Thank you for the quick review @BenjaminBossan !

I appreciate you pointing out the issue with the `</hfoption>` tag in the docs. I've corrected it now. Additionally, I've updated the copyright headers to 2024 as requested.

However, I noticed that one image in the docs is broken. Could you please provide some guidance on how to fix this? 🙏

Regarding the profiler's performance, I'm also interested in investigating the slowdown you mentioned. On my local server, it seems to work fine. Could you provide more details on where you're experiencing the slowness?

Thanks again for your feedback :)

examples/by_feature/profiler.py
```diff
@@ -0,0 +1,254 @@
+# Copyright 2021 The HuggingFace Inc. team. All rights reserved.
```
BenjaminBossan, line 1: Still old year.

src/accelerate/accelerator.py
```diff
@@ -3356,6 +3363,68 @@ def autocast(self, cache_enabled: bool = False, autocast_handler: AutocastKwargs
...
+        with profile_handler.build() as profiler:
+            yield profiler
+
+        if profile_handler.output_trace_dir is None:
+            return
+
+        os.makedirs(profile_handler.output_trace_dir, exist_ok=True)
+        profiler.export_chrome_trace(
+            os.path.join(
+                profile_handler.output_trace_dir, PROFILE_PATTERN_NAME.format(suffix=f"_{self.process_index}")
```
BenjaminBossan, line 3423: How about adding the `_` to the `PROFILE_PATTERN_NAME` directly?

src/accelerate/utils/dataclasses.py
```diff
@@ -350,6 +350,117 @@ def __post_init__(self):
                 raise ValueError(f"`optimization_level` must be one of {' or '.join(get_args(OptLevel))}")
 
 
+# Literal
+ProfilerActivity = Literal["cpu", "xpu", "mtia", "cuda"]
+
+
+@dataclass
+class ProfileKwargs(KwargsHandler):
```
BenjaminBossan, line 358: I think for this docstring to be included in the API docs, you need to add an entry to `docs/source/package_references/kwargs.md`.

tests/test_examples.py
```diff
@@ -248,6 +249,10 @@ def test_early_stopping(self):
         testargs = ["examples/by_feature/early_stopping.py"]
         run_command(self.launch_args + testargs)
 
+    def test_profiler(self):
```
BenjaminBossan, line 252: At least on my machine, this takes a significant amount of time to run, should `@slow` be added?

tests/test_kwargs_handlers.py
```diff
@@ -96,6 +96,46 @@ def test_autocast_kwargs(self):
...
+            cycles = effective_steps // steps_per_cycle
+            if option["repeat"] > 0:
+                expected_count = min(cycles, option["repeat"])
+            else:
+                expected_count = cycles
+
+            def on_trace_ready(prof):
+                nonlocal count
+                count += 1
+                print(prof.key_averages().table(sort_by="cpu_memory_usage", row_limit=-1))
```
BenjaminBossan, line 124: I wonder: Instead of printing, could we, for instance, just add this to a list, then do a basic sanity check on the output, for example what columns we should expect?

BenjaminBossan (changes requested):
This looks pretty good already, I have only some minor comments. As mentioned, the full review should come from Zach when he's back.

> However, I noticed that one image in the docs is broken.

I wonder if it would ahve been okay as it was, once the PR would be merged. Currently, the image is indeed not at the original location, but that would change after the merge. Unfortunately, I couldn't find any example in the existing docs to verify this.

> Regarding the profiler's performance, I'm also interested in investigating the slowdown you mentioned. On my local server, it seems to work fine. Could you provide more details on where you're experiencing the slowness?

From memory, when I tested it, I found some sources of slow down:

- Processing the data seems to take quite some time when the model is big.
- Some options (IIRC `with_stack`) would considerably slow down the run itself.
- The generated trace could be huge (>1GB) and bring Chrome to its knees.

Of course, this is to be expected to some extent but it made me not want to use the feature ;)

examples/by_feature/profiler.py
yhna940: Done `20f377b`

src/accelerate/accelerator.py
yhna940: Done `9cfbbf3`

src/accelerate/utils/dataclasses.py
yhna940: Done `06a2430`

tests/test_examples.py
yhna940: Thank you :) I totally agree with your comment. I added a `slow` decorator to the function.

yhna940:
Thank you for the review and the feedback! I appreciate the insights and will wait for Zach's full review for any additional comments or suggestions.

I will conduct some experiments to further investigate the performance impacts of `with_stack` and data processing times.

Additionally, I propose updating the user guide to include tips or warnings about handling large or complex tasks using the PyTorch profiler's API for long-running jobs. As highlighted in the [PyTorch official documentation](https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html):

> PyTorch profiler offers an additional API to handle long-running jobs (such as training loops). Tracing all of the execution can be slow and result in very large trace files. To avoid this, use optional arguments:
> 
> - `schedule` - specifies a function that takes an integer argument (step number) as an input and returns an action for the profiler, the best way to use this parameter is to use the `torch.profiler.schedule` helper function that can generate a schedule for you;
> - `on_trace_ready` - specifies a function that takes a reference to the profiler as an input and is called by the profiler each time the new trace is ready.

By updating the documentation with these details, we can provide more comprehensive guidance to users, helping them to effectively manage performance profiling for their models.

Let me know if there's anything else I should address or if you have further suggestions!

Thanks again for the review :)

BenjaminBossan (approved):
Thanks a lot for the updates. From my point of view, this looks good. As mentioned, we'll wait for Zach for the final review.

> I will conduct some experiments to further investigate the performance impacts of with_stack and data processing times.

> By updating the documentation with these details, we can provide more comprehensive guidance to users, helping them to effectively manage performance profiling for their models.

I wouldn't put too much time into this, as I'm sure this is a moving target and we can't constantly monitor the PyTorch docs for updates. I think it's sufficient to refer users to the PyTorch docs for caveats and performance tips.

muellerzr (approved):
This is an exceedingly clean, thorough, and excellent PR @yhna940, very nice work! Let's get this in 🔥 

cc @stas00

Issue #2859, opened 2024-06-15
Improve `skip_first_batches` method to efficiently support `IterableDataset` and `StatefulDataloader`


yzhangcs:
Hi all, Thank you for developing this great project.
Currently, the implementation naively iterates through all batches until the specified number have been consumed, which can be extremely slow for very large datasets. 
The latest version of the [datasets](https://github.com/huggingface/datasets/releases/tag/2.20.0) library now supports resumable iterable datasets, as well as the StatefulDataloader to allow for efficient resumption of training states.
![a3049bf3e1246379c558fe8133c6d34e](https://github.com/huggingface/accelerate/assets/18402347/1916f2a3-9c17-453a-9cb9-d617e632886c)

I'm wondering if there are any plans to leverage these new features in Accelerate to make `skip_first_batches` more efficient and compatible with the latest datasets capabilities? 
If not, are there plans to add support for this in the future?
Efficiently skipping batches on huge datasets would significantly speed up resuming interrupted training runs. Let me know if you need any additional information or have thoughts on the best way to approach this.

Thanks for considering this suggestion!

muellerzr:
You can ping myself (@muellerzr) or @SunMarc on these things, Sylvain hasn't worked at HF for well over a year or two now :)

muellerzr:
Yes, we are indeed actively looking into this!

byi8220:
Ran into something annoying while looking at this. Merely *importing* StatefulDataLoader (i.e. putting the line `from torchdata.stateful_dataloader import StatefulDataLoader` anywhere in the code) causes one of the unit test, `check_seedable_sampler` to fail.

I suspect it has something to do with torchdata overriding torch's BatchSampler [here](https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/sampler.py#L134-L135). This is supported by the fact if I import this and add some logging, it seems SeedableRandomSampler.__iter__() is called one less time than expected:
```
# We should see the epoch and seed sequence [(0, 42), (1, 43), (2, 44)] twice, but the first call with seed 42 is missing
# It looks like the first sample is being drawn without setting a seed

stdout: stdout: Shuffled central dataloader passing.
stdout: stdout: {'x': tensor([-1.3022,  0.1278], device='cuda:0'), 'y': tensor([0.3097, 3.2926], device='cuda:0')}
[...]

Issue #2894, opened 2024-06-26
Importing `torchdata.stateful_dataloader` causes the test `check_seedable_sampler` to fail


byi8220:
Merely *importing* StatefulDataLoader from the nightly [`torchdata`](https://github.com/pytorch/data/tree/main) package (i.e. putting the line `from torchdata.stateful_dataloader import StatefulDataLoader` anywhere in the code) causes one of the unit test, `check_seedable_sampler` to fail.

Stack trace obtained by running tests with the import
```
stderr: Traceback (most recent call last):
stderr:   File "redacted/accelerate/src/accelerate/test_utils/scripts/test_script.py", line 827, in <module>
stderr:     main()
stderr:   File "redacted/accelerate/src/accelerate/test_utils/scripts/test_script.py", line 802, in main
stderr:     check_seedable_sampler()
stderr:   File "redacted/accelerate/src/accelerate/test_utils/scripts/test_script.py", line 381, in check_seedable_sampler
stderr:     assert torch.allclose(original_items, new_items), "Did not obtain the same items with the same seed and epoch."
stderr:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
stderr: AssertionError: Did not obtain the same items with the same seed and epoch.
```

I suspect it has something to do with `torchdata` overriding torch's BatchSampler [in this code](https://github.com/pytorch/data/blob/main/torchdata/stateful_dataloader/sampler.py#L134-L135). This is supported by the fact if I import this and add some logging, it seems SeedableRandomSampler.\_\_iter\_\_() is called one less time than expected:
```
# We should see the epoch and seed sequence [(0, 42), (1, 43), (2, 44)] twice, but the first call with seed 42 is missing
# It looks like the first sample is being drawn without setting a seed

stdout: stdout: Shuffled central dataloader passing.
stdout: stdout: {'x': tensor([-1.3022,  0.1278], device='cuda:0'), 'y': tensor([0.3097, 3.2926], device='cuda:0')}
stdout: stdout: {'x': tensor([-1.0400,  0.7505], device='cuda:0'), 'y': tensor([0.9978, 4.5075], device='cuda:0')}
stdout: stdout: {'x': tensor([ 0.3047, -0.0168], device='cuda:0'), 'y': tensor([3.6974, 3.0542], device='cuda:0')}
[...]

PR #2895, opened 2024-06-26
Add early support for `torchdata.stateful_dataloader.StatefulDataLoader` within the `Accelerator`

byi8220:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/2859

This PR does the following:

1. Added a new field `use_stateful_dataloader` in `DataLoaderConfiguration`. Passing this into the config makes it so that all `DataLoader`s prepared and returned by the Accelerator are [`StatefulDataLoader` objects from the torchdata library](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader)
2. Create a class `DataLoaderAdapter` which can wrap around and act as either PyTorch's `DataLoader`, or other variants of it such as `StatefulDataLoader`
3. Refactor `DataLoaderShard`, `DataLoaderDispatcher`, and `SkipDataLoader` to inherit from `DataLoaderAdapter` instead of `DataLoader`

## Testing

Added new unit tests to test that StatefulDataLoader can be dropped in and loaded and saved from.

#### Caveats

- The `torchdata` package may have conflicts with `accelerate`, see https://github.com/huggingface/accelerate/issues/2894
    - However, if `torchdata` is not installed, all existing tests pass, suggesting this is not regressive. 
- `torchdata.stateful_dataloader.StatefulDataLoader` is only available in the beta build of `torchdata`, this is not a stable feature.
- Adding another dependency (on a nightly package) means that almost none of the tests added in this PR is done underneath the existing images or imports.
- This has only been tested on my local workstation using a single GPU.
- The implementation of `DataLoaderAdapter` is somewhat invasive and uses some questionable reflection

## Before submitting
      Pull Request section?
- [X] Was this discussed/approved via a Github issue (see above)
- [X] Did you write any new necessary tests?

## Who can review?

@muellerzr


src/accelerate/data_loader.py
```diff
@@ -388,9 +394,63 @@ def end(self):
...
+    """
+    A class which wraps around a PyTorch `DataLoader` (or variants of it) to be used with the `Accelerator`. For
+    compatability reasons, this class inherits from the class it wraps around, so it can be used as a drop-in.
+    """
+
+    def __init__(self, dataset, use_stateful_dataloader=False, batch_sampler=None, **kwargs):
+        self.use_stateful_dataloader = use_stateful_dataloader
+
+        if use_stateful_dataloader and not is_torchdata_stateful_dataloader_available():
+            raise ValueError("StatefulDataLoader is not available. Please install torchdata to use it.")
```
muellerzr, line 407:
```suggestion
            raise ImportError("StatefulDataLoader is not available. Please install torchdata to use it.")
```

src/accelerate/data_loader.py
```diff
@@ -19,6 +19,12 @@
 import torch
 from torch.utils.data import BatchSampler, DataLoader, IterableDataset, RandomSampler
 
+from accelerate.utils.imports import is_torchdata_stateful_dataloader_available
+
+
+if is_torchdata_stateful_dataloader_available():
+    from torchdata.stateful_dataloader import StatefulDataLoader
```
muellerzr, line 26: To keep base imports fast, checks like this should happen directly when needed. In this case the import should occur during`DataLoaderAdapter`'s `__init__`

src/accelerate/test_utils/testing.py
```diff
@@ -29,6 +29,7 @@
 import torch
 
 import accelerate
+from accelerate.utils.imports import is_torchdata_stateful_dataloader_available
```
muellerzr, line 32: We should add` is_torchdata_stateful_dataloader_available` to `accelerate.utils.__init__.py` and import from there please :)

src/accelerate/utils/dataclasses.py
```diff
@@ -720,6 +720,14 @@ class DataLoaderConfiguration:
             " prepared dataloader has `pin_memory` set to `True` to work properly."
         },
     )
+    use_stateful_dataloader: bool = field(
+        default=False,
+        metadata={
+            "help": "If set to true, the dataloader prepared by the Accelerator will be backed by "
```
muellerzr, line 726:
```suggestion
            "help": "If set to `True`, the dataloader prepared by the Accelerator will be backed by "
```

muellerzr (approved):
Thanks! This LG2M bar some nits. (Esp nice now that it's no longer a monkey-patch :) ).

Can you verify the tests pass on a multi-GPU system?

Also would be good to add `torchdata` in the test requirements in the `setup.py`

src/accelerate/data_loader.py
byi8220: Done, but I think I also need to put this import in `skip_first_batches` since there's a usage of StatefulDataLoader there as well.

src/accelerate/test_utils/testing.py
byi8220: Done

byi8220:
> Can you verify the tests pass on a multi-GPU system?

I have not written any tests specifically for multi-gpu. However, the PR as is passes on a 2 GPU cloud instance I spun up on GCP for [the tests added so far](https://gist.github.com/byi8220/b5909da3d3ad132e0274d47d76d4f176). 

Would the tests I've put into the PR so far be enough, or should I add specific multi-GPU tests? (I only have 1 GPU on my physical workstation so it costs me money to run these tests :sweat:)

> Also would be good to add torchdata in the test requirements in the setup.py

I think `StatefulDataLoader` is still a beta feature and not officially included in any stable release. Their [docs](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader#installation) state it's only available in nightly builds for now, which are not found on PyPI.

According to https://stackoverflow.com/q/68809295 you really shouldn't be trying to specify `extra_index_url` in setup.py.

I'm not sure of a solution besides between trying to hack it out or waiting for the required features in torchdata to be released into stable.

muellerzr:
Hi @byi8220, will be reviewing this in the next few days!

src/accelerate/data_loader.py
```diff
@@ -19,6 +19,8 @@
 import torch
 from torch.utils.data import BatchSampler, DataLoader, IterableDataset, RandomSampler
 
+from accelerate.utils import is_torchdata_stateful_dataloader_available
```
muellerzr, line 22: This can now be added under the `from .utils import (` block in line 26 :)

muellerzr (changes requested):
Thanks for doing this. Thinking on this more, this is far too magical. Is there an issue with just duplicating the code into each `DataLoader` type instead of introducing a new subclass? (e..g the `load_state_dict` logic). 

I'd like some clarification on that, as even if it is more LOC, I'd rather not deal with more subclasses and have code duplication instead.

src/accelerate/test_utils/scripts/external_deps/test_metrics.py
```diff
@@ -235,6 +235,11 @@ def test_gather_for_metrics_drop_last():
 
     # Should return a full set of complete batches from each GPU
     num_expected_items = per_device_batch_size * accelerator.num_processes
+    print("dataloader.batch_size:", dataloader.batch_size)
+    print("accelerator.num_processes:", accelerator.num_processes)
+    print("gathered_items:", gathered_items)
+    print("batch:", batch)
+    print("len(dataloader):", len(dataloader))
```
byi8220, line 242: These lines should probably be removed. I left these in for debugging my tests

byi8220:
> Is there an issue with just duplicating the code into each DataLoader type instead of introducing a new subclass?

There's one obvious way to get this feature working without some magical reflection tricks:

1. Create equivalent `StatefulDataLoaderDispatcher`, `StatefulDataLoaderShard`, `StatefulSkipDataLoader` classes which are identical except that it inherits from `StatefulDataLoader` and implement the state_dict functions.
2. Create helper functions such as `create_data_loader_dispatcher()` which take in all the constructor args and pick whether to create a StatefulDataLoader or regular DataLoader
3. Replace all constructor calls with these helper functions.

Doing that would add a ton of code duplication, but it should be mostly functionally equivalent to what I have now. The only thing you would need to be careful with moving forward is now you must *always* use the helper functions to create these objects instead of the class constructors.

If you want to avoid evil dynamic reflection, I could refactor this PR to do the above instead without too much effort.

There's some other ideas I thought about writing this PR, mostly involving composing DataLoader into these classes, but most of them felt kinda wrong or that they might break something.

src/accelerate/data_loader.py
byi8220: done

byi8220:
Actually, trying to draft up the change I just mentioned reminded me why I gave up on duplicating the code in the first place. It's a mess. Issues I needed to workaround include:
1. Since StatefulDataLoader inherits from DataLoader, attempting something like having StatefulDataLoaderDispatcher inherit from DataLoaderDispatcher creates a diamond problem. This leads to a ton of messy MRO issues which feel like too much mental effort to process over simply just *not* having this inheritance.
2. By separating the types, there is a bunch of boilerplate handling code that needs to consider the exact details of the DataLoader being passed around.
3. The existing issues with this being a feature depending on a nightly, unstable build of `torchdata` remain (CI gap, needing dependency checking boilerplate, weirdly written tests, so on)

I'll defend my `DataLoaderAdapter<T> : T` solution since I find it elegant, but I recognize it's horribly confusing and invokes some pretty cursed stuff, so I'm in full support of getting rid of it for something more overt and readable.

If you want to see what it would look like, I've made those changes in [a separate branch](https://github.com/byi8220/accelerate/tree/stateful-dataloader-2).

Here's the diff between this PR and the above branch: https://github.com/byi8220/accelerate/compare/stateful-dataloader...byi8220:accelerate:stateful-dataloader-2)

@muellerzr If the code above looks better, I can merge those into this PR's working branch and move on with that instead.

muellerzr:
TBH I'm not sure if I like either. Let me work on this today and tommorow and see where I end up. Whichever solution winds up being similar to yours (unless a third one pops up) we'll go with it.

byi8220:
> TBH I'm not sure if I like either. 

It's tricky. My understanding of the problem is this: We want to do something like this"If `use_stateful_dataloader==True`, then create a `StatefulDataLoader` instead of a `DataLoader`." 

However, based on how classes like `DataLoaderShard` are currently implemented, what we have to do in practice is "If `use_stateful_dataloader==True`, then make this class *inherit* from `StatefulDataLoader` instead of `DataLoader`." This is a lot more awkward to code.

I have one more idea for a solution, which is to have those classes create a `base_dataloader` and manually code the passthrough to methods and properties. I feel like this is a bit fragile, and isn't too different from my original solution.

muellerzr (approved):
The more I've been working on this, the more I actually think this is the best solution we can get. Thanks a bunch for doing this, I think even though it's annoying with the patches, there's no other clear way to get there.

muellerzr:
@byi8220 can you resolve the PR's and then I think we're okay to merge this.

muellerzr:
As a final step, we likely want to update `save_state/load_state` to resume the dataloaders at this point.

src/accelerate/accelerator.py
```diff
@@ -1593,9 +1597,9 @@ def _prepare_deepspeed(self, *args):
 
         deepspeed_plugin = self.state.deepspeed_plugin
 
-        is_dataloader_present = any(isinstance(obj, torch.utils.data.DataLoader) for obj in args)
+        is_dataloader_present = any((isinstance(obj, torch.utils.data.DataLoader)) for obj in args)
         result = [
-            self._prepare_one(obj, first_pass=True) if isinstance(obj, torch.utils.data.DataLoader) else obj
+            self._prepare_one(obj, first_pass=True) if (isinstance(obj, torch.utils.data.DataLoader)) else obj
```
BenjaminBossan, line 1602: These added parentheses are not required.

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,65 @@ def end(self):
...
+    compatability reasons, this class inherits from the class it wraps around, so it can be used as a drop-in.
+    """
+
+    def __init__(self, dataset, use_stateful_dataloader=False, batch_sampler=None, **kwargs):
+        self.use_stateful_dataloader = use_stateful_dataloader
+        if is_torchdata_stateful_dataloader_available():
+            from torchdata.stateful_dataloader import StatefulDataLoader
+
+        if use_stateful_dataloader and not is_torchdata_stateful_dataloader_available():
+            raise ImportError("StatefulDataLoader is not available. Please install torchdata to use it.")
```
BenjaminBossan, line 404: Indicate min version of torchdata.

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,65 @@ def end(self):
...
+        # This code is incredibly awkward but it's the only way to make `isinstance(obj, StatefulDataLoader)` work
+        # transparently.
+        #
+        # A more robust solution is for DataLoaderAdapter to not inherit from DataLoader (compose rather than inherit),
+        # but this would not be backwards compatible with existing code which assumes
+        # DataLoaderShard/DataLoaderDispatcher are DataLoaders.
+        base_cls = self.__class__
+        base_cls_name = self.__class__.__name__
+        parent_cls_name = self.base_dataloader.__class__
+        self.__class__ = type(base_cls_name, (base_cls, parent_cls_name), {})
```
BenjaminBossan, line 427: Let me just bring up (again) that another solution could be monkey-patching `__instancecheck__` on `DataLoader`. Not saying that it's less hacky, just wanted to raise awareness :)

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,65 @@ def end(self):
...
+
+        # Allow this class to transparently pass through attributes from the underlying class
+        if hasattr(self.base_dataloader, "state_dict"):
+            self.dl_state_dict = self.base_dataloader.state_dict()
+
+        for attr in self.base_dataloader.__dict__.keys():
+            setattr(self, attr, getattr(self.base_dataloader, attr))
```
BenjaminBossan, line 434: Kinda looks dangerous. For example, this skips `@property`, is that intended? We could instead use `__getattr__` to dispatch to `self.base_dataloader`.

If we want to stick this this, more succinct code could be: `self.__dict__.update(self.base_loader.__dict__)` or `vars(self).update(self.base_loader.__dict__)`

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,65 @@ def end(self):
...
+
+    def state_dict(self):
+        return self.dl_state_dict
+
+    def load_state_dict(self, state_dict):
+        super().load_state_dict(state_dict)
+        self.dl_state_dict = self.state_dict
+
+    def _save_state_dict(self):
```
BenjaminBossan, line 443: IMO, the name is not quite fitting, isn't it more like `update_state_dict` or so? Also, maybe we can avoid this all by not having a static `self.dl_state_dict` attribute but instead the `state_dict` method just returns `self.base_dataloader.state_dict()`.

src/accelerate/data_loader.py
```diff
@@ -428,11 +487,12 @@ def __init__(
         rng_types=None,
         synchronized_generator=None,
         skip_batches=0,
+        use_stateful_dataloader=False,
         _drop_last: bool = False,
         _non_blocking: bool = False,
         **kwargs,
     ):
-        super().__init__(dataset, **kwargs)
+        super().__init__(dataset, use_stateful_dataloader, **kwargs)
```
BenjaminBossan, line 495: Let's pass `use_stateful_dataloader=use_stateful_dataloader`.

src/accelerate/data_loader.py
```diff
@@ -601,7 +665,7 @@ def __init__(
             # We need to save the shuffling state of the DataPipe
             if isinstance(dataset, ShufflerIterDataPipe):
                 shuffle = dataset._shuffle_enabled
-        super().__init__(dataset, **kwargs)
+        super().__init__(dataset, use_stateful_dataloader, **kwargs)
```
BenjaminBossan, line 668: Let's pass `use_stateful_dataloader=use_stateful_dataloader`.

src/accelerate/data_loader.py
```diff
@@ -868,6 +935,10 @@ def prepare_data_loader(
         non_blocking (`bool`, *optional*, defaults to `False`):
             If set to `True`, dataloader will utilize non-blocking host-to-device transfers. If the dataloader has
             `pin_memory` set to `True`, this will help to increase overlap between data transfer and computations.
+        use_stateful_dataloader (`bool`, *optional*, defaults to `False`):
+            "If set to true, the dataloader prepared by the Accelerator will be backed by "
+            "[torchdata.StatefulDataLoader](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader).
+            This requires a version" " of `torchdata` with StatefulDataLoader to be installed."
```
BenjaminBossan, line 941: `version" "`?

BenjaminBossan (commented):
Thanks a lot for the PR and the good discussion of possible designs. The end result is still something that I'm afraid will one day cause a hard to debug issue, but I can't say what exactly would be a better solution.

I added a couple of comments, which I think can help clean up this PR a bit, but don't consider them to be blockers. I have to admit I only skimmed the tests but they look very well done, so together with the existing ones they should hopefully avoid regressions.

One thing I would like to see is an addition to the docs to explain what stateful data loaders are, why users may want to use them, and how they can use them.

src/accelerate/data_loader.py
byi8220: Done, I will once again warn that this depends on the nightly build of torchdata.

src/accelerate/data_loader.py
byi8220: > Kinda looks dangerous. 

Kinda agree with you, but all dynamic reflection looks dangerous to me. 

I did write up an alternative which avoids the wizardry and just duplicates all the code required over here in: https://github.com/byi8220/accelerate/compare/stateful-dataloader...byi8220:accelerate:stateful-dataloader-2 

That code is messier and involves way more duplication, but much more explicit in what it does. If enough people feel the reflection approach is way too hacky and this feature doesn't justify it, I'm fine with doing that instead.

> We could instead use __getattr__ to dispatch to self.base_dataloader.

I updated the PR to do that instead.

src/accelerate/data_loader.py
byi8220: Done

src/accelerate/data_loader.py
byi8220: > IMO, the name is not quite fitting, isn't it more like update_state_dict or so? 

Changed to `_update_state_dict`

> Also, maybe we can avoid this all by not having a static self.dl_state_dict attribute but instead the state_dict method just returns self.base_dataloader.state_dict().

I'm not sure if we can. The base dataloader's state dict is one ahead of what we're yielding, so we couldn't do a passthrough. Some additional context in https://github.com/huggingface/accelerate/pull/2991/commits/a6e192c02dc57fb7cc4bba2152a22b7a68a8f5e3#r1704736815

src/accelerate/data_loader.py
byi8220: Fixed

byi8220:
@BenjaminBossan Thanks for the review. Just addressed the comments on the PR.

>  The end result is still something that I'm afraid will one day cause a hard to debug issue, but I can't say what exactly would be a better solution.

The ultimate intent of this code is something like "Sometimes I want a DataLoaderDispatcher that inherits from DataLoader, but other times I want a DataLoaderDispatcher that inherits from StatefulDataLoader."

Imo, the less magical alternative would be to explicitly duplicate each DataLoader derivative that accelerate defines into a stateful version. I.e. manually create the classes StatefulDataLoaderDispatcher, StatefulDataLoaderShard, StatefulSkipDataLoader. I wrote up this alternative in a separate branch (diffed by https://github.com/byi8220/accelerate/compare/stateful-dataloader...byi8220:accelerate:stateful-dataloader-2), but it leads to quite a lot of code duplication and also looks messy.

> I have to admit I only skimmed the tests but they look very well done, so together with the existing ones they should hopefully avoid regressions.

I've tested this locally on my 1 GPU home workstation + a 2xGPU cloud instance (which costs me a few dollars every time I want to run the test suite :disappointed: ...)
The fact that all tests pass for me when not using this feature, regardless of if the required torchdata version is not installed gives confidence that it's not causing a breaking regression.

This is my first real PR into `accelerate`, so I added the sanity and happy test cases I could think of based on my limited context.

The tests highlighted one small thing though, the fact that to fully stop using a dataloader in the middle you have to call `dataloader.end()`, but this might just be unavoidable. If the use case of StatefulDataLoader is to restart the entire program from a checkpoint maybe it's not a big issue: https://github.com/huggingface/accelerate/pull/2895/files#diff-68b278b14afa2e1ea337bb5e13d122f6d074c8bf0f0b83bef779eac6f4ba7f9aR724-R726

src/accelerate/data_loader.py
byi8220: Just checked, it turns out torchdata officially supports this feature from version 0.8.0 onwards: https://github.com/pytorch/data/releases/tag/v0.8.0

Updated docs for this, can probably officially support this without some nightly hacks now

muellerzr:
> The tests highlighted one small thing though, the fact that to fully stop using a dataloader in the middle you have to call dataloader.end(), but this might just be unavoidable. If the use case of StatefulDataLoader is to restart the entire program from a checkpoint maybe it's not a big issue
I believe this has been a known "issue" in accelerate (I've seen it pop up in other issues sparingly). Agreed that it's less of an issue here, since this is pretty much just called once at the start of training. As long as we have the state properly (which your tests check!) it's a different bug/issue to solve

> Imo this might be better in a separate PR, once the code is checked in?
We tend to like full FC PR's that include doc updates. Less likely it'll be forgotten about and it's done all at once so users who want the bleeding edge can read immediately :)

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,68 @@ def end(self):
...
+        parent_cls_name = self.base_dataloader.__class__
+        self.__class__ = type(base_cls_name, (base_cls, parent_cls_name), {})
+
+        # Allow this class to transparently pass through attributes from the underlying class
+        if hasattr(self.base_dataloader, "state_dict"):
+            self.dl_state_dict = self.base_dataloader.state_dict()
+
+    def __getattr__(self, name):
+        # Delegate attribute access to the internal dataloader
+        return getattr(self.base_dataloader, name)
```
BenjaminBossan, line 437: A bit of an edge case: Let's also check if the name is not `"base_dataloader"`, and if it is to raise an `AttributeError`, to avoid an infinite recursion.

BenjaminBossan (approved):
Thanks for the updates.

> I'm not sure if we can. The base dataloader's state dict is one ahead of what we're yielding, so we couldn't do a passthrough. Some additional context in the comments of https://github.com/huggingface/accelerate/commit/a6e192c02dc57fb7cc4bba2152a22b7a68a8f5e3#r1704736815

I see. If Zach is fine with the proposed solution, then we're good.

> Well, I have no idea how to solve such a problem in python. In the C++ world this is what destructors and RAII are for, I guess.

There is the `__del__` magic method in Python but let's not touch it.

src/accelerate/data_loader.py
byi8220: Could you give a code example of how infinite recursion would happen here?

If I'm reading the python3 docs for [`__getattr__()`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__) correctly, it states "Note that if the attribute is found through the normal mechanism, `__getattr__()` is not called." IIUC, `base_dataloader` should always be retrievable through the normal mechanism.

If I add the following block into `test_dataloader_inheritance()` in `test_data_loader.py` (without making any changes), the tests pass without causing an infinite recursion:

```
        assert isinstance(skip_dl.base_dataloader, DataLoader)
        assert isinstance(dl_shard.base_dataloader, DataLoader)
        assert isinstance(dl_dispatcher.base_dataloader, DataLoader)
```

byi8220:
> I see. If Zach is fine with the proposed solution, then we're good.

sgtm

> There is the __del__ magic method in Python but let's not touch it.

I see. Destructors in python don't seem very reliable, but my knowledge of the python memory model isn't great.

src/accelerate/data_loader.py
BenjaminBossan: > Could you give a code example of how infinite recursion would happen here?

Yes, that would be for the edge case of an attribute getting called on the _class_, i.e. before it is instantiated. In that case, the `base_dataloader` attribute does not exist. Now you could say "who would do such a pernicious thing?", but it's a bug that actually happened in another project and for some reason DeepSpeed would do this (on a module, not a data loader, but let's rather be safe than sorry).

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,68 @@ def end(self):
...
+        #
+        # A more robust solution is for DataLoaderAdapter to not inherit from DataLoader (compose rather than inherit),
+        # but this would not be backwards compatible with existing code which assumes
+        # DataLoaderShard/DataLoaderDispatcher are DataLoaders.
+        base_cls = self.__class__
+        base_cls_name = self.__class__.__name__
+        parent_cls_name = self.base_dataloader.__class__
+        self.__class__ = type(base_cls_name, (base_cls, parent_cls_name), {})
+
+        # Allow this class to transparently pass through attributes from the underlying class
```
BenjaminBossan, line 431: This comment does not fit here anymore, right?

src/accelerate/utils/imports.py
```diff
@@ -431,3 +431,16 @@ def is_xpu_available(check_device=False):
 
 def is_dvclive_available():
     return _is_package_available("dvclive")
+
+
+def is_torchdata_available():
+    return _is_package_available("torchdata")
+
+
+# TODO: Remove this function once stateful_dataloader is a stable feature in torchdata.
```
BenjaminBossan, line 440: Can this now be adjusted to use a version check?

tests/test_accelerator.py
```diff
@@ -73,6 +87,23 @@ def forward(self, x):
         return self.linear2(self.batchnorm(self.linear1(x)))
 
 
+def create_dataloaders_for_test(
+    a=2, b=3, batch_size=3, n_train_batches: int = 12, n_valid_batches: int = 2, num_workers=0
```
BenjaminBossan, line 91: What is with the `a` and `b` parameters? Can they be removed?

src/accelerate/data_loader.py
```diff
@@ -388,9 +389,68 @@ def end(self):
...
+
+    def state_dict(self):
+        return self.dl_state_dict
+
+    def load_state_dict(self, state_dict):
+        super().load_state_dict(state_dict)
+        self.dl_state_dict = self.state_dict
+
+    def _update_state_dict(self):
+        if hasattr(self.base_dataloader, "state_dict"):
```
BenjaminBossan, line 447: Let's add a comment here when this needs to be called and with the context on why it's required.

tests/test_accelerator.py
```diff
@@ -615,3 +651,134 @@ def test_can_unwrap_model(self):
...
+    @parameterized.expand(
+        itertools.product([True, False], [True, False], [0, 2], [True, False]),
+        name_func=parameterized_custom_name_func,
+    )
+    @require_torchdata_stateful_dataloader
+    def test_save_model_with_stateful_dataloader(self, use_safetensors, tied_weights, num_workers, dispatch_batches):
+        """
+        Test that saving and loading a model with a stateful dataloader returns the same model,
+        and that the dataloader's iterator is restored properly."""
+        print()
```
BenjaminBossan, line 688: remove?

BenjaminBossan (commented):
Did a final review and also more carefully reviewed the tests. I didn't find anything big, but a few minor things that could be improved. After that, this can be merged from my POV.

src/accelerate/data_loader.py
byi8220: Added a comment here, kinda clunky though.

tests/test_accelerator.py
byi8220: I think that's leftover from a different approach, removed

tests/test_accelerator.py
```diff
@@ -35,9 +37,22 @@
...
+    require_torchdata_stateful_dataloader,
+)
+from accelerate.utils import FP8RecipeKwargs, is_torchdata_stateful_dataloader_available, patch_environment
+from accelerate.utils.dataclasses import DataLoaderConfiguration
 from accelerate.utils.modeling import get_state_dict_from_offload, load_checkpoint_in_model
+from accelerate.utils.random import set_seed
+
+
+if is_torchdata_stateful_dataloader_available():
+    from torchdata.stateful_dataloader import (
```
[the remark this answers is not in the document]
byi8220: Done

src/accelerate/utils/imports.py
byi8220: Yes, modeled after other version checks in file

src/accelerate/utils/dataclasses.py
```diff
@@ -777,6 +777,14 @@ class DataLoaderConfiguration:
             " prepared dataloader has `pin_memory` set to `True` to work properly."
         },
     )
+    use_stateful_dataloader: bool = field(
+        default=False,
+        metadata={
+            "help": "If set to `True`, the dataloader prepared by the Accelerator will be backed by "
+            "[torchdata.StatefulDataLoader](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader). This requires a version"
+            " of `torchdata` with StatefulDataLoader to be installed."
```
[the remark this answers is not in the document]
byi8220: Done

src/accelerate/data_loader.py
byi8220: Removed

byi8220:
Thanks!

> I didn't find anything big, but a few minor things that could be improved. 

I fixed the nits above, but I also made one, maybe important, change, done in https://github.com/huggingface/accelerate/pull/2895/commits/74e2f53d841f5701d476f3fe5f6df8f97ad82e5c

Basically, I literally realized just now that I have been delegating the work of iteration to the superclass, instead of the backing dataloader. That felt wrong, so I did the commit above.

To confirm, replacing `super()->self.base_dataloader` here is the sensible thing to do, right? Like I want to fully delegate everything I can to the base_dataloader, and the only reason it worked as written before is because of getattr spaghetti?

src/accelerate/data_loader.py
byi8220: I am still not entirely sure how this could happen, but I added a check in `__getattr__`.

muellerzr (approved):
Very nicely done! 🤩 

Thanks so much for all your hard work on this, let's get it merged in ✅ 

(And yes, I think it is a sensible thing to do rather than the getattr spaghetti)

muellerzr:
as a next step I'll work on getting this working with `Accelerator.save_state`/`Accelerator.load_state` today

PR #2909, opened 2024-07-04
[tests] fix bug in torch_device

faaany:
## What does this PR do?
Fix the bug introduced in #https://github.com/huggingface/accelerate/pull/2879/files. I reverted the change I made in that PR because once our PyTorch PR is merged, it will work. 

@muellerzr


muellerzr (approved):
Sounds good, thanks!

PR #2927, opened 2024-07-11
delete CCL env var setting

Liangliang-Ma:
This PR deletes some CCL environment variable setting. With versions updating of software stack, current torch-CCL don't recommend manually setting these CCL env vars like 'CCL_LOCAL_SIZE', 'CCL_LOCAL_RANK' or 'CCL_PROCESS_LAUNCHER'.

And  `os.environ.get("LOCAL_WORLD_SIZE", "1")
os.environ.get("LOCAL_RANK", "0")`
may not get the correct value. For example, when run fine-tuning with deepspeed on multiple gpus, it will get wrong value and pass them to torchCCL comm backend, which will result in crash.

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -194,12 +194,6 @@ class PartialState:
                                 "DeepSpeed is not available => install it using `pip3 install deepspeed` or build it from source"
                             )
                         from deepspeed import comm as dist
-
-                        if is_xpu_available() and is_ccl_available():
-                            os.environ["CCL_PROCESS_LAUNCHER"] = "none"
-                            os.environ["CCL_LOCAL_SIZE"] = os.environ.get("LOCAL_WORLD_SIZE", "1")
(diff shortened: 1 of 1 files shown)


Liangliang-Ma:
@muellerzr @faaany pls help to review. Thanks!

faaany:
yes, I can verify that multi-card fine-tuning works without these CCL environment variables using the latest torch-ccl&ipex. They were added at a time when users had to handle them manually. But now we can remove them. Pls have a review, thx! @muellerzr

faaany:
@yao-matrix

muellerzr (approved):
Thanks for the cleanup!

muellerzr:
Can you run `make style; make quality`? (Don't worry about the import test)

Liangliang-Ma:
@muellerzr Thanks. Format fixed.

PR #2936, opened 2024-07-15
Consider pynvml available when installed through the nvidia-ml-py distribution

matthewdouglas:
# What does this PR do?

This PR will enable detection of the `pynvml` package when installed via [`nvidia-ml-py`](https://pypi.org/project/nvidia-ml-py/).

@muellerzr


muellerzr (approved):
Nice corner case!

PR #2945, opened 2024-07-18
Add `torch.float8_e4m3fn` format `dtype_byte_size`

SunMarc:
# What does this PR do?
This PR adds the `torch.float8_e4m3fn` in `dtype_byte_size` function, so that we can calculate the memory required for this specific datatype

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -161,7 +161,7 @@ def dtype_byte_size(dtype: torch.dtype):
         return 1 / 4
     elif dtype == CustomDtype.INT4:
         return 1 / 2
-    elif dtype == CustomDtype.FP8:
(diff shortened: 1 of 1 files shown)


muellerzr (commented):
Nice! Note we need a check here, this dtype was only added in torch 2.1.0: https://github.com/pytorch/pytorch/commit/b64bd4a5dd9ee21d269997c4a28f6389870cd540

SunMarc:
> Nice! Note we need a check here, this dtype was only added in torch 2.1.0: https://github.com/pytorch/pytorch/commit/b64bd4a5dd9ee21d269997c4a28f6389870cd540

muellerzr (approved):
Nice :) LG2M post quality pass ;)

PR #2964, opened 2024-07-29
Run compiled model once after torch.compile to exclude the compile time from measurement

zxd1997066:
This PR mainly focus on excluding the compilation time from evaluation loop.
- Add `torch.no_grad()` before `torch.compile` to avoid recompilation due to different grad mode.
- Run the model with example inputs after `torch.compile` to do the compilation before going to the evaluation loop.


BenjaminBossan:
> This PR mainly focus on excluding the compilation time from evaluation loop.

Could you please explain a bit further what exactly the problem is that you're trying to solve with this PR?

zxd1997066:
> > This PR mainly focus on excluding the compilation time from evaluation loop.
> 
> Could you please explain a bit further what exactly the problem is that you're trying to solve with this PR?

Hi, thanks for the comment. We planned to add a warmup, but it seems not appropriate to modify here. Close the PR temporarily.

PR #2982, opened 2024-07-31
Fix torchvision to be compatible with torch version in CI

SunMarc:
# What does this PR do?

This PR make sure that the CI is using a version of torch that is compatible with the version of torchvision installed. The latest torchvision requires torch>=2.4 which created an issue on our CI.


muellerzr (approved):
Thanks!

krishna1761:
Thanks

PR #2983, opened 2024-07-31
Enable FSDP & Deepspeed + FP8

muellerzr:
# Enable FSDP with FP8

## What does this add?

This PR enables users to utilize TransformersEngine FP8 with fully sharded data parallelism and DeepSpeed (1-3)

## Who is it for?

Solves https://github.com/huggingface/accelerate/issues/2970

### User-facing:

*  If no `FP8RecipeKwargs` are passed in, we go based on what is available by first prioritizing MS-AMP then `transformersengine`, letting users just pass in `--mixed_precision fp8` during the example scripts easily

* 4 new scripts have been added, which show 1:1 accelerate <-> transformers engine comparisons to ensure that performance works as best as possible. These include
  * Single GPU
  * DDP
  * FSDP
  * DeepSpeed (ZeRO 1-3)

### Internal structure:

The logic behind preparing `TransformersEngine` models has been adjusted, since how these really should be prepared is before *anything* touches FSDP/DeepSpeed/etc. As a result we have a new `Accelerator._prepare_te` similar to `Accelerator._prepare_msamp`, since we need the optimizers for us to convert the param mappings back *prior* to FSDP.

This initial implementation assumes only a single model and single optimizer. I'm unsure of TE supports multiple models, but will check. If so, this will be done as part of the multiple model deepspeed implementation/fix.

## **IMPORTANT**

A big bug was discovered, wherein evaluated accuracy is greatly decreased when using FP8 autocasting during evaluation. This is now configurable using the `use_during_evaluation` key as part of the `FP8RecipeKwargs`. It is set to `False` by default to get the best metrics. 

cc @SunMarc


muellerzr:
(And the branch name is confusing, I know. Switched over to FSDP from DS midway through)

src/accelerate/state.py
```diff
@@ -897,7 +897,10 @@ def __init__(
                 if os.environ.get("ACCELERATE_USE_FSDP", "false") == "true":
                     self.distributed_type = DistributedType.FSDP
                     if self._mixed_precision != "no":
-                        fsdp_plugin.set_mixed_precision(self._mixed_precision)
+                        mixed_precision = self._mixed_precision
+                        if self._mixed_precision == "fp8":
+                            mixed_precision = "bf16"
+                        fsdp_plugin.set_mixed_precision(mixed_precision)
```
SunMarc, line 903: Why are we setting it to bf16 ?

src/accelerate/accelerator.py
```diff
@@ -393,12 +396,11 @@ def __init__(
 
         self.delayed_fp8_autocast = False
         if self.fp8_recipe_handler is not None:
-            # We already check if FP8 is available during `self.state`
-            if self.state.mixed_precision != "fp8":
+            if self.state.mixed_precision != "fp8" and self.distributed_type != DistributedType.FSDP:
```
SunMarc, line 399: We should maybe update the error message since you added an new condition ? Or we can do this new right below with its own error message.

SunMarc (approved):
Nice ! Thanks for adding this @muellerzr. Just a few nits !

src/accelerate/state.py
muellerzr: With training in FP8, only certain layers are done in FP8 (such as linear, etc). Otherwise we run bf16 precision for everything else (gradients, backprop, etc).

muellerzr:
Okay current status:

1. We have working FSDP, DS 1 and DS 2
2. We can maintain eval accuracy by disabling the FP8 autocast manager during eval if prompted. Generally their examples show not using it during eval, and I found as well during my testings that this would reduce end accuracy. This could also explain why users were seeing poor performance during evals when using accelerate + FP8
3. We have 4 example testing scripts added which show a baseline measurement with TE vs Accelerate to make sure we can be as accurate as possible to their work

@BenjaminBossan pinged you here for review as there's quite a few design implementations that needed to be done to get all of this working, would love to know your thoughts on any

src/accelerate/utils/dataclasses.py
```diff
@@ -324,6 +326,7 @@ class FP8RecipeKwargs(KwargsHandler):
     """
 
     backend: Backend = "MSAMP"
+    use_during_eval: bool = False
```
BenjaminBossan, line 329: Strictly speaking, this change is not 100% backwards compatible, right? E.g. some existing eval code using FP8 could now OOM when previously it didn't. It's probably still fine to change it, as this is an edge case, but I wanted to flag it.

src/accelerate/utils/dataclasses.py
```diff
@@ -1077,12 +1080,12 @@ def set_mixed_precision(self, mixed_precision):
         ds_config = self.deepspeed_config
         kwargs = {
             "fp16.enabled": mixed_precision == "fp16",
-            "bf16.enabled": mixed_precision == "bf16",
+            "bf16.enabled": mixed_precision in ("bf16", "fp8"),
```
BenjaminBossan, line 1083: Why is this? Could you maybe add a comment here?

src/accelerate/utils/transformer_engine.py
```diff
@@ -82,3 +82,41 @@ def has_transformer_engine_layers(model):
...
+    import transformer_engine.common.recipe as te_recipe
+
+    kwargs = fp8_recipe_handler.to_kwargs() if fp8_recipe_handler is not None else {}
+    if "fp8_format" in kwargs:
+        kwargs["fp8_format"] = getattr(te_recipe.Format, kwargs["fp8_format"])
+    use_during_eval = kwargs.pop("use_during_eval", False)
+    fp8_recipe = te_recipe.DelayedScaling(**kwargs)
+    new_forward = contextual_fp8_autocast(model.forward, fp8_recipe, use_during_eval)
+
+    from types import MethodType
```
BenjaminBossan, line 115: Could be a root import.

src/accelerate/utils/transformer_engine.py
```diff
@@ -82,3 +82,41 @@ def has_transformer_engine_layers(model):
         if isinstance(m, (te.LayerNorm, te.Linear, te.TransformerLayer)):
             return True
     return False
+
+
+def contextual_fp8_autocast(fn, fp8_recipe, use_during_eval=False):
```
BenjaminBossan, line 87: Why "contextual" in the name?

src/accelerate/utils/transformer_engine.py
```diff
@@ -82,3 +82,41 @@ def has_transformer_engine_layers(model):
...
+
+    def inner(self, *args, **kwargs):
+        enabled = use_during_eval or self.training
+        with fp8_autocast(enabled=enabled, fp8_recipe=fp8_recipe):
+            return fn(*args, **kwargs)
+
+    return inner
+
+
+def apply_fp8_autowrap(model, fp8_recipe_handler):
```
BenjaminBossan, line 101: Is it necessary to also provide a method that unwraps the `model` (i.e. restores the `forward` method to the original)? For most use cases, I guess not, as this function basically only ensures that the fp8 autocast context is invoked, but perhaps for serialization or something like that, it could be important to unwrap? Not sure, as it seems that so far there was no such issue.

src/accelerate/accelerator.py
```diff
@@ -1578,16 +1570,46 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+            model = apply_fp8_autowrap(model, self.fp8_recipe_handler)
         # torch.compile should be called last and only if the model isn't already compiled.
         if self.state.dynamo_plugin.backend != DynamoBackend.NO and not is_compiled_module(model):
             if not is_torch_version(">=", "2.0"):
                 raise ValueError("Using `torch.compile` requires PyTorch 2.0 or higher.")
             model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
         return model
 
+    def _prepare_te(self, *args):
+        model, optimizer = None, None
```
BenjaminBossan, line 1582: Is an import check needed, same as for MS-AMP?

src/accelerate/accelerator.py
```diff
@@ -1578,16 +1570,46 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+        if optimizer is None and model is None:
+            return result
+        elif optimizer is None or model is None:
+            raise ValueError(
+                "You must pass a model and an optimizer together to `accelerate.prepare()` when using TransformerEngine."
+            )
+        elif num_models > 1 or num_optimizers > 1:
+            raise ValueError(
+                f"You can't use multiple models ({num_models}) or optimizers {num_optimizers} with TransformerEngine."
+            )
```
BenjaminBossan, line 1601: I guess this could be factored out into a separate function to avoid duplication, but it's only 1 dup so far so it's fine.

src/accelerate/test_utils/scripts/external_deps/transformerengine/ddp.py
```diff
@@ -0,0 +1,170 @@
...
+    accelerator_not_trained, accelerator_trained = train_integration()
+
+    assert (
+        baseline_not_trained["accuracy"] == accelerator_not_trained["accuracy"]
+    ), f'Accuracy should be the same for the baseline and accelerator: {baseline_not_trained["accuracy"]} == {accelerator_not_trained["accuracy"]}'
+    assert (
+        baseline_not_trained["f1"] == accelerator_not_trained["f1"]
+    ), f'F1 score should be the same for the baseline and accelerator: {baseline_not_trained["f1"]} == {accelerator_not_trained["f1"]}'
+    assert (
+        baseline_trained["accuracy"] == accelerator_trained["accuracy"]
```
BenjaminBossan, line 164: This replicates 100% the accuracy? Nice.

src/accelerate/test_utils/scripts/external_deps/transformerengine/experiment_setup.py
BenjaminBossan, line 1: Name was a bit surprising, as it's more about utility functions than setup?

src/accelerate/test_utils/scripts/external_deps/transformerengine/ddp.py
```diff
@@ -0,0 +1,170 @@
...
+def get_named_parameters(model):
+    """
+    Same thing as `Accelerator.get_named_parameters` Returns a list of the named parameters of the model (extracted
+    from parallel)
+    """
+    model = extract_model_from_parallel(model)
+    return {n: p for n, p in model.named_parameters()}
+
+
+def evaluate_model(model, dataloader, fp8_recipe=None, accelerator=None):
```
BenjaminBossan, line 47: These 2 functions are identical between scripts (except DS), right? If yes, they could be moved to `experiment_setup.py`.

BenjaminBossan (commented):
Generally, this LGTM and is is also cleaner than previously, thanks for this PR. I added some comments but none or critical, so feel free to ignore if you want.

Also well done adding the scripts to test for equivalent outputs. I guess they are not run automatically right now, so it's up to contributors to ensure that future changes don't break them. Unfortunately, I could not run examples as I got an error when trying to install TE, and the first few searches didn't help with resolving it.

src/accelerate/utils/transformer_engine.py
muellerzr: Originally it was going to be based on if the model is in `.eval()`, but I decided to switch it and give user choice.

src/accelerate/utils/dataclasses.py
muellerzr: It could but the old method was "wrong". We can make a note about it though

src/accelerate/utils/transformer_engine.py
muellerzr: I think `unwrap_model` should catch this, but I'll make a test

muellerzr: Made a test to make sure that `unwrap_model` properly gets rid of this.

src/accelerate/accelerator.py
muellerzr: Done!

src/accelerate/test_utils/scripts/external_deps/transformerengine/experiment_setup.py
muellerzr: Calling it `utils` made the lib get angry with me 😅 Let me see if I still have issues though when I move all the files over

PR #2989, opened 2024-08-04
Add missing docs for two KwargsHandlers

salcc:
TorchDynamoPlugin and ‎GradientAccumulationPlugin were not included in the documentation.

@muellerzr


salcc:
Just saw that they are documented here https://huggingface.co/docs/accelerate/en/package_reference/utilities

Issue #2923, opened 2024-07-07
KeyError: 'step' when resume from checkpoint


kxhit:
### System Info

```Shell
accelerate 0.32.0
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

My training script works fine with accelerate==0.23.0 and when using 0.32.0, and resume from checkpoint (saved by 0.32.0 version as well), I got an error 

"accelerate/accelerator.py", line 3147, in load_state
    self.step = override_attributes["step"]
KeyError: 'step'" 

### Expected behavior

I believe this [line](https://github.com/huggingface/accelerate/blob/2471eacdd648446f2f63eccb9b18a7731304f401/src/accelerate/accelerator.py#L3147) causes the error and in accelerate==0.23.0, there is no "step". 

Hope to get some suggestion in avoiding this bug or get it fixed.

I downgraded my accelerate to 0.31.0 and it works.

alexanderswerdlow:
For anyone who has a similar issue, I also encountered this loading of the internal step to be problematic. Specifically, after adding a try/catch, I found that it succeeds on the master rank but not other ranks. In turn, this causes the ranks to become out of sync, in my case with different amounts of gradient accum in the first step. Ultimately, this can result in a hang later on.

rahji:
I'm having the same problem. How did you downgrade accelerate to 0.31.0, @kxhit ?

BenjaminBossan:
Thanks for reporting, as correctly stated, downgrading accelerate is the correct workaround.

This was most likely caused by #2765. IMO it would be best if checkpoints were compatible between accelerate versions, so ideally there is a fix that makes the `step` key optional to have. Let's see what @muellerzr thinks about this when he's back in office.

priyammaz:
I just build my environment so i was running the newest 0.33.0 accelerate version. I saved a checkpoint with this version, when when trying to load it, it throws the key error for "step". I downgraded to 0.31 and its totally fine now, but just thought id mention that even within the same version of accelerate there may be a slight issue.

rbli-john:
May I ask is there any plan to fix this issue?

simonhessner:
[...]

PR #2992, opened 2024-08-06
Explicit check for `step` when loading the state

muellerzr:
# What does this PR do?

To not break prior releases, checks if `step` exists in our checkpoint to try and load it in.

Fixes https://github.com/huggingface/accelerate/issues/2923

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -3159,7 +3159,8 @@ class Accelerator:
             map_location,
             **load_model_func_kwargs,
         )
-        self.step = override_attributes["step"]
+        if "step" in override_attributes:
+            self.step = override_attributes["step"]
         custom_checkpoints = [
             f for f in os.listdir(input_dir) if re.search(r"^custom_checkpoint_\d+\.pkl$", f) is not None
         ]
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -252,7 +252,8 @@ def load_accelerator_state(
     # Random states
     try:
         states = torch.load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"))
-        override_attributes["step"] = states["step"]
+        if "step" in states:
+            override_attributes["step"] = states["step"]
         random.setstate(states["random_state"])
         np.random.set_state(states["numpy_random_seed"])
         torch.set_rng_state(states["torch_manual_seed"])


BenjaminBossan (commented):
For my understanding, if a user loads an older checkpoint, they won't have `step`, so `override_attributes` will just be an empty dict, right? Would this now fail?

https://github.com/huggingface/accelerate/pull/2765/files#diff-2d7515874eaecac2687c7fc1a9c720be53f802bf14b4c3dcebe14ad443d075dcR3114

muellerzr:
Yes indeed, thanks

BenjaminBossan (approved):
Thanks for fixing this.

alexanderswerdlow:
Just wanted to comment on this since I believe this PR has a bug [or can cause a bug/unexpected behavior in some cases]. I tried this exact fix on a local fork and found that this caused issue where the master rank had the step saved but other ranks didn't. I have a suspicion it is because I was gating my `state_state` to only occur on the main process which was fine for DDP, but not in this case, see my comment [here](https://github.com/huggingface/accelerate/issues/2923#issuecomment-2231537097). In my case, a gather worked to make sure all ranks agreed, but this took a few hours of debugging to figure out this was needed:

```
            starting_steps = gather_object([self.accelerator.step])
            print(f"Starting steps: {starting_steps}")
            if not all([x > 0 for x in starting_steps]):
                print(f"Not all ranks have >0 step, setting to: {starting_steps[0]}")
                self.accelerator.step = starting_steps[0]
```

muellerzr:
Oh that's very interesting. Would you like to make a follow-up with this solution @alexanderswerdlow ? (Interesting they got out of sync like that). Otherwise I can do so shortly

alexanderswerdlow:
@muellerzr Apologies, I don't have a ton of bandwidth atm, but happy to test it. It unexpected to say the least and took a sec to find because steps being out of sync doesn't always matter!

Not sure how problematic a silent internal gather is inside accelerate [or if this is common in accelerate], but I think a warning would also suffice if the step isn't found [since this gather behavior may not be the most robust in all cases].

simonhessner:
Has this PR already been released in a new version? I am using 0.33.0 and I get `KeyError: 'step'`

ReppiksProductions:
@muellerzr How exactly do PR work if I need this fix?  Since this was a simple change in two scripts I changed them manually.  Is there a way to automatically do this from the command line?

I'm on a Windows 11 system

Issue #2973, opened 2024-07-29
Make plugin classes less environmental-variable focused, let users have an easier time just using `torchrun`, etc if they want


muellerzr:
E.g. just let the user do:

```python
from accelerate import Accelerator
from accelerate.utils import FullyShardedDataParallelPlugin

plugin = FullyShardedDataParallelPlugin(...)
accelerator = Accelerator(fsdp_plugin=plugin)
```
(Currently we have much that needs to be set via environmental variables/`accelerate config`/`accelerate launch`

PR #2998, opened 2024-08-07
Make env variables optional for FSDP

muellerzr:
# Remove env variable requirement for FSDP

## What does this add?

This PR removes the need to lock the user into using `accelerate launch` when using FSDP + accelerate. Instead, the user can now fully create a `FullyShardedDataParallelPlugin` manually.

## Who is it for?

Users of accelerate who want the flexibility of FSDP without needing `accelerate launch`

Closes https://github.com/huggingface/accelerate/issues/2973

## Why is it needed?

Limiting users to using global variables long-term is not good because it hides much of what's going on in the env, with no way to pull them down easily.

### User-facing:

A user can now use `FullyShardedDataParallelPlugin` manually:

```python
from accelerate import FullyShardedDataParallelPlugin
	strategy="FULL_SHARD"
)
```

### Internal structure:

This was an entire refactor to the internals of the dataclass, allowing for every variable to be passed in, new documentation, and new checks/type hints towards what's happening underneath. 

If a user wants to avoid using `accelerate launch`, or knows they just want FSDP and manually set it, this avoids needing (too many) hacky env variables

Still todo, we require users to still have the env variable set for doing efficient loading. I need to test it still, but I intend on making a util such as `from accelerate.utils import enable_fsdp_low_cpu_mem` to trigger this flag, which lets us then set and use it inside of `transformers` manually if a user passes it in/wants to utilize it *without* needing to set it first in the env.

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -29,6 +29,7 @@ from datetime import timedelta
 from typing import Any, Callable, Dict, Iterable, List, Literal, Optional, Tuple, get_args
 
 import torch
+from torch.distributed.fsdp.api import ShardingStrategy, BackwardPrefetch, CPUOffload, StateDictType
 
 from .constants import FSDP_AUTO_WRAP_POLICY, FSDP_BACKWARD_PREFETCH, FSDP_SHARDING_STRATEGY, FSDP_STATE_DICT_TYPE
 from .environment import str_to_bool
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
-            "for reduced memory usage."
+            "for reduced memory usage. (`False` by default)"
+        },
+    )
+    ram_efficient_loading: bool = field(
+        default=None,
+        metadata={
+            "help": "If True, only the first process loads the pretrained model checkoint while all other processes have empty weights. "
+            "Only applicable for 🤗 Transformers. When using this, `sync_module_states` needs to be `True`"
+        },
```
SunMarc, line 1305: specify default value

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
     use_orig_params: bool = field(
-        default=True,
-        metadata={
-            "help": "If `True`, allows non-uniform `requires_grad` during init, which means support for interspersed frozen and trainable parameters. "
-            "Useful in cases such as parameter-efficient fine-tuning. "
-            "Please refer this [blog](https://dev-discuss.pytorch.org/t/rethinking-pytorch-fully-sharded-data-parallel-fsdp-from-first-principles/1019). "
-            "This also enables multiple optimizer param groups. This should be `True` when creating an optimizer object before preparing/wrapping the model with FSDP."
-        },
+        default=None,
+        metadata={"help": "Whether to use the original parameters for the optimizer. Should be a `bool`."},
```
SunMarc, line 1268: specify default value

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
+            "help": "Whether to have FSDP explicitly prefetches the next upcoming "
+            "all-gather while executing in the forward pass. only use with Static graphs. Defaults to `False`"
         },
     )
     activation_checkpointing: bool = field(
-        default=False,
+        default=None,
         metadata={
-            "help": "If True, activation checkpointing is a technique to reduce memory usage by clearing activations of "
+            "help": "A technique to reduce memory usage by clearing activations of "
```
SunMarc, line 1295: specify default value

SunMarc (approved):
Nice refactor @muellerzr ! This looks good on my side ! The only nit I have and maybe future users is that we don't really know the default value for each of these param unless we dig inside the `post_init` ? Maybe it can make sense to tell the user which will be the default used if the variable is to set + env variable not set.

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
+        if isinstance(self.backward_prefetch, str):
+            if self.backward_prefetch.upper() == "NO_PREFETCH":
+                self.backward_prefetch = None
+            else:
+                if self.backward_prefetch in FSDP_BACKWARD_PREFETCH:
+                    self.backward_prefetch = FSDP_BACKWARD_PREFETCH.index(self.backward_prefetch.upper()) + 1
+                if isinstance(self.backward_prefetch, int) or self.backward_prefetch.isdigit():
+                    self.backward_prefetch = BackwardPrefetch(int(self.backward_prefetch))
+                else:
+                    self.backward_prefetch = BackwardPrefetch[self.backward_prefetch.upper()]
```
BenjaminBossan, line 1357: Could this be simplified or am I missing something?

```suggestion
            self.backward_prefetch = os.environ.get(env_prefix + "BACKWARD_PREFETCH", None)
        if isinstance(self.backward_prefetch, str):
            if self.backward_prefetch in FSDP_BACKWARD_PREFETCH:
                self.backward_prefetch = FSDP_BACKWARD_PREFETCH.index(self.backward_prefetch.upper()) + 1
            if isinstance(self.backward_prefetch, int) or self.backward_prefetch.isdigit():
                self.backward_prefetch = BackwardPrefetch(int(self.backward_prefetch))
            else:
                self.backward_prefetch = BackwardPrefetch[self.backward_prefetch.upper()]
```

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
-        self.activation_checkpointing = str_to_bool(os.environ.get(prefix + "ACTIVATION_CHECKPOINTING", "False")) == 1
-
-        if str_to_bool(os.environ.get("FSDP_CPU_RAM_EFFICIENT_LOADING", "False")) == 1 and not self.sync_module_states:
+            self.state_dict_type = os.environ.get(env_prefix + "STATE_DICT_TYPE", "FULL_STATE_DICT")
+        if isinstance(self.state_dict_type, str):
+            if self.state_dict_type.isdigit():
+                self.state_dict_type = StateDictType(int(self.state_dict_type))
+            else:
+                self.state_dict_type = StateDictType[self.state_dict_type.upper()]
+        self.set_state_dict_type()
```
BenjaminBossan, line 1366: What does this do?

matthewdouglas (approved):
LGTM!

src/accelerate/utils/dataclasses.py
```diff
@@ -1428,6 +1460,80 @@ def set_state_dict_type(self, state_dict_type_policy):
...
+            ",".join(model._no_split_modules) if no_split_modules is not None else ""
+        )
+        if self.auto_wrap_policy == transformer_auto_wrap_policy:
+            if self.transformer_cls_names_to_wrap is None:
+                self.transformer_cls_names_to_wrap = default_transformer_cls_names_to_wrap
+            transformer_cls_to_wrap = set()
+            for layer_class in self.transformer_cls_names_to_wrap:
+                transformer_cls = get_module_class_from_name(model, layer_class)
+                if transformer_cls is None:
+                    raise Exception(f"Could not find the transformer layer class {layer_class} in the model.")
```
BenjaminBossan, line 1482: Maybe `ValueError` or `TypeError`?

src/accelerate/utils/dataclasses.py
```diff
@@ -1428,6 +1460,80 @@ def set_state_dict_type(self, state_dict_type_policy):
...
+            # If zero, we silently ignore it.
+            if self.min_num_params > 0:
+                self.auto_wrap_policy = functools.partial(self.auto_wrap_policy, min_num_params=self.min_num_params)
+            else:
+                self.auto_wrap_policy = None
+
+    def set_mixed_precision(self, mixed_precision=None, buffer_autocast=False, override=False):
+        "Sets the mixed precision policy for FSDP"
+        if mixed_precision is None and self.mixed_precision_policy is None:
+            raise ValueError("Either `mixed_precision` or `mixed_precision_policy` must be provided.")
```
BenjaminBossan, line 1499: Could this raise an error where previously there was none?

BenjaminBossan (commented):
Nice rewrite, this is more readable and better documented than previously. I didn't check each and every line to verify the logic, but this should be covered by the tests. Still I have a few comments, but nothing major.

Btw. will you rewrite all of this to use pattern matching once Python 3.10 is the min version? ;-)

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
         metadata={
             "help": "A config to enable mixed precision training with FullyShardedDataParallel. "
-            "The 3 flags that are set are `param_dtype`, `reduce_dtype`, `buffer_dtype`. "
-            "Each flag expects `torch.dtype` as the value. "
-            "It is of type `torch.distributed.fsdp.fully_sharded_data_parallel.MixedPrecision`."
+            "If passing in a `dict`, it should have the following keys: `param_dtype`, `reduce_dtype`, and `buffer_dtype`."
         },
     )
-    auto_wrap_policy: Optional[Callable] = field(
+    auto_wrap_policy: Optional[Union[str, Callable]] = field(
```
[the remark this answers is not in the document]
muellerzr: Personally I'm okay with this

src/accelerate/utils/dataclasses.py
```diff
@@ -1197,147 +1196,230 @@ class FullyShardedDataParallelPlugin:
...
         default=None,
-        metadata={"help": "A callable specifying a policy to recursively wrap layers with FSDP"},
+        metadata={
+            "help": "A callable or string specifying a policy to recursively wrap layers with FSDP. If a string, it must be one of `transformer_based_wrap`, `size_based_wrap`, or `no_wrap`."
```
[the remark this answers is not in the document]
muellerzr: Best I can really get for this is to have users look at one of the FSDP versions for now (no docs discussing this really well)

src/accelerate/utils/dataclasses.py
muellerzr: You're missing the fact that `NO_PREFETCH` is in `FSDP_BACKWARD_PREFETCH`

src/accelerate/utils/dataclasses.py
muellerzr: I didn't like that we hid the rest of our earlier patterns behind `set_state_dict_type` because it did a *lot* more than that, it's more of a `set_state_dict_type_and_config`. It's far reaching to get rid of it now, so for now I reverted the refactor

src/accelerate/utils/dataclasses.py
BenjaminBossan: So this should work?

```python
            self.backward_prefetch = os.environ.get(env_prefix + "BACKWARD_PREFETCH", None)
                if isinstance(self.backward_prefetch, str) and (self.backward_prefetch != "NO_PREFETCH"):
	            if self.backward_prefetch in FSDP_BACKWARD_PREFETCH:
	                self.backward_prefetch = FSDP_BACKWARD_PREFETCH.index(self.backward_prefetch.upper()) + 1
	            if isinstance(self.backward_prefetch, int) or self.backward_prefetch.isdigit():
	                self.backward_prefetch = BackwardPrefetch(int(self.backward_prefetch))
	            else:
	                self.backward_prefetch = BackwardPrefetch[self.backward_prefetch.upper()]
```

My main concern is to avoid setting `self.backward_prefetch` from `None` to `"NO_PREFETCH"` back to `None`, which is confusing.

BenjaminBossan (approved):
LGTM, just one nit about potentially simplifying some logic.

Issue #2975, opened 2024-07-29
Barebones dataloader to allow for any type of iterable object to be used. Should just handle device placement


muellerzr:

alex-jw-brooks:
This sounds cool! I'd like to try to contribute this, I'll start working on it this week 😄

alex-jw-brooks:
Hi @ muellerzr, I should hopefully have a PR up for this pretty soon! Just to clarify the goal here, is there a reason to write a new dataloader here, as opposed to adding support for custom types to the existing dataloader in Accelerate?

I assume the intent of this is to sidestep the `TypeError` that you get if you try to make a torch `DataLoader` that also causes problems if you have `prepare` a `DataLoader` through `Accelerate`, e.g.,

```python
from torch.utils.data import DataLoader

class MyIterable:
    def __init__(self, data):
        self.data = data
    def __iter__(self):
        return iter(self.data)

list(DataLoader(MyIterable(data=[1,2,3,4]), batch_size=2))
```
which throws `TypeError: object of type 'MyIterable' has no len()`, but should be able to be treated like `DataLoader([1,2,3,4], batch_size=2)` instead of throwing because it yields things that can be cast to torch tensors. Is that correct?

I think that can be done relatively easily, e.g., by adding a really thin wrapper to encapsulate the custom type as an `IterableDataset`, but I assume that could also be done to extend support for existing dataloader types without a new flag 🤔 Just want to make sure I'm on the same page if a new option + data loader is added for how device placement & options like `dispatch_batches` should behave, e.g., if `dispatch_batches=True` and `custom_types=True` 🙂

alex-jw-brooks:
Hi @muellerzr, I should hopefully have a PR up for this pretty soon! Just to clarify the goal here, is there a reason to write a new dataloader here, as opposed to adding support for custom types to the existing dataloader in Accelerate?

I assume the intent of this is to sidestep the `TypeError` that you get if you try to make a torch `DataLoader` that also causes problems if you have `prepare` a `DataLoader` through `Accelerate`, e.g.,

```python
from torch.utils.data import DataLoader

class MyIterable:
    def __init__(self, data):
        self.data = data
    def __iter__(self):
        return iter(self.data)

list(DataLoader(MyIterable(data=[1,2,3,4]), batch_size=2))
```
[...]

Issue #2083, opened 2023-10-25
WebLoader Support


karamavusibrahim:
### System Info

```Shell
CUDA 11.8
My environment is:
      - accelerate==0.24.0
      - braceexpand==0.1.7
      - certifi==2022.12.7
      - charset-normalizer==2.1.1
      - diffusers==0.21.4
      - filelock==3.9.0
      - fsspec==2023.4.0
      - huggingface-hub==0.17.3
      - idna==3.4
      - imageio==2.31.6
      - importlib-metadata==6.8.0
      - jinja2==3.1.2
      - lazy-loader==0.3
      - markupsafe==2.1.2
      - mpmath==1.3.0
      - natsort==8.4.0
      - networkx==3.0
      - numpy==1.24.1
      - opencv-python==4.8.1.78
      - packaging==23.2
      - pandas==2.1.1
      - pillow==9.3.0
      - psutil==5.9.6
      - python-dateutil==2.8.2
      - pytz==2023.3.post1
      - pyyaml==6.0.1
      - regex==2023.10.3
      - requests==2.28.1
      - safetensors==0.4.0
      - scikit-image==0.22.0
      - scipy==1.11.3
      - six==1.16.0
      - sympy==1.12
      - tifffile==2023.9.26
      - tokenizers==0.14.1
      - torch==2.1.0+cu118
      - torchaudio==2.1.0+cu118
      - torchvision==0.16.0+cu118
      - tqdm==4.66.1
      - transformers==4.34.1
      - triton==2.1.0
      - typing-extensions==4.4.0
      - tzdata==2023.3
      - urllib3==1.26.13
      - webdataset==0.2.62
      - zipp==3.17.0
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I am trying to use webdataset and webloader in  my training code.I can run my code without any error if i am using webdataset and dataloader of torch,however if i want to use webdataset and webloader,i am getting this error:

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:1! (when checking argument for argument weight in method wrapper_CUDA___slow_conv2d_forward)

### Expected behavior

Works without any error as webdataset+dataloader configuration

[...]

PR #3008, opened 2024-08-12
Custom Types DataLoader

alex-jw-brooks:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/2975

This PR adds a `CustomTypesDataLoader`, which allows for the passing of custom iterable types (either under a PyTorch DataLoader, which would normally throw a TypeError once you start to iterate over it, or directly) to the Accelerator state's device. This adds two args to the `DataLoaderConfiguration` class:
- `custom_types: bool`; indicates whether or not an instance of this class should be created when `prepare()` is invoked  on a dataloader or iterable type
- `custom_type_batch_size: int`; batch size to be used for the `CustomTypesDataLoader`; if the iterable is already held under a PyTorch dataloader, this is optional, and the batch size will be pulled off of the data loader.

Minimal example:
```python
from accelerate.utils import DataLoaderConfiguration

class MyIterableType:
    def __init__(self, data):
        self.data = data
    
    def __iter__(self):
        return iter(self.data)

some_iterable = MyIterableType(list(range(16)))
accelerator = Accelerator(dataloader_config=DataLoaderConfiguration(custom_types=True, custom_type_batch_size=4))

# Passing DataLoader(some_iterable, batch_size=4) 
# would also be fine - if passed in this way, we could omit custom_type_batch_size
custom_type_loader = accelerator.prepare(some_iterable)
for batch in custom_type_loader:
    print(batch)
```
Running with `export ACCELERATE_TORCH_DEVICE=cuda` provides the following output:
```
tensor([0, 1, 2, 3], device='cuda:0')
tensor([4, 5, 6, 7], device='cuda:0')
tensor([ 8,  9, 10, 11], device='cuda:0')
tensor([12, 13, 14, 15], device='cuda:0')
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/concept_guides/internal_mechanism.md
+++ b/docs/source/concept_guides/internal_mechanism.md
@@ -28,7 +28,7 @@ Then, when calling [`~Accelerator.prepare`], the library:
 - wraps your model(s) in the container adapted for the distributed setup,
 - wraps your optimizer(s) in an [`~optimizer.AcceleratedOptimizer`],
 - wraps your scheduler(s) in an [`~scheduler.AcceleratedScheduler`]
-- creates a new version of your dataloader(s) in a [`~data_loader.DataLoaderShard`] or [`~data_loader.DataLoaderDispatcher`]
+- creates a new version of your dataloader(s) in a [`~data_loader.DataLoaderShard`], [`~data_loader.DataLoaderDispatcher`], or [`~data_loader.CustomTypesDataLoader`]
 
 While the model(s), optimizer(s), and scheduler(s) are just put in simple wrappers, the dataloader(s) are re-created. This is mostly
 because PyTorch does not let the user change the `batch_sampler` of a dataloader once it's been created and the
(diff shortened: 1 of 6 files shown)


BenjaminBossan:
To me, it's still not clear what we want to achieve here, even after reading the associated issue. Is the plan really to allow users to pass arbitrary iterables to `accelerator.prepare` (in which case the docstring of `prepare` needs to be updated)? Or is it that users still need to pass some type of `DataLoader` instance but that accelerate should handle this data loader instance better for arbitrary iterables inside that data loader?

alex-jw-brooks:
Good questions @BenjaminBossan! To be honest, after implementing it, I have mixed feelings too. My thoughts are:

- If the goal is to sidestep the `TypeError` that would normally be thrown when iterating over a `DataLoader` and handle the device placement - this does do that, but the way that makes sense with the current Accelerate API is a bit strange, i.e., wrapping it in a `DataLoader` that normally would not be usable, then using `prepare` to get one that is

- For passing arbitrary iterables to `accelerator.prepare()` - I have mixed feelings about that as well. I'm not entirely sure that the benefit this DataLoader adds is worth extending the API that way, and it's also kind of strange to handle in the way `_prepare_one()` gets called 🤔 

Those are mostly the reasons it's currently implemented the way it is, where you can pass it either as an iterable or as a wrapped dataloader - I think that it is interesting and could imagine it being useful somewhere, but either way feels a bit awkward with respect to the current accelerator API.

I'd be really interesting in hearing @muellerzr's thought as well. In any case, if there is a common enough use-case that it's compelling to add this feature that we can elucidate, maybe it would also be useful to add a demo somewhere showing how it's intended to be leveraged, either as part of this PR or a follow-up!

muellerzr:
Hi! Thanks so much for your first attempt at this, I had to trace back through the issues to figure out what really I was after with this "generic only place on device" dataloader, and it was from this FR for WebLoader: https://github.com/huggingface/accelerate/issues/2083

The general idea at the time was to support a generic iterable (which would be a *dataloader type class*) which we can then pipe into custom implementations easier such as WebLoader. 

IMO it should just place the outputs on the device, and have all the mixins our dataloaders have, to make it work with things like gradient accumulation properly (since that's how we check for it).

Let me know if you want to rethink your PR/simplify it at all/etc given this new information. 

It's a great effort, and I agree I did not really expound too much on the FR issue at the time. Let me know if this path forward makes sense to you both and we can continue 🚀

alex-jw-brooks:
Thanks for the guidance @muellerzr! It makes sense to try to rework this PR a bit so that we can discuss further - I think assuming that everything that should be prepared as a `DataLoader` is a DataLoader is a good idea. Then maybe the workaround @BenjaminBossan suggested in the webloader issue to debug, e.g.,

```
class MyLoader(wds.WebLoader, torch.utils.data.DataLoader):
    pass

train_dataloader = MyLoader(train_dataset, ...)
accelerator = Accelerator(device_placement=True) # maybe another flag is needed here? will check current behavior
dl = accelerator.prepare(train_dataloader)
```

could be a recommended solution for data-loader like classes that aren't instances, which is a bit gross looking, but is at least gross looking outside of accelerate. If I have the bandwidth, I'll try to reproduce the webloader issue to see what it would take for the revisions in this PR is able to fix it, it would be nice to validate its usefulness 🤞

PR #3009, opened 2024-08-13
fix default value for rank size in cpu threads_per_process assignment logic

rbrugaro:
When a distributed run is launched with `accelerate launch myscript.py` under the hood it launches mpirun. 
The logic that sets the num_of_threads_per_process for good OOB performance relies on the environment variables that are yet unassigned because mpirun is called after the environment variables are read.
```
local_size = get_int_from_env(
                ["MPI_LOCALNRANKS", "OMPI_COMM_WORLD_LOCAL_SIZE", "MV2_COMM_WORLD_LOCAL_SIZE"],
                max(args.num_processes/args.num_machines,1)
            )
threads_per_process = int(psutil.cpu_count(logical=False) / local_size)
```
https://github.com/huggingface/accelerate/blob/cd5698bb324352a80a2a2b5d8bfccf495e14bf6b/src/accelerate/commands/launch.py#L1063 

The solution is to calculate the rank per node assuming total number of processes are equally distributed across the number of machines.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -1061,7 +1061,8 @@ def _validate_launch_command(args):
         args.num_cpu_threads_per_process = get_int_from_env(["OMP_NUM_THREADS"], 1)
         if args.use_cpu and args.num_processes >= 1 and get_int_from_env(["OMP_NUM_THREADS"], 0) == 0:
             local_size = get_int_from_env(
-                ["MPI_LOCALNRANKS", "OMPI_COMM_WORLD_LOCAL_SIZE", "MV2_COMM_WORLD_LOCAL_SIZE"], max(args.num_processes/args.num_machines,1)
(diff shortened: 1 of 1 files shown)


sywangyi:
@muellerzr please help review

BenjaminBossan:
Could you please run `make style`? Also, should it be ensured that `args.num_processes/args.num_machines` is an int?

muellerzr (approved):
Thanks, overall this is fine to me I think. @BenjaminBossan num_processes and num_machines must be ints, who default to 1, so this should be okay.

rbrugaro:
@BenjaminBossan fixed style

BenjaminBossan (approved):
LGTM, thanks for the PR. There is still a linter issue, not sure why that is -- maybe a different ruff version is used in CI? Apart from that, the PR can be merged.

muellerzr:
@rbrugaro you can try `pip install -e .[quality] -U` followed by `make style; make quality`

PR #3012, opened 2024-08-14
destroy process group in `end_training`

SunMarc:
# What does this PR do ?

With the latest version of torch, when we use multi-gpu, torch will trigger a warning asking us to call `destoy_process_group()`. This PR fixes this by adding that in end_training method. Note that this function needs to be called on all process. 

For trackers, we are already only executing the methods on the main process. See [here](https://github.com/huggingface/accelerate/blob/851cf3435135d4be28a7569e61fb7c9466d68291/src/accelerate/tracking.py#L108). So it should be safe to remove the on_main_process decorator in the end_training method. 
However, I see for [WandBTracker](https://github.com/huggingface/accelerate/blob/851cf3435135d4be28a7569e61fb7c9466d68291/src/accelerate/tracking.py#L289) that `main_process_only = False`. Is there a specific reason @muellerzr ?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2678,11 +2678,10 @@ class Accelerator:
         for tracker in self.trackers:
             tracker.log(values, step=step, **log_kwargs.get(tracker.name, {}))
 
-    @on_main_process
     def end_training(self):
         """
-        Runs any special end training behaviors, such as stopping trackers on the main process only. Should always be
-        called at the end of your script if using experiment tracking.
+        Runs any special end training behaviors, such as stopping trackers on the main process only or destoying all process.
+        Should always be called at the end of your script if using experiment tracking.
 
         Example:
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan (commented):
With the removal of `@on_main_process`, would `tracker.finish()` not be called on all processes where previously that was only called on the main process?

muellerzr:
Agreed, I'll make a push to Marc's branch since he's OOO. We're missing a `if is_main_process:` flag here :)

muellerzr (approved):
Nice!

muellerzr:
All of the trackers come loaded with `@on_main_process` decorators already for their functions (if needed/applicable) so this is fine how it is/the right method: https://github.com/huggingface/accelerate/blob/main/src/accelerate/tracking.py#L571-L576

src/accelerate/accelerator.py
```diff
@@ -2678,11 +2678,10 @@ def log(self, values: dict, step: int | None = None, log_kwargs: dict | None = {
         for tracker in self.trackers:
             tracker.log(values, step=step, **log_kwargs.get(tracker.name, {}))
 
-    @on_main_process
     def end_training(self):
```
muellerzr, line 2681: Note that eventually we'll get to a point where users will want to pass in their own pg's and init them, e.g. this is needed for hybrid shard w/ FSDP. So we may need to take that in there potentially. Will see how it ends up looking

PR #3034, opened 2024-08-22
Fixup dataloader state dict bugs + incorporate load/save_state API

muellerzr:
# What does this PR do?

This PR builds on https://github.com/huggingface/accelerate/pull/2895, finishing it and solving a few issues that come up during distributed training.

1. As pointed out in `_update_state_dict`:
```python
    def _update_state_dict(self):
        # The state_dict of the underlying base_dataloader may be ahead of what is currently being yielded.
        # E.g. the implementation of DataLoaderShard involves having an underlying iterator 1 element ahead of
        # what it wants to yield.
        #
        # _update_state_dict is called to snapshot the state_dict that would properly recover the DataLoaderAdapter.
        if hasattr(self.base_dataloader, "state_dict"):
            self.dl_state_dict = self.base_dataloader.state_dict()
```
technically we're 1 ahead of what we want to yield. This causes issues later on when we want to restore, as during DDP our `_iter_yielded`, `_num_yielded`, and `samples_yielded` will be off by exactly one full batch pulled - 1 (or `num_processes -1`). With this fix in `load_state_dict`, we can properly restore the batch to the right spot. 
2. I noticed that I could only get reproducible behavior if we enabled deterministic algorithms. Unrelated to the prior PR, it's a flaky thing I noticed that goes away when enabling
3. Reworked the test to fix a few issues, namely `skip_first_batches` is meant to be used on raw pytorch dataloaders. This API technically removes the need for this (and we have tests for this elsewhere). The real test we want to ensure is if we resume from that state, the fully trained model should equal the partially trained model resumed from that state, which the tests now do.
4. Enables `save_state`/`load_state` to actually load in the dataloader states and adds tests for them
5. Ensures that this can actually run on multiple GPUs, namely by removing the part of the `skip_first_batches` since it was being used on a prepared dataloader

cc @byi8220

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc


src/accelerate/data_loader.py
```diff
@@ -539,6 +553,7 @@ def __iter__(self):
                 current_batch = next_batch
             except StopIteration:
                 self.end_of_dataloader = True
+                self._update_state_dict()
```
muellerzr, line 556: additionally, we need to update the state dict when we've hit the stop iteration to know if we've hit the end of the iterator

tests/test_accelerator.py
```diff
@@ -703,77 +703,57 @@ def test_save_model_with_stateful_dataloader(self, use_safetensors, tied_weights
...
+                accelerator.backward(loss)
+                prepared_optimizer.step()
+                prepared_scheduler.step()
+                prepared_optimizer.zero_grad()
                 batches_seen_with_loaded_dl += 1
 
-            new_linear1 = prepared_model.linear1.weight
-            new_batchnorm = prepared_model.batchnorm.weight
-            new_linear2 = prepared_model.linear2.weight
+            unwrapped_model_2 = accelerator.unwrap_model(prepared_model)
```
muellerzr, line 749: during any distributed process, we need to unwrap the model first before we can toy with the layers

src/accelerate/checkpointing.py
```diff
@@ -241,6 +246,11 @@ def load_accelerator_state(
             sampler = dataloader.get_sampler()
             if isinstance(sampler, SeedableRandomSampler):
                 sampler = dataloader.set_sampler(torch.load(input_sampler_file))
+        if getattr(dataloader, "use_stateful_dataloader", False):
+            dataloader_state_dict_name = "dl_state_dict.bin" if i == 0 else f"dl_state_dict_{i}.bin"
```
BenjaminBossan, line 250: Probably there is no way for this to happen, but I just wanted to bring it up: Could there be a situation where this condition is entered from an older checkpoint, where the dataloader checkpoint does not exist yet, thus leading to  a crash?

src/accelerate/checkpointing.py
```diff
@@ -241,6 +246,11 @@ def load_accelerator_state(
...
+            input_dataloader_state_dict_file = input_dir.joinpath(dataloader_state_dict_name)
+            state_dict = torch.load(input_dataloader_state_dict_file)
```
BenjaminBossan, line 252: Probably not something for this PR, but this line could cause trouble when the `weights_only` switch to `torch.load` will come.

src/accelerate/data_loader.py
```diff
@@ -442,8 +442,21 @@ def state_dict(self):
         return self.dl_state_dict
 
     def load_state_dict(self, state_dict):
+        # The state dict will be off by a factor of `n-1` batch too many during DDP,
+        # so we need to adjust it here
+        if PartialState().distributed_type != DistributedType.NO:
```
BenjaminBossan, line 447: Should this be fixed during loading or rather during saving?

src/accelerate/data_loader.py
```diff
@@ -1164,12 +1180,15 @@ class SkipDataLoader(DataLoaderAdapter):
     def __init__(self, dataset, skip_batches=0, use_stateful_dataloader=False, **kwargs):
         super().__init__(dataset, use_stateful_dataloader=use_stateful_dataloader, **kwargs)
         self.skip_batches = skip_batches
+        self.gradient_state = GradientState()
```
BenjaminBossan, line 1183: It would be helpful if `DataLoaderStateMixin` docstring mentioned that `self.gradient_state` is required.

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -77,12 +77,17 @@ def create_accelerator(even_batches=True):
...
+):
     """
     Create a simple DataLoader to use during the test cases
     """
+    values = torch.as_tensor(range(dataset_size))
+    if shuffle:
+        values = values[torch.randperm(values.size(0))]
     if iterable:
-        dataset = DummyIterableDataset(torch.as_tensor(range(dataset_size)))
+        dataset = DummyIterableDataset(values)
```
BenjaminBossan, line 90: Why not the same for `TensorDataset`?

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -260,6 +265,67 @@ def test_data_loader(data_loader, accelerator):
...
+    prepared_dl.load_state_dict(state_dict)
+    resumed_batches = []
+    for batch in prepared_dl:
+        resumed_batches.append(batch)
+    resumed_batches = accelerator.gather(resumed_batches)
+    for b1, b2 in zip(not_skipped_batches, resumed_batches):
+        for v1, v2 in zip(b1, b2):
+            assert torch.equal(v1, v2), f"Batch {b1} and {b2} are not equal"
+
+    accelerator.dataloader_config = old_dataloader_config
```
BenjaminBossan, line 295: Does this need to be inside a `try ... finally` if this test fails but others have yet to run?

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -260,6 +265,67 @@ def test_data_loader(data_loader, accelerator):
     ), "Not all the dataset elements have been iterated in an epoch due to duplication of samples across processes."
 
 
+def test_stateful_dataloader(accelerator):
```
BenjaminBossan, line 268: A short comment on what is tested here could be useful.

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
```diff
@@ -260,6 +265,67 @@ def test_data_loader(data_loader, accelerator):
...
+        resumed_batches.append(batch)
+    resumed_batches = accelerator.gather(resumed_batches)
+    for b1, b2 in zip(not_skipped_batches, resumed_batches):
+        for v1, v2 in zip(b1, b2):
+            assert torch.equal(v1, v2), f"Batch {b1} and {b2} are not equal"
+
+    accelerator.dataloader_config = old_dataloader_config
+
+
+def test_stateful_dataloader_save_state(accelerator):
```
BenjaminBossan, line 298: Same comments as for the test above.

tests/test_accelerator.py
```diff
@@ -703,77 +703,57 @@ def test_save_model_with_stateful_dataloader(self, use_safetensors, tied_weights
...
-            stateful_train_dl, _ = create_dataloaders_for_test(num_workers=num_workers)
-            prepared_stateful_train_dl = accelerator.prepare_data_loader(stateful_train_dl)
-            prepared_stateful_train_dl.load_state_dict(state_dict)
+                if step == num_batches_to_skip - 1:
+                    # Save the state once we've gone through a few batches
+                    accelerator.save_state(f"{tmpdirname}/state", safe_serialization=use_safetensors)
+                    # When breaking out without fully going through the iterator, must call end() to unregister this iterator from gradient state.
+                    # TODO: Maybe this could be done automatically?
+                    # prepared_train_dl.end()
+                    # break
```
BenjaminBossan, line 722: Should these be uncommented? If not, the comment is confusing.

BenjaminBossan (commented):
Thanks for investigating the issues and providing a fix, as well as harden the tests.

Generally, this LGTM, I have a few comments though, please check.

byi8220:
Thanks for catching this. Just to get an understanding of what's going on, this is an issue when training on multiple GPUs, where the inner state_dict (for the base dataloader on the dispatcher) is fetching 1 extra item for each additional shard beyond the first? 

Did I miss this because I didn't have a test in `test_distributed_data_loop.py`, and my assumption that my test in `test_accelerator.py` would test this didn't hold because I didn't unwrap the model?

src/accelerate/data_loader.py
muellerzr: If we make it during saving, that isolates it to users who just do `Accelerator.save_state`/`Accelerator.load_state`, which (esp in the trainer) might not be what users want to end up doing, since it's always entirely optional. I'd rather it happen in `load` IMO but will think if there's a better way

muellerzr:
@byi8220 exactly. If it's not in those scripts, it's only ever ran on a single GPU since those scripts are the ones that test during DDP/multiple GPUs :) 

It's not catching an extra item, it's one complete batch ahead. (due to prefetching)

src/accelerate/data_loader.py
muellerzr: The other (probably more right option) is to fix it in `state_dict()`

byi8220: Adding onto this, is it appropriate to hack around with the state_dict's fields such as `_iter_yielded`, `_num_yielded`? 

IIUC this works in the basic case, but `torchdata` seems to support [`custom state functions`](https://github.com/pytorch/data/tree/main/torchdata/stateful_dataloader#saving-custom-state-with-iterable-style-datasets).

This makes me feel like this needs to be fixed during saving, and maybe in a way that doesn't make assumptions about the contents of state_dict (make sure to save the right state dict?).

muellerzr: We must, it's not a matter of if its appropriate or not. If we don't, the sampler/resuming simply wont work :) As we need to modify *their* values in their sampler. 

This is a naiive implementation to start, and if we hit edge cases with more things to be adjusted later, we can. But the base case is supported

muellerzr: I've migrated this logic to `_update_state_dict` to accomplish this properly

src/accelerate/test_utils/scripts/test_distributed_data_loop.py
muellerzr: `TensorDataset` can be shuffled, basically. Our iterable here can't

tests/test_accelerator.py
muellerzr: Yeah it can be removed now since we just go through the entire DL

src/accelerate/data_loader.py
byi8220: If it works, it works.

I was just wondering if there was a more robust solution, based on the idea that maybe at **some** point the state_dict could reflect what actually needs to be saved.

muellerzr:
It's why we have the scripts, it's not a perfect solution but it's what we have to do :) 

(And also, it does. They are found in `tests/test_multigpu.py`)

src/accelerate/data_loader.py
muellerzr: The only point that would be is on a non-DDP process. Otherwise they're likely not using Accelerate dataloaders

byi8220:
> (And also, it does. They are found in tests/test_multigpu.py)

Ah, so it runs [these tests](https://github.com/huggingface/accelerate/blob/main/tests/test_multigpu.py#L40-L45)? Should the `test_accelerator.py` suite be integrated somewhere here, or is that not possible?

I think I get the workflow, I still feel it's something that's really easy to miss if you're not paying attention or missing context (as I have literally just done).

BenjaminBossan (approved):
Thanks for addressing my comments.

I have added one suggestion on how to possibly deal with custom stateful dataloaders, otherwise this LGTM.

src/accelerate/data_loader.py
muellerzr: I'm more a fan of this, but not sure if we can get there in our current state since it's either a `StatefulDataLoader` or a native torch.utils.data.DataLoader that gets built

BenjaminBossan: Got confused, the type of the data loader is fix.

muellerzr: I went with an `adjust_state_dict_for_prefetch` func in `DataLoaderAdapter`, with documentation on how overriding it should work.

muellerzr:
@byi8220 it's not possible as these are only CPU runners. We then run them on merge to `main` since those are GPU runners and can do so (it's how I flagged/saw this had issues)

BenjaminBossan (approved):
Thanks for the latest changes, LGTM. `adjust_state_dict_for_prefetch` is implemented even more elegantly than I was thinking of.

byi8220:
> We then run them on merge to main since those are GPU runners and can do so (it's how I flagged/saw this had issues)

Well, I guess if it's caught on merge it's not that big of an issue

PR #3036, opened 2024-08-23
Trigger `weights_only=True` by default for all compatible objects

muellerzr:
# What does this PR do?

Sets `weights_only=True` by default when using `torch.save()` for all compatible objects:

* Model weights
* Optimizer states
* DataLoader states
* Scheduler states

The only one that we can't do right now are the random states, since the numpy random states will raise an exception if we add the right classes for safe loading:

```
  File "/home/zach/work/accelerate/test.py", line 24, in <module>
    _ = load(p / "r2.pkl", weights_only=True)
  File "/home/zach/work/accelerate/src/accelerate/utils/other.py", line 245, in load
    loaded_obj = torch.load(f, map_location=map_location, **kwargs)
  File "/home/zach/miniconda3/envs/accelerate/lib/python3.10/site-packages/torch/serialization.py", line 1096, in load
    raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
_pickle.UnpicklingError: Weights only load failed. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
 Please file an issue with the following so that we can make `weights_only=True` compatible with your use case: WeightsUnpickler error: Can only build Tensor, parameter or OrderedDict objects, but got <class 'numpy.dtypes.UInt32DType'>
```

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -268,7 +268,7 @@ def load_accelerator_state(
 
     # Random states
     try:
-        states = load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"), weights_only=False)
+        states = load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"), weights_only=True)
         if "step" in states:
             override_attributes["step"] = states["step"]
         random.setstate(states["random_state"])
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -35,12 +35,14 @@ from .utils import (
     is_mlu_available,
     is_torch_xla_available,
     is_xpu_available,
+    load,
     save,
 )
 
 
 if is_torch_xla_available():
     import torch_xla.core.xla_model as xm
+
 
 from .logging import get_logger
 from .state import PartialState
@@ -217,7 +219,7 @@ def load_accelerator_state(
         else:
             # Load with torch
             input_model_file = input_dir.joinpath(f"{MODEL_NAME}{ending}.bin")
-            state_dict = torch.load(input_model_file, map_location=map_location)
+            state_dict = load(input_model_file, map_location=map_location, weights_only=True)
             model.load_state_dict(state_dict, **load_model_func_kwargs)
     logger.info("All model weights loaded successfully")
 
@@ -225,7 +227,7 @@ def load_accelerator_state(
     for i, opt in enumerate(optimizers):
         optimizer_name = f"{OPTIMIZER_NAME}.bin" if i == 0 else f"{OPTIMIZER_NAME}_{i}.bin"
         input_optimizer_file = input_dir.joinpath(optimizer_name)
-        optimizer_state = torch.load(input_optimizer_file, map_location=map_location)
+        optimizer_state = load(input_optimizer_file, map_location=map_location, weights_only=True)
         optimizers[i].load_state_dict(optimizer_state)
     logger.info("All optimizer states loaded successfully")
 
@@ -233,7 +235,8 @@ def load_accelerator_state(
     for i, scheduler in enumerate(schedulers):
         scheduler_name = f"{SCHEDULER_NAME}.bin" if i == 0 else f"{SCHEDULER_NAME}_{i}.bin"
         input_scheduler_file = input_dir.joinpath(scheduler_name)
-        scheduler.load_state_dict(torch.load(input_scheduler_file))
+        scheduler_state = load(input_scheduler_file, weights_only=True)
+        scheduler.load_state_dict(scheduler_state)
     logger.info("All scheduler states loaded successfully")
 
     for i, dataloader in enumerate(dataloaders):
@@ -245,24 +248,27 @@ def load_accelerator_state(
         if isinstance(dataloader.dataset, IterableDatasetShard):
             sampler = dataloader.get_sampler()
             if isinstance(sampler, SeedableRandomSampler):
-                sampler = dataloader.set_sampler(torch.load(input_sampler_file))
+                sampler = dataloader.set_sampler(torch.load(input_sampler_file, weights_only=True))
         if getattr(dataloader, "use_stateful_dataloader", False):
             dataloader_state_dict_name = "dl_state_dict.bin" if i == 0 else f"dl_state_dict_{i}.bin"
             input_dataloader_state_dict_file = input_dir.joinpath(dataloader_state_dict_name)
             if input_dataloader_state_dict_file.exists():
-                state_dict = torch.load(input_dataloader_state_dict_file)
+                state_dict = torch.load(input_dataloader_state_dict_file, weights_only=True)
                 dataloader.load_state_dict(state_dict)
+                sampler_state = load(input_sampler_file, weights_only=True)
+                sampler = dataloader.set_sampler(sampler_state)
     logger.info("All dataloader sampler states loaded successfully")
 
     # GradScaler state
     if scaler is not None:
         input_scaler_file = input_dir.joinpath(SCALER_NAME)
-        scaler.load_state_dict(torch.load(input_scaler_file))
+        scaler_state = load(input_scaler_file, weights_only=True)
+        scaler.load_state_dict(scaler_state)
         logger.info("GradScaler state loaded successfully")
 
     # Random states
     try:
-        states = torch.load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"))
+        states = load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"), weights_only=False)
         if "step" in states:
             override_attributes["step"] = states["step"]
         random.setstate(states["random_state"])
@@ -295,8 +301,9 @@ def save_custom_state(obj, path, index: int = 0, save_on_each_node: bool = False
 
 def load_custom_state(obj, path, index: int = 0):
     """
-    Loads the state of `obj` at `{path}/custom_checkpoint_{index}.pkl`
+    Loads the state of `obj` at `{path}/custom_checkpoint_{index}.pkl`. Will always set `weights_only=False` when
+    loading the state.
     """
     load_location = f"{path}/custom_checkpoint_{index}.pkl"
     logger.info(f"Loading the state of {get_pretty_name(obj)} from {load_location}")
-    obj.load_state_dict(torch.load(load_location, map_location="cpu"))
+    obj.load_state_dict(load(load_location, map_location="cpu", weights_only=False))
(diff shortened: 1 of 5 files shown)


src/accelerate/utils/imports.py
```diff
@@ -444,3 +444,9 @@ def is_torchdata_stateful_dataloader_available():
         torchdata_version = version.parse(importlib.metadata.version("torchdata"))
         return compare_versions(torchdata_version, ">=", "0.8.0")
     return False
+
+
+def is_weights_only_available():
+    # Weights only with allowlist was added in 2.4.0
+    # ref: https://github.com/pytorch/pytorch/pull/124331
+    return is_torch_version(">=", "2.4.0")
```
muellerzr, line 452: The other option is to ignore the allowlist and force `weights_only` from v2.0.0, WDYT @BenjaminBossan.

(Basically this will wind up allowing only the optimizer, model, and scheduler)

src/accelerate/checkpointing.py
```diff
@@ -240,18 +243,20 @@ def load_accelerator_state(
...
         input_scaler_file = input_dir.joinpath(SCALER_NAME)
-        scaler.load_state_dict(torch.load(input_scaler_file))
+        scaler_state = load(input_scaler_file, weights_only=True)
+        scaler.load_state_dict(scaler_state)
         logger.info("GradScaler state loaded successfully")
 
     # Random states
     try:
-        states = torch.load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"))
+        states = load(input_dir.joinpath(f"{RNG_STATE_NAME}_{process_index}.pkl"), weights_only=False)
```
BenjaminBossan, line 259: Let's add a `TODO` comment here that this should be adjusted once it works on torch side.

src/accelerate/utils/other.py
```diff
@@ -207,6 +213,42 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
...
+    """
+    if is_weights_only_available():
+        if "weights_only" not in kwargs:
+            kwargs["weights_only"] = True
+        torch.serialization.add_safe_globals(TORCH_SAFE_GLOBALS)
+    else:
+        kwargs.pop("weights_only", None)
+    loaded_obj = torch.load(f, map_location=map_location, **kwargs)
+    if is_weights_only_available():
+        torch.serialization.clear_safe_globals()
```
BenjaminBossan, line 248: Probably better to put this all in a `try ... finally` to ensure that the safe globals are cleared at the end.

src/accelerate/utils/other.py
```diff
@@ -207,6 +213,42 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
...
+    # numpy arrays are just numbers, not objects, so we can reconstruct them safely
+    np.core.multiarray._reconstruct,
+    np.ndarray,
+]
+
+
+def load(f, map_location=None, **kwargs):
+    """
+    Compatible drop-in replacement of `torch.load()` which allows for `weights_only` to be used if `torch` version is
+    2.0.0 or higher. Otherwise will ignore the kwarg.
```
BenjaminBossan, line 228: Should it be 2.4.0?

src/accelerate/utils/imports.py
BenjaminBossan: The way that we decided to solve this in torch is to stay with the current default in torch, which is `weights_only=None` (despite the docs saying otherwise), which gives the warning and then switches to `weights_only=False`. For torch 2.6 (which is projected to make the switch), we also switch to `weights_only=True`.

The reasoning was that we stick with what torch does and once torch makes the switch, users would run into potential errors no matter what.

If you feel confident that this won't disrupt accelerate users, or if it does it's better to do it earlier rather than later, it's also fine to switch now. I just think that when the corresponding torch version releases, a lot of code will break and there will be general chatter and awareness of the problem, making it easier for users to find help for their breaking code.

src/accelerate/checkpointing.py
```diff
@@ -212,23 +214,24 @@ def load_accelerator_state(
         else:
             # Load with torch
             input_model_file = input_dir.joinpath(f"{MODEL_NAME}{ending}.bin")
-            state_dict = torch.load(input_model_file, map_location=map_location)
+            state_dict = load(input_model_file, map_location=map_location, weights_only=True)
```
BenjaminBossan, line 217: Does this pass for models in general? I wonder because of the example I had posted on slack:

```python
import torch
from torch import nn

print(torch.__version__)  # 2.4.0

class MyModule(nn.Module):
    def __init__(self):
        super().__init__()
        self.dense = nn.Linear(1, 1)

net = MyModule()
torch.save(net, "/tmp/checkpoint.pt")

# uncomment to remove the error
# torch.serialization.add_safe_globals([MyModule, set, nn.Linear])
torch.load("/tmp/checkpoint.pt", weights_only=True)
```

src/accelerate/utils/other.py
```diff
@@ -207,6 +213,42 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
...
+# The following are considered "safe" globals to reconstruct various types of objects when using `weights_only=True`
+# These should be added and then removed after loading in the file
+TORCH_SAFE_GLOBALS = [
+    # numpy arrays are just numbers, not objects, so we can reconstruct them safely
+    np.core.multiarray._reconstruct,
+    np.ndarray,
+]
+
+
+def load(f, map_location=None, **kwargs):
```
BenjaminBossan, line 225: I see why you chose the name, personally I would prefer something more descriptive though.

BenjaminBossan (changes requested):
Thanks for tackling this.

Honestly, I'm a bit surprised that by just allowing these few numpy types, everything can be loaded fine (except for the rng). I think it's super important to ensure that this will not lead to breakage.

src/accelerate/utils/other.py
mikaylagawarecki: Yes, sorry about this, there is a `safe_globals` context manager, but that is only available in nightlies at the moment :/ https://pytorch.org/docs/main/notes/serialization.html#torch.serialization.safe_globals

src/accelerate/checkpointing.py
muellerzr: no, only state dicts (which is what we save here)

src/accelerate/utils/imports.py
muellerzr: If I ran into any sort of flag/issue with the checkpointing script, I'd have been more likely to not. Just in case I'll run the slow suite, but if nothing breaks I think we're safe here.

src/accelerate/checkpointing.py
muellerzr: Just triggered this normally since we found the solution

src/accelerate/utils/other.py
```diff
@@ -237,15 +242,17 @@ def load(f, map_location=None, **kwargs):
...
+        if is_weights_only_available():
+            if "weights_only" not in kwargs:
+                kwargs["weights_only"] = True
+            torch.serialization.add_safe_globals(TORCH_SAFE_GLOBALS)
+        else:
+            kwargs.pop("weights_only", None)
+        loaded_obj = torch.load(f, map_location=map_location, **kwargs)
+    finally:
+        if is_weights_only_available():
+            torch.serialization.clear_safe_globals()
```
BenjaminBossan, line 255: I wonder if this could happen: A user adds some safe globals, then uses accelerate, then wants to manually load something but now the safe globals are cleared. Maybe an edge case, but it would probably be better to restore the previous status quo instead of clearing completely.

src/accelerate/utils/imports.py
BenjaminBossan: Makes sense.

src/accelerate/utils/other.py
```diff
@@ -207,6 +214,48 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
...
+# The following are considered "safe" globals to reconstruct various types of objects when using `weights_only=True`
+# These should be added and then removed after loading in the file
+TORCH_SAFE_GLOBALS = [
+    # numpy arrays are just numbers, not objects, so we can reconstruct them safely
+    np.core.multiarray._reconstruct,
+    np.ndarray,
+    # The following are needed for the RNG states
+    encode,
+    np.dtype,
+    np.dtypes.UInt32Dtype,
```
BenjaminBossan, line 226:
```suggestion
    np.dtypes.UInt32DType,
```

BenjaminBossan (changes requested):
Thanks for the updates. A bunch of tests are failing. I added a suggestion that should fix some, but I'm not sure if it'll fix all.

Generally, if the tests indicate that this switch will work fine, I'm okay with merging. Still I'm wondering if there should not at least be a fail-safe for users to be able to switch to `weights_only=False`, as otherwise there is no way they can circumvent this, short of downgrading accelerate.

src/accelerate/utils/other.py
mikaylagawarecki: Yea I agree that this is a valid case as some libraries are adding to add_safe_globals on import (e.g. torchao, float8_experimental), perhaps we should gate on version >=2.5 and use the `safe_globals` ctx

(an alternative for now is you could technically create your own version of a context manager using a combination of `get/clear/add_{safe_globals}`, which are all available in 2.4)

Again apologies for the churn here

BenjaminBossan: I like the idea of backporting the context manager. Do you have standalone code to implement it? Probably something like:

```python
@contextmanager
def torch_safe_globals(types):
    # TODO check if torch has safe_globals, if yes use that, else:
    old_safe_globals = torch.serialization.get_safe_globals()
    torch.serialization.add_safe_globals(types)
    try:
        yield
    finally:
        torch.serialization.clear_safe_globals()
        torch.serialization.add_safe_globals(old_safe_globals)
```

src/accelerate/checkpointing.py
```diff
@@ -245,24 +247,27 @@ def load_accelerator_state(
...
+                sampler = dataloader.set_sampler(load(input_sampler_file))
         if getattr(dataloader, "use_stateful_dataloader", False):
             dataloader_state_dict_name = "dl_state_dict.bin" if i == 0 else f"dl_state_dict_{i}.bin"
             input_dataloader_state_dict_file = input_dir.joinpath(dataloader_state_dict_name)
             if input_dataloader_state_dict_file.exists():
-                state_dict = torch.load(input_dataloader_state_dict_file)
+                state_dict = load(input_dataloader_state_dict_file)
                 dataloader.load_state_dict(state_dict)
+                sampler_state = load(input_sampler_file)
+                sampler = dataloader.set_sampler(sampler_state)
```
BenjaminBossan, line 258: These are technically changes that are unrelated to the `load` functionality, right?

BenjaminBossan (approved):
Change LGTM, thanks.

There is a CI issue caused by a numpy import. It might be worth checking a few numpy versions (especially < 2.0 and >= 2.0) to be sure that all the numpy imports work.

PR #3042, opened 2024-08-25
Speed up tests by shaving off subprocess when not needed

muellerzr:
# What does this PR do?

As title states, subprocess is slow. Each time we call it adds a whole second or so to the testing suite. This PR removes that (when save/applicable) from our testing suite, and instead does a 1:1 replacement by capturing the `stdout` instead of a func (said func being the equivalent `accelerate.commands.X`. 

Total CPU CI time:

```
Before: 2m53s
After: 2m36s
```

For specific areas:

* `TpuConfigTester`: (10x reduction)
```
Before: 10.77s
After: 0.95s
```
* `AccelerateLauncherTester` (50% reduction)
```
Before: 13.46s
After: 6.99s
```

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/test_utils/__init__.py
+++ b/src/accelerate/test_utils/__init__.py
@@ -15,6 +15,7 @@ from .testing import (
     DEFAULT_LAUNCH_COMMAND,
     are_the_same_tensors,
     assert_exception,
+    capture_call_output,
     device_count,
     execute_subprocess_async,
     get_launch_command,
--- a/src/accelerate/test_utils/testing.py
+++ b/src/accelerate/test_utils/testing.py
@@ -672,10 +672,10 @@ def assert_exception(exception_class: Exception, msg: str = None) -> bool:
     if was_ran:
         raise AssertionError(f"Expected exception of type {exception_class} but ran without issue.")
 
+
 def capture_call_output(func, *args, **kwargs):
     """
-    Takes in a `func` with `args` and `kwargs`
-    and returns the captured stdout as a string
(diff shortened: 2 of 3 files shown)


BenjaminBossan (commented):
Honestly, I'm a bit confused by the command launchers. It feels that there are multiple ways to achieve the same, or very similar things, and can't wrap my heard around what the difference is.

E.g. let's take the TPU tests. Previously, they were calling `run_command`, which in turn calls `subprocess.check_output`. Now they call `tpu_command_launcher` which calls `subprocess.run`. So how come the latter is faster? Is it really related to spawning subprocesses, as it appears that both ways do that.

Maybe you plan on harmonizing this further later on (say, removing `run_command` completely) and then this will become more obvious to me, but as is I'm confused.

That said, don't let my confusion block you from proceeding if this is a clear gain in test speed.

muellerzr:
@BenjaminBossan yep exactly as it sounds, doing 2x subprocess is slower than 1x by a decent margin apparently.

BenjaminBossan:
> yep exactly as it sounds, doing 2x subprocess is slower than 1x by a decent margin apparently

Hmm, you mean that `run_command` runs subprocess twice? How is that? Also, is it not surprising that this would lead to a 10x reduction?

muellerzr:
Only when the underlying command calls subprocess (like the tpu ones)

BenjaminBossan (approved):
I'm still not 100% sure why we see the speedup, but it's fine to merge from my point of view.

PR #3046, opened 2024-08-26
Update CONTRIBUTING.md Setup Instructions

siddk:
# What does this PR do?

CONTRIBUTING.md has developers only install quality dependencies (`black, hf-doc-builder, ruff`) rather than full set of testing dependencies. Since the documentation requires folks run tests anyway, update instructions to install `dev` dependencies.

Not sure if this is overkill (and we just want `pytest` / `parameterized` as specified with the `test_prod` target).

## Before submitting
- [X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

@muellerzr

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,12 +123,15 @@ Follow these steps to start contributing:
 4. Set up a development environment by running the following command in a conda or a virtual environment you've created for working on this library:
 
    ```bash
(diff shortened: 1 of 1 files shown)


muellerzr (approved):
Thanks for going through and refreshing our directions 🤗

PR #3052, opened 2024-08-27
fix bug in `_get_named_modules`

faaany:
## What does this PR do?
I got the following error when running the UT in transformers: `tests/models/glpn/test_modeling_glpn.py::GLPNModelTest::test_model_parallelism`
```bash
src/transformers/modeling_utils.py:3780: in from_pretrained
    device_map = infer_auto_device_map(model, dtype=target_dtype, **device_map_kwargs)
../accelerate/src/accelerate/utils/modeling.py:1238: in infer_auto_device_map
    tied_parameters = find_tied_parameters(model)
../accelerate/src/accelerate/utils/modeling.py:708: in find_tied_parameters
    all_named_parameters = {name: param for name, param in _get_named_parameters(model, remove_duplicate=False)}
../accelerate/src/accelerate/utils/modeling.py:708: in <dictcomp>
    all_named_parameters = {name: param for name, param in _get_named_parameters(model, remove_duplicate=False)}
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

module = None, prefix = '', recurse = True, remove_duplicate = False

    def _get_named_parameters(module: torch.nn.Module, prefix="", recurse=True, remove_duplicate: bool = True):
        """
        Help yield various names + members of modules. Copied from PyTorch `torch.nn.Module.named_modules` for
        compatability with torch < 2.0 versions with `remove_duplicate` option added.
        """
        memo = set()
        modules = (
            _get_named_modules(module, prefix=prefix, remove_duplicate=remove_duplicate) if recurse else [(prefix, module)]
        )
        for module_prefix, module in modules:
>           members = module._parameters.items()
E           AttributeError: 'NoneType' object has no attribute '_parameters'

../accelerate/src/accelerate/utils/modeling.py:667: AttributeError
```
Pls have a review. Thx! 
@muellerzr


src/accelerate/utils/modeling.py
```diff
@@ -648,7 +648,7 @@ def _get_named_modules(
             memo.add(module)
         yield prefix, module
         for name, sub_module in module._modules.items():
-            if module is None:
+            if sub_module is None:
```
SunMarc, line 651: these function were copied from pytorch function so I don't think this is the right fix. Could you check if replacing `_get_named_parameters` in find_tied_parameters by `named_parameters` fixes the issue ? If not, it means that another fix is needed.

SunMarc (commented):
Thanks for the PR @faaany  ! I left a comment

src/accelerate/utils/modeling.py
faaany: I tried to use `named_parameters` instead of `_get_named_parameters`, but `named_parametes` doesn't have `remove_duplicate`, so it still fails at the next line.

faaany: I tried the version `0.32.0`, and it works. To be honest, from the current code of `_get_named_modules`, module would definitively be None, as long as `module._modules` contains `None`. Would it be possible the problem lies in the model constructor itself? Usually, `module.modules` wouldn't contain None.

SunMarc: > I tried to use named_parameters instead of _get_named_parameters, but named_parametes doesn't have remove_duplicate, so it still fails at the next line.

SunMarc: After investigation the pytorch implementation, I confirm that we made a mistake when copy pasting and this is the right fix ! Thanks again for the fix !

SunMarc (approved):
Nice job @faaany ! Thanks for your help !

muellerzr (approved):
Thanks for the fix!

SunMarc:
CI failing test is unrelated, merging !

PR #3057, opened 2024-08-28
checks if MLU is available via cndev

huismiling:
# What does this PR do?

Checks if mlu is available via an cndev-based check which won't trigger the drivers and leave mlu uninitialized.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


huismiling:
Hi @muellerzr , could you help to review this PR ?

BenjaminBossan (commented):
I think this could be simplified by using the [`patch_environment`](https://github.com/huggingface/accelerate/blob/939ce400cbc0478fd281b9fcc9a2a85691ba9f7d/src/accelerate/utils/other.py#L246) function.

muellerzr (commented):
Agree with @BenjaminBossan here, otherwise that's a good solution to me!

huismiling:
@BenjaminBossan `patch_environment` is very good . 
But there is Error when `from .other import patch_environment` is used in accelerate/utils/imports.py.
`ImportError: cannot import name 'ComputeEnvironment' from partially initialized module 'accelerate.utils' (most likely due to a circular import)`

BenjaminBossan:
Damn circular imports. You would have to import `patch_environment` locally (with a comment that it's to avoid circular imports) or move the function to a different module. Probably the former is more practical, WDYT @muellerzr.

muellerzr:
Agreed, if we must, former would be the way to go

huismiling:
What is the `former` ? I have no idea what I should do.

muellerzr:
Importing the func inside the function using it, with a comment explaining that it's there to avoid circular imports

huismiling:
> Importing the func inside the function using it, with a comment explaining that it's there to avoid circular imports

like this ?  This cause `circular import`, too. 

```
def is_mlu_available(check_device=False):
    if importlib.util.find_spec("torch_mlu") is None:
        return False

    import torch_mlu  # noqa: F401
    from .other import patch_environment

    with patch_environment(PYTORCH_CNDEV_BASED_MLU_CHECK="1"):
        available = torch.mlu.is_available()

    return available
```

BenjaminBossan:
> This cause `circular import`, too.

Ouch, this is because `is_mlu_available()` is called on root level here:

https://github.com/huggingface/accelerate/blob/3fcc9461c4fcb7228df5e5246809ba09cfbb232e/src/accelerate/state.py#L57-L58

So IMO the right way would be to move `patch_environment` to a separate module. Reimporting it to `other.py` should still work, so for users nothing would change. WDYT @muellerzr?

PR #3062, opened 2024-08-30
remove cpu restriction for bnb training

jiqing-feng:
Refer to [cpu_bnb](https://github.com/huggingface/transformers/pull/31098), we will support bnb 8bit and 4bit training on CPU.


BenjaminBossan:
Would it make sense to add a version check on bitsandbytes before exempting CPU usage from raising, for those users who are still using older versions?

jiqing-feng:
> Would it make sense to add a version check on bitsandbytes before exempting CPU usage from raising, for those users who are still using older versions?

Hi @BenjaminBossan , thanks for your review, I have fixed it. As transformers has supported [CPU BNB path](https://github.com/huggingface/transformers/pull/31098). I think we can start to review this PR.

src/accelerate/accelerator.py
```diff
@@ -1443,9 +1443,16 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                             "you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device()}` or `device_map={'':torch.xpu.current_device()}`"
                         )
 
-            if "cpu" in model_devices or "disk" in model_devices:
+            bnb_multi_backends = False
+            try:
+                from transformers.utils import is_bitsandbytes_multi_backend_available
+                bnb_multi_backends = is_bitsandbytes_multi_backend_available()
+            except ImportError:
+                bnb_multi_backends = False
```
SunMarc, line 1451: Let's not use the transformers import in accelerate. You can just create the same function in accelerate. This way, we don't need to have the try/catch also.

SunMarc (commented):
Thanks for the PR @jiqing-feng ! Left a suggestion

src/accelerate/accelerator.py
jiqing-feng: Right. Done.

SunMarc (approved):
Thanks for iterating !

jiqing-feng:
Hi @SunMarc . Thanks, can you merge it since all tests passed?

SunMarc:
Oh just a small nit before merging ! Also, I don't think this is specific to 8-bit training,

jiqing-feng:
> Oh just a small nit before merging ! Also, I don't think this is specific to 8-bit training,

Yes, I have added 4-bit in the log.

jiqing-feng:
> > Oh just a small nit before merging ! Also, I don't think this is specific to 8-bit training,
> 
> Yes, I have added 4-bit in the log.

Please let me know if there is anything I need to change, thx!

PR #3074, opened 2024-09-03
Allow DataLoaderAdapter subclasses to be pickled by implementing `__reduce__`

### LINKED ISSUE #3070: Pickling the accelerator after preparing data loader no longer possible
### System Info

```Shell
accelerate v0.34.0
```

### Information

- [X] My own modified scripts

### Tasks

### Reproduction

```python
import pickle
import torch

data = torch.arange(10)
ds = torch.utils.data.TensorDataset(data)
dl = torch.utils.data.DataLoader(ds)

accelerator = Accelerator()
pickle.dumps(accelerator)  # works
accelerator.prepare(dl)
pickle.dumps(accelerator)  # fails
```

The error is:

```
    pickle.dumps(accelerator)  # fails
    ^^^^^^^^^^^^^^^^^^^^^^^^^
_pickle.PicklingError: Can't pickle <class 'accelerate.data_loader.DataLoaderShard'>: it's not the same object as accelerate.data_loader.DataLoaderShard
```

### Expected behavior

Pickling and unpickling the `Accelerator` instance used to work but now it fails if a data loader has been prepared. I could trace down the issue to this PR: #2895. Although I haven't investigated further, I'm fairly certain that messing with the `__class__` attribute is the reason:

https://github.com/huggingface/accelerate/blob/b5235f21d83c7f714a44d876cfd6fd0b1339dd38/src/accelerate/data_loader.py#L436

If that's so, there might be the necessity to adjust `__getstate__` or `__reduce__` on `DataLoaderAdapter` to account for the fact that the class is dynamically changed.

byi8220:
# What does this PR do?

Fixes #3070

This PR contains the following:

1. Cleanup the dynamic class resolution of DataLoaderAdapter a bit. Instead of performing dynamic class overriding in `__init__`, we instead override the property to pass through the base dataloader's class.

    *Note:* This means `isinstance(obj, DataLoaderAdapter)` breaks. However, since `DataLoaderAdapter` was introduced very recently, it's very unlikely this will break anything, since there [shouldn't](https://www.hyrumslaw.com/) be any reason why existing downstream code would be checking this.
2. Implement `__reduce__` in DataLoaderAdapter subclasses to allow pickling of these classes, despite the `__class__` override. This implementation just repasses the object's init arguments and `__dict__` back into the constructor.
3. Some cleanup in DataLoaderDispatcher (use base_dataloader instead of `super()`), and some cleanup in SkipDataLoader (implement `__len__`)

Some caveats:

1. StatefulDataLoader's state_dict [might not be pickleable](https://github.com/pytorch/pytorch/blob/main/torch/utils/data/dataloader.py#L724-L730).
2. Pickling in general seems dubious and fragile.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@muellerzr
@BenjaminBossan

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -572,7 +572,20 @@ class DataLoaderShard(DataLoaderAdapter, DataLoaderStateMixin):
         self.end()
 
     def __reduce__(self):
-        return super().__reduce__()
+        return (
+            DataLoaderShard,
+            (
+                self.base_dataloader.dataset,
+                self.device,
+                self.rng_types,
+                self.synchronized_generator,
+                self.skip_batches,
+                self.use_stateful_dataloader,
+                self._drop_last,
+                self._non_blocking,
+            ),
+            self.__dict__,
+        )
 
     def set_epoch(self, epoch: int):
         # In case it is manually passed in, the user can set it to what they like
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -572,20 +572,9 @@ class DataLoaderShard(DataLoaderAdapter, DataLoaderStateMixin):
         self.end()
 
     def __reduce__(self):
-        return (
-            DataLoaderShard,
-            (
-                self.base_dataloader.dataset,
-                self.device,
-                self.rng_types,
-                self.synchronized_generator,
-                self.skip_batches,
-                self.use_stateful_dataloader,
-                self._drop_last,
-                self._non_blocking,
-            ),
-            self.__dict__,
-        )
+        args = super().__reduce__()
+        return (DataLoaderShard, *args[1:])
+
 
     def set_epoch(self, epoch: int):
         # In case it is manually passed in, the user can set it to what they like
@@ -881,19 +870,8 @@ class DataLoaderDispatcher(DataLoaderAdapter, DataLoaderStateMixin):
             return math.ceil(whole_length / self.state.num_processes)
 
     def __reduce__(self):
-        return (
-            DataLoaderDispatcher,
-            (
-                self.base_dataloader.dataset,
-                self.split_batches,
-                self.skip_batches,
-                self.use_stateful_dataloader,
-                self._drop_last,
-                self._non_blocking,
-                self.slice_fn,
-            ),
-            self.__dict__,
-        )
+        args = super().__reduce__()
+        return (DataLoaderDispatcher, *args[1:])
 
     @property
     def total_batch_size(self):
@@ -1238,11 +1216,9 @@ class SkipDataLoader(DataLoaderAdapter, DataLoaderStateMixin):
         return len(self.base_dataloader) - self.skip_batches
 
     def __reduce__(self):
-        return (
-            SkipDataLoader,
-            (self.base_dataloader.dataset, self.skip_batches, self.use_stateful_dataloader),
-            self.__dict__,
-        )
+        args = super().__reduce__()
+        return (SkipDataLoader, *args[1:])
+
 
 
 def skip_first_batches(dataloader, num_batches=0):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -1212,7 +1212,6 @@ class SkipDataLoader(DataLoaderAdapter, DataLoaderStateMixin):
         self.end()
 
     def __len__(self):
-        print("len called")
         return len(self.base_dataloader) - self.skip_batches
 
     def __reduce__(self):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -435,9 +435,8 @@ class DataLoaderAdapter:
     @property
     def __class__(self):
         """
-        In order to maintain backwards compatability with other code, we need to ensure `isinstance(obj, DataLoader)`
-        returs true. This is because some downstream code assumes that the `DataLoader` is the base class of the
-        object.
+        In order to maintain backwards compatability with other code, we need to ensure `isinstance(obj, DataLoader)` returs true.
+        This is because some downstream code assumes that the `DataLoader` is the base class of the object.
         """
         return self.base_dataloader.__class__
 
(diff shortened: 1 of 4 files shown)


src/accelerate/data_loader.py
```diff
@@ -580,6 +571,22 @@ def __iter__(self):
...
+                self.base_dataloader.dataset,
+                self.device,
+                self.rng_types,
+                self.synchronized_generator,
+                self.skip_batches,
+                self.use_stateful_dataloader,
+                self._drop_last,
+                self._non_blocking,
+            ),
+            self.__dict__,
```
BenjaminBossan, line 587: Instead of hard-coding these arguments, how about calling `super().__reduce__()`?

src/accelerate/data_loader.py
```diff
@@ -865,14 +872,29 @@ def set_epoch(self, epoch: int):
             self.dataset.set_epoch(epoch)
 
     def __len__(self):
-        whole_length = super().__len__()
+        whole_length = self.base_dataloader.__len__()
```
BenjaminBossan, line 875: How about: `len(self.base_dataloader)`?

src/accelerate/data_loader.py
```diff
@@ -865,14 +872,29 @@ def set_epoch(self, epoch: int):
...
+            (
+                self.base_dataloader.dataset,
+                self.split_batches,
+                self.skip_batches,
+                self.use_stateful_dataloader,
+                self._drop_last,
+                self._non_blocking,
+                self.slice_fn,
+            ),
+            self.__dict__,
```
BenjaminBossan, line 895: Same comment as above.

src/accelerate/data_loader.py
```diff
@@ -1211,6 +1233,17 @@ def __iter__(self):
...
 
+    def __len__(self):
+        print("len called")
+        return len(self.base_dataloader) - self.skip_batches
+
+    def __reduce__(self):
+        return (
+            SkipDataLoader,
+            (self.base_dataloader.dataset, self.skip_batches, self.use_stateful_dataloader),
+            self.__dict__,
```
BenjaminBossan, line 1244: Same comment as above.

BenjaminBossan (commented):
Thanks for working on this issue so quickly. I agree that pickling the data loader is (even if indirectly, via accelerator) is not generally a good idea. I'll add a note to the skorch docs that this should be avoided. Still, I think it's better to fix this, as we don't know what other code may depend on it, even if only indirectly (e.g. pickle is often being used under the hood to pass around data between processes).

As to the solution, I made some comments which I think will make the code more robust. Regarding the use of `@property` for `__class__`, I really feel like we're dancing on a razor's edge here (the limits of the Python data model), though this is probably no worse than what we had before.

byi8220:
>Regarding the use of @property for __class__, I really feel like we're dancing on a razor's edge here (the limits of the Python data model),

We really are. This idea comes up a few times if you look for it online, and unfortunately given the nature of the problem, this is the absolute last resort: https://stackoverflow.com/a/52172876. The problem is that the more sensible solutions such as using a metaclass or tampering with `__instancecheck__` requires us to have control over the base class. 

I will reiterate, the **only** reason for this magic code to exist is to make sure that `isinstance(dl_shard, DataLoader)` works. If this requirement is lifted, we don't need to mess with the class structure at all.

src/accelerate/data_loader.py
byi8220: I don't know if that works. IIUC, `__reduce__` is called to specify how we want to pickle an object for reconstruction. In this context, if I'm understanding right, the current code states "We want to pickle a DataLoaderShard by constructing a DataLoaderShard with these arguments and filling it's `__dict__` with my elements."

If we replace this with `return super().__reduce__()`, then doesn't this change this to "We want to pickle a DataLoaderShard by turning it into a DataLoader." In that case, the following would happen:

```python
## Behavior of code using super().__reduce__():

# accelerator prepares a dataloader shard from a dataloader
dl_shard = original_accel.prepare(dl)
assert isinstance(dl_shard, DataLoaderShard)
assert isinstance(original_accel._dataloaders[0], DataLoaderShard)

# Try to pickle and unpickle the dataloader
loaded_accel = pickle.loads(pickle.dumps(original_accel)

# The unpickled accelerator's underlying dataloaders are no longer sharded
assert isinstance(loaded_accel._dataloaders[0], DataLoader)
```
Is this the intended behavior?

src/accelerate/data_loader.py
byi8220: Done

src/accelerate/data_loader.py
BenjaminBossan: Sorry, what I meant is something like:

```python
args = super().__reduce__()
return (DataLoaderShard, *args[1:])
```

That way, the 2nd and 3rd argument (which are roughly the arguments needed to construct the instance) don't need to be hardcoded, only the constructor class. If we hard-code the arguments, we need to remember to change them each time they're changed on the class.

byi8220: Done. The above block doesn't include the object `__dict__`, but I'm not sure if we need to add that or not. It passes my existing tests as written

src/accelerate/data_loader.py
```diff
@@ -1211,6 +1211,15 @@ def __iter__(self):
                 yield batch
         self.end()
 
+    def __len__(self):
+        print("len called")
```
muellerzr, line 1215: Leftover print statement :) 
```suggestion
```

tests/test_accelerator.py
```diff
@@ -647,6 +648,53 @@ def test_can_unwrap_model(self):
...
+        if dispatch_batches:
+            assert isinstance(loaded_dl, DataLoaderDispatcher)
+        else:
+            assert isinstance(loaded_dl, DataLoaderShard)
+        assert len(loaded_dl) == len(original_dl)
+        assert [i for i in loaded_dl] == [i for i in original_dl]
+
+        # Test skip dataloader works as expected as well
+        loaded_skip_dl = model_loaded._dataloaders[1]
+        print(model_loaded._dataloaders)
```
muellerzr, line 689: Leftover
```suggestion
```

muellerzr (approved):
Thanks for fixing this! Looks like some leftover print statements, I'll also run it locally to make sure there's nothing else lingering/unexpected too.

src/accelerate/data_loader.py
byi8220: removed

BenjaminBossan (approved):
LGTM, thanks for quickly taking up this task and finding a solution. :crossed_fingers: that this will avoid further trouble down the line.

A nit based on my personal preference, I like to add a comment to non-obvious code like the `__reduce__` method to indicate why it's needed. This way, the reader doesn't have to git blame and search through github to understand the code.

muellerzr:
@byi8220 can you do a quick `make style; make quality` then we can include it in the patchfix coming out soon :)

byi8220:
Added docstrings for `__reduce__` methods and ran `make style && make quality`

PR #3076, opened 2024-09-04
use the correct available memory API for XPU

faaany:
## What does this PR do?
This PR introduces the new API for get xpu available memory. Below is an example of before and after:

```python
from transformers import AutoModelForSeq2SeqLM

model = AutoModelForSeq2SeqLM.from_pretrained("bigscience/T0pp", device_map="auto")
print(model.hf_device_map)
```

Before
```python
{'': 'cpu'}
```

After
```python
{'shared': 0, 'decoder.embed_tokens': 0, 'encoder.embed_tokens': 0, 'encoder.block.0': 0, 'encoder.block.1': 0, 'encoder.block.2': 0, 'encoder.block.3': 0, 'encoder.block.4': 0, 'encoder.block.5': 0, 'encoder.block.6': 1, 'encoder.block.7': 1, 'encoder.block.8': 1, 'encoder.block.9': 1, 'encoder.block.10': 1, 'encoder.block.11': 1, 'encoder.block.12': 1, 'encoder.block.13': 1, 'encoder.block.14': 2, 'encoder.block.15': 2, 'encoder.block.16': 2, 'encoder.block.17': 2, 'encoder.block.18': 2, 'encoder.block.19': 2, 'encoder.block.20': 2, 'encoder.block.21': 2, 'encoder.block.22': 3, 'encoder.block.23': 3, 'encoder.final_layer_norm': 3, 'encoder.dropout': 3, 'decoder.block.0': 3, 'decoder.block.1': 3, 'decoder.block.2': 3, 'decoder.block.3': 3, 'decoder.block.4': 4, 'decoder.block.5': 4, 'decoder.block.6': 4, 'decoder.block.7': 4, 'decoder.block.8': 4, 'decoder.block.9': 4, 'decoder.block.10': 5, 'decoder.block.11': 5, 'decoder.block.12': 5, 'decoder.block.13': 5, 'decoder.block.14': 5, 'decoder.block.15': 5, 'decoder.block.16': 6, 'decoder.block.17': 6, 'decoder.block.18': 6, 'decoder.block.19': 6, 'decoder.block.20': 6, 'decoder.block.21': 6, 'decoder.block.22': 7, 'decoder.block.23': 7, 'decoder.final_layer_norm': 7, 'decoder.dropout': 7, 'lm_head': 7}
```

### DIFF
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -427,12 +427,14 @@ def is_xpu_available(check_device=False):
 
 def get_xpu_available_memory(device_index: int):
     if is_ipex_available():
-
         try:
             from intel_extension_for_pytorch.xpu import mem_get_info
+
             return mem_get_info(device_index)[0]
         except Exception:
-            warnings.warn("The XPU `mem_get_info` API is available in IPEX version >=2.5. To get the correct available memory, pls consider upgrading your IPEX version")
+            warnings.warn(
+                "The XPU `mem_get_info` API is available in IPEX version >=2.5. The current returned availabel memory is incorrect. Pls consider upgrading your IPEX version."
+            )
             return torch.xpu.max_memory_allocated(device_index)
     else:
         # this branch is for stock PyTorch 2.4
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -425,22 +425,6 @@ def is_xpu_available(check_device=False):
     return hasattr(torch, "xpu") and torch.xpu.is_available()
 
 
-def get_xpu_available_memory(device_index: int):
-    if is_ipex_available():
-        try:
-            from intel_extension_for_pytorch.xpu import mem_get_info
-
-            return mem_get_info(device_index)[0]
-        except Exception:
-            warnings.warn(
-                "The XPU `mem_get_info` API is available in IPEX version >=2.5. The current returned availabel memory is incorrect. Pls consider upgrading your IPEX version."
-            )
-            return torch.xpu.max_memory_allocated(device_index)
-    else:
-        # this branch is for stock PyTorch 2.4
-        return torch.xpu.max_memory_allocated(device_index)
-
-
 def is_dvclive_available():
     return _is_package_available("dvclive")
 
--- a/src/accelerate/utils/imports.py
+++ b/src/accelerate/utils/imports.py
@@ -425,14 +425,18 @@ def is_xpu_available(check_device=False):
     return hasattr(torch, "xpu") and torch.xpu.is_available()
 
 
-def get_xpu_available_memory(i: int):
+def get_xpu_available_memory(device_index: int):
     if is_ipex_available():
-        from intel_extension_for_pytorch.xpu import mem_get_info
 
-        return mem_get_info(i)[0]
+        try:
+            from intel_extension_for_pytorch.xpu import mem_get_info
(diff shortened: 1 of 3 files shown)


muellerzr (approved):
Thanks! Overall this looks good to me, let's use `device_index` however rather than `i` (or something similar to avoid single-letter input names)

faaany:
> Thanks! Overall this looks good to me, let's use `device_index` however rather than `i` (or something similar to avoid single-letter input names)

thanks for pointing this out! My fault..

I also added a version check.

faaany:
Hi @muellerzr , how do you think about the new update?

src/accelerate/utils/imports.py
```diff
@@ -425,6 +425,22 @@ def is_xpu_available(check_device=False):
...
+
+            return mem_get_info(device_index)[0]
+        except Exception:
+            warnings.warn(
+                "The XPU `mem_get_info` API is available in IPEX version >=2.5. The current returned availabel memory is incorrect. Pls consider upgrading your IPEX version."
+            )
+            return torch.xpu.max_memory_allocated(device_index)
+    else:
+        # this branch is for stock PyTorch 2.4
+        return torch.xpu.max_memory_allocated(device_index)
```
muellerzr, line 441: Let's move this to `utils.memory` instead of imports.

For the implementation itself, we can do:

```python
if is_ipex_available():
    ipex_version = importlib.metadata.version("intel_extension_for_pytorch")
    if compare_versions(ipex_version, ">=", "2.5"):
        from intel_extension_for_pytorch.xpu import mem_get_info
        return mem_get_info(device_index)[0]
    warnings.warn(
        "The XPU `mem_get_info` API is available in IPEX version >=2.5. The current returned available memory is incorrect. Please consider upgrading your IPEX version."
    )
return torch.xpu.max_memory_allocated(device_index)
```

muellerzr (commented):
Thought about it a bit more, let's:

1. Move this to `utils.memory`
2. Use the new implementation I've shown, which takes care of things a bit cleaner.

src/accelerate/utils/imports.py
faaany: thanks for the nice refactoring! code updated

src/accelerate/utils/memory.py
```diff
@@ -159,3 +162,16 @@ def decorator(*args, **kwargs):
...
+
+
+def get_xpu_available_memory(device_index: int):
+    if is_ipex_available():
+        ipex_version = importlib.metadata.version("intel_extension_for_pytorch")
+        if compare_versions(ipex_version, ">=", "2.5"):
+            from intel_extension_for_pytorch.xpu import mem_get_info
+
+            return mem_get_info(device_index)[0]
+        warnings.warn(
```
BenjaminBossan, line 174: Just to be sure: When ipex is _not_ available, `torch.xpu.max_memory_allocated(device_index)` will return the correct result? I'm asking since the warning is only given if ipex is available.

BenjaminBossan (approved):
Thanks, LGTM. I have a clarifying question, but it's not a blocker.

src/accelerate/utils/memory.py
faaany: you are right! `torch.xpu.max_memory_allocated(device_index)` doesn't return the correct result. So I updated the warning. Thx!!

faaany:
ok, I think we can merge now. 😊

PR #3077, opened 2024-09-04
Update README.md to include distributed image generation gist

sayakpaul:
As discussed through Slack.

### DIFF
--- a/examples/inference/distributed/README.md
+++ b/examples/inference/distributed/README.md
@@ -23,3 +23,7 @@ Or:
 ```bash
 torchrun --nproc-per-node {NUM_GPUS} phi2.py
 ```
+
+## Notes
+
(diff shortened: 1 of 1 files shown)


sayakpaul:
@muellerzr okay to merge?

muellerzr (commented):
Thanks! Would you be maybe open to having the gist live inside the `distributed` folder instead of there?

sayakpaul:
I think that might confuse people because there’s already an existing example. I think the purpose of the examples in that folder is to show bare minimum usage.

sayakpaul:
Okay I can add. Any name you have in mind for the script?

muellerzr:
`distributed_image_generation.py` would work just fine unless you can think of something more :)

sayakpaul:
@muellerzr done!

Issue #2496, opened 2024-02-27
[RFC] Supporting multiple models with DeepSpeed


pacman100:
### System Info

```Shell
PyTorch 2.2.1
DeepSpeed 0.13.4
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [ ] My own task or dataset (give details below)

## Abstract
We are considering supporting multiple models with DeepSpeed when using Accelerate. We will be using the term model and DeepSpeed engine interchangeably. 

## Motivation and Background
Currently, when using Accelerate's integration of DeepSpeed, only a single model is supported. This limits the use cases such as RLHF, GANs, Knowledge Distillation etc which involve multiple models. We also have interest in this feature as per the below feature requests:
1. https://huggingface.slack.com/archives/C06CEE9C1M4/p1706821695816299
2. [Passing multiple models with DeepSpeed will fail · Issue #253 · huggingface/accelerate (github.com)](https://github.com/huggingface/accelerate/issues/253)
3. [Passing multiple models with DeepSpeed will fail · Issue #1388 · huggingface/accelerate (github.com)](https://github.com/huggingface/accelerate/issues/1388)

The reasons for restricting to only a single model support is given below:
1. The user can only provide a single DeepSpeed config plugin/DeepSpeed config file corresponding to a single model. Ideally, the user would have different DeepSpeed configs for different models.
2. DeepSpeed needs to keep track of the model, its optimizer and scheduler. Therefore, we currently have only one global DeepSpeed engine wrapper to control the backward and optimizer/scheduler step. 

## Proposal
The aim would be to solve the 2 challenges above. This would need:
1. Support multiple DeepSpeed configurations. I believe the questionnaire with minimal DeepSpeed plugin shouldn’t be changed and should continue to support a single model. This should act like the default config to be used by all the models.
5. Flexibility to use different deepspeed config should be part of `prepare` method. For example, given 4 models in rlhf scenario, I should be able to do the below:

```python
# rlhf
...
model_1 = actor_model()
model_2 = critic_model()
model_3 = reference_model()
model_4 = reward_model()

optimizer_1 = torch.optim.AdamW(model_1.parameters(), lr=lr_1)
[...]

PR #3097, opened 2024-09-10
POC: multiple model/configuration DeepSpeed support

muellerzr:
# Multiple model/multiple configuration support for DeepSpeed

## What does this add?

While we await for potentially rewriting much of the deepspeed (DS) implementation to get rid of the `engine` mechanic, this PR provides *a way* for users to use multiple deepspeed plugins (potentially with a single accelerator) to have different models prepared different ways/using different DS configurations.

## Who is it for?

Solves https://github.com/huggingface/accelerate/issues/2496 (*Finally*)

## Why is it needed?

There are *many* cases where users may want to use different DS configs/prepare multiple models when using DS. One such case is TRL where we may want an untrained model distributed in Zero3, while another in Zero2. 

**Note**: for the case of multiple models being trained, a user should create a second Accelerator object (since we require the prepared engine to do `backward()`). Since the plugins are stored in the `state`, you don't need to re-pass in the plugins just create a secondary `Accelerator`.

### User-facing:

Users may now pass in a *list* of `DeepSpeedPlugin`'s to the `Accelerator`. The first plugin will be the "default" (so if you *don't* set an active plugin, this is what's used during all calls to `.prepare()`). When a user wants a different configuration, simply call `plugin.enable()` and this will in-turn setup that particular plugin to be the one `accelerator.prepare` will call (see examples below)

### Internal structure:

* `Accelerator`/`AcceleratorState` now stores a list of `DeepSpeedPlugins`
* `get_active_deepspeed_plugin` will go through and find of the list of `DeepSpeedPlugins` which one is enabled. 
* Only a singular DS plugin should be "active"
* We always try and setup the cached transformers plugin since we may constantly be swapping out its internals being used

## Basic Usage Example(s):

```python
from accelerate import Accelerator, DeepSpeedPlugin
plugin1 = DeepSpeedPlugin(config_file_a)
plugin2 = DeepSpeedPlugin(config_file_b)

plugins = {"student": plugin1, "teacher": plugin2}

accelerator = Accelerator(deepspeed_plugin=plugins)
# The ones we're training on
model, optimizer, scheduler = accelerator.prepare(model, optimizer, scheduler)
# Our inference/secondary config
accelerator.state.enable_deepspeed_plugin("teacher")
model2 = accelerator.prepare(model2)
```

Ideally this stays how it is, and we can remove that secondary/n-model `Accelerator`'s later once we've worked with the DS team to make the DS api more pytorch-ic 

## Who can review?

@BenjaminBossan 
@SunMarc 
@stas00


src/accelerate/accelerator.py
```diff
@@ -285,11 +287,18 @@ def __init__(
...
+                        raise TypeError("`deepspeed_plugin` must be a DeepSpeedPlugin object.")
+            else:
+                if not isinstance(deepspeed_plugin, DeepSpeedPlugin):
+                    raise TypeError("`deepspeed_plugin` must be a DeepSpeedPlugin object.")
+
+        if deepspeed_plugin is not None:
             os.environ["ACCELERATE_USE_DEEPSPEED"] = "true"  # use DeepSpeed if plugin is provided
-        if deepspeed_plugin:
+            if not isinstance(deepspeed_plugin, (tuple, list)):
+                deepspeed_plugin = [deepspeed_plugin]
```
stas00, line 301: turn l286 into a list right away.

then move this code to line 290, and then you don't need to make the check if it's a plugin twice.

stas00: possibly while at it rename to `deepspeed_plugins` - not sure if it breaks any public apis?

src/accelerate/accelerator.py
```diff
@@ -532,6 +544,10 @@ def __init__(
 
         check_os_kernel()
 
+    @property
+    def deepspeed_plugin(self):
+        return self.state.deepspeed_plugin
```
stas00, line 549: `s/plugin/plugins/`?

src/accelerate/accelerator.py
```diff
@@ -1836,7 +1852,14 @@ def _prepare_deepspeed(self, *args):
...
                     result[i] = scheduler
             # pointing for deepspeed_engine_wrapped.backward()
-            self.deepspeed_engine_wrapped = DeepSpeedEngineWrapper(engine)
+            if self.deepspeed_engine_wrapped is None:
+                self.deepspeed_engine_wrapped = DeepSpeedEngineWrapper(engine)
+            else:
+                logger.warning(
+                    "A wrapped DeepSpeed engine reference is currently tied for this `Accelerator()` instance. "
+                    "If you want to call `accelerator.backward()` referencing a new model/engine, "
+                    "please create a seperate `Accelerator()` instance and call `accelerator.prepare()` on it."
```
stas00, line 1861:
```suggestion
                    "please create a separate `Accelerator()` instance and call `accelerator.prepare()` on it."
```

src/accelerate/utils/deepspeed.py
```diff
@@ -21,6 +21,10 @@
 from ..scheduler import AcceleratedScheduler
 
 
+def get_active_deepspeed_plugin(state):
+    return next((plugin for plugin in state.deepspeed_plugin if plugin.enabled), None)
```
stas00, line 25: document that it'll return `None` if none are active?

But is this even possible? There must be at least one plugin active if ds has been configured, so if this plugin is called when there none active shouldn't it assert instead since it'd point to a user trying to call deepspeed w/o configuring it?

otherwise what is a user to do with `None`?

src/accelerate/accelerator.py
muellerzr: It does/people tend to use `deepspeed_plugin`, however `.deepspeed_plugin` can refer to the first or active one and `.deepspeed_plugins` can be the full list.

stas00: What you propose would be somewhat confusing but at the same time it's more backward compatible. But if it's well documented I think it works.

`.deepspeed_plugins` is unlikely to be used often if ever by end users.

But I'd ask for a 2nd opinion.

muellerzr: Agreed, in this case (as I was still 50/50) it would *only* be the active plugin, while users can look in `accelerator.deepspeed_plugins` for all of them.

muellerzr: Decided for this to be the active plugin, and then we have `deepspeed_plugins` aliased to grab them all

src/accelerate/accelerator.py
muellerzr: Resolving, added another for plugins

src/accelerate/accelerator.py
```diff
@@ -304,8 +312,11 @@ def __init__(
             mixed_precision = (
                 os.environ.get("ACCELERATE_MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
             )
-            deepspeed_plugin.set_mixed_precision(mixed_precision)
-            deepspeed_plugin.set_deepspeed_weakref()
+            # The first plugin is always the active one
+            for plugin in deepspeed_plugin:
+                plugin.set_mixed_precision(mixed_precision)
+            deepspeed_plugin[0].enable()
```
stas00, line 318:
```suggestion
            for plugin in deepspeed_plugin:
                plugin.set_mixed_precision(mixed_precision)
            # The first plugin is always the active one
            deepspeed_plugin[0].enable()
```

src/accelerate/utils/dataclasses.py
```diff
@@ -1098,6 +1098,7 @@ def __post_init__(self):
         if self.zero3_init_flag and not self.hf_ds_config.is_zero3():
             warnings.warn("DeepSpeed Zero3 Init flag is only applicable for ZeRO Stage 3. Setting it to False.")
             self.zero3_init_flag = False
+        self.enabled = False
```
stas00, line 1101: perhaps add a short note to tell where it gets enabled? just an idea if it resonates

src/accelerate/utils/deepspeed.py
```diff
@@ -19,6 +19,20 @@
...
+
+def get_active_deepspeed_plugin(state: AcceleratorState) -> DeepSpeedPlugin:
+    """
+    Returns the currently active DeepSpeedPlugin.
+
+    Raises:
+        ValueError: If DeepSpeed was not enabled and this function is called.
+    """
+    if state.distributed_type != DistributedType.DEEPSPEED:
+        raise ValueError("DeepSpeed was not enabled. This function should not be called.")
```
stas00, line 34: could this be made more user-friendly, when one asserts it helps to guide the user to how the issue can be fixed. "This function should not be called" is not actionable. Possibly:

> Can't retrieve the active DeepSpeed plugin since none were enabled. Use <fill in> to enable one.

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,121 @@
...
+                hf_ds_config=self.config_zero3,
+            ),
+        ]
+
+    def test_enable_disable(self):
+        ds_zero2, ds_zero3 = self.get_ds_plugins()
+        _ = Accelerator(
+            deepspeed_plugin=[ds_zero2, ds_zero3],
+        )
+        # Accelerator should flag the first plugin as enabled
```
stas00, line 79:
```suggestion
        # Accelerator constructor should automatically enable the first plugin 
```

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,121 @@
...
+        assert ds_zero2.enabled
+        assert not ds_zero3.enabled
+        ds_zero3.enable()
+        assert not ds_zero2.enabled
+        assert ds_zero3.enabled
+        ds_zero2.enable()
+        assert not ds_zero3.enabled
+        assert ds_zero2.enabled
+
+    def test_get_active_plugin(self):
```
stas00, line 89: this can be probably merged into the previous one as they do 90% the same things. again just a suggestion.

src/accelerate/utils/deepspeed.py
muellerzr: Went with:

```python
        raise ValueError(
            "Couldn't retrieve the active `DeepSpeedPlugin` as none were enabled. "
            "Please make sure that either `Accelerator` is configured for `deepspeed` "
            "and make sure that a `DeepSpeedPlugin` has been enabled (`plugin.enable()`) before calling this."
        )
```

Which gets the point across

stas00 (commented):
Testing: I think you want an actual test where you do fwd/bwd with 2 models. This is insufficient to test that it works correctly IMHO.

src/accelerate/utils/dataclasses.py
muellerzr: agreed :)

src/accelerate/utils/deepspeed.py
```diff
@@ -19,6 +19,24 @@
...
+    Returns the currently active DeepSpeedPlugin.
+
+    Raises:
+        ValueError: If DeepSpeed was not enabled and this function is called.
+    """
+    if state.distributed_type != DistributedType.DEEPSPEED:
+        raise ValueError(
+            "Couldn't retrieve the active `DeepSpeedPlugin` as none were enabled. "
+            "Please make sure that either `Accelerator` is configured for `deepspeed` "
+            "and make sure that a `DeepSpeedPlugin` has been enabled (`plugin.enable()`) before calling this."
```
stas00, line 37:
```suggestion
            "Couldn't retrieve the active `DeepSpeedPlugin` as none were enabled. "
            "Please make sure that either `Accelerator` is configured for `deepspeed` "
            "or make sure that the desired `DeepSpeedPlugin` has been enabled (`plugin.enable()`) before calling this function."
```

src/accelerate/utils/deepspeed.py
stas00: super!

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,114 @@
...
+        accelerator = Accelerator(deepspeed_plugin=[ds_zero2, ds_zero3])
+        # Using Zero-2 first
+        model1 = self.model_init()
+        optimizer = DummyOptim(model1.parameters())
+        scheduler = DummyScheduler(optimizer)
+
+        dataset = RegressionDataset()
+        dataloader = torch.utils.data.DataLoader(dataset, batch_size=1)
+        model1, optimizer, scheduler, dataloader = accelerator.prepare(model1, optimizer, scheduler, dataloader)
+        ds_zero3.enable()
```
lewtun, line 103: do i understand correctly that calling this after `model1` is prepared will shard the model with z3? in other words, if i want to use z2 for model2 i would do something like:

```python
model1, optimizer, scheduler, dataloader = accelerator.prepare(model1, optimizer, scheduler, dataloader)
ds_zero3.enable()
model2 = self.model_init()
ds_zero2.enable()
```

muellerzr: It will shard the following models with z2 if you call `.prepare`

For your case you'd do (realistically):

```python
# Zero2 requires real model and optimizer
ds_zero2.enable()
model1, optimizer, scheduler, dataloader = accelerator.prepare(model1, optimizer, scheduler, dataloader)
# Enable before model init so ZeroInit takes into effect
ds_zero3.enable()
model2 = self.model_init()
# Actually prepare with zero3
model2 = accelerator.prepare(model2)
```

src/accelerate/state.py
```diff
@@ -946,7 +950,10 @@ def initialized(self) -> bool:
     def __repr__(self):
         repr = PartialState().__repr__() + f"\nMixed precision type: {self.mixed_precision}\n"
         if self.distributed_type == DistributedType.DEEPSPEED:
-            repr += f"ds_config: {self.deepspeed_plugin.deepspeed_config}\n"
+            from accelerate.utils.deepspeed import get_active_deepspeed_plugin
```
BenjaminBossan, line 953: Can this not use `self.deepspeed_plugin`?

src/accelerate/state.py
```diff
@@ -900,7 +900,11 @@ def __init__(
                         os.environ["XLA_DOWNCAST_BF16"] = str(0)
                         self.downcast_bfloat = False
             elif os.environ.get("ACCELERATE_USE_DEEPSPEED", "false") == "true" and not cpu:
-                self.deepspeed_plugin = deepspeed_plugin
+                # Just incase a user manually creates an `AcceleratorState`/bypasses the `Accelerator`
+                if not isinstance(deepspeed_plugin, (list, tuple)):
```
BenjaminBossan, line 904: It's not intended for users to later add another DS plugin, right? I.e. they need to pass all they'll ever need directly during `__init__`. I'm asking because otherwise, a tuple would not allow appending new plugins later.

src/accelerate/accelerator.py
```diff
@@ -1640,7 +1673,7 @@ def _prepare_deepspeed(self, *args):
 
             ds_initialize = msamp_deepspeed.initialize
 
-        deepspeed_plugin = self.state.deepspeed_plugin
+        deepspeed_plugin = get_active_deepspeed_plugin(self.state)
```
BenjaminBossan, line 1676: Can use `self.deepspeed_plugin` instead.

src/accelerate/accelerator.py
```diff
@@ -541,6 +557,23 @@ def __init__(
 
         check_os_kernel()
 
+    @property
+    def deepspeed_plugin(self):
+        """
+        Returns the currently active DeepSpeedPlugin.
```
BenjaminBossan, line 563: Mention that this can return `None` if DS is not being used.

src/accelerate/state.py
```diff
@@ -1092,6 +1102,20 @@ def local_main_process_first(self):
         with PartialState().local_main_process_first():
             yield
 
+    @property
+    def deepspeed_plugin(self):
+        """
+        Returns the currently active DeepSpeedPlugin.
```
BenjaminBossan, line 1108: Mention that this can return `None` if DS is not being used.

src/accelerate/accelerator.py
```diff
@@ -179,9 +180,10 @@ class Accelerator:
             the execution on one process only.
         dataloader_config (`DataLoaderConfiguration`, *optional*):
             A configuration for how the dataloaders should be handled in distributed scenarios.
-        deepspeed_plugin ([`~utils.DeepSpeedPlugin`], *optional*):
+        deepspeed_plugin ([`~utils.DeepSpeedPlugin`] or list of [`~utils.DeepSpeedPlugin`], *optional*):
```
BenjaminBossan, line 183: Type annotation below also needs to be adjusted.

BenjaminBossan (changes requested):
Thanks for pushing this topic forward. I don't have expertise on DeepSpeed, so I can't really comment on the parts that are DS-implementation specific. But I left a couple of comments regarding how accelerate handles this.

It's not clear to me yet in what circumstances I should create multiple `Accelerator` instances and when I should instead pass multiple DS plugins with a single instance. Do you plan on adding documentation on using multiple DS plugins?

muellerzr:
@BenjaminBossan yes, but the TL;DR:

Am I training multiple models? For now, create multiple Accelerators.
If I'm not (use case: using Zero-3 for a sharded reference model, Zero2 for training/optimize the active model since it's faster than Zero3), create multiple plugins

src/accelerate/state.py
muellerzr: Correct

src/accelerate/utils/dataclasses.py
```diff
@@ -1284,6 +1287,22 @@ def set_moe_leaf_modules(self, model):
                     transformer_moe_cls.append(transformer_cls)
             set_z3_leaf_modules(model, transformer_moe_cls)  # z3_leaf
 
+    def enable(self):
```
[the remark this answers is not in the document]
muellerzr: We've toyed with such APIs, and they don't make sense with how DS works right now. 

I also don't think we should fully tie it to the Accelerator, as the accelerator is a "generic" API for things, this is hyper-specific for DeepSpeed. 

Re; users may call `.enabled=True`, I'll put some blockers in place to avoid users doing the wrong thing. 

Re; when to set it, yeah we probably should move this to the `AcceleratorState` itself.

Re; dict: on paper, that's fine. But it's a very new pattern which breaks backward compatibility, and the overall API with how things are **currently** makes it very clunky, because we cannot prepare multiple models/engines at once with a singular Accelerator.

Once we can do that (aka no need for two `Accelerator`'s to do two models), I am 100% behind `dict` support.

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,114 @@
...
+            WORLD_SIZE="1",
+        )
+
+        self._test_file_path = inspect.getfile(self.__class__)
+        path = Path(self._test_file_path).resolve()
+        self.test_file_dir_str = str(path.parents[0])
+
+        self.ds_config_file = dict(
+            zero2=f"{self.test_file_dir_str}/ds_config_zero2.json",
+            zero3=f"{self.test_file_dir_str}/ds_config_zero3_model_only.json",
```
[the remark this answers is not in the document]
muellerzr: It's necessary for this particular test. Zero2 requires an optimizer and scheduler, zero3 can be used for inference/just the model, which is what this is testing

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+        # use self.get_config_dict(stage) to use these to ensure the original is not modified
```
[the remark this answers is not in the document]
muellerzr: I think that was leftover copy/paste 😅

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,114 @@
...
+        with self.assertLogs(level="WARNING") as captured:
+            model2 = accelerator.prepare(model2)
+            self.assertIn(
+                "A wrapped DeepSpeed engine reference is currently tied for this `Accelerator()` instance.",
+                captured.output[0],
+            )
+
+        assert accelerator.deepspeed_engine_wrapped.engine is model1
+
+    def test_train_multiple_models(self):
```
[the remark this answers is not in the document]
muellerzr: There now! :D

src/accelerate/utils/dataclasses.py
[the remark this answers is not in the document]
stas00: > Re; dict: on paper, that's fine. But it's a very new pattern which breaks backward compatibility, and the overall API with how things are currently makes it very clunky, because we cannot prepare multiple models/engines at once with a singular Accelerator. 

It'd work if you make it dual: list and dict. The init 2+ plugins is new anyway, so you can have it init'ed as a dict - the bonus is that it'd be much easier to retrieve the right plugin from the Accelerator and not requiring the user to have their own object they need to pass through to many calls.

So you can return a list of plugins, or a dict, or a specific plugin by its key.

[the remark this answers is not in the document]
muellerzr: What I'm stuck on is what is the key if the user doesn't pass in a dict? Do we do them ordered as the key? (so key `0` gets the first plugin, key `1` gets the second, ...)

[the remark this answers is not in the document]
stas00: since the original proposal of yours didn't enforce or prescribe any particular order, then the sorting doesn't matter as long as it's consistent. So just using the standard `d.values()` should sort it consistently.

As I suggested elsewhere, I think `deepspeed_plugins` is unlikely to be used since what would be the point if you don't know what they are - how would you know plugin is which?

But `deepspeed_plugin(key="teacher")` vs `deepspeed_plugin(key="student")` is very useful and most likely will be the way users will retrieve the desired plugin in their code.

Based on that probably just using the dict interface is sufficient and there is no need for a list at all.

This is a new API, so I'm not sure why you said it'd be a BC issue. If `deepspeed_plugin` continues returning the active plugin then everything should just work as before with a single plugin.

before w/ a single plugin:
```
accelerate.deepspeed_plugin().foo()
```
after w/ a single plugin:
```
accelerate.deepspeed_plugin().foo() # returns active plugin
```
after w/ multiple plugins:
```
accelerate.deepspeed_plugin(key="teacher").foo() # returns named plugin
accelerate.deepspeed_plugin(key="student").foo() # returns named plugin
```
and as I suggested elsewhere a magical accessor will make it even nicer:
```
accelerate.deepspeed_plugin.teacher.foo()
accelerate.deepspeed_plugin.student.foo()
```

so the only problem with the last one is that I'm using the same `deepspeed_plugin` as an attribute and as a method - so I don't think this would work and would require a different accessor name.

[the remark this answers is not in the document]
muellerzr: This would wind up being quite a bit breaking, as all HF codebases that use accelerate for DS do `accelerate.deepspeed_plugin` rather than `accelerate.deepspeed_plugin()`.  (and examples, etc)

If we go functional, I'd rather do `accelerator.deepspeed_plugins(key="teacher")`.

*Or* this waits until after 1.0 for this particular side of the API. Are you willing to wait a few weeks before introducing it? :)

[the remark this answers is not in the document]
stas00: `accelerator.deepspeed_plugins(key="teacher")` always returning a singular but named plural is not ideal.

ok, what about any plugin object knowing how to get other objects? e.g. 
```
accelerate.deepspeed_plugin.teacher.foo()
```
so it'd `getattr` and find `teacher` in `deepspeed_plugins`? Too magical?

> Or this waits until after 1.0 for this particular side of the API. Are you willing to wait a few weeks before introducing it? 

Personally I'm totally not in a rush, I'm ramping up on inference at the moment.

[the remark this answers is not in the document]
stas00: so:
```
# return current active plugin
accelerate.deepspeed_plugin

# return a specific named plugin
accelerate.deepspeed_plugin.teacher
accelerate.deepspeed_plugin.student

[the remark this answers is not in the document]
Comment (muellerzr) [reply src/accelerate/utils/dataclasses.py:1290]: Feels a bit too magical. I don't expect (without being explicit) the `deepspeed_plugin` to interact with the rest. If it were `deepspeed_plugins` then, and only then, I would. 

Or: `deepspeed_plugin_container` if we wanted to get very explicit.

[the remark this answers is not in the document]
Comment (muellerzr) [reply src/accelerate/utils/dataclasses.py:1290]: Even with being explicit, I'm still not sure such magic should exist in Accelerate, which is aimed to be non magical as much as possible (even if convenient).

Requiring users maintain their `DeepSpeedPlugins` IMO isn't a terrible requirement when working on multi-plugin based tasks, + a k/v system to access if needed.

The `Accelerator` shouldn't have something like `accelerator.get_deepspeed_plugin(name)` however I could be okay with:

```python
accelerator.state.get_deepspeed_plugin(key)
```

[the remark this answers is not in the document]
stas00: > Requiring users maintain their DeepSpeedPlugins IMO isn't a terrible requirement when working on multi-plugin based tasks, + a k/v system to access if needed.

I disagree here, because it then requires bouncing yet another function argument between potentially many calls. Being able to get it from state which is a global is a way easier.

> `accelerator.state.get_deepspeed_plugin(key)`

works for me.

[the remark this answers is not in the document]
stas00: as I suggested earlier I recommend asking those users who asked for this feature to kick the tires once you are happy with your work and see how well it works for them.

The biggest challenge here is to think of the various use cases and ensure they are all catered to.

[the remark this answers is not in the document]
muellerzr: Okay pushed new changes, I think this fixes most of the code smell for you @BenjaminBossan. We still need to check `self`, but users can't do `disable()` now, they *must* pick an active plugin to use at all times, and just swap the active one out.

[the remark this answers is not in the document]
muellerzr: @stas00 lmk if you have any other thoughts post seeing the new push

src/accelerate/accelerator.py
```diff
@@ -316,10 +312,24 @@ def __init__(
...
+                # Can either be a list or a dict at this point:
+                # list if we haven't initialized the plugins yet
+                # dict if we have
+                for plugin in deepspeed_plugin:
+                    if not isinstance(plugin, DeepSpeedPlugin):
+                        plugin = deepspeed_plugin[plugin]
+                    plugin.set_mixed_precision(mixed_precision)
+                if isinstance(deepspeed_plugin, (list, tuple)):
+                    # The first plugin is always the active one
+                    deepspeed_plugin[0].enable()
```
stas00, line 328: aren't you missing `enable` in the case of dict?

src/accelerate/accelerator.py
```diff
@@ -561,20 +571,10 @@ def deepspeed_plugin(self):
         """
         Returns the currently active DeepSpeedPlugin.
 
-        If using multiple plugins, the first one will be the active one by default. Manually call `plugin.enable()` to
-        activate a different plugin.
-
```
stas00: where did this doc go to in the last version?

src/accelerate/state.py
```diff
@@ -1112,6 +1109,14 @@ def deepspeed_plugin(self):
 
         return get_active_deepspeed_plugin(self)
 
+    def get_deepspeed_plugin(self, plugin_key: str):
```
stas00, line 1112: `plugin.plugin_key` is awkward - perhaps `plugin.name`?
```suggestion
    def get_deepspeed_plugin(self, name: str):
```
the fact that it's used as a dict key is irrelevant and doesn't have to be exposed in the api.

src/accelerate/accelerator.py
muellerzr: My thoughts here are:

Users set the name as part of the config object itself. Rather than having many ways to do one thing, I think it'd make sense to have users just set the name only during the config, and users still passing a list/tuple. They're already passing in a configured DS plugin anyways.

src/accelerate/utils/dataclasses.py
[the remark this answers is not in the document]
BenjaminBossan: I like the new changes, but since I've never had to deal with this use case, please don't put too much weight on my perspective here :)

One thing I would have implemented differently is that I would have had something like a `_active_deepspeed_plugin_key` or so on the `state` instead of an `enabled` attribute on the plugin. The `_active_deepspeed_plugin_key` would always be one of the keys from the dict (or `None` if there is no DS plugin). This would be in line with the idea of making "illegal states unrepresentable", i.e. it would be impossible to have multiple enabled DS plugins or that none are enabled.

Maybe there is a good reason not to do it like this that I'm missing (say, in the future we want to allow multiple active DS plugins), in that case please ignore my comment.

src/accelerate/accelerator.py
```diff
@@ -179,9 +179,10 @@ class Accelerator:
             the execution on one process only.
         dataloader_config (`DataLoaderConfiguration`, *optional*):
             A configuration for how the dataloaders should be handled in distributed scenarios.
-        deepspeed_plugin ([`~utils.DeepSpeedPlugin`], *optional*):
+        deepspeed_plugin ([`~utils.DeepSpeedPlugin`] or list of [`~utils.DeepSpeedPlugin`], *optional*):
```
BenjaminBossan, line 182: This should now also mention the dict possibility (same with the type annotation below). I would also like to see a bit more explanation on what happens if there is more than one plugin. I think, ideally, once this is properly added to the accelerate docs, a reference could be added here instead of inflating the docstring, as I imagine most users don't need that option.

src/accelerate/accelerator.py
muellerzr: Okay I'm convinced the other way, you have to specify it as a `dict`, you can't set it as part of the plugin config :)

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+There are a few use cases where one may want to run multiple models when using `accelerate` and `DeepSpeed`:
+
+* Knowledge distillation
+* RLHF techniques (see the `trl` repo for [more examples](https://github.com/huggingface/trl))
+* Training multiple models at once.
+
+As it stands, Accelerate has a **very experimental API** to help you accomplish this. 
+
+This tutorial will focus on two common use cases:
+1. Using `zero2` to train a model, and `zero3` to perform inference for assisting of the training for the `zero2` model (as `zero2` is faster to train on)
```
BenjaminBossan, line 35: Mention "RLHF" to make it clear that this example refers to the next section.

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+
+Normally, you would use a single [`utils.DeepSpeedPlugin`] at a time, however in this case we have two seperate configurations. Accelerate allows you to create and use multiple plugins **if and only if** they are in a `dict` so that we can reference/enable the proper plugin when needed:
+
+```python
+from accelerate.utils import DeepSpeedPlugin
+
+zero2_plugin = DeepSpeedPlugin(hf_ds_config="zero2_config.json")
+zero3_plugin = DeepSpeedPlugin(hf_ds_config="zero3_config.json")
+
+deepspeed_plugins = {"student": zero2_plugin, "teacher": zero3_plugin, }
```
BenjaminBossan, line 48:
```suggestion
deepspeed_plugins = {"student": zero2_plugin, "teacher": zero3_plugin}
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+When doing so,  `zero2_config.json` should be configured for full training (so specifying `scheduler` and `optimizer` if not utilizing your own) while `zero3_config.json` should *only* be configured for the model,
```
BenjaminBossan, line 51:
```suggestion
When doing so, `zero2_config.json` should be configured for full training (so specifying `scheduler` and `optimizer` if not utilizing your own) while `zero3_config.json` should *only* be configured for the model,
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+Now let's see how to use them.
+
+### The student model
+
+By default, `accelerate` will set the first item in the `dict` as the default/enabled plugin (so here our `"student"` plugin). We can verify this by using the [`utils.deepspeed.get_active_deepspeed_plugin]` function to see which one is being used:
+
+```python
+active_plugin = get_active_deepspeed_plugin(accelerator.state)
```
BenjaminBossan, line 100: Can we not use `accelerator.state.deepspeed_plugin` or `accelerator.deepspeed_plugin`. Or is the idea to make it more explicit?

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+### The teacher model
+
+First we need to tell the `Accelerator` that the zero3/teacher ZeRO configuration should be used:
+
+```python
+accelerator.state.enable_deepspeed_plugin("teacher")
+```
+
+Doing so will disable the `"student"` plugin and enable the `"teacher"` one instead. This will update the
+DeepSpeed stateful config inside of `transformers`, and change which plugin configuration gets called when using
```
BenjaminBossan, line 122: It's not clear how `transformers` comes into play here. What if the user is not using transformers? It should still work, right?

docs/source/usage_guides/deepspeed_multiple_model.md
BenjaminBossan, line 1: Very nice examples, thanks for putting in the effort. I leave it to others to comment on the concrete examples, as I didn't have those use cases yet.

A small comment: There are some line breaks here and there that are not necessary, e.g. lines 35-36. But no big deal.

BenjaminBossan (approved):
Nicely done, implementation-wise I don't have anything to add, just some nits.

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: Idea was to make it more explicit. (I'll add a comment about `accelerator.deepspeed_plugin` though)

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: Yes and no. If not using `transformers` they need to manually call `ZeroInit`, which is what this is talking about

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+    training this particular model
+
+</Tip>
+
+From here, we can create a single [`Accelerator`] and pass in both configurations:
+
+```python
+from accelerate import Accelerator
+
+accelerator = Accelerator(deepspeed_plugin=deepspeed_plugins)
```
stas00, line 91: can this be fixed to `deepspeed_plugins` as the key, otherwise this is a poor API.

Perhaps for BC it could be either?

src/accelerate/accelerator.py
```diff
@@ -304,8 +312,16 @@ def __init__(
...
                 os.environ.get("ACCELERATE_MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
             )
-            deepspeed_plugin.set_mixed_precision(mixed_precision)
-            deepspeed_plugin.set_deepspeed_weakref()
+            if not isinstance(deepspeed_plugin, dict):
+                deepspeed_plugin.set_mixed_precision(mixed_precision)
+                deepspeed_plugin.enable()
+            else:
+                for plugin in deepspeed_plugin.values():
+                    plugin.set_mixed_precision(mixed_precision)
```
SunMarc, line 320: can two deepspeed plugins have different mixed precision ? Not sure about the use case also, just asking

src/accelerate/accelerator.py
```diff
@@ -280,16 +281,23 @@ def __init__(
...
         else:
-            assert isinstance(
-                deepspeed_plugin, DeepSpeedPlugin
-            ), "`deepspeed_plugin` must be an `accelerate.utils.DeepSpeedPlugin` object."
+            if isinstance(deepspeed_plugin, dict):
+                for plugin in deepspeed_plugin.values():
+                    if not isinstance(plugin, DeepSpeedPlugin):
+                        raise TypeError("`deepspeed_plugin` must be a DeepSpeedPlugin object.")
+
+        if deepspeed_plugin is not None:
```
SunMarc, line 299: maybe raise a warning if the user pass another deepspeed plugin in the second accelerator ?

SunMarc (approved):
Thanks for the huge work @muellerzr ! Mostly went through the examples and the tutorial. Just in case (even though dict are ordered for python >=3.7), I think we should advise the user to always enable the deepspeed plugin before preparing the model `accelerator_0.state.enable_deepspeed_plugin("model1")`.

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+From here, we need to deal with the teacher model.
+
+### The teacher model
+
+First we need to tell the `Accelerator` that the zero3/teacher ZeRO configuration should be used:
+
+```python
+accelerator.state.enable_deepspeed_plugin("teacher")
```
stas00, line 124: yeah, now reading the demo, I think I agree with the earlier discussion - `enable` is a bad terminology for something that requires only 1 of many state. Enable implies that more than one thing can be enabled at the same time, but this is not the case.

Toggle could work for 2 objects. But then we may have more than 2 models, in which case toggle doesn't fit either.

So probably `select` is the right verb here, as it implies one among many.

But yet you did say in the docs that only one plugin can be enabled at any given time... not sure. I just think select is more self-documenting.

src/accelerate/test_utils/scripts/external_deps/test_ds_multiple_model.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # Now prepare the model under zero3 plugin
+    accelerator.state.enable_deepspeed_plugin("inference")
+    assert get_active_deepspeed_plugin(accelerator.state) is zero3_plugin
+    inference_model = NoiseModel()
+    inference_model = accelerator.prepare(inference_model)
+    inference_model.eval()
+
+    # Run training loop
+    accelerator.state.enable_deepspeed_plugin("training")
```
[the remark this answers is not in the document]
muellerzr: On the accelerate side, not particularly much/anything at all, since it's mostly for `deepspeed.init()`. But for the transformers side eventually I believe this should run into some errors between zero2 and zero3, since with zero3 we need to gather params, and this sets a flag that tells transformers we're using zero3 right now.

src/accelerate/accelerator.py
muellerzr: Good question. I suppose technically they could, e.g. one has the model in bf16 and another in float32. 

A direct use case would be the zero3 model is in fp8 vs the trained model in bf16 (`te` fp8 specifically).

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+Doing so will disable the `"student"` plugin and enable the `"teacher"` one instead. This will update the
+DeepSpeed stateful config inside of `transformers`, and change which plugin configuration gets called when using
+`deepspeed.initialize()`, allowing us to use the no-code `Zero3Init` `deepspeed` context manager `transformers` provides:
```
stas00, line 129:
```suggestion
`deepspeed.initialize()`, allowing us to use the no-code `deepspeed.zero.Init`  context manager `transformers` provides:
```

stas00: Also, I don't know what "no-code" means

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+* Training multiple models at once.
+
+As it stands, Accelerate has a **very experimental API** to help you accomplish this.
+
+This tutorial will focus on two common use cases:
+
+1. Using `zero2` to train a model, and `zero3` to perform inference for assisting of the training for the `zero2` model (as `zero2` is faster to train on), which is shown in the RLHF example
+2. Training multiple *disjoint* models at once, shown later
+
+## RLHF (multiple models, only one is trained)
```
stas00, line 39:
```suggestion
## Multiple models, only one is trained

RLHF is a good example of such use-case
```

The documentation is about many-models-one-trained, not RLHF - so proposing ot re-arrange this

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+    accelerator.backward(loss)
+    optimizer.step()
+    scheduler.step()
+    optimizer.zero_grad()
+```
+
+## Multiple models all being trained
+
+Multiple models being trained is a more complicated scenario and is actively being looked upon by the team.
+In its current state, we assume that each model **completely disjoint** from the other during training.
```
stas00, line 164:
```suggestion
In its current state, we assume that each model is **completely disjoint** from the other during training.
```

perhaps add a note or an example of what you mean by disjoint?

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+
+With this situation, we still require two [`utils.DeepSpeedPlugin`]'s to be made, however we also further
+require a second `Accelerator`, since we need to call the proper `deepspeed` engine, and `Accelerator` only
```
stas00, line 167:
```suggestion
require a second `Accelerator`, since we need to call different `deepspeed` engines at different times, and a single `Accelerator` only
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+accelerator_0 = Accelerator(deepspeed_plugin=deepspeed_plugins)
+accelerator_1 = Accelerator()
+```
+
+Similar to before, we can call either `accelerator_0.state.enable_deepspeed_plugin()` to enable/disable
+a particular plugin, and call `.prepare` like normal:
+
+```python
+# can be `accelerator_0`, `accelerator_1`, or by calling `AcceleratorState().enable_deepspeed_plugin(...)`
+accelerator_0.state.enable_deepspeed_plugin("model1")
```
stas00, line 183: if you are using the `_\d` pattern can we use it consistently? i.e. `model_1`, etc? or w/o `_` but then again consistently w/o it.

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,203 @@
...
+With this situation, we still require two [`utils.DeepSpeedPlugin`]'s to be made, however we also further
+require a second `Accelerator`, since we need to call the proper `deepspeed` engine, and `Accelerator` only
+carries one instance of it at a time.
+
+Since the [`state.AcceleratorState`] is a stateful object though, it is already aware of both `DeepSpeedPlugin`'s available,
+meaning you can just instantiate a second `Accelerator` with no extra arguments:
+
+```python
+accelerator_0 = Accelerator(deepspeed_plugin=deepspeed_plugins)
+accelerator_1 = Accelerator()
```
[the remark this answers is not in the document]
muellerzr: Well, now we'll raise an error :)

docs/source/usage_guides/deepspeed_multiple_model.md
stas00: i think the accounting is weird too, you have model0 and model1 and model2 - something is off here - you should have only 2 models here, no?

src/accelerate/accelerator.py
muellerzr: Though at first glance... quite a challenge. I'll leave this to a future PR/enhancement

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,215 @@
...
+    accelerator_0.backward(outputs_1.loss)
+    optimizer_0.step()
+    scheduler_0.step()
+    optimizer_0.zero_grad()
+    outputs_2 = model_1(**batch)
+    accelerator_1.backward(outputs_2.loss)
+    optimizer_1.step()
+    scheduler_1.step()
+    optimizer_1.zero_grad()
+```
```
stas00, line 210: I find this example very difficult to follow since the numbers don't match.

perhaps once you explained why this is so and they have to be that way, non-numerical differentiator would work better? perhaps the same teacher/student naming would be more appropriate?

also we are facing humans here, model_0 and model_1 isn't how we think - we think model_1 and model_2 instead. I'd suggest to leave 0-indexing to computer indices.

stas00: also adding some vertical white space to group things together would help the readability in this last code section.

src/accelerate/test_utils/scripts/external_deps/test_ds_multiple_model.py
```diff
@@ -0,0 +1,328 @@
...
+
+    deepspeed_plugins = {"zero2": zero2_plugin, "zero3": zero3_plugin}
+
+    # Initialize accelerator
+    zero2_accelerator = Accelerator(
+        deepspeed_plugin=deepspeed_plugins,
+        mixed_precision="bf16",
+    )
+
+    zero3_accelerator = Accelerator()
```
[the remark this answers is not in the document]
muellerzr: For rn we assume the same mixed precision as noted earlier, will look towards configuring multiple precision types later

src/accelerate/accelerator.py
stas00: that would require 2 accelerate configs.

docs/source/usage_guides/deepspeed_multiple_model.md
tjruwase: This is a DS bug due to sanity checking never getting updated for inference. Please share a repro and issue. Thanks!

stas00:
@muellerzr, where is `  zero3_init_flag: true` treating in this PR? As surely it's possible not all models will want the same treatment.

muellerzr:
@stas00 it's done in that tweak to when we create the HF config file (notice how it's done always, rather than just under zero3). We update the deepspeed config reference transformers looks at. See here: https://github.com/huggingface/accelerate/pull/3097/files#diff-bcbdd609996df7224cc9c21ecd97092122d45263d1323892474021d510ec1eefR1199-R1224

And then the first thing `enable` does is set the reference on that plugin

src/accelerate/accelerator.py
muellerzr: Yep, which rn we don't/can't do

src/accelerate/test_utils/scripts/external_deps/test_ds_multiple_model.py
```diff
@@ -0,0 +1,328 @@
...
+    best_performance = 0
+    metric = evaluate.load("glue", "mrpc")
+    performance_metric = {}
+    for epoch in range(starting_epoch, num_epochs):
+        model.train()
+        inference_model.train()
+        for step, batch in enumerate(train_dataloader):
+            with accelerator.accumulate(model):
+                outputs_1 = model(**batch)
+                outputs_2 = inference_model(outputs_1.loss)
```
stas00, line 158: FWIW, I often try to read tests as a practical example since docs usually fall out of sync (or don't exist), so perhaps we can use these tests as the guide - you can quote relevant parts and then send the reader to see the full working program which they can easily adopt to their needs.

That way you only need to main a single readable working example and not two and everybody benefits - the maintainers and the users.

If it resonates then let's make this test easy to read, that is something like:

```suggestion
                train_outputs = train_model(**batch)
                infrence_outputs = inference_model(outputs_1.loss)
```

stas00: so not mixing numbers and string identifiers but using them consistently.

tests/deepspeed/test_deepspeed_multiple_model.py
```diff
@@ -0,0 +1,152 @@
...
+        model1, optimizer, scheduler, dataloader = accelerator.prepare(model1, optimizer, scheduler, dataloader)
+        accelerator.state.enable_deepspeed_plugin("zero3")
+        model2 = self.model_init()
+        with self.assertLogs(level="WARNING") as captured:
+            model2 = accelerator.prepare(model2)
+            self.assertIn(
+                "A wrapped DeepSpeed engine reference is currently tied for this `Accelerator()` instance.",
+                captured.output[0],
+            )
+
```
stas00, line 142: just flagging that this test is missing the crucial part of actually running training.

muellerzr: Correct, that's for the later test/the script

src/accelerate/utils/dataclasses.py
```diff
@@ -1196,13 +1199,13 @@ def set_mixed_precision(self, mixed_precision):
     def set_deepspeed_weakref(self):
```
stas00, line 1199: let's move the discussion on zero.Init / zero3_init_flag flag here - so it's self-contained.

Yes, I'm aware of this - but I don't see it being tested in the new multi-model tests. how do you know that it works correctly?

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: Will move forward with `.select()` then. If we get to a point where for some reason we have multiple active, we can deprecate and change it back (or just add)

src/accelerate/utils/dataclasses.py
muellerzr: Thanks. 

And true, will do a very quick test rn. Thanks for catching it

muellerzr: added tests, and this does indeed work :)

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: Fixed :)

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: Went with `first_` and `second_` to be more readable (and the prior example used words)

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+* Knowledge distillation
+* RLHF techniques (see the `trl` repo for [more examples](https://github.com/huggingface/trl))
+* Training multiple models at once.
+
+As it stands, Accelerate has a **very experimental API** to help you accomplish this.
+
+This tutorial will focus on two common use cases:
+
+1. Using `zero2` to train a model, and `zero3` to perform inference for assisting of the training for the `zero2` model (as `zero2` is faster to train on if the model fits onto a single GPU), which is shown in the RLHF example
+2. Training multiple *disjoint* models at once, shown later
```
stevhliu, line 37:
```suggestion
Running multiple models with Accelerate and DeepSpeed is useful for:

* Knowledge distillation
* RLHF techniques (see the [TRL](https://github.com/huggingface/trl) library for more examples)
* Training multiple models at once

Currently, Accelerate has a **very experimental API** to help you use multiple models.

This tutorial will focus on two common use cases:

1. An RLHF use case, using ZeRO2 to train a model and ZeRO3 for inference to assist in training the ZeRO2 model. ZeRO2 is faster to train with if the model fits on a single GPU.
2. Training multiple *disjoint* models at once.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Multiple models, only one is trained
```
stevhliu, line 39:
```suggestion
## RLHF
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+RLHF is a good example of such use-case
+
+Normally, you would use a single [`utils.DeepSpeedPlugin`] at a time, however in this case we have two separate configurations. Accelerate allows you to create and use multiple plugins **if and only if** they are in a `dict` so that we can reference/enable the proper plugin when needed:
```
stevhliu, line 43:
```suggestion
RLHF is a good example of using multiple models, but only training one of them.

Normally, you would use a single [`utils.DeepSpeedPlugin`] at a time. However, in this case, there are two separate configurations. Accelerate allows you to create and use multiple plugins **if and only if** they are in a `dict` so that you can reference and enable the proper plugin when needed.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+from accelerate.utils import DeepSpeedPlugin
+
+zero2_plugin = DeepSpeedPlugin(hf_ds_config="zero2_config.json")
+zero3_plugin = DeepSpeedPlugin(hf_ds_config="zero3_config.json")
+
+deepspeed_plugins = {"student": zero2_plugin, "teacher": zero3_plugin}
+```
+
+When doing so,  `zero2_config.json` should be configured for full training (so specifying `scheduler` and `optimizer` if not utilizing your own) while `zero3_config.json` should *only* be configured for the inference model,
+such as the following example for a `zero3_config.json`:
```
stevhliu, line 55:
```suggestion
The `zero2_config.json` should be configured for full training (so specify `scheduler` and `optimizer` if you are not utilizing your own), while `zero3_config.json` should only be configured for the inference model, as shown in the example below.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+        "stage3_prefetch_bucket_size": "auto",
+        "stage3_param_persistence_threshold": "auto",
+        "stage3_max_live_parameters": "auto",
+        "stage3_max_reuse_distance": "auto",
+    },
+    "train_micro_batch_size_per_gpu": 1
+}
+```
+
+And the zero2 counterpart:
```
stevhliu, line 75:
```suggestion
An example `zero2_config.json` configuration is shown below.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+    "gradient_clipping": "auto",
+    "train_batch_size": "auto",
+    "train_micro_batch_size_per_gpu": "auto",
+}
+```
+
+<Tip>
+
+    DeepSpeed will raise an error if we don't specify a `train_micro_batch_size_per_gpu`, despite not really 
+    training this particular model
```
stevhliu, line 116:
```suggestion
    DeepSpeed will raise an error if `train_micro_batch_size_per_gpu` isn't specified, even if this particular model isn't being trained.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+</Tip>
+
+From here, we can create a single [`Accelerator`] and pass in both configurations:
```
stevhliu, line 120:
```suggestion
From here, create a single [`Accelerator`] and pass in both configurations.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+```python
+from accelerate import Accelerator
+
+accelerator = Accelerator(deepspeed_plugins=deepspeed_plugins)
+```
+
+Now let's see how to use them.
+
+### The student model
```
stevhliu, line 130:
```suggestion
### Student model
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+By default, `accelerate` will set the first item in the `dict` as the default/enabled plugin (so here our `"student"` plugin). We can verify this by using the [`utils.deepspeed.get_active_deepspeed_plugin]` function to see which one is being used:
```
stevhliu, line 132:
```suggestion
By default, Accelerate sets the first item in the `dict` as the default or enabled plugin (`"student"` plugin). Verify this by using the [`utils.deepspeed.get_active_deepspeed_plugin]` function to see which plugin is enabled.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+```python
+active_plugin = get_active_deepspeed_plugin(accelerator.state)
+assert active_plugin is deepspeed_plugins["student"]
+```
+
+Also the `AcceleratorState` will keep the active DeepSpeed plugin saved in `state.deepspeed_plugin`:
```
stevhliu, line 139:
```suggestion
[`AcceleratorState`] also keeps the active DeepSpeed plugin saved in `state.deepspeed_plugin`.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+```python
+assert active_plugin is accelerator.deepspeed_plugin
+```
+
+Since the `student` is the currently active plugin, let's go ahead and prepare our model, optimizer, and scheduler like normal:
```
stevhliu, line 144:
```suggestion
Since `student` is the currently active plugin, let's go ahead and prepare the model, optimizer, and scheduler.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+```python
+student_model, optimizer, scheduler = ...
+student_model, optimizer, scheduler, train_dataloader = accelerator.prepare(student_model, optimizer, scheduler, train_dataloader)
+```
+
+From here, we need to deal with the teacher model.
+
+### The teacher model
+
+First we need to tell the `Accelerator` that the zero3/teacher ZeRO configuration should be used:
```
stevhliu, line 155:
```suggestion
Now it's time to deal with the teacher model.

### Teacher model

First, you need to specify in [`Accelerator`] that the `zero3_config.json` configuration should be used.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+```python
+accelerator.state.select_deepspeed_plugin("teacher")
+```
+
+Doing so will disable the `"student"` plugin and enable the `"teacher"` one instead. This will update the
+DeepSpeed stateful config inside of `transformers`, and change which plugin configuration gets called when using
+`deepspeed.initialize()`, allowing us to use the no-code `deepspeed.zero.Init`  context manager `transformers` provides:
```
stevhliu, line 163:
```suggestion
This disables the `"student"` plugin and enables the `"teacher"` plugin instead. The
DeepSpeed stateful config inside of Transformers is updated, and it changes which plugin configuration gets called when using
`deepspeed.initialize()`. This allows you to use the no-code `deepspeed.zero.Init`  context manager Transformers provides.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+```python
+teacher_model = AutoModel.from_pretrained(...)
+teacher_model = accelerator.prepare(teacher_model)
+```
+
+Otherwise you should manually initialize them under `deepspeed.zero.Init`:
```
stevhliu, line 170:
```suggestion
Otherwise, you should manually initialize the model with `deepspeed.zero.Init`.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+```python
+with deepspeed.zero.Init(accelerator.deepspeed_plugin.config):
+    model = MyModel(...)
+```
+
+### Training
+
+From here, your training loop can be whatever you like, making sure that `teacher_model` is never being trained on:
```
stevhliu, line 178:
```suggestion
From here, your training loop can be whatever you like, as long as `teacher_model` is never being trained on.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+    output_student = student_model(**batch)
+    # Combine the losses or modify it in some way
+    loss = output_teacher.loss + output_student.loss
+    accelerator.backward(loss)
+    optimizer.step()
+    scheduler.step()
+    optimizer.zero_grad()
+```
+
+## Multiple models all being trained
```
stevhliu, line 195:
```suggestion
## Train multiple disjoint models
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+Multiple models being trained is a more complicated scenario and is actively being looked upon by the team.
+In its current state, we assume that each model is **completely disjoint** from the other during training.
+
+With this situation, we still require two [`utils.DeepSpeedPlugin`]'s to be made, however we also further
+require a second `Accelerator`, since we need to call different `deepspeed` engines at different times, and a single `Accelerator` only
+carries one instance of it at a time.
+
+Since the [`state.AcceleratorState`] is a stateful object though, it is already aware of both `DeepSpeedPlugin`'s available,
+meaning you can just instantiate a second `Accelerator` with no extra arguments:
```
stevhliu, line 205:
```suggestion
Training multiple models is a more complicated scenario.
In its current state, we assume each model is **completely disjointed** from the other during training.

This scenario still requires two [`utils.DeepSpeedPlugin`]'s to be made. However, you also need a second [`Accelerator`], since different `deepspeed` engines are being called at different times. A single [`Accelerator`] can only carry one instance at a time.

Since the [`state.AcceleratorState`] is a stateful object though, it is already aware of both [`utils.DeepSpeedPlugin`]'s available. You can just instantiate a second [`Accelerator`] with no extra arguments.
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+```python
+first_accelerator = Accelerator(deepspeed_plugins=deepspeed_plugins)
+second_accelerator = Accelerator()
+```
+
+Similar to before, we can call either `first_accelerator.state.select_deepspeed_plugin()` to enable/disable
+a particular plugin, and call `.prepare` like normal:
```
stevhliu, line 213:
```suggestion
You can call either `first_accelerator.state.select_deepspeed_plugin()` to enable or disable
a particular plugin, and then call [`prepare`].
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+    first_optimizer.zero_grad()
+    
+    outputs2 = model2(**batch)
+    second_accelerator.backward(outputs2.loss)
+    second_optimizer.step()
+    second_scheduler.step()
+    second_optimizer.zero_grad()
+```
+
+## More Resources
```
stevhliu, line 251:
```suggestion
## Resources
```

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,254 @@
...
+
+To see more examples, please check out the [related tests](https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/scripts/external_deps/test_ds_multiple_model.py) we currently have in `Accelerate`, and we will be improving this further
+with better examples, documentation, and functionality as we continue to flesh out this API
```
stevhliu, line 254:
```suggestion
To see more examples, please check out the [related tests](https://github.com/huggingface/accelerate/blob/main/src/accelerate/test_utils/scripts/external_deps/test_ds_multiple_model.py) currently in [Accelerate].
```

stevhliu (approved):
Very cool use case examples! Left some suggestions to reduce wordiness and be more direct, and made it clearer that the sections correspond to these use cases :)

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,246 @@
...
+## RLHF
+
+RLHF is a good example of using multiple models, but only training one of them.
+
+Normally, you would use a single [`utils.DeepSpeedPlugin`] at a time. However, in this case, there are two separate configurations. Accelerate allows you to create and use multiple plugins **if and only if** they are in a `dict` so that you can reference and enable the proper plugin when needed.
+
+```python
+from accelerate.utils import DeepSpeedPlugin
+
+zero2_plugin = DeepSpeedPlugin(hf_ds_config="zero2_config.json")
```
lewtun, line 46: Would be good to document somewhere if/how `accelerate launch` works when two configs are needed. As a user, I rely heavily on `accelerate launch --config_file=/path/to/deepspeed_zero{1,2,3}.yaml` and I wonder if we can enable something like:

```
# comma separated?
accelerate launch --config_file=deepspeed_zero2.yaml,deepspeed_zero3.yaml
```

And then the training code assumes the first plugin comes from the first config, etc

docs/source/usage_guides/deepspeed_multiple_model.md
```diff
@@ -0,0 +1,246 @@
...
+from accelerate import Accelerator
+
+accelerator = Accelerator(deepspeed_plugins=deepspeed_plugins)
+```
+
+Now let's see how to use them.
+
+### Student model
+
+By default, Accelerate sets the first item in the `dict` as the default or enabled plugin (`"student"` plugin). Verify this by using the [`utils.deepspeed.get_active_deepspeed_plugin]` function to see which plugin is enabled.
```
lewtun, line 128:
```suggestion
By default, Accelerate sets the first item in the `dict` as the default or enabled plugin (`"student"` plugin). Verify this by using the [`utils.deepspeed.get_active_deepspeed_plugin`] function to see which plugin is enabled.
```

lewtun (commented):
Really cool feature @muellerzr - this is going to unlock a ton of cool use cases for post-training LLMs! I only commented on the docs from the perspective of a user - overall looks great 🔥

docs/source/usage_guides/deepspeed_multiple_model.md
muellerzr: We can't do that currently, it is solely based on creating `DeepSpeedPlugin`'s manually.

muellerzr: Most probable, setting up the following:

```yaml
deepspeed_configs:
    config_1:
        deepspeed_config_file: ...
        name: ...
    config_2:
        deepspeed_config_file: ...
        name: ...
```

lewtun: Indeed, having a "master" config file would be nice to have. That way I can have my Z2/Z3 configs fixed and just toggle their use based on the task at hand

lzy37ld:
Thanks for this great effort!

I just have a question here (might be trivial): I still can not understand why training two disjoint models needs two accelerators while training the student model (teacher model is fixed) only needs one accelerator.

I see that each example would call the `.prepare()` function twice. From my current shallow understanding, the example about training two models need two optimizers. Is that the reason we need two accelerators?

PR #3099, opened 2024-09-10
[docs] Doc sprint

stevhliu:
Addresses the docs improvements internally discussed:

 - add headers for more classes/functions in the API docs so they're easier and faster to find
 - fix unsupported `<Note>` tags
 - align title in the `toctree` with the header referenced in the doc
 - remove 🤗 emoji to avoid cluttering


docs/source/package_reference/fsdp.md
```diff
@@ -13,12 +13,20 @@ specific language governing permissions and limitations under the License.
 rendered properly in your Markdown viewer.
 -->
 
-# Utilities for Fully Sharded Data Parallelism
+# Fully Sharded Data Parallel utilities
+
+## enable_fsdp_ram_efficient_loading
```
muellerzr, line 18: Wish we didn't have to do this everywhere but I suppose can't be helped :/

muellerzr (approved):
Nice! 🤩 

cc @SunMarc for the inference related docs

docs/source/usage_guides/big_modeling.md
```diff
@@ -66,42 +56,25 @@ model = load_checkpoint_and_dispatch(
...
-
-    Multiple GPUs can be utilized, however this is considered "model parallelism" and as a result only one GPU will be active at a given moment, waiting for the prior one to send it the output. You should launch your script normally with `python`
-    and not need `torchrun`, `accelerate launch`, etc.
+Each time an input is passed through a layer, it is sent from the CPU to the GPU (or disk to CPU to GPU), the output is calculated, and the layer is removed from the GPU going back down the line. While this adds some overhead to inference, it enables you to run any size model on your system, as long as the largest layer fits on your GPU.
 
-</Tip>
+Multiple GPUs, or "model parallelism", can be utilized but only one GPU will be active at any given moment. This forces the GPU to wait for the previous GPU to send it the output. You should launch your script normally with Python instead of other tools like torchrun and accelerate launch.
 
-For a visual representation of this, check out the animation below:
+<Youtube id="MWCSGj9jEAo"/>
```
SunMarc, line 75: Maybe we can add a link here to redirect to torch.distributed.pipelining section as it is relevant. torch.distributed.pipelining is basically doing the same thing as big model inference but it will use all gpus instead of one at a time. Of course it is less flexible.

SunMarc (approved):
Thanks for this sprint @stevhliu ! The big model inference part is a lot nicer now ! I left a few suggestions

docs/source/usage_guides/distributed_inference.md
```diff
@@ -144,22 +144,20 @@ You can find more complex examples [here](https://github.com/huggingface/acceler
...
 
 The general idea with pipeline parallelism is: say you have 4 GPUs and a model big enough it can be *split* on four GPUs using `device_map="auto"`. With this method you can send in 4 inputs at a time (for example here, any amount works) and each model chunk will work on an input, then receive the next input once the prior chunk finished, making it *much* more efficient **and faster** than the method described earlier. Here's a visual taken from the PyTorch repository:
 
 ![PiPPy example](https://camo.githubusercontent.com/681d7f415d6142face9dd1b837bdb2e340e5e01a58c3a4b119dea6c0d99e2ce0/68747470733a2f2f692e696d6775722e636f6d2f657955633934372e706e67)
 
 To illustrate how you can use this with Accelerate, we have created an [example zoo](https://github.com/huggingface/accelerate/tree/main/examples/inference) showcasing a number of different models and situations. In this tutorial, we'll show this method for GPT2 across two GPUs.
 
-Before you proceed, please make sure you have the latest pippy installed by running the following:
+Before you proceed, please make sure you have the latest PyTorch version installed by running the following:
```
stevhliu, line 155: I also updated the installation instructions here, let me know if its incorrect!

muellerzr: Nope it's correct :)

muellerzr: in a follow-up after this I'll fix the `PiPPy example` chunk so we don't have a merge conflict :)

PR #3102, opened 2024-09-11
MAINT: Permission for GH token in stale.yml

BenjaminBossan:
# What does this PR do?

See https://github.com/huggingface/peft/pull/2061 for the equivalent PR in PEFT.

This restores the functionality of the stale bot after permissions for the token have been limited. The action still shows errors for PEFT but the bot appears to work fine.

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

PR #3112, opened 2024-09-13
MNT Permission for PRs for GH token in stale.yml

BenjaminBossan:
Continuation of #3102.

The equivalent PR in PEFT (https://github.com/huggingface/peft/pull/2064) was successful to restore stale bot function to PRs as well. Hence also making the same change for accelerate.


muellerzr (approved):
Thanks!

PR #3116, opened 2024-09-17
typo of "scalar" instead of "scaler"

tonyzhaozh:
causes bug in the case of is_musa_available()


BenjaminBossan (approved):
Thanks for catching this and providing the fix, LGTM.

PR #3119, opened 2024-09-18
[tests] skip pippy tests for XPU 

faaany:
## What does this PR do?
XPU doesn't support pippy yet. So we should skip the pippy-related tests for now.   

@muellerzr @SunMarc


SunMarc (approved):
Thanks! LGTM

Issue #3078, opened 2024-09-04
[Community Contributions] examples on distributed inference using 🤗 Accelerate


sayakpaul:
The [`inference/distributed` directory](https://github.com/huggingface/accelerate/tree/main/examples/inference/distributed) houses examples on running distributed inference with `accelerate`:

* Phi2 for language generation 
* Stable Diffusion for image generation 

The strategy followed there is to load an entire model onto each GPU and sending chunks of a batch through each GPU’s model copy at a time. Synthetic data generation has become an essential toolkit for every ML Engineer. So, it'd be beneficial to extend these examples to include some more use cases:

* Image captioning 
* Speech data generation 

Some nice to haves:

* Include artifact serialization as done in [this](https://gist.github.com/sayakpaul/cfaebd221820d7b43fae638b4dfa01ba) 
* Keep the artifact serialization code under a thread to not block GPU execution

sayakpaul:
For image recaptioning, I have a quick and dirty PoC here that assumes streaming from `webdataset` and uses `multiprocessing` from torch: https://gist.github.com/sayakpaul/dfb26d6562b62ba8c2f613dc912a28a8. Would be nice to make it more efficient and leverage `accelerate` to make things cleaner and simpler.

VladOS95-cyber:
Hey @sayakpaul @muellerzr! If it is still actual, I am going to add one more example for Video captioning using LLaVA-NeXT-Video-7B-hf

sayakpaul:
Of course it's up, that would be so cool! Cc: @a-r-r-o-w

GANESH-NADKARNI:
Hi! I'd like to add an image captioning example using BLIP-2. Is this still open for contributions?

krishna2426:
is it still open?

Tweezky66:
hey @muellerzr @sayakpaul ! Is it still open,if yes,i would like to work on it

lokeshps321:
Hi @sayakpaul @muellerzr! Is this still available for people to contribute? I want to create a distributed batch inference example script using Qwen2.5 or Whisper for audio or Gemma 2. Let me know if you have a model that would be most useful!

LeonxLJX:
I'd like to take this one (`[Community Contributions] examples on distributed inference `). I'll dig into the root cause and follow up with a PR shortly. (claiming via @LeonxLJX)

LeonxLJX:
[...]

PR #3123, opened 2024-09-19
Florence2 distributed inference example

hlky:
# What does this PR do?

This PR adds a distributed inference example for Florence-2.

#3078 

## Notes

Based on https://gist.github.com/sayakpaul/dfb26d6562b62ba8c2f613dc912a28a8#file-generate_captions_multigpu-py and https://gist.github.com/sayakpaul/cfaebd221820d7b43fae638b4dfa01ba

- `nodesplitter=None` in `wds.WebDataset` seems to be an important implementation detail, by default I'm getting `ValueError: you need to add an explicit nodesplitter to your input pipeline for multi-node training`, with `wds.split_by_node` only 1 GPU is utilized and the script hangs.
- Similarly `empty_check=False` avoids the need to set num_workers to less than or equal to the number of shards, otherwise: `ValueError: No samples found in dataset; perhaps you have fewer shards than workers.`
- We have to convert `BatchFeature` to dict: `dict(inputs)` for `split_between_processes` to work.

## Who can review?

@sayakpaul

### DIFF
--- a/examples/inference/distributed/florence2.py
+++ b/examples/inference/distributed/florence2.py
@@ -32,16 +32,12 @@ from accelerate import PartialState
 
 
 """
-Additional requirements: flash_attn einops timm
-pip install flash_attn einops timm
+Additional requirements: flash_attn einops timm webdataset fire tqdm huggingface_hub
+pip install flash_attn einops timm webdataset fire tqdm huggingface_hub
 
 Example:
 
 accelerate launch --num_processes=2 florence2.py --data_path "https://huggingface.co/datasets/pixparse/cc3m-wds/resolve/main/cc3m-train-0000.tar" --output_path outputs --batch_size 12 --num_workers 1 --prompt "<CAPTION>"
-
-On 2x4090: 420it [03:15,  2.15it/s] (~25.8 images/s)
-
-With --prompt "<DETAILED_CAPTION>": 420it [08:16,  1.18s/it] (~10.17 images/s)
 """
 
 
--- a/examples/inference/distributed/florence2.py
+++ b/examples/inference/distributed/florence2.py
@@ -1,0 +1,205 @@
+# Copyright 2024 The HuggingFace Inc. team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+import json
+import os
+import pathlib
+import queue
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from typing import Union
+
+import fire
+import torch
+import webdataset as wds
+from huggingface_hub.utils import insecure_hashlib
+from PIL import Image
+from tqdm import tqdm
+from transformers import AutoModelForCausalLM, AutoProcessor
+
+from accelerate import PartialState
+
+
+def main(
+    data_path: str,
+    output_path: str,
+    batch_size: int,
+    num_workers: int,
+    prompt: str = "<MORE_DETAILED_CAPTION>",
+    model_name: str = "microsoft/Florence-2-large",
+    max_new_tokens: int = 1024,
+    num_beams: int = 3,
+):
+    output_dir = pathlib.Path(output_path)
+
+    distributed_state = PartialState()
+
+    if distributed_state.is_main_process:
+        output_dir.mkdir(exist_ok=True)
+
+    model = AutoModelForCausalLM.from_pretrained(
+        model_name,
+        device_map=distributed_state.device,
+        torch_dtype=torch.float16,
+        trust_remote_code=True,
+    )
+
+    processor = AutoProcessor.from_pretrained(model_name, trust_remote_code=True, clean_up_tokenization_spaces=True)
+
+    class ExistsFilter:
+        def __init__(self, output_dir: Union[pathlib.Path, str]):
+            current_training_img_hashes = [f.split(".jpg")[0] for f in os.listdir(output_dir) if f.endswith(".jpg")]
+            self.current_training_img_hashes = set(current_training_img_hashes)
+            if distributed_state.is_main_process:
+                print(f"Existing images found: {len(self.current_training_img_hashes)}.")
+
+        def __call__(self, x):
+            if len(self.current_training_img_hashes) > 0:
+                if x["img_hash"] in self.current_training_img_hashes:
+                    return False
+                else:
+                    return True
+            else:
+                return True
+
+    def pil_to_bytes(pil_image: Image.Image, format="PNG"):
+        byte_arr = io.BytesIO()
+        pil_image.save(byte_arr, format=format)
+        im_bytes = byte_arr.getvalue()
+        return im_bytes
+
+    def preprocess_fn(sample):
+        image: Image.Image = sample["jpg"].convert("RGB")
+        img_hash = insecure_hashlib.sha1(image.tobytes()).hexdigest()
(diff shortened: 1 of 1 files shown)


sayakpaul (commented):
Thank you for working on this! 

Could we perhaps make the data_path configurable such that it accepts the supported paths by `webdataset`?

hlky:
I've added CLI args using `Fire`. `--data_path` can any of the supported paths by `webdataset` e.g. `https://huggingface.co/datasets/pixparse/cc3m-wds/resolve/main/cc3m-train-{0000..0011}.tar`, `/local/dir/shard-{0000..0011}.tar`, `pipe:aws s3 cp s3://data/shard-{0000..0011}.tar -`

examples/inference/distributed/florence2.py
```diff
@@ -0,0 +1,203 @@
...
+    dataset = (
+        wds.WebDataset(
+            data_path,
+            handler=wds.warn_and_continue,
+            nodesplitter=None,
+            shardshuffle=False,
+        )
+        .decode("pil", handler=wds.warn_and_continue)
+        .map(preprocess_fn, handler=wds.warn_and_continue)
+        .select(ExistsFilter(output_dir))
```
sayakpaul, line 126: Should this be applied only when `len(current_training_img_hashes) > 0`?

examples/inference/distributed/florence2.py
```diff
@@ -0,0 +1,203 @@
...
+                )
+    finally:
+        output_queue.put(None)
+        save_thread.shutdown(wait=True)
+
+    save_future.result()
+
+
+if __name__ == "__main__":
+    fire.Fire(main)
```
sayakpaul, line 203: Would be nice to add some notes above the script about the time it takes to get the artifacts done (maybe on a single shard)?

sayakpaul (commented):
Thanks! I left two minor comments.

examples/inference/distributed/florence2.py
hlky:
```python
def __call__(self, x):
    if len(self.current_training_img_hashes) > 0:
        if x["img_hash"] in self.current_training_img_hashes:
            return False
        else:
            return True
    else:
        return True
```
Can do, shouldn't make much difference though, and it's a little awkward due to the way [`select`](https://github.com/webdataset/webdataset/blob/fe2ac2b39635093ed9bed70ee31b435fcd2dbb71/webdataset/compat.py#L161-L173) [works](https://github.com/webdataset/webdataset/blob/fe2ac2b39635093ed9bed70ee31b435fcd2dbb71/webdataset/filters.py#L418-L431), we'd have to do something like
```python
exist_filter = ExistsFilter(output_dir)

dataset = (
    wds.WebDataset(
        data_path,
        handler=wds.warn_and_continue,
        nodesplitter=None,
        shardshuffle=False,
    )
    .decode("pil", handler=wds.warn_and_continue)
    .map(preprocess_fn, handler=wds.warn_and_continue)
)
if len(exist_filter.current_training_img_hashes) > 0:
    dataset = dataset.select(ExistsFilter(output_dir))
dataset = dataset.batched(
    batch_size,
    partial=False,
    collation_fn=partial(collate_fn, processor=processor),
)
```
WDYT?

examples/inference/distributed/florence2.py
hlky: Sure I'll do some benchmarking, any preference on GPU model and count?

sayakpaul: I think as long as the script runs on 2 GPUs (maybe 4090s), it should be okay. But I will let @muellerzr to comment further.

examples/inference/distributed/florence2.py
sayakpaul: Yes, that is exactly what we will have to do.

hlky: Done.

examples/inference/distributed/florence2.py
hlky: I've tested this on 2x4090. It appears to be CPU bound due to `sha1`, `pil_to_bytes` in preprocess, `Image.open`, `processor` on the batch in collate.

Testing with 1 shard, `https://huggingface.co/datasets/pixparse/cc3m-wds/resolve/main/cc3m-train-0000.tar`, batch size 12 (6 per gpu), and `<CAPTION>` task:
```
420it [10:33,  1.51s/it]
```

Compared to my project [florence-tool](https://github.com/bigdata-pw/florence-tool) using 1 GPU:
```
5046it [06:22, 13.18it/s]
```
In that project I save the results with `__key__` and `__url__`, use `processor` in `preprocess` then concat in `collate` and don't resave the image, although I do need to test whether `processor` is more efficient in `preprocess` or `collate`. Note: I increment the progress bar by batch size rather than batch index.

Testing a version of this script that does the same get roughly the same speed accounting for 1 vs 2 GPUs:
```
420it [03:16,  2.13it/s]
```

WDYT?

Also the last partial batch is missing in this script:
```
ls -f1 outputs/ | grep .json | wc -l
5040
```
Looks like we'd need to use `apply_padding=True` in `split_between_processes`.

sayakpaul: Thanks!

Not serializing the images is fine. I think it's dependent on how one is ultimately using the datasets in their workflows. But how are we doing the captions? They should be serialized, right? 

What I usually do is first serialize the images and caption, and then package them as `webdataset` shards. I think we could directly do that here instead of serializing the captions. WDYT? 

Additionally, for speeding up the PIL stuff, I find https://github.com/uploadcare/pillow-simd to be quintessential.

hlky: I've pushed an optimized version that still saves the images and original captions, seems the issue was `pil_to_bytes` and `Image.open`. See the [diff here](https://github.com/huggingface/accelerate/pull/3123/commits/9767e7000594714d2c89a8f6d6d8661df4935565).

Performance:
```
On 2x4090: 420it [03:15,  2.15it/s] (~25.8 images/s)

With --prompt "<DETAILED_CAPTION>": 420it [08:16,  1.18s/it] (~10.17 images/s)
```

sayakpaul: Nice. IIUC is this same performance you obtained with your tool? 

Additionally, would it be possible to check if pillow-simd leads to any speedup?

hlky: Yes it's the same performance.

No improvement using pillow-simd.

examples/inference/distributed/florence2.py
```diff
@@ -0,0 +1,206 @@
...
+from PIL import Image
+from tqdm import tqdm
+from transformers import AutoModelForCausalLM, AutoProcessor
+
+from accelerate import PartialState
+
+
+"""
+Additional requirements: flash_attn einops timm
+pip install flash_attn einops timm
```
muellerzr, line 36: We also need webdataset here

examples/inference/distributed/florence2.py
```diff
@@ -0,0 +1,206 @@
...
+
+Example:
+
+accelerate launch --num_processes=2 florence2.py --data_path "https://huggingface.co/datasets/pixparse/cc3m-wds/resolve/main/cc3m-train-0000.tar" --output_path outputs --batch_size 12 --num_workers 1 --prompt "<CAPTION>"
+
+On 2x4090: 420it [03:15,  2.15it/s] (~25.8 images/s)
```
muellerzr, line 42: IMO not sure worth having this, as it can vary largely by what CPU someone is using + internet speed. E.g. this baseline I got `420it [01:53, 3.70it/s]`

muellerzr (approved):
Thanks for this wonderful example! (and patience while I was OOO for a bit). Code wise this looks solid, just two suggestions in regards to the documentation at the top :)

examples/inference/distributed/florence2.py
hlky: I've added `webdataset` and `fire tqdm huggingface_hub` for completeness.

sayakpaul:
Thanks for your contributions here, @hlky! I didn't recognize you first but now after https://github.com/huggingface/diffusers/pull/9509, I can immediately recall!

sayakpaul:
@muellerzr is the failing test related?

muellerzr:
@sayakpaul nope it's not

PR #3144, opened 2024-10-08
add multi gpu image captioning

dame-cell:
# What does this PR do?
This PR adds a distributed image captioning example  using blip.

Based on https://gist.github.com/sayakpaul/dfb26d6562b62ba8c2f613dc912a28a8#file-generate_captions_multigpu-py 

This was discussed on the #3078

## Who can review?
@sayakpaul 

## Notes
- I was able to generate image captions using two T4 GPUs in less than 20 minutes for a dataset of 50,000 rows.
- If there are any concerns regarding the choice of model or dataset, I am open to making adjustments as needed.

<!-- Your PR will be replied to more quickly if you can figure out the right person to tag with @

 If you know how to use git blame, that is the easiest way, otherwise, here is a rough guide of **who to tag**.

- Big modeling: @SunMarc
- Fully-Sharded Data Parallism: @muellerzr
- DeepSpeed: @muellerzr
- Documentation: @muellerzr
- Maintained examples: @muellerzr or @SunMarc

### DIFF
--- a/examples/inference/distributed/distributed_image_captioning.py
+++ b/examples/inference/distributed/distributed_image_captioning.py
@@ -13,15 +13,16 @@
 # limitations under the License.
 
 """
-Originally by jiwooya1000, put together together by sayakpaul.
+Additional requirements: einops 
+pip install  einops 
+
+put together together by damerajee.
 Documentation: https://huggingface.co/docs/diffusers/main/en/training/distributed_inference
 
 Run:
 
 accelerate launch distributed_image_generation.py --batch_size 8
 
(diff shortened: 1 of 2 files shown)


sayakpaul:
Thanks for your contributions! But we already have #3123. Does it not have an overlap in the sense that it also adds captioning?

dame-cell:
@sayakpaul  i guess the only thing left if for the Speech data generation i will try contributing this  if no one has done it

sayakpaul:
Sure, speech-data generation would be very nice.

Issue #2985, opened 2024-08-01
conflict in deepsepped frame and accelerate.save_mdoel


yangtian6781:
### System Info

```Shell
- `Accelerate` version: 0.33.0
- Platform: Linux-5.4.0-182-generic-x86_64-with-glibc2.31
- `accelerate` bash location: /home/tianyang/miniconda3/envs/bunny/bin/accelerate
- Python version: 3.10.14
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.4.0+cu124 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- PyTorch MUSA available: False
- System RAM: 503.79 GB
- GPU type: NVIDIA GeForce RTX 3090
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: DEEPSPEED
        - mixed_precision: bf16
        - use_cpu: False
        - debug: False
        - num_processes: 2
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - enable_cpu_affinity: False
        - deepspeed_config: {'gradient_accumulation_steps': 1, 'offload_optimizer_device': 'none', 'offload_param_device': 'none', 'zero3_init_flag': True, 'zero3_save_16bit_model': True, 'zero_stage': 3}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

```python
import torch
from accelerate import Accelerator
from accelerate.utils import ProjectConfiguration
from torch.utils.data import Dataset, DataLoader
import torch.nn as nn
from torch.optim import AdamW



prefix_date = 'stage3_bf16'
projectconfig = ProjectConfiguration(project_dir=f'/home/tianyang/transformers-code/work_dir/{prefix_date}',
                                     logging_dir=f'/home/tianyang/transformers-code/work_dir/{prefix_date}')
class Simple(Dataset):
    def __init__(self, root='./data.bin') -> None:
        super().__init__()
        self.data = torch.load(root, weights_only=True)
[...]

PR #3146, opened 2024-10-08
Fixup Zero3 + `save_model`

muellerzr:
# What does this PR do?

When `save_model` is called under Zero3 only a single rank has all of the parameters. As a result `Accelerator.save_model` will throw an error under zero3 saying `'NoneType' object has no attribute 'items'` because in this case the `state_dict` of the model on non-0-ranks won't exist. 

Fixes https://github.com/huggingface/accelerate/issues/2985

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @BenjaminBossan

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2838,8 +2838,6 @@ class Accelerator:
             logger.error(f"Provided path ({save_directory}) should be a directory, not a file")
             return
 
-        os.makedirs(save_directory, exist_ok=True)
-
         # get the state_dict of the model
         if any(
             [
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -2854,6 +2854,10 @@ def save_model(
                 raise RuntimeError("You can't save the model since some parameters are on the meta device.")
             state_dict = self.get_state_dict(model)
 
+        # Case: DeepSpeed zero3 gets gathered and `state_dict` is empty
+        if state_dict is None:
+            return
```
SunMarc, line 2859: We can maybe make it more specific to deepspeed ? 
In their code, this is how they do this: https://github.com/microsoft/DeepSpeed/blob/f3943cf9109226ed3ecf2d5dbb639a11cd925555/deepspeed/runtime/engine.py#L3414
Seems like the model only gets saved on rank 0

SunMarc (approved):
LGTM ! Left a small suggestion but feel free to ignore.

src/accelerate/accelerator.py
muellerzr: I went more generic just in case somehow this happens in other places that we don't know yet (so we enable and don't block them)

PR #3150, opened 2024-10-09
POC: Allow for a `data_seed`

muellerzr:
# What does this PR do?

Adds in the ability to pass in a `data_seed` when using the `seedable_random_sampler`. Alternative to https://github.com/huggingface/accelerate/pull/3130

Essentially all we *really* want to do is make sure that the `initial_seed` in the `SeedableRandomSampler` can be changed to not be the same as the one used by torch. so as a result, adds it to the `DataLoaderConfiguration` class

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc @MekkCyber

### DIFF
--- a/src/accelerate/test_utils/scripts/test_script.py
+++ b/src/accelerate/test_utils/scripts/test_script.py
@@ -34,10 +34,12 @@ from accelerate.utils import (
     DistributedType,
     gather,
     is_bf16_available,
+    is_datasets_available,
     is_ipex_available,
     is_mlu_available,
     is_musa_available,
     is_npu_available,
+    is_pytest_available,
     is_xpu_available,
     set_seed,
     synchronize_rng_states,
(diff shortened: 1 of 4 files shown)


src/accelerate/test_utils/scripts/test_script.py
```diff
@@ -400,6 +400,33 @@ def check_seedable_sampler_in_batch_sampler_shard():
     ), "Sampler in BatchSamplerShard is not SeedableRandomSampler."
 
 
+def check_seedable_sampler_with_data_seed():
```
BenjaminBossan, line 403: I'm not sure if I quite get this test. I ran it with the current main branch, i.e. without the fix, and it passed. Of course, I removed the `data_seed=data_seed` argument, as it doesn't exist. But should the test not fail on main?

muellerzr: That's the weird part. Not sure what writing a test for this would look like tbh since it's just a passthrough in-a-sense. I guess check that the output when using it vs not differ?

muellerzr: Better test added 🫡

SunMarc (approved):
LGTM ! Thanks for adding this !

BenjaminBossan (approved):
Thanks, LGTM.

PR #3154, opened 2024-10-11
Models With Tied Weights Need Re-Tieing After FSDP Param Init

fabianlim:
# What does this PR do?

Currently in `FullyShardedDataParallelPlugin`, the `param_init_fn` [is set when sync_module_states=True](https://github.com/huggingface/accelerate/blob/6f79b63b865a33b92a3f1c9e2562b88ee7a4d89d/src/accelerate/utils/dataclasses.py#L1699). This is required by `FSDP` to initialize the shards (i.e. rank > 0) params in a [variety of situations](https://github.com/pytorch/pytorch/blob/8321eec009c8c79145ebccd51fdfc336e5f8b848/torch/distributed/fsdp/_init_utils.py#L588-L603), including the important one where the parameter was on `torch.device("meta")` because `low_cpu_mem_mode` was used. 

However `FullyShardedDataParallelPlugin.param_init_fn` is now set to 
```python
self.param_init_fn = lambda x: x.to_empty(device=device, recurse=False)
```
which causes problems when there are tied weights. Consider the following scenario
1. there are two modules `A` and `B` that share a tied weight `A.weight = B.weight`
2. `A.to_empty()` is called, then `A.weight` will be reassigned to a new tensor.
3. Similarly `B.empty()` is called, and then now `A.weight != B.weight`.

This is observed to cause problems in the `low_cpu_mem_mode=True` case, because now when getting the `managed_params` in FSDP, see [here](https://github.com/pytorch/pytorch/blob/8321eec009c8c79145ebccd51fdfc336e5f8b848/torch/distributed/fsdp/_init_utils.py#L626), 
- in rank = 0, `param_init_fn` is not called, because `low_cpu_mem_mode` will load weights in this shard. So `managed_params` will not have duplicates if weights are tied.
- In other ranks > 0, `param_init_fn` is called and `managed_params` may have duplicates. 

Then when FSDP calls `_sync_module_params_and_buffers`, the `torch.distributed._broadcast_coalesced` will be trying to communicate different number of tensors. This is not as intended and causes unexpected behaviors.

Fixes inconsistency in current logic in general, in particular is required to fix https://github.com/huggingface/trl/pull/2089 to completion.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -75,6 +75,7 @@ from .utils import (
     convert_model,
     convert_outputs_to_fp32,
     extract_model_from_parallel,
+    ensure_weights_retied,
     gather,
     gather_object,
     get_grad_scaler,
@@ -1473,12 +1474,11 @@ class Accelerator:
                     self.state.fsdp_plugin.set_auto_wrap_policy(model)
                     fsdp_plugin = self.state.fsdp_plugin
 
-                    # need to do this here to access the model
-                    # - there is a default param_init_fn inside but we ignore it
-                    # - qlora low_mem_mode requires trl > 0.11.2
-                    # - we should make this a passthrough if there are no meta devices
-                    from .utils.fsdp_utils import build_param_init_fn
-                    fsdp_plugin.param_init_fn = build_param_init_fn(model, self.device)
+                    # need to ensure that params are re-tied after running
+                    # param_init_fn
+                    fsdp_plugin.param_init_fn = ensure_weights_retied(
+                        fsdp_plugin.param_init_fn, model, self.device, 
+                    )
 
                     kwargs = {
                         "sharding_strategy": fsdp_plugin.sharding_strategy,
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1472,6 +1472,14 @@ class Accelerator:
                 if not is_type_fsdp:
                     self.state.fsdp_plugin.set_auto_wrap_policy(model)
                     fsdp_plugin = self.state.fsdp_plugin
+
+                    # need to do this here to access the model
+                    # - there is a default param_init_fn inside but we ignore it
+                    # - qlora low_mem_mode requires trl > 0.11.2
+                    # - we should make this a passthrough if there are no meta devices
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -1472,6 +1473,13 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                 if not is_type_fsdp:
                     self.state.fsdp_plugin.set_auto_wrap_policy(model)
                     fsdp_plugin = self.state.fsdp_plugin
+
+                    # need to ensure that params are re-tied after running
+                    # param_init_fn
+                    fsdp_plugin.param_init_fn = ensure_weights_retied(
+                        fsdp_plugin.param_init_fn, model, self.device, 
+                    )
```
muellerzr, line 1481: Would it make more sense to have this just exist as part of the `__post_init__` in `FullyShardedDataParallelPlugin` instead? Around when we declare what `param_init_fn` is?

muellerzr (commented):
Thanks for the fix, just one question on when this should actually be being set

src/accelerate/accelerator.py
fabianlim: right that was my initial thought to, but in that `__post_init__`. I cannot access the model. It is a similar pattern to the [call of `_fsdp_qlora_plugin_updates`](https://github.com/huggingface/transformers/blob/617b21273a349bd3a94e2b3bfb83f8089f45749b/src/transformers/trainer.py#L2226), which is done in the initial part of `Trainer._inner_training_loop`; this call internally accesses `self.model`, so its done outside of dataclass preperation

fabianlim:
@muellerzr noticed that one of the tests failed.. the test `test_save_load_model_use_pytorch` seems to have some kind of random component in it

enesmsahin:
This change creates an error if the model doesn't have `_tied_weights_keys`:
`AttributeError: 'FluxControlNetModel' object has no attribute '_tied_weights_keys'`

fabianlim:
@enesmsahin sorry for the overlook this is fixed here https://github.com/huggingface/accelerate/pull/3226

PR #3157, opened 2024-10-11
[docs] use nn.module instead of tensor as model 

faaany:
## What does this PR do?

When running the self-contained example on CUDA or XPU, I will get the following error: 
```bash
initial model weight is 0.00000
initial model weight is 0.00000
0 tensor([1., 2.], device='cuda:0')
  File "/mnt/disk4/fanlilin/workspace/test.py", line 35, in <module>
    outputs = inputs @ model
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument mat2 in method wrapper_CUDA_mm)
```
The reason is that `accelerator.prepare()` will not move the tensor model to device. We should have a model with type `torch.nn.Module`.

So this PR updates the example code on this. 

After the update, below is the result:
```bash
Detected kernel version 4.18.0, which is below the recommended minimum of 5.5.0; this can cause the process to hang. It is recommended to upgrade the kernel to the minimum version or higher.
initial model weight is 0.00000
initial model weight is 0.00000
0 tensor([1., 2.], device='cuda:0')
1 tensor([3., 4.], device='cuda:0')
2 tensor([5., 6.], device='cuda:0')
3 tensor([7., 8.], device='cuda:0')
w/ accumulation, the final model weight is 2.04000
```

### DIFF
--- a/docs/source/package_reference/utilities.md
+++ b/docs/source/package_reference/utilities.md
@@ -215,6 +215,8 @@ These include general utilities that should be used when working in parallel.
 
 [[autodoc]] utils.save
 
+[[autodoc]] utils.load
+
 [[autodoc]] utils.wait_for_everyone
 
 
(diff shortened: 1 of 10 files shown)


faaany:
I think the failed CI is not caused by my change. Could you pls help retrigger the CI? Thanks a lot! @muellerzr

PR #3159, opened 2024-10-12
enable cpu bnb distributed lora finetune

jiqing-feng:
Hi @SunMarc , this PR enables cpu bnb model distributed finetune by removing check index because we don't need index in cpu device. Please review it. Thanks!

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1421,7 +1421,10 @@ class Accelerator:
                     current_device.index if isinstance(current_device, torch.device) else current_device
                 )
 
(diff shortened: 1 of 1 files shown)


jiqing-feng:
Hi @SunMarc @ArthurZucker  , if we don't make this change, the error be like:
image

SunMarc (commented):
Thanks for the PR @jiqing-feng ! Instead of removing this check, could you add the multi-backend bnb check ?

jiqing-feng:
Hi @SunMarc . I have checked bnb multi-backend, please review it. Thx!

SunMarc (approved):
LGTM! Thanks!

PR #3161, opened 2024-10-12
🚀 [Feature] Add deprecated Decorator

yhna940:
# What does this PR do?

## Summary
This PR introduces a `deprecated` decorator, designed to mark functions as deprecated with detailed warnings, following the approach in [torch/onnx/_deprecation.py](https://github.com/pytorch/pytorch/blob/6001b16597d94ab46764941ff69fb2305ce5f41c/torch/onnx/_deprecation.py). The decorator provides a standard way to issue deprecation warnings and update docstrings, informing users of the version when a function was deprecated, its removal target, and recommended alternative actions.

## Changes
- **New `deprecated` decorator:** Adds a decorator that marks functions as deprecated, integrates a warning when the function is called, and updates docstrings accordingly.
- **Integration with `FindTiedParametersResult`:** Applied the `deprecated` decorator to the `values` method in the `FindTiedParametersResult` class, signaling to users that this method will be removed in Accelerate v1.3.0 and suggesting alternative approaches.

## Example Usage
After this PR, the following usage:

```python
from accelerate.utils.modeling import FindTiedParametersResult
results = FindTiedParametersResult([[1, 2, 3]])
results.values()
```

will produce output similar to:
```
<stdin>:1: FutureWarning: 'accelerate.utils.modeling.values' is deprecated in version 1.0.0rc0 and will be removed in 1.3.0. Please use another method instead.
[2, 3]
```

## Testing
Included a test for `deprecated` to validate:
- Warning output when a deprecated function is called.
- Docstring formatting to ensure it includes the deprecation note with the appropriate information.


BenjaminBossan:
Maybe not relevant, but Python added its own [`warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) decorator in Python 3.13. Using the same name could be confusing. The Python decorator apparently also allows type checkers to flag the usage, which would be nice to have. Of course, we can't use it directly since we want to support older Python versions.

yhna940:
> Maybe not relevant, but Python added its own [`warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) decorator in Python 3.13. Using the same name could be confusing. The Python decorator apparently also allows type checkers to flag the usage, which would be nice to have. Of course, we can't use it directly since we want to support older Python versions.

Hello @BenjaminBossan, thank you for the insightful review! I wasn’t aware of the `warnings.deprecated` decorator introduced in Python 3.13, and I appreciate you pointing it out. As I am currently working with a lower Python version (likely 3.8, according to the Dev Container), I implemented a custom decorator to ensure compatibility.

That said, the name overlap could indeed be confusing. Would using an alternative name for the decorator help clarify things? I’d be glad to make adjustments if needed!

BenjaminBossan:
As to the Python version, 3.8 is going to be end of life by the end of the month, so the containers will be updated soon. Still, it will be quite some time before 3.13 will be the lowest supported Python version, so we can't assume that the decorator exists. More importantly, I wonder if we can "borrow" some of the [implementation details](https://peps.python.org/pep-0702/#runtime-behavior) of `warnings.deprecated`, as I'm sure the Python devs put a lot of thought into it. Regarding the naming, I'll let Zach decide on that.

yhna940:
Thank you for your feedback, I agree with your suggestions regarding naming and compatibility! Using Python’s `warnings.deprecated` decorator in 3.13 and above, while defaulting to our custom decorator in lower versions, sounds like a balanced approach. This way, we maintain compatibility across versions while taking advantage of the built-in decorator when possible.

I’ve resolved the conflicts in the code for now, and I’ll be sure to apply any further adjustments based on additional feedback. Thanks again!

src/accelerate/utils/deprecation.py
```diff
@@ -0,0 +1,80 @@
...
+
+    It will result in a warning when the function is called and a note in the docstring.
+
+    Args:
+        since (`str`):
+            The version when the function was first deprecated.
+        removed_in (`str`):
+            The version when the function will be removed.
+        instructions (`str`):
+            The action users should take.
```
BenjaminBossan, line 38: Just a nit: I'd find it more intuitive if the `instructions` was renamed `instruction` and if the `"Please "` was not added in front, as I would assume as a caller that I need to pass a full sentence here.

tests/test_utils.py
```diff
@@ -417,6 +419,61 @@ def test_convert_dict_to_env_variables(self):
             valid_env_items = convert_dict_to_env_variables(env)
         assert valid_env_items == ["ACCELERATE_DEBUG_MODE=1\n", "OTHER_ENV=2\n"]
 
+    def test_deprecated(self):
```
BenjaminBossan, line 422: I'm wondering if the decorator would also work for a class? If yes, let's add a test and update the docstring. If not, we can leave it as is.

BenjaminBossan (approved):
Thanks for the updates. I only have a nit left, no blocker.

> Using Python’s warnings.deprecated decorator in 3.13 and above, while defaulting to our custom decorator in lower versions

I don't think we really need to switch over come Python 3.13, since the Python decorator works quite differently from this one.

src/accelerate/utils/deprecation.py
yhna940: Done in [a7a43aa](https://github.com/huggingface/accelerate/pull/3161/commits/a7a43aa2f2268885b5db25a1324b04b59921ec65) and [2cb4648](https://github.com/huggingface/accelerate/pull/3161/commits/2cb46485f9bfda5c6108978fe786eeda3227ef0f) Thank you :)

tests/test_utils.py
yhna940: Done in [89b1974](https://github.com/huggingface/accelerate/pull/3161/commits/89b1974c0539bbcf21d70abf91505021d8991de4). Thank  you!

matthewdouglas:
Hi folks,

Could we consider using the implementation from `typing_extensions>=4.5` instead? It could maybe be wrapped still to provide help with formatting for UX/docs, but that way we get improved IDE support and alignment with PEP 702.

BenjaminBossan:
@matthewdouglas AFAICT, [that implementation](https://github.com/python/typing_extensions/blob/67c16e190e7ebe6a9130b63839520ce01bdd6422/src/typing_extensions.py#L2867-L2872) is a backport of `warnings.deprecated` that we discussed above. I agree that typing support would be nice to have, but feature-wise, the two are not on par, so not sure if we can simply replace it.

yhna940:
I hope you're all doing well! I wanted to kindly follow up on this PR to check if there are any remaining concerns or additional feedback that I should address to move it forward. Please let me know if there's anything else I can do to improve this PR :) Thank you for your time and guidance!

BenjaminBossan:
_not stale_

src/accelerate/utils/deprecation.py
```diff
@@ -0,0 +1,80 @@
...
+from typing import Callable, TypeVar
+
+from typing_extensions import ParamSpec
+
+
+_T = TypeVar("_T")
+_P = ParamSpec("_P")
+
+
+def deprecated(since: str, removed_in: str, instruction: str) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]:
```
XuehaiPan, line 27: You can implement the decorator based on [`warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) (since Python 3.13) or `typing_extensions.deprecated`. It has better linter and IDE support.

PR #3163, opened 2024-10-13
add the missing xpu for local sgd

faaany:
## What does this PR do?
As the tile suggests, local sgd also works on XPU.


muellerzr (approved):
Thanks!

faaany:
I think the failed test is not relevant to my change in LocalSGD. Could you help retrigger the CI? Thanks a lot! @muellerzr

Issue #3147, opened 2024-10-09
Duplicated process and memory leakage for evaluation process in all_gather


SangbumChoi:
### System Info

```Shell
For all accelerate version
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

Related issue 
https://github.com/huggingface/transformers/issues/15466
https://github.com/huggingface/transformers/pull/28769/files

### Expected behavior

https://github.com/huggingface/accelerate/blob/55136b8dc4a1f5bf8a33f38f25b279debdabcc00/src/accelerate/utils/operations.py#L353

All the accelerate gather function is stricted to `all_gather`. However, there are also the way of using `gather` in main process to calculate the evaluation process. If we use `all_gather` for the evaluation process and pass it to cpu it will cost n times (n is number of process). However we only require to gather the distributed variable to one place to calculate.

What do you think about this?

https://github.com/facebookresearch/detectron2/blob/ebe8b45437f86395352ab13402ba45b75b4d1ddb/detectron2/utils/comm.py#L188

SangbumChoi:
https://github.com/huggingface/accelerate/issues/2898

muellerzr:
@SangbumChoi definitely open to trying out something more efficient! Best case scenario we have a flag to use `all_gather` instead, and default to this new method. Would you like to take a stab at a PR?

PR #3164, opened 2024-10-14
add use_all_gather for option

SangbumChoi:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/3147
@muellerzr

I think testing might need but not sure where to add this corresponding test

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2454,7 +2454,7 @@ class Accelerator:
                 and inefficient as it incurs GPU -> CPU transfer since tensors would be pickled.
             use_all_gather(`bool`):
                 Whether to use all_gather instead of gather. all_gather collects a list of tensors from all processes,
-                while gather collects tensors into a single process. Using all_gather can be beneficial in scenarios 
+                while gather collects tensors into a single process. Using all_gather can be beneficial in scenarios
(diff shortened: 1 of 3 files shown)


SangbumChoi:
Also have additional question for LLM inference evaluation process. Does LLM evaluation process requires to have `all_gather` function to make all the prediction in all processes?

muellerzr:
Thanks! Tests should be in `src/accelerate/test_utils/scripts/test_ops.py`

The `all_gather` is so that we can get the results across all processes. It's there as an example and for simplicity for the user. (But it is not generally required, no)

SangbumChoi:
@muellerzr Hi I haved added test. However, to be more strict test I think we can add additional checking when the process is not in the `state.is_main_process` and check the return is empty list. Let me know how do you think about this.

SangbumChoi:
@muellerzr Gentle Ping!

muellerzr (changes requested):
Since using `all_gather` uses more memory by default, let's set a default to `False` please (even though we said the opposite initially I think)

muellerzr:
re; test:I think it's fine how it is right now

SangbumChoi:
@muellerzr Does this resolved in another way?

PR #3165, opened 2024-10-14
fix version check bug in `get_xpu_available_memory`

faaany:
## What does this PR do?
This PR fixed the version check bug introduced in #3076. Below are the verified results with ipex "2.5.10+gitad5c117":

```bash
from accelerate.utils.memory import get_xpu_available_memory

mem = get_xpu_available_memory(1)
print(mem)
# 68719476736
```

@muellerzr


muellerzr (approved):
Thanks!

PR #3181, opened 2024-10-21
[docs] update neptune API

faaany:
## What does this PR do?
The neptune init API in the documentation is out-dated. This PR updates the API to the latest API when `pip install neptune` is used. 

@muellerzr


BenjaminBossan (approved):
Thanks for this update to the docs. I quickly checked the Neptune docs and this looks consistent.

PR #3186, opened 2024-10-22
fix bnb

eljandoubi:
# What does this PR do?
Fixed the type of `bnb_4bit_compute_dtype`, which was incorrectly set as bool. Changed it to str to ensure that `transformers.HfArgumentParser` can parse it from the CLI, as Union is not valid in this context.
## Who can review?

### DIFF
--- a/src/accelerate/utils/bnb.py
+++ b/src/accelerate/utils/bnb.py
@@ -334,26 +334,25 @@ def _replace_with_bnb_layers(
                     break
             if proceed:
                 # Load bnb module with empty weight and replace ``nn.Linear` module
-                with torch.device(module.weight.device): # in case of meta device
-                    if bnb_quantization_config.load_in_8bit:
-                        bnb_module = bnb.nn.Linear8bitLt(
-                            module.in_features,
-                            module.out_features,
-                            module.bias is not None,
-                            has_fp16_weights=False,
-                            threshold=bnb_quantization_config.llm_int8_threshold,
-                        )
-                    elif bnb_quantization_config.load_in_4bit:
-                        bnb_module = bnb.nn.Linear4bit(
-                            module.in_features,
-                            module.out_features,
-                            module.bias is not None,
-                            bnb_quantization_config.bnb_4bit_compute_dtype,
-                            compress_statistics=bnb_quantization_config.bnb_4bit_use_double_quant,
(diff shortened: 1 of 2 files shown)


eljandoubi:
Fixes #3189

src/accelerate/utils/bnb.py
```diff
@@ -334,25 +334,26 @@ def _replace_with_bnb_layers(
...
-                        module.in_features,
-                        module.out_features,
-                        module.bias is not None,
-                        bnb_quantization_config.bnb_4bit_compute_dtype,
-                        compress_statistics=bnb_quantization_config.bnb_4bit_use_double_quant,
-                        quant_type=bnb_quantization_config.bnb_4bit_quant_type,
-                    )
-                else:
-                    raise ValueError("load_in_8bit and load_in_4bit can't be both False")
+                with torch.device(module.weight.device): # in case of meta device
```
SunMarc, line 337: Why did you add this ? This function already under `init_empty_weights` in the `load_and_quantize_model` function

SunMarc (commented):
Thanks for fixing ! Left a comment

src/accelerate/utils/bnb.py
eljandoubi: In docs https://huggingface.co/docs/accelerate/usage_guides/quantization, it was not clear. Plus `load_and_quantize_model` gives a simple warning in case model_device.type != "meta"`` . So I thought `load_and_quantize_model` works for all types of devices.

SunMarc: Oh thanks, if you initialized the model with `init_empty_weights`, `model_device.type` should be "meta" no ? Could you check why it is not returning meta ? Maybe we need to change the `model_device.type` check instead

eljandoubi: I run my script from accelerate cli to perform deepspeed stage 2 on cluster of machines. Do not know if this has an effect on code.

SunMarc: Not sure also. Feel free to open a PR if you manage to fix the issue ! We can discuss about that in the issue you create. If you can share a minimal reproducer , that would be helpful. The best would be to reproduce this without deepspeed.

SunMarc (approved):
Thanks !

BenjaminBossan (approved):
LGTM, thanks.

PR #3187, opened 2024-10-23
MLU devices : Checks if mlu is available via an cndev-based check which won't trigger the drivers and leave mlu

huismiling:
# What does this PR do?

MLU devices : Checks if mlu is available via an cndev-based check which won't trigger the drivers and leave mlu

Check devices(cuda and mlu) is available with `patch_environment` . 

Move `clear_environment` and `patch_environment` into `src/accelerate/utils/environment.py` to avoid circular import .

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


huismiling:
@BenjaminBossan @muellerzr 
Previous PR https://github.com/huggingface/accelerate/pull/3057 is closed.
Sorry for taking so long time to fix circular import. 
`clear_environment `and `patch_environment` are moved into [src/accelerate/utils/environment.py](https://github.com/huismiling/accelerate/blob/main/src/accelerate/utils/environment.py) .

muellerzr (approved):
Thanks! Makes sense that we do the realignment here of things & how they ended up.

BenjaminBossan (approved):
Thanks for coming back to this PR, LGTM.

Short question for @muellerzr: We should be fine with the functions no longer being importable from `accelerate.utils.other` because users should always import from the higher level, right?

muellerzr:
Yep exactly. We can add a deprecation warning here but we've never advertised it in that way at that import level so I think we should be fine.

BenjaminBossan:
Okay, great, just wanted to be sure. This PR is good to be merged from my point of view.

PR #3193, opened 2024-10-24
Give example on how to handle gradient accumulation with cross-entropy

ylacombe:
# What does this PR do?

Following the recent [highlights](https://huggingface.co/blog/gradient_accumulation) on how gradient accumulation with the cross-entropy loss is usually off, it could be great to have it mentioned in the doc. I've thus added some code and explanation of it in the gradient accumulation page.

cc @SunMarc and @muellerzr, let me know what you think of it or if I can make this any clearer!

### DIFF
--- a/docs/source/usage_guides/gradient_accumulation.md
+++ b/docs/source/usage_guides/gradient_accumulation.md
@@ -252,6 +252,8 @@ In other words, some adjustements must be made on losses that operate on a token
 ```python
 from accelerate import Accelerator
 import math
+import contextlib
+
 gradient_accumulation_steps = 2
 accelerator = Accelerator(gradient_accumulation_steps=gradient_accumulation_steps)
 model, optimizer, training_dataloader, scheduler = accelerator.prepare(
@@ -279,16 +281,17 @@ for update_step in range(total_updates):
         # to compute it correctly in a multi-device DDP training, we need to gather the total number of items in the full batch.
         num_items_in_batch = accelerator.gather(num_items_in_batch).sum().item()
             
-        for batch in batch_samples:
+        for i, batch in enumerate(batch_samples):
+            # if we perform gradient accumulation in a multi-devices set-up, we want to avoid unecessary communications when accumulating
+            # cf: https://muellerzr.github.io/blog/gradient_accumulation.html
+            if (i < len(batch_samples) - 1 and accelerator.num_processes > 1):
+                ctx = model.no_sync
+            else:
+                ctx = contextlib.nullcontext
+            
             total_batched_samples += 1
 
-            with accelerator.accumulate(model):
-                # Since we performed prefetching, we need to manually set sync_gradients
-                if total_batched_samples % gradient_accumulation_steps != 0:
-                    accelerator.gradient_state._set_sync_gradients(False)
-                else:
-                    accelerator.gradient_state._set_sync_gradients(True)
-
+            with ctx():
                 inputs, targets = batch
                 outputs = model(inputs)
                 loss = loss_function(outputs, targets) # the loss function shoud sum over samples rather than averaging
@@ -298,9 +301,11 @@ for update_step in range(total_updates):
                 loss = (loss * gradient_accumulation_steps * accelerator.num_processes) / num_items_in_batch
                 
                 accelerator.backward(loss)
-                optimizer.step()
-                scheduler.step()
-                optimizer.zero_grad()
+
+        # Sync gradients and perform optimization steps once every gradient_accumulation_steps
+        optimizer.step()
+        scheduler.step()
+        optimizer.zero_grad()
 ```
 
 ### Self-contained causal LM example
@@ -313,6 +318,7 @@ from accelerate.utils import set_seed
 from accelerate.logging import  get_logger
 from torch.utils.data import Dataset, DataLoader
 import math
+import contexlib
 
 # seed
 set_seed(0)
@@ -385,16 +391,16 @@ for update_step in range(total_gradient_updates):
         num_items_in_batch = accelerator.gather(local_num_items_in_batch).sum().item()
         logger.warning(f"Total num items {num_items_in_batch}")
 
-        for batch in batch_samples:
+        for i, batch in enumerate(batch_samples):
             inputs, labels = batch["input_ids"], batch["labels"]
             total_batched_samples += 1
-
-            with accelerator.accumulate(model):
-                # Since we performed prefetching, we need to manually set sync_gradients
-                if total_batched_samples % gradient_accumulation_steps != 0:
-                    accelerator.gradient_state._set_sync_gradients(False)
-                else:
-                    accelerator.gradient_state._set_sync_gradients(True)
+            # if we perform gradient accumulation in a multi-devices set-up, we want to avoid unecessary communications when accumulating
+            # cf: https://muellerzr.github.io/blog/gradient_accumulation.html
+            if (i < len(batch_samples) - 1 and accelerator.num_processes > 1):
+                ctx = model.no_sync
+            else:
+                ctx = contextlib.nullcontext
+            with ctx():
 
                 outputs = model(inputs)
                 loss = criterion(outputs.view(-1, 2), labels.view(-1).to(torch.int64))
@@ -403,8 +409,8 @@ for update_step in range(total_gradient_updates):
                 # Same reason for gradient_accumulation_steps, but this times it's Accelerate that calculate the average gradient across the accumulated steps 
                 loss = (loss * gradient_accumulation_steps * accelerator.num_processes) / num_items_in_batch
                 accelerator.backward(loss)
-                model_optimizer.step()
-                model_optimizer.zero_grad()
+        model_optimizer.step()
+        model_optimizer.zero_grad()
                 
 
 logger.warning(f"Device {accelerator.process_index} - w/ accumulation, the final model weight is {accelerator.unwrap_model(model).weight.detach().cpu().squeeze()}", main_process_only=False)
@@ -429,7 +435,7 @@ if accelerator.is_main_process:
 
 ```
 
-Results on a single device:
+Results on a single device - gradient accumulation steps set to 1 and batch_size set to 8:
 ```
 initial model weight is tensor([-0.0075,  0.5364])
 initial model clone weight is tensor([-0.0075,  0.5364])
@@ -439,7 +445,7 @@ Device 0 - w/ accumulation, the final model weight is tensor([0.0953, 0.4337])
 w/o accumulation, the final model weight is tensor([0.0953, 0.4337])
 ```
 
-Results on a two devices set-up:
+Results on a two devices set-up - gradient accumulation steps set to 2 and batch_size set to 4.
 ```
 initial model weight is tensor([-0.0075,  0.5364])
 initial model clone weight is tensor([-0.0075,  0.5364])
@@ -450,3 +456,15 @@ Device 1 - w/ accumulation, the final model weight is tensor([0.2117, 0.3172])
 Device 0 - w/ accumulation, the final model weight is tensor([0.2117, 0.3172])
 w/o accumulation, the final model weight is tensor([0.2117, 0.3172])
 ```
+
+### To go further:
+
+Please find a complete example script on a real world training run in the examples folder at the path [`accelerate/examples/by_feature/gradient_accumulation_for_autoregressive_models.py`](https://github.com/huggingface/accelerate/blob/main/examples/by_feature/gradient_accumulation_for_autoregressive_models.py).
+
+Running it on several training configurations with constant global batch size equal to 32 gives the following graph:
+
+<div style="text-align: center">
+<img src="https://huggingface.co/datasets/hf-audio/gradient_accumulation_example/resolve/main/training_losses.png">
+</div>
+
+Note that the training losses are exactly the same up to training step 20. The small deviation after this training step occurs at the very end of the first epoch, because, by [default](https://huggingface.co/docs/accelerate/en/package_reference/torch_wrappers#accelerate.data_loader.prepare_data_loader.even_batches), the dataloader duplicates the samples at the beginning of the dataset when the total batch size doesn't exactly divide the dataset.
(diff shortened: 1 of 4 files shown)


docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+```
+initial model weight is tensor([-0.0075,  0.5364])
+initial model clone weight is tensor([-0.0075,  0.5364])
+Step 0 - Device 0 - num items in the local batch 52
+Step 0 - Device 1 - num items in the local batch 84
+Total num items 136
+Device 1 - w/ accumulation, the final model weight is tensor([0.2117, 0.3172])
+Device 0 - w/ accumulation, the final model weight is tensor([0.2117, 0.3172])
+w/o accumulation, the final model weight is tensor([0.2117, 0.3172])
+```
```
muellerzr, line 452: Honestly if we can let's even toss up some `wandb` graphs 🔥

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+                
+                # We multiply by num_processes because the DDP calculates the average gradient across all devices whereas dividing by num_items_in_batch already takes into account all devices
+                # Same reason for gradient_accumulation_steps, but this times it's Accelerate that calculate the average gradient across the accumulated steps 
+                loss = (loss * gradient_accumulation_steps * accelerator.num_processes) / num_items_in_batch
+                accelerator.backward(loss)
+                model_optimizer.step()
+                model_optimizer.zero_grad()
+                
+
+logger.warning(f"Device {accelerator.process_index} - w/ accumulation, the final model weight is {accelerator.unwrap_model(model).weight.detach().cpu().squeeze()}", main_process_only=False)
```
muellerzr, line 410: Rather than `logger.warning`, we can do `print()` here or change the default logging level :) (Just `logging.warning` rather than `logging.info` weirds me out)

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+        for _ in range(num_batches_in_step):
+            batch_samples += [next(training_iterator)]
+            
+        # get local num items in batch 
+        local_num_items_in_batch = sum([(batch["labels"].ne(-100)).sum() for batch in batch_samples])
+        logger.warning(f"Step {update_step} - Device {accelerator.process_index} - num items in the local batch {local_num_items_in_batch}", main_process_only=False)
+
+        # to compute it correctly in a multi-device DDP training, we need to gather the total number of items in the full batch.
+        num_items_in_batch = accelerator.gather_for_metrics(local_num_items_in_batch).sum().item()
+        logger.warning(f"Total num items {num_items_in_batch}", main_process_only=True)
```
muellerzr, line 386: also if we ever do `true`, no need to include it since it's the default :)

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+
+# define model, model_optimizer and loss function
+model = torch.nn.Linear(1, 2, bias=False)
+model_clone = copy.deepcopy(model)
+criterion = torch.nn.CrossEntropyLoss(reduction="sum") # must sum over samples rather than averaging
+model_optimizer = torch.optim.SGD(model.parameters(), lr=0.08)
+
+
+logger.warning(f"initial model weight is {model.weight.detach().cpu().squeeze()}", main_process_only=True)
+logger.warning(f"initial model clone weight is {model_clone.weight.detach().cpu().squeeze()}", main_process_only=True)
```
muellerzr, line 359: Same thing, no need for `main_process_only`

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+        return self.len
+    
+def collate_fn(features):
+    input_ids = torch.nn.utils.rnn.pad_sequence([f["input_ids"] for f in features], batch_first=True, padding_value=-100)
+    labels = torch.nn.utils.rnn.pad_sequence([f["labels"] for f in features], batch_first=True, padding_value=-100)
+    return {"input_ids": input_ids[..., None], "labels": labels[..., None]}
+
+# define toy inputs and labels
+gradient_accumulation_steps = 2
+batch_size = 4
```
muellerzr, line 341: Let's call it a bit more explicitly what it is:

```python
per_device_batch_size
```

muellerzr (approved):
Looks great! (We need to do the `gather()` + div by num processes in the trainer still).

Left a few nits, I think it'd be really cool if we can show full training graphs. After doing stuff with FP8 just taking "the end result is the same" at face value I don't fully trust :)

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+        # we need to pre-load the full local batch - i.e the next per_device_batch_size * accumulation_steps samples
+        batch_samples = []
+        num_batches_in_step = gradient_accumulation_steps if update_step != (total_updates - 1) else remainder
+        for _ in range(num_batches_in_step):
+            batch_samples += [next(training_iterator)]
+            
+        # get local num items in batch 
+        num_items_in_batch = sum([(batch["labels"].ne(-100)).sum() for batch in batch_samples])
+        # to compute it correctly in a multi-device DDP training, we need to gather the total number of items in the full batch.
+        num_items_in_batch = accelerator.gather_for_metrics(num_items_in_batch).sum().item()
```
muellerzr, line 280: Also across all we should use `gather()` rather than `gather_for_metrics`, we don't want to drop anything in the case of loss calculation!

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
 w/ accumulation, the final model weight is 2.04000
 w/o accumulation, the final model weight is 2.04000
 ```
+
+## Gradient accumulation on training samples of variable size
+
+As was pointed out in this [blog-post](https://huggingface.co/blog/gradient_accumulation), which points out a common error that occurs when perfoming gradient accumulation on training samples of variable size:
```
SunMarc, line 244:
```suggestion
As was pointed out in this [blog-post](https://huggingface.co/blog/gradient_accumulation), which points out a common error that occurs when performing gradient accumulation on training samples of variable size:
```

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+total_gradient_updates = math.ceil(num_samples_in_epoch / gradient_accumulation_steps)
+
+total_batched_samples = 0
+for update_step in range(total_gradient_updates):
+        # In order to correctly the total number of non-padded tokens on which we'll compute the cross-entropy loss
+        # we need to pre-load the full local batch - i.e the next per_device_batch_size * accumulation_steps samples
+        batch_samples = []
+        num_batches_in_step = gradient_accumulation_steps if update_step != (total_gradient_updates - 1) else remainder
+        for _ in range(num_batches_in_step):
+            batch_samples += [next(training_iterator)]
```
SunMarc, line 378: This only works when we know the size of the dataloader. Can we think of a solution that doesn't require this information ? I think we can just iter on the dataloader until we have `gradient_accumulation_steps` to create the batch_sample. If we can't iter anymore, then we stop also. I think that code will be easier to understand.

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+            inputs, labels = batch["input_ids"], batch["labels"]
+            total_batched_samples += 1
+
+            with accelerator.accumulate(model):
+                # Since we performed prefetching, we need to manually set sync_gradients
+                if total_batched_samples % gradient_accumulation_steps != 0:
+                    accelerator.gradient_state._set_sync_gradients(False)
+                else:
+                    accelerator.gradient_state._set_sync_gradients(True)
+
```
SunMarc, line 398: The issue here is due to `end_of_dataloader` that will modify incorrectly the `sync_gradient` due to the prefetching. 
Maybe we can add an option to disable `do_sync` in accumulate ? This way, we won't have to put this specific piece of code under accumulate and the user will have total control of when we do sync the gradient. cc @muellerzr

docs/source/usage_guides/gradient_accumulation.md
ylacombe: Indeed, it'd be great, but here we do only one single global batch size, I don't think it's worth adding a graph. Maybe should I do the same but with multiple global batch size ?

ylacombe: Or add some wandb graphs from the upcoming modif of `examples/by_feature/gradient_accumulation` ?

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+    model_clone_optimizer.zero_grad()
+    loss.backward()
+    model_clone_optimizer.step()
+    
+    # We should have the same final weights.
+    logger.warning(f"w/o accumulation, the final model weight is {model_clone.weight.detach().cpu().squeeze()}")
+
+```
+
+Results on a single device:
```
SunMarc, line 432: Maybe we can precise the exact setup ? I think that we are doing the following ? 
- dp=1 grad_acc= 2 batch_size = 4 vs dp=1 grad_acc= 1 batch_size = 8 ? 
If we are only doing one update, then we won't be able to get a graph. Maybe we do this on a larger dataset where batch_size != len(data_loader) and add the graphs.

docs/source/usage_guides/gradient_accumulation.md
```diff
@@ -238,3 +238,215 @@ initial model weight is 0.00000
...
+initial model weight is tensor([-0.0075,  0.5364])
+initial model clone weight is tensor([-0.0075,  0.5364])
+Step 0 - Device 0 - num items in the local batch 36
+Total num items 36
+Device 0 - w/ accumulation, the final model weight is tensor([0.0953, 0.4337])
+w/o accumulation, the final model weight is tensor([0.0953, 0.4337])
+```
+
+Results on a two devices set-up:
+```
```
SunMarc, line 443: On a two devices set-up, the modification you did to take into account the dp won't be reflected here as we are only changing grad acc and batch_size. So the loss will be the same nevertheless. However, it's nice to see that the total_num_items really changed: 
- dp=2 grad_acc= 2 batch_size = 4 vs dp=2 grad_acc=1 batch_size=8 

Maybe we should probably do a separate section/experiment to show the following will have the same loss graph
- dp=2 batch_size =2 is the same as dp=1 batch_size=4

SunMarc (commented):
Nice job @ylacombe ! Left a few suggestions !

docs/source/usage_guides/gradient_accumulation.md
muellerzr: agreed, you can put it as part of this PR if you want @ylacombe

ylacombe: I haven't take into account this case, but we should also set `sync_gradient=True` when reaching the very last `total_batched_samples` btw

ylacombe: Not sure to understand exactly what your point @SunMarc,  reaching `end_of_dataloader` also set `accelerator.step` to 0. If I disable it, we'd have [issues when saving the accelerator state](https://github.com/ylacombe/accelerate/blob/395c572dd30ba7c8d18c6a7cff506a567897d69e/src/accelerate/accelerator.py#L3074), right ?

SunMarc: I think it should be fine as we don't care about step in Trainer also. cc @muellerzr but we can leave that for a follow up PR if you want !

muellerzr: Follow up would be fine by me :)

tests/test_examples.py
```diff
@@ -245,6 +246,10 @@ def test_gradient_accumulation(self):
         testargs = ["examples/by_feature/gradient_accumulation.py"]
         run_command(self.launch_args + testargs)
 
+    def test_gradient_accumulation_for_autoregressive_models(self):
+        testargs = ["examples/by_feature/gradient_accumulation_for_autoregressive_models.py"]
+        run_command(self.launch_args + testargs)
```
muellerzr, line 251: Just a nit: this doesn't use gradient accumulation here since it uses the default of `1`

examples/by_feature/gradient_accumulation_for_autoregressive_models.py
```diff
@@ -0,0 +1,341 @@
...
+        "--gradient_accumulation_steps",
+        type=int,
+        default=1,
+        help="The number of minibatches to be ran before gradients are accumulated.",
+    )
+    parser.add_argument(
+        "--per_device_batch_size",
+        type=int,
+        default=2,
+        help="The number of minibatches to be ran before gradients are accumulated.",
```
muellerzr, line 326: Shouldn't this be "The size of each minibatch"?
```suggestion
        help="The size of each minibatch",
```

JaheimLee:
any updates?

muellerzr (approved):
Beautiful! LG2M, @SunMarc are you happy with it as well? :)

SunMarc (approved):
Really nice PR ! Thanks for iterating !

JaheimLee:
Hi, I have a question. `transformers` has already shifted source and target. Why do you need to shift them here?

SunMarc:
Could you detail a bit your point ?

JaheimLee:
> Could you detail a bit your point ?

I mean you shift `input_ids` and `labels` [here](https://github.com/ylacombe/accelerate/blob/cc8bcc8820cd45f19dca5f0517012131cbd1be8d/examples/by_feature/gradient_accumulation_for_autoregressive_models.py#L117), but `transformers` has shifted them [here].(https://github.com/huggingface/transformers/blob/d5aebc64653d09660818109f2fac55b5e1031023/src/transformers/loss/loss_utils.py#L38)

SunMarc:
cc @ylacombe

PR #3196, opened 2024-10-24
Update transformers.deepspeed references from transformers 4.46.0 release

loadams:
# What does this PR do?

Fixes instances of `transformers.deepspeed` that were broken with the deprecation in transformers 4.46.0 released earlier today.  Updates those to use `transformers.integrations.deepspeed` since this is where those have been moved.

## Who can review?

- DeepSpeed: @muellerzr

### DIFF
--- a/tests/deepspeed/test_deepspeed.py
+++ b/tests/deepspeed/test_deepspeed.py
@@ -271,7 +271,7 @@ class DeepSpeedConfigIntegration(AccelerateTestCase):
 
         with mockenv_context(**self.dist_env):
             accelerator = Accelerator(deepspeed_plugin=deepspeed_plugin)  # noqa: F841
-            from transformers.deepspeed import is_deepspeed_zero3_enabled
+            from transformers.integrations import is_deepspeed_zero3_enabled
 
             assert is_deepspeed_zero3_enabled()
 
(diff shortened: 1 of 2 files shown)


tests/deepspeed/test_deepspeed.py
```diff
@@ -271,7 +271,7 @@ def test_init_zero3(self):
 
         with mockenv_context(**self.dist_env):
             accelerator = Accelerator(deepspeed_plugin=deepspeed_plugin)  # noqa: F841
-            from transformers.deepspeed import is_deepspeed_zero3_enabled
+            from transformers.integrations import is_deepspeed_zero3_enabled
```
loadams, line 274: @muellerzr - do you prefer if we update these to be `transformers.integrations.deepspeed` - same in other file?  Also any preference if we should add a version check here too to support older transformers versions?

muellerzr: Let's do `transformers.integrations` in case we change something later on with the file structure, as bare minimum that import structure will remain intact (Not saying it'll happen, just a form of future proofing)

muellerzr (approved):
Thanks! Let me know when we're set to merge 🫡

loadams:
> Thanks! Let me know when we're set to merge 🫡

DeepSpeed tests passed here: https://github.com/microsoft/DeepSpeed/actions/runs/11508916220

And look good to merge.

muellerzr (approved):

PR #3204, opened 2024-10-29
[Utils] `align_module`

kylesayrs:
## Purpose ##
* Add an `align_module` utility function which temporarily moves a module's parameters to its execution device. The execution device may be offloaded and defaults to the execution device of the AlignDevicesHook if present
* Prepare for other utility functions to be implemented in future PRs

## Changes ##
* Implement utility function in `accelerate.utils.modeling`, exposed through `accelerate.utils`
* Replace repeat code use in `accelerate.utils.modeling` and with newly added function

## Testing ##
* Added tests in `tests/test_modeling_utils.py.py`

## Who can review?
@SunMarc
@LysandreJik
@muellerzr
@mgoin

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1531,22 +1531,12 @@ def get_state_dict_offloaded_model(model: nn.Module):
     for name, module in model.named_modules():
         if name == "":
             continue
-        if has_offloaded_params(module):
-            original_device = module._hf_hook.execution_device
-            # assign hook execution device to cpu
-            module._hf_hook.execution_device = "cpu"
-            # onload meta tensors to execution device
-            try:
-                module._hf_hook.pre_forward(module)
-            except MemoryError:
-                raise MemoryError("Offloaded module must fit in CPU memory to call save_model!") from None
-            module_state_dict = module.state_dict()
-            # offload meta tensors from cpu
-            module._hf_hook.post_forward(module, torch.tensor([]))
-            # re-assign hook to original execution device
-            module._hf_hook.execution_device = original_device
-        else:
-            module_state_dict = module.state_dict()
+
+        try:
+            with align_module(module, "cpu"):
+                module_state_dict = module.state_dict()
+        except MemoryError:
+            raise MemoryError("Offloaded module must fit in CPU memory to call save_model!") from None
 
         for key in module_state_dict:
             # ignore placeholder parameters that are still on the meta device
@@ -1586,22 +1576,13 @@ def get_state_dict_from_offload(
     """
 
     root = module_name[: module_name.rfind(".")]  # module name without .weight or .bias
-    preforward = False
-    if has_offloaded_params(module):
-        # assign the device to which the offloaded parameters will be sent
-        original_device = module._hf_hook.execution_device
-        module._hf_hook.execution_device = device_to_put_offload
-        module._hf_hook.pre_forward(module)
-        preforward = True
 
-    for m_key in module.state_dict():
-        params = module.state_dict()[m_key]
-        if (root + f".{m_key}") in state_dict:
-            state_dict[root + f".{m_key}"] = params
-
-    if preforward:
-        module._hf_hook.post_forward(module, torch.tensor([]))
-        module._hf_hook.execution_device = original_device
+    # assign the device to which the offloaded parameters will be sent
+    with align_module(module, device_to_put_offload):
+        for m_key in module.state_dict():
+            params = module.state_dict()[m_key]
+            if (root + f".{m_key}") in state_dict:
+                state_dict[root + f".{m_key}"] = params
 
     return state_dict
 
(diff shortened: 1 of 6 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1900,51 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+    """
+    Moves a module's parameters to the specified execution device.
+
+    Args:
+        module (torch.nn.Module): Module with parameters to align.
+        execution_device (Optional[torch.device]): If provided, overrides the
+            module's execution device within the context.
+
+    Yields:
+        None: Yields control while the module's parameters are aligned to the execution device.
```
muellerzr, line 1916:
```suggestion
    Args:
        module (`torch.nn.Module`): 
            Module with parameters to align.
        execution_device (`torch.device`, *optional*): 
            If provided, overrides the module's execution device within the context.

    Yields:
        None: Yields control while the module's parameters are aligned to the execution device.
```

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1900,51 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+            )
+
+        yield
+
+        for name, param in module.named_parameters():
+            set_module_tensor_to_device(
+                module,
+                name,
+                devices[name],
+            )
```
muellerzr, line 1947: Just a suggestion, we can refactor this with `try/finally`:

```suggestion
    elif execution_device is not None:
        devices = {name: param.device for name, param in module.named_parameters()}
        try:
            for name in devices:
                set_module_tensor_to_device(module, name, execution_device)
            yield
        finally:
            for name, device in devices.items():
                set_module_tensor_to_device(module, name, device)
```

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1900,51 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+        if execution_device is not None:
+            original_device = module._hf_hook.execution_device
+            module._hf_hook.execution_device = execution_device
+
+        module._hf_hook.pre_forward(module)
+        yield
+        module._hf_hook.post_forward(module, None)
+
+        if execution_device is not None:
+            module._hf_hook.execution_device = original_device
```
muellerzr, line 1928: We can do it here too if we wanted: (it's a little more readable IMO)

```suggestion
    if has_offloaded_params(module):
        if execution_device is not None:
            original_device = module._hf_hook.execution_device
            module._hf_hook.execution_device = execution_device
        
        try:
            module._hf_hook.pre_forward(module)
            yield
        finally:
            module._hf_hook.post_forward(module, None)
            if execution_device is not None:
                module._hf_hook.execution_device = original_device
```

muellerzr (approved):
Thanks for adding the tests! Overall this looks great to me, I left a few readability suggestions 🤗

src/accelerate/utils/modeling.py
kylesayrs: Thanks!

kylesayrs:
@muellerzr Thank you for the suggestions!

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1899,44 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+@contextlib.contextmanager
+def align_module(module: torch.nn.Module, execution_device: Optional[torch.device] = None):
+    """
+    Moves a module's parameters to the specified execution device.
+
+    Args:
+        module (`torch.nn.Module`):
+            Module with parameters to align.
+        execution_device (`torch.device`, *optional*):
+            If provided, overrides the module's execution device within the context.
```
BenjaminBossan, line 1913: Let's document what happens if execution device is not provided.

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1899,44 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+
+    Yields:
+        None: Yields control while the module's parameters are aligned to the execution device.
```
BenjaminBossan, line 1916: Just my personal preference, but for context managers that don't yield anything, there is no need to document it.

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1899,44 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
     from ..hooks import AlignDevicesHook  # avoid circular import
 
     return hasattr(module, "_hf_hook") and isinstance(module._hf_hook, AlignDevicesHook) and module._hf_hook.offload
+
+
+@contextlib.contextmanager
+def align_module(module: torch.nn.Module, execution_device: Optional[torch.device] = None):
```
BenjaminBossan, line 1905: I wonder if the name could be more descriptive, like `align_module_device`.

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1899,44 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+    """
+    Moves a module's parameters to the specified execution device.
```
BenjaminBossan, line 1907:
```suggestion
    Context manager that moves a module's parameters to the specified execution device.
```

src/accelerate/utils/modeling.py
```diff
@@ -1919,3 +1899,44 @@ def has_offloaded_params(module: torch.nn.Module) -> bool:
...
+
+        try:
+            module._hf_hook.pre_forward(module)
+            yield
+        finally:
+            module._hf_hook.post_forward(module, None)
+            if execution_device is not None:
+                module._hf_hook.execution_device = original_device
+
+    elif execution_device is not None:
```
BenjaminBossan, line 1931: Just for my understanding: The functionality of this branch was not there before, right? Is this some kind of fix or is it in case the context manager is used in other situations where it may apply?

BenjaminBossan (commented):
Thanks a lot for providing this new utility function, it makes a lot of sense to me. Also, nice tests.

I just some smaller comments, no blocker.

src/accelerate/utils/modeling.py
kylesayrs: Maybe `align_module_params`?

src/accelerate/utils/modeling.py
kylesayrs: This branch is not a bug fix, `accelerate`'s behavior should be exactly the same in this branch as in main

kylesayrs: This branch simply introduces a utility function which can help simplify existing code in the `accelerate` package as well as be useful for downstream `accelerate` users, such as [llm-compressor](https://github.com/vllm-project/llm-compressor)

src/accelerate/utils/modeling.py
kylesayrs: Makes sense!

src/accelerate/utils/modeling.py
BenjaminBossan: Okay, so this is not strictly for use within accelerate in mind, makes sense.

src/accelerate/utils/modeling.py
BenjaminBossan: I'd be fine with that. I wanted to add `device` to the name as "align" does not imply this, at least not for me. It could also, for instance, be about aligning the dtype.

kylesayrs: There's a little bit of a lexical gap, as I haven't seen many uses of the word "onload". `align` comes from `AlignDevicesHook`, so on second thought I think I'd also prefer `align_module_device`

SunMarc (commented):
Very nice utility function ! Thanks for adding it !

PR #3213, opened 2024-11-03
creeate _preprare_fsdp to pre- prepare fsdp model training

### LINKED ISSUE #3209: The optimizer is not receiving the FSDP model parameters.
### System Info

```Shell
- `Accelerate` version: 1.0.1
- Platform: Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
- `accelerate` bash location: /home/a/anaconda3/envs/trans/bin/accelerate
- Python version: 3.12.7
- Numpy version: 2.1.2
- PyTorch version (GPU?): 2.5.0+cu124 (True)
- System RAM: 15.46 GB
- GPU type: NVIDIA GeForce RTX 3070 Laptop GPU
	Not found
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

See code in the pdf
[fsdp_acc.pdf](https://github.com/user-attachments/files/17599827/fsdp_acc.pdf)

### Expected behavior

The optimizer has the FSDP model parameters.

eljandoubi:
Fixes #3209

## Who can review?

- Fully-Sharded Data Parallism: @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -23,7 +23,7 @@ import re
 import shutil
 import sys
 import warnings
-from collections import OrderedDict
+from collections import Counter, OrderedDict
 from contextlib import contextmanager
 from functools import partial
 from types import MethodType
@@ -32,6 +32,7 @@ from typing import Any, Callable, Union
 import torch
 import torch.utils.hooks as hooks
 from huggingface_hub import split_torch_state_dict_into_shards
+from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
 
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
 from .data_loader import DataLoaderDispatcher, prepare_data_loader, skip_first_batches
@@ -1319,6 +1320,10 @@ class Accelerator:
                 args = self._prepare_ipex_or_xpu(*args)
         if self.fp8_backend == "TE":
             args = self._prepare_te(*args)
+        if self.distributed_type == DistributedType.FSDP:
+            # Wrap models with FSDP and update the optimizer parameters.
+            # Other types of wrapping are handled in the next if-else block.
+            args = self._prepare_fsdp(*args, device_placement=device_placement)
         if self.distributed_type == DistributedType.DEEPSPEED:
             result = self._prepare_deepspeed(*args)
         elif self.distributed_type == DistributedType.MEGATRON_LM:
@@ -1463,7 +1468,6 @@ class Accelerator:
                         self.ddp_handler.register_comm_hook(model)
             elif self.distributed_type == DistributedType.FSDP:
                 # We need to fix the optimizer *before* sharding the model
-                from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
 
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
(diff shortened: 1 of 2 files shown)


eljandoubi:
@muellerzr have you any feedback?

src/accelerate/accelerator.py
```diff
@@ -2075,6 +2079,173 @@ def _prepare_msamp(self, *args, device_placement):
...
+            return args
+
+        # Flattening weights implies that the optimizers have already been processed.
+        if next(next(iter(models.values())).named_parameters())[0].endswith("_flat_param"):
+            return args
+
+        if len(models) != len(optimizers):
+            raise ValueError(
+                f"The number of models ({len(models)}) must be equal to the number of optimizers ({len(optimizers)})."
+            )
```
BenjaminBossan, line 2105: Would it make sense to move these checks to the very start of the method?

src/accelerate/accelerator.py
```diff
@@ -2075,6 +2079,173 @@ def _prepare_msamp(self, *args, device_placement):
...
+                            model_layer_group_map[model_idx][name] = group_idx
+                            break  # Exit after the first match to avoid redundant mappings
+
+                    if name in model_layer_group_map[model_idx]:
+                        break
+
+        # Clear parameter lists.
+        for opt in optimizers.values():
+            for group in opt.param_groups:
+                group["params"] = []
```
BenjaminBossan, line 2164: Is it better to call `group["params"].clear()`? That would affect references.

BenjaminBossan (commented):
Thanks for working on this important issue.

I wonder on the overall approach to solving this issue. Right now, IIUC, the idea is to re-initialize the optimizer(s) using the FSDP-wrapped models. This could potentially be error prone and ideally should be well tested before we merge.

Would it be possible instead to check if this has happened and raise an error, then direct users towards not initializing the optimizer prematurely, and handle the initialization with accelerate so that it's in the right order? Maybe that can't work, but  I think it's worth considering different solutions.

Also, in case this is really the same issue as [this one reported in PEFT](https://github.com/huggingface/peft/issues/2205), it means that it used to work correctly in previous versions of accelerate/transformers. In that case, I wonder what changed that resulted in the issue.

src/accelerate/accelerator.py
eljandoubi: @BenjaminBossan The method! Do you mean `.prepare`? What are the benefits of doing so?

eljandoubi:
@BenjaminBossan  
- You are right; the solution needs more thorough testing. I have conducted a small test in a multi-GPU environment, but it should be tested against other configurations and models.  
- Raising an error when the optimizer is initialized prematurely could be a solution, but two cases emerge:  
  1. If Accelerate initializes the optimizer internally, either it creates a default optimizer with limited flexibility, or the user provides optimizer specifications. This would require external special treatment for FSDP, causing Accelerate to lose its unified interface for distributed training.  
  2. If the user initializes the optimizer post-FSDP setup, Accelerate also loses generalization in this case.  
- For PEFT training, which involves mixing frozen and non-frozen parameters, `use_orig_params=True` must be used. This ensures non-flattened parameters, meaning no changes are needed in the optimizer (I assume).  
- The Transformers Trainer internally uses `delay_optimizer_creation` and creates the optimizer after FSDP wrapping.

src/accelerate/accelerator.py
BenjaminBossan: I meant `_prepare_fsdp`. It is a common pattern to perform all checks as early as possible, so following this makes the code easier to understand for readers. This is especially so if there are early returns.

The reason why it's common is so that we don't do any unnecessary work if the checks fail anyway. In this case, there is no need to determine `models` or `optimizers` if we're going to raise an error later. By skipping the unnecessary work, we ensure faster execution and prevent possibly unwanted side-effects (this might not be relevant here _right now_ but code will change in the future and then it could be true).

BenjaminBossan (commented):
Thanks for the updates and explaining your testing and reasoning further.

> Raising an error when the optimizer is initialized prematurely could be a solution, but two cases emerge

These are valid concerns. At the end of the day, it's a trade off and we need to decide which cost we'd rather pay.

> The Transformers Trainer internally uses delay_optimizer_creation and creates the optimizer after FSDP wrapping.

I wonder if this logic could be moved/copied to accelerate.

> For PEFT training, which involves mixing frozen and non-frozen parameters, use_orig_params=True must be used

I don't think this is strictly necessary, we have examples in PEFT with `use_orig_params=False`. But I have to admit I don't know what exactly changes under the hood in FSDP when this parameter is set. Note also that in the linked PEFT issue, I tried setting `use_orig_params=True` and it didn't help.

eljandoubi:
- The [[PyTorch FSDP documentation](https://pytorch.org/docs/stable/fsdp.html)](https://pytorch.org/docs/stable/fsdp.html) mentions that mixed frozen and non-frozen parameters are only supported when `use_orig_params=True`. I believe this is due to the flattening process — we cannot flatten parameters with gradients together with those without. My contribution does not handle the case where `use_orig_params=False` because, in that scenario, the parameters remain unchanged.  
- Another solution would be to adapt an approach similar to the DeepSpeed integration. This involves using a single model and optimizer, and after wrapping the model with `model = FSDP(model)`, we create the optimizer as `optimizer = optimizer_class(model.parameters(), **defaults)`. See #3255

BenjaminBossan:
> * The [PyTorch FSDP documentation](https://pytorch.org/docs/stable/fsdp.html) mentions that mixed frozen and non-frozen parameters are only supported when `use_orig_params=True`. I believe this is due to the flattening process — we cannot flatten parameters with gradients together with those without.

This is true but using the right FSDP auto wrap policy, trainable and frozen parameters should be prevented from being mixed.

> Another solution would be to adapt an approach similar to the DeepSpeed integration.

If this is an option here, it sounds like the more robust solution to me. Let's wait for @muellerzr return to office and get his opinion on this.

muellerzr:
This feels a bit overengineered and the commit to transformers broke a lot of users code with saving checkpoints. For now let's not move forward with this IMO, since otherwise this would imply that the prior change broke all FSDP code, which isn't the case.

BenjaminBossan:
@eljandoubi Could you please check if https://github.com/huggingface/transformers/pull/35212 solved the issue (installing transformers 4.47.1 should contain the fix)?

eljandoubi:
@BenjaminBossan Since the `trainer` delays the creation of the optimizer, this resolves the issue for `transformers`. On the other hand, the core concept of `accelerate` is to pass the model, optimizer, dataloader, and other components simultaneously.

BenjaminBossan:
@eljandoubi Right. As [another user noted](https://github.com/huggingface/accelerate/issues/3256#issuecomment-2558266582), even with `Trainer`, it can still be problematic if users pass their own optimizer instance. I'm not sure if we can find a simpler solution, I think it's worth evaluating the possibilities. @muellerzr do you have a better idea for this situation?

PR #3220, opened 2024-11-05
[docs] use real path for `checkpoint`

faaany:
## What does this PR do?
I got the following error, when running the example code in quicktour docs: 

```bash
ValueError: `checkpoint` should be the path to a file containing a whole state dict, or the index of a sharded checkpoint, or a folder containing a sharded checkpoint or the whole state dict, but got mistralai/Mixtral-8x7B-Instruct-v0.1.
```
I think using the model_id name for checkpoint in `load_checkpoint_and_dispatch` is a bit misleading. It is better to tell users to pass their local model folder for `checkpoint`.

@muellerzr


BenjaminBossan (approved):
Thanks, makes sense.

faaany:
@muellerzr can we merge this one as well? thx so much!

PR #3223, opened 2024-11-05
remove hook for bnb 4-bit 

SunMarc:
# What does this PR do ?

This PR relax the condition that 4-bit bnb models needs to have a hook to work. It is not necessary as we can move the 4-bit model since bnb 0.43.2.

cc @sayakpaul

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -355,7 +355,9 @@ def dispatch_model(
     # We need to force hook for quantized model that can't be moved with to()
     if getattr(model, "quantization_method", "bitsandbytes") == "bitsandbytes":
         # since bnb 0.43.2, we can move 4-bit model
-        if getattr(model, "is_loaded_in_8bit", False) or (getattr(model, "is_loaded_in_4bit", False) and not is_bnb_available(min_version="0.43.2")):
+        if getattr(model, "is_loaded_in_8bit", False) or (
+            getattr(model, "is_loaded_in_4bit", False) and not is_bnb_available(min_version="0.43.2")
+        ):
             force_hooks = True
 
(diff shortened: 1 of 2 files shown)


matthewdouglas (approved):
LGTM apart from the quality check. Thanks!

BenjaminBossan:
Some tests in PEFT are failing since this PR has been merged. Here is a simple reproducer:

```python
import torch
from transformers import AutoModelForSequenceClassification, BitsAndBytesConfig

model_id = "facebook/opt-125m"
model = AutoModelForSequenceClassification.from_pretrained(
    model_id,
    quantization_config=BitsAndBytesConfig(load_in_4bit=True),
)
input = torch.arange(10).view(-1, 1)  # <= on CPU

# RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
model(input)
# _hf_hook no longer exists
assert hasattr(model.score, "_hf_hook")
```

I think it can be argued that the code was incorrect to begin with but there is the risk that with the next accelerate release, a bunch of similar code out there could start breaking.

SunMarc:
> I think it can be argued that the code was incorrect to begin with but there is the risk that with the next accelerate release, a bunch of similar code out there could start breaking.

I think we should add a note in the next release that the user should specify which device the bnb model will be loaded on + move the inputs to the right device. 

```python 
import torch
from transformers import AutoModelForSequenceClassification, BitsAndBytesConfig

model_id = "facebook/opt-125m"
model = AutoModelForSequenceClassification.from_pretrained(
    model_id,
    quantization_config=BitsAndBytesConfig(load_in_4bit=True),
    device_map="cuda"
)
input = torch.arange(10).view(-1, 1)  # <= on CPU
input.to("cuda")

model(input)
```

PR #3228, opened 2024-11-08
[docs] update set-seed

faaany:
## What does this PR do?
The current `set-seed` in documentation only includes the cuda part, which is out-dated. This PR updates it with the current implementation.

 @muellerzr

### DIFF
--- a/docs/source/concept_guides/performance.md
+++ b/docs/source/concept_guides/performance.md
@@ -43,13 +43,22 @@ Why is this important? Under the hood this will set **5** different seed setting
     random.seed(seed)
     np.random.seed(seed)
     torch.manual_seed(seed)
-    torch.cuda.manual_seed_all(seed)
+    if is_xpu_available():
+        torch.xpu.manual_seed_all(seed)
+    elif is_npu_available():
+        torch.npu.manual_seed_all(seed)
+    elif is_mlu_available():
+        torch.mlu.manual_seed_all(seed)
+    elif is_musa_available():
+        torch.musa.manual_seed_all(seed)
+    else:
+        torch.cuda.manual_seed_all(seed)
     # ^^ safe to call this function even if cuda is not available
     if is_torch_xla_available():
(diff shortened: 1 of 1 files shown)


faaany:
@muellerzr @BenjaminBossan

BenjaminBossan (approved):
Thanks for updating the docs.

IMO, it would also be fine to remove the code snippet completely (as it's just a matter of time until it gets out of date again) and just explain in words what's happening.

faaany:
> Thanks for updating the docs.
> 
> IMO, it would also be fine to remove the code snippet completely (as it's just a matter of time until it gets out of date again) and just explain in words what's happening.

agree! let me update

BenjaminBossan:
> I think code is easier to read and understand. Only using cuda is a bit unfair. Maybe I can update other devices in a comment?

My main concern is mostly the example and the actual code getting out of sync. It's not a big deal, would just be nice to prevent it.

muellerzr (commented):
Thanks, there is always a fear of what would happen if code gets out of sync, esp as even this code doesn't quite fully align to what we're up to. (e.g. there's now a by_index). 

While it's not per-se "fair", I'm not sure if it's more or less readable now than before for our docs. It is a concept guide, so we can show more code. Just wish we could show the source itself. 

I think we could get by just saying:
```python
    random.seed(seed)
    np.random.seed(seed)
    torch.manual_seed(seed)
    torch.cuda.manual_seed_all(seed) # or torch.npu.manual_seed_all, etc
    # ^^ safe to call this function even if cuda/backend is not available
    if is_torch_xla_available():
        xm.set_rng_state(seed)
```

Since we care more about `manual_seed_all` than the backend specifically

faaany:
Thanks for the feedback! Agree. Code looks much cleaner now. @muellerzr @BenjaminBossan

muellerzr (approved):
Thanks!

PR #3231, opened 2024-11-08
[RFC] Support FSDP2

### LINKED ISSUE #2873: Plan to support FSDP2?
### System Info

```Shell
FSDP2 provides smaller memory footprint, compatibility with torch compile, and more flexibility due to per param sharding. Does huggingface have plan to support FSDP2?

https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md
```

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

n/a

### Expected behavior

n/a

kmehant:
# What does this PR do?

Prototype implementation for porting from FSDP V1 to FSDP V2. There are couple of open questions in this PR that would need comments and discussion.

1. Do we want to maintain FSDP V1 as is and add a experimental parallel to FSDP V2?
2. When we want to maintain 2 versions, should we maintain separate FSDP plugins and distributed types for each versions?
3. For HF/transformers users, using fsdp_config, how we want to allow them to choose between these versions?
4. How we want prepare 2D mesh for HSDP, should that be an input from user?

### Preliminary run of this PR and results
The current version of the PR has been tested for basic functionality (full shard) and compared with previous FSDP V1 implementation.

|  Key    |  Value |
| -------- | ------- |
| Model  | Maykeye/TinyLLama-v0   |
| Mesh size | 2 GPUs     |
| sharding    | full shard  |

#### Memory

Screenshot 2024-11-09 at 12 50 10 AM

#### Loss Parity

Screenshot 2024-11-09 at 12 59 56 AM

#### Throughput

TODO

Fixes #2873

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1499,6 +1499,7 @@ class Accelerator:
                         "limit_all_gathers": fsdp_plugin.limit_all_gathers,
                         "device_id": self.device,
                     }
+                    print("using FSDP in accelerate")
                     model = FSDP(model, **kwargs)
                     if fsdp_plugin.activation_checkpointing:
                         from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1464,13 +1464,15 @@ class Accelerator:
             elif self.distributed_type == DistributedType.FSDP:
                 # We need to fix the optimizer *before* sharding the model
                 from torch.distributed.fsdp.fully_sharded_data_parallel import FullyShardedDataParallel as FSDP
-
+                from torch.distributed._composable.fsdp import fully_shard, MixedPrecisionPolicy, CPUOffloadPolicy, FSDPModule, OffloadPolicy
+                from torch.distributed.fsdp.api import ShardingStrategy
                 # Check if the model is already a FSDP model due to `Manual Wrapping` and if so,
                 # don't wrap it again
                 # In case the model is already compiled using PyTorch 2.0 and the wrapped model in it
                 # is a FSDP model, don't wrap it again
-                is_type_fsdp = isinstance(model, FSDP) or (
-                    is_compiled_module(model) and isinstance(model._orig_mod, FSDP)
+                # We check for FSDPModule instead of FSDP class for FSDP v2
+                is_type_fsdp = isinstance(model, FSDPModule) or (
+                    is_compiled_module(model) and isinstance(model._orig_mod, FSDPModule)
                 )
 
                 if not is_type_fsdp:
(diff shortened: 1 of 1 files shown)


raghukiran1224:
@ByronHsu FYI - thoughts?

SumanthRH:
@kmehant thanks for starting this PR! I was looking at FSDP2 support in `accelerate` and landed here! 

> Do we want to maintain FSDP V1 as is and add a experimental parallel to FSDP V2?

Yes please! It looks like FSDP2 will be in a public API in the next torch release (2.6): https://github.com/pytorch/pytorch/commit/d815efcc3632f4a670b0aeba61d644cd52987e29 , so maybe things are somewhat stable ?  But many of the older config parameters (like `auto_wrap_policy`) are simply not there in V2 so I'd prefer if `accelerate` users get time to migrate. 

> When we want to maintain 2 versions, should we maintain separate FSDP plugins and distributed types for each versions?

Hmm if the new API had supported most of the V1 configurations, I would think having only a feature flag would be enough -i.e something like `ACCELERATE_FSDP2_ENABLED`. But looking at the API differences: https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md#fsdp1--fsdp2-api-differences 
I think just having a feature flag + a different plugin `FullyShardedDataParallelPluginV2` is better. 

> For HF/transformers users, using fsdp_config, how we want to allow them to choose between these versions?

It looks like `accelerate` doesn't do much validation for the config parameters in `fsdp_config` until plugin initialization. So in the accelerate config, specifying `enable_fsdp2` should be fine , and users would be expected to list only v2 parameters in `fsdp_config`  - this is validated when `FullyShardedDataParallelPluginV2` is initialized.

SumanthRH:
cc @muellerzr curious to know if this already in the pipeline internally from HF!

src/accelerate/accelerator.py
```diff
@@ -1498,8 +1500,100 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+                        fsdp2_kwargs["mp_policy"] = CPUOffloadPolicy(
+                            # pin_memory= cannot be deduced from FSDP1 args and has to come from user
+                            # offloads params is the default behaviour
+                            )
+
+                    #######
+                    # auto_wrap_policy is not yet supported by FSDP2
+                    # therefore manual wrapping has to be done like below
+                    #######
+                    for layer in model.model.layers:
```
kyleliang919, line 1590: This one doesn't seem to apply to general use case.
`stack = [model]
ordered_modules = []
while stack:
	current_modules = stack.pop()
	for _, attr in current_module.__dict__.items():
		if isinstance(attr, torch.nn.Module):
		    stack.append(attr)
	ordered_modules.append(current_module)

for each in ordered_modules[::-1]:
	fully_shard(each, **fsdp2_kwargs)`

src/accelerate/accelerator.py
```diff
@@ -1498,8 +1500,100 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+                        fully_shard(layer, **fsdp2_kwargs)
+                    fully_shard(model, **fsdp2_kwargs)
```
psinger, line 1592: I am a bit unsure what is the reason for this final `fully_shard` here? I also see this in original tutorial. It is causing some issue related to:
`RuntimeError: aten.mul.Tensor: got mixed torch.Tensor and DTensor, need to convert all torch.Tensor to DTensor before calling distributed operators!`

kmehant:
heartbeat ping

kwanUm:
Any ETA on this?

kmehant:
There are three threads to this:
1. current PR - https://github.com/huggingface/accelerate/pull/3231 (from me)
2. proposing nd parallelism with TP + FSDP combinations + not removing FSDP1 as is - https://github.com/huggingface/accelerate/pull/3395 (from me)
3. As well adding FSDP2 https://github.com/huggingface/accelerate/pull/3394 (I am guessing this effort from HF internal team)

I am guessing (3) can supersede (1) and ublock (2).

I will let @muellerzr @SunMarc to comment. Thanks

SunMarc:
cc @ S1ro1

muellerzr:
Yes, that is correct. 3 will supersede as we need to be careful to *still maintain the original FSDP in its entirety* for backwards compatibility purposes. However we will make sure the original authors here still get credit for their efforts in starting this.

src/accelerate/accelerator.py
```diff
@@ -1498,9 +1500,100 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
...
+                    
+                    #######
+                    # offload policy can be mapped from FSDP1 to FSDP2 arg classes
+                    # pinning memory seems to be a new feature to FSDP2
+                    # offloading params is the default behaviour
+                    #######
+
+                    if kwargs["cpu_offload"] is not None and kwargs["cpu_offload"].offload_params:
+                        # CPUOffloadPolicy is from the new _composable design
+                        fsdp2_kwargs["mp_policy"] = CPUOffloadPolicy(
```
cyr0930, line 1581: This should be
```fsdp2_kwargs["offload_policy"] = CPUOffloadPolicy(...)```

src/accelerate/accelerator.py
cyr0930: current_module to current_modules

cyr0930:
This PR can be closed as https://github.com/huggingface/accelerate/pull/3394 this already merged into main?

S1ro1:
@cyr0930 Yes, if @kmehant doesn't think there's anything extra then it can be closed.

kmehant:
Please see https://github.com/huggingface/accelerate/pull/3394

PR #3234, opened 2024-11-11
[docs] add XPU to profiler documentation and fix minor bugs

faaany:
## What does this PR do?
As the title suggest, this PR add one-line comment for XPU users and also fix 2 minor bugs. 

The first one is that the missing input data preparation, which leads to the following error on CUDA:
```bash
RuntimeError: Input type (torch.FloatTensor) and weight type (torch.cuda.FloatTensor) should be the same or input should be a MKLDNN tensor and weight is a dense tensor
```
The second one is the small training steps in the example code, which leads to no print message and no `trace.json` is saved. 

@muellerzr


faaany:
I realized that the change in this PR is also included in PR #3227. So we can close this one.

PR #3244, opened 2024-11-18
Fix : get_balanced_memory when using multi gpus with small models or quantized models with a large vocabulary

MekkCyber:
# What does this PR do?
Fixes https://github.com/huggingface/transformers/issues/34751, and part of https://github.com/huggingface/transformers/issues/34706. 
### Explanation of `get_balanced_memory` Behavior and Handling of Large `embed_tokens` Layers  

When a small model or a relatively large quantized model is loaded using `device_map=auto`, the function `get_balanced_memory` calculates the maximum memory usage for each visible device. Its goal is to distribute the model evenly across all devices while reserving all available memory on the last device.  

#### Issue with Large `embed_tokens` Layers  

For models with a small number of parameters but a large vocabulary size (e.g., *Gemma2 2B*), the `embed_tokens` layer can consume a significant amount of memory. This layer might exceed the `max_memory` limit on the initial devices. As a result, the `infer_auto_device_map` function bypasses all earlier devices, placing the `embed_tokens` layer—and all subsequent layers—on the last device.  

A similar issue arises with quantized models. Since embedding layers are often **not quantized**, the `max_memory` per device might be insufficient to accommodate the `embed_tokens` layer, leading to the same behavior.  

#### Improvement in This PR  

This pull request addresses the issue by comparing the size of the `embed_tokens` layer to the memory available per GPU (`per_gpu`). It adjusts the memory allocation strategy to ensure that `embed_tokens` can be distributed appropriately across devices, improving the handling of models with large embeddings.  

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1066,11 +1066,11 @@ def get_balanced_memory(
     )
     # The last device is left with max_memory just in case the buffer is not enough.
     for idx in gpus_idx_list[:-1]:
-        if idx == 0 and not low_zero and module_sizes['model.embed_tokens'] > per_gpu * 0.9: 
-            max_memory[idx] = min(module_sizes['model.embed_tokens'] * 1.3, max_memory[idx])
-        elif idx == 1 and low_zero and module_sizes['model.embed_tokens'] > per_gpu * 0.9 :
-            max_memory[idx] = min(module_sizes['model.embed_tokens'] * 1.3, max_memory[idx])
-        else :
+        if idx == 0 and not low_zero and module_sizes["model.embed_tokens"] > per_gpu * 0.9:
+            max_memory[idx] = min(module_sizes["model.embed_tokens"] * 1.3, max_memory[idx])
+        elif idx == 1 and low_zero and module_sizes["model.embed_tokens"] > per_gpu * 0.9:
+            max_memory[idx] = min(module_sizes["model.embed_tokens"] * 1.3, max_memory[idx])
+        else:
             max_memory[idx] = min(max_memory[0] if low_zero and idx == 0 else per_gpu, max_memory[idx])
 
     if low_zero:
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1066,7 +1066,12 @@ def get_balanced_memory(
     )
     # The last device is left with max_memory just in case the buffer is not enough.
     for idx in gpus_idx_list[:-1]:
-        max_memory[idx] = min(max_memory[0] if low_zero and idx == 0 else per_gpu, max_memory[idx])
+        if idx == 0 and not low_zero and module_sizes['model.embed_tokens'] > per_gpu * 0.9: 
+            max_memory[idx] = min(module_sizes['model.embed_tokens'] * 1.3, max_memory[idx])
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/modeling.py
```diff
@@ -1066,7 +1066,12 @@ def get_balanced_memory(
     )
     # The last device is left with max_memory just in case the buffer is not enough.
     for idx in gpus_idx_list[:-1]:
-        max_memory[idx] = min(max_memory[0] if low_zero and idx == 0 else per_gpu, max_memory[idx])
+        if idx == 0 and not low_zero and module_sizes["model.embed_tokens"] > per_gpu * 0.9:
+            max_memory[idx] = min(module_sizes["model.embed_tokens"] * 1.3, max_memory[idx])
+        elif idx == 1 and low_zero and module_sizes["model.embed_tokens"] > per_gpu * 0.9:
+            max_memory[idx] = min(module_sizes["model.embed_tokens"] * 1.3, max_memory[idx])
+        else:
+            max_memory[idx] = min(max_memory[0] if low_zero and idx == 0 else per_gpu, max_memory[idx])
```
SunMarc, line 1074: Not every model have their embedding layers as `model.embed_tokens`. Since this is specific to transformers, I don't think this should live in accelerate too. We can modify `max_memory` directly there as it is computed. We are also trying to tackle similar issues with this PR https://github.com/huggingface/accelerate/pull/3066#issuecomment-2417065301. 

Maybe a good solution would be to check if there is a module that is > per_gpu, return a message saying that the model is unbalanced which will lead to the whole model being put on only one device and propose to the user to use "sequential" mode instead ? Or we could also do as you suggested and modify `max_memory` with the largest module size.

MekkCyber: Thanks for the suggestion @SunMarc, I updated the code to use the largest leave module size instead of hardcoding the embed_tokens layer

src/accelerate/utils/modeling.py
```diff
@@ -1065,8 +1065,14 @@ def get_balanced_memory(
         )
     )
     # The last device is left with max_memory just in case the buffer is not enough.
+    max_leave_size = max([module_sizes[leave] for leave in leaves])
     for idx in gpus_idx_list[:-1]:
-        max_memory[idx] = min(max_memory[0] if low_zero and idx == 0 else per_gpu, max_memory[idx])
+        if idx == 0 and not low_zero and max_leave_size > per_gpu * 0.9:
+            max_memory[idx] = min(max_leave_size * 1.3, max_memory[idx])
+        elif idx == 1 and low_zero and max_leave_size > per_gpu * 0.9:
+            max_memory[idx] = min(max_leave_size * 1.3, max_memory[idx])
```
SunMarc, line 1073: You are taking the minimum, is this expected ?

MekkCyber: Yes we take the minimum between the max_memory of the gpu and the gpu space needed on the device. So if the space needed exceeds the space available on the gpu, we only allocate the space available.

SunMarc (commented):
Could you add a test that fails before this PR but is fixed now ? Also it would be nice to try with the models that have this issue.

PR #3247, opened 2024-11-20
Use `numpy._core` instead of `numpy.core`

qgallouedec:
# What does this PR do?

We've been getting this warning:

>   /fsx/qgallouedec/miniconda3/envs/trl/lib/python3.11/site-packages/accelerate/utils/other.py:220: DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.multiarray.
    np.core.multiarray._reconstruct,

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -215,7 +215,7 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
 
 # The following are considered "safe" globals to reconstruct various types of objects when using `weights_only=True`
 # These should be added and then removed after loading in the file
-np_core = np._core if Version(np.__version__) >= Version("2.0.0") else np.core
+np_core = np._core if is_numpy_available("2.0.0") else np.core
 TORCH_SAFE_GLOBALS = [
     # numpy arrays are just numbers, not objects, so we can reconstruct them safely
     np_core.multiarray._reconstruct,
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -23,6 +23,7 @@ from typing import OrderedDict
 
 import numpy as np
 import torch
+from packaging import version
 from packaging.version import Version
 from safetensors.torch import save_file as safe_save_file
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
The issue here is that `np._core` does not exist in older numpy versions (probably < 2.0 but I haven't checked), so this would fail there (as happens in the failing doc build). We would need to do a version check and use either `np.core` or `np._core`.

qgallouedec:
good point!

The very first version that includes `np._core` is 1.26.1.

What about something like

```python
np_core = np._core if version.parse(np.__version__) >= version.parse("1.26.1") else np.core
```

alternatively we can do something like 

```python
np_core = getattr(np, "_core", np.core)
```

but I find it less intuitive to know when the condition is met or not, and we'll probably forget to remove this if/else once we've increased the min version of numpy

qgallouedec:
wait it seems like `np._core` for np<2.0 is not the same as for np>=2.0. Checking right now

BenjaminBossan (approved):
Thanks for fixing the warning.

SunMarc (approved):
Nice thanks !

src/accelerate/utils/other.py
```diff
@@ -215,9 +215,10 @@ def save(obj, f, save_on_each_node: bool = False, safe_serialization: bool = Fal
 
 # The following are considered "safe" globals to reconstruct various types of objects when using `weights_only=True`
 # These should be added and then removed after loading in the file
+np_core = np._core if Version(np.__version__) >= Version("2.0.0") else np.core
```
SunMarc, line 218: could you use is_numpy_available function instead ?

qgallouedec: Done in bd0e022e6b6e0e505cf11b6b8ce2d103f9e96743

PR #3248, opened 2024-11-22
Fix: Resolve #3060, `preload_module_classes` is lost for nested modules

### LINKED ISSUE #3060: `preload_module_classes` is lost in attach_execution_device_hook.
### System Info

https://github.com/huggingface/accelerate/blob/3fcc9461c4fcb7228df5e5246809ba09cfbb232e/src/accelerate/hooks.py#L439
Should we pass preload_module_classes  to the next calling? such as
```
        attach_execution_device_hook(child, execution_device, tied_params_map=tied_params_map, 
 preload_module_classes=preload_module_classes)
```

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

Step into the code is enough.

### Expected behavior

Identify the bug.

wejoncy:
# What does this PR do?
We ran into this issue when try to load [VPTQ](https://github.com/microsoft/VPTQ/tree/main)  models by multiple devices, the `VQLinear` layer has embending as its sub-module. This function should move all constant buffers/parameters to the corresponding devices from Meta device.
So `add_hook_to_module` should pass `preload_module_classes` to the next recursive calling for the nested modules.

https://github.com/microsoft/VPTQ/blob/ac7258f461e214ca705f5895513f314576750528/vptq/layers/vqlinear.py#L160C18-L160C18

This PR is to fix https://github.com/huggingface/accelerate/issues/3060

- Big modeling: @SunMarc

### DIFF
--- a/tests/test_accelerator.py
+++ b/tests/test_accelerator.py
@@ -762,3 +762,65 @@ class AcceleratorTester(AccelerateTestCase):
             assert torch.allclose(original_linear1, new_linear1)
             assert torch.allclose(original_batchnorm, new_batchnorm)
             assert torch.allclose(original_linear2, new_linear2)
+
+    @require_cuda
+    def test_nested_hook(self, use_safetensors):
+        from transformers.modeling_utils import PretrainedConfig, PreTrainedModel
+
+        class MyLinear(torch.nn.Module):
+            def __init__(self, device=None, dtype=None):
+                factory_kwargs = {"device": device, "dtype": dtype}
+                super().__init__()
+                self.centroid = torch.nn.Embedding(1, 2)
+                self.indices = torch.nn.parameter(torch.empty((1, 2, 2), **factory_kwargs))
+
+            def forward(self, x):
+                orig_shape = x.shape
+                x = torch.abs(x + self.indices).long()
+                x = x % 2
+                x = x.sum(-1)
+                x = (self.centroid.weight + x).reshape(orig_shape)
+                return x
+
+        class MySubModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.layer = MyLinear()
+
+            def forward(self, x):
+                return self.layer(x)
+
+        class MyModel(PreTrainedModel):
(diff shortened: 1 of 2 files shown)


SunMarc (approved):
LGTM ! Thanks for fixing ! Could you share which specific issue you where having ? Also if you can implement a test, it would be nice but not needed to merged this PR. You can fix the CI with `make style`

muellerzr (commented):
agreed with Marc, some more details to this issue and adding a test is needed.

wejoncy:
Hi @muellerzr  and @SunMarc  Thanks for your review.  I have briefly described the issue in overview of this PR. Besides I will try to construct such a test case.

tests/test_accelerator.py
```diff
@@ -762,3 +762,64 @@ def test_save_model_with_stateful_dataloader(self, use_safetensors, tied_weights
...
+            with init_empty_weights():
+                my_model = MyModel(config)
+            my_model = load_checkpoint_and_dispatch(
+                my_model,
+                checkpoint=check_point,
+                max_memory={"cpu": 60, 0: 60},
+                device_map="auto",
+                no_split_module_classes=["MySubModel"],
+                offload_folder=offload_folder,
+                preload_module_classes=["VQuantLinear"],
```
SunMarc, line 820: There is no VQuantLinear created

tests/test_accelerator.py
```diff
@@ -762,3 +762,64 @@ def test_save_model_with_stateful_dataloader(self, use_safetensors, tied_weights
             assert torch.allclose(original_linear1, new_linear1)
             assert torch.allclose(original_batchnorm, new_batchnorm)
             assert torch.allclose(original_linear2, new_linear2)
+
+    @require_cuda
```
SunMarc, line 766: we need a require_transformers decorator if you are going to import transformers here. Otherwise, we can try to not use it and just save the model with 

```
checkpoint = os.path.join(tmp_dir, "pt_model.bin")
torch.save(model.state_dict(), checkpoint)
```

PR #3252, opened 2024-11-22
[WIP] FEAT Decorator to purge accelerate env vars

BenjaminBossan:
# What does this PR do?

In some circumstances, calling certain classes or functions can result in accelerate env vars being set and not being cleaned up afterwards. As an example, when calling:

`TrainingArguments(fp16=True, ...)`

The following env var will be set:

`ACCELERATE_MIXED_PRECISION=fp16`

This can affect subsequent code, since the env var takes precedence over `TrainingArguments(fp16=False)`. This is especially relevant for unit testing, where we want to avoid the individual tests to have side effects on one another. Decorate the unit test function or whole class with this decorator to ensure that after each test, the env vars are cleaned up. This works for both `unittest.TestCase` and normal classes (pytest) and standalone functions; it also works when decorating the parent class and in conjunction with other decorators.

## Notes

In its current state, this PR adds the new decorator and tests it, but the decorator is not yet applied to potentially problematic functions or classes.

Personally, I'm not super happy with the implementation, as it turned out to be more complicated than I initially thought (mainly because of the subclassing). I tested this decorator on the initial test failure in PEFT and it solved it. However, it requires to decorate each and every class/function that could potentially have side effects. It is thus easy to miss something.

## Before submitting
      Pull Request section?
- [ ] ~~Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.~~ Internal discussion on Slack

### DIFF
--- a/docs/source/package_reference/utilities.md
+++ b/docs/source/package_reference/utilities.md
@@ -174,6 +174,8 @@ When setting up 🤗 Accelerate for the first time, rather than running `acceler
 
 [[autodoc]] utils.environment.override_numa_affinity
 
+[[autodoc]] utils.purge_accelerate_environment
+
 ## Memory
 
 [[autodoc]] utils.find_executable_batch_size
--- a/src/accelerate/utils/__init__.py
+++ b/src/accelerate/utils/__init__.py
@@ -71,6 +71,7 @@ from .environment import (
     parse_choice_from_env,
     parse_flag_from_env,
     patch_environment,
+    purge_accelerate_environment,
     set_numa_affinity,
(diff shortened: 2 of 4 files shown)


muellerzr (approved):
Thanks! LG2M since we both agreed offline :)

BenjaminBossan:
@muellerzr how much will this still be needed if https://github.com/huggingface/transformers/pull/34886 is merged?

SunMarc (approved):
LGTM ! Thanks for the nice tests

PR #3255, opened 2024-11-24
Select the DeepSpeedCPUOptimizer based on the original optimizer class.

eljandoubi:
# What does this PR do?

For DeepSpeed optimizer offloading, select the appropriate optimizer based on the original optimizer class.

## Who can review?

- DeepSpeed: @muellerzr

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -30,6 +30,7 @@ from types import MethodType
 from typing import Any, Callable, Union
 
 import torch
+import torch.optim.adagrad
 import torch.utils.hooks as hooks
 from huggingface_hub import split_torch_state_dict_into_shards
 
@@ -83,6 +84,7 @@ from .utils import (
     has_offloaded_params,
     is_bf16_available,
     is_bitsandbytes_multi_backend_available,
+    is_bnb_available,
     is_deepspeed_available,
     is_ipex_available,
     is_lomo_available,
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
Thanks for this update. Just wondering aloud: If the optimizer is neither Adam, nor Adagrad, nor Lion, what would be the right thing to do? As is, Adam is used, is that correct?

eljandoubi:
@BenjaminBossan By défaut, I have kept the Adam as it was for everyone torch optimizer.

src/accelerate/accelerator.py
```diff
@@ -1839,10 +1841,42 @@ def _prepare_deepspeed(self, *args):
...
+                        # For DeepSpeedCPULion
+                        if is_bnb_available(min_version="0.38.0") and compare_versions("deepspeed", ">=", "0.11.0"):
+                            from bitsandbytes.optim import Lion, Lion32bit
+
+                            if isinstance(optimizer, (Lion, Lion32bit)) and optimizer.optim_bits == 32:
+                                from deepspeed.ops.lion import DeepSpeedCPULion
+
+                                optimizer_class = DeepSpeedCPULion
+
+                        optimizer = optimizer_class(optimizer.param_groups, **defaults)
```
muellerzr, line 1879: This seems a bit bulky for the `Accelerator`, can we abstract this out to a deepspeed util of `map_pytorch_optim_to_deepspeed`?

muellerzr (commented):
Thanks for doing this! Overall I'm a fan, however let's abstract this out to a util func to keep the Accelerator a bit more readable

eljandoubi:
@muellerzr What do you think of it like this?

PR #3270, opened 2024-12-02
[docs] no hard-coding cuda

faaany:
## What does this PR do?
This PR adds xpu besides cuda and replaces the hard-coded cuda with device-agnostic device. 

@muellerzr

### DIFF
--- a/docs/source/usage_guides/big_modeling.md
+++ b/docs/source/usage_guides/big_modeling.md
@@ -64,7 +64,8 @@ Now that the model is fully dispatched, you can perform inference.
 
 ```py
 input = torch.randn(2,3)
-input = input.to(model.device.type)
+device_type = next(iter(model.parameters())).device.type
+input = input.to(device_type)
 output = model(input)
 ```
 
--- a/docs/source/usage_guides/big_modeling.md
+++ b/docs/source/usage_guides/big_modeling.md
@@ -92,7 +92,8 @@ model = load_checkpoint_and_dispatch(
 )
 
 input = torch.randn(2,3)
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
LGTM.

docs/source/usage_guides/big_modeling.md
```diff
@@ -64,7 +64,7 @@ Now that the model is fully dispatched, you can perform inference.
 
 ```py
 input = torch.randn(2,3)
-input = input.to("cuda")
+input = input.to(model.device.type)
```
XuehaiPan, line 67: I'm afraid the `model` does not always have a `device` member.

faaany: also when the model is loaded with `load_checkpoint_and_dispatch`? I think when we consider the context, model should be a HF ModelClass, which will have device as an attribute. But pls correct me if I am wrong.

muellerzr: We don't, esp as the prior sentence considers a regular pytorch model.

We can instead do one of two options:

1. Bring in the `PartialState` for this tutorial
2. Go a bit easier and use the model itself

(More a fan of 2 here)

```suggestion
device_type = next(iter(model.parameters())).device.type
input = input.to(device_type)
```

muellerzr (commented):
Thanks! Let's make this a bit more agnostic as @XuehaiPan recommends :)

faaany:
Thanks for the suggestion! Code Updated. @muellerzr

muellerzr (approved):
Thanks!

Issue #3281, opened 2024-12-09
`Accelerate.unwrap` should fully unwrap the model.


ggoggam:
### System Info

```Shell
- `Accelerate` version: 0.28.0
- Platform: Linux-5.15.0-124-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Numpy version: 1.26.3
- PyTorch version (GPU?): 2.5.1+cu118 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 250.55 GB
- GPU type: NVIDIA T4
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

When the model is wrapped in both distributed wrapper (e.g. `DistributedDataParallel` or `DeepSpeedEngine`) and compiled module (e.g. `OptimizedModule`), calling `Accelerator.unwrap_model` should return a fully unwrapped model (`torch.nn.Module`). Instead, it currently returns the distributed wrapper. 

```python
import accelerate

# some model
model = ...
optimizer = ...

accelerator = accelerate.Accelerator(
    deepspeed_plugin=accelerate.utils.DeepSpeedPlugin(
        **{"zero_stage": 2, "train_micro_batch_size_per_gpu": 4}
    ),
    dynamo_backend="INDUCTOR"
)
model, optimizer = accelerator.prepare(model, optimizer)

# unwrap
unwrapped = accelerator.unwrap_model(model)

# assert
# currently raises error
assert model == unwrapped
```

### Expected behavior

`Accelerator.unwrap_model` should return fully unwrapped model, i.e. `torch.nn.Module`.

PR #3282, opened 2024-12-09
Add `keep_torch_compile` param to `unwrap_model` and `extract_model_from_parallel` for distributed compiled model.

ggoggam:
# What does this PR do?
This PR fixes the unexpected behavior of `Accelerator.unwrap_model` (Issue https://github.com/huggingface/accelerate/issues/3281). Right now, if the model is wrapped in both distributed wrapper (e.g. `DistributedDataParallel` or `DeepSpeedEngine`) and compiled module (`OptimizedModule`) it only unwraps the distributed module. This behavior arises from the following code in L80 of `utils/others.py`: https://github.com/huggingface/accelerate/blob/cb8b7c637a8588668c52bd306f9b2828f69d9585/src/accelerate/utils/other.py#L80

Instead of checking for compiled model both **before and after** unwrapping distributed wrapper, the current code only checks for compilation before unwrapping the distributed wrapper. If the model is wrapped in both, `is_compiled` will be set to `False` and won't unwrap the model fully, resulting in unexpected behavior (users expect fully unwrapped model before saving, but gets `OptimizedModule` instead, which may result in an error when loading the state dict due to key mismatch).

## Before submitting
      Pull Request section?
      to it if that's the case.

## Related Issue
- https://github.com/huggingface/accelerate/issues/3281

## Who can review?
@muellerzr @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -2601,7 +2601,7 @@ class Accelerator:
         """
         return pad_across_processes(tensor, dim=dim, pad_index=pad_index, pad_first=pad_first)
 
-    def unwrap_model(self, model, keep_fp32_wrapper: bool = True):
+    def unwrap_model(self, model, keep_fp32_wrapper: bool = True, keep_torch_compile: bool = True):
         """
         Unwraps the `model` from the additional layer possible added by [`~Accelerator.prepare`]. Useful before saving
         the model.
@@ -2611,7 +2611,8 @@ class Accelerator:
                 The model to unwrap.
             keep_fp32_wrapper (`bool`, *optional*, defaults to `True`):
                 Whether to not remove the mixed precision hook if it was added.
-
+            keep_torch_compile (`bool`, *optional*, defaults to `True`):
+                Whether to not unwrap compiled model if compiled.
         Returns:
             `torch.nn.Module`: The unwrapped model.
 
@@ -2632,7 +2633,7 @@ class Accelerator:
         MyModel
         ```
         """
-        return extract_model_from_parallel(model, keep_fp32_wrapper)
+        return extract_model_from_parallel(model, keep_fp32_wrapper, keep_torch_compile)
 
     def wait_for_everyone(self):
         """
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -59,7 +59,9 @@ def is_compiled_module(module):
     return isinstance(module, torch._dynamo.eval_frame.OptimizedModule)
 
 
-def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive: bool = False):
+def extract_model_from_parallel(
+    model, keep_fp32_wrapper: bool = True, keep_torch_compile: bool = True, recursive: bool = False
+):
     """
     Extract a model from its distributed containers.
 
@@ -68,6 +70,8 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive
             The model to extract.
         keep_fp32_wrapper (`bool`, *optional*):
             Whether to remove mixed precision hooks from the model.
+        keep_torch_compile (`bool`, *optional*):
+            Whether to unwrap compiled model.
         recursive (`bool`, *optional*, defaults to `False`):
             Whether to recursively extract all cases of `module.module` from `model` as well as unwrap child sublayers
             recursively, not just the top-level distributed containers.
@@ -124,7 +128,7 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive
         if getattr(model, "_converted_to_transformer_engine", False):
             convert_model(model, to_transformer_engine=False)
 
-    if is_compiled:
+    if keep_torch_compile and is_compiled:
         compiled_model._orig_mod = model
         model = compiled_model
 
--- a/tests/test_accelerator.py
+++ b/tests/test_accelerator.py
@@ -611,6 +611,30 @@ class AcceleratorTester(AccelerateTestCase):
         model_loaded = pickle.loads(pickle.dumps(model))
         model_loaded(inputs)
 
+    def test_can_unwrap_distributed_compiled_model_keep_torch_compile(self):
+        model = create_components()[0]
+        accelerator = Accelerator()
+
+        compiled_model = torch.compile(model)
+
+        distributed_model = torch.nn.DataParallel(model)
+        distributed_compiled_model = torch.compile(distributed_model)
+        unwrapped_model = accelerator.unwrap_model(distributed_compiled_model, keep_torch_compile=True)
+
+        assert compiled_model._orig_mod == unwrapped_model._orig_mod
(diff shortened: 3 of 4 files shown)


SunMarc (commented):
Hi @ggoggam, this is actually intended behavior as you can see from this PR https://github.com/huggingface/accelerate/pull/1437. 

What I can suggest modify either `unwrap_model` or add a new argument to `extract_model_from_parallel`. cc @muellerzr

ggoggam:
I see. I think it would make more sense to modify `unwrap_model` to exhibit expected behavior. I will wait for other reviews before I implement the fix.

BenjaminBossan:
Thanks for the PR. Regarding the implementation, `extract_model_from_parallel` is already supposed to handle both compilation and models wrapped by DS/DDP. If it doesn't work, I'd rather see `extract_model_from_parallel` fixed than calling it repeatedly. Also, could you please add a unit test [similar to this one](https://github.com/huggingface/accelerate/blob/cb8b7c637a8588668c52bd306f9b2828f69d9585/tests/test_utils.py#L243), your original example could be adjusted for that.

ggoggam:
> Thanks for the PR. Regarding the implementation, `extract_model_from_parallel` is already supposed to handle both compilation and models wrapped by DS/DDP. If it doesn't work, I'd rather see `extract_model_from_parallel` fixed than calling it repeatedly. Also, could you please add a unit test [similar to this one](https://github.com/huggingface/accelerate/blob/cb8b7c637a8588668c52bd306f9b2828f69d9585/tests/test_utils.py#L243), your original example could be adjusted for that.

My first commit actually fixes `extract_model_from_parallel`. While I was fixing this, I noticed that the unit test you 
mentioned checks for the **inner module** of the unwrapped model, which is not expected.

https://github.com/huggingface/accelerate/blob/cb8b7c637a8588668c52bd306f9b2828f69d9585/tests/test_utils.py#L252

I would fix this to
```suggestion
assert compiled_model._orig_mod == compiled_model_unwrapped
```

BenjaminBossan:
> If I understand correctly, it should be
> 
> ```python
> assert compiled_model._orig_mod == compiled_model_unwrapped
> ```
> 
> if `extract_model_from_parallel` is supposed to handle both compilation and models by DS/DDP?

Hmm, good question, I'll leave that to the others to answer, as I'm not sure.

SunMarc:
> If I understand correctly, it should be
> ```py
> assert compiled_model._orig_mod == compiled_model_unwrapped
> ```
> if extract_model_from_parallel is supposed to handle both compilation and models by DS/DDP?

That's right. But the current behavior is that we don't unwrap the compiled model with `extract_model_from_parallel`. Right now, we just make sure that we are able to unwrap the distributed part even if the model is compiled. 

cc @muellerzr what do you prefer, modify how `extract_model_from_parallel` works or modify `unwrap` function ?

src/accelerate/utils/other.py
```diff
@@ -77,8 +77,7 @@ def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive
     """
     options = (torch.nn.parallel.DistributedDataParallel, torch.nn.DataParallel)
 
-    is_compiled = is_compiled_module(model)
```
ggoggam: To clarify: Is `is_compiled` being assigned here and then checked after DDP/DS is unwrapped? If so, this seems  potentially unnecessary or wrong.

muellerzr: Yes I agree that we should check for compilation after

ggoggam: Sorry I think I misunderstood the code. `is_compiled` is assigned here before unwrapping to keep the torch compilation. I still think this could be a problem if the model is in the form of `Distributed(Compiled(model))`. Refer to my newest commit.

ggoggam:
> Hey! Thanks for trying to tackle this. I think what I'd rather see instead is in `extract_model_from_parallel` we should add a new param arg called `keep_torch_compile`, which similar to `keep_fp32_wrapper` should default to `False` for a number of versions and then we likely should flip that to True after awhile, & modify the logic in `extract_model_from_parallel` to reflect this change.
> 
> Can you tweak this PR to so? :)

Sure thing. I added `keep_torch_compile` argument to both `Accelerator.unwrap_model` (defaults to `False`) and `extract_model_from_parallel` (defaults to `True`). Let me know if this makes sense.

I am also curious about what you think in the case of `Distributed(Compiled(model))`, since the current code only accounts for `Compiled(Distributed(model))` case. I think `Accelerator.prepare` does the latter, but I am not sure if we should still consider the former case.

src/accelerate/utils/other.py
```diff
@@ -59,7 +59,9 @@ def is_compiled_module(module):
     return isinstance(module, torch._dynamo.eval_frame.OptimizedModule)
 
 
-def extract_model_from_parallel(model, keep_fp32_wrapper: bool = True, recursive: bool = False):
+def extract_model_from_parallel(
+    model, keep_fp32_wrapper: bool = True, keep_torch_compile: bool = True, recursive: bool = False
+):
```
SunMarc, line 64: I think that you want to keep keep_torch_compile=False for a few version before setting it to True @muellerzr ?

SunMarc (approved):
Thanks for adding this and the tests !
Also, note that we have to also upstream the modification done to `extract_model_from_parallel` in transformers if needed.

src/accelerate/utils/other.py
ggoggam: I thought setting `keep_torch_compile=True` is keeping it as-is since the current version keeps compilation.

SunMarc: Oh yeah sorry I misread that, you can ignore my comment.

src/accelerate/accelerator.py
```diff
@@ -2601,7 +2601,7 @@ def pad_across_processes(self, tensor, dim=0, pad_index=0, pad_first=False):
         """
         return pad_across_processes(tensor, dim=dim, pad_index=pad_index, pad_first=pad_first)
 
-    def unwrap_model(self, model, keep_fp32_wrapper: bool = True):
+    def unwrap_model(self, model, keep_fp32_wrapper: bool = True, keep_torch_compile: bool = False):
```
SunMarc, line 2604: I think you wanted to set keep_torch_compile to True for a couple of version @muellerzr ?

muellerzr: Correct, let's default to `True` until 2.5.0, to give users ~3 months. 

Essentially we should default to `None`, and then if it gets `None` warn that the default to this will be changing

ggoggam: I've defaulted `keep_torch_compile=True` in the latest commit.

ggoggam:
@SunMarc Please let me know if there is any changes needed for this PR!

SunMarc (approved):
LGTM !

SunMarc:
cc @muellerzr for final

muellerzr (approved):
Thanks for all the work on this! Looks great to me

PR #3289, opened 2024-12-11
Fix test_nested_hook

SunMarc:
# What does this PR do ?

This PR fixes test_nested_hook as this is failing in our CI currently


muellerzr (approved):
Thanks for fixing! 🤩

PR #3298, opened 2024-12-16
[FSDP][Example] Fix FSDP example for `use_orig_params=False` 

SumanthRH:
# What does this PR do?

Fixes the FSDP example for `use_orig_params=False`. I recently started using FSDP via Accelerate and hit some roadblocks due to rough edges in documentation and examples. Some improvements in this PR: 

- `use_orig_params=False` should be the recommended route all the time (full param/ lora), since it's typically more efficient.  
- `use_orig_params=False` requires model preparation before optimizer instantiation. This is a silent failure - I noticed that loss didn't decrease in this case. I believe it is best if `.prepare` has extra validation for this and raise an Error to the user. (Might do this in a separate PR if I get time). The example prepares everything together because it assumes `use_orig_params` is True from this PR: https://github.com/huggingface/accelerate/pull/2177
- Documentation has removed this caveat with `use_orig_params`. I found a note from an older version so I brought it back. 
- `SHARD_GRAD_OP` is a bit different from Zero Stage 2, this PR adds a tiny note. 

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

 cc @muellerzr

### DIFF
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -186,6 +186,23 @@ accelerate merge-weights pytorch_model_fsdp_0/ output_path
 
 ## A few caveats to be aware of
 
+- PyTorch FSDP auto wraps sub-modules. With `use_orig_params=False`, it flattens the parameters in each sub-module and shards them in place.
+  Due to this, any optimizer created before model wrapping gets broken and occupies more memory. Further, you might also observe correctness issues during training. 
+  Hence, it is highly recommended and efficient to prepare the model before creating the optimizer. Example: 
+```diff
+  model = AutoModelForSequenceClassification.from_pretrained("bert-base-cased", return_dict=True)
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
> * `use_orig_params=False` requires model preparation before optimizer instantiation. This is a silent failure - I noticed that loss didn't decrease in this case.

Nice catch, I didn't know about that.

SumanthRH:
not stale

muellerzr:
@SumanthRH what this hints at is when we do `accelerator.prepare()` we need to:
1. Find the model (there are many cases of us doing so)
2. Prepare said model in FSDP *before* everything else

Would you be willing to tackle it this way instead? It's critical we maintain the `.prepare()` only workflow

SumanthRH:
Yes that's much better behaviour. Let me take a stab at it this week.

PR #3305, opened 2024-12-19
Fix 'AdamW' object has no attribute 'optim_bits' error when deepspeed zero2 is enabled

bursteratom:
# What does this PR do?
This PR fixes the `has no attribute 'optim_bits' error` when using AdamW or lion_32bit optimizers with deepspeed zero2 enabled. 

Also popping `"adamw_mode"` from `defaults` dict if using lion or adagrad so that `DeepSpeedCPUAdagrad` and `DeepSpeedCPULion` can be respectively initialized without unexpected argument in `"adamw_mode"`.

Fixes # (issue)
https://github.com/axolotl-ai-cloud/axolotl/issues/2191


kalomaze:
Is there anything preventing this from being merged?

bursteratom:
@kalomaze we are combining it with this pr: https://github.com/huggingface/accelerate/pull/3311 to account for backwards compatibility

BenjaminBossan:
Can this PR be closed?

bursteratom:
done :)

Issue #2100, opened 2023-10-31
It seems that FSDP's cpu_ram_efficient_loading setting doesn't work?


lllukehuang:
I run my FSDP scripts with the following configs:

```
compute_environment: LOCAL_MACHINE
debug: false
distributed_type: FSDP
downcast_bf16: 'no'
fsdp_config:
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_backward_prefetch_policy: BACKWARD_PRE
  fsdp_cpu_ram_efficient_loading: true
  fsdp_forward_prefetch: false
  fsdp_offload_params: false
  fsdp_sharding_strategy: 1
  fsdp_state_dict_type: FULL_STATE_DICT
  fsdp_sync_module_states: true
  fsdp_use_orig_params: false
machine_rank: 0
main_training_function: main
mixed_precision: 'bf16'
num_machines: 1
num_processes: 4
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

But it seems that there are still 4 processes loading the full model checkpoint and transferring data from CPU to GPU?  Does `fsdp_cpu_ram_efficient_loading: true` still work?

lllukehuang:
Seems that the `fsdp_cpu_ram_efficient_loading` config is loaded through 
https://github.com/huggingface/accelerate/blob/55747318a0f47cdfbc281e11269ba96214e4092d/src/accelerate/utils/launch.py#L106
and finally transferred into torch.distributed.run API:
https://github.com/huggingface/accelerate/blob/55747318a0f47cdfbc281e11269ba96214e4092d/src/accelerate/commands/launch.py#L654

Am I getting this right?

But I couldn't find any handling code for `fsdp_cpu_ram_efficient_loading` in Pytorch, does that mean the `fsdp_cpu_ram_efficient_loading` can't change anything?

Thanks for your reply.

pacman100:
Hello, this only works with Transformers Models as is mentioned when setting that config parameter. It is used in Transformers when doing `from_pretrained`. See these:

https://github.com/huggingface/transformers/blob/c5be38cd27bee92be73c73ba09aec8bedf841423/src/transformers/modeling_utils.py#L131


https://github.com/huggingface/transformers/blob/c5be38cd27bee92be73c73ba09aec8bedf841423/src/transformers/modeling_utils.py#L476-L483


[...]

PR #3307, opened 2024-12-19
[FSDP] Make env var and dataclass flag consistent for `cpu_ram_efficient_loading` 

SumanthRH:
# What does this PR do?

Makes environment variable and dataclass consistent for `cpu_ram_efficient_loading`. Currently, there's an inconsistency when the FSDP plugin is initialized in a standalone fashion: the `cpu_ram_efficient_loading` flag can be set to `True` but if the user doesn't use `accelerate launch` , then the environment variable `FSDP_CPU_RAM_EFFICIENT_LOADING` would not be set. This environment variable is actually used by transformers to initialize parameters efficiently in the process group (See: https://github.com/huggingface/accelerate/issues/2100#issuecomment-1822330625 for details on how this happens)

I can see a previous PR #2998 that tried to make  all env vars optional, but looks like this was left. 

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

 cc @muellerzr

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -1678,6 +1678,16 @@ class FullyShardedDataParallelPlugin:
             )
             self.sync_module_states = True
 
+        if self.cpu_ram_efficient_loading != bool(
+            str_to_bool(os.environ.get(env_prefix + "CPU_RAM_EFFICIENT_LOADING", "False"))
+        ):
+            env_var = env_prefix + "CPU_RAM_EFFICIENT_LOADING"
+            warnings.warn(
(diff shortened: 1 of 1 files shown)


SumanthRH:
not stale

SumanthRH:
not stale

S1ro1:
not stale, gonna take a look!

src/accelerate/utils/dataclasses.py
```diff
@@ -1678,6 +1678,16 @@ def __post_init__(self):
             )
             self.sync_module_states = True
 
+        if self.cpu_ram_efficient_loading != bool(
```
S1ro1, line 1681: Could you elaborate on why we need to set the variable in env? Is it not enough to have it only in the accelerator FSDP_plugin?

SunMarc (approved):
Thx !

PR #3340, opened 2025-01-13
[tests] make cuda-only test case device-agnostic

faaany:
## What does this PR do?
This PR enables 1 cuda-only case on non-cuda hardware accelerators. Below is the test result on XPU: 
```bash
================================================= short test summary info ==================================================
PASSED tests/test_accelerator.py::AcceleratorTester::test_nested_hook
======================================= 1 passed, 417 deselected, 1 warning in 5.53s =======================================
```

cc @SunMarc and @muellerzr


muellerzr (approved):
Nice! 😍

PR #3350, opened 2025-01-17
[tests] enable more bnb tests on XPU  

faaany:
## What does this PR do?
Follow up of PR #3349 .

PR #3351, opened 2025-01-17
[tests] Fix bnb cpu error

faaany:
## What does this PR do?

Follow up on PR #3350. 

The reason for the failure is that `is_bitsandbytes_multi_backend_available()` returns True on XPU, but False on CUDA. And this is because CUDA and XPU have different ways to install bitsandbytes (see [here](https://huggingface.co/docs/bitsandbytes/main/en/installation)). 

To align the behavior on CUDA and XPU, we need to add an XPU check, thus this PR. 

cc @muellerzr @SunMarc


src/accelerate/accelerator.py
```diff
@@ -1471,8 +1471,11 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                             "You can't train a model that has been loaded in 8-bit or 4-bit precision on a different device than the one "
                             "you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device()}` or `device_map={'':torch.xpu.current_device()}`"
                         )
-
-            if ("cpu" in model_devices and not is_bitsandbytes_multi_backend_available()) or "disk" in model_devices:
+            if (
+                ("cpu" in model_devices and not is_bitsandbytes_multi_backend_available())
+                or ("cpu" in model_devices and is_xpu_available())
+                or "disk" in model_devices
+            ):
```
SunMarc, line 1478: does this mean that with xpu, it doesn't work if the model in on cpu even if bnb multi-backend is installed ?

faaany: yes, cpu and xpu share the same bnb installation method, namely the multi-backend support.

SunMarc (approved):

PR #3359, opened 2025-01-21
Support more functionalities for MUSA backend

fmo-mt:
# What does this PR do?
We have already supported MUSA backend  since [v0.33.0](https://github.com/huggingface/accelerate/releases/tag/v0.33.0), and accelerate have supported more functionalities, like FSDP, of which MUSA backend doesn't included.

This PR just make MUSA backend adapted with all these features.

cc @SunMarc @muellerzr


SunMarc (approved):
Thanks !

PR #3368, opened 2025-01-25
Fix immediate StopIteration in DataLoaderShard

### LINKED ISSUE #3367: DataLoaderShard wrongly yields None instead of StopIteration when its dataloader returns StopIteration immediately
### System Info

```Shell
- `Accelerate` version: 1.3.0
- Platform: Linux-5.14.0-508.el9.x86_64-x86_64-with-glibc2.39
- `accelerate` bash location: /usr/local/bin/accelerate
- Python version: 3.12.3
- Numpy version: 2.2.2
- PyTorch version (GPU?): 2.5.1+cu124 (True)
- System RAM: 62.24 GB
- GPU type: NVIDIA GeForce RTX 2060 SUPER
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - enable_cpu_affinity: False
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

The following code snippet will print "None" and then raise an UnboundLocalError:

```python
import torch

accelerator = Accelerator()
dataset = range(10)
dataloader = torch.utils.data.DataLoader(dataset, batch_size=8, shuffle=True)

for x in accelerator.skip_first_batches(dataloader, len(dataloader)):
    print(x)

```

### Expected behavior

The iterator should immediately return StopIteration, such that nothing is printed and the script exits normally.

Aleko2286:
Fixes DataLoaderShard to also immediately return StopIteration when its dataloader immediately returns StopIteration.

Fixes  #3367 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -559,20 +559,28 @@ class DataLoaderShard(DataLoaderAdapter, DataLoaderStateMixin):
         self.set_epoch(self.iteration)
         dataloader_iter = self.base_dataloader.__iter__()
 
+        # We iterate one batch ahead to check when we are at the end
+        try:
+            next_batch = next(dataloader_iter)
+        except StopIteration:
+            self.end_of_dataloader = True
+
         batch_index = 0
-        while True:
-            try:
-                current_batch = next(dataloader_iter)
-                if self.device is not None:
+        while not self.end_of_dataloader:
+            current_batch = next_batch
+            # But we still move it to the device so it is done before `StopIteration` is reached
+            if self.device is not None:
                     current_batch = send_to_device(current_batch, self.device, non_blocking=self._non_blocking)
-                self._update_state_dict()
-                if batch_index >= self.skip_batches:
-                    yield current_batch
-                batch_index += 1
+
+            try:
+                next_batch = next(dataloader_iter)
             except StopIteration:
                 self.end_of_dataloader = True
-                self._update_state_dict()
-                break
+
+            self._update_state_dict()
+            if batch_index >= self.skip_batches:
+                yield current_batch
+            batch_index += 1
 
         self.iteration += 1
         self.end()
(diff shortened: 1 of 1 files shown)


Aleko2286:
This should be functionally identical to main now, except for two things:

1) It doesn't error on empty dataloaders
2) On main, self._update_state_dict() is called twice before yielding the final element while here it is only called once:

```python
def __iter__(self):
        if self.rng_types is not None:
            synchronize_rng_states(self.rng_types, self.synchronized_generator)
        self.begin()

        self.set_epoch(self.iteration)
        dataloader_iter = self.base_dataloader.__iter__()
        # We iterate one batch ahead to check when we are at the end
        try:
            current_batch = next(dataloader_iter)
        except StopIteration:
            yield

        batch_index = 0
        while True:
            try:
                # But we still move it to the device so it is done before `StopIteration` is reached
                if self.device is not None:
                    current_batch = send_to_device(current_batch, self.device, non_blocking=self._non_blocking)
                # First call for final element
                self._update_state_dict()
                # This gets a StopIteration for the final element
                next_batch = next(dataloader_iter)
                if batch_index >= self.skip_batches:
                    yield current_batch
                batch_index += 1
                current_batch = next_batch
            except StopIteration:
                self.end_of_dataloader = True
                # Second call
                self._update_state_dict()
                if batch_index >= self.skip_batches:
                    yield current_batch
                break

        self.iteration += 1
        self.end()

```

From the description of what self._update_state_dict() does, this shouldn't be intentional as the lookahead would be accounted for twice. If it is intentional however, the PR needs to be updated again.

Aleko2286:
Now it should be completely identical except for not crashing on empty iterators. I am still not sure if calling self._update_state_dict() a second time is intentional, but it's setting "_iterator_finished", which would never be set otherwise with the current position of self._update_state_dict() in the loop.

Issue #3379, opened 2025-02-05
estimate-memory gives ~64GB for meta-llama/Llama-3.1-70B-Instruct instead of anticipated ~140GB


dvrogozh:
With:
* accelerate: https://github.com/huggingface/accelerate/commit/f07649558051a9332a383dedfdb70918a55345de
* https://github.com/huggingface/transformers/commit/8d73a38606bc342b370afe1f42718b4828d95aaa

Accelerate's `estimate-memory` gives 64.73GB for meta-llama/Llama-3.1-70B-Instruct at fp16 precision instead of anticipated ~140GB, i.e. `num_params * num_bytes_in_dtype`. The later is basically overall recommendation for the memory capacity to accommodate inference. See for example https://huggingface.co/blog/llama31#inference-memory-requirements.

Is this a mistake by `estimate-memory` or there is some magic happening behind the scenes when Transformers load models which allows to save memory?

If `estimate-memory` indeed makes a mistake, that's worrisome since this might impact `device_map=auto` if it reuses estimate code to distribute layers across devices. At least `estimate-memory` does call regular accelerator's utils method:
https://github.com/huggingface/accelerate/blob/f07649558051a9332a383dedfdb70918a55345de/src/accelerate/commands/estimate.py#L21-L22

```
# accelerate estimate-memory meta-llama/Llama-3.1-70B-Instruct
Loading pretrained config for `meta-llama/Llama-3.1-70B-Instruct` from `transformers`...
┌────────────────────────────────────────────────────────────────┐
│  Memory Usage for loading `meta-llama/Llama-3.1-70B-Instruct`  │
├───────┬─────────────┬──────────┬───────────────────────────────┤
│ dtype │Largest Layer│Total Size│      Training using Adam      │
├───────┼─────────────┼──────────┼───────────────────────────────┤
│float32│   1.96 GB   │129.46 GB │           517.84 GB           │
│float16│  1002.0 MB  │ 64.73 GB │           258.92 GB           │
│  int8 │   501.0 MB  │ 32.36 GB │              N/A              │
│  int4 │   250.5 MB  │ 16.18 GB │              N/A              │
└───────┴─────────────┴──────────┴───────────────────────────────┘
```

CC: @muellerzr @ArthurZucker @SunMarc

SunMarc:
Indeed, I fixed it in the PR above. The issue was that we didn't correctly initialize the model in FP32 (used the torch_dtype from the config that could be different)

[...]

PR #3383, opened 2025-02-06
fix torch_dtype in estimate memory

SunMarc:
# What does this PR do ?

Fixes https://github.com/huggingface/accelerate/issues/3379
This PR fixes the estimate memory command as it gave wrong results when `torch_dtype` is specified in the config we are loading.


BenjaminBossan (approved):
LGTM, thanks for the quick fix. Maybe let's add a test or at least a comment on why passing the torch dtype is needed.

SunMarc:
Added a comment !

BenjaminBossan:
Linter is complaining, probably the line is too long.

Issue #3360, opened 2025-01-23
FP8 training can't works with deepspeed


XiaobingSuper:
### System Info

```Shell
- `Accelerate` version: 1.3.0
- Platform: Linux-5.4.250-4-velinux1u1-amd64-x86_64-with-glibc2.35
- `accelerate` bash location: /usr/local/bin/accelerate
- Python version: 3.10.12
- Numpy version: 1.24.4
- PyTorch version (GPU?): 2.5.1+cu124 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- PyTorch MUSA available: False
- System RAM: 1928.86 GB
- GPU type: NVIDIA H20
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

Create deepspeed configure json:
```json
{
    "train_batch_size": 16,
    "train_micro_batch_size_per_gpu": 16,
    "gradient_accumulation_steps": 1,
    "zero_optimization": {
        "stage": 3,
        "stage3_gather_16bit_weights_on_model_save": false
    },
    "gradient_clipping": 1.0,
    "bf16": {"enabled": true},
    "fp16": {"enabled": false},
    "zero_allow_untested_optimizer": true
}
````
and accelerate configure yaml:
```yaml
distributed_type: DEEPSPEED
deepspeed_config:
  deepspeed_config_file: "config.json"
  zero3_init_flag: true
num_processes: 1
```

Then run the following script:
```python
from accelerate import Accelerator
from accelerate.utils import has_transformer_engine_layers, FP8RecipeKwargs
from fp8_utils import  get_training_utilities

MODEL_NAME = "bert-base-cased"

FP8_RECIPE_KWARGS = {"fp8_format": "HYBRID", "amax_history_len": 32, "amax_compute_algo": "max"}
kwargs_handlers = [FP8RecipeKwargs(backend="TE", **FP8_RECIPE_KWARGS)]

accelerator = Accelerator(
    mixed_precision="fp8", kwargs_handlers=kwargs_handlers
)

model, optimizer, train_dataloader, eval_dataloader, lr_scheduler = get_training_utilities(
[...]

PR #3385, opened 2025-02-07
allow to set mxied_precision in accelerate config when using a given deepspeed config

XiaobingSuper:
# What does this PR do?

Fixes https://github.com/huggingface/accelerate/issues/3360#issuecomment-2642180644.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -1335,7 +1335,6 @@ class DeepSpeedPlugin:
             "ACCELERATE_DEEPSPEED_OFFLOAD_PARAM_NVME_PATH",
             "ACCELERATE_DEEPSPEED_OFFLOAD_OPTIMIZER_NVME_PATH",
             "ACCELERATE_DEEPSPEED_ZERO3_SAVE_16BIT_MODEL",
-            "ACCELERATE_MIXED_PRECISION",
         ]
         env_variable_names_to_ignore = [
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/dataclasses.py
```diff
@@ -1335,7 +1335,6 @@ def _deepspeed_config_checks(self):
             "ACCELERATE_DEEPSPEED_OFFLOAD_PARAM_NVME_PATH",
             "ACCELERATE_DEEPSPEED_OFFLOAD_OPTIMIZER_NVME_PATH",
             "ACCELERATE_DEEPSPEED_ZERO3_SAVE_16BIT_MODEL",
-            "ACCELERATE_MIXED_PRECISION",
```
XiaobingSuper, line 1338: The ds config doesn't have such an attribute, and it only works when calling in 
https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L351

XiaobingSuper:
@muellerzr, please help review it. Thanks!

muellerzr (commented):
I disagree with this change, we rely on the deepspeed config specifically for things like the FP8 setup as well if applicable, so I'm not a fan of merging this.

ronald-d-rogers:
@muellerzr I think the issue might be that there is no way to set DeepSpeed up for fp8 besides wrapping the modules in TransformerEngine modules, and unless you set `mixed_precision` to `fp8` that won't happen.

I think I'm running into the same issue.

PR #3387, opened 2025-02-08
[examples] upgrade code for seed setting  

faaany:
## What does this PR do?
As the title suggests, this PR updates the code for seed setting in `cv_example.py`. The current code only applies for cuda device.

cc @SunMarc and @muellerzr

### DIFF
--- a/examples/cv_example.py
+++ b/examples/cv_example.py
@@ -22,9 +22,9 @@ from timm import create_model
 from torch.optim.lr_scheduler import OneCycleLR
 from torch.utils.data import DataLoader, Dataset
 from torchvision.transforms import Compose, RandomResizedCrop, Resize, ToTensor
-from transformers import set_seed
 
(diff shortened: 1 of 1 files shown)


examples/cv_example.py
```diff
@@ -22,6 +22,7 @@
 from torch.optim.lr_scheduler import OneCycleLR
 from torch.utils.data import DataLoader, Dataset
 from torchvision.transforms import Compose, RandomResizedCrop, Resize, ToTensor
+from transformers import set_seed
```
SunMarc, line 25: You can use set_seed from accelerate.utils instead !

SunMarc (approved):
Thanks ! Just a nit

examples/cv_example.py
faaany: Sure, updated.

PR #3394, opened 2025-02-11
Initial FSDP2 support

S1ro1:
# FSDP2

## What does this add?

This PR adds support for enabling `FSDP2` in  Accelerate as another way to parallelize your training 🚀 

## Why is it needed?

FSDP2 offers a lot of advantages over current `FSDP1` implementation, mostly not user facing, but this enables further development of features in Accelerate such as combining TP + FSDP.
For users, this enables small improvements in memory usage

### User-facing:

Allows users to specify `--fsdp-version=2` to use FSDP2 instead of current FSDP1, this also changes some of the configuration options for `FullyShardedDataParallelPlugin`. Old configurations are still supported as before.

## Basic Usage Example(s):

```python
from accelerate import FullyShardedDataParallelPlugin, Accelerator

fsdp_plugin = FullyShardedDataParallelPlugin(
    fsdp_version=2
    # other options...
)
accelerator = Accelerator(fsdp_plugin=fsdp_plugin)
```

Using FSDP2 as a default is reccomended, with swapping to FSDP1 if something isn't supported.


src/accelerate/commands/to_fsdp2.py
```diff
@@ -0,0 +1,180 @@
...
+        "HYBRID_SHARD_ZERO2": False,
+    },
+    #  TODO: do we need to handle mp/offload policy
+}
+
+if is_rich_available():
+    from rich.logging import RichHandler
+
+    FORMAT = "%(message)s"
+    logging.basicConfig(format=FORMAT, datefmt="[%X]", handlers=[RichHandler()])
```
muellerzr, line 64: We've basically gotten rid of rich so let's not include this

src/accelerate/utils/dataclasses.py
```diff
@@ -1594,26 +1611,53 @@ def __post_init__(self):
...
+                # We need to remap based on custom enum values for user readability
+                if self.reshard_after_forward.upper() in FSDP_SHARDING_STRATEGY:
+                    self.reshard_after_forward = FSDP_SHARDING_STRATEGY.index(self.reshard_after_forward.upper()) + 1
+                if isinstance(self.reshard_after_forward, int) or self.reshard_after_forward.isdigit():
+                    self.reshard_after_forward = ShardingStrategy(int(self.reshard_after_forward))
+                else:
+                    self.reshard_after_forward = ShardingStrategy[self.reshard_after_forward.upper()]
+        if self.fsdp_version != 2 and isinstance(self.reshard_after_forward, bool):
+            raise ValueError(
+                "reshard_after_forward set to bool. This is not supported in FSDP1, please set to a `str` or an instance of `torch.distributed.fsdp.fully_sharded_data_parallel.ShardingStrategy`"
```
muellerzr, line 1641:
```suggestion
                f"reshard_after_forward set to {self.reshard_after_forward}. This is not supported with FSDP1, please set to a `str` or an instance of `torch.distributed.fsdp.fully_sharded_data_parallel.ShardingStrategy`"
```

src/accelerate/utils/dataclasses.py
```diff
@@ -1594,26 +1611,53 @@ def __post_init__(self):
...
+            else:
+                self.cpu_offload = CPUOffload(offload_params=self.cpu_offload)
+        if self.fsdp_version != 2 and isinstance(self.cpu_offload, CPUOffloadPolicy):
+            raise ValueError(
+                "cpu_offload set to `torch.distributed.fsdp.CPUOffloadPolicy`. This is not supported in FSDP1, please set to a `bool` or an instance of `torch.distributed.fsdp.CPUOffload`"
+            )
+        if self.fsdp_version == 2 and not isinstance(self.cpu_offload, CPUOffloadPolicy):
+            raise ValueError(
+                "cpu_offload set to `bool` or `torch.distributed.fsdp.CPUOffload`. This is not supported in FSDP2, please set to an instance of `torch.distributed.fsdp.CPUOffloadPolicy`"
+            )
```
muellerzr, line 1660: I feel like we can simplify this

```suggestion
    if isinstance(self.cpu_offload, CPUOffloadPolicy):
        err = "`cpu_offload` set to `torch.distributed.fsdp.CPUOffloadPolicy`."
        if self.fsdp_version != 2:
            raise ValueError(f"{err} This is not supported in FSDP1, please set to a `bool` or an instance of `torch.distributed.fsdp.CPUOffload`")
        else:
            raise ValueError(f"{err} This is not supported in FSDP2, please set to an instance of `torch.distributed.fsdp.CPUOffloadPolicy`")
```

src/accelerate/utils/dataclasses.py
```diff
@@ -1594,26 +1611,53 @@ def __post_init__(self):
...
+                "reshard_after_forward set to bool. This is not supported in FSDP1, please set to a `str` or an instance of `torch.distributed.fsdp.fully_sharded_data_parallel.ShardingStrategy`"
+            )
 
         if self.cpu_offload is None:
             self.cpu_offload = str_to_bool(os.environ.get(env_prefix + "OFFLOAD_PARAMS", "False")) == 1
         if isinstance(self.cpu_offload, bool):
-            self.cpu_offload = CPUOffload(offload_params=self.cpu_offload)
+            if self.fsdp_version == 2:
+                if not self.cpu_offload:
+                    warnings.warn("Offload_params is set to False, however FSDP2 always offloads parameters and runs optimizer step on CPU. This will be overridden to True.")
```
muellerzr, line 1649: Note that in general, users will ignore `warnings` (and will get annoyed it bloats the logs). So we should instead use `logger.warn` if we don't want to explicitly raise an error about this.

src/accelerate/utils/dataclasses.py
```diff
@@ -1814,6 +1882,23 @@ def set_mixed_precision(self, mixed_precision, buffer_autocast=False, override=F
                 )
             self.mixed_precision_policy = MixedPrecision(**self.mixed_precision_policy)
 
+    def validate_mixed_precision_policy(self):
+        """
+        Validates the mixed precision policy, abstracted away to not bring in the imports if not needed.
+        """
+        from torch.distributed.fsdp import MixedPrecision, MixedPrecisionPolicy
```
muellerzr, line 1889: Will this import lead to issues on old pytorch versions?

kmehant:
@muellerzr @S1ro1 

I see potential collaboration on this thread with my PRs
1. https://github.com/huggingface/accelerate/pull/3395
2. https://github.com/huggingface/accelerate/pull/3231

With changes on this PR I understand that the design is to convert FSDP1 args to FSDP2 args and put to use. With my initial discussions started at https://github.com/huggingface/accelerate/pull/3231, In the PR https://github.com/huggingface/accelerate/pull/3395 I propose to have FSDP2 as a separate distributed type training protocol alongsides FSDP1. Also, to support more complex combinations of parallelisms I as well propose a design as described in the PR's description.

Looking forward.

S1ro1:
@kmehant I think collaborating on support for HSDP later is the best option, we have discussed the user-facing design of this internally and most of the work is ~done, however I'd be open to collaborate on HSDP later.

src/accelerate/utils/dataclasses.py
```diff
@@ -1623,27 +1651,63 @@ class FullyShardedDataParallelPlugin:
     def __post_init__(self):
         from torch.distributed.fsdp import (
             BackwardPrefetch,
-            CPUOffload,
             ShardingStrategy,
         )
 
+        _fsdp2_warnings = {}
```
muellerzr, line 1657: Note this needs to be a `set()` for what you want
```suggestion
        _fsdp2_warnings = set()
```

docs/source/concept_guides/fsdp1_vs_fsdp2.md
```diff
@@ -0,0 +1,67 @@
...
+# FSDP1 vs FSDP2
+
+This guide explains the key differences between `FSDP1` and `FSDP2` and helps you migrate your existing code to use `FSDP2` with minimal changes.
+
+## What is FSDP2?
+
+`FSDP2` is a new and improved version of PyTorch's fully-sharded data parallel training API. Compared to `FSDP1`, it offers:
+- Simpler internal implementation
+- Flexible parameter freezing
+- Support for `fp8` parameters
```
muellerzr, line 25: We might want to be more specific here? Because I might think "well, FSDP with TransformerEngine or `torchao` was already a thing without it"

src/accelerate/utils/dataclasses.py
```diff
@@ -1856,7 +1950,61 @@ def set_mixed_precision(self, mixed_precision, buffer_autocast=False, override=F
                     f"Must be a `dict` with keys `param_dtype`, `reduce_dtype`, and `buffer_dtype`. "
                     f"Values must be one of {list(mixed_precision_mapping.values())}"
                 )
-            self.mixed_precision_policy = MixedPrecision(**self.mixed_precision_policy)
+            if self.fsdp_version == 1:
+                self.mixed_precision_policy = MixedPrecision(**self.mixed_precision_policy)
+            elif self.fsdp_version == 2:
+                self.mixed_precision_policy = MixedPrecisionPolicy(**self.mixed_precision_policy)
```
muellerzr, line 1956: This then also just becomes reduced the the old code

src/accelerate/utils/environment.py
```diff
@@ -56,7 +56,7 @@ def convert_dict_to_env_variables(current_env: dict):
     return valid_env_items
 
 
-def str_to_bool(value) -> int:
+def str_to_bool(value, to_bool: bool = False) -> Union[int, bool]:
```
muellerzr, line 59: What's the need for the change here?

S1ro1: This is required due to a funny *feature* `FSDP2` has. It supports `int` for `reshard_after_forward` as shown [here](https://pytorch.org/docs/stable/distributed.fsdp.fully_shard.html). If `int` is specified then it has a different meaning than if `bool` is specified. Therefore when we convert `str` to `bool` and it returns `1` it has a different meaning than if it returns `true`.

src/accelerate/utils/constants.py
```diff
@@ -40,6 +40,7 @@
 FSDP_PYTORCH_VERSION = (
     "2.1.0.a0+32f93b1"  # Technically should be 2.1.0, but MS-AMP uses this specific prerelease in their Docker image.
 )
+FSDP2_PYTORCH_VERSION = "2.6.0"  # TODO(s1ro): check if this is 100% correct
```
SalmanMohammadi, line 43: AFAIK it was torch 2.5.1

src/accelerate/accelerator.py
```diff
@@ -1374,10 +1380,30 @@ def prepare(self, *args, device_placement=None):
...
+        # this is a workaround to make memory footprint match if `Optimizer` is created before preparing the model
+        if fsdp2_should_fix_optimizer:
+            for obj in args:
+                if isinstance(obj, torch.optim.Optimizer):
+                    for param_group in obj.param_groups:
+                        for i, p in enumerate(param_group["params"]):
+                            # We drop a reference to the original param here, so that _move_states_to_device triggers a reallocation
+                            # We reassign the data_ptr to the original param, so that we preserve the mapping to the new ones
+                            param_group["params"][i] = torch.empty_like(p)
+                            param_group["params"][i].data_ptr = p.data_ptr()
```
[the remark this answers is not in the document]
cyr0930: isn't it?
```param_group["params"][i].data_ptr = p.data_ptr```

File: src/accelerate/utils/launch.py [hunk 12]
```diff
@@ -282,7 +282,15 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> Dict[str, str]:
         if args.fsdp_cpu_ram_efficient_loading and not args.fsdp_sync_module_states:
```
cyr0930, line 282: It is okay with getting args from yaml file, but if we get args from cli, this doesn't work as they are string, and result of ```args.fsdp_cpu_ram_efficient_loading``` will always be true. (because it is "true" or "false" not True or False)

src/accelerate/accelerator.py
```diff
@@ -1374,10 +1380,30 @@ def prepare(self, *args, device_placement=None):
...
+        # however that goes against `Accelerate's` design of `bring your own`
+        # this is a workaround to make memory footprint match if `Optimizer` is created before preparing the model
+        if fsdp2_should_fix_optimizer:
+            for obj in args:
+                if isinstance(obj, torch.optim.Optimizer):
+                    for param_group in obj.param_groups:
+                        for i, p in enumerate(param_group["params"]):
+                            # We drop a reference to the original param here, so that _move_states_to_device triggers a reallocation
+                            # We reassign the data_ptr to the original param, so that we preserve the mapping to the new ones
+                            param_group["params"][i] = torch.empty_like(p)
```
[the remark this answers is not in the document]
S1ro1: Yes, you are right, current changes were broken due to a typo. Fix is on the way.

src/accelerate/utils/launch.py
S1ro1: Will take a look at this in a different PR as this wasn't changed in this PR, could you open an issue with a small reproduction if possible?

src/accelerate/accelerator.py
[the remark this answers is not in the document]
S1ro1: No, torch has `data_ptr()` as a getter, `data_ptr` as an attribute to set.

muellerzr (commented):
Nice job on the benchmarks, some initial notes on the README

benchmarks/fsdp2/main.py
```diff
@@ -0,0 +1,123 @@
...
+def train(
+    model: torch.nn.Module,
+    optimizer: torch.optim.Optimizer,
+    train_dataloader: torch.utils.data.DataLoader,
+    accelerator: Accelerator,
+) -> torch.Tensor:
+    losses = []
+    for step, batch in enumerate(train_dataloader):
+        optimizer.zero_grad()
+        batch = {k: v.to(accelerator.device) for k, v in batch.items()}
```
muellerzr, line 42: Since all baselines should wrap the `dataloader` via accelerate, we don't need to do this :)

(At some point we need to change all the docs since this is now the default and has been for a long while)
```suggestion
```

benchmarks/fsdp2/main.py
```diff
@@ -0,0 +1,123 @@
...
+
+    loss = train(model, optimizer, dataloader, accelerator)
+
+    memory_tracker.stop()
+    if accelerator.is_main_process:
+        print(f"Results for {run_name} (rank 0):")
+        print(f"\tLoss: {loss[-1].item()}")
+        print(f"Peak Allocated Memory: {float(memory_tracker.peak_allocated_memory):.2f} MB")
+        print(f"Peak Reserved Memory: {float(memory_tracker.peak_reserved_memory):.2f} MB")
+        print("----------------------------------")
```
muellerzr, line 64:
```suggestion
    msg = f"Results for {run_name} (rank 0):\tLoss: {loss[-1].item()}\nPeak Allocated Memory: {float(memory_tracker.peak_allocated_memory):.2f} MB\nPeak Reserved Memory: {float(memory_tracker.peak_reserved_memory):.2f} MB\n{'-'*34}"
    accelerator.print(msg)
```

benchmarks/fsdp2/measure_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        log_interval: float = 0.01,
+    ):
+        """Class for tracking gpu and cpu memory usage of the process.
+
+        Args:
+            device (torch.device): Cuda device to monitor.
+            output_directory (str): Directory to save the memory usage data to, will be created if it doesn't exist.
+            run_name (str): Name of the run, will be used to name the output files.
+            save_memory_snapshot (bool): Whether to also save `torch.cuda.memory._dump_snapshot` to the output directory.
+            log_interval (float, optional): Interval in seconds between memory measurements. Defaults to 0.01.
```
muellerzr, line 43:
```suggestion
            device (`torch.device`): 
                CUDA device to monitor.
            output_directory (`str`): 
                Directory to save the memory usage data to, will be created if it doesn't exist.
            run_name (`str`): 
                Name of the run, will be used to name the output files.
            save_memory_snapshot (`bool`): 
                Whether to also save `torch.cuda.memory._dump_snapshot` to the output directory.
            log_interval (`float`, *optional*): 
                Interval in seconds between memory measurements. Defaults to `0.01`.
```

benchmarks/fsdp2/measure_utils.py
```diff
@@ -0,0 +1,131 @@
...
+            virtual_memory = self._process.memory_info().rss / (1024 * 1024)
+
+            self.allocated_memory.append(allocated)
+            self.reserved_memory.append(reserved)
+            self.virtual_memory.append(virtual_memory)
+            self.timestamps.append(time.time() - self.start_time)
+
+            time.sleep(self.log_interval)
+
+    def _pad_tensor_to_power_of_2(self, tensor):
```
muellerzr, line 78: Small docstring of why we're doing this would be good

benchmarks/fsdp2/measure_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        self.timestamps = []
+        self.allocated_memory = []
+        self.reserved_memory = []
+        self.virtual_memory = []
+
+        self.start_time = None
+        self.running = False
+
+        self._thread = None
+        self._accelerator = Accelerator()
```
muellerzr, line 59: Rather than accelerator, let's use the `PartialState()` class instead since it's made for situations like these :)

muellerzr (commented):
Rest of the benchmark nits, nice work :)

src/accelerate/accelerator.py
```diff
@@ -546,7 +554,12 @@ def __init__(
             ) or is_torch_xla_available(check_is_tpu=True):
                 raise ValueError(f"fp16 mixed precision requires a GPU (not {self.device.type!r}).")
             kwargs = self.scaler_handler.to_kwargs() if self.scaler_handler is not None else {}
-            self.scaler = get_grad_scaler(self.distributed_type, **kwargs)
+
+            # FSDP2 doesn't use ShardedGradScaler, don't want to modify `get_grad_scaler`, rather create a simple utility
+            if self.distributed_type == DistributedType.FSDP and self.state.fsdp_plugin.fsdp_version == 2:
```
muellerzr, line 559: Let's add an attribute to the accelerator and `AcceleratorState` of something like the following:

```python
# Inside AcceleratorState
@property 
def is_fsdp2(self) -> bool:
    return self.distributed_type == DistributedType.FSDP and self.fsdp_plugin.fsdp_version == 2
```
```python
# Inside accelerator
@property 
def is_fsdp2(self) -> bool:
    return self.state.is_fsdp2
```

Should simplify a ton of stuff

src/accelerate/accelerator.py
```diff
@@ -1372,12 +1385,47 @@ def prepare(self, *args, device_placement=None):
...
+            optimizer_count = 0
+            for obj in args:
+                if isinstance(obj, torch.nn.Module):
+                    model_count += 1
+                elif isinstance(obj, torch.optim.Optimizer):
+                    optimizer_count += 1
+
+            # This needs to be written as such, so that passing other objects other than models/optimizers doesn't raise an error
+            if (model_count < 1 and optimizer_count > 0) or (model_count > 0 and optimizer_count < 1):
+                raise ValueError(  #
```
muellerzr, line 1399:
```suggestion
                raise ValueError(
```

src/accelerate/accelerator.py
```diff
@@ -1372,12 +1385,47 @@ def prepare(self, *args, device_placement=None):
...
+                    "With FSDP2, you need to pass model and optimizer to a single `Accelerator.prepare()` call."
+                    "This is due to the fact that we apply a fix to the optimizer, which replaces its parameters after applying FSDP2."
```
muellerzr, line 1401:
```suggestion
                    "When using FSDP2, a model and optimizer must be passed together to `Accelerator.prepare()`"
                    " as the optimizer needs to have its parameters modified after the model is converted."
```

src/accelerate/accelerator.py
```diff
@@ -1372,12 +1385,47 @@ def prepare(self, *args, device_placement=None):
...
+                )
+
         # If we're dealing with device placement, this deals with that by...
         tpu_should_fix_optimizer = self.device_placement and self.distributed_type == DistributedType.XLA
+        fsdp2_should_fix_optimizer = (
+            self.distributed_type == DistributedType.FSDP and self.state.fsdp_plugin.fsdp_version == 2
+        )
+        should_fix_optimizer = tpu_should_fix_optimizer or fsdp2_should_fix_optimizer
```
muellerzr, line 1409: Can be simplified with the stated change

src/accelerate/utils/fsdp_utils.py
```diff
@@ -71,16 +75,22 @@ def save_fsdp_model(fsdp_plugin, accelerator, model, output_dir, model_index=0,
...
-    ):
+    ctx = (
+        FSDP.state_dict_type(
+            model, fsdp_plugin.state_dict_type, fsdp_plugin.state_dict_config, fsdp_plugin.optim_state_dict_config
+        )
+        if fsdp_plugin.fsdp_version == 1
+        else nullcontext()
+    )
+
+    with ctx:
```
muellerzr, line 93: Nice!

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +424,149 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+            full_tensor = torch.empty(sharded_param.size(), device="cuda", dtype=sharded_param.dtype)
+            mesh = sharded_param.device_mesh
+            dist.broadcast(full_tensor, src=0, group=mesh.get_group())
+            sharded_tensor = distribute_tensor(full_tensor, mesh, sharded_param.placements)
+            sharded_sd[param_name] = sharded_tensor
+
+    model.load_state_dict(sharded_sd)
+
+
+def fsdp2_switch_optimizer_parameters(optimizer: torch.optim.Optimizer, mapping: dict):
```
muellerzr, line 452: Let's document

muellerzr: And also add to the fsdp reference docs

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +424,149 @@ def param_init_fn_tied_param(module: torch.nn.Module):
         return module
 
     return param_init_fn_tied_param
+
+
+def fsdp2_load_full_state_dict(accelerator, model: torch.nn.Module, full_sd: dict):
```
muellerzr, line 429: Similarly document and add to the fsdp reference docs

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +424,149 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+            param_group["params"] = [mapping[p.data_ptr] for p in param_group["params"]]
+    except KeyError:
+        # This shouldn't ever happen, but we want to fail here else training wouldn't be numerically correct
+        # This basically means that we're missing a mapping from the original parameter to the sharded parameter
+        raise KeyError(
+            "A parameter in the optimizer couldn't be switched to its sharded version. This breaks the training. Please raise an issue on GitHub."
+        )
+
+
+def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
```
muellerzr, line 464: Similarly

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +424,149 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+        # We upcast the model according to `deepspeed`'s implementation
+        # More info about this can be found in `accelerator.py:prepare_model`s FSDP1 section
+        model = model.to(torch.float32)
+        if accelerator.is_main_process:
+            # TODO(siro1): Add a warning for each parameter that was upcasted
+            warnings.warn("FSDP upcast of low precision parameters may affect the precision of model checkpoints.")
+    return model
+
+
+def fsdp2_prepare_auto_wrap_policy(fsdp2_plugin, auto_wrap_policy_type: str, model: torch.nn.Module):
```
muellerzr, line 538: Similarly

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +424,149 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+
+        def policy(module: torch.nn.Module) -> bool:
+            return module.numel() > fsdp2_plugin.min_num_params
+    else:
+        return None
+
+    return policy
+
+
+def get_fsdp2_grad_scaler(**kwargs):
```
muellerzr, line 569: (This one is fine, maybe just a comment explaining why we need to do this/where it's used)

src/accelerate/utils/other.py
```diff
@@ -350,3 +350,16 @@ def _getattr(obj, attr):
         return getattr(obj, attr)
 
     return reduce(_getattr, [obj] + attr.split("."))
+
+
+def get_children_bottom_up(model: torch.nn.Module) -> list[torch.nn.Module]:
```
muellerzr, line 355: Add to the reference docs :)

muellerzr (approved):
Nicely done @S1ro1, this is great. Bar moving the images to the docs repo and the nits here, nice work ✅

muellerzr (commented):
Offloaded the images to hf hub

docs/source/concept_guides/fsdp1_vs_fsdp2.md
```diff
@@ -0,0 +1,105 @@
...
+This is why `FSDP2` was introduced. It doesn't use `FlatParameter`, instead it uses `DTensor` which is short for "Distributed Tensor". Each `DTensor` basically represents a vanilla `torch.Tensor` that has been sharded across ranks. It contains metadata about the original `torch.Tensor` and how it's sharded, what is the [placement type](https://pytorch.org/docs/stable/distributed.tensor.html#module-torch.distributed.tensor.placement_types) and so on. This is why it's called `per-parameter sharding`. The following figure shows the difference:
+
+<div align="center">
+  <img src="../imgs/fsdp1_vs_fsdp2/Fsdp2.png" alt="FSDP2">
+</div>
+
+Each Parameter of the original `Layer` is sharded across the 0th dimension, and split between 2 GPUs. Now, each `Linear` layer is a separate `DTensor` and storing metadata per-parameter is possible and straightforward.
+
+
+> [!NOTE]
```
stevhliu, line 60: The doc-builder only supports `TIP` or `WARNING`

```suggestion
> [!TIP]
```

stevhliu (commented):
Good breakdown of their differences, and the table at the end really helps with the comparison as well! 👏 My only feedback is:

The main feature of FSDP2 is `DTensor`, but I didn't really realize it's significance until the following section that illustrated the limitation of FSDP1. Now that first sentence "Simpler internal implementation, where each `Parameter` is a separate `DTensor`" is more impactful to me.

I wonder if it'd be better to start with that section so users understand what `DTensor` is and how it enables FSDP2 to overcome the limitations of FSDP1. Then you can discuss the other benefits of FSDP2 in the following section.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -233,15 +274,28 @@ def load_fsdp_optimizer(fsdp_plugin, accelerator, optimizer, model, input_dir, o
...
             logger.info(f"Optimizer loaded from {ckpt_dir}")
-        flattened_osd = FSDP.optim_state_dict_to_load(model=model, optim=optimizer, optim_state_dict=optim_state)
-        optimizer.load_state_dict(flattened_osd)
+        if fsdp_plugin.fsdp_version == 1:
+            flattened_osd = FSDP.optim_state_dict_to_load(model=model, optim=optimizer, optim_state_dict=optim_state)
+            optimizer.load_state_dict(flattened_osd)
+        else:
+            # we can't do `set_state_dict` here because it does a step of the optimizer which breaks with grad-scaler
+            # TODO(siro1): investigate
+            optimizer.load_state_dict(optim_state)
```
SunMarc, line 298: is there an issue with load_state_dict compared to set_state_dict ?

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +425,203 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+            ),
+            auto_wrap_policy=fsdp2_plugin.auto_wrap_policy,
+        )
+
+    fsdp2_kwargs = {
+        "reshard_after_forward": fsdp2_plugin.reshard_after_forward,
+        "offload_policy": fsdp2_plugin.cpu_offload,
+        # `fully_shard` doesn't accept `None` in case of `MixedPrecisionPolicy`
+        "mp_policy": fsdp2_plugin.mixed_precision_policy or MixedPrecisionPolicy(),
+    }
```
SunMarc, line 546: If we always need to set mixed_precision_policy for fdspv2, can we set it in the post_init instead of here

src/accelerate/utils/fsdp_utils.py
```diff
@@ -371,3 +425,203 @@ def param_init_fn_tied_param(module: torch.nn.Module):
...
+        if accelerator.is_main_process:
+            # TODO(siro1): Add a warning for each parameter that was upcasted
+            warnings.warn("FSDP upcast of low precision parameters may affect the precision of model checkpoints.")
+    return model
+
+
+def fsdp2_prepare_auto_wrap_policy(
+    fsdp2_plugin, auto_wrap_policy_type: str, model: torch.nn.Module
+) -> Callable[[torch.nn.Module], bool]:
+    """Prepares the auto wrap policy based on its type, done to mimic the behaviour of FSDP1 auto wrap policy.
```
SunMarc, line 575: if this is mostly copied from the fsdp code, put the comment above the function it will be better

tests/fsdp/test_fsdp.py
```diff
@@ -15,6 +15,7 @@
```
SunMarc, line 15: No strong opinion but maybe create a separate file for FSDPv2 tests. This will potentially make things easier in the future. WDTY @muellerzr ?  Also, does fdsp is still being maintained by pytorch or they totally switched to fsdpv2 ?

docs/source/concept_guides/fsdp1_vs_fsdp2.md
```diff
@@ -0,0 +1,106 @@
...
+    fsdp_version=2
+    # other options...
+)
+accelerator = Accelerator(fsdp_plugin=fsdp_plugin)
+```
+
+### If using YAML config:
+Use our conversion tool:
+```bash
+accelerate to-fsdp2 --config_file config.yaml --output_file new_config.yaml
```
SunMarc, line 103: Good idea but maintaining that might be a bit of a pain. You already have added the deprecation msg for some args + not sure what will happen with this tool you will fully deprecate `fsdp_sharding_strategy`. wdyt @muellerzr ?

benchmarks/fsdp2/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Overview
+
+This benchmark consists of two parts:
+- `main.py` is the main script that runs the benchmark
+- `visualize.py` is the script that visualizes the results
+
+## Motivation
+
+We want to showcase that 🤗 `accelerate`'s integration of `FSDP2` is on par raw PyTorch, and highlight a "broken" part in PyTorch that creating an optimizer before applying `FSDP2` **doesn't result in a working training loop**. (more on this later)
```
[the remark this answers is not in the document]
S1ro1: No, there isn't any. We found this via testing. In PyTorch docs they only showcase creating optimizer *after* applying `fully_shard`. No mention of doing it before.

benchmarks/fsdp2/utils.py
```diff
@@ -0,0 +1,290 @@
...
+    config = AutoConfig.from_pretrained(model_name, trust_remote_code=True)
+    model = AutoModelForCausalLM.from_config(config)
+    return model
+
+
+def get_tokenizer(model_name: str):
+    tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
+    tokenizer.pad_token = tokenizer.eos_token
+    return tokenizer
+
```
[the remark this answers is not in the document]
S1ro1: To be sure if users switch to their model so that the benchmark doesn't break. It doesn't do anything for the default case, IMO no harm in keeping that in then.

docs/source/concept_guides/fsdp1_vs_fsdp2.md
```diff
@@ -0,0 +1,106 @@
...
+- With `DTensor`, `FSDP2` supports mixing `fp8` and other parameter types in the same model out of the box
+- Faster and simpler checkpointing without extra communication across ranks using `SHARDED_STATE_DICT` and [`torch.distributed.checkpoint`](https://pytorch.org/docs/stable/distributed.checkpoint.html), this way, each rank only saves its own shard and corresponding metadata
+- For loading, it uses a `state_dict` of the sharded model to directly load the sharded parameters
+- Support for asynchronous checkpointing, where parameters are first copied to CPU memory, after this, main thread continues training while another thread stores the parameters on disk
+- Memory efficiency and deterministic memory usage, `FSDP2` doesn't use `recordStream` anymore and uses stream-to-stream synchronization (for more technical details see [this forum post](https://dev-discuss.pytorch.org/t/fsdp-cudacachingallocator-an-outsider-newb-perspective/1486) and [this issue](https://github.com/pytorch/pytorch/issues/114299))
+- In the future, optimizations of the communication patterns via `torch.compile` are planned, further improving the performance and memory efficiency
+
```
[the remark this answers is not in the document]
S1ro1: Not 100% sure as of now, should break with the current optimizers in bnb, but is a high priority TODO.

src/accelerate/accelerator.py
```diff
@@ -1552,6 +1609,13 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                         and _tp_plan attribute to model class."
                     )
                 model.tensor_parallel(self.state.torch_tp_plugin.torch_device_mesh["tp"])
+            elif self.is_fsdp2:
+                model = fsdp2_prepare_model(self, model)
+
+                if len(self._models) > 1 and (self._models[-2] is self._models[-1]):
+                    del self._models[-2]
+                self._models[-1] = model
```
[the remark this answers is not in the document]
S1ro1: A copy of what we do in FSDP1, for not totally clear reason to me. Though just keeps that there are no duplicate models, which isn't wrong whatsoever.

src/accelerate/utils/dataclasses.py
```diff
@@ -1783,6 +1875,13 @@ def set_state_dict_type(self, state_dict_type=None):
             if self.optim_state_dict_config is None:
                 self.optim_state_dict_config = ShardedOptimStateDictConfig(offload_to_cpu=True)
 
+        # TODO(s1ro1): add support for FULL_STATE_DICT in FSDP2
+        if self.fsdp_version == 2 and self.state_dict_type != StateDictType.SHARDED_STATE_DICT:
+            raise ValueError(
+                "FSDP2 only supports SHARDED_STATE_DICT for now. "
+                "Please set `fsdp_state_dict_type` to `SHARDED_STATE_DICT`."
+            )
+
```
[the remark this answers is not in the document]
S1ro1: It is removed as of the official support, though we can simulate it by gathering the state dict on rank 0 and saving from there. Though this results in memory usage spike so should be done in a clever-ish way.

src/accelerate/utils/dataclasses.py
```diff
@@ -1635,27 +1663,69 @@ class FullyShardedDataParallelPlugin:
...
+        if self.fsdp_version is None:
+            self.fsdp_version = int(os.environ.get(env_prefix + "VERSION", "1"))
+
+        if self.sharding_strategy is not None:
+            # We cannot properly detect all of the cases, as by default `args.fsdp_sharding_strategy` is set to `fully_shard`
+            # Therefore we issue a warning only if the user has explicitly set it inside their plugin
+            _fsdp2_warnings.add(
+                "sharding_strategy is deprecated in favor of reshard_after_forward. "
+                "This will be removed in a future version of Accelerate."
+            )
```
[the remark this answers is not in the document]
S1ro1: We also want to warn in-case users specify it inside the plugin itself, which that wouldn't cover. That for me feels like the most important use-case of this warning.

src/accelerate/utils/dataclasses.py
```diff
@@ -1697,14 +1770,25 @@ def __post_init__(self):
...
+                "sync_module_states is obsolete in FSDP2, as it is not needed anymore."
+                "Setting sync_module_states to None."
+            )
+            self.sync_module_states = None
 
-        if self.forward_prefetch is None:
+        if self.forward_prefetch is None and self.fsdp_version == 1:
             self.forward_prefetch = str_to_bool(os.environ.get(env_prefix + "FORWARD_PREFETCH", "False")) == 1
+        if self.fsdp_version == 2 and self.forward_prefetch is not None:
+            raise ValueError("forward_prefetch is not yet implemented in FSDP2, set to None or use `fsdp_version=1`")
```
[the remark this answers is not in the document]
S1ro1: Seemed like a big enough of configuration decision to just overwrite it. Not a strong opinion.

src/accelerate/utils/fsdp_utils.py
S1ro1: Most likely not, `set_state_dict` is the reccomended way by Torch in all the guides, but that's why it's just a possible investigation.

src/accelerate/utils/fsdp_utils.py
S1ro1: Not properly, long explanation of how internal interactions work would be needed but haven't found a way to do it. Can take a look later, though not very important IMO.

tests/fsdp/test_fsdp.py
S1ro1: For testing FSDP2 only, we just remove that from `FSDP_VERSIONS` list. Though there are some if/else cases in the example tests, but not very intrusive to remove later.

src/accelerate/accelerator.py
```diff
@@ -1372,12 +1389,45 @@ def prepare(self, *args, device_placement=None):
...
+                    model_count += 1
+                elif isinstance(obj, torch.optim.Optimizer):
+                    optimizer_count += 1
+
+            # This needs to be written as such, so that passing other objects other than models/optimizers doesn't raise an error
+            if (model_count < 1 and optimizer_count > 0) or (model_count > 0 and optimizer_count < 1):
+                raise ValueError(
+                    "When using FSDP2, a model and optimizer must be passed together to `Accelerator.prepare()`"
+                    " as the optimizer needs to have its parameters modified after the model is converted."
+                )
```
kmehant, line 1406: Thank you for this work! Couple of comments/questions.

Wouldn't this approach conflict with the way transformers does it with `delay_optimizer_creation` (https://github.com/huggingface/transformers/blob/348f3285c5114159d2ff4933b4b8ae36866d01a7/src/transformers/trainer.py#L2314) knob?

src/accelerate/utils/constants.py
```diff
@@ -37,9 +37,11 @@
 FSDP_AUTO_WRAP_POLICY = ["TRANSFORMER_BASED_WRAP", "SIZE_BASED_WRAP", "NO_WRAP"]
 FSDP_BACKWARD_PREFETCH = ["BACKWARD_PRE", "BACKWARD_POST", "NO_PREFETCH"]
 FSDP_STATE_DICT_TYPE = ["FULL_STATE_DICT", "LOCAL_STATE_DICT", "SHARDED_STATE_DICT"]
+FSDP2_STATE_DICT_TYPE = ["SHARDED_STATE_DICT"]
 FSDP_PYTORCH_VERSION = (
     "2.1.0.a0+32f93b1"  # Technically should be 2.1.0, but MS-AMP uses this specific prerelease in their Docker image.
 )
+FSDP2_PYTORCH_VERSION = "2.5.1"
```
kmehant, line 44: Should we make this `2.6.0`? 

Since imports such as  https://github.com/huggingface/accelerate/blob/d7c741a6bc575878ccafcbdc957cc8fcb5ba5c34/src/accelerate/utils/dataclasses.py#L1987 would break for any version below 2.6.0

src/accelerate/accelerator.py
```diff
@@ -388,6 +392,10 @@ def __init__(
                 raise TypeError("`fsdp_plugin` must be a FullyShardedDataParallelPlugin object.")
             os.environ["ACCELERATE_USE_FSDP"] = "true"  # use FSDP if plugin is provided
 
+        if fsdp_plugin is not None and fsdp_plugin.fsdp_version == 2:
+            if not is_torch_version(">=", FSDP2_PYTORCH_VERSION):
+                raise ImportError(f"FSDP2 requires PyTorch >= {FSDP2_PYTORCH_VERSION}")
```
kmehant, line 397: Thank you for this check, you might also want to protect imports since the imports (such as https://github.com/huggingface/accelerate/blob/d7c741a6bc575878ccafcbdc957cc8fcb5ba5c34/src/accelerate/utils/dataclasses.py#L1987) seem to break first even before we hit this condition.

S1ro1: Thank you for the sanity checks. Will do more thorough version checks in the follow-up PRs.

src/accelerate/accelerator.py
S1ro1: Absolutely, if you do create a PR feel free to mention me, if not I'll take a look next week ish.

byi8220: Thanks! I actually do need to get Trainer to work with FSDP2, so if I have time I might take a look over the weekend if I have time. I'm not sure how much effort getting everything together would be. 

Hopefully it can literally be a 1 line change in `transformers`, but I ran into some OOM issue when I tried hacking it together, so maybe it won't be that easy.

S1ro1: The only requirement should be now to prepare model + optimizer with `accelerate` at the same time. Before that, the model shouldn't have anything applied to it, except of tp possibly (not sure on the interaction).

benchmarks/fsdp2/README.md
```diff
@@ -0,0 +1,74 @@
...
+This script showcases **matching memory usage and convergence between `accelerate` and `torch`'s baseline.**
+To deal with this breaking change (and maintain backward compatibility with FSDP1 in terms of an API), `accelerate` had to come up with a workaround since `accelerate` assumes that the user will nearly always create a model, optimizer, scheduler, etc beforehand and bring them themselves. This lead to an issue of a stark increase in memory as well as the model not even training if the user creates an optimizer beforehand. 
+To workaround this, we replace the parameters inside the optimizer with the newly created FSDP2 sharded ones. More about this can be found in this [blog post (TBD)](TODO)
+> [!WARNING]
+> This script is intended to fit on 2x 24GB GPUs, though on so few GPUs it's not possible to see the memory difference (discrepancies in grad allocation result in lower memory usage in the non-fixed case), only the difference in convergence. Below are attached results from 8x H100 GPUs where the difference is visible.
```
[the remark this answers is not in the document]
SunMarc: Thanks for the explanation !

src/accelerate/utils/fsdp_utils.py
SunMarc: yeah not very important, it was just a nit

tests/fsdp/test_fsdp.py
SunMarc: SG !

src/accelerate/utils/dataclasses.py
```diff
@@ -1635,27 +1663,69 @@ class FullyShardedDataParallelPlugin:
...
+            reshard_after_forward = os.environ.get(
+                env_prefix + "RESHARD_AFTER_FORWARD", "true" if self.fsdp_version == 2 else "FULL_SHARD"
+            )
+            if self.fsdp_version == 2:
+                self.reshard_after_forward = str_to_bool(reshard_after_forward.lower(), to_bool=True)
+            else:
+                self.reshard_after_forward = reshard_after_forward
+        if isinstance(self.reshard_after_forward, str):
+            if self.fsdp_version == 2:
+                self.reshard_after_forward = str_to_bool(self.reshard_after_forward.lower(), to_bool=True)
```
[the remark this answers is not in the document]
S1ro1: Yeah there are for sure caveats of this that are not totally figured out. I've spent hell of a time to make this work with all of the use-cases I could think of. 

I'd even be in favor of a somewhat radical solution:
- fsdp1 -> we use `sharding_strategy`
- fsdp2 -> we use `reshard_after_forward`

That makes this whole mess basically non-existent.

cc @muellerzr on what you think?

src/accelerate/accelerator.py
byi8220: > The only requirement should be now to prepare model + optimizer with accelerate at the same time. 

I drafted a tiny PR to do exactly that: https://github.com/huggingface/transformers/pull/37147 but I haven't sent it out yet since I haven't run a full multi-gpu test run yet.

However, when I am trying to fine tune a LORA, memory usage goes up dramatically. The VRAM usage is *significantly* higher when I kick off the job with `accelerate launch` compared to just running a single node/single gpu. 

I can't rule out the chance I'm doing something wrong, [maybe this is already solved](https://huggingface.co/docs/bitsandbytes/main/en/fsdp_qlora)

For example, when I tune an float16 version of `Llama-3.2-1B-Instruct` with a lora, during training my GPU memory usage peaks at 3944 MiB used, 4246 MiB reserved. I haven't dug into but this seems reasonable. However, when I try to do the exact same thing, except passing in an FSDP2 config and launching it via accelerate, it spikes up to 8264 MiB allocated, 13650 MiB reserved. 

I'm not super familiar with FSDP and I'm trying to follow https://github.com/pytorch/torchtune/blob/main/recipes/lora_finetune_distributed.py#L426 as a guide on how to get a memory efficient lora+hf trainer+fsdp2 working.

My guess is maybe we have to wrap the lora modules in isolation, and replicate what [lora_fsdp_wrap_policy](https://github.com/pytorch/torchtune/blob/2e5f6475f3d0b2c4f9ffa83e8db3623862636bcd/torchtune/training/_distributed.py#L245) is doing?

Not sure if this is relevant either, but the cause of https://github.com/huggingface/transformers/issues/34366 might be giving me issues as well.

S1ro1: @byi8220 I haven't done *any* research/work on using LoRA yet. This PR was just an initial work that supports simple FSDP2, I'll try to check how it interacts with LORA in the near future. Thanks for the checks on that. I think that preliminary merging your PR for transformers integration is fine, as it should support all that is currently supported by Accelerate, so I approved for now.

byi8220: SG. I'll try to find a second GPU and run the tests before I officially raise the PR. 

At worst this shouldn't break anything, since this is a no-op unless the user explicitly opts in.

> I'll try to check how it interacts with LORA in the near future. Thanks for the checks on that. 

I've mostly been hacking around, and I'm pretty unfamiliar with FSDP, so there is a high chance I am misinterpreting, or that I am not setting up my trainer properly. However everything I am suspecting as memory wastage seems logical to me (need to wrap separately, need to be clever with mmap). I might file an issue if I can make a clean reproduction

S1ro1: I'd suggest trying to do LORA with accelerate only first, do trainer after. As of the `mmap` stuff: getting that correctly has been a hassle in my limited experience with this bug/feature. Again, would leave it out of this and try to make simplest reproduction. If you won't come up with anything I'll try to look at this ~next week.

byi8220: > I'll try to find a second GPU and run the tests before I officially raise the PR.

PR was raised, passed 1/2 new tests. The other test had to be skipped (its FSDP 1 counterpart is skipped)

> I'd suggest trying to do LORA with accelerate only first, do trainer after. 

SG, I'll try to get something small and raise a bug for it soon. Getting fsdp2+lora in trainer working is a high priority blocker for me, so I will look into this tomorrow

> For FSDP2 we can just shard as we would usually.

Makes sense, I experimented with this a bit and found no savings.

> As of the mmap stuff: getting that correctly has been a hassle in my limited experience with this bug/feature. 

I can relate...

byi8220: Filed a bug about this, https://github.com/huggingface/accelerate/issues/3474

src/accelerate/accelerator.py
kmehant: Thanks @S1ro1, appreciate your review to PR https://github.com/huggingface/accelerate/pull/3499 since I needed these on my setup. Thanks.

docs/source/concept_guides/fsdp1_vs_fsdp2.md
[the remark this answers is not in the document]
dataScientistMS: Hey, any updates ?

PR #3396, opened 2025-02-14
[docs] add the missing `import torch`

faaany:
## What does this PR do?
minor optimization: both `torch.float16` need torch to be imported. 

cc @SunMarc and @muellerzr


SunMarc (approved):

faaany:
can we merge? :) @SunMarc

PR #3401, opened 2025-02-18
xpu: enable xccl distributed backend

dvrogozh:
xccl distributed backend is available for XPU device backend starting from torch 2.7 (requires torch built with `USE_XCCL=1 USE_C10D_XCCL=1`).

This change is verified with the following Transformers tests:
* `tests/extended/test_trainer_ext.py`
* `tests/trainer/test_trainer_distributed.py`

This commit does not impact IPEX which currently remains using custom distributed backend.

CC: @muellerzr

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -48,6 +48,7 @@ from .utils import (
     is_npu_available,
     is_sdaa_available,
     is_torch_xla_available,
+    is_xccl_available,
     is_xpu_available,
     parse_choice_from_env,
     parse_flag_from_env,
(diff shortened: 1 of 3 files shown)


muellerzr (approved):
Thanks, are there further changes needed for big model inference? Or is this sufficient

dvrogozh:
I am not aware of other changes specifically needed for big model inferencing - this should be sufficient per my understanding.  That being said, I do plan to continue looking to enable multi-gpu tests in Transformers and Accelerate libraries for XPU. If some tests will be failing, I will eventually post follow up PRs. Per my current try-out, this PR does help to fix some cases without introducing new errors.

FYI, I did rebase PR on top of latest main and updated commit message. Source code change remained intact.

SunMarc (approved):
Thanks !

SunMarc:
should be safe to safe, failing test unrelated

dvrogozh:
Thank you for merging, @SunMarc.

PR #3409, opened 2025-02-24
fix: ensure CLI args take precedence over config file.

cyr0930:
# What does this PR do?
Reapply this PR (https://github.com/huggingface/accelerate/pull/2757) with some fixes.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@muellerzr @lain-S

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -1043,35 +1043,20 @@ def _validate_launch_command(args):
             # Update args with the defaults
             for name, attr in defaults.__dict__.items():
                 if isinstance(attr, dict):
-                    for k in defaults.deepspeed_config:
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
Thanks for the PR. Could you please provide an example where currently the CLI args don't take precedence and that would be resolved by your fix?

cyr0930:
I added a test case.
As you can see in the test case, if config file options are instance of dict, they preceded cli options.
"deepspeed_config_file" option in the test is the case.

cyr0930:
I think there's nothing to change in the document. Mention this in the release note would be fine.

muellerzr (approved):
Thanks for fixing this with deepspeed!

Issue #3438, opened 2025-03-12
Version 1.5.0 explicitly depends on pytest


adigioacchino:
# Issue
Using the latest release (v1.5.0) I get a `ModuleNotFoundError: No module named 'pytest'` when launching any script with `accelerate launch` if pytest is not installed in the virtual environment.

Using v1.4.0 everything works fine, and I think this is the expected behavior.

# Reproducibility
I run
`uv run accelerate launch hello.py`
with the following files:

**hello.py**
```hello.py
def main():
    print("Hello from temp!")


if __name__ == "__main__":
    main()
```

**pyproject.toml**
```
[project]
name = "temp"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    "accelerate>=1.5.0",
]
```

The resulting full traceback:
```
  File "/home/ssm-user/temp/.venv/bin/accelerate", line 4, in <module>
    from accelerate.commands.accelerate_cli import main
  File "/home/ssm-user/temp/.venv/lib/python3.13/site-packages/accelerate/commands/accelerate_cli.py", line 22, in <module>
    from accelerate.commands.test import test_command_parser
  File "/home/ssm-user/temp/.venv/lib/python3.13/site-packages/accelerate/commands/test.py", line 19, in <module>
    from accelerate.test_utils import execute_subprocess_async, path_in_accelerate_package
  File "/home/ssm-user/temp/.venv/lib/python3.13/site-packages/accelerate/test_utils/__init__.py", line 14, in <module>
    from .testing import (
    ...<43 lines>...
    )
  File "/home/ssm-user/temp/.venv/lib/python3.13/site-packages/accelerate/test_utils/testing.py", line 30, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

muellerzr:
Thanks for flagging, should be fixed now with 1.5.1. Can you confirm?

adigioacchino:
Yes, v1.5.1 fixes this. Thanks @muellerzr!

Closing the issue 😄

PR #3420, opened 2025-03-04
fix: Add `device=torch.get_default_device()` in `torch.Generator`s

saforem2:
# What does this PR do?

Replaces calls to

```python
torch.Generator()
```

with

```python
torch.Generator(device=torch.get_default_device())
```

in [`src/accelerate/data_loader.py`](src/accelerate/data_loader.py).

Prior to this fix, I was seeing:

<details closed><summary>🐛 <b>Bug</b>:</summary>

```python
File /lus/flare/projects/datascience/foremans/projects/argonne-lcf/AuroraGPT/venvs/aurora_nre_models_frameworks-2024.2.1_u1/lib/python3.10/site-packages/accelerate/data_loader.py:99, in SeedableRandomSampler.__iter__(self)
     97 # print("Setting seed at epoch", self.epoch, seed)
     98 self.generator.manual_seed(seed)
---> 99 yield from super().__iter__()
    100 self.set_epoch(self.epoch + 1)

File /opt/aurora/24.180.3/frameworks/aurora_nre_models_frameworks-2024.2.1_u1/lib/python3.10/site-packages/torch/utils/data/sampler.py:167, in RandomSampler.__iter__(self)
    165 else:
    166     for _ in range(self.num_samples // n):
--> 167         yield from torch.randperm(n, generator=generator).tolist()
    168     yield from torch.randperm(n, generator=generator).tolist()[:self.num_samples % n]

File /opt/aurora/24.180.3/frameworks/aurora_nre_models_frameworks-2024.2.1_u1/lib/python3.10/site-packages/torch/utils/_device.py:78, in DeviceContext.__torch_function__(self, func, types, args, kwargs)
     76 if func in _device_constructors() and kwargs.get('device') is None:
     77     kwargs['device'] = self.device
---> 78 return func(*args, **kwargs)

RuntimeError: Expected a 'xpu' device type for generator but found 'cpu'
```

</details>

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
[...]

PR #3441, opened 2025-03-13
Fix prod issues

muellerzr:
# What does this PR do?

Fixes an issue caused by https://github.com/huggingface/accelerate/pull/3420 where said code only works if we're on torch 2.6+. Used a `hasattr` check instead here to make it work. 

Fixes https://github.com/huggingface/accelerate/issues/3438

(Did a new PR since the old one had outdated main).

I'll do a secondary patch with these fixes once merged

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @BenjaminBossan @IlyasMoutawwakil

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -90,7 +90,7 @@ class SeedableRandomSampler(RandomSampler):
     def __iter__(self):
         if self.generator is None:
             self.generator = torch.Generator(
-                device=torch.get_default_device() if hasattr(torch, "get_default_device") else None
+                device=torch.get_default_device() if hasattr(torch, "get_default_device") else "cpu"
             )
             self.generator.manual_seed(self.initial_seed)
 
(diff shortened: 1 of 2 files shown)


src/accelerate/data_loader.py
```diff
@@ -89,7 +89,9 @@ def __init__(self, *args, **kwargs):
 
     def __iter__(self):
         if self.generator is None:
-            self.generator = torch.Generator(device=torch.get_default_device())
+            self.generator = torch.Generator(
+                device=torch.get_default_device() if hasattr(torch, "get_default_device") else None
```
BenjaminBossan, line 93: Note that according to the [docs](https://pytorch.org/docs/stable/generated/torch.Generator.html), the default value of `device` for `Generator` is `"cpu"`. Maybe `None` would also work but I think `"cpu"` is the safer bet.

PR #3390, opened 2025-02-10
tensor parallel dataloder for deepspeed accelerator

inkcherry:
This PR adds support for TP data loader in the DeepSpeed accelerator.  
The related DeepSpeed APIs are already supported.  
Currently, TP+DP training and validation (ZeRO-0 and ZeRO-1) for Llama have been successfully completed.  
(https://github.com/deepspeedai/DeepSpeed/pull/6922)   @muellerzr 

cc @sywangyi  @delock
cc DeepSpeed Member: @tjruwase

PR #3443, opened 2025-03-13
Fix attribute issue with deepspeed tp

SunMarc:
# What does this PR do ?

This PR fixes https://github.com/huggingface/accelerate/pull/3390#issuecomment-2722005648. deepspeed_config don't necessarily have the `tensor_parallel` key


muellerzr (approved):

inkcherry:
thanks for the fix!

Issue #3402, opened 2025-02-20
Transformers test_cpu_offload tests fail with KeyError: 'xpu:0'


dvrogozh:
With:
* (accelerate) https://github.com/huggingface/accelerate/commit/8039158d71418e4520113e43bfe3567ffeedd7db
* https://github.com/huggingface/transformers/commit/31bb662db19355e56b3125ab5c48b780ec13111c
* Pytorch XPU backend (pytorch 2.7 xpu nightly)
* Note: no IPEX

On:
* Intel Data Center GPU Max (single device)

```
FAILED tests/models/blip/test_modeling_blip.py::BlipVQAModelTest::test_cpu_offload - KeyError: 'xpu:0'
FAILED tests/models/blip/test_modeling_blip.py::BlipVQAModelTest::test_disk_offload_bin - KeyError: 'xpu:0'
FAILED tests/models/blip/test_modeling_blip.py::BlipVQAModelTest::test_disk_offload_safetensors - KeyError: 'xpu:0'
FAILED tests/models/blip/test_modeling_blip.py::BlipTextImageModelTest::test_cpu_offload - KeyError: 'xpu:0'
FAILED tests/models/blip/test_modeling_blip.py::BlipTextImageModelTest::test_disk_offload_bin - KeyError: 'xpu:0'
FAILED tests/models/blip/test_modeling_blip.py::BlipTextImageModelTest::test_disk_offload_safetensors - KeyError: 'xpu:0'
FAILED tests/models/dab_detr/test_modeling_dab_detr.py::DabDetrModelTest::test_cpu_offload - KeyError: 'xpu:0'
FAILED tests/models/dab_detr/test_modeling_dab_detr.py::DabDetrModelTest::test_disk_offload_bin - KeyError: 'xpu:0'
FAILED tests/models/dab_detr/test_modeling_dab_detr.py::DabDetrModelTest::test_disk_offload_safetensors - KeyError: 'xpu:0'
FAILED tests/models/vilt/test_modeling_vilt.py::ViltModelTest::test_cpu_offload - KeyError: 'xpu:0'
FAILED tests/models/vilt/test_modeling_vilt.py::ViltModelTest::test_disk_offload_bin - KeyError: 'xpu:0'
FAILED tests/models/vilt/test_modeling_vilt.py::ViltModelTest::test_disk_offload_safetensors - KeyError: 'xpu:0'
```

[...]

PR #3448, opened 2025-03-20
remove device index workaround on xpu since xpu supports integer device index as cuda now

yao-matrix:
remove device index workaround on xpu since it supports integer device index as cuda now

this can fix https://github.com/huggingface/accelerate/issues/3402, this issue is brought by inconsistent device key(integer(x) vs string("xpu:x")).

- blip, dab_detr and vilt can pass
- roberta fail, but CUDA fail w/ same log, so it's another issue, may open another PR to fix

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -495,8 +495,6 @@ def dispatch_model(
             device = f"sdaa:{device}"
         elif is_musa_available() and isinstance(device, int):
             device = f"musa:{device}"
-        elif is_xpu_available() and isinstance(device, int):
-            device = f"xpu:{device}"
         if device != "disk":
             model.to(device)
         else:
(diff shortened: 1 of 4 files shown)


yao-matrix:
@faaany , since it's a pretty fundamental changes, could you help run a round of UTs on XPU for this PR? I suppose we should run all accelerate UTs and transformers modeling related UTs. Thx.

yao-matrix:
@dacorvo

faaany:
I ran this PR on Accelerate and Transformers' UT XPU. There was no error related to the change in this PR.

muellerzr (commented):
Nice! Thanks for the verify @faaany

muellerzr (approved):

muellerzr:
@yao-matrix can you do `pip install -e .[quality]; make style; make quality;`? This should fix the last failure

yao-matrix:
seems the 3 CI fails are triggered by CI issue rather than my PR.

SunMarc (approved):
Thanks ! Indeed, the failing tests are not related, so I will merge it

PR #3451, opened 2025-03-20
check that an object has .float() method

eljandoubi:
# What does this PR do?

Fixes issue from https://github.com/huggingface/transformers/pull/35873

## Who can review

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -778,9 +778,14 @@ def convert_to_fp32(tensor):
         return tensor.float()
 
     def _is_fp16_bf16_tensor(tensor):
-        return (is_torch_tensor(tensor) or hasattr(tensor, "dtype")) and tensor.dtype in (
-            torch.float16,
-            torch.bfloat16,
+        return (
+            hasattr(tensor, "dtype")
+            and tensor.dtype
+            in (
+                torch.float16,
+                torch.bfloat16,
+            )
+            and callable(getattr(tensor, "float", None))
         )
 
     return recursively_apply(_convert_to_fp32, tensor, test_type=_is_fp16_bf16_tensor)


src/accelerate/utils/operations.py
```diff
@@ -778,9 +778,14 @@ def _convert_to_fp32(tensor):
         return tensor.float()
 
     def _is_fp16_bf16_tensor(tensor):
-        return (is_torch_tensor(tensor) or hasattr(tensor, "dtype")) and tensor.dtype in (
```
BenjaminBossan, line 781: I'm a bit surprised that cache objects would pass this check successfully. Could you please provide an example where this happens? It would be very useful to have that as a unit test. Moreover, is it intended that the `is_torch_tensor(tensor)` part was removed?

eljandoubi: @BenjaminBossan Any torch tensor has `dtype` attribute. So no need for `is_torch_tensor` with the or `hasattr(tensor, "dtype"))`. For the exemple, the Cache class is.

BenjaminBossan: > Any Torch tensor has a dtype attribute, so there is no need for `is_torch_tensor`. `is_torch_tensor(tensor) or hasattr(tensor, "dtype") == hasattr(tensor, "dtype")`

Note that this was specifically added in #2387 to avoid `hasattr(tensor, "dtype")` for `torch.compile`. I'm not sure if this issue still persists with newer PyTorch versions, but let's not change it without verifying first.

> For the example, in `transformers=4.50.0` the `StaticCache`, `HybridCache`, `MambaCache` and `OffloadedStaticCache` initialized with `dtype=torch.bfloat16` pass the current test `_is_fp16_bf16_tensor`

I see, the issue is that `Cache` objects seem to have a `dtype` attribute. I think it is okay to check for `callable(getattr(tensor, "float", None))`. But again, let's first check how this interacts with `torch.compile`. If it causes issues, we could instead do `if not isinstance(tensor, Cache) and ...`.

eljandoubi:
@gante The PR LGTM. Still, if an object has dtype, it needs to have a caster method, doesn't it?

PR #3173, opened 2024-10-16
feat: support tensor parallel & Data loader

kmehant:
# What does this PR do?

1. Implements `TorchTensorParallelPlugin` to support TP with Pytorch 2.0. This work should be seen along with the PR https://github.com/huggingface/transformers/pull/34194.
2. https://github.com/huggingface/transformers/pull/34184
3. Modifies dataloader to support passing same samples across TP ranks

Please review in conjunction with https://github.com/huggingface/transformers/pull/34194

# Results

See significant improvement in both memory and throughput compared against single gpu training, and FSDP across different settings (checkpointing on/off) and context lengths.

Done on two models
1. ibm-granite/granite-8b-code-base-128k 
4. codellama/CodeLlama-7b-hf

Tables below show the max cuda memory and throughput for various configurations showing the potential of TP contributed in this PR. There is gains in both memory and throughput.

**Note:** Please be aware that the effective TPS for FSDP would be multiplicative of the parallel factor (number of GPUs/devices engaged in distributed training) whereas that is not the case with TP. Therefore, when effective throughput is considered we can find FSDP is better than TP in terms of throughput. However, that **may** be compensated by increasing the batch size utilizing the memory gains etc. 

Model | Method | # of GPUs | Context Length | Batch Size | Grad Checkpointing | Cuda Max Mem (GiB) | Tokens/Sec/GPU
-- | -- | -- | -- | -- | -- | -- | --
ibm-granite/granite-8b-code-base-128k | Single GPU non-distributed | 1 | 8192 | 1 | FALSE | OOM | NA
ibm-granite/granite-8b-code-base-128k | FSDP | 4 | 8192 | 1 | FALSE | OOM | NA
ibm-granite/granite-8b-code-base-128k | TP (This PR) | 4 | 8192 | 1 | FALSE | 52.4 | 7675.4

Model | Method | # of GPUs | Context Length | Batch Size | Grad Checkpointing | Cuda Max Mem (GiB) | Tokens/Sec/GPU
-- | -- | -- | -- | -- | -- | -- | --
ibm-granite/granite-8b-code-base-128k | Single GPU non-distributed | 1 | 8192 | 1 | TRUE | OOM | NA
ibm-granite/granite-8b-code-base-128k | FSDP | 4 | 8192 | 1 | TRUE | 29.975586 | 2256.896
ibm-granite/granite-8b-code-base-128k | TP (This PR) | 4 | 8192 | 1 | TRUE | 26.5 | 5935.5

[...]

PR #3457, opened 2025-03-25
(Part 1) fix: make TP training compatible with new transformers

### LINKED ISSUE #3456: tensor_parallel public API is deprecated and removed
### System Info

```Shell
transformers==v4.50.0
accelerate==v1.5.2
```

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

https://github.com/huggingface/transformers/pull/36033 and some recent TP refactors in HF/transformers break use of tensor parallel with accelerate since `tensor_parallel()` API no longer exists.

We would need to update https://github.com/huggingface/accelerate/blob/806f661cd31a8f058ef7dd4d47f77dea9c69e89f/src/accelerate/accelerator.py#L1554 to utilize the latest sharding functions.

### Expected behavior

Tensor parallel training should work with latest accelerate and transformer releases.

kmehant:
# What does this PR do?

Fixes #3456

## Before submitting
      Pull Request section?
      to it if that's the case. #3456

## Who can review?

Thanks to @SunMarc for valuable discussion over #3456

@muellerzr or @SunMarc


src/accelerate/accelerator.py
```diff
@@ -1551,7 +1551,11 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                         Tensor parallelism plan can be added as base_model_tp_plan to model config class \
                         and _tp_plan attribute to model class."
                     )
-                model.tensor_parallel(self.state.torch_tp_plugin.torch_device_mesh["tp"])
+                # For transformers>4.49, model should be TP sharded outside the trainer while loading using .from_pretrained(..., tp_plan="auto")
+                if compare_versions("transformers", ">=", BETA_TP_AVAILABLE_TRANSFORMERS_VERSION) and compare_versions(
+                    "transformers", "<=", "4.49.0"
+                ):
+                    model.tensor_parallel(self.state.torch_tp_plugin.torch_device_mesh["tp"])
```
SunMarc, line 1558: Let's request the user to download `transformers 4.50` for TP training -> just update `BETA_TP_AVAILABLE_TRANSFORMERS_VERSION`. I don't think we want to maintain this as this was available only for one release.

SunMarc (commented):
Thanks ! Left a comment

src/accelerate/accelerator.py
kmehant: Absolutely, agree with you. Have made the change. Thanks

SunMarc: But now I feel like having `DistributedType.TP` does nothing in accelerate, maybe we should just get rid of it ?

SunMarc: We can add a quick attribute in transformers to check that the model is indeed initialized with tp

kmehant: > We can add a quick attribute in transformers to check that the model is indeed initialized with tp

Let me look it up and happy to add it! But I am afraid since this wont extend to "bring your own model" case where the model could have gone through TP sharding outside of from_pretrained (I would say a rare case).

SunMarc: > But if you are referring to the whole DistributedType.TP logic in accelerate then I am afraid, since we have specific dataloader settings that we do for TP. more details in this PR - https://github.com/huggingface/accelerate/pull/3173
> 
> Let me know your thoughts.

Oh yes I forgot. It relies on the TorchTensorParallelPlugin object with a tp_size that we passed in accelerate config. However, now the model is initialized with transformers, we need make sure that we create the right `torch_device_mesh` with the right `tp_size` or simply make sure that we `tp_size` from the plugin is the same of the model, otherwise we return an error.

SunMarc: maybe add an attribute on the loaded model called `tp_size` ?

SunMarc: > Let me look it up and happy to add it! But I am afraid since this wont extend to "bring your own model" case where the model could have gone through TP sharding outside of from_pretrained (I would say a rare case).

indeed we can add more check in the future if needed to take care of this edge case

muellerzr (approved):
Thanks!

src/accelerate/accelerator.py
kmehant: Hi @SunMarc 

To what we have discussed I have done this piece of work in two parts.

1. Part 1: https://github.com/huggingface/transformers/pull/37054
2. Part 2: https://github.com/huggingface/accelerate/pull/3457 (current PR)

Thank you.

SunMarc (commented):
I'll review that after merging the transformers PR ! But for a quick look it looks nice

src/accelerate/accelerator.py
S1ro1: > But now I feel like having `DistributedType.TP` does nothing in accelerate, maybe we should just get rid of it ? Or we can use that to perform a couple of check to see if the model is correctly initialized with tp

As stated below, we do need it for some dataloader settings (not 100% sure on the internals)
THOUGH refactoring `--use-tp` to not be exclusive with other parallelism settings would be nice. With `FSDP2`, `DDP` we have quite few use-cases where extending with `TP` could be nice. Also bearing in mind `FSDP`/`ZERO` + `TP` is one of the common and more effective scaling techniques on larger clusters.

Possibly keeping `--use-tp` as an extension, not a unique setting. I think combining with `FSDP2`/`DDP` should be pretty seamless.

SunMarc: yeah, we should try to make it easy to use tp with ddp and fsdp, we already have support for deepspeed tp which is nice also.

src/accelerate/state.py
```diff
@@ -966,7 +966,7 @@ def __init__(
                     self.distributed_type = DistributedType.MEGATRON_LM
                     megatron_lm_plugin.set_mixed_precision(self._mixed_precision)
                     self.megatron_lm_plugin = megatron_lm_plugin
-                if os.environ.get("ACCELERATE_USE_TP", "false") == "true" or self.torch_tp_plugin is not None:
+                if self.torch_tp_plugin is not None:
                     self.distributed_type = DistributedType.TP
```
S1ro1, line 970: Not sure if we want to cover this in this PR or another one, but with changes planned to support TP with DDP/FSDP2, it would be helpful to not use `self.distributed_type` rather something as `self.is_tp`. 

If not, we need to do that in another PR.

src/accelerate/test_utils/scripts/external_deps/test_performance.py
```diff
@@ -255,6 +254,12 @@ def main():
         default=False,
         help="To add pad token if not exists.",
     )
+    parser.add_argument(
+        "--tp_plan",
+        type=str,
+        default=None,
+        help="To use TP or not",
```
S1ro1, line 261: Maybe make this bit more helpful: I.e. what options are supported - this implies we want yes/no answer, even though we want "auto" etc

src/accelerate/commands/config/config_args.py
```diff
@@ -194,8 +194,6 @@ class ClusterConfig(BaseConfig):
     deepspeed_config: dict = None
     # args for fsdp
     fsdp_config: dict = None
-    # args for tp
-    tp_config: dict = None
```
S1ro1, line 198: Also can we remove this attr in `__post_init__` of `ClusterConfig` I think it's gonna complain about non-existing attribute there

kmehant: nice catch thanks!

src/accelerate/test_utils/scripts/external_deps/test_performance.py
kmehant: review, well taken, thank you, have modified, hope its more sensible now. Thanks

src/accelerate/state.py
S1ro1: @kmehant I have opened #3498 to discuss exactly that, it builds on top of your PR. It's very much work in progress and I need to think the api changes thoroughly, I agree we can leave it as is and move it there.

kmehant:
Failing test is unrelated. Thanks

S1ro1:
> Thanks ! Now that we've merged the PR about `tp_size` in transformers, maybe we can use that to infer automatically the `tp_size` so that we create the plugin accordingly. Not sure how well this will integrate with the current code as we don't have access to the model when creating accelerator

I think we can also cover this in #3498 as for example Trainer does create the `TensorParallelPlugin` for us if model was sharded. I'm all for removing it and leaving it on accelerate to create plugins based on `tp_size` provided.

src/accelerate/accelerator.py
```diff
@@ -396,14 +394,8 @@ def __init__(
...
-        if torch_tp_plugin is None:
-            torch_tp_plugin = (
-                TorchTensorParallelPlugin() if os.environ.get("ACCELERATE_USE_TP", "false") == "true" else None
-            )
-        else:
-            if not isinstance(torch_tp_plugin, TorchTensorParallelPlugin):
-                raise TypeError("`torch_tp_plugin` must be a TorchTensorParallelPlugin object.")
-            os.environ["ACCELERATE_USE_TP"] = "true"
+        if torch_tp_plugin is not None and not isinstance(torch_tp_plugin, TorchTensorParallelPlugin):
+            raise TypeError("`torch_tp_plugin` must be a TorchTensorParallelPlugin object.")
```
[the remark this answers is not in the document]
kmehant: Approach 1: Should we allow for passing model while creating accelerator and doc string it mentioning it would only be used for TP? This approach would also come in handy for any parallelism that modifies the model like context parallel. This would allow users to not needing to create the plugin and pass model while creating accelerator. 

Approach 2: modify TorchTensorParallelPlugin to take model as input rather tp_size, this way we enforce to users that to enable TP, they would need to pass a tplized model only. then we extract tp_size from model to create device mesh to be used by data loader.

Approach 3:  We use tp_size to only validate if it matches with what model has been sharded to + creating device mesh for data loader (that we have it already). However, it would still feel redundant, from user's PoV since they would need to pass the same tp_size while sharding in transformers and while using the plugin. 


Let me know which one sounds good for this PR. Thanks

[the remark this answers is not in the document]
S1ro1: I'm up for approach 3, though we shouldn't use it to validate, but as a fallback if model has no attr `tp_size`, this way there's no redundancy. Also you'll notice in #3498 I also do the same for `DP/FSDP_size` as those will be needed as well to recreate the same device mesh as was created for TP (gpu order is different in 1d/2d case), so my proposition is:

Extra dataclass - i.e. ParallelismConfig
Which would hold all parallelism sizes, would be used as a fallback though.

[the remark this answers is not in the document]
kmehant: Thanks @S1ro1 

@SunMarc your thoughts?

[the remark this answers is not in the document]
SunMarc: Hmmm i think approach 3 will be better. I don't want to force users to create the model before initializing accelerateor or `TorchTensorParallelPlugin`. Maybe in the future, accelerate also take care for sharding the any model for tp. 
We can just check when preparing the model that if the model is tp-sharded, we check the tp size vs the one passed in `TorchTensorParallelPlugin`. If the model is not tp-sharded but the user passed a `TorchTensorParallelPlugin`, we return an error. WDYT ?

SunMarc:
> I think we can also cover this in https://github.com/huggingface/accelerate/pull/3498 as for example Trainer does create the TensorParallelPlugin for us if model was sharded. I'm all for removing it and leaving it on accelerate to create plugins based on tp_size provided.

I don't mind if you think if it can make things simpler for you. What are the issue currently with `TensorParallelPlugin` ?

src/accelerate/accelerator.py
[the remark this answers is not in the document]
kmehant: Thanks @SunMarc have gone with this approach and updated the PR. Performance test for TP passes as well. Can we merge this PR as a self contained piece for now and visit the discussion at https://github.com/huggingface/accelerate/pull/3498 for optimizations for nd parallel? Thanks

Screenshot 2025-04-11 at 4 03 06 PM

cc: @S1ro1

src/accelerate/accelerator.py
```diff
@@ -1598,15 +1590,17 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                     if self.ddp_handler is not None:
                         self.ddp_handler.register_comm_hook(model)
             elif self.distributed_type == DistributedType.TP:
-                if hasattr(model, "supports_tp_plan") and not model.supports_tp_plan:
-                    if not compare_versions("transformers", ">=", BETA_TP_AVAILABLE_TRANSFORMERS_VERSION):
-                        raise ValueError(f"TP requires transformers >= {BETA_TP_AVAILABLE_TRANSFORMERS_VERSION}")
+                if not compare_versions("transformers", ">=", BETA_TP_AVAILABLE_TRANSFORMERS_VERSION):
+                    raise ValueError(f"TP requires transformers >= {BETA_TP_AVAILABLE_TRANSFORMERS_VERSION}")
+                if not hasattr(model, "tp_size"):
```
SunMarc, line 1595: This was added only recently, so we have to update `BETA_TP_AVAILABLE_TRANSFORMERS_VERSION` to `4.52.0` or the dev version.

SunMarc (approved):
Thanks, minor nit on the version check but other than that LGTM ! You can merge it @S1ro1 if you are fine if that

src/accelerate/accelerator.py
kmehant: Apologies on missing that, have updated it to `4.52.0` Thanks.

S1ro1:
> Sounds good for 1. I'm all for removing `TensorParallelPlugin` and using `tp_size` instead. Note that right now, for deepspeed, we are creating a separate device_mesh also. Check `_prepare_device_mesh` method.

As for the device mesh, that is fine with `DeepSpeedPlugin` as that encapsulates all levels of parallelism (dp, fsdp, tp), though in our case, we want TP to be composable with either of DDP or FSDP or both, where we don't have a clear "leader" that will be responsible for device mesh. Therefore the best option is to create it centrally and it being accessible from all FSDP/DDP plugins, possibly other such as PP if we decide to support that. Later we can move to having a central device mesh for DeepSpeed as well, but that's possibly breaking and bigger refactor.

However, this is to be discussed later, merging and let's move the discussion to #3498 .

PR #3459, opened 2025-03-26
Fix seeding of new generator for multi GPU

### LINKED ISSUE #3398: Randomization with seed not working
### System Info

```Shell
- `Accelerate` version: 0.32.1
- Platform: Linux-6.5.0-1022-aws-x86_64-with-glibc2.35
- Python version: 3.10.14
- Numpy version: 1.26.3
- System RAM: 1121.80 GB
- GPU type: NVIDIA A100-SXM4-40GB
```

### Information

- [x] My own modified scripts

### Tasks

- [x] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)

### Reproduction

Basically, what I am trying to do, is run a finetuning script on multi-GPU setting (Distributed Data Parallel). My setup looks like this : [https://github.com/huggingface/transformers/blob/main/examples/pytorch/language-modeling/run_mlm_no_trainer.py#L311](url).

Let me provide a toy example to exhibit what is the problem.

```python
import torch

set_seed(0)

accelerator = Accelerator()
dataset = range(10)
dataloader = torch.utils.data.DataLoader(dataset, batch_size=1, shuffle=True)
accelerator.print(type(dataloader))

accelerator.print(type(dataloader))

epochs=2

for epoch in range(epochs):
    accelerator.print(f"==============================================")
    accelerator.print(f"Epoch {epoch}")
    
    for idx, batch in enumerate(dataloader):
        if idx == 0:
            print(f'Batch {idx}: {batch}')
        
```

### Expected behavior

If i run this script with python (one process), setting the random seed ensures reproducibility. 
Changing the seed ensures that I have different shuffling  of the data.

But, if I run the script with accelerate launch (multi-process), the dataloader shuffles the data in the same way every time, regardless of the random seed. For example, by changing the random seed, the first batch  remains the exact same, and the data gets shuffled in the same way.
This is a bit strange to me, as the set_seed function (transformers' or accelerate's) achieves reproducibility in my one-process experiments. But I wanted to run some experiments on a distributed scenario (Multi-GPU), and I noticed this.

You can run the script that i provided with multiple seeds (or even no seed if you comment the set_seed line), and the fetched batches will be the exact same.
I also tried to include 

```python
dataloader_config=DataLoaderConfiguration(
        use_seedable_sampler=True,
    )
```
in the Accelerator initialization. by doing this, the random seeds worked for the first epoch (I mean that when I changed the seed, the data was fetched in different ways). BUT, another problem occured: the shuffling before each epoch did not work. So this was not a proper fix (?).

Please let me know if there is any proper way of fixing this, to ensure reproducible experiments with multiple seeds. I cannot find any sufficient answer  on this, it seems like a bug to me.

albertcthomas:
# What does this PR do?

Fixes #3398 

The initial state of the new generator was always the same, whatever the seed passed to `accelerate.utils.set_seed`. This PR makes sure that we get reproducible results but different numbers for different seeds passed to `set_seed`. This is already the case when multi GPU is disabled because generator is None (see https://github.com/pytorch/pytorch/blob/main/torch/utils/data/sampler.py#L170).
Before adding a test, I will wait for a first reviewer to confirm this is OK, I am not yet very familiar with accelerate.

Maybe the same should be done for https://github.com/huggingface/accelerate/blob/main/src/accelerate/data_loader.py#L1184 where the seed is set to 42 and thus will always produce the same number whatever the first global initial seed passed to `set_seed`

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

Based on git blame I would say @muellerzr :)

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -1203,6 +1203,8 @@ def prepare_data_loader(
                     sampler.generator = torch.Generator(
                         device=torch.get_default_device() if hasattr(torch, "get_default_device") else "cpu"
                     )
+                    seed = int(torch.empty((), dtype=torch.int64).random_().item())
+                    sampler.generator.manual_seed(seed)
                 synchronized_generator = sampler.generator
             batch_sampler = dataloader.sampler if sampler_is_batch_sampler else dataloader.batch_sampler
             new_batch_sampler = BatchSamplerShard(
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -1181,7 +1181,9 @@ def prepare_data_loader(
         # isinstance(dataloader.sampler, RandomSampler) indicates the original dataloader has `shuffle` enabled.
         generator = torch.Generator(
             device=torch.get_default_device() if hasattr(torch, "get_default_device") else "cpu"
-        ).manual_seed(42)
+        )
+        seed = int(torch.empty((), dtype=torch.int64).random_().item())
+        generator.manual_seed(seed)
         dataloader.generator = generator
         dataloader.sampler.generator = generator
     # No change if no multiprocess
--- a/tests/test_data_loader.py
+++ b/tests/test_data_loader.py
@@ -34,7 +34,7 @@ from accelerate.data_loader import (
 )
 from accelerate.state import GradientState
 from accelerate.test_utils.testing import AccelerateTestCase, require_torchdata_stateful_dataloader
-from accelerate.utils import is_torchdata_stateful_dataloader_available
+from accelerate.utils import is_torchdata_stateful_dataloader_available, set_seed
 
 
 if is_torchdata_stateful_dataloader_available():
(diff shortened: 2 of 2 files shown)


muellerzr (approved):
Nice catch, thanks!

muellerzr:
Yes the same should be done there. Do you want to include it in this PR?

albertcthomas:
> Yes the same should be done there. Do you want to include it in this PR?

Thanks for the quick check. Yes I will add it in this PR. I'll try to add a test as well.

albertcthomas:
> Yes I will add it in this PR. I'll try to add a test as well.

Done.

albertcthomas:
The added test fails on main, pass on this branch. Not sure this covers the other line I just changed in the code.

muellerzr (approved):
Thanks for adding a test!

albertcthomas:
Thanks for the quick review!

PR #3460, opened 2025-03-27
remove use_xpu to fix ut issues, we don't need this since XPU is OOB …

yao-matrix:
**Sympton**
When run transformers e2e training ut `pytest -rA tests/trainer/test_trainer.py::TrainerIntegrationTest::test_end_to_end_example` on XPU, we will get below error:

> RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and xpu:0! (when checking argument for argument index in method wrapper_XPU__index_select)

**Root Cause**
When run this test case while there is no default accelerate config in “/root/.cache/huggingface/accelerate/default_config.yaml”（generated by `accelerate config`）, the `use_xpu` flag in `launch.py `(default False) blocked the using of XPU, so accelerate is launched both in single process and in cpu, but when data comes, it automatically detected xpu and put data into xpu, so model is in cpu and data is in xpu, lead to this error.

**Fix**

XPU is now native citizen of PyTorch, we need align the same behavior w/ CUDA, so remove these special handling.

### DIFF
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -158,7 +158,6 @@ The following arguments are useful for selecting which training paradigm to use.
 * `--use_deepspeed` (`bool`) -- Whether or not to use DeepSpeed for training.
 * `--use_fsdp` (`bool`) -- Whether or not to use FullyShardedDataParallel for training.
 * `--use_megatron_lm` (`bool`) -- Whether or not to use Megatron-LM for training.
-* `--use_xpu` (`bool`) -- Whether to use IPEX plugin to speed up training on XPU specifically.
 
 **Distributed GPU Arguments**:
 
--- a/docs/source/package_reference/cli.md
+++ b/docs/source/package_reference/cli.md
@@ -158,6 +158,7 @@ The following arguments are useful for selecting which training paradigm to use.
 * `--use_deepspeed` (`bool`) -- Whether or not to use DeepSpeed for training.
 * `--use_fsdp` (`bool`) -- Whether or not to use FullyShardedDataParallel for training.
 * `--use_megatron_lm` (`bool`) -- Whether or not to use Megatron-LM for training.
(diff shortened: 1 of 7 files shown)


yao-matrix:
@sywangyi  @faaany , this is a pretty fundamental change, pls review first

yao-matrix:
@faaany , since it changes the default behavior, could you help run accelerate UTs for this? Thx

faaany:
> @faaany , since it changes the default behavior, could you help run accelerate UTs for this? Thx

sure

yao-matrix:
@SunMarc @muellerzr , pls help review, thx.

docs/source/package_reference/cli.md
```diff
@@ -158,7 +158,6 @@ The following arguments are useful for selecting which training paradigm to use.
 * `--use_deepspeed` (`bool`) -- Whether or not to use DeepSpeed for training.
 * `--use_fsdp` (`bool`) -- Whether or not to use FullyShardedDataParallel for training.
 * `--use_megatron_lm` (`bool`) -- Whether or not to use Megatron-LM for training.
-* `--use_xpu` (`bool`) -- Whether to use IPEX plugin to speed up training on XPU specifically.
```
muellerzr, line 161: Note we should deprecate this instead 2 versions of accelerate, we're preparing a release this week so even if it's "noop", we should still support the argument passing for 2 versions and then drop it. That'd be 1.8.0. Can we do so? Thanks!

SunMarc: + make sure to add a deprecation msg

yao-matrix: sure, make a lot of sense. update per comments, pls let me know for any changes need

yao-matrix:
@muellerzr @SunMarc, updated per your comments. Pls help review whether any other changes needed.

SunMarc (approved):
Thanks !

PR #3465, opened 2025-03-28
Fix: clip grad norm in fsdp2

S1ro1:
# What does this PR do?

Fixes a small bug in `clip_grad_norm` that was introduced in #3394.

cc @muellerzr


SunMarc (approved):
Thanks !

Issue #3322, opened 2025-01-05
support customized dataLoaders?


ifiaposto:
DataLoaderAdapter, and as a result DataLoaderShard returned by prepare, sets:

 self.base_dataloader = DataLoader(dataset, batch_sampler=batch_sampler, **kwargs)

I think it would be very useful if prepare_dataloader passed the type of the original dataloader when creating the DataLoaderShard and extend DataloaderShard to support customized dataloader types. Otherwise, the functionality seems very limited.

Issue #3014, opened 2024-08-14
batch sampler vs sampler?


hjc3613:
![image](https://github.com/user-attachments/assets/6c337dea-0993-4f41-95de-085a1148602b)
as shown in the picture, when **sampler_is_batch_sampler**, why pass the **original sampler** to **sampler argument** instead of **batch_sampler argument**? I do not understand. I think this is not consistent with the original dataloader

hjc3613:
I wander whether is it needed to pass dataloader to accelerator.prepare since I have write the distributed sampler aready, according rank, which set by: rank = dist.get_rank()

Issue #2091, opened 2023-10-27
Wrong judgement of batch_sampler in prepare_data_loader


XuHwang:
### System Info

```Shell
- `Accelerate` version: 0.24.0
- Platform: Linux-4.15.0-213-generic-x86_64-with-glibc2.27
- Python version: 3.9.0
- Numpy version: 1.26.1
- PyTorch version (GPU?): 1.13.1+cu117 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- System RAM: 503.79 GB
- GPU type: Tesla V100-PCIE-32GB
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I defined my own batch sampler, aiming to output a batch of indices each time. It works well when it's passed as the `batch_sampler` argument to `torch.util.data.DataLoader`. Meanwhile, it functions after wrapping the dataloader with `accelerate.data_loader.prepare_data_loader()` for the version `accelerate==0.20.3`. But now I upgrade the `accelerate` to the latest version `0.24.0`, it raises an `Attribute Error: 'MySampler' object has no attribute 'sampler'`. The code snippets are pasted below. 

I have compared the source code of `accelerate/data_loader.py` in version 0.20.3 and 0.24.0. I found the main difference to this bug is the `sampler_is_batch_sampler` variable in `prepare_data_loade()` function. In line 718 in version 0.20.3,  `sampler_is_batch_sampler` is set to `False`, while it is set as `sampler_is_batch_sampler = isinstance(dataloader.sampler, BatchSampler)` in line 834 in version 0.24.0. The condition may be not right in my case, where `sampler`
 of dataloader is `None` and `batch_sampler` is set to my own sampler (no member `sampler` in the batch sampler)  instead.

```python
import accelerate
import numpy as np
from torch.utils.data import Sampler, DataLoader, Dataset

class MyDataset(Dataset):
    def __init__(self, data) -> None:
        super().__init__()
        self.data = data
    
    def __len__(self) -> int:
        return len(self.data)

    def __getitem__(self, index):
        return self.data[index]


class MySampler:
    """
[...]

PR #3469, opened 2025-03-31
(fix) remove sampler_is_batch_sampler code in prepare_data_loader(..)

suzyahyah:
### What does this PR do?

This PR fixes various confusions wrt to `torch.utils.data.DataLoader`, `torch.utils.data.BatchSampler`, and `accelerate/data_loader.py:: prepare_data_loader(..)`.

~~https://github.com/huggingface/accelerate/issues/3322~~ *Edit, not quite this*
https://github.com/huggingface/accelerate/issues/3014
https://github.com/huggingface/accelerate/issues/2091

### Motivation

1. `accelerate/data_loader.py` has various patches allowing for a `BatchSampler` object to be passed as an argument to `sampler`. 

2. In the code, it allows this behavior using `sampler_is_batch_sampler = isinstance(dataloader.sampler, BatchSampler)`.

3. Allowing this is unintuitive for developers as it directly conflicts with `torch.utils.data.DataLoader` documentation, when accelerate should only wrap the pytorch dataloader and not change/allow different logic or arguments. 

4. It also permits various unintended behavior from other libraries relying on accelerate. For instance, when passing a custom BatchSampler to the dataloader argument in HuggingFace Trainer, the `sampler` kwargs is currently used, regardless of whether the sampler is BatchSampler or just RandomSampler. https://github.com/huggingface/transformers/blob/3b07ca78bb696825feee3e976795fab58f2b6d0c/src/transformers/trainer.py#L1026 (I'll be making a separate PR in HuggingFace on this) 

5. I believe this is due to a misunderstanding stemming from this Issue #https://github.com/huggingface/accelerate/issues/679. As @sgugger  initially suspected, this is probably a typo or misunderstanding from HuggingFace. 

This should not be allowed behavior, based on the following basic test case which will throw the datasets/formatting/formatting.py: `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'list'`

```
from torch.utils.data import BatchSampler, RandomSampler

ds = load_dataset("wikitext", "wikitext-103-raw-v1", split='validation')

train_dataloader = DataLoader(
    ds,
    sampler=BatchSampler(RandomSampler(dev_ds), batch_size=32, drop_last=False),
    num_workers=0,  # Adjust based on your setup
    pin_memory=True,
)

for batch in train_dataloader:
  print(batch)
```

[Official Pytorch Documentation](https://pytorch.org/docs/stable/data.html) for reference:
```
DataLoader(dataset, batch_size=1, shuffle=False, sampler=None, 
           batch_sampler=None, num_workers=0, collate_fn=None,
           pin_memory=False, drop_last=False, timeout=0,
           worker_init_fn=None, *, prefetch_factor=2,
           persistent_workers=False)
```

### This PR

1. Reverses this historical PR: https://github.com/huggingface/accelerate/pull/687 which supports the wrong logic.
2. Removes all traces of `sampler_is_batch_sampler`, simplifying the code
3. ~~Immediately throws an error~~  *Edit: Throws a warning* if a `BatchSampler` had been passed as an argument to `sampler`

### Tests

Passes all tests in tests/test_data_loader.py, but does not introduce any new tests. Open to suggestions.

### Considerations

1. The PR Immediately throws an error if a BatchSampler had been passed as an argument to Sampler. Technically this error should be thrown earlier or in `torch.utils.data.Sampler` but we can make it explicit that the problem is not coming from the Accelerate Library, since it had previously been allowed.   

2. Upon reading the [torch.utils.data source code](https://github.com/pytorch/pytorch/blob/v2.6.0/torch/utils/data/dataloader.py) v2.6.0, I figured that `torch.utils.data.DataLoader` will attempt to construct a BatchSampler from a Sampler if `batch_sampler=None`, and will also construct a default sampler if `sampler=None`. 

This means, we should always be able to recover a `dataloader.batch_sampler` when wrapping accelerate around an already constructed pytorch `dataloader`, and there is no need to check whether `sampler_is_batch_sampler` again, if the intent is just to get "`new_batch_sampler`". 

### Before submitting
- [ Yes ] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ No ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [NA; no docs mention batch_sampler ] Did you make sure to update the documentation with your changes? Here are the
- [No, open to suggestion ] Did you write any new necessary tests?

### Who can review?
@SunMarc @BenjaminBossan  @zach-huggingface @muellerzr

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -19,7 +19,7 @@ from typing import Callable, Optional, Union
 
 import torch
 from packaging import version
-from torch.utils.data import BatchSampler, DataLoader, IterableDataset, RandomSampler
+from torch.utils.data import BatchSampler, DataLoader, IterableDataset, RandomSampler, Sampler
 
 from .logging import get_logger
 from .state import DistributedType, GradientState, PartialState, is_torch_xla_available
@@ -631,13 +631,10 @@ class DataLoaderShard(DataLoaderAdapter, DataLoaderStateMixin):
         return get_sampler(self)
 
     def set_sampler(self, sampler):
-        sampler_is_batch_sampler = isinstance(self.sampler, BatchSampler)
-        if sampler_is_batch_sampler:
+        if isinstance(sampler, BatchSampler):
+            self.sampler.batch_sampler = sampler
+        elif isinstance(sampler, Sampler):
             self.sampler.sampler = sampler
-        else:
-            self.batch_sampler.sampler = sampler
-            if hasattr(self.batch_sampler, "batch_sampler"):
-                self.batch_sampler.batch_sampler.sampler = sampler
 
 
 if is_torch_xla_available():
@@ -958,13 +955,12 @@ class DataLoaderDispatcher(DataLoaderAdapter, DataLoaderStateMixin):
         return get_sampler(self)
 
     def set_sampler(self, sampler):
-        sampler_is_batch_sampler = isinstance(self.sampler, BatchSampler)
-        if sampler_is_batch_sampler:
+        if isinstance(sampler, BatchSampler):
+            self.sampler.batch_sampler = sampler
+        elif isinstance(sampler, Sampler):
             self.sampler.sampler = sampler
         else:
-            self.batch_sampler.sampler = sampler
-            if hasattr(self.batch_sampler, "batch_sampler"):
-                self.batch_sampler.batch_sampler.sampler = sampler
+            raise ValueError(f"{sampler} must be of type torch.utills.data.Sampler or torch.utils.data.BatchSampler")
 
 
 def get_sampler(dataloader):
@@ -977,10 +973,8 @@ def get_sampler(dataloader):
     Returns:
         `torch.utils.data.Sampler`: The sampler associated to the dataloader
     """
-    sampler_is_batch_sampler = isinstance(dataloader.sampler, BatchSampler)
-    if sampler_is_batch_sampler:
-        sampler = getattr(dataloader.sampler, "sampler", None)
-    else:
+    sampler = getattr(dataloader.sampler, "sampler", None)
+    if not sampler:
         sampler = getattr(dataloader.batch_sampler, "sampler", None)
     return sampler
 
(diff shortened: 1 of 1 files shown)


src/accelerate/data_loader.py
```diff
@@ -1155,11 +1149,16 @@ def prepare_data_loader(
 
     new_dataset = dataloader.dataset
     # Iterable dataset doesn't like batch_sampler, but data_loader creates a default one for it
+    if isinstance(dataloader.sampler, BatchSampler):
+        raise ValueError(
+            "Should not pass a BatchSampler do dataloader sampler argument. As per pytorch>2.1.0 documentation, please pass this to sampler instead"
+        )
```
muellerzr, line 1155: Can you please show the exact paragraph/section of the pytorch docs that state this? You reference many issues, but I dont' see where you found this

suzyahyah: Thanks for the quick reply! Yes it’s referenced under motivation #5. 

“Official PyTorch Documentation for Reference” showing two arguments for sampler and batch_sampler,

And in the docs they write this is “mutually exclusive with sampler” 

`batch_sampler (Sampler or Iterable, optional) – like sampler, but returns a batch of indices at a time. Mutually exclusive with batch_size, shuffle, sampler, and drop_last.`

suzyahyah: If we pass a BatchSampler (yields a list of indices) as an argument to sampler, we will get a list of list of indices, because pytorch will convert whatever was passed in the sampler argument to “batch mode”.

SunMarc:
can you rebase again ? the diff doesn't look correct

suzyahyah:
Thanks, @SunMarc 

The tests were failing because I misunderstood and had wrongly changed the logic in the function  `src/accelerate.data_loader.py: set_sampler()`. _I have reverted the logic in that function back to main branch._

After rebase against main, the tests successfully run (w/o hardware accelerators): 

`make test_core`
`make test_prod`
`pytest tests/test_data_loader.py`
`make test`

Quality checks run:

`make style`
`make quality`

SunMarc (commented):
Thanks for your work ! I feel like it is better if we first do a deprecation message instead of breaking everything right now WDYT @suzyahyah ? This will help users to fix their code. Also, it would be nice to fix the HF docs, otherwise users will still use sampler arg by default

PR #3470, opened 2025-03-31
Update ruff target-version to py39 and apply more fixes

cyyever:
# What does this PR do?

Use Python 3.9 as base version for ruff.


SunMarc (approved):
Thanks !

PR #3471, opened 2025-04-01
Bump ruff to 0.11.2

cyyever:
# What does this PR do?
Bump ruff to 0.11.2 and format code.


SunMarc (approved):
LGTM. I don't mind switching to 11.2 since transformers moved to that version also.

PR #3475, opened 2025-04-02
use device agnostic torch.OutOfMemoryError from pytorch 2.5.0

yao-matrix:
From PyTorch 2.5.0, we don't need platform-dependent OutOfMemoryError API, there is a unified one `torch.OutOfMemoryError`.


yao-matrix:
@SunMarc @muellerzr , pls help review, thx.

SunMarc (approved):
Thanks !

PR #3484, opened 2025-04-04
Remove deprecated PyTorch/XLA APIs

zpcore:
# What does this PR do?

PyTorch/XLA has deprecated bunch of APIs (https://github.com/pytorch/xla/issues/8915) for the upcoming 2.7 release. This PR integrates the change into accelerate package.

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -22,6 +22,7 @@ from .utils import DistributedType, honor_type, is_lomo_available, is_torch_xla_
 
 if is_torch_xla_available():
     import torch_xla.core.xla_model as xm
+    import torch_xla.runtime as xr
 
 
 def move_to_device(state, device):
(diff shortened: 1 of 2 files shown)


zpcore:
cc @SunMarc see if anyone can take a review. Thanks

src/accelerate/optimizer.py
```diff
@@ -22,6 +22,7 @@
 
 if is_torch_xla_available():
     import torch_xla.core.xla_model as xm
+    import torch_xla.runtime as xr
```
SunMarc, line 25: do we require a specific version of `torch_xla` to use xr.world_size() or xr.global_ordinal() ?

SunMarc (approved):
Thanks for the updates ! Just a small question

jeffhataws:
Hi @zpcore  @SunMarc can you also update https://github.com/huggingface/accelerate/blob/main/src/accelerate/state.py#L284 ?

SunMarc:
What needs to be changed @jeffhataws ?

Issue #3489, opened 2025-04-07
Accelerate Config produces broken FP8 config for DeepSpeed


AdamCuellar:
### System Info

```Shell
- `Accelerate` version: 1.6.0
- Platform: Linux-4.18.0-477.27.1.el8_8.x86_64-x86_64-with-glibc2.28
- `accelerate` bash location: /home/acuellar/miniforge-pypy3/envs/mllama/bin/accelerate
- Python version: 3.11.10
- Numpy version: 2.2.4
- PyTorch version (GPU?): 2.6.0+cu124 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- PyTorch SDAA available: False
- PyTorch MUSA available: False
- System RAM: 2015.39 GB
- GPU type: NVIDIA H100 80GB HBM3
- `Accelerate` default config:
        Not found
```

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

Steps to reproduce behavior:

1. Create config: accelerate config --config_file zero2_cpu_fp8_config.yaml
    Use default value for question: "Do you want to to execute `fprop`, `dgrad`, and `wgrad` GEMMS in higher precision? [yes/NO]: "
                                                                             
2. Launch training script: OMP_NUM_THREADS=8 accelerate launch --config_file zero2_cpu_fp8_config.yaml --num_processes 4 train.py

3. Get error:

  File "/home/acuellar/miniforge-pypy3/envs/mllama/lib/python3.11/site-packages/accelerate/utils/launch.py", line 90, in setup_fp8_env
    current_env[prefix + "FP8_OVERRIDE_FPROP"] = value[0]
                                                 ~~~~~^^^
TypeError: 'bool' object is not subscriptable

4. Config has:

fp8_config:
  amax_compute_algorithm: max
  amax_history_length: 1024
  backend: TE
  fp8_format: HYBRID
  interval: 1
  margin: 0
  override_linear_precision: false
  use_autocast_during_eval: false

5. Fix issue by changing override_linear_precision: false -> override_linear_precision: [false, false, false]

### Expected behavior

Fix issue by changing override_linear_precision: false -> override_linear_precision: [false, false, false]

[...]

PR #3492, opened 2025-04-08
Fix fp8 deepspeed config

SunMarc:
# What does this PR do?
Fixes https://github.com/huggingface/accelerate/issues/3489 
`override_linear_precision` is incorrectly set when using `accelerate config`


S1ro1 (approved):
LGTM

PR #3497, opened 2025-04-10
fix: apply torchfix to set `weights_only=True`

bzhong-solink:
Since `accelerate` depends on PyTorch >= 2.0 at the minimum, this was introduced in 2.0 which is backward compatible since then. 🎉

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -256,7 +256,7 @@ def load(f, map_location=None, **kwargs):
             torch.serialization.add_safe_globals(TORCH_SAFE_GLOBALS)
         else:
             kwargs.pop("weights_only", None)
-        loaded_obj = torch.load(f, map_location=map_location, **kwargs, weights_only=True)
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/other.py
```diff
@@ -256,7 +256,7 @@ def load(f, map_location=None, **kwargs):
             torch.serialization.add_safe_globals(TORCH_SAFE_GLOBALS)
         else:
             kwargs.pop("weights_only", None)
-        loaded_obj = torch.load(f, map_location=map_location, **kwargs)
+        loaded_obj = torch.load(f, map_location=map_location, **kwargs, weights_only=True)
```
BenjaminBossan, line 259: Some tests are failing with

> TypeError: torch.serialization.load() got multiple values for keyword argument 'weights_only'

Probably, `weights_only` needs to be added to `kwargs` here.

SunMarc (approved):
Thanks !

PR #3501, opened 2025-04-11
add support for port 0 auto-selection in multi-GPU environments

hellobiondi:
# What does this PR do?

This PR implements support for port 0 auto-selection in multi-GPU environments (`prepare_multi_gpu_env()`). The documentation already mentions that setting port to 0 will automatically select the next available port, but this functionality wasn't actually implemented in the code.

When `main_process_port` is set to `0`, the code now:
- Automatically finds an available port through socket binding
- Updates relevant arguments (`master_port`, `rdzv_endpoint`) with the selected port
- Provides a more seamless experience for users working in environments where specific ports might be occupied

## Before submitting
- [x] Did you read the [[contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr)](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr), Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [[forum](https://discuss.huggingface.co/)](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Here are the [[documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs)](https://github.com/huggingface/accelerate/tree/main/docs), and [[here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification)](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [x] Did you write any new necessary tests? `tests/test_launch.py`

## Who can review?

@SunMarc @zach-huggingface - This relates to the Command Line Interface and distributed training functionality.

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -195,6 +195,13 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> dict[str, str]:
     """
     Prepares and returns an environment with the correct multi-GPU environment variables.
     """
+    # get free port and update configurations
+    if args.main_process_port == 0:
+        args.main_process_port = get_free_port()
+
+    elif args.main_process_port is None:
+        args.main_process_port = 29500
+
     num_processes = args.num_processes
     num_machines = args.num_machines
     main_process_ip = args.main_process_ip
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/launch.py
```diff
@@ -219,6 +219,18 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> dict[str, str]:
...
     # for some reasons like splitting log files.
     need_port_check = num_machines <= 1 or int(args.machine_rank) == 0
+
+    # get free port and update configurations
+    if need_port_check and main_process_port == 0:
+        main_process_port = get_free_port()
+        if num_machines > 1 and getattr(args, "same_network", False):
+            args.master_port = str(main_process_port)
+        elif num_machines <= 1:
+            args.master_port = str(main_process_port)
```
BenjaminBossan, line 229: These two lines assign the same value, is this intended?

hellobiondi: the lines assign the same value to mimic the structure of that block in lines 202 - 214, but i can change it to be more concise!

also, i notice that these changes might be inconsequential since in lines 216 - 217, main_process_port is updated after the previous arguments were already assigned prior.

thanks for reviewing 🥇

BenjaminBossan: > but i can change it to be more concise!

Yes, I think that would clarify the code.

> also, i notice that these changes might be inconsequential since in lines 216 - 217, main_process_port is updated after the previous arguments were already assigned prior.

Do you mean that your newly added snippet should be moved up?

SunMarc:
friendly ping @hellobiondi

SunMarc (approved):
Thanks ! One last thing that would be nice is to confirm that it actually works when training a model, do you think you can quickly test that ?

PR #3503, opened 2025-04-14
Fix deepspeed tests

S1ro1:
This test fails on main: `tests/deepspeed/test_deepspeested.yy::DeepSpeedConfigIntegration::test_basic_run` because of #3457 where test `test_performance` script was changed to include `TorchTensorParallelPlugin`. 

This PR changes this to include this plugin only if `args.tp_size` is specified. Also included a sticter check in the `TorchTensorParallelPlugin` as the previous error was slightly cryptic (we passed in None)

cc @kmehant

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -2034,8 +2034,11 @@ class TorchTensorParallelPlugin:
     torch_device_mesh: Optional["torch.distributed.DeviceMesh"] = field(default=None)
 
     def __post_init__(self):
-        if self.tp_size == 1:
-            raise ValueError("Provide TP degree > 1.")
+        if not isinstance(self.tp_size, int):
+            raise ValueError(f"`tp_size` set to {self.tp_size}, please set to an `int`.")
+
+        if self.tp_size <= 1:
+            raise ValueError("`tp_size` must be greater than 1.")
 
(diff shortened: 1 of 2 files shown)


kmehant (commented):
@S1ro1 
Ah right, didnt notice, that this file is being used by other tests. Thanks for the catch!

can you as well change this part of the code? 

https://github.com/huggingface/accelerate/pull/3503/files#diff-0415726aa66758c678c2aa714624ffe64e0ff70a3c30212891741184bae3a370R101-R103

kmehant (approved):
Code looks organized now, thanks for the changes!

src/accelerate/utils/dataclasses.py
```diff
@@ -2034,7 +2034,8 @@ class TorchTensorParallelPlugin:
     torch_device_mesh: Optional["torch.distributed.DeviceMesh"] = field(default=None)
 
     def __post_init__(self):
-        if self.tp_size == 1:
+        # we don't load from env variable anymore so we check for None
+        if self.tp_size is None or self.tp_size <= 1:
```
SunMarc, line 2038: not sure what you mean wit the env variable. Also maybe we should check for `int` type and raise an error.

SunMarc (approved):
Thanks a lot ! Just a minor nit

PR #3504, opened 2025-04-15
Fix: require transformers version for tp tests

S1ro1:
With #3457 this introduces another bug in our CI, where it fails on nightly runners for `TPIntegrationPlugin` on `ValueError("... requires transformers version 4.52.0...")` due to the version of `transformers` on that runner being old (viz [here](https://github.com/huggingface/accelerate/actions/runs/14459958612)). I believe this fail also introduces further unrelated fail, so this should be handled.


S1ro1:
Failing tests seem to hub related.

SunMarc (approved):
LGTM

PR #3514, opened 2025-04-18
Add torchao to FP8 error message

jphme:
# What does this PR do?

When trying to get FP8 training to work, i was confused by the error message (as the docs say that FP8 via torchao is supported as well). 
`is_fp8_available` tests for all 3 FP8 methods, so the error message should also state all FP8 options :). 

## Before submitting
- [X ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

@muellerzr @zach-huggingface


SunMarc (approved):
Indeed, thanks !

jphme:
@SunMarc can you trigger a rerun failed? 
`OSError: We couldn't connect to 'https://huggingface.co/' to load the files, and couldn't find them in the cached files.` probably not related to this string change ;).

PR #3517, opened 2025-04-18
Fix logic in `accelerator.prepare` + IPEX for 2+ `nn.Models` and/or `optim.Optimizers`

### LINKED ISSUE #3516: `accelerator.prepare` with IPEX incorrectly handles 2+ `nn.Module` and/or `optim.Optimizer` objects
### System Info

```Shell
- `Accelerate` version: 1.7.0.dev0
- Platform: Linux-6.8.0-49-generic-x86_64-with-glibc2.35
- `accelerate` bash location: .venv/bin/accelerate
- Python version: 3.11.10
- Numpy version: 2.1.2
- PyTorch version (GPU?): 2.6.0+cpu (False)
- PyTorch SDAA available: False
- System RAM: 755.50 GB
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: True
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - rdzv_backend: static
        - same_network: True
        - enable_cpu_affinity: False
        - ipex_config: {'ipex': True}
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

To reproduce, first install PyTorch and IPEX according to https://pytorch-extension.intel.com/installation?platform=cpu&version=v2.6.0%2Bcpu&os=linux%2Fwsl2&package=pip.

Then, install the latest `accelerate/main`.

Then, launch the following script using `accelerate launch --ipex`:

```python
import torch

accelerator = Accelerator()

class BigModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.a = torch.nn.Parameter(torch.randn(()))
        self.b = torch.nn.Parameter(torch.randn(()))
        self.c = torch.nn.Parameter(torch.randn(()))

    def forward(self, x):
        return self.a + self.b * x + self.c * x ** 2
    
class SmallModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.a = torch.nn.Parameter(torch.randn(()))

    def forward(self, x):
        return self.a * x

big_model = BigModel()
small_model = SmallModel()

print(f"Before accelerator.prepare\n{type(big_model) = }\n{type(small_model) = }\n")

# Prepare models
big_model, small_model = accelerator.prepare(big_model, small_model)

print(f"After accelerator.prepare\n{type(big_model) = }\n{type(small_model) = }")
```

### Expected behavior

The output will be:
```
Before accelerator.prepare
type(big_model) = <class '__main__.BigModel'>
type(small_model) = <class '__main__.SmallModel'>
(big_model is small_model) = False

After accelerator.prepare
type(big_model) = <class '__main__.SmallModel'>
type(small_model) = <class '__main__.SmallModel'>
(big_model is small_model) = True
```

This is a silent error: `big_model` is forever lost and has been overridden by `small_model`. The root cause is a logical error in [`accelerator._prepare_ipex`](https://github.com/huggingface/accelerate/blob/423fbbfdea0eda4e5b2db7bad2194da293dc69f9/src/accelerate/accelerator.py#L2173).

Only the last `nn.Module` is captured as `model` (same issue for the optimizer):
https://github.com/huggingface/accelerate/blob/423fbbfdea0eda4e5b2db7bad2194da293dc69f9/src/accelerate/accelerator.py#L2189-L2194

After which it gets assigned in place of all input `nn.Module` instances:
https://github.com/huggingface/accelerate/blob/423fbbfdea0eda4e5b2db7bad2194da293dc69f9/src/accelerate/accelerator.py#L2204-L2208

The bug is not present when the `--ipex` flag is not used.

mariusarvinte:
# What does this PR do?

Fixes #3516.

Fixes the logic in `accelerator._prepare_ipex` such that:
- Multiple `nn.Module` objects can now be prepared in parallel, without silent failures.
- An error is raised if a user attempts to prepare a combination that is non-trivial to infer (e.g., two modules and two optimizers at the same time).
- A cautionary note and a suggested workaround for this limitation (due to `intel-extension-for-pytorch`) is added to the documentation.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@yao-matrix @BenjaminBossan @SunMarc @zach-huggingface

### DIFF
--- a/docs/source/usage_guides/ipex.md
+++ b/docs/source/usage_guides/ipex.md
@@ -95,7 +95,7 @@ accelerate launch examples/nlp_example.py
 ```
 
 > [!CAUTION]
-> `accelerator.prepare` can currently only handle simultaneously preparing multiple models for  (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.
+> `accelerator.prepare` can currently only handle simultaneously preparing multiple models  (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.
 
 **Scenario 2**: Acceleration of distributed CPU training
 we use Intel oneCCL for communication, combined with Intel® MPI library to deliver flexible, efficient, scalable cluster messaging on Intel® architecture. you could refer the [here](https://huggingface.co/docs/transformers/perf_train_cpu_many) for the installation guide
--- a/docs/source/usage_guides/ipex.md
+++ b/docs/source/usage_guides/ipex.md
@@ -95,7 +95,7 @@ accelerate launch examples/nlp_example.py
 ```
 
 > [!CAUTION]
-> `accelerator.prepare` can currently only handle simultaneously preparing multiple models  (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.
+> `accelerator.prepare` can currently only handle simultaneously preparing multiple models (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.
 
 **Scenario 2**: Acceleration of distributed CPU training
 we use Intel oneCCL for communication, combined with Intel® MPI library to deliver flexible, efficient, scalable cluster messaging on Intel® architecture. you could refer the [here](https://huggingface.co/docs/transformers/perf_train_cpu_many) for the installation guide
--- a/docs/source/usage_guides/ipex.md
+++ b/docs/source/usage_guides/ipex.md
@@ -94,6 +94,9 @@ use_cpu: true
 accelerate launch examples/nlp_example.py
 ```
 
+> [!CAUTION]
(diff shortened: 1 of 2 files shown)


docs/source/usage_guides/ipex.md
```diff
@@ -94,6 +94,9 @@ use_cpu: true
 accelerate launch examples/nlp_example.py
 ```
 
+> [!CAUTION]
+> `accelerator.prepare` can currently only handle simultaneously preparing multiple models for  (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.
+
```
SunMarc, line 99:
```suggestion
> `accelerator.prepare` can currently only handle simultaneously preparing multiple models  (and no optimizer) OR a single model-optimizer pair for training. Other attempts (e.g., two model-optimizer pairs) will raise a verbose error. To work around this limitation, consider separately using `accelerator.prepare` for each model-optimizer pair.

```

src/accelerate/accelerator.py
```diff
@@ -2183,29 +2183,52 @@ def _prepare_ipex(self, *args):
                     " to https://github.com/intel/intel-extension-for-pytorch."
                 )
 
-        model = None
-        optimizer = None
+        # ipex.optimize() is available only for IPEX, both IPEX-CPU and IPEX-XPU
+        if is_ipex_available():
+            import intel_extension_for_pytorch as ipex
```
SunMarc, line 2188: Let's remove the `if self.state.use_ipex:` condition above, this function is always called when `use_ipex` is `True`. Can you also combine both logic in a if else condition ?

SunMarc (approved):
Thanks a lot ! Just a nit

src/accelerate/accelerator.py
mariusarvinte: Thanks, implemented

SunMarc (approved):
Thanks !

PR #3528, opened 2025-04-23
[FSDP2] Issues in Wrap Policy and Mixed Precision

jhliu17:
# What does this PR do?

This PR fixes three issues in FSDP2's wrap policy and mixed precision preparations.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @zach-huggingface @S1ro1

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -693,7 +693,7 @@ def fsdp2_prepare_auto_wrap_policy(
             The auto wrap policy function to be applied to the model
     """
     if auto_wrap_policy_type == "transformer":
-        no_split_modules = model._no_split_modules
+        no_split_modules = getattr(model, "_no_split_modules", None)
         if no_split_modules is None:
             no_split_modules = []
         transformer_cls_names_to_wrap = list(no_split_modules)
@@ -715,7 +715,8 @@ def fsdp2_prepare_auto_wrap_policy(
     elif auto_wrap_policy_type == "size":
 
         def policy(module: torch.nn.Module) -> bool:
-            return module.numel() > fsdp2_plugin.min_num_params
+            module_num_params = sum(p.numel() for p in module.parameters())
+            return module_num_params > fsdp2_plugin.min_num_params
     else:
         return None
 
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -665,7 +665,7 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
 
     # There is no `dtype` attribution for nn.Module
     # Set it to None if it doesn't exist and do the upcast always
-    model_dtype = getattr(model, 'dtype', None)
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -559,7 +559,7 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
         # Other ranks have an empty model on `meta` device, so we need to distribute the weights properly
         fsdp2_load_full_state_dict(accelerator, model, original_sd)
 
-    if accelerator.mixed_precision != "no" and model.dtype != torch.float32:
+    if accelerator.mixed_precision != "no":
```
S1ro1, line 562: There should be a similar check in-place as we only want to warn if we're really upcasting. Such as [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/accelerator.py#L1704-L1708)

S1ro1 (commented):
Thank you for the PR, left a small note, except of that it looks good to me!

src/accelerate/utils/fsdp_utils.py
SunMarc: what was the issue here ?

SunMarc (commented):
Thanks ! Just a nit

src/accelerate/utils/fsdp_utils.py
S1ro1: Vanilla PyTorch modules don't have a `dtype` attribute actually, the underlying parameters only have one.

SunMarc: oh yeah, i was thinking of transformers model this is why. Maybe we should be something similar to what we have in fdsp ?

jhliu17: Any good idea on how to access the dtype in model's parameters? For the strategy adopted in FSDP, what file should I look at? Thanks!

jhliu17: I have reverted the commit on `model.dtype != torch.float32`. I think the checking can be improved later so it keeps consistent with the check in FSDP. Thank you.

S1ro1: @jhliu17 Oh okay, fine by me. Though make it a `getattr` call, it was a mishap on my side, not realizing torch modules don't have dtype by default. We don't want it crashing, so just do `getattr`. I'll take a look later on how to make it work as FSDP1.

jhliu17: @S1ro1 great! I just committed a change that uses `getattr` to check whether `dtype` is existed. If `dtype` is not existed, the upcast operation would be always performed.

S1ro1 (approved):
Thanks a lot! Fix style/quality (`make style/quality`) and good to merge.

SunMarc:
There are some conflits are you need to fix before we merge @jhliu17

jhliu17:
> There are some conflits are you need to fix before we merge @jhliu17

Thank you. Just rebased the branch onto the upstream/main

S1ro1:
Thank you, merged!

PR #3069, opened 2024-09-03
support torch dynamo for deepspeed>=0.14.4

oraluben:
# What does this PR do?

Fixes # (issue)

Related: https://github.com/microsoft/DeepSpeed/pull/6502

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #3529, opened 2025-04-23
Dynamo regional compilation

IlyasMoutawwakil:
# What does this PR do?

An adapted recipe from https://pytorch.org/tutorials/recipes/regional_compilation.html
Instead of trying to compile the whole model, which usually has a big problem space for optimization. Regional compilation compiles the repeated block first (usually a decoder/encoder layer), so that the compiler would re-use its cached/optimized generated code for the other blocks, reducing the cold start compilation time observed on the first inference call.

Some reuslts on llama:

```
[-------------------------------------------------- NousResearch/Llama-3.2-1B ---------------------------------------------------]
                            |  Inference time (1x128)  |  Inference time (4x128)  |  Compile time (1x128)  |  Compile time (4x128)
1 threads: -----------------------------------------------------------------------------------------------------------------------
      Baseline              |           18.3           |           18.4           |                        |                      
      Full compilation      |            6.3           |           10.0           |        10696.4         |        10248.0       
      Regional compilation  |            9.7           |           10.0           |         1952.7         |         2903.9       

Times are in milliseconds (ms).

[---------------------------------------------- NousResearch/Hermes-3-Llama-3.2-3B ----------------------------------------------]
                            |  Inference time (1x128)  |  Inference time (4x128)  |  Compile time (1x128)  |  Compile time (4x128)
1 threads: -----------------------------------------------------------------------------------------------------------------------
      Baseline              |           33.4           |           33.6           |                        |                      
      Full compilation      |           11.2           |           23.9           |        17857.5         |        17736.5       
      Regional compilation  |           17.3           |           23.7           |         2993.2         |         2478.8       

Times are in milliseconds (ms).

[---------------------------------------------- NousResearch/Hermes-3-Llama-3.1-8B ----------------------------------------------]
                            |  Inference time (1x128)  |  Inference time (4x128)  |  Compile time (1x128)  |  Compile time (4x128)
1 threads: -----------------------------------------------------------------------------------------------------------------------
      Baseline              |           40.3           |           59.5           |                        |                      
      Full compilation      |           18.9           |           54.4           |        20437.8         |        20152.3       
      Regional compilation  |           19.7           |           54.0           |         2903.1         |         2438.0       

Times are in milliseconds (ms).

[--------------------------------------------- NousResearch/Nous-Hermes-Llama2-13b ----------------------------------------------]
                            |  Inference time (1x128)  |  Inference time (4x128)  |  Compile time (1x128)  |  Compile time (4x128)
1 threads: -----------------------------------------------------------------------------------------------------------------------
      Baseline              |           45.5           |          100.4           |                        |                      
      Full compilation      |           29.4           |           89.7           |        23099.4         |        22885.9       
      Regional compilation  |           29.4           |           87.5           |         2945.5         |         2526.2       

Times are in milliseconds (ms).
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/accelerate/accelerator.py
```diff
@@ -3833,3 +3835,28 @@ def fp8_backend(self):
...
+        if isinstance(module, torch.nn.ModuleList):
+            # Create a new ModuleList
+            new_modulelist = torch.nn.ModuleList()
+            for submodule in module:
+                new_modulelist.append(torch.compile(submodule, **compile_kwargs))
+            return new_modulelist
+        elif module._modules:  # Non-leaf node
+            # Create a new module of the same type
+            new_module = module.__class__.__new__(module.__class__)
+            new_module.__dict__.update(module.__dict__)
```
SunMarc, line 3855: not sure who robust this is. Is this how torch.compile works also ?

src/accelerate/accelerator.py
```diff
@@ -3833,3 +3835,28 @@ def fp8_backend(self):
...
+            new_module._modules = {}
+            # Process children
+            for name, submodule in module.named_children():
+                new_module.add_module(name, TorchDynamoPlugin.compile_regions(submodule, **compile_kwargs))
```
SunMarc, line 3859: I think you mean Accelerator no ? 
```suggestion
                new_module.add_module(name, Accelerator.compile_regions(submodule, **compile_kwargs))
```

SunMarc (commented):
Thanks for this ! Maybe worth to add a few tests and update the docs (with your benchmark) ?

src/accelerate/accelerator.py
IlyasMoutawwakil: Yes, do you think this should be in the accelerator or `TorchDynamoPlugin` ?

src/accelerate/accelerator.py
```diff
@@ -3833,3 +3835,28 @@ def fp8_backend(self):
         elif self.state.deepspeed_plugin is not None and self.state.deepspeed_plugin.enable_msamp:
             return "MSAMP"
         return None
+
+    @staticmethod
+    def compile_regions(module: torch.nn.Module, **compile_kwargs) -> torch.nn.Module:
+        """
+        Compiles regions of the model and returns a new instance of the model without copying parameters (like
+        `torch.compile`).
+        """
```
SunMarc, line 3844: update docstring for compile_kwargs

src/accelerate/accelerator.py
IlyasMoutawwakil: okay, I'll move it there (makes sense to have it close to `is_compiled_module`)

src/accelerate/accelerator.py
IlyasMoutawwakil: no, `torch.compile` simply keeps a reference to the original model in `_orid_mod` and uses a compiled `__call__/forward`. In this case we wanna continue to use the original `__call__/forward` and only compile some submodules, so It's a bit more complicated to do without creating a new instance.
This should be fairly robust as it's a generic way to create a new instance with references to the attributes instead of copies. It could fail but only in very niche cases (e.g. a class that overrides its __new__ method and adds some args/kwargs to it, like a tensor subclass), but I don't think anyone would do that with `torch.nn.Module`.

SunMarc:
Nice findings ! Good to see that it works nicely for big bs as it is crucial for training !

tests/test_compile.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2021 The HuggingFace Team. All rights reserved.
```
SunMarc, line 1:
```suggestion
# Copyright 2025 The HuggingFace Team. All rights reserved.
```

tests/test_compile.py
```diff
@@ -0,0 +1,101 @@
...
+COMPILE_ITERS = 2
+INFERENCE_ITERS = 100
+
+INFRENCE_STMT = "model(input_ids, use_cache=False)"
+COMPILE_STMT = f"torch._dynamo.reset(); torch._inductor.utils.clear_inductor_caches(); {INFRENCE_STMT}"
+
+
+@require_non_cpu
+@require_huggingface_suite
+class RegionalCompilationTester(unittest.TestCase):
```
SunMarc, line 34: Can you add one more test to show that we can unwrap a model that went through regional compilation ?

benchmarks/compile/README.md
```diff
@@ -0,0 +1,96 @@
+# Regional Compilation Benchmark
+
+This benchmark compares different compilation strategies using PyTorch's `torch.compile` and Accelerate's `compile_regions` utility, which is based on the recipe in [PyTorch documentation](https://pytorch.org/tutorials/recipes/regional_compilation.html).
+
+## Overview
+
```
SunMarc, line 6: Really nice docs ! I feel like this is a bit of a shame to keep it hidden in the benchmark docs, maybe we should start a new doc around compilation and link this benchmark in `https://huggingface.co/docs/accelerate/en/usage_guides/` under the training section ?

SunMarc (approved):
Really nice ! Thanks for iterating ! Just a few nits

IlyasMoutawwakil:
@SunMarc I added more docs and figures wdyt ?

benchmarks/compile/README.md
IlyasMoutawwakil: first time adding docs to accelerate, so tell me if what I did makes sense 😅

src/accelerate/utils/other.py
```diff
@@ -50,15 +50,74 @@
...
+    them sequentially to hit the compiler's cache. This allows us to speed up the compilation time / cold start of
+    models like LLMs and Transformers in general.
+
+    See: https://pytorch.org/tutorials/recipes/regional_compilation.html
+
+    Args:
+        module (`torch.nn.Module`):
+            The model to compile.
+        **compile_kwargs:
+            Additional keyword arguments to pass to `torch.compile()`.
```
sayakpaul, line 90: Could you provide an example here?

docs/source/usage_guides/compilation.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Initialize accelerator with the plugin
+accelerator = Accelerator(dynamo_plugin=dynamo_plugin)
+# This will apply torch.compile to your model
+model = accelerator.prepare(model)
+```
+
+## Regional Compilation
+
+Instead of trying to compile the whole model, which usually has a big problem space for optimization. Regional compilation finds and compiles a repeated block first (usually a decoder/encoder layer), so that the compiler would re-use its cached/optimized generated code for the other blocks, reducing the cold start compilation time observed on the first inference call.
```
sayakpaul, line 39: We could provide some concrete examples of what constitutes repeated blocks.

src/accelerate/utils/dataclasses.py
```diff
@@ -999,6 +999,18 @@ class TorchDynamoPlugin(KwargsHandler):
     options: Any = field(default=None, metadata={"help": "A dictionary of options to pass to the backend."})
     disable: bool = field(default=False, metadata={"help": "Turn torch.compile() into a no-op for testing"})
 
+    use_regional_compilation: bool = field(
+        default=False,
+        metadata={
+            "help": (
+                # https://pytorch.org/tutorials/recipes/regional_compilation.html
+                "This recipe to reduces the cold start compilation time of torch.compile() by choosing to "
```
sayakpaul, line 1007:
```suggestion
                "Use it to reduce the cold start compilation time of torch.compile() by choosing to "
```

tests/test_compile.py
```diff
@@ -0,0 +1,126 @@
...
+        distributed_compiled_model = compile_regions(distributed_model)
+        compiled_model_unwrapped = extract_model_from_parallel(distributed_compiled_model, keep_torch_compile=True)
+
+        assert compiled_model._orig_mod == compiled_model_unwrapped._orig_mod
+
+    def test_dynamo_extract_model_remove_torch_compile(self):
+        model, _ = self._get_model_and_inputs()
+        compiled_model = compile_regions(model)
+
+        # could also do a test with DistributedDataParallel, but difficult to run on CPU or single GPU
```
sayakpaul, line 59: Should the test be decorated with something that says it needs multiple GPUs? 👁️

tests/test_compile.py
```diff
@@ -0,0 +1,126 @@
...
+        from transformers import AutoConfig, AutoModelForCausalLM
+
+        with torch.device(torch_device):
+            config = AutoConfig.from_pretrained(MODEL_ID)
+            model = AutoModelForCausalLM.from_config(config)
+            input_ids = torch.randint(0, 1000, (4, 128), dtype=torch.int64)
+
+        return model, input_ids
+
+    def test_dynamo_extract_model_keep_torch_compile(self):
```
sayakpaul, line 44: Would it make sense to have a test that checks if all the modules under a `ModuleList` were compiled when doing regional compilation?

docs/source/usage_guides/compilation.md
```diff
@@ -0,0 +1,69 @@
...
+- **Ease of Use**: Requires minimal code changes to implement, making it highly accessible.
+- **Compatibility**: Works seamlessly with existing PyTorch code and models.
+
+When used with Accelerate, `torch.compile` integrates smoothly into distributed training workflows, allowing you to benefit from both distributed execution and compilation optimizations simultaneously.
+
+The first execution of compiled code typically takes longer as it includes the compilation time, but subsequent runs are significantly faster. For optimal performance in different scenarios, `torch.compile` offers various modes like `"default"`, `"reduce-overhead"` (which uses CUDA graphs to further reduce overhead), and `"max-autotune"` (which performs extensive autotuning to find the best kernels for your model).
+
+## Using `torch.compile` with Accelerate
+
+Accelerate provides `TorchDynamoPlugin` for easy integration with `torch.compile`.
```
SunMarc, line 17: maybe we can also write somewhere that this is also compatible with ddp, fsdp, [deepspeed](https://github.com/huggingface/accelerate/pull/3069). cc @S1ro1

docs/source/_toctree.yml
```diff
@@ -71,6 +71,8 @@
       title: Big Model Inference
     - local: usage_guides/distributed_inference
       title: Distributed inference
+    - local: usage_guides/compilation
+      title: Compilation
```
SunMarc, line 75: I think this should probably live in the training section instead of the inference one.

SunMarc (approved):
Thanks ! Just some nits

src/accelerate/utils/other.py
IlyasMoutawwakil: done !

docs/source/usage_guides/compilation.md
IlyasMoutawwakil: done !

sayakpaul (approved):
Thanks a bunch @IlyasMoutawwakil! I learned tons!

benchmarks/torch.compile/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+Full results are available in the tables below:
+
+```markdown
+[-------------------------------------------------- NousResearch/Llama-3.2-1B ---------------------------------------------------]
+                            |  Inference time (1x128)  |  Inference time (4x128)  |  Compile time (1x128)  |  Compile time (4x128)
+1 threads: -----------------------------------------------------------------------------------------------------------------------
+      Baseline              |           18.3           |           18.4           |                        |                      
+      Full compilation      |            6.3           |           10.0           |        10696.4         |        10248.0       
+      Regional compilation  |            9.7           |           10.0           |         1952.7         |         2903.9
```
anijain2305, line 54: Are you clearing your cache for measuring cold start compilation latency?

Cold start - Compiling the model for the very first time. This is first time ever compile saw this model.
Warm start - The model was compiled earlier (in a separate process) and therefore has compiled artifacts on the disk at `/tmp/torchinductor_$USER`. So, this time it will skip some parts of `torch.compile` stack because it can use the artifacts from disk. More info at https://docs.pytorch.org/tutorials/recipes/torch_compile_caching_tutorial.html


The reason I ask is because these numbers look way too good for cold start :)

It is possible that we really care about warm-start because many users would run the model on their local machines multiple times. Bt if we are talking about inference server where we are spinning new machines without any remote compiled artifacts, we will see higher compile time latency on the first run.

anijain2305: Note that regional compilation will reduce both cold and warm start.

src/accelerate/utils/other.py
```diff
@@ -50,15 +50,106 @@
...
+            else:
+                new_module = torch.compile(module, **compile_kwargs)
+        elif module._modules:  # Non-leaf node
+            new_module = module.__class__.__new__(module.__class__)
+            new_module.__dict__.update(module.__dict__)
+            new_module._modules = {}
+            for name, submodule in module.named_children():
+                new_module.add_module(name, _compile_regions(submodule, **compile_kwargs))
+        else:  # Leaf node
+            new_module = torch.compile(module, **compile_kwargs)
```
anijain2305, line 140: This seems little bit hacky. Another way to do this would be to inplace compile a nn.Module- https://github.com/pytorch/pytorch/blob/8b9c9a327f0eced63233674d04883d1b73ddc4d1/torch/nn/modules/module.py#L2998

Basically instead of doing

`new_mod = torch.compile(old_mod, ..)`

you can do

`old_mod.compile(...)`

This would remove the need of `OptimizedModule` inconvenience at other places as well.

anijain2305 (commented):
LGTM. A couple of comments. I think we should seriously look into inplace nn.module `compile`. This is not a very publicized version, but for this usecase it makes a lot of sense, and would clean up `compile_regions` significantly.

benchmarks/torch.compile/README.md
IlyasMoutawwakil: check the compilation statement (compilation_stmt)

src/accelerate/utils/other.py
anijain2305: Other comment is that recursively marking every nn.Module as regional compilation will not yield good results. We need to keep the regional compilation region fairly large to get a good compile-time-and-performance tradeoff. I suspect that the reason for large cliff on batch size 1 is because this code is marking too many children as regional compilation.

IlyasMoutawwakil: but the idea here is to have compile_regions work like an in-place replacement for torch.compile (which returns a new instance of the model in a stateless way)

IlyasMoutawwakil: an in-place version of compile_regions could be interesting but decompiling in that case will not be the simple compiled._orig_mod but would rather require another method for decompiling.

IlyasMoutawwakil: > keep the regional compilation region fairly large to get a good compile-time-and-performance tradeoff

I actually have wrote a variant that does exactly that (maximizing regions) but wasn't confident enough to include it in this PR, will benchmark it extensively and see if it's any better.

Issue #3524, opened 2025-04-22
'notebook_launcher' Now Broken in Kaggle Notebook Environment


jmurph1:
### System Info

```Shell
- `Accelerate` version: 1.6.0
- Platform: Linux-6.6.56+-x86_64-with-glibc2.35
- `accelerate` bash location: /usr/local/bin/accelerate
- Python version: 3.11.11
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.6.0+cu124 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- PyTorch SDAA available: False
- PyTorch MUSA available: False
- System RAM: 31.35 GB
- GPU type: Tesla T4
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

```python
from accelerate import notebook_launcher, Accelerator

# A completely dummy function doing nothing with CUDA/ML
def dummy_fn():
    import os
    accelerator = Accelerator()
    print(f"Dummy function started in process {os.environ.get('LOCAL_RANK', 0)}.")
    import time
    time.sleep(5)
    print(f"Dummy function finished in process {os.environ.get('LOCAL_RANK', 0)}.")

print("Calling notebook_launcher...")
try:
    notebook_launcher(dummy_fn, num_processes=2)
    print("notebook_launcher call finished successfully.")
except Exception as e:
    print(f"notebook_launcher failed: {e}")
    import traceback
    traceback.print_exc() # Print full traceback if it fails



Calling notebook_launcher...
Launching training on 2 GPUs.
E0422 05:07:49.013000 204 torch/distributed/elastic/multiprocessing/api.py:732] failed (exitcode: 1) local_rank: 0 (pid: 240) of fn: dummy_fn (start_method: fork)
E0422 05:07:49.013000 204 torch/distributed/elastic/multiprocessing/api.py:732] Traceback (most recent call last):
E0422 05:07:49.013000 204 torch/distributed/elastic/multiprocessing/api.py:732]   File "/usr/local/lib/python3.11/dist-packages/torch/distributed/elastic/multiprocessing/api.py", line 687, in _poll
[...]

PR #3530, opened 2025-04-23
Fix unwanted cuda init due to torchao

SunMarc:
# What does this PR do?

This PR protects better the import as it initializes cuda somehow, breaking the notebook launcher. This was not pick-up by the CI as it requires to install torchao

Fixes https://github.com/huggingface/accelerate/issues/3524
Thanks for the report @jmurph1

To reproduce with the following script or one of our tests:

`CUDA_VISIBLE_DEVICES=0,1 pytest tests/test_cli.py::AccelerateLauncherTester::test_notebook_launcher -s -vvvv`

```python 
from accelerate import notebook_launcher, Accelerator

# A completely dummy function doing nothing with CUDA/ML
def dummy_fn():
    import os
    accelerator = Accelerator()
    print(f"Dummy function started in process {os.environ.get('LOCAL_RANK', 0)}.")
    import time
    time.sleep(5)
    print(f"Dummy function finished in process {os.environ.get('LOCAL_RANK', 0)}.")

print("Calling notebook_launcher...")
try:
    notebook_launcher(dummy_fn, num_processes=2)
    print("notebook_launcher call finished successfully.")
except Exception as e:
    print(f"notebook_launcher failed: {e}")
    import traceback
    traceback.print_exc() # Print full traceback if it fails
```

### DIFF
--- a/src/accelerate/utils/ao.py
+++ b/src/accelerate/utils/ao.py
@@ -17,15 +17,16 @@ Needed utilities for torchao FP8 training.
 """
 
 from functools import partial
-from typing import Callable, Optional
+from typing import TYPE_CHECKING, Callable, Optional
 
 import torch
 
 from .imports import is_torchao_available, torchao_required
 
 
-if is_torchao_available():
-    from torchao.float8.float8_linear import Float8LinearConfig
+if TYPE_CHECKING:
+    if is_torchao_available():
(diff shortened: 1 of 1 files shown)


S1ro1 (approved):
LGTM. Thanks.

Issue #3539, opened 2025-04-29
Error loading optimizer state with torch.load weights_only=True default in PyTorch 2.6


luiz0992:
### System Info

```Shell
- `Accelerate` version: 1.6.0
- Platform: Linux-5.10.0-34-cloud-amd64-x86_64-with-glibc2.31
- `accelerate` bash location: /home/luiz/delphos/training/.venv/bin/accelerate
- Python version: 3.12.4
- Numpy version: 1.26.4
- PyTorch version (GPU?): 2.6.0+cu124 (True)
- PyTorch XPU available: False
- PyTorch NPU available: False
- PyTorch MLU available: False
- PyTorch SDAA available: False
- PyTorch MUSA available: False
- System RAM: 1842.60 GB
- GPU type: NVIDIA H100 80GB HBM3
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: MULTI-GPU
        - mixed_precision: bf16
        - use_cpu: False
        - debug: False
        - num_processes: 8
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - enable_cpu_affinity: False
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

Error Message:

```
  [rank7]: _pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint. 
  [rank7]:        (1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will 
  likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
  [rank7]:        (2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
[...]

PR #3540, opened 2025-04-29
Add kwargs to optimizer, scheduler and dataloader using function `accelerator().load_state()`

luiz0992:
# What does this PR do?

 This PR fixes issue #3539
  
This PR adds support for passing keyword arguments to the load functions of optimizers, schedulers, and dataloaders when using the Accelerator's load_state method. This enhancement allows for more
  flexibility and control when loading optimizer, scheduler, and dataloader states.

  The changes include:
  1. Extending the load_state method signature to accept dedicated kwargs dictionaries for each component
  2. Passing these kwargs to the respective components' load functions
  3. Updating docstrings to document the new parameters
  

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

 @BenjaminBossan @SunMarc @zach-huggingface


SunMarc (commented):
Thanks ! Shared about another solution in the issue.

src/accelerate/checkpointing.py
```diff
@@ -200,6 +203,12 @@ def load_accelerator_state(
             An optional *GradScaler* instance to load
         map_location (`str`, *optional*):
             What device to load the optimizer state onto. Should be one of either "cpu" or "on_device".
+        optimizer_load_kwargs (`dict`, *optional*):
+            Additional arguments that can be passed to the optimizer's `load` function.
+        scheduler_load_kwargs (`dict`, *optional*):
+            Additional arguments that can be passed to the scheduler's `load` function.
+        dataloader_load_kwargs (`dict`, *optional*):
+            Additional arguments that can be passed to the dataloader's `load` function.
```
SunMarc, line 211: instead of create kwargs for each one of them, let's just have `load_kwargs` that we use for all load method in this function. Also it is better to set `load_kwargs` to None by default and in the function set it to dict() as it is mutable.

src/accelerate/checkpointing.py
```diff
@@ -261,7 +270,7 @@ def load_accelerator_state(
             dataloader_state_dict_name = "dl_state_dict.bin" if i == 0 else f"dl_state_dict_{i}.bin"
             input_dataloader_state_dict_file = input_dir.joinpath(dataloader_state_dict_name)
             if input_dataloader_state_dict_file.exists():
-                state_dict = load(input_dataloader_state_dict_file)
+                state_dict = load(input_dataloader_state_dict_file, map_location=None, **dataloader_load_kwargs)
                 dataloader.load_state_dict(state_dict)
     logger.info("All dataloader sampler states loaded successfully")
```
SunMarc, line 276: let's also include it for scaler and states

SunMarc (commented):
Thanks ! just some nits

SunMarc:
@luiz0992, let me know if you are planning to finish this pr or not

src/accelerate/checkpointing.py
luiz0992: I am not sure if I understood correct your suggestions. You want to use `load_kwargs`  for all load functions? I implemented that but I am not sure if this is the best way to go. On my case the error only happend with the optimizer load. Anyway, let me know what you think

810c69eed6a5759cecf80b70f31ceb39c15b854b

src/accelerate/checkpointing.py
luiz0992: 810c69eed6a5759cecf80b70f31ceb39c15b854b

src/accelerate/accelerator.py
```diff
@@ -3357,10 +3357,7 @@ def register_load_state_pre_hook(self, hook: Callable[..., None]) -> hooks.Remov
     def load_state(
         self,
         input_dir: str = None,
-        optimizer_load_kwargs: dict[str, Any] = {},
-        scheduler_load_kwargs: dict[str, Any] = {},
-        dataloader_load_kwargs: dict[str, Any] = {},
-        **load_model_func_kwargs,
```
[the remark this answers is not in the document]
luiz0992: The `load_model` does not accept kwargs. Let me know if I am mistaken.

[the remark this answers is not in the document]
SunMarc: in safetensors, we have the following
```python
def load_model(
    model: torch.nn.Module, filename: Union[str, os.PathLike], strict: bool = True, device: Union[str, int] = "cpu"
) -> Tuple[List[str], List[str]]:
```
If you can revert the changes related to `load_model_func_kwargs` and only update `load_kwargs` to where we use `load`, it will be better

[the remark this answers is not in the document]
luiz0992: The way it is being used it is missing only the `strict` argument, but that's ok. I reverted the `load_model_func_kwargs`.

c3d7f1fef1430c4a786dafd38941871175ac7e6c

SunMarc (approved):
Thanks !

src/accelerate/checkpointing.py
```diff
@@ -180,6 +180,7 @@ def load_accelerator_state(
     process_index,
     scaler=None,
     map_location=None,
+    load_kwargs=None,
```
SunMarc, line 183: just make sure to init a dict afterwards, toherwise, it will raise an error: 

```python 
if load_kwargs is None:
        load_kwargs = {}
```

SunMarc (commented):
Just a small nit

src/accelerate/checkpointing.py
luiz0992: e54694ce06cc74716a375f33f72436e40a4a9a64

PR #3544, opened 2025-05-01
set backend correctly for CUDA+FSDP2+cpu-offload

winglian:
# What does this PR do?

with FSDP2, fsdp_offload_params: true doesn't work and gives this error https://gist.github.com/winglian/700e5ab9021911f6d1c095b4dbdec5e5

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@S1ro1 

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -214,6 +214,8 @@ class PartialState:
                             if self.backend == "tccl":
                                 local_rank = os.environ.get("LOCAL_RANK", -1)
                                 torch.sdaa.set_device(f"sdaa:{local_rank}")
+                            if self.backend == "nccl" and os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" and os.environ.get("FSDP_VERSION", "1") == "2":
+                                self.backend = "cuda:nccl,cpu:gloo"
                             dist.init_distributed(dist_backend=self.backend, auto_mpi_discovery=False, **kwargs)
                         # We need to flag to `use_deepspeed` to be True to override `distributed_type` later
                         use_deepspeed = True


src/accelerate/state.py
```diff
@@ -214,6 +214,8 @@ def __init__(self, cpu: bool = False, **kwargs):
                             if self.backend == "tccl":
                                 local_rank = os.environ.get("LOCAL_RANK", -1)
                                 torch.sdaa.set_device(f"sdaa:{local_rank}")
+                            if self.backend == "nccl" and os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" and os.environ.get("FSDP_VERSION", "1") == "2":
```
joecummings, line 217: Where does this check for CPU offload? If you aren't offloading, you don't need to initialize a GLOO backend as well as the NCCL one.

winglian: @joecummings does it hurt anything to leave cpu:gloo in if not offloading?
@S1ro1 I can also add another check for `FSDP_OFFLOAD_PARAMS=="true"`, lmk

S1ro1: Can do that, it won't cover all the cases but at least most of them.

joecummings: Heard from Tristan (resident GLOO expert) that you can probably expect _some_ overhead in terms of time if initializing a GLOO backend for CPU as well as a NCCL backend. I don't have hard numbers here, however.

FWIW, I believe the only two operations you need to check for are 1) FSDP CPU Offload and 2) Async checkpointing b/c these are the only two that utilize a separate process group in order to not block comms  as much as possible. You could have just a flag that checks for parameters that complete cpu operations in the background and then check for this in the setting of the backend. It would be similar to how we do it in torchtune: https://github.com/pytorch/torchtune/blob/5d51c25cedfb6ba7b00e03cb2fef4f9cdb7baebd/recipes/full_finetune_distributed.py#L142

SunMarc:
Superseded by https://github.com/huggingface/accelerate/pull/3574

PR #3545, opened 2025-05-02
[FSDP2] Fix: "..." is not a buffer or a paremeter

S1ro1:
Sometimes, the infer logic fails when we use LORA, this fixes it

See mentioned PR in trl for *test*


SunMarc (approved):
LGTM !

PR #3561, opened 2025-05-13
[docs] update deepspeed config path

faaany:
## What does this PR do?
The deepspeed config path in the deepspeed doc is not up-to-date. This PR updates it.


faaany:
cc @SunMarc

PR #3562, opened 2025-05-13
simplify model.to logic

yao-matrix:
@SunMarc @faaany , pls help review, thx.

### DIFF
--- a/src/accelerate/utils/bnb.py
+++ b/src/accelerate/utils/bnb.py
@@ -143,12 +143,13 @@ def load_and_quantize_model(
                         param.to(torch.float32)
             elif torch.is_floating_point(param):
                 param.to(dtype)
-        if model_device.type == "cuda" or torch.cuda.is_available():
-            model.to(torch.cuda.current_device())
+        if model_device.type == "cuda":
+            model.cuda(torch.cuda.current_device())
             torch.cuda.empty_cache()
+        elif torch.cuda.is_available():
+            model.to(torch.cuda.current_device())
         elif torch.xpu.is_available():
             model.to(torch.xpu.current_device())
-            torch.xpu.empty_cache()
         else:
             raise RuntimeError("No GPU or Intel XPU found. A GPU or Intel XPU is needed for quantization.")
         logger.info(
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/bnb.py
```diff
@@ -143,19 +143,17 @@ def load_and_quantize_model(
                         param.to(torch.float32)
             elif torch.is_floating_point(param):
                 param.to(dtype)
-        if model_device.type == "cuda":
-            # move everything to cpu in the first place because we can't do quantization if the weights are already on cuda
```
yao-matrix, line 147: this comment is quite confusing, the code move the model to cuda, but the comment says move the model to cpu, @SunMarc , maybe you can educate me the background.

SunMarc: This is probably some comments I forgot to delete. Can you keep the code as it as it behaves slightly differently (`cuda` vs `to` in bnb). When calling to, we check if we are on cpu, otherwise, we don't quantize the weights.

SunMarc (commented):
Thanks, can you just revert some of the changes ?

src/accelerate/utils/bnb.py
yao-matrix: aha, thx for educating me the background, good to learn. And could you also educate me why you use `empty_cache` in `model_device.type == "cuda"` path, but not otherwise? Thx very much for taking time to educate me.

yao-matrix: > This is probably some comments I forgot to delete. Can you keep the code as it as it behaves slightly differently (`cuda` vs `to` in bnb). When calling to, we check if we are on cpu, otherwise, we don't quantize the weights.

@SunMarc , i  did a bit deeper look, this seems a deprecated behavior in bnb, i can see a 2024 Jan PR [here](https://github.com/bitsandbytes-foundation/bitsandbytes/commit/dcfb6f81433e37a8546f7dab3f648eaf858b29ff#diff-4d235c7e595546c6656c229dfa139298ce6602b356c2d0bafcb2352eb2cfae79L167) changed the `.cuda` behavior back to normal, before it has implicit  callings as below:
` w = self.data.contiguous().half().cuda(device)
  w_4bit, quant_state = bnb.functional.quantize_4bit(w, blocksize=self.blocksize, 
  compress_statistics=self.compress_statistics, quant_type=self.quant_type)`

SunMarc (approved):
LGTM !

yao-matrix:
@SunMarc , pls let me know who i need to @ to merge the PR, thx.

PR #3574, opened 2025-05-15
set backend correctly for CUDA+FSDP2+cpu-offload

SunMarc:
# What does this PR do?

Supersedes https://github.com/huggingface/accelerate/pull/3544

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -214,7 +214,11 @@ class PartialState:
                             if self.backend == "tccl":
                                 local_rank = os.environ.get("LOCAL_RANK", -1)
                                 torch.sdaa.set_device(f"sdaa:{local_rank}")
-                            if self.backend == "nccl" and os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" and os.environ.get("FSDP_OFFLOAD_PARAMS", "false") == "true":
+                            if (
+                                self.backend == "nccl"
+                                and os.environ.get("ACCELERATE_USE_FSDP", "false") == "true"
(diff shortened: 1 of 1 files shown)


universuen:
@SunMarc Hi, it is really a nice patch! However, I found a corner case when setting fsdp using kwargs like this:

```python
Accelerator(
    gradient_accumulation_steps=1,
    mixed_precision='bf16',
    fsdp_plugin=FullyShardedDataParallelPlugin(
        fsdp_version=2,
        cpu_offload=True,
    ),
)
```

Currently, I have to set the backend explicitly to avoid the error, but didn't have time to find a final solution to this.

```python
Accelerator(
    gradient_accumulation_steps=1,
    mixed_precision='bf16',
    fsdp_plugin=FullyShardedDataParallelPlugin(
        fsdp_version=2,
        cpu_offload=True,
    ),
    kwargs_handlers=[
        InitProcessGroupKwargs(
            backend="cuda:nccl,cpu:gloo"
        ),
    ]
)
```

SunMarc:
Indeed that's an edge case that we might need to fix if we want to allow users to depend only on the plugin in the future. cc @S1ro1

SunMarc:
I guess the easiest way for now is to update kwargs that is passed in partial state depending on `fsdp_plugin`

S1ro1:
> I guess the easiest way for now is to update kwargs that is passed in partial state depending on fsdp_plugin

I think we should just set gloo by default together with nccl if fsdp2 is happening, i.e. async checkpointing I work on also requires gloo so I feel like defaulting to both is sensible, even costing a little overhead in launch

SunMarc:
Okay, then we can do that in the async checkpoint pr

PR #3575, opened 2025-05-16
Set ccl and KMP param in simple launch

jiqing-feng:
Don't know why it assigns `CCL_WORKER_COUNT` only when machine> 1 because 1 CPU machine can also run distributed training or Tensor Parallelism.

I also added KMP params to get better performance on CPU.

With this PR we can run transformers TP model and got 40% speed-up on Intel 4th Gen Xeon.
The accelerate config is :
```
debug: false
distributed_type: MULTI_CPU
downcast_bf16: 'no'
enable_cpu_affinity: false
ipex_config:
  ipex: false
machine_rank: 0
main_training_function: main
mixed_precision: 'no'
mpirun_config:
  mpirun_ccl: '1'
  mpirun_hostfile: /home/jiqingfe/hostfile
num_machines: 1
num_processes: 2
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: true
```

The script is as follows
`accelerate launch script.py`
```python
import os
from transformers import AutoTokenizer, AutoModelForCausalLM
import oneccl_bindings_for_pytorch

import time
import torch

print(f"Using {torch.get_num_threads()} threads (PyTorch)")
print(f"OMP_NUM_THREADS={os.getenv('OMP_NUM_THREADS')}")

model_id = "meta-llama/Llama-3.1-8B-Instruct"

os.environ['RANK'] = str(os.environ.get('PMI_RANK', 0))
os.environ['WORLD_SIZE'] = str(os.environ.get('PMI_SIZE', 1))

def main(is_tp, rank, world_size) -> None:
    print("is_tp, rank, world_size: ", is_tp, rank, world_size)
    model_kwargs = dict(torch_dtype=torch.bfloat16)
    if is_tp:
        model_kwargs["tp_plan"] = "auto"
    else:
        model_kwargs["device_map"] = "cpu"

    # Retrieve tensor parallel model
    model = AutoModelForCausalLM.from_pretrained(model_id, **model_kwargs)
    if dist.is_initialized():
        print("Backend:", dist.get_backend())
    else:
        print("Distributed process group is not initialized.")

    # Prepare input tokens
    prompt = "It is done, and submitted. You can play 'Survival of the Tastiest' on Android, and on the web. Playing on the web works, but you have to simulate multiple touch for table moving and that can be a bit confusing. There is a lot I'd like to talk about. I will go through every topic, insted of making the typical what went right/wrong list. Concept Working over the theme was probably one of the hardest tasks which I had to face. Originally, I had an idea of what kind of game I wanted to develop, gameplay wise - something with a lot of enemies/actors"
    inputs = tokenizer(prompt, return_tensors="pt", max_length=512).to(model.device)
    print(f"inpu shape is {inputs.input_ids.shape}")

    model.generation_config.cache_implementation = "static"

    if is_tp:
        model.config.hidden_size = model.config.hidden_size // world_size
        model.config.num_key_value_heads = model.config.num_key_value_heads // world_size

    for i in range(1):
        with torch.no_grad():
            start = time.time()
            outputs = model.generate(**inputs, do_sample=False, max_new_tokens=128, min_new_tokens=128)
            end = time.time()
            print(f"time cost {(end-start)*1000} ms")

    # warm-up
    if is_tp:
        dist.barrier()

    if rank == 0:
        print(tokenizer.batch_decode(outputs, skip_special_tokens=True))

    model.forward = torch.compile(model.forward)
    # warm-up
    if is_tp:
        dist.barrier()

    for i in range(4):
        with torch.no_grad():
            start = time.time()
            outputs = model.generate(**inputs, do_sample=False, max_new_tokens=128, min_new_tokens=128)
            if is_tp:
                dist.barrier()

            end = time.time()
            print(f"time cost {(end-start)*1000} ms")

    if rank == 0:
        print(tokenizer.batch_decode(outputs, skip_special_tokens=True))


if __name__ == "__main__":
    rank = int(os.environ["RANK"]) if "RANK" in os.environ else 0
    world_size = int(os.environ["WORLD_SIZE"]) if "WORLD_SIZE" in os.environ else 1
    is_tp = world_size > 1
    main(is_tp, rank, world_size)

```

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -105,11 +105,11 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
     if args.no_python and args.module:
         raise ValueError("--module and --no_python cannot be used together")
 
+    num_processes = getattr(args, "num_processes", None)
     if args.mpirun_hostfile is not None:
         mpi_app_name, hostfile_arg, num_proc_arg, proc_per_node_arg, bind_to_arg = _get_mpirun_args()
         bind_to = getattr(args, "bind-to", "socket")
         num_machines = args.num_machines
-        num_processes = getattr(args, "num_processes", None)
         nproc_per_node = str(num_processes // num_machines) if num_processes and num_machines else "1"
         cmd += [
             mpi_app_name,
@@ -157,7 +157,7 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
         )
 
     ccl_worker_count = getattr(args, "mpirun_ccl", 0) if is_ccl_available() else 0
-    if getattr(args, "num_processes", 1) > 1:
+    if num_processes is not None and num_processes > 1:
         current_env["MASTER_ADDR"] = args.main_process_ip if args.main_process_ip is not None else "127.0.0.1"
         current_env["MASTER_PORT"] = str(args.main_process_port) if args.main_process_port is not None else "29500"
         current_env["CCL_WORKER_COUNT"] = str(ccl_worker_count)
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -106,10 +106,10 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
         raise ValueError("--module and --no_python cannot be used together")
 
     num_processes = getattr(args, "num_processes", None)
+    num_machines = args.num_machines
     if args.mpirun_hostfile is not None:
         mpi_app_name, hostfile_arg, num_proc_arg, proc_per_node_arg, bind_to_arg = _get_mpirun_args()
         bind_to = getattr(args, "bind-to", "socket")
-        num_machines = args.num_machines
         nproc_per_node = str(num_processes // num_machines) if num_processes and num_machines else "1"
         cmd += [
             mpi_app_name,
@@ -148,7 +148,7 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
             current_env["HABANA_VISIBLE_MODULES"] = args.gpu_ids
         else:
             current_env["CUDA_VISIBLE_DEVICES"] = args.gpu_ids
-    if args.num_machines > 1:
+    if num_machines > 1:
         assert args.main_process_ip is not None, (
             "When using multiple machines, you need to specify the main process IP."
         )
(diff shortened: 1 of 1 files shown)


jiqing-feng:
@sywangyi  @yao-matrix . Please review this PR, thanks!

yao-matrix:
does `ipex: true` work？

jiqing-feng:
Hi @yao-matrix . I have verified that both ipex is True and no config can work as before.

jiqing-feng:
Hi @SunMarc . Could you please review this PR? Thanks!

src/accelerate/utils/launch.py
```diff
@@ -149,14 +149,20 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
...
+        assert args.main_process_ip is not None, (
+            "When using multiple machines, you need to specify the main process IP."
+        )
+        assert args.main_process_port is not None, (
+            "When using multiple machines, you need to specify the main process port."
+        )
+
+    ccl_worker_count = getattr(args, "mpirun_ccl", 0) if is_ccl_available() else 0
+    current_env["MASTER_ADDR"] = args.main_process_ip if args.main_process_ip is not None else "127.0.0.1"
+    current_env["MASTER_PORT"] = str(args.main_process_port) if args.main_process_port is not None else "29500"
```
SunMarc, line 161: let's not set only when needed (args.num_machines > 1 or args.num_processes > 1)

src/accelerate/utils/launch.py
```diff
@@ -149,14 +149,20 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
         else:
             current_env["CUDA_VISIBLE_DEVICES"] = args.gpu_ids
     if args.num_machines > 1:
-        current_env["MASTER_ADDR"] = args.main_process_ip
-        current_env["MASTER_PORT"] = str(args.main_process_port)
-
-        if args.mpirun_hostfile is not None:
```
SunMarc, line 155: we don't need this check anymore ?

SunMarc (commented):
Thanks, left a couple of comments

src/accelerate/utils/launch.py
jiqing-feng: As I said in the PR statement, only 1 machine can also run CPU distributed with multi process. I suppose only 1 machine does not need to assign the master address or master port, right? Please let me know if there is anything I missed.

src/accelerate/utils/launch.py
jiqing-feng: Yes. Same as the previous comment. Even 1 machine without hostfile, we can still run multi process with ccl .

src/accelerate/utils/launch.py
SunMarc: yeah i just don't want to assign this when we have one machine and one process. For multiple processes, we should still assign it.

src/accelerate/utils/launch.py
```diff
@@ -149,14 +149,21 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
...
+        )
+        assert args.main_process_port is not None, (
+            "When using multiple machines, you need to specify the main process port."
+        )
 
-        if args.mpirun_hostfile is not None:
-            current_env["CCL_WORKER_COUNT"] = str(mpirun_ccl)
-    elif args.num_processes > 1:
+    ccl_worker_count = getattr(args, "mpirun_ccl", 0) if is_ccl_available() else 0
+    if num_processes is not None and num_processes > 1:
```
SunMarc, line 160: Although it rarely happens (num_machine>1 but num_processes=1), we should also check `if args.num_machines > 1:`

```suggestion
    if (num_processes is not None and num_processes > 1) or args.num_machines > 1: 
```

SunMarc (commented):
LGTM, just a small nit

SunMarc (approved):
Thanks ! LGTM !

PR #3578, opened 2025-05-19
enable test_cli & test_example cases on XPU

yao-matrix:
all passed. @SunMarc , pls help review, thx very much.

### DIFF
--- a/src/accelerate/utils/memory.py
+++ b/src/accelerate/utils/memory.py
@@ -183,13 +183,7 @@ def find_executable_batch_size(
 
 
 def get_xpu_available_memory(device_index: int):
-    if is_ipex_available():
-        ipex_version = version.parse(importlib.metadata.version("intel_extension_for_pytorch"))
-        if compare_versions(ipex_version, ">=", "2.5"):
-            from intel_extension_for_pytorch.xpu import mem_get_info
-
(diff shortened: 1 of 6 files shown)


src/accelerate/utils/memory.py
```diff
@@ -200,6 +194,12 @@ def get_xpu_available_memory(device_index: int):
             return torch.xpu.mem_get_info(device_index)[0]
         except Exception:
             pass
+    elif is_ipex_available():
+        ipex_version = version.parse(importlib.metadata.version("intel_extension_for_pytorch"))
+        if compare_versions(ipex_version, ">=", "2.5"):
+            from intel_extension_for_pytorch.xpu import mem_get_info
+
+            return mem_get_info(device_index)[0]
```
yao-matrix, line 202: here i changed the order of checking, since XPU is now a PyTorch built-in device, we'll check PyTorch API firstly, then check IPEX.

SunMarc: SG !

SunMarc (approved):
Thanks for fixing these !

SunMarc:
test_cli seems to be failing, can you have a look ?

yao-matrix:
fixed, thx very much.

PR #3583, opened 2025-05-22
feat: use datasets.IterableDataset shard if possible.

### LINKED ISSUE #3547: `datasets` Iterable Dataset sharding support
When I'm using huggingface `datasets`, I think a good faeture is the IterableDataset create by `load_dataset(..., streaming=True)`. It has `shard`, `shuffle` and `state_dict`, which is useful in multi node training. I can just call `shard` to make different node with different data, easy to randomize and recover.

However, it failed when working with `accelerate`. when passing dataloader with IterableDatset `accelerate` assume that the dataset instance is all same in different rank. and the `IterableDatasetShard` class is doing a overhead sampling by sampling n batch in all rank and yield out the sub batches belongs to the coresspoding rank [src ref](https://github.com/huggingface/accelerate/blob/v1.6.0/src/accelerate/data_loader.py#L265).

Not prepare the dataloader is a temproary solution in this case I think. But I think is nice to have the auto shard and shuffle in accelerate prepare. So I'm writing the issue for this. And If this feature is permmitted, but leaf no one to do, I think I can do the feature pr by myself

I have search the relative issue and find https://github.com/huggingface/accelerate/issues/2859
but this issue is aiming to add `load_state_dict` for dataloader

ValMystletainn:
# What does this PR do?

Add support for `datasets.IterableDataset` sharding if pass it to `accelerator.prepare`.
Use the `n_shard` rather than `IterableDatasetShard` to reduce the data reading overhead, and make different rank reading different data shard efficient.

Fixes #3547 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?
@SunMarc

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -14,6 +14,7 @@
 
 import importlib
 import math
+import sys
 from contextlib import suppress
 from typing import Callable, Optional, Union
 
@@ -1194,7 +1195,13 @@ def prepare_data_loader(
         dataloader.sampler.generator = generator
     # No change if no multiprocess
     if (num_processes != 1 or state.distributed_type == DistributedType.MEGATRON_LM) and not dispatch_batches:
-        if isinstance(new_dataset, IterableDataset):
+        if (
+            isinstance(new_dataset, getattr(sys.modules.get("datasets"), "IterableDataset", type(None)))
+            and not split_batches
+            and new_dataset.n_shard > num_processes
+        ):
+            new_dataset = new_dataset.shard(num_shards=num_processes, index=process_index)
+        elif isinstance(new_dataset, IterableDataset):
             if getattr(dataloader.dataset, "generator", None) is not None:
                 synchronized_generator = dataloader.dataset.generator
             new_dataset = IterableDatasetShard(


ValMystletainn:
I guess I shoud write a test for it and change somewhere in docs.

However, I readthough the `tests/test_data_loader.py` and find no where to starts a multi process test suit for this function.

src/accelerate/data_loader.py
```diff
@@ -1194,7 +1195,13 @@ def prepare_data_loader(
         dataloader.sampler.generator = generator
     # No change if no multiprocess
     if (num_processes != 1 or state.distributed_type == DistributedType.MEGATRON_LM) and not dispatch_batches:
-        if isinstance(new_dataset, IterableDataset):
+        if (
+            isinstance(new_dataset, getattr(sys.modules.get("datasets"), "IterableDataset", type(None)))
```
SunMarc, line 1199: That could work but let's use check if dataset is available (`is_datasets_available`) and import the class `IterableDataset` from there to perform the check.

SunMarc (commented):
Thanks ! A test would be nice to have !

src/accelerate/data_loader.py
ValMystletainn: I write in this style rather than use 
```
if is_datasets_available():
    from datasets import IterableDataset as DatasetsIterableDatasets
...

if isinstance(new_dataset, DatasetsIterableDatasets):
    ...
```


is aiming to reduce  import overhead like this [codesnippet](https://github.com/gaogaotiantian/coredumpy/blob/31becdc08169513999713d6ef61c001c658ce6c3/src/coredumpy/dap_server.py#L396). it check the object is `torch.Tensor` or not, and skip to import the heavy pytorch package if there is no `torch.Tensor` object at all.

however, I think the import overhead of datasets is not so heavy like `torch`, so if it's for the  readability and maintainability, I will change to this style
```
if is_datasets_available():
    from datasets import IterableDataset as DatasetsIterableDatasets
...
```


So what do you think about it, the original version or the import and check version.

SunMarc: yeah it should be fine with the overhead. We only call this function once so it shouldn't create a huge overhead. Please go with the import + check version.

SunMarc:
I've overtaken your PR to finish it ! Thanks a lot for adding this @ValMystletainn

PR #3585, opened 2025-05-22
[FSDP2, Do not merge] FP8 adjustments

S1ro1:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -648,10 +648,11 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
     if auto_wrap_policy is not None:
         # We skip the model itself, as that one is always wrapped
         for module in get_module_children_bottom_up(model)[:-1]:
-            if auto_wrap_policy(module):
+            if auto_wrap_policy(module) and not isinstance(module, FSDPModule):
                 fully_shard(module, **fsdp2_kwargs)
 
-    fully_shard(model, **fsdp2_kwargs)
+    if not isinstance(model, FSDPModule):
+        fully_shard(model, **fsdp2_kwargs)
 
     if fsdp2_plugin.cpu_ram_efficient_loading:
         # If `cpu_ram_efficient_loading` is enabled, only rank 0 loads the weights


examples/fsdp2/fsdp2_fp8.py
```diff
@@ -0,0 +1,255 @@
...
+
+    args = parse_args()
+
+    fsdp2_plugin = FullyShardedDataParallelPlugin(
+        fsdp_version=2,
+        cpu_ram_efficient_loading=False,  # CPU RAM efficient loading CANNOT work with fp8 torchao
+        auto_wrap_policy="transformer_based_wrap",
+        transformer_cls_names_to_wrap=["LlamaDecoderLayer"],
+    )
+    fsdp2_plugin.set_mixed_precision("bf16")
```
SunMarc: you can set it to `fp8` or `bf16` depending on if we have `args.bf16`, the mapping inside of `set_mixed_precision` will take care of selecting the right torch_dtype. It will be less confusing this way

examples/fsdp2/README.md
```diff
@@ -0,0 +1,32 @@
...
+which replaces the default `torch.nn.Linear` with `Float8Linear`. We also utilize `TorchDynamoPlugin` together with regional compilation to compile the model,
+gaining even more speed and memory savings, as `ao` doesn't ship with any kernels by default, so we have to gain the performance from compiling the model.
+
+Replacing linear layers with `Float8Linear` can greatly improve performance, if used correctly and on hardware that supports FP8 tensor cores. This highly depends on the model dimensions and sequence length used for training.
+You can view the performance of `Float8Linear` as a function of matrix dimensions in [this document](https://github.com/pytorch/ao/blob/main/torchao/float8/README.md#performance). 
+
+In our example, we use a 8B Llama3.1 model, which has a hidden dimension of 4096 and we train on sequence length of 8192. In the below images, we can see that this improves performance by ~25% compared to `bf16`, reaching ~10000 tokens per second, per device on 8x H100 GPUs, compared to ~8000 tokens per second using `bf16`, while loss function stays roughly the same.
```
SunMarc: can we also add a graph for the tflops since you also added it in your script ?

src/accelerate/accelerator.py
```diff
@@ -1396,30 +1397,21 @@ def prepare(self, *args, device_placement=None):
...
                     " as the optimizer needs to have its parameters modified after the model is converted."
                 )
+            if model_count > 1:
+                raise ValueError("Only one model is supported when using FSDP2")
+
+            # TODO: S1ro - this is probably gonna be a problem with other fp8 backends too
+            if self.fp8_backend == "AO" and self.state.fsdp_plugin.cpu_ram_efficient_loading:
+                raise ValueError(
+                    "torchao with FSDP2 and cpu_ram_efficient_loading is not supported, setting `cpu_ram_efficient_loading` to False will fix the issue and work as intended."
+                )
```
SunMarc: let's move this part in the init of accelerator, it will be cleaner this way as we shouldn't need to call prepare to raise this error

src/accelerate/accelerator.py
```diff
@@ -1463,6 +1452,75 @@ def prepare(self, *args, device_placement=None):
...
+        # Apply AC if needed
+        if self.state.fsdp_plugin.activation_checkpointing:
+            model = fsdp2_apply_ac(self, model)
+
+        # Apply compile if needed, has to be *after* applying AC
+        if self.state.dynamo_plugin.backend != DynamoBackend.NO and not is_compiled_module(model):
+            if self.state.dynamo_plugin.use_regional_compilation:
+                model = compile_regions(model, **self.state.dynamo_plugin.to_kwargs())
+            else:
+                model = torch.compile(model, **self.state.dynamo_plugin.to_kwargs())
```
SunMarc: add a copied from comment so that we can easily decide to update this part of the code if needed

src/accelerate/accelerator.py
```diff
@@ -3547,10 +3617,26 @@ def clear(self, *objects):
...
+                    accessor_mapping[WeightWithDynamicFloat8CastTensor] = "_tensor"
+
+                named_parameters.update(
+                    {
+                        n: getattr(p, accessor_mapping[type(p)]).data_ptr()
+                        if type(p) in accessor_mapping
+                        else p.data_ptr()
+                        for n, p in obj.named_parameters()
+                    }
+                )
```
SunMarc, line 3643: explain this bits

src/accelerate/utils/fsdp_utils.py
```diff
@@ -694,9 +699,7 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
     return model
 
 
-def fsdp2_prepare_auto_wrap_policy(
-    fsdp2_plugin, auto_wrap_policy_type: str, model: torch.nn.Module
-) -> Callable[[torch.nn.Module], bool]:
```
SunMarc, line 699: not the best pratice to change a public method args but should be okay hopefully

src/accelerate/utils/fsdp_utils.py
```diff
@@ -768,4 +778,5 @@ def fsdp2_canonicalize_names(named_params: dict) -> dict:
     named_params = {
         k.replace("_orig_mod.", "") if k.startswith("_orig_mod.") else k: v for k, v in named_params.items()
     }
+    named_params = {k.replace("._orig_mod", ""): v for k, v in named_params.items()}
```
SunMarc, line 781: should we check that it ends with `._orig_mod` ?

SunMarc (commented):
Thanks ! LGTM overall ! Left some minor comments

examples/fsdp2/fsdp2_fp8.py
S1ro1: This is just mixed_precision, we don't want to really change it, we want to always train in `bf16` MP, just change layers to Float8Linear if needed. Probably gonna rename the argument to be less confusing instead.

examples/fsdp2/README.md
S1ro1: Sure, I didn't think it's really needed as it basically is the same *scale* as TPS, as they're related.

src/accelerate/utils/fsdp_utils.py
S1ro1: Probably not, I think it was not a good decision for this and several other methods to be *public* as they depend on many conditions that we make sure are met when calling them, though users might not.

src/accelerate/utils/fsdp_utils.py
S1ro1: Not really, this method now works in a bit different way, i.e. it really "canonicalises" the state dict. Meaning it removes all the modifiers that just change names (currently ac + compile). `.replace()` returns the original string if it doesn't contain the searched word, therefore this only modifies state dicts that contain params with `._orig_mod`, it keeps the other param names intact.

examples/fsdp2/README.md
SunMarc: it's more to compare to the theoretical tflofs of hardware to see how much we could potentially improve

examples/fsdp2/fsdp2_fp8.py
SunMarc: check the `set_mixed_precision` definition, `fp8` is mapped to `bf16` inside

examples/fsdp2/README.md
S1ro1: That is relatively hard to do, as with `Float8Linear` we actually do part of matmuls in fp8, so harder to get the theoretical maximum, as we're not entirely fp8, nor bf16.

SunMarc: yeah but still I think this will be a good information for users that might be interested

SunMarc (approved):
LGTM !

src/accelerate/accelerator.py
```diff
@@ -1396,30 +1403,15 @@ def prepare(self, *args, device_placement=None):
                     "When using FSDP2, a model and optimizer must be passed together to `Accelerator.prepare()`"
                     " as the optimizer needs to have its parameters modified after the model is converted."
                 )
+            if model_count > 1:
+                raise ValueError("Only one model is supported when using FSDP2")
```
yh8899, line 1407: Why only one model is supported when using fsdp2? Can you explained?

S1ro1: I have had very weird behaviour show when 2 models were used. I don't think supporting just 1 model is final, though I didn't have time to check it really.

PR #3597, opened 2025-05-28
Make compile regions inplace

IlyasMoutawwakil:
# What does this PR do?

Some frameworks apparently do very custom stuff with their Modules (deepspeed 😭) to the point that any alteration of it breaks everything.
I made an in-place option and used it in prepare_deepspeed, also provided a method that allows to decompile the model inplace.

Here's a minimal reproducible of the problem that has nothing to do with regional compilation:

```python
from torch.optim import AdamW
from transformers import GPT2LMHeadModel, GPT2Tokenizer
from deepspeed import initialize


model_name = "gpt2"
tokenizer = GPT2Tokenizer.from_pretrained(model_name)
model = GPT2LMHeadModel.from_pretrained(model_name)


deepspeed_engine, optimizer, _, _ = initialize(
    model=model, optimizer=AdamW(model.parameters()), config={"train_batch_size": 1}
)
deepspeed_engine.module = deepspeed_engine.module

# both below fail
print(list(deepspeed_engine.module.named_parameters()))
print(dir(deepspeed_engine.module))
```

This will fail because once the engine.module is overwritten deepspeed will start looping infinitely when trying to get attributes from its module, because of this method https://github.com/deepspeedai/DeepSpeed/blob/b4cc079eee835d8dbcbd5d9b894380830e5c5abf/deepspeed/runtime/engine.py#L593

Doing this works tho:

```python
compile_regions(deepspeed_model.module, inplace=True)
print(dir(deepspeed_model.module))
print(deepspeed_model.module)
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


src/accelerate/utils/other.py
```diff
@@ -60,22 +60,7 @@ def is_compiled_module(module: torch.nn.Module) -> bool:
...
-    if module._modules:
-        for submodule in module.modules():
-            if isinstance(submodule, torch._dynamo.eval_frame.OptimizedModule):
-                return True
-
-    return False
-
-
-def compile_regions(module: torch.nn.Module, **compile_kwargs) -> torch.nn.Module:
+def compile_regions(module: torch.nn.Module, inplace: bool = False, **compile_kwargs) -> torch.nn.Module:
```
SunMarc, line 63: maybe we can use inplace by default ? it might be a slightly breaking but since this is a new feature that will be patched, it seems reasonable.

src/accelerate/utils/other.py
```diff
@@ -124,13 +113,11 @@ def compile_regions(module: torch.nn.Module, **compile_kwargs) -> torch.nn.Modul
 
     def _compile_regions(module: torch.nn.Module, **compile_kwargs) -> torch.nn.Module:
         if isinstance(module, torch.nn.ModuleList):
-            if all(isinstance(submodule, module[0].__class__) for submodule in module):
```
SunMarc, line 127: why are we removing this ?

src/accelerate/utils/other.py
```diff
@@ -173,9 +224,8 @@ def extract_model_from_parallel(
     options = (torch.nn.parallel.DistributedDataParallel, torch.nn.DataParallel)
 
     is_compiled = is_compiled_module(model)
-    has_compiled = has_compiled_regions(model)
```
SunMarc, line 176: `is_compiled_module` is enough ?

tests/test_compile.py
```diff
@@ -41,37 +42,27 @@ def _get_model_and_inputs(self):
...
-
-        assert compiled_model._orig_mod is compiled_model_unwrapped._orig_mod
-
-    def test_extract_model_remove_torch_compile(self):
-        model, _ = self._get_model_and_inputs()
-        compiled_model = compile_regions(model)
-
-        distributed_model = torch.nn.parallel.DataParallel(model)
-        distributed_compiled_model = compile_regions(distributed_model)
-        compiled_model_unwrapped = extract_model_from_parallel(distributed_compiled_model, keep_torch_compile=False)
```
SunMarc, line 72: why remove those ?

SunMarc (commented):
Thanks ! Left a couple of comments

src/accelerate/utils/other.py
IlyasMoutawwakil: inplace doesn't work /breaks the benchmark for example, and inplace is only necessary here for deepspeed because of the way the engine shares state with the underlying module.

src/accelerate/utils/other.py
IlyasMoutawwakil: good catch ! I was experimenting with regional comp + deepspeed on Gaudi and removed it while debugging

src/accelerate/utils/other.py
IlyasMoutawwakil: yes, I believe it was wrong to consider a regionally compiled model the same as a compiled model.
`keep_torch_compile` should only handle `is_compiled_module` because that's the only case where it has _orig_mod now.

src/accelerate/utils/other.py
SunMarc: wdym it break the benchmark ?

src/accelerate/utils/other.py
SunMarc: stil should we call `decompile_regions` in that case ?

tests/test_compile.py
SunMarc: see my point about extract_model_from_parallel above. Maybe we can change slightly the test then ? Feels like we should still be able to test those

SunMarc (commented):
Left a few remarks

src/accelerate/utils/other.py
IlyasMoutawwakil: unfortunately decompile_regions is inplace as well 😅 doing something like:
```python
    if is_compiled:
        compiled_model = model
        model = model._orig_mod
    elif has_compiled:
        compiled_model = model
        model = decompile_regions(model)
```
will result in both compiled_model and model being the same, and then at the end you will have to regionally compile it again.

tests/test_compile.py
IlyasMoutawwakil: I changed the test, and added a comment explaining why you don't need to decompile regions when extracting from a parallel model ; because compile_rgions doesn't add a thin wrapper around the top level module. and if it did (edge case where model was compiled entirely) the model will be handled in:
```python
    if is_compiled:
        compiled_model = model
        model = model._orig_mod
```

src/accelerate/utils/other.py
SunMarc: that's too bad ;) The issue is that if we don't decompile the model in extract_model_from_parallel, we won't be able to save the original model when training it for example with Trainer. Maybe we can add a warning in  `extract_model_from_parallel` if the model is regionally compiled so that the users knows that he has to call `decompile_regions` + recompile afterwards

IlyasMoutawwakil: what we can do is, at the end of the extract method, if not keep_torch_compile: we decompile regions.

SunMarc: Yeah we can do that but still I think we should warn the user that the model needs to be recompiled afterwards unfortunately.

IlyasMoutawwakil: I dont see why tho ? keep_torch_compile is true by default, and compile_regions doesn't interfere with that, and if a user sets it to false, they are explicitly asking for the decompiled model. why warn ?

SunMarc: In Trainer, we have `keep_torch_compile=False,` so users might not understand why the model is suddenly training slower after it has been saved

SunMarc: also the usual behavior of `unwrap` is that it is not an `inplace` function, so that we can save the model then restart to train without having to wrap again

PR #3599, opened 2025-05-29
enable torchao and pippy test cases on XPU

yao-matrix:
torchao has xpu support since 0.11.0, so enable torchao cases. ALL cases passed.
enable pippy cases on XPU, ALL passed.

@SunMarc , pls help review.

Till now, all existing accelerate ut cases which are valid on XPU are enabled, thx very much @SunMarc @IlyasMoutawwakil for your great support!


SunMarc (approved):
Thx !

PR #3604, opened 2025-06-02
[FSDP2] Context parallel

S1ro1:
Building on top of #3585, this PR enables context parallelism together with FSDP2 (branch stems from that PR, that's why so many changes).
~~ Possibly, this will be moved to global, not fsdp local namespace, though it requires a bit more experiments from me to see what it works with. ~~
We probably want to always use this with fsdp/zero, as from my profiling it ends up in a perfect overlap and u get model sharding for free. For zero I'll need to check Ulysses then for context parallel.

This has been verified with benchmarks against torchtitan, where our implementation reaches ~equal performance and memory usage (against my reruns of torchtitan, not their reported numbers).

Managed to scale training to ~500k sequence length on 32x H100 with 8B Llama, torchtitan reports ~1m context length on this scale, but had the same results with their repo on this scale. With doubling sequence length, we need to double the GPU count, so I reasonably stopped at ~500k, as doing benchmarks at 64 GPU scale felt not too productive to me yet.

TODO:
- [x] Tests
- [ ] Proper benchmarks
- [x] Add to `accelerate config`
- [x] concept guide
- [x] #3585 needs to get merged

### DIFF
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -20,7 +20,7 @@ This guide will cover basics of using context parallelism in 🤗`accelerate`, f
 ## Why context parallelism?
 
 With the advent of large language models, and recently reasoning models, the sequence length has been growing rapidly. This, combined with quadratic memory complexity of attention, has lead to a need for more efficient ways to train models with long sequences.
-With sequence length of 128k, the memory requirement of `S` in the attention computation is `128k * 128k * 2 * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
+With sequence length of 128k, the memory requirement of the attention matrix is `128k * 128k * 2 * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
 
 Context parallelism allows us to shard the inputs to the attention computation along the sequence dimension and compute the attention in parallel on multiple GPUs. With this, we can train models with long sequences, scaling potentially to 1M+ sequence length.
 
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -126,14 +126,14 @@ You can directly see this issue in the profiler output in the image below:
 <p align="center">
   <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/accelerate/examples/fsdp2/cp_all_to_all.png" alt="all-to-all profiler output" />
   <br>
-  <em>Figure 1: In red you can see the idle time, while we wait for the all-to-all kernel to finish. It's also highlighted in the first blue bar, where we can see it takes ~250us to finish, which is repeated N-1 times for each attention call, where N is the context parallel size.</em>
+  <em>Figure 1: In red you can see the idle time, while we wait for the all-to-all kernel to finish. Highlighted in the first blue bar, you can see that it takes ~250us to finish, which is repeated N-1 times for each attention call, where N is the context parallel size.</em>
 </p>
 
 
 ## Why only FSDP2?
 
 We only support context parallelism with `FSDP2` for now, as we create a joint mesh of `context_parallel_size` and `dp_shard_size` to 
-utilize its full potential. In the profiler output in the image below, you can see that the computation of `attn_i`
+utilize its full potential. In the profiler output in the image below, you can see why this is the case.
 
 <p align="center">
   <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/accelerate/examples/fsdp2/cp_why_fsdp2.png" alt="why FSDP2+CP" />
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -66,6 +66,18 @@ for batch in dataloader:
 > [!Warning]
 > This context manager has to be recreated with each training step, as shown in the example above. It's crucial to do so.
 
+This can scale your context size to 1M+ sequence length potentially. Below, we showcase speed and memory usage of context parallelism for up-to 256k context size. We can see that when we double the context size and number of GPUs, we can achieve consistent memory usage, potentiall enabling endless context length scaling.
+
+<p align="center">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/accelerate/examples/fsdp2/cp_perf.png" alt="context parallelism memory usage" />
+  <br>
+  <em>Figure 1: Memory usage and speed of context parallelism for up-to 256k context size.</em>
+</p>
+
+> [!Tip]
+> These examples were created with a script you can find [in the examples folder](https://github.com/huggingface/accelerate/blob/main/examples/fsdp2/fsdp2_context_parallel.py). For instructions on how to run it, see the [README](https://github.com/huggingface/accelerate/blob/main/examples/fsdp2/README.md) in the same folder.
+
+
 ## Accelerate's interface
 
 The context manager takes a few arguments, that are used to configure the context parallelism.
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -43,8 +43,8 @@ from accelerate.utils import FullyShardedDataParallelPlugin
 plugin = FullyShardedDataParallelPlugin(
     ...
     fsdp_version=2,
-+   context_parallel_size=8,
-+   context_parallel_shard_rotation="allgather",
++   cp_size=8,
++   cp_comm_strategy="allgather",
 )
 accelerator = Accelerator(fsdp_plugin=plugin)
 ```
@@ -77,8 +77,8 @@ The context manager takes a few arguments, that are used to configure the contex
 ## Configurable options
 Accelerate provides only a few options to configure context parallelism, which are:
 
-- `context_parallel_size`: The number of ranks to shard the inputs to the attention computation across the sequence dimension.
-- `context_parallel_shard_rotation`: The rotation method to use for the shards. We strongly reccomend keeping this as `"allgather"`, as it's very likely it will outperform `"alltoall"` in most cases.
+- `cp_size`: The number of ranks to shard the inputs to the attention computation across the sequence dimension.
+- `cp_comm_strategy`: The rotation method to use for the shards. We strongly reccomend keeping this as `"allgather"`, as it's very likely it will outperform `"alltoall"` in most cases.
 
 Context parallel size is rather self-explanatory, it's the number of ranks across which the inputs are to be-sharded.
 Context parallel shard rotation defines how the shards of the inputs are rotated across ranks. We'll cover the 2 options in more detail in the next section.
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -20,7 +20,7 @@ This guide will cover basics of using context parallelism in 🤗`accelerate`, f
 ## Why context parallelism?
 
 With the advent of large language models, and recently reasoning models, the sequence length has been growing rapidly. This, combined with quadratic memory complexity of attention, has lead to a need for more efficient ways to train models with long sequences.
-With sequence length of 128k, the memory requirement of the attention matrix is `128k * 128k * 2 * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
+With sequence length of 128k, the memory requirement of the attention matrix is `128k * 128k * 2 bytes * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
 
 Context parallelism allows us to shard the inputs to the attention computation along the sequence dimension and compute the attention in parallel on multiple GPUs. With this, we can train models with long sequences, scaling potentially to 1M+ sequence length.
 
--- a/docs/source/concept_guides/context_parallel.md
+++ b/docs/source/concept_guides/context_parallel.md
@@ -1,0 +1,144 @@
+<!--Copyright 2025 The HuggingFace Team. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+specific language governing permissions and limitations under the License.
+
+⚠️ Note that this file is in Markdown but contain specific syntax for our doc-builder (similar to MDX) that may not be
+rendered properly in your Markdown viewer.
+-->
+
+# Context Parallel in 🤗`accelerate`
+
+This guide will cover basics of using context parallelism in 🤗`accelerate`, for the more curious readers, we will also cover some technicalities in the later sections.
+
+## Why context parallelism?
+
+With the advent of large language models, and recently reasoning models, the sequence length has been growing rapidly. This, combined with quadratic memory complexity of attention, has lead to a need for more efficient ways to train models with long sequences.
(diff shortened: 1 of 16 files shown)


examples/fsdp2/fsdp2_context_parallel.py
```diff
@@ -0,0 +1,176 @@
...
+
+    tokenizer = setup_tokenizer(MODEL_ID)
+    optimizer = torch.optim.AdamW(model.parameters(), lr=1e-5)
+
+    model, optimizer = accelerator.prepare(model, optimizer)
+
+    accelerator.print("Preparing dataset... this might take a while")
+    dataset = get_dataset(accelerator, tokenizer, args.sequence_length)
+    dataloader = DataLoader(dataset, batch_size=1, collate_fn=create_collate_fn())
+    dataloader = accelerator.prepare(dataloader)
```
yh8899, line 131: Does the dataloader have compatibility with context parallel?

S1ro1: Yes it does and no it doesn't. This is something that is handled by using a joint mesh for FSDP + CP -> we dispatch a different batch (bc we're in a data parallel regime) on each rank, there is no need to change anything for context parallel.

yh8899: I have a little confusion. rank 0 and 1 in one cp group, should don't they have the same batch data? Have i miss something? 

> a joint mesh for FSDP + CP -> we dispatch a different batch (bc we're in a data parallel regime) on each rank, there is no need to change anything for context parallel.

docs/source/concept_guides/context_parallel.md
```diff
@@ -0,0 +1,144 @@
...
+-->
+
+# Context Parallel in 🤗`accelerate`
+
+This guide will cover basics of using context parallelism in 🤗`accelerate`, for the more curious readers, we will also cover some technicalities in the later sections.
+
+## Why context parallelism?
+
+With the advent of large language models, and recently reasoning models, the sequence length has been growing rapidly. This, combined with quadratic memory complexity of attention, has lead to a need for more efficient ways to train models with long sequences.
+With sequence length of 128k, the memory requirement of the attention matrix is `128k * 128k * 2 * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
```
SunMarc:
```suggestion
With sequence length of 128k, the memory requirement of the attention matrix is `128k * 128k * 2 bytes * num_heads = ~32 GB * num_heads` for `bf16` precision, given vanilla attention implementation. Granted, with usage of `flash attention` or `SDPA` which do not materialize these attention weights, this decreases drastically, but the growth in memory requirements is still considerable.
```
Nice calculation

docs/source/concept_guides/context_parallel.md
```diff
@@ -0,0 +1,144 @@
...
+In ideal scenario, all-gather finishes in the exact moment as the calculation of `attn_i` is done. However, this never happens in practice, so the ideal real overlap is achieved when the full `attn_i` is overlapped with a part of the communication, then to start the computation with `K_j` and `V_j`, we wait for the all-gather to finish.
+
+### all-to-all
+All-to-all, or sometimes called `ring-rotation` utilizes a ring-like communication pattern. After concluding `attn_i` computation, an all-to-all is launched to send `K_i` and `V_i` to the neighbouring ranks. We then repeat this `context_parallel_size-1` times, so that each rank sees all the shards of `K` and `V` from all other ranks once. In ideal scenario, we prefetch shards `K_i+1` and `V_i+1` from the neighbouring rank and this communication is exactly overlapped with computation of our current `attn_i`. Again, realistically, this perfect overlap doesn't ever happen. Given the nature of this approach, if we don't achieve perfect overlap, the penalty is way larger than with all-gather.
```
SunMarc, line 132: this is ring attention basically no ?

docs/source/concept_guides/context_parallel.md
```diff
@@ -0,0 +1,144 @@
...
+You can also enable context parallelism programatically, by passing it in the `FullyShardedDataParallelPlugin` constructor:
+
+```diff
+from accelerate.utils import FullyShardedDataParallelPlugin
+
+plugin = FullyShardedDataParallelPlugin(
+    ...
+    fsdp_version=2,
++   context_parallel_size=8,
++   context_parallel_shard_rotation="allgather",
```
SunMarc: instead of `context_parallel_shard_rotation`, maybe call this `context_parallel_comm_stategy` WDYT ?

SunMarc: also should we shorten the `context_parallel` to `cp` ?

examples/fsdp2/fsdp2_context_parallel.py
```diff
@@ -0,0 +1,176 @@
...
+    fsdp_plugin = FullyShardedDataParallelPlugin(
+        auto_wrap_policy="transformer_based_wrap",
+        transformer_cls_names_to_wrap=["LlamaDecoderLayer"],
+        cpu_ram_efficient_loading=True,
+        fsdp_version=2,
+        context_parallel_size=args.context_parallel_size,
+        context_parallel_shard_rotation=args.context_parallel_shard_rotation,
+        activation_checkpointing=True,
+    )
+    fsdp_plugin.set_mixed_precision("bf16")
```
SunMarc: not a good practice to set it here, maybe put that in the accelerator instead that will handle this ?

examples/fsdp2/fsdp2_context_parallel.py
SunMarc: for the same cp group, we should indeed need to have the same data, let's add some tests to make sure that the data is correctly dispatched.

src/accelerate/accelerator.py
```diff
@@ -1254,6 +1254,65 @@ def join_uneven_inputs(self, joinables, even_batches=None):
             with contextlib.nullcontext(joinables):
                 yield
 
+    @contextmanager
+    def context_parallel(
+        self,
+        buffers: list[torch.Tensor] | None = None,
+        buffer_seq_dims: list[int] | None = None,
+        no_restore_buffers: set[torch.Tensor] | None = None,
```
SunMarc, line 1285: add `Optional` typing

src/accelerate/accelerator.py
```diff
@@ -1254,6 +1254,65 @@ def join_uneven_inputs(self, joinables, even_batches=None):
...
+        ...         outputs = model(batch)
+        ...         ...
+        ```
+        """
+
+        if (
+            getattr(self.state.fsdp_plugin, "context_parallel_size", None) is None
+            or (cp_context := getattr(self, "_cp_context", None)) is None
+        ):
+            warnings.warn("Context parallel is not configured, this context manager will have no effect.")
```
SunMarc: precise fsdpv2 also in the warning + Use logger.warning also
btw, we don't need to use `getattr` as `fsdp_plugin` will definitely have `context_parallel_size` attribute

src/accelerate/accelerator.py
```diff
@@ -1464,6 +1523,39 @@ def _prepare_fsdp2(self, *args):
         # Needs to be done first, to make sure AC + fully_shard will work as expected
         self.state.fsdp_plugin.set_auto_wrap_policy(model)
 
+        _fully_shard_kwargs = {}
+
+        if (context_parallel_size := getattr(self.state.fsdp_plugin, "context_parallel_size", None)) is not None:
```
SunMarc: same here

src/accelerate/accelerator.py
```diff
@@ -1464,6 +1523,39 @@ def _prepare_fsdp2(self, *args):
...
+                raise ValueError(
+                    f"context_parallel_size set to {context_parallel_size}, which is greater than the number of processes {self.state.num_processes}. Please set to None or use a smaller value."
+                )
+
+            from torch.distributed.device_mesh import init_device_mesh
+            from torch.distributed.tensor.experimental import context_parallel
+            from torch.distributed.tensor.experimental._attention import set_rotate_method
+
+            context_parallel_shard_rotation = getattr(
+                self.state.fsdp_plugin, "context_parallel_shard_rotation", "allgather"
```
SunMarc: here also

src/accelerate/utils/dataclasses.py
```diff
@@ -1691,6 +1697,18 @@ class FullyShardedDataParallelPlugin:
             "help": "The minimum number of parameters a module must have to be wrapped. Only applicable when `auto_wrap_policy` is `size_based_wrap`."
         },
     )
+    context_parallel_size: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": "The size of the context parallel group. Only applicable when `fsdp_version` is set to 2, else error will be raised."
+        },
```
SunMarc: precise default value here also

src/accelerate/utils/dataclasses.py
```diff
@@ -1853,6 +1871,41 @@ def __post_init__(self):
...
+
+        if self.context_parallel_size is not None and self.fsdp_version != 2:
+            raise ValueError(
+                f"context_parallel_size set to {self.context_parallel_size}. This is not supported with FSDP1, please set to None or use `fsdp_version=2`"
+            )
+
+        if self.context_parallel_size is not None and not is_torch_version(">=", CONTEXT_PARALLEL_PYTORCH_VERSION):
+            raise ValueError(
+                f"context_parallel_size set to {self.context_parallel_size}. This is not supported with PyTorch < {CONTEXT_PARALLEL_PYTORCH_VERSION}, please set to None or upgrade your PyTorch version."
+            )
```
SunMarc: simplify this a bit

src/accelerate/utils/dataclasses.py
```diff
@@ -1853,6 +1871,41 @@ def __post_init__(self):
...
+
+        if self.context_parallel_shard_rotation is None:
+            self.context_parallel_shard_rotation = os.environ.get(env_prefix + "CONTEXT_PARALLEL_SHARD_ROTATION", None)
```
SunMarc:
```suggestion
        if self.context_parallel_shard_rotation is None:
            self.context_parallel_shard_rotation = os.environ.get(env_prefix + "CONTEXT_PARALLEL_SHARD_ROTATION", "allgather")
```

src/accelerate/utils/dataclasses.py
```diff
@@ -1853,6 +1871,41 @@ def __post_init__(self):
...
+        if self.context_parallel_shard_rotation is not None:
+            if self.context_parallel_shard_rotation not in ["allgather", "alltoall"]:
+                raise ValueError(
+                    f"context_parallel_shard_rotation set to {self.context_parallel_shard_rotation}. Must be one of ['allgather', 'alltoall']."
+                )
+
+            if self.fsdp_version == 1:
+                raise ValueError(
+                    f"context_parallel_shard_rotation set to {self.context_parallel_shard_rotation}. This is not supported with FSDP1, please set to None or use `fsdp_version=2`"
+                )
```
SunMarc: shouldn't be needed

src/accelerate/utils/dataclasses.py
```diff
@@ -1853,6 +1871,41 @@ def __post_init__(self):
...
+
+            if self.context_parallel_size is None or self.context_parallel_size == 1:
+                raise ValueError(
+                    f"context_parallel_size set to {self.context_parallel_size} but context_parallel_shard_rotation is set to {self.context_parallel_shard_rotation}. Please set context_parallel_size to a value greater than 1."
+                )
```
SunMarc: check not needed

src/accelerate/utils/launch.py
```diff
@@ -329,6 +329,14 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> dict[str, str]:
         current_env["FSDP_SYNC_MODULE_STATES"] = str(args.fsdp_sync_module_states).lower()
         current_env["FSDP_ACTIVATION_CHECKPOINTING"] = str(args.fsdp_activation_checkpointing).lower()
 
+        if (fsdp_context_parallel_size := getattr(args, "fsdp_context_parallel_size", 1)) != 1:
+            current_env["FSDP_CONTEXT_PARALLEL_SIZE"] = str(fsdp_context_parallel_size)
+
+        if (
+            fsdp_context_parallel_shard_rotation := getattr(args, "fsdp_context_parallel_shard_rotation", None)
+        ) is not None:
+            current_env["FSDP_CONTEXT_PARALLEL_SHARD_ROTATION"] = str(fsdp_context_parallel_shard_rotation).lower()
```
SunMarc: shouldn't need getattr
make sure to update the `launch_command_parser` to include the new args btw

tests/fsdp/test_fsdp.py
```diff
@@ -420,6 +420,24 @@ def test_cpu_ram_efficient_loading(self):
...
+                fsdp_plugin = FullyShardedDataParallelPlugin()
+                assert fsdp_plugin.context_parallel_shard_rotation == context_parallel_shard_rotation
+
+            env = self.fsdp_envs[fsdp_version].copy()
+            with patch_environment(**env):
+                fsdp_plugin = FullyShardedDataParallelPlugin(
+                    context_parallel_shard_rotation=context_parallel_shard_rotation
+                )
+                assert fsdp_plugin.context_parallel_shard_rotation == context_parallel_shard_rotation
+
```
SunMarc: nice let's also add some slow tests to check if the data is correctly handled + test if it works with 2 gpus

SunMarc (commented):
Really nice integration ! Btw this should be compatible with other parallelisms oob (TP/PP) as described in this [doc](https://discuss.pytorch.org/t/distributed-w-torchtitan-breaking-barriers-training-long-context-llms-with-1m-sequence-length-in-pytorch-using-context-parallel/215082) , not sure if this is the best way forward to attach it to fsdpv2 this tightly. Maybe we should create a `cp_config` instead ?

src/accelerate/accelerator.py
```diff
@@ -1464,6 +1523,39 @@ def _prepare_fsdp2(self, *args):
...
+
+            device_mesh = init_device_mesh(
+                device_type=self.device.type,
+                mesh_shape=(dp_shard_size, context_parallel_size),
+                mesh_dim_names=("dp_shard", "cp"),
+            )
+            device_mesh["dp_shard", "cp"]._flatten("dp_shard_cp")
+            _fully_shard_kwargs["mesh"] = device_mesh[
+                "dp_shard_cp"
+            ]  # apply fully_shard to a joint mesh of cp and dp_shard
```
SunMarc: since we are flattening the whole mesh, do we still need to pass this to fsdpv2 ?

examples/fsdp2/fsdp2_context_parallel.py
S1ro1: Yes, I just realised what you mean by this. Will have to check how this is exactly handled. Right now, it is not correct (IMO).

examples/fsdp2/fsdp2_context_parallel.py
```diff
@@ -0,0 +1,176 @@
...
+        accelerator: Accelerator instance
+
+    Returns:
+        loss: Training loss
+    """
+    input_ids = batch["input_ids"]
+    labels = batch["labels"]
+
+    # Use context parallel for efficient long sequence processing
+    with accelerator.context_parallel(
```
yh8899, line 61: This is enough for LLMs. But for DiT, like Wan2.1, they have patch embeding(conv3d), Dtensor haven't support conv3d as i know. Can we change this context for transformer blocks?

S1ro1: We can probably explore these options later, I'm not too familiar with vision models so I can't be of help on that much. How would you see this work for models with ops that aren't dtensor supported?

src/accelerate/accelerator.py
S1ro1: We use | None most of the time, staying consistent with that here.

src/accelerate/utils/dataclasses.py
S1ro1: We use a pattern like this quite often: we keep as None, then check from env and default to default. I don't want to really change it part by part. I can take a look at refactoring it later.

examples/fsdp2/README.md
```diff
@@ -34,3 +34,25 @@ The figures above were generated on 8x H100 SXM GPUs, with 8192 sequence length
...
 ```
+
+## FSDP2 + context parallelism (`fsdp2_context_parallel.py`)
+
+In this file, we showcase integration of context parallelism with FSDP2. Context parallelism is a technique that allows us to scale the training to sequence length of up to a million tokens. With `accelerator.context_parallel` context manager, we replace the attention implementation with a context parallel version, which enables us to train on a sequence length of up to 128k tokens on 8x H100 GPUs, with possibility of endless scaling if we have enough GPUs.
+
+For a detailed explanation and more details, please refer to [this guide](https://huggingface.co/docs/accelerate/concept_guides/context_parallel). You can run the example with the following command:
+
+```bash
+accelerate launch --fsdp2_context_parallel.py --sequence_length 128000 --num_steps 1000 --log_with wandb --cp_size 8 --cp_comm_strategy allgather
```
qgallouedec:
```suggestion
accelerate launch fsdp2_context_parallel.py --sequence_length 128000 --num_steps 1000 --log_with wandb --cp_size 8 --cp_comm_strategy allgather
```

kashif:
```suggestion
accelerate launch --num_processes 8 fsdp2_context_parallel.py --sequence-length 128000 --num-steps 1000 --log-with wandb --cp-size 8 --cp-comm-strategy allgather
```

PR #3608, opened 2025-06-06
remove ipex.optimize in accelerate

yao-matrix:
Most of the optimizations were upstreamed to PyTorch, and new optimizations will directly upstream to PyTorch directly, so retire `ipex.optimize` from `accelerate`.


yao-matrix:
@sywangyi @jiqing-feng @[kaixuanliu](https://github.com/kaixuanliu), pls help review, thx

yao-matrix:
@SunMarc , pls help review, thx very much

docs/source/usage_guides/intel_cpu.md
```diff
@@ -156,27 +126,29 @@ use_cpu: true
...
 oneccl_bindings_for_pytorch_path=$(python -c "from oneccl_bindings_for_pytorch import cwd; print(cwd)")
 source $oneccl_bindings_for_pytorch_path/env/setvars.sh
 
 accelerate launch examples/nlp_example.py
 ```
-Otherwise, if you selected not to have Accelerate launch `mpirun`, run the following command in node0 and **16DDP** will
-be enabled in node0,node1,node2,node3 with BF16 mixed precision. When using this method, the python script, python
-environment, and accelerate config file need to be present on all of the machines used for multi-CPU training.
+
+You can also directly luanch distributed training with `mpirun` command, you need to run the following command in node0 and **16DDP** will be enabled in node0,node1,node2,node3 with BF16 mixed precision. When using this method, the python script, python environment, and accelerate config file need to be available on all of the machines used for multi-CPU training.
```
kaixuanliu: small spelling mistake, 'luanch'->'launch'

yao-matrix: done, thx

SunMarc (commented):
I don't mind removing ipex from accelerate since it was finally upstreamed to pytorch. However, I think it would be nice to have a deprecation cycle of 2 versions at least and add more information on which torch version is needed instead

yao-matrix:
> I don't mind removing ipex from accelerate since it was finally upstreamed to pytorch. However, I think it would be nice to have a deprecation cycle of 2 versions at least and add more information on which torch version is needed instead

Thx @SunMarc. I am using to keep the `ipex` options and add a future warning in `accelerate launch`, and ignore it(i.e. don't call `ipex.optimize` anymore), is it what in your mind? Or you think we need keep calling `ipex.optimize` in the deprecation window? Thx very much.
```
    if args.ipex is not None:
        logger.warning(
            "ipex flag is deprecated and ignored, will be removed in Accelerate v1.20. "
            "From 2.7, PyTorch has all needed optimizations for Intel CPU and XPU."
        )
```

SunMarc:
> Or you think we need keep calling ipex.optimize in the deprecation window? Thx very much.

Yeah, I think we should still call `ipex.optimize` in the deprecation window. Or what we could do is for users that have torch >= 2.7, we don't use `ipex.optimize` otherwise, we keep using it until 2.7 gets deprecated WDYT ? We can add a warning msg for those who have <=2.7 to nudge them to upgrade torch. Also, in theory, should we expect any difference with ipex.optimize or with torch >=2.7 ?

yao-matrix:
> > Or you think we need keep calling ipex.optimize in the deprecation window? Thx very much.
> 
> Yeah, I think we should still call `ipex.optimize` in the deprecation window. Or what we could do is for users that have torch >= 2.7, we don't use `ipex.optimize` otherwise, we keep using it until 2.7 gets deprecated WDYT ? We can add a warning msg for those who have <=2.7 to nudge them to upgrade torch. Also, in theory, should we expect any difference with ipex.optimize or with torch >=2.7 ?

we don't expect difference since most of optimizations are and will continuously upstream to PyTorch. And we plan to deprecate `ipex.optimize` API after top-of-stack libraries(like HF) done the deprecation. Let me do per your suggestion.

yao-matrix:
@SunMarc , done, thx.

src/accelerate/accelerator.py
```diff
@@ -1422,7 +1422,12 @@ def prepare(self, *args, device_placement=None):
...
 
         if self.distributed_type in [DistributedType.MULTI_CPU, DistributedType.MULTI_XPU, DistributedType.NO]:
-            if (self.device.type == "cpu" or self.device.type == "xpu") and self.state.use_ipex:
+            if (
+                is_torch_version("<", "2.7.0")
+                and (self.device.type == "cpu" or self.device.type == "xpu")
+                and self.state.use_ipex
+            ):
+                logger.warning("You are using lower version of PyTorch(< 2.7.0) with ipex acceleration on Intel CPU or XPU, Intel has upstreamed most of the optimizations into stock PyTorch from 2.7.0, we enourage you to install the latest stock PyTorch and enjoy the out-of-experience on Intel CPU/XPU.")
                 args = self._prepare_ipex(*args)
```
SunMarc: maybe also add a `logger.warning` in case the user specify `use_ipex` with `torch >= 2.7.0`, saying that this flag is not needed anymore as optimization is upstreamed to `torch` ( meaning that we won't be using ipex) + to remove the warning, he just need to not specify `use_ipex`. 

Also is there a way to capture that we are dealing with intel cpu ? I don't want to show warning to anyone using a cpu.

docs/source/package_reference/cli.md
```diff
@@ -139,7 +139,7 @@ values. They can also be passed in manually.
 * `--cpu` (`bool`) -- Whether or not to force the training on the CPU.
 * `--multi_gpu` (`bool`) -- Whether or not this should launch a distributed GPU training.
 * `--tpu` (`bool`) -- Whether or not this should launch a TPU training.
-* `--ipex` (`bool`) -- Whether or not this should launch an Intel Pytorch Extension (IPEX) training.
+* `--ipex` (`bool`) -- Whether or not this should launch an Intel Pytorch Extension (IPEX) training. **This argument is deprecated, will be removed in Accelerate v1.10**
```
SunMarc, line 142: So you still want to deprecate the `ipex` flag and force the user to use `torch 2.7` onwards ? Note that this won't be possible for device.type == "cpu" as this might be too breaking for all the users. For xpu, we can potentially enforce that.

SunMarc (commented):
Thanks for iterating ! Left some comments !

src/accelerate/accelerator.py
yao-matrix: @SunMarc actually I'm deprecating ipex both for cpu and xpu cases, cpu is the same case as xpu. :)

docs/source/package_reference/cli.md
yao-matrix: @SunMarc actually I'm deprecating ipex both for cpu and xpu cases, cpu is the same case as xpu. ipex has done historical mission as Intel's upstreaming staging package for cpu and xpu, and will retire. If we have API/user interface backward-compatibility concern, could we do like this: still keep this flag but ignore it in the code, which means the legacy code can still pass the arg in w/o breaking their legacy code but this arg will take no effect.

SunMarc: > actually I'm deprecating ipex both for cpu and xpu cases

Ok fine with me ! Let's do the following as you want to deprecate ipex once for all: 
1) merge the current PR -> users with torch < 2.7 will still use ipex 
2) after two releases : 
- remove `prepare_ipex` + we `use_ipex` flag to raise an error to old users to switch to `torch 2.7` instead of the current warning
Then after a while maybe we can totally remove `use_ipex` flag 

WDTY ?

yao-matrix: Your plan is perfect to me, will follow your plan.

SunMarc (approved):
Thanks for iterating !

SunMarc:
@bot /style

PR #3611, opened 2025-06-06
better handle FP8 with and without deepspeed

IlyasMoutawwakil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1508,7 +1508,7 @@ class Accelerator:
                 model.forward = MethodType(convert_outputs_to_fp32(model.forward.__func__), model)
 
         # We prepare TE after, allowing for bf16 autocast to happen first
-        if self.fp8_backend == "TE" and not self.delayed_fp8_autocast:
+        if not self.delayed_fp8_autocast:
             model = apply_fp8_autowrap(model, self.te_recipe_handler or self.fp8_recipe_handler)
 
         if (getattr(model, "is_loaded_in_8bit", False) or getattr(model, "is_loaded_in_4bit", False)) and getattr(
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -462,10 +462,9 @@ class Accelerator:
             **kwargs,
         )
 
-        self._mixed_precision = mixed_precision
         # Check for automatic FP8 recipe creation
-        if self._mixed_precision == "fp8" and not self.has_fp8_handler:
-            # Prioritize TE -> AO -> MSAMP
+        if self.state.mixed_precision == "fp8" and not self.has_fp8_handler:
+            # Prioritize AO -> TE -> MSAMP
             if is_torchao_available():
                 logger.info("Found `torchao` installed, using it for FP8 training.")
                 self.ao_recipe_handler = AORecipeKwargs()
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -354,18 +354,6 @@ class Accelerator:
             elif compare_versions("deepspeed", "<", "0.9.3"):
                 raise ImportError("DeepSpeed version must be >= 0.9.3. Please update DeepSpeed.")
 
-            mixed_precision = (
-                os.environ.get("ACCELERATE_MIXED_PRECISION", "no") if mixed_precision is None else mixed_precision
-            )
-            if not isinstance(deepspeed_plugins, dict):
-                deepspeed_plugins.set_mixed_precision(mixed_precision)
-                deepspeed_plugins.select(_from_accelerator_state=True)
-            else:
-                for plugin in deepspeed_plugins.values():
-                    plugin.set_mixed_precision(mixed_precision)
-                # The first plugin passed in is always the active one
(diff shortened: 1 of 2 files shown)


src/accelerate/accelerator.py
```diff
@@ -462,10 +462,9 @@ def __init__(
             **kwargs,
         )
 
-        self._mixed_precision = mixed_precision
         # Check for automatic FP8 recipe creation
-        if self._mixed_precision == "fp8" and not self.has_fp8_handler:
-            # Prioritize TE -> AO -> MSAMP
+        if self.state.mixed_precision == "fp8" and not self.has_fp8_handler:
+            # Prioritize AO -> TE -> MSAMP
```
SunMarc: for deepspeed case 

```suggestion
        if (self.state.mixed_precision == "fp8" or mixed_precision== "fp8) and not self.has_fp8_handler:
            # Prioritize AO -> TE -> MSAMP
```

SunMarc (approved):
LGTM, just a nit

src/accelerate/accelerator.py
IlyasMoutawwakil: oh yes this totally makes sense !

src/accelerate/accelerator.py
```diff
@@ -462,10 +462,9 @@ def __init__(
             **kwargs,
         )
 
-        self._mixed_precision = mixed_precision
         # Check for automatic FP8 recipe creation
-        if self._mixed_precision == "fp8" and not self.has_fp8_handler:
-            # Prioritize TE -> AO -> MSAMP
+        if (self.state.mixed_precision == "fp8" or mixed_precision== "fp8) and not self.has_fp8_handler:
```
IlyasMoutawwakil:
```suggestion
        if (self.state.mixed_precision == "fp8" or mixed_precision == "fp8") and not self.has_fp8_handler:
```

src/accelerate/accelerator.py
```diff
@@ -480,6 +469,7 @@ def __init__(
                     "Tried to train with `fp8` and auto-detect backend, but no FP8-compatible backend was installed. "
                     "Valid backends are: `torchao`, `transformer-engine`, and `msamp`."
                 )
+            self.has_fp8_handler = True
```
IlyasMoutawwakil, line 472: idk how it worked before but this was missing 😅

IlyasMoutawwakil:
> Not too sure of the DeepSpeed related changes, rest looks fine.

looking at the structure of the accelerator state, I guess moving its mixed precision logic there makes more sense and avoids reloading the env var (see fsdp logic which does the same).

Initial deepspeed tests are passing, I'm currently running the slow ones.

src/accelerate/accelerator.py
IlyasMoutawwakil: if `has_fp8_handler` stays false, the property `fp8_backend` becomes erroneous as it can only return MSAMP or None.

S1ro1:
> > Not too sure of the DeepSpeed related changes, rest looks fine.
> 
> looking at the structure of the accelerator state, I guess moving its mixed precision logic there makes more sense and avoids reloading the env var (see fsdp logic which does the same).
> 
> Initial deepspeed tests are passing, I'm currently running the slow ones.

Yes, I think we want to slowly move into having logic local to the plugins. This centralised handling caused many issues i.e. with FSDP2 (see #3585 which did the same basically), it introduces a bit of copy-paste, but gives us a lot more freedom to make stuff work properly. Just noticed that Marc already reviewed, so just lmk when it's ready to merge.

src/accelerate/accelerator.py
SunMarc: Yeah indeed ;)

src/accelerate/state.py
```diff
@@ -954,8 +954,17 @@ def __init__(
...
+                if not isinstance(deepspeed_plugin, dict):
+                    deepspeed_plugin.set_mixed_precision(mixed_precision)
+                    deepspeed_plugin.select(_from_accelerator_state=True)
+                else:
+                    for plugin in deepspeed_plugin.values():
+                        plugin.set_mixed_precision(mixed_precision)
+                    # The first plugin passed in is always the active one
+                    first_plugin = next(iter(deepspeed_plugin.values()))
+                    first_plugin.select(_from_accelerator_state=True)
+                self.deepspeed_plugins = deepspeed_plugin
```
SunMarc, line 967: Indeed that's better to put everything in state. Not sure why it was handled both in accelerator and in the state. 
Found this PR https://github.com/huggingface/accelerate/pull/432/files but the change should be safe. Just make sure that the tests about multi models + deepspeed works as expected

IlyasMoutawwakil: all tests passed 🤗

src/accelerate/accelerator.py
IlyasMoutawwakil: should I add some test for different expected values of fp8_backend ?

src/accelerate/state.py
SunMarc: merging !

PR #3613, opened 2025-06-07
warn for invalid keys

### LINKED ISSUE #3584: [suggestion] potentially add warning with "invalid" `device_map` keys
Hello dear accelerate team,

thanks for all your hard work on this amazing package, I am having truly the best time playing around with it. 

I wanted to suggest one small enhancement that might make the experience for new users a bit nicer:

In the attached examples, I am trying to dispatch a model myself with a dummy device map. As I now understand, this will obviously not work because the device map specifies tensors (e.g. `weight`, `bias`) and not torch modules. Do you think it would make sense to add a warning for any `device_map` entries that are not actually `torch.nn.Module`s in the given input module?

Thanks a lot again! If you feel like this is useful, I am glad to help with a PR on this.

```python
import torch
import torch.nn as nn
from accelerate import dispatch_model
from accelerate.utils.modeling import clean_device_map
from collections import OrderedDict

class MLP(nn.Module):    
    def __init__(self, input_size, hidden_size, output_size):
        super(MLP, self).__init__()
        self.layer1 = nn.Linear(input_size, hidden_size)
        self.relu = nn.ReLU()
        self.layer2 = nn.Linear(hidden_size, hidden_size)
    
    def forward(self, x):
        x = self.layer1(x)
        x = self.relu(x)
        x = self.layer2(x)
        return x

# Create a random input tensor
batch_size, input_size, hidden_size, output_size = 32, 10, 20, 5

# Initialize the MLP
model = MLP(input_size, hidden_size, output_size)
model.to("cpu")
print("Model is now on CPU")

# Generate random input data and perform forward pass
X = torch.randn(batch_size, input_size)
output = model(X)
print("CPU inference successful.")

device_map = {k: 0 if "layer1" in k else 1 for k in model.state_dict()}
print("Device Map: ",device_map)
print("Type of Device Map: ", type(device_map))

# Dispatch the model according to the device map
model = dispatch_model(model, device_map=device_map)
print("Model devices:", model.layer1.weight.device, model.layer2.weight.device)

# Forward pass with the dispatched model
try:
    output = model(X)
    print("Dispatched inference successful.")
except Exception as e:
    print("Dispatched inference unsuccessful: ", e)
```
Output:
```
Model is now on CPU
CPU inference successful.
Device Map:  {'layer1.weight': 0, 'layer1.bias': 0, 'layer2.weight': 1, 'layer2.bias': 1}
Type of Device Map:  <class 'dict'>
Model devices: cpu cpu
Dispatched inference unsuccessful:  Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
```
accelerate version: 1.7

ved1beta:
# What does this PR do?

add's warning for invalid keys in check_device_map as discussed in #3584 

Fixes #3584

## Before submitting
      Pull Request section?
- [X] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.

## Who can review?
- Big modeling: @SunMarc

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1594,6 +1594,12 @@ def check_device_map(model: nn.Module, device_map: dict[str, Union[int, str, tor
         model (`torch.nn.Module`): The model to check the device map against.
         device_map (`Dict[str, Union[int, str, torch.device]]`): The device map to check.
     """
+    all_module_names = dict(model.named_modules())
+    invalid_keys = [k for k in device_map if k != "" and k not in all_module_names]
+
+    if invalid_keys:
+        print(f"Warning: The following device_map keys do not match any submodules in the model: {invalid_keys}")
+
     all_model_tensors = [name for name, _ in model.state_dict().items()]
     for module_name in device_map.keys():
         if module_name == "":
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1599,8 +1599,7 @@ def check_device_map(model: nn.Module, device_map: dict[str, Union[int, str, tor
 
     if invalid_keys:
         warnings.warn(
-            f"The following device_map keys do not match any submodules in the model: {invalid_keys}",
-            UserWarning
+            f"The following device_map keys do not match any submodules in the model: {invalid_keys}", UserWarning
         )
 
     all_model_tensors = [name for name, _ in model.state_dict().items()]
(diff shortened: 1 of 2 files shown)


SunMarc (commented):
Thanks @ved1beta  ! Can you add a test for that ?

SunMarc:
@bot /style

PR #3626, opened 2025-06-12
cp dataloader 

SunMarc:
# What does this PR do?

To try CP support for dataloader. Make sure to set `dispatch_batches` to False and `split_batches` to False in accelerate config 

cc @qgallouedec

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -1137,6 +1137,11 @@ def prepare_data_loader(
             process_index = process_index // submesh_tp_size
             num_processes = submesh_fsdp_size * submesh_dp_size
 
+    # put cp size here
+    cp_size = 
+    process_index = process_index // cp_size
+    num_processes = num_processes // cp_size
+    
     # Sanity check
     if split_batches:
         if dataloader.batch_size is not None:


src/accelerate/data_loader.py
```diff
@@ -1137,6 +1138,10 @@ def prepare_data_loader(
             process_index = process_index // submesh_tp_size
             num_processes = submesh_fsdp_size * submesh_dp_size
 
+    if cp:
+        process_index = 0
+        num_processes = 1
+
```
winglian, line 1144: Does only 1 process break n-d parallel? Maybe something like?
```suggestion
            process_index = process_index // (submesh_tp_size * submesh_cp_size)
            num_processes = submesh_fsdp_size * submesh_dp_size // (submesh_tp_size * submesh_cp_size)

    if cp:
        process_index = 0
        num_processes = 1
    
```

SunMarc: indeed we will have something like that. I just opened this PR to not forget about this but we will upstream the changes to main in another pr when n-d parallelism pr will be finished.

SunMarc:
this should cover this PR https://github.com/huggingface/accelerate/pull/3682

Issue #3628, opened 2025-06-14
[rank0]: AssertionError: found no DeviceMesh from dtensor args for c10d.broadcast_.default!


lucasjinreal:
A simple accelerate code got error

policy, optimizer, dataloader, lr_scheduler = accelerator.prepare(
        policy, optimizer, dataloader, lr_scheduler
    )

policy is model object which loaded sucessfully, but can not be prepared for multi GPU training. help wanted!

S1ro1:
Could you provide a bit more info? Config/plugin you used, code snippet we could try? Thanks

lucasjinreal:
@S1ro1 Sure. I had debugged a little bit, using raw DDP written didn't work as well with torch2.7, but still wanna accelerate take a look for a solution.

accelerate code:

```
@parser.wrap()
def train(cfg: TrainPipelineConfig):
    cfg.validate()
    logging.info(pformat(cfg.to_dict()))

    # Initialize accelerator
    from accelerate.utils import DistributedDataParallelKwargs

    from lerobot.common.utils.wandb_utils import cfg_to_group, get_wandb_run_id_from_filesystem

    ddp_kwargs = DistributedDataParallelKwargs(find_unused_parameters=True)
    accelerator = Accelerator(
        mixed_precision="fp16" if cfg.policy.use_amp else "no",
        gradient_accumulation_steps=cfg.policy.gradient_accumulation_steps,
        log_with="wandb" if cfg.wandb.enable else None,
        kwargs_handlers=[ddp_kwargs],
        project_dir=cfg.output_dir,
    )

    accelerator.init_trackers(
        project_name=cfg.wandb.project,
        init_kwargs={
            "wandb": {
                "entity": cfg.wandb.entity,
                "name": cfg.job_name,
                "notes": cfg.wandb.notes,
                "tags": cfg_to_group(cfg, return_list=True),
                "dir": cfg.output_dir,
                "config": cfg.to_dict(),
                "save_code": False,
                "job_type": "train_eval",
                "mode": cfg.wandb.mode if cfg.wandb.mode in ["online", "offline", "disabled"] else "online",
                "resume": "must" if cfg.resume else None,
                "id": cfg.wandb.run_id
                if cfg.wandb.run_id
                else (get_wandb_run_id_from_filesystem(cfg.output_dir) if cfg.resume else None),
            }
        },
    )

    # Set seed for reproducibility
    if cfg.seed is not None:
        accelerate_set_seed(cfg.seed)

[...]

PR #3629, opened 2025-06-15
Fix: properly error when DDP + Dtensor model

S1ro1:
A rather tiny PR, "fixes" or more likely errors properly on an issue introduced in #3628. The issue is that DDP doesn't work if model has DTensors which it has if it was loaded with device_map="auto" (it applies TP). Now we properly raise an Error.

Snippet to test:

```python

model_id = "NousResearch/Hermes-3-Llama-3.1-8B"
accelerator = Accelerator()

model = AutoModelForCausalLM.from_pretrained(model_id, device_map="auto")

```

Before: 
`[rank0]: AssertionError: found no DeviceMesh from dtensor args for c10d.broadcast_.default!`

After:
`[rank0]: ValueError: Your model contains `DTensor` parameters, which is incompatible with DDP. Maybe you loaded your model with `device_map='auto'`?`

cc @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -108,6 +108,7 @@ from .utils import (
     is_xpu_available,
     load_fsdp_model,
     load_fsdp_optimizer,
+    model_has_dtensor,
     pad_across_processes,
     parse_choice_from_env,
     recursively_apply,
@@ -1620,6 +1621,10 @@ class Accelerator:
                 DistributedType.MULTI_XPU,
                 DistributedType.MULTI_HPU,
             ):
+                if model_has_dtensor(model):
+                    raise ValueError(
+                        "Your model contains `DTensor` parameters, which is incompatible with DDP. Maybe you loaded your model with `device_map='auto'`?"
+                    )
                 if any(p.requires_grad for p in model.parameters()):
                     kwargs = self.ddp_handler.to_kwargs() if self.ddp_handler is not None else {}
                     # TODO: Look at enabling native TP training directly with a proper config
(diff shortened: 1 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -1620,6 +1621,10 @@ def prepare_model(self, model: torch.nn.Module, device_placement: bool = None, e
                 DistributedType.MULTI_XPU,
                 DistributedType.MULTI_HPU,
             ):
+                if model_has_dtensor(model):
+                    raise ValueError(
+                        "Your model contains `DTensor` parameters, which is incompatible with DDP. Maybe you loaded your model with `device_map='auto'`?"
+                    )
```
SunMarc: Would be also nice to include a way to fix this -> specify `device_map = {"": torch.cuda.current_device()}` for example

src/accelerate/utils/other.py
```diff
@@ -194,6 +194,33 @@ def compile_regions_deepspeed(module: torch.nn.Module, **compile_kwargs):
...
+    """
+    dtensor_available = False
+    # We have to do this disgusting import dance because DTensor changed location in 2.5.0
+    try:
+        from torch.distributed.tensor import DTensor
+
+        dtensor_available = True
+    except ImportError:
+        try:
+            from torch.distributed._tensor import DTensor
```
SunMarc: let's just check torch version then, i'm not a fan of try except here

SunMarc (commented):
Thanks a lot for that ! Just some nits

PR #3635, opened 2025-06-18
feat: use datasets.IterableDataset shard if possible

SunMarc:
# What does this PR do?

Supersedes https://github.com/huggingface/accelerate/pull/3583

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -1198,7 +1198,8 @@ def prepare_data_loader(
         if is_datasets_available():
             from datasets import IterableDataset as DatasetsIterableDataset
         if (
-            is_datasets_available() and isinstance(new_dataset, DatasetsIterableDataset) 
+            is_datasets_available()
+            and isinstance(new_dataset, DatasetsIterableDataset)
             and not split_batches
             and new_dataset.n_shard > num_processes
         ):


TonyCongqianWang:
IterableDataset does not have the attribute "n_shard" but "n_shards" instead. See:

https://github.com/huggingface/datasets/blob/3.6.0/src/datasets/iterable_dataset.py#L1947

SunMarc:
thanks for noticing, will be included in a patch release soon !

dongruliu:
@SunMarc I notice that it will still use the IterableDatasetShard when i use datasets.IterableDataset and n_shards = num_processes. Then it will yield out the sub samples of the corresponding rank, which is not the expected behavior. As the n_shards > 1 and equals to num_processese, each shard should be separated to each rank and yield out all the samples

image

SunMarc:
The check above `isinstance(new_dataset, DatasetIterableDataset)` is False ?

dongruliu:
> The check above `isinstance(new_dataset, DatasetIterableDataset)` is False ?

Both isinstance(new_dataset, DatasetsIterableDataset) and isinstance(new_dataset, IterableDataset) are True.
But new_dataset.n_shard is same as num_processes, instead of new_dataset.n_shard > num_processes, so it goes to the elif and uses the IterableDatasetShard.

SunMarc:
Will this solves your case ? https://github.com/huggingface/accelerate/pull/3958

dongruliu:
> Will this solves your case ? #3958

Yes. It solves the issue. Thanks for helping create the PR!

PR #3637, opened 2025-06-19
Fix: correct labels for fsdp2 examples

S1ro1:
Very small PR, doesn't change anything functionality-wise, I just got bit by LLM generated code and my lack of attention and was passing labels even though they were shifted. Explicitly pass shifted labels now, doesn't change ANYTHING in the benchmark.


SunMarc (approved):
SG !

PR #3654, opened 2025-06-25
xpu enablement on left cases 

yao-matrix:
1. enable xpu for launcher -> validated
2. expand cuda only ds uts to xpu -> all 3 passed
3. expand profiler example to xpu -> validated

@SunMarc , pls help review, thx.

### DIFF
--- a/src/accelerate/launchers.py
+++ b/src/accelerate/launchers.py
@@ -60,8 +60,8 @@ def notebook_launcher(
 
     <Tip warning={true}>
 
-    To use this function absolutely zero calls to an accelerator must be made in the notebook session before calling. If
-    any have been made, you will need to restart the notebook and make sure no cells use any accelerator capability.
+    To use this function absolutely zero calls to an accelerator must be made in the notebook session before calling.
+    If any have been made, you will need to restart the notebook and make sure no cells use any accelerator capability.
 
     Setting `ACCELERATE_DEBUG_MODE="1"` in your environment will run a test before truly launching to ensure that none
     of those calls have been made.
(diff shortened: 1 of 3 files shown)


src/accelerate/launchers.py
```diff
@@ -60,8 +60,8 @@ def notebook_launcher(
 
     <Tip warning={true}>
 
-    To use this function absolutely zero calls to a CUDA device must be made in the notebook session before calling. If
-    any have been made, you will need to restart the notebook and make sure no cells use any CUDA capability.
+    To use this function absolutely zero calls to an accelerator must be made in the notebook session before calling.
```
S1ro1: NIT: I would maybe try to find better wording as `accelerator` - maybe `device`. Since the whole accelerate is built on top `Accelerator` class. If you had the same naming convention even before, we can keep it.

yao-matrix: done, thx, the original thought to use `accelerator` is to align w/ PyTorch's vocabulary. But yes, it's confusing in `accelerate` since it has an `Accelerator` class.

yao-matrix: @S1ro1 indeed, especially in context that the src code already uses `Accelerator` instance. Done, thx.

yao-matrix:
@S1ro1 @SunMarc , could you pls review again? Thx very much.

yao-matrix:
@SunMarc , could you pls take a look? thx very much.

SunMarc (approved):
Thanks !

SunMarc:
@bot /style

PR #3671, opened 2025-07-07
Bunch of FSDP improvements

S1ro1:
First I intended just to make the tests better, but I found some issues with the underlying code so I'm fixing that here too (can split if needed), most of it related to state loading, which somehow didn't get caught in the CI

1) remove dynamic FSDP test shenanigans (tests/fsdp/test_fsdp.py), we now just inherit the class and run like that
2) Scaler loading broken with FSDP2 because scale wasn't set correctly
3) broadcast_from_rank_0 requires all ranks to get into the distributed section with FSDP2 model loading, that wasn't the case as with `sync_module_states` which we used in FSDP1 we just exit on non-main ranks.

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -181,7 +181,7 @@ def load_fsdp_model(fsdp_plugin, accelerator, model, input_dir, model_index=0, a
     sd_options = _prepare_sd_options(fsdp_plugin)
     with ctx:
         if fsdp_plugin.state_dict_type == StateDictType.FULL_STATE_DICT:
-            if type(model) is not FSDP and accelerator.process_index != 0:
+            if type(model) is not FSDP and accelerator.process_index != 0 and not accelerator.is_fsdp2:
                 if not fsdp_plugin.sync_module_states and fsdp_plugin.fsdp_version == 1:
                     raise ValueError(
                         "Set the `sync_module_states` flag to `True` so that model states are synced across processes when "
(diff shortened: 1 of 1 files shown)


src/accelerate/state.py
```diff
@@ -213,12 +213,6 @@ def __init__(self, cpu: bool = False, **kwargs):
                             if self.backend == "tccl":
                                 local_rank = os.environ.get("LOCAL_RANK", -1)
                                 torch.sdaa.set_device(f"sdaa:{local_rank}")
-                            if (
```
S1ro1, line 216: This had to be moved, as it was in a `deepspeed` branch before, which was wrong

src/accelerate/utils/fsdp_utils.py
```diff
@@ -179,10 +179,11 @@ def load_fsdp_model(fsdp_plugin, accelerator, model, input_dir, model_index=0, a
         else nullcontext()
     )
     sd_options = _prepare_sd_options(fsdp_plugin)
-
     with ctx:
         if fsdp_plugin.state_dict_type == StateDictType.FULL_STATE_DICT:
-            if type(model) is not FSDP and accelerator.process_index != 0:
+            if (
+                type(model) is not FSDP and not isinstance(model, torch.distributed.fsdp.FSDPModule)
```
S1ro1: In here, we actually early exit if FSDP1, which shouldn't be the case in FSDP2 as the `set_model_state_dict` timeouts

PR #3674, opened 2025-07-10
Speedup model loading by 4-5x in Diffusers ⚡

a-r-r-o-w:
Sister PR in diffusers: https://github.com/huggingface/diffusers/pull/11904

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -295,9 +295,9 @@ def set_module_tensor_to_device(
 
         if dtype is None:
             # For compatibility with PyTorch load_state_dict which converts state dict dtype to existing dtype in model
-            value = value.to(old_value.dtype)
+            value = value.to(old_value.dtype, non_blocking=True)
         elif not str(value.dtype).startswith(("torch.uint", "torch.int", "torch.bool")):
-            value = value.to(dtype)
+            value = value.to(dtype, non_blocking=True)
 
     device_quantization = None
     with torch.no_grad():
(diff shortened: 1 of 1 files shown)


src/accelerate/utils/memory.py
```diff
@@ -138,11 +138,10 @@ def find_executable_batch_size(
 
 
     >>> @find_executable_batch_size(starting_batch_size=128)
-    ... def train(batch_size, model, optimizer):
```
a-r-r-o-w: Change is unrelated. Seems to come from ruff, and I'm not sure why. I'm using the ruff version from the setup.py

src/accelerate/utils/modeling.py
```diff
@@ -222,6 +222,8 @@ def set_module_tensor_to_device(
     dtype: Optional[Union[str, torch.dtype]] = None,
     fp16_statistics: Optional[torch.HalfTensor] = None,
     tied_params_map: Optional[dict[int, dict[torch.device, torch.Tensor]]] = None,
+    non_blocking: bool = False,
+    _empty_cache: bool = True,
```
SunMarc: can you update the docstring also ? Also, maybe `empty_cache` or `clear_cache`  ?

SunMarc (approved):
Thanks for this ! A minor suggestion

src/accelerate/utils/modeling.py
a-r-r-o-w: @SunMarc Thanks, updated! Could you take a look again?

PR #3677, opened 2025-07-10
Fix FP8 tests, enable FP8 to be used without direct `Accelerator()` configuring

pstjohn:
# What does this PR do?

It looks like the [current fp8 tests are not passing](https://github.com/huggingface/accelerate/actions/workflows/fp8_runner.yml), this PR slightly refactors these tests and makes a few fixes to get them to pass.

It also adds (and fixes) new tests that ensure FP8 functionality can be configured entirely from an accelerate config yaml in the cases where the user doesn't have the ability to change how the `Accelerator()` object is created. This currently seems broken, since you need to be able to pass a `kwargs_handlers` to choose the FP8 backend. With the `transformers.Trainer` class, for instance, the `Accelerator()` object is created under-the-hood, so this should enable FP8 training with that class simply by changing the accelerator config yaml.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@muellerzr 
@zach-huggingface 

All tests pass locally with the `huggingface/accelerate:gpu-fp8-transformerengine-nightly` container after installing deepspeed, with the exception of 

```
FAILED tests/test_metrics.py::MetricTester::test_metric_accelerator_multi - RuntimeError: 'accelerate launch --num_processes=2 --monitor_interval=0.1 /workspaces/accelerate/src/accelerate/test_utils/scripts/external_deps/test_metrics.py' failed with returncode 1

stderr: [rank1]:   File "/root/.cache/huggingface/modules/evaluate_modules/metrics/evaluate-metric--glue/05234ba7acc44554edcca0978db5fa3bc600eeee66229abe79ff9887eacaf3ed/glue.py", line 84, in simple_accuracy
stderr: [rank1]:     return float((preds == labels).mean())
stderr: [rank1]:                  ^^^^^^^^^^^^^^^^^^^^^^
stderr: [rank1]: AttributeError: 'bool' object has no attribute 'mean'
```

but, that test fails in the same way for me on `main`, so I don't think it's related to this change.

### DIFF
--- a/benchmarks/fp8/transformer_engine/Dockerfile
+++ b/benchmarks/fp8/transformer_engine/Dockerfile
@@ -7,7 +7,7 @@ RUN pip install transformers evaluate datasets
 RUN git clone https://github.com/huggingface/accelerate.git
 
 RUN cd accelerate && \
-    pip install -e . && \
+    pip install -e .[deepspeed] && \
     cd benchmarks/fp8
 
 RUN /bin/bash
--- a/examples/config_yaml_templates/fp8.yaml
+++ b/examples/config_yaml_templates/fp8.yaml
@@ -11,7 +11,7 @@ fp8_config:
   fp8_format: E4M3
   interval: 1
   margin: 0
-  override_linear_precision: (false, false, false)
+  override_linear_precision: [false, false, false]
   # Generally this should always be set to `false` to have the most realistic fp8 eval performance
   use_autocast_during_eval: false
   # If using MS-AMP, we ignore all of the prior and set a opt_level
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -306,10 +306,10 @@ class Accelerator:
 
         if mixed_precision is not None:
             mixed_precision = str(mixed_precision)
-        else:
-            mixed_precision = parse_choice_from_env("ACCELERATE_MIXED_PRECISION", "no")
-        if mixed_precision not in PrecisionType:
-            raise ValueError(f"Unknown mixed_precision mode: {mixed_precision}. Choose between {PrecisionType.list()}")
+            if mixed_precision not in PrecisionType:
+                raise ValueError(
+                    f"Unknown mixed_precision mode: {mixed_precision}. Choose between {PrecisionType.list()}"
+                )
 
         if dynamo_plugin is not None and dynamo_backend is not None:
             raise ValueError("You cannot pass in both `dynamo_plugin` and `dynamo_backend`, please only pass in one.")
@@ -488,7 +488,7 @@ class Accelerator:
         self.delayed_fp8_autocast = False
         if self.has_fp8_handler:
             # We already check if FP8 is available during `self.state`
-            if mixed_precision != "fp8" and (
+            if not self.fp8_enabled and (
                 self.distributed_type not in (DistributedType.FSDP, DistributedType.DEEPSPEED)
             ):
                 raise ValueError("Passing in an FP8 configuration requires setting `mixed_precision='fp8'`.")
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -33,6 +33,8 @@ import torch
 import torch.utils.hooks as hooks
 from huggingface_hub import split_torch_state_dict_into_shards
 
+from accelerate.utils.dataclasses import FP8BackendType
+
 from .checkpointing import load_accelerator_state, load_custom_state, save_accelerator_state, save_custom_state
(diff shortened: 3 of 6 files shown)


S1ro1:
Thank you, very happy to iterate on this when I get back from vacation, until then cc @SunMarc for general design. 
> All tests pass locally with the huggingface/accelerate:gpu-fp8-transformerengine-nightly container after installing deepspeed, with the exception of

Yes, that test was broken on yesterday's CI because of a BC in `datasets`, should work now.

benchmarks/fp8/transformer_engine/Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN pip install transformers evaluate datasets
 RUN git clone https://github.com/huggingface/accelerate.git
 
 RUN cd accelerate && \
-    pip install -e . && \
+    pip install -e .[deepspeed] && \
```
pstjohn, line 10: this is the container used for fp8 CI tests, which include deepspeed. so even though we don't use it in these benchmark scripts (which we should double-check are still functional 😄), this allows the `requires_deepspeed` tests to run in the fp8 tests.

examples/config_yaml_templates/fp8.yaml
```diff
@@ -11,8 +11,8 @@ fp8_config:
   fp8_format: E4M3
   interval: 1
   margin: 0
-  override_linear_precision: (false, false, false)
+  override_linear_precision: [false, false, false]
```
pstjohn, line 14: this isn't exercised in CI anywhere but I caught this bug while using this to debug locally 🤷 . The `()` expression just gets evaluated to a string rather than a list

src/accelerate/accelerator.py
```diff
@@ -33,6 +33,8 @@
 import torch.utils.hooks as hooks
 from huggingface_hub import split_torch_state_dict_into_shards
 
+from accelerate.utils.dataclasses import FP8BackendType
```
pstjohn, line 36: we already had this enum, so I figured it was worth using this here instead of the string comparisons

src/accelerate/accelerator.py
```diff
@@ -458,27 +461,34 @@ def __init__(
 
         # Check for automatic FP8 recipe creation
         if self.fp8_enabled and not self.has_fp8_handler:
-            # Prioritize AO -> TE -> MSAMP
-            if is_torchao_available():
-                logger.info("Found `torchao` installed, using it for FP8 training.")
+            if self.fp8_backend == FP8BackendType.AO:
```
pstjohn, line 464: Here we first defer to the fp8_backend specified in the yaml, and only if we didn't specify one do we revert to the AO -> TE -> MSAMP preference.

tests/test_fp8.py
```diff
@@ -163,16 +209,26 @@ def test_can_prepare_model_multi_accelerator_deepspeed(self):
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--test_te", action="store_true", default=False)
+    parser.add_argument("--test_ao", action="store_true", default=False)
+    parser.add_argument("--from_config", action="store_true", default=False)
+    args = parser.parse_args()
+
+    if not args.test_te and not args.test_ao:
+        raise ValueError("Must specify at least one of --test_te or --test_ao")
+
+    if args.test_te:
```
pstjohn, line 227: rather than checking if TE is available twice (once when dispatching the test and once inside the test), we just check once when deciding what tests to run. This way we don't end up running the same tests twice if we have both TE and AO installed; and the failures are more fine-grained.

src/accelerate/accelerator.py
```diff
@@ -488,7 +498,11 @@ def __init__(
             )
 
         # TODO: S1ro - this is probably gonna be a problem with other fp8 backends too
-        if self.fp8_backend == "AO" and self.state.fsdp_plugin.cpu_ram_efficient_loading:
+        if (
+            self.fp8_backend == FP8BackendType.AO
+            and self.state.distributed_type == DistributedType.FSDP
+            and self.state.fsdp_plugin.cpu_ram_efficient_loading
+        ):
```
djsaunde, line 505: I was about to submit a PR that made this exact change :)

djsaunde: Related: https://github.com/huggingface/transformers/pull/39370

pstjohn: yeah IIRC this was where some of the existing fp8 tests were failing

examples/config_yaml_templates/fp8.yaml
SunMarc: nice catch

Issue #1878, opened 2023-08-23
Title: Support for non-tensor based datasets in Accelerate


hzphzp:
### System Info

```Shell
accelerate==0.21.0
python==3.9
```

### Information

- [X] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

I encountered an issue while using the Accelerate library, as it does not support non-tensor based datasets. For instance, consider the following situation:
```python
        return {  
            "image": tensor,  
            "text_input": str  
        }  
 ```
In this case, an error occurs: `Can only concatenate tensors but got <class 'str'>`. The problem lies in the following code:

[raise TypeError(f"Can only concatenate tensors but got {type(data[0])}")](https://github.com/huggingface/accelerate/blob/69e4c3c54da3201eda288b500d138761e7a5221c/src/accelerate/utils/operations.py#L365C1-L365C1)

The concat function here can only handle tensor type data, otherwise, it raises an error. 

### Expected behavior

However, in many practical applications, we often need to work with non-tensor datasets in IO operations. I would like to request support for non-tensor type datasets in Accelerate.

hzphzp:
Waiting for kindly reply ...

irowberryFS:
I'm getting the same issue. I'm working with PyTorch Geometric Data objects. According to various other threads I've read, it was supported in the past.

PR #3686, opened 2025-07-18
Concatenate str support for `IterableDataset`

### LINKED ISSUE #3624: dataloader with IterableDataset cannot work after accelerator.prepare()
### System Info

```Shell
Name: accelerate
Version: 1.7.0
Summary: Accelerate
Name: torch
Version: 2.5.1
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```
import torch
import webdataset as wds

# ---- Step 1: Initialize Accelerator ---- #
accelerator = Accelerator()
device = accelerator.device

# ---- Step 2: Define WebDataset with IterableDataset ---- #
wds_urls = ["/path/to/shard1.tar", "/path/to/shard2.tar"]  # Add your shard paths here

def custom_collate(batch):
    # Create batch as dict of lists
    collated = {}
    for sample in batch:
        for k, v in sample.items():
            if k not in collated:
                collated[k] = []
            collated[k].append(v)
    # Convert tensors to stacked tensors
    for k in collated:
        if torch.is_tensor(collated[k][0]):
            collated[k] = torch.stack(collated[k])
    return collated

wds_dataset = (
    wds.WebDataset(wds_urls, handler=wds.handlers.warn_and_continue)
    .decode("torchrgb")
    .to_tuple("mp4", "txt", "json")
    .map(lambda sample: {
        "video": sample[0],
        "text": sample[1],
        "meta": sample[2]
    })
)

# ---- Step 3: Define DataLoader ---- #
dataloader = DataLoader(
    wds_dataset,
    batch_size=32,
    num_workers=4,
    pin_memory=True,
    collate_fn=custom_collate
)

# ---- Step 4: Define model and optimizer ---- #
model = torch.nn.Linear(1000, 10)
optimizer = torch.optim.Adam(model.parameters(), lr=1e-4)

# ---- Step 5: Prepare model, optimizer, and dataloader ---- #
model, optimizer, dataloader = accelerator.prepare(model, optimizer, dataloader)

# ---- Step 6: Training Loop ---- #
for epoch in range(10):
    for batch in dataloader:
        batch = {k: v.to(device) if torch.is_tensor(v) else v for k, v in batch.items()}
```

with this code, the dataloader cannot work normally as it return TypeError: Can only concatenate tensors but got <class 'str'>, however, if i do not wrap dataloader with accelerator.prepare(), the dataloder can successfully return data.

### Expected behavior

Work normally when i use accelerator.prepare to wrap a dataloder with IterableDataset dataset.

TitouanCh:
# What does this PR do?

This PR changes concatenate so it doesn't crash when batches contain non-tensor values like strings.
Right now it seems to throw a TypeError if you had anything other than tensors, for example, string labels. 
This is pretty useful for `IterableDataset`. I also added a new test.

Fixes #3624 #1878

## Proposal

<del>When concatenating lists, check if the first element is a `str`, if so, concat as strings as a flat python list instead of tensor (I could also check all elements if you would like to avoid cases like `["test", 0] + ["test", "test] = ["test", 0, "test", "test"]`).</del>
Changed the logic to check all elements when concatenating lists. If all elements are strings, they are concatenated as a flat Python list.
If this is accepted, this can easily be adapted to support more types if needed.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@BenjaminBossan @SunMarc @zach-huggingface 

## Minimal example script showing the problem

```python
import torch
import numpy as np
from collections.abc import Mapping
from accelerate.utils import concatenate


def test_batches():
    print("=== Test 1: Tensor batches ===")
    batch1 = {
        "x": torch.rand(4, 1),
        "y": torch.from_numpy(
            np.array(
                [[1.0, 2.0, 3.0]] * 4,
                dtype=np.float32,
            )
        ),
    }

    batch2 = {
        "x": torch.rand(4, 1),
        "y": torch.from_numpy(
            np.array(
                [[1.0, 2.0, 3.0]] * 4,
                dtype=np.float32,
            )
        ),
    }

    batch = concatenate([batch1, batch2], dim=0)

    print(batch)
    print("x shape:", batch["x"].shape)  # Should be (8, 1)
    print("y shape:", batch["y"].shape)  # Should be (8, 3)

    print("\n=== Test 2: Mixed types (with lists) ===")
    batch1 = {"x": torch.rand(4, 1), "animals": ["dog", "cat", "baby", "penguin"]}
    batch2 = {
        "x": torch.rand(4, 1),
        "animals": ["koala", "samurai", "iguana", "rabbit"],
    }

    batch = concatenate([batch1, batch2], dim=0)

    print(batch)
    print("x shape:", batch["x"].shape)  # Should be (8, 1)
    print("animals:", batch["animals"])  # Should be batch1["animals"] + batch2["animals"]


if __name__ == "__main__":
    test_batches()
```
```bash
>> "src/accelerate/utils/operations.py", line 637 in concatenate
>> TypeError: Can only concatenate tensors but got <class 'str'>
```

## Example with `IterableDataset`
```python
import torch
from torch.utils.data import IterableDataset, DataLoader
import random


class SyntheticIterableDataset(IterableDataset):
    def __init__(self, num_samples, input_dim, vocab):
        self.num_samples = num_samples
        self.input_dim = input_dim
        self.vocab = vocab

    def __iter__(self):
        for _ in range(self.num_samples):
            x = torch.randint(0, 256, (self.input_dim,), dtype=torch.float32)
            y_str = random.choice(self.vocab)  # str attribute
            yield {"x": x, "y": y_str}


def custom_collate(batch):
    collated = {}
    for sample in batch:
        for k, v in sample.items():
            if k not in collated:
                collated[k] = []
            collated[k].append(v)
    for k in collated:
        if torch.is_tensor(collated[k][0]):
            collated[k] = torch.stack(collated[k])
    return collated


if __name__ == "__main__":
    # Setup model
    vocab = ["cat", "dog", "snake"]
    vocab_to_idx = {label: i for i, label in enumerate(vocab)}
    num_classes = len(vocab)
    input_dim = 1000 + num_classes

    dataset = SyntheticIterableDataset(1000, 1000, vocab=vocab)
    dataloader = DataLoader(dataset, batch_size=32, collate_fn=custom_collate)

    model = torch.nn.Linear(input_dim, 10)  # Dummy linear model
    optimizer = torch.optim.Adam(model.parameters(), lr=1e-4)

    accelerator = Accelerator()
    model, optimizer, dataloader = accelerator.prepare(model, optimizer, dataloader)
    device = accelerator.device

    for epoch in range(10):
        for batch in dataloader:
            print(batch)
            x = batch["x"].to(device)

            # y to one-hot
            y_str_list = batch["y"]
            y_idx = torch.tensor([vocab_to_idx[y] for y in y_str_list], device=device)
            y_onehot = torch.nn.functional.one_hot(
                y_idx, num_classes=num_classes
            ).float()

            model_input = torch.cat([x, y_onehot], dim=1)

            output = model(model_input)
            loss = output.sum()

            optimizer.zero_grad()
            accelerator.backward(loss)
            optimizer.step()

        print(f"Epoch {epoch} complete.")

```
```bash
>> "src/accelerate/utils/operations.py", line 637 in concatenate
>> TypeError: Can only concatenate tensors but got <class 'str'>
```

## Discussion

Could be extended to support constants? Like

```python
batch1 = {
    "const": True,
    "const_num": 1,
    "x": torch.tensor([
        [1, 2, 3],
        [2, 3, 4]
    ]),
    "y": ["c", "d"],
}

batch2 = {
    "const": True,
    "const_num": 1,
    "x": torch.tensor([
        [1, 2, 3],
        [1, 4, 5]
    ]),
    "y": ["a", "b"],
}

batch_concat = {
    "const": True,
    "const_num": 1,
    "x": torch.vstack([batch1["x"], batch2["x"]]),
    "y": batch1["y"] + batch2["y"],
}
```

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -58,7 +58,7 @@ from accelerate.utils import (
     save,
     send_to_device,
 )
-from accelerate.utils.operations import is_namedtuple, concatenate
+from accelerate.utils.operations import concatenate, is_namedtuple
 
 
 if is_torch_xla_available():
@@ -454,14 +454,13 @@ class UtilsTester(unittest.TestCase):
         batch = concatenate([batch1, batch2], dim=0)
 
         assert batch["tuple_key"][0].shape == (8, 2)
-        assert  batch["tuple_key"][1] == batch1["tuple_key"][1] + batch2["tuple_key"][1]
+        assert batch["tuple_key"][1] == batch1["tuple_key"][1] + batch2["tuple_key"][1]
 
         batch1 = {"mix": torch.rand(4, 1)}
         batch2 = {"mix": ["Basketball", "Baseball", "Surf", "Bilboquet"]}
 
         with pytest.raises(TypeError):
             concatenate([batch1, batch2], dim=0)
-
 
     def test_send_to_device_compiles(self):
         compiled_send_to_device = torch.compile(send_to_device, fullgraph=True)


src/accelerate/utils/operations.py
```diff
@@ -611,9 +615,17 @@ def concatenate(data, dim=0):
         The same data structure as `data` with all the tensors concatenated.
     """
     if isinstance(data[0], (tuple, list)):
-        return honor_type(data[0], (concatenate([d[i] for d in data], dim=dim) for i in range(len(data[0]))))
+        first_inner = data[0][0] if len(data[0]) > 0 else None
+            
+        if isinstance(first_inner, str):
+            return honor_type(data[0], [item for sublist in data for item in sublist])
+        else:
```
SunMarc: we shouldn't modify that, we should check elif isinstance(data[0], str) and do the respective changes there. Try to add for test case also involving tuple and dictionaries. Also we should check that no tensors is being concatenated with str

SunMarc (commented):
Left a comment !

src/accelerate/utils/operations.py
TitouanCh: Done!
Instead of checking for str directly, I check whether it's a list of strings, this made the logic simpler.
Let me know if that works for you, and I'm happy to explore another approach if needed.

TitouanCh:
Fixed ruff issues to pass the quality check.

TitouanCh:
up

PR #3700, opened 2025-07-31
Feat: context parallel v2.0

S1ro1:
Integrates CP seamlessly with previously merged `ParallelismConfig`. Builds on top of #3604 (which it supersedes) and so on.


S1ro1:
Supersedes #3604 as I can't bother with fixing git on that branch.

docs/source/concept_guides/context_parallelism.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+## Accelerate's interface
+
+The context manager takes a few arguments, that are used to configure the context parallelism.
+
+- `buffers`: This is a list of tensors that are to be sharded across the sequence dimension. These tensors are usually input ids, labels and attention mask.
+- `buffer_seq_dims`: This is a list of integers, that specify the sequence dimension of the buffers, in the order of the `buffers` list.
+- `no_restore_buffers`: The implementation of context parallelism modifies the buffers in-place, converting them to `torch.distributed.tensor.Dtensor`s. After the context manager is exited, a communication kernel would need to be launched to restore the buffers to their original state (usually all-gather). This takes some time, so it is recommended to pass the same tensors as in the `buffers` argument, to avoid unnecessary communication, unless you are sure that you need to use the buffers after the context manager is exited.
+
```
SunMarc: since this is recommended to pass the same tensor, can't we change the default to that ?

docs/source/concept_guides/context_parallelism.md
```diff
@@ -0,0 +1,187 @@
...
+> [!Tip]
+> These examples were created with a script you can find [in the examples folder](https://github.com/huggingface/accelerate/blob/main/examples/fsdp2/nd_parallel.py). For instructions on how to run it, see the [README](https://github.com/huggingface/accelerate/blob/main/examples/fsdp2/README.md) in the same folder.
+
+
+## Accelerate's interface
+
+The context manager takes a few arguments, that are used to configure the context parallelism.
+
+- `buffers`: This is a list of tensors that are to be sharded across the sequence dimension. These tensors are usually input ids, labels and attention mask.
+- `buffer_seq_dims`: This is a list of integers, that specify the sequence dimension of the buffers, in the order of the `buffers` list.
```
SunMarc: can we add more details to that somewhere ? some ppl might not understand what you mean by sequence dimension of the buffers. Maybe in the example ?

examples/fsdp2/nd_parallel.py
```diff
@@ -52,17 +53,24 @@ def parse_args():
...
+    input_ids, labels = batch["input_ids"], batch["labels"]
+    with accelerator.maybe_context_parallel(
+        buffers=[input_ids, labels], buffer_seq_dims=[1, 1], no_restore_buffers={input_ids, labels}
+    ):
+        # To get the proper loss value, we need to average across devices that are participating in data parallel/context parallel training
+        loss_reduce_grp = (
+            accelerator.torch_device_mesh["dp_cp"].get_group()
+            if accelerator.parallelism_config.dp_cp_dim_names
+            else None
+        )
```
SunMarc: nice use of the device mesh

examples/fsdp2/nd_parallel.py
```diff
@@ -71,21 +79,22 @@ def train(args):
...
     # FSDP needs extra configuration, so we properly shard the model
-    if parallelism_config.dp_shard_enabled:
+    fsdp2_plugin = None
+    if parallelism_config.dp_shard_enabled or parallelism_config.cp_enabled:
         fsdp2_plugin = FullyShardedDataParallelPlugin(
             fsdp_version=2,
             auto_wrap_policy="transformer_based_wrap",
             transformer_cls_names_to_wrap=["LlamaDecoderLayer"],
+            state_dict_type="SHARDED_STATE_DICT",
+            activation_checkpointing=True,
```
SunMarc: why add that ?

src/accelerate/accelerator.py
```diff
@@ -448,6 +449,12 @@ def __init__(
 
         parallelism_config = self._setup_parallelism_config(parallelism_config, torch_tp_plugin)
 
+        # TODO: Siro - figure out a better place where this can go (needs to be above AcceleratorState init)
+        if parallelism_config and parallelism_config.cp_enabled and fsdp_plugin is None:
+            raise ValueError(
+                "`cp_enabled` is set to `True` in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use `cp_enabled=True`, as we also shard the model across the device mesh to save more memory"
+            )
+
```
SunMarc: why ? Can't we put that in `_validate_accelerator` ?

src/accelerate/accelerator.py
```diff
@@ -1561,6 +1575,23 @@ def _prepare_tp(self, *args):
...
+        cp_comm_strategy = self.parallelism_config.cp_handler.cp_comm_strategy
+        set_rotate_method(cp_comm_strategy)
+
+        self._cp_context = functools.partial(context_parallel, mesh=self.torch_device_mesh["cp"])
+
+        for arg in args:
+            if not isinstance(arg, torch.nn.Module):
+                continue
+
+            attach_context_parallel_hooks(arg)
```
SunMarc:
```suggestion
        for arg in args:
            if isinstance(arg, torch.nn.Module):
                attach_context_parallel_hooks(arg)
```

src/accelerate/accelerator.py
```diff
@@ -3903,6 +3934,69 @@ def register_for_checkpointing(self, *objects):
...
+            with self._cp_context(
+                buffers=buffers, buffer_seq_dims=buffer_seq_dims, no_restore_buffers=no_restore_buffers
+            ):
+                yield
+        else:
+            if not getattr(self, "_warned_about_cp", False) and self.is_main_process:
+                logger.warning("Context parallel training is not enabled. This context manager will have no effect.")
+            # As this context manager is recreated each training step, we only warn once
+            self._warned_about_cp = True
+            yield
```
SunMarc: maybe a good time to add `warning_once` that we have in transformers ?

src/accelerate/big_modeling.py
```diff
@@ -747,3 +747,43 @@ def _attach_layerwise_casting_hooks(
             non_blocking,
             _prefix=layer_name,
         )
+
+
+def attach_context_parallel_hooks(
+    model: nn.Module,
+):
```
SunMarc: let's make this private

src/accelerate/commands/launch.py
```diff
@@ -994,6 +1039,16 @@ def _validate_launch_command(args):
...
+    # TODO: Merge into 1 if
+    if not args.use_fsdp and args.use_parallelism_config:
+        raise ValueError(
+            "You cannot use `--use_parallelism_config` without `--use_fsdp`. Please set `--use_fsdp` to True if you want to use parallelism config."
+        )
+    elif args.fsdp_version == 1 and args.use_parallelism_config:
+        raise ValueError(
+            "You cannot use `--use_parallelism_config` with FSDP version 1. Please set `--fsdp_version=2` if you want to use parallelism config."
+        )
+
```
SunMarc: todo

src/accelerate/parallelism_config.py
```diff
@@ -206,6 +210,23 @@ def _get_mesh(self) -> tuple[tuple[int, ...], tuple[str, ...]]:
...
     def __post_init__(self):
         # Basic size validation
+        if self.dp_replicate_size == 1:
+            self.dp_replicate_size = int(os.environ.get("PARALLELISM_CONFIG_DP_REPLICATE_SIZE", "1"))
+        if self.dp_shard_size == 1:
+            self.dp_shard_size = int(os.environ.get("PARALLELISM_CONFIG_DP_SHARD_SIZE", "1"))
+        if self.tp_size == 1:
+            self.tp_size = int(os.environ.get("PARALLELISM_CONFIG_TP_SIZE", "1"))
+        if self.cp_size == 1:
+            self.cp_size = int(os.environ.get("PARALLELISM_CONFIG_CP_SIZE", "1"))
```
SunMarc: For this kind of situation, it is better to set tp_size/cp_size ... to None. Then in the post init, if it is still None, we use the global var value or default to its default value. If the user explicitely created the parallelconfig with some values, we keep those.

src/accelerate/parallelism_config.py
```diff
@@ -206,6 +210,23 @@ def _get_mesh(self) -> tuple[tuple[int, ...], tuple[str, ...]]:
...
+
+        if self.tp_size > 1:
+            if self.tp_handler is None:
+                self.tp_handler = TorchTensorParallelConfig()
+
+        if self.cp_size > 1:
+            if self.cp_handler is None:
+                self.cp_handler = TorchContextParallelConfig()
```
SunMarc, line 228: yay

src/accelerate/state.py
```diff
@@ -981,7 +981,9 @@ def __init__(
                 DistributedType.MULTI_XPU,
                 DistributedType.MULTI_HPU,
             ]:
-                if os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" or fsdp_plugin is not None:
+                if (os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" or fsdp_plugin is not None) or (
+                    self.parallelism_config is not None and self.parallelism_config.cp_enabled
+                ):
```
SunMarc: not sure this is needed no ? in which case fsdp is not used but still we set `self.distributed_type = DistributedType.FSDP`

tests/test_dataclasses.py
```diff
@@ -73,6 +103,14 @@ def test_get_mesh(
         expected_shape,
         expected_dim_names,
     ):
+        # Skip tests based on version requirements
+        if _should_skip_cp_test(cp_size):
+            pytest.skip(f"CP tests require torch >= {BETA_CP_AVAILABLE_PYTORCH_VERSION}")
+        if _should_skip_tp_test(tp_size):
+            pytest.skip(
+                f"TP tests require torch >= {BETA_TP_AVAILABLE_PYTORCH_VERSION}, transformers available and >= {BETA_TP_AVAILABLE_TRANSFORMERS_VERSION}"
+            )
```
SunMarc: the error msg might be misleading when tp_size or cp_size == 1. same for the other tests. btw it should be fine for one of the nd size to be equal to 1 if the other is not no ? the device mesh will still be created.

SunMarc (commented):
Nice job ! Thanks for `integrating` CP using `parallelism_config` ! Left a couple of comments

docs/source/concept_guides/context_parallelism.md
S1ro1: IMO this should be `opt-in` instead of `opt-out` as by default we want to leave the buffers unchanged.

examples/fsdp2/nd_parallel.py
S1ro1: mb, was supposed to be removed, this is needed for the 128k seq-len example I was testing so I forgot it in

src/accelerate/accelerator.py
S1ro1: In AcceleratorState we already need to set stuff based on cp being enabled, we would get a default fsdp plugin out -> no way to detect if it was not passed in.

EDIT: moved this into acceleratorState (still more viable imo)

tests/test_dataclasses.py
S1ro1: This will error out (we have a check for that in post_init) if tp_size > 1 or cp_size > 1 and we're not on a required version. If either of this is not met, it's fine

src/accelerate/state.py
```diff
@@ -981,7 +981,13 @@ def __init__(
                 DistributedType.MULTI_XPU,
                 DistributedType.MULTI_HPU,
             ]:
-                if os.environ.get("ACCELERATE_USE_FSDP", "false") == "true" or fsdp_plugin is not None:
+                if self.parallelism_config and self.parallelism_config.cp_enabled and fsdp_plugin is None:
+                    raise ValueError(
+                        "`cp_size > 1` in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use `cp_enabled=True`, as we also shard the model across the device mesh to save more memory"
+                    )
```
SunMarc: cc might not be optimal for downstream lib like axolotl cc @winglian

src/accelerate/accelerator.py
```diff
@@ -1497,6 +1502,9 @@ def prepare(self, *args, device_placement=None):
         if self.parallelism_config and self.parallelism_config.tp_enabled:
             args = self._prepare_tp(*args)
 
+        if self.parallelism_config and self.parallelism_config.cp_enabled:
+            args = self._prepare_cp(*args)
```
winglian, line 1560: I am worried that in Axolotl, that automatically handling this might break existing context-parallel handling we have. @djsaunde

src/accelerate/parallelism_config.py
```diff
@@ -206,6 +210,23 @@ def _get_mesh(self) -> tuple[tuple[int, ...], tuple[str, ...]]:
...
     def __post_init__(self):
         # Basic size validation
+        if self.dp_replicate_size is None:
+            self.dp_replicate_size = int(os.environ.get("PARALLELISM_CONFIG_DP_REPLICATE_SIZE", "1"))
+        if self.dp_shard_size is None:
+            self.dp_shard_size = int(os.environ.get("PARALLELISM_CONFIG_DP_SHARD_SIZE", "1"))
+        if self.tp_size is None:
+            self.tp_size = int(os.environ.get("PARALLELISM_CONFIG_TP_SIZE", "1"))
+        if self.cp_size is None:
+            self.cp_size = int(os.environ.get("PARALLELISM_CONFIG_CP_SIZE", "1"))
```
winglian, line 220: aren't most accelerate env vars prefixed with `ACCELERATE_`?

SunMarc: oh yes good catch, can you fix that @S1ro1 ?

S1ro1: Not entirely true actually, paradigm variables are prefixed with ACCELERATE (such as use_fsdp, use_deepspeed), then variables configuring the underlying implementations are prefixed with the impl name (i.e. FSDP_, MEGATRON_LM_) etc. Such as [here](https://github.com/huggingface/accelerate/blob/cb343c63d7736baf082739393e52c9c1edcdf676/src/accelerate/utils/dataclasses.py#L1764C9-L1768C1) or [here](https://github.com/huggingface/accelerate/blob/cb343c63d7736baf082739393e52c9c1edcdf676/src/accelerate/utils/dataclasses.py#L2433C5-L2436C74)

Afaik only DeepSpeed is special and prefixes with ACCELERATE_DEEPSPEED_ making the env variables insanely long

src/accelerate/big_modeling.py
```diff
@@ -747,3 +747,43 @@ def _attach_layerwise_casting_hooks(
...
+def _attach_context_parallel_hooks(
+    model: nn.Module,
+):
+    """
+    Monkeypatch huggingface's `transformers` model to fix attention mask issues when using context parallelism.
+
+    This function attaches forward_pre_hooks to each self_attn module of the model, where each hook checks the
+    args/kwargs, if they contain an attention mask, if it does, it will remove this mask, check if it is a causal mask,
+    if yes, will add a kwarg `is_causal=True`, otherwise will raise an error. This is because context parallelism does
+    not support attention masks. This function modifies the model in place.
```
djsaunde, line 761: Nice, I was wondering how to handle this myself.

djsaunde: Oh, I see you don't actually check whether the mask is causal yet.

S1ro1: yes we should do that eventually, will revisit, currently we only warn in docs.

src/accelerate/accelerator.py
djsaunde: Yeah, I think it will. `ring-flash-attn` (what we use) supports non-causal masks so e.g. the mask deletion / replacement with `causal=True` is not good for us. We could patch over this maybe? I like the way pre-hooks are handled here in the accelerator, so we could swap to that vs. our context manager.

S1ro1: You can probably just delete the hook? It's gonna be the first one always, as I use `prepend=True` so should be pretty simple! The hook is the only thing we add for cp to work, then you only would need to use the context manager (which you won't)

src/accelerate/state.py
winglian: yeah, we definitely support CP without FSDP, so this would break that feature. Maybe some other sort of explicit setting that a user is enabling accelerate to handle CP for them? @djsaunde

src/accelerate/utils/launch.py
```diff
@@ -347,6 +347,20 @@ def prepare_multi_gpu_env(args: argparse.Namespace) -> dict[str, str]:
     current_env["OMP_NUM_THREADS"] = str(args.num_cpu_threads_per_process)
     if args.enable_cpu_affinity:
         current_env["ACCELERATE_CPU_AFFINITY"] = "1"
+
+    if not args.use_parallelism_config:
+        return current_env
+
+    prefix = "PARALLELISM_CONFIG_"
```
winglian, line 356:
```suggestion
    prefix = "ACCELERATE_PARALLELISM_CONFIG_"
```
Similar to comment above about accelerate using namespaced env vars

src/accelerate/state.py
djsaunde: I'm curious why this is gated in the first place, can we not use CP in accelerate sans FSDP? They should be independent.

S1ro1: > I'm curious why this is gated in the first place, can we not use CP in accelerate sans FSDP? They should be independent.

They are (of sorts) but FSDP is a free lunch with CP, so imo it should be the default. While we're computing the ring-attention we can prefetch next fsdp_layer for free, giving us 1/cp_size*fsdp_size savings in model/optimizer/grads. I have some profiling for this in the concept guide.

TLDR: it can be independent, but there's (almost) no world where it's worth to not do FSDP on top.

PR #3701, opened 2025-08-01
[WIP] Async checkpointing

S1ro1:
Very much WIP, overrides bunch of stuff I'm not sure that is stable to do. 
TODO: discuss if we want to do a bit different approach (and more easily maintainable)

### DIFF
--- a/src/accelerate/dist_checkpointing.py
+++ b/src/accelerate/dist_checkpointing.py
@@ -1,0 +1,182 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.import queue
+
+import dataclasses
+import os
+import pickle
+import queue
+from io import UnsupportedOperation
+from pathlib import Path
+from typing import TYPE_CHECKING, cast
+
+import torch
+import torch.distributed.checkpoint as dcp
+import torch.distributed.checkpoint.state_dict as dcs
+from torch.distributed.checkpoint.filesystem import (
+    FileSystemWriter,
+    SavePlan,
+    SavePlanner,
+    _generate_uuid,
+    _split_by_size_and_type,
+)
+from torch.distributed.checkpoint.metadata import Metadata, MetadataIndex, StorageMeta
+from torch.distributed.checkpoint.storage import WriteResult
+
+
+if TYPE_CHECKING:
+    from accelerate import Accelerator
+
+
+class AccelerateStorageWriter(FileSystemWriter):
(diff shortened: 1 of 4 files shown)


src/accelerate/dist_checkpointing.py
```diff
@@ -0,0 +1,182 @@
...
+)
+from torch.distributed.checkpoint.metadata import Metadata, MetadataIndex, StorageMeta
+from torch.distributed.checkpoint.storage import WriteResult
+
+
+if TYPE_CHECKING:
+    from accelerate import Accelerator
+
+
+class AccelerateStorageWriter(FileSystemWriter):
```
S1ro1, line 41: This class is the issue: I'm overriding quite interesting stuff from Pytorch that idk if I should (asked on their slack if it's safe). If we don't have this, we can't save optimizer into 1 directory and model into another, which we currently do

src/accelerate/dist_checkpointing.py
```diff
@@ -0,0 +1,182 @@
...
+                    )
+                )
+
+            return tuple(result)
+
+        model_metadata, optim_metadata = _split_metadata(metadata)
+        model_storage_md, optim_storage_md = {}, {}
+        for wr_list in results:
+            for wr in wr_list:
+                new_index = dataclasses.asdict(wr.index)
```
S1ro1, line 114: `WriteResult` dataclass is frozen (which tells a lot about what kind of war crimes I do here), so we have to use some fancy python things to avoid that.

src/accelerate/dist_checkpointing.py
```diff
@@ -0,0 +1,182 @@
...
+        except TypeError:
+            pass
+
+        def _split_metadata(
+            metadata: Metadata,
+        ) -> tuple[Metadata, Metadata]:
+            result = []
+            for to_get in ["model", "optim"]:
+                result.append(
+                    Metadata(
```
S1ro1, line 96: By default DCP thinks we're saving an object called "state" into 1 directory, which we're not. We're saving "optimizer" into 1 subdirectory and "model" into another. That's why we have to update the metadata (remove the "state" prefix and split it into 2)

src/accelerate/dist_checkpointing.py
```diff
@@ -0,0 +1,182 @@
...
+                        os.sync()
+
+            metadata_path = self.fs.concat_path(final_path, ".metadata")
+            if self.fs.exists(metadata_path):
+                self.fs.rm_file(metadata_path)
+
+            self.fs.rename(tmp_path, metadata_path)
+
+
+def save_model_and_optimizer(
```
S1ro1, line 155: This is the only "public" api that we expose, not even. We only use this internally in `accelerator.save_state`.

S1ro1:
bump

romitjain:
Hi @S1ro1, I am eagerly waiting for this to be merged. Any idea of how much time it might take?
Is there any help that I can provide (tests/dev)?

Also, a side question, looking at the PR, it seems like it doesn't support `safetensor` serialization. Are there plans to support it, or are you open to contributions for that?

Thanks

romitjain:
@S1ro1 Just a gentle reminder. Let me know if you'd like me to take a stab at any pending changes, and you can review?
Thanks

romitjain:
cc: @SunMarc

SunMarc:
@S1ro1 is not working at HF anymore, so feel free to take over his PR if you want @romitjain !

romitjain:
Sure @SunMarc.
Let me get back on this next week

bhack:
What is the status of this?

Issue #3702, opened 2025-08-01
[ Enhancement] Getter methods for parallel ranks (TP/DP/CP) to Accelerator class


WoosungMyung:
Hello,

Thanks for the amazing works for enabling parallelism easily in huggingface for users!
I would like to suggest getter methods to the accelerator class that return the parallelism rank (DP,TP,CP).

### Suggest
Added the following methods to the `Accelerator` class:
- `get_tensor_parallel_rank(self) -> int`
- `get_data_parallel_rank(self) -> int`
- `get_context_parallel_rank(self) -> int`

Each method will be implemented using parallelism_config, and torch_device_mesh. 
If the parallelism is not enabled, the method returns 0.

### Motivation
Currently, `Accelerate` provides access to the parallelism configuration and the device mesh, but does not have simple method to directly retrieve the rank of each parallel dimension. ( Although it is possible, not simple)

By adding these lightweight getter methods, we improve usability for developers and researchers working with hybrid parallelism (e.g., FSDP + TP or HSDP + TP/CP).

I'd be happy to submit a PR if this is aligned with your thoughts.

S1ro1:
@WoosungMyung First of all thank you for the nice words regarding our parallelism work 🤗 Implementing such methods would surely be helpful and wanted. If you want to create a PR, I'd be happy to review! Thank you. 

Small note, I think the methods should return -1/None if not enabled as 0 is a valid rank which could make validation troublesome.

PR #3703, opened 2025-08-02
Add Parallelism getter property to Accelerator class

WoosungMyung:
This PR adds a @property method  for simply get parallelism rank in HF Accelerate.

Motivation:
- To make it easier to retrieve the process rank in Distributed Learning using Accelerate.

Changes:
- Added `rank` property in accelerator.py

This PR is from issue #3702 with S1ro1.
Let me know if there’s anything I should improve. 

Thanks a lot for giving opportunity for this amazing PJ!

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -784,6 +784,34 @@ class Accelerator:
 
         return parallelism_config
 
+    @property
+    def tensor_parallel_rank(self) -> Optional[int]:
+        if self.parallelism_config and self.parallelism_config.tp_enabled:
+            return self.torch_device_mesh.get_local_rank("tp")
+        return None
+
+    @property
+    def pipeline_parallel_rank(self) -> Optional[int]:
+        raise NotImplementedError("Pipeline parallelism is currently not supported in Accelerate.")
+
+    @property
+    def context_parallel_rank(self) -> Optional[int]:
+        raise NotImplementedError("Context parallelism is currently not supported in Accelerate.")
+      
+    # Replicate based Data Parallelism (e.g. DDP)
+    @property
+    def data_parallel_rank(self) -> Optional[int]:
+        if self.parallelism_config and self.parallelism_config.dp_replicate_enabled:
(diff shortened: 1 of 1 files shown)


src/accelerate/accelerator.py
```diff
@@ -784,6 +784,52 @@ def _setup_parallelism_config(
...
+            raise RuntimeError("Replicate-based data parallelism is not enabled. Please check your configuration.")
+        return self.torch_device_mesh.get_local_rank("dp_replicate")
+
+    @property
+    def data_parallel_shard_rank(self) -> int:
+        """
+        Returns the local rank for shard-based data parallelism (e.g., FSDP).
+        Raises an error if not enabled.
+        """
+        if not self.parallelism_config or not self.parallelism_config.dp_shard_enabled:
```
S1ro1: I think we want to adapt this a little, sorry for wrong assumption yesterday. I'd say the best thing UX wise is: return the rank if enabled, return 0 if parallelism config is enabled but no parallelism {x} is enabled (in this case all ranks are esentially 0) and raise a RuntimeError if neither of these 2 met. Do you think that is reasonable?

WoosungMyung: @S1ro1 
Thanks for your fast feedback! I think it makes sense that parallelism configuration is set & parallelism is not enable,  it should return rank 0 for all process. I'd appreciate it if you could take a look and let me know your thoughts.

S1ro1 (approved):
Thank you, LGTM now. Congratulation on your first contribution! Just fixed style checks for you and will merge!

WoosungMyung:
@S1ro1 
Thank you so much for open-minded feedback throughout my first PR. I really appreciate the constructive discussions and your welcoming attitude.

S1ro1:
> @S1ro1
Thank you so much for open-minded feedback throughout my first PR. I really appreciate the constructive discussions and your welcoming attitude.

We thank you for the contribution and look forward to more 🤗 Pleasure to be of help!

PR #3707, opened 2025-08-04
Fix: prepare works even if nothing except tp specified (rare)

S1ro1:
If we only had TP specified, the training would probably be still correct, though the optimizer/dataloader wouldn't be accelerator prepared (in some cases). This was and is a rare case, as pure TP is not fully supported yet anyway. Though this PR fixes atleast this.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1584,9 +1584,17 @@ class Accelerator:
         return result if len(result) > 1 else result[0]
 
     def _prepare_tp(self, *args):
+        # First pass: prepare everything except schedulers (and model, which is prepared separately below)
+        result = [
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
LGMT. Just to be sure, this should be safe wrt to fsdpv2 for example ? Do we handle the states differently for these cases ? If tp is used, this will always to executed first

SunMarc:
TP training alone should work no ?

S1ro1:
> TP training alone should work no ?

Yes it should, though there are some limitations which I didn't have time to address yet (saving etc), though usage of tp only training is very limited so not a big deal IMO.

> LGMT. Just to be sure, this should be safe wrt to fsdpv2 for example ? Do we handle the states 
> differently for these cases ? If tp is used, this will always to executed first

No, this would only cause `_prepare_one` to be called multiple times which becomes a no-op the 2nd time anyway.

PR #3713, opened 2025-08-06
Parallel config env

winglian:
# What does this PR do?

This is a fix for how Trainer by default resets the PartialState shared state, which means we can't read the ParallelismConfig from PartialState anymore. Since we can infer the ParallelismConfig from the env vars like we do all the other `ACCELERATE_*` options (e.g. FSDP, Deepspeed,etc), we can repopulate this inside the constructor. 

Addtionally, there is a fix to the indent level of setting the fsdp_plugin, as if it the conditional it currently is in would bypass setting self.fsdp_plugin if the env var was set to allow standalone CP, which is independent of FSDP.

We'll need to add some fixes down the line in Trainer, but this will get us to a working release.

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -786,6 +786,8 @@ class Accelerator:
                         "Partial state contains a `parallelism_config` which is not None, but you configured `parallelism_config` from the `accelerate launch` CLI. We don't know which to use, please remove one of those configuration methods."
                     )
                 parallelism_config = PartialState().parallelism_config
+            elif ParallelismConfig.has_env_variables():
+                parallelism_config = ParallelismConfig()
             else:
                 # TODO: Remove after deprecating tp_plugin
                 tp_size = 1 if torch_tp_plugin is None else torch_tp_plugin.tp_size
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -903,6 +903,7 @@ class AcceleratorState:
         fsdp_plugin=None,
         torch_tp_plugin=None,
         megatron_lm_plugin=None,
+        parallelism_config=None,
         _from_accelerator: bool = False,
         **kwargs,
     ):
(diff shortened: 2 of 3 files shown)


src/accelerate/accelerator.py
```diff
@@ -786,6 +786,8 @@ def _setup_parallelism_config(
                         "Partial state contains a `parallelism_config` which is not None, but you configured `parallelism_config` from the `accelerate launch` CLI. We don't know which to use, please remove one of those configuration methods."
                     )
                 parallelism_config = PartialState().parallelism_config
+            elif ParallelismConfig.has_env_variables():
+                parallelism_config = ParallelismConfig()
```
S1ro1: I actually don't understand really, because if `has_env_variables` evaluates to True, we create an empty parallelism config (that reads from env), but we would do the same beforehand as we'd go to the `else` branch which would always evaluate to True and construct a parallelism_config from env vars.

src/accelerate/state.py
```diff
@@ -995,13 +997,13 @@ def __init__(
...
-                    ) or (self.parallelism_config is not None and self.parallelism_config.cp_enabled):
-                        self.distributed_type = DistributedType.FSDP
-                        if self._mixed_precision != "no":
-                            fsdp_plugin.set_mixed_precision(self._mixed_precision)
-                        self.fsdp_plugin = fsdp_plugin
+                if (os.environ.get("ACCELERATE_USE_FSDP", "false").lower() == "true" or fsdp_plugin is not None) or (
+                    self.parallelism_config is not None and self.parallelism_config.cp_enabled
+                ):
+                    self.distributed_type = DistributedType.FSDP
+                    if self._mixed_precision != "no":
```
S1ro1: We should then add `and fsdp_plugin is not None` as sometimes fsdp_plugin could be None here, raising from `set_mixed_precision`

PR #3716, opened 2025-08-06
Fix: tp size wouldn't read from env

S1ro1:
With change in #3700 where we default to None, passing 1 as tp_size would result in env vars not being read.


SunMarc (approved):
thanks ! nice catch

PR #3720, opened 2025-08-06
Remove `ParallelismConfig` from `PartialState`

SunMarc:
# What does this PR do?

Since we moved `ParallelismConfig` to AcceleratorState, it should be fine to remove that in `PartialState`. Trainer in transformers actually reset `PartialState` so we shouldn't keep it there as we can't reconstruct if we don't create global var.
Apart from that, we don't initialize `parallelismconfig` if it is not passed or we didn't set any env var.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -451,10 +451,11 @@ class Accelerator:
                 if "recipe_handler" in handler_attr and not self.has_fp8_handler:
                     self.has_fp8_handler = True
 
-        parallelism_config = self._setup_parallelism_config(parallelism_config, torch_tp_plugin)
+        # TODO: Remove after deprecating tp_plugin
+        if parallelism_config is None:
+            parallelism_config = ParallelismConfig(tp_size = None if torch_tp_plugin is None else torch_tp_plugin.tp_size)
 
         kwargs = self.init_handler.to_kwargs() if self.init_handler is not None else {}
-        kwargs["parallelism_config"] = parallelism_config
         self.state = AcceleratorState(
             mixed_precision=mixed_precision,
             cpu=cpu,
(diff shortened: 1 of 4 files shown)


src/accelerate/parallelism_config.py
```diff
@@ -194,6 +196,14 @@ def build_device_mesh(self, device_type: str):
 
         return device_mesh
 
+    def get_device_mesh(self, device_type: Optional[str] = None):
+        if self.device_mesh is None:
+            if device_type is not None:
+                self.device_mesh = self.build_device_mesh(device_type)
+            else:
+                raise ("You need to pass a device_type e.g cuda to build the device mesh")
+        return self.device_mesh
```
S1ro1, line 211: Maybe not needed/required, but maybe check if we passed in device_type (from accelerator) and device_mesh already existed, that it's device_type matches the one we passed in?

S1ro1 (approved):
Except of the one nit inline, lgtm!

src/accelerate/parallelism_config.py
SunMarc: done !

src/accelerate/accelerator.py
```diff
@@ -451,23 +451,28 @@ def __init__(
                 if "recipe_handler" in handler_attr and not self.has_fp8_handler:
                     self.has_fp8_handler = True
 
-        parallelism_config = self._setup_parallelism_config(parallelism_config, torch_tp_plugin)
+        if parallelism_config is None:
+            # TODO: Remove after deprecating tp_plugin
+            if torch_tp_plugin is not None:
+                parallelism_config = ParallelismConfig(tp_size=torch_tp_plugin.tp_size)
+            elif os.environ.get("ACCELERATE_USE_PARALLELISM_CONFIG", "false").lower() == "true":
```
S1ro1, line 458: This actually needs to be set somewhere, I think it's not set rn.

SunMarc: i added it back no ?

SunMarc:
```
 if args.use_parallelism_config:
        current_env["ACCELERATE_USE_PARALLELISM_CONFIG"] = "true"
```

S1ro1: Oh right, missed it, thanks!

PR #3722, opened 2025-08-07
Fix tests

SunMarc:
# What does this PR do?

This PR fixing the CI tests. 
- we might deprecate pippy or at least not test it anymore 
- skip bnb tests as i can't reproduce it
- try to run deepspeed tests first -> worked and this fixed the fsdp issues we saw 
- fix deepspeed tests
- fix skorch integration tests 
- fix release docker
- skip compile test instead of pinning torch 2.8. Not sure if this was the issue but there was a segmentation fault. 

Note that with torch 2.8 dynamo test is not passing anymore. Will open a PR about this [test](https://github.com/pytorch/pytorch/issues/160084)

### DIFF
--- a/.github/workflows/self_hosted_integration_tests.yml
+++ b/.github/workflows/self_hosted_integration_tests.yml
@@ -112,7 +112,7 @@ jobs:
           cd skorch;
           git config --global --add safe.directory '*'
           git checkout master && git pull
-          pip install .[testing]
+          pip install .[test]
           pip install flaky
 
       - name: Show installed libraries
--- a/.github/workflows/build-docker-images-release.yml
+++ b/.github/workflows/build-docker-images-release.yml
@@ -15,7 +15,7 @@ jobs:
     outputs:
       version: ${{ steps.step1.outputs.version }}
     steps:
-      - uses: actions/checkout@4
+      - uses: actions/checkout@v4
       - id: step1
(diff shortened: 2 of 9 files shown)


.github/workflows/self_hosted_integration_tests.yml
```diff
@@ -112,7 +112,7 @@ jobs:
           cd skorch;
           git config --global --add safe.directory '*'
           git checkout master && git pull
-          pip install .[testing]
+          pip install .[test]
```
S1ro1, line 115: This is wanted, right? Or just a typo

BenjaminBossan:
Thanks for taking care of fixing the tests.

> * fix skorch integration tests

What is the error? I don't see a fix in the diff.

SunMarc:
> What is the error? I don't see a fix in the diff.

We were not installing skorch correctly. Previously we had [testing] but we should have installed [test] dependency and we were hitting a pytest issue was pytest-cov wasn't installed ! btw, there is a failing test in the integration test https://github.com/huggingface/accelerate/actions/runs/16804693816/job/47594034737. Can you fix that on your side ? @BenjaminBossan

BenjaminBossan:
> We were not installing skorch correctly. Previously we had [testing] but we should have installed [test] dependency and we were hitting a pytest issue was pytest-cov wasn't installed !

Ah, subtle.

> Can you fix that on your side ?

Should be fixed now on skorch master branch.

PR #3737, opened 2025-08-18
[ND Parallel] Update examples, cleanup

S1ro1:
- Renames `fsdp2` examples to `torch_native_parallelism` (subject to change)
- Adds example using transformers trainer
- Cleans up the old FP8 example.

### DIFF
--- a/examples/torch_native_parallelism/nd_parallel_trainer.py
+++ b/examples/torch_native_parallelism/nd_parallel_trainer.py
@@ -33,6 +33,8 @@ def parse_args():
 
 
 def main():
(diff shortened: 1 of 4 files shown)


examples/torch_native_parallelism/nd_parallel_trainer.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2025 The HuggingFace Inc. team. All rights reserved.
```
SunMarc, line 1: maybe add a check for transformers version here to warn the users that it requires the upcoming version / main branch to run this

SunMarc (approved):
Nice thanks for adding this ! Just a nit

PR #3750, opened 2025-08-28
feat(utils): add collect_epoch_for_metrics function

### LINKED ISSUE #3749: collect_epoch_for_metrics — stream step outputs, auto-dedup, optional trim/shard
## Summary / problem
With even_batches=True (default), Accelerate keeps ranks aligned by duplicating a few samples at the tail. The docs recommend accelerator.gather_for_metrics(...) to gather per step and drop those duplicates. In practice, many users want an epoch-level result equal to len(dataset) without writing the same gather-append-concat boilerplate (and occasional final slice).

## Proposal
Add a helper that streams per-step outputs, uses gather_for_metrics internally, and returns an epoch-level tensor (exactly expected_len if provided). Optional: a callback to stream shards to disk instead of keeping everything in RAM.

##utils
collect_epoch_for_metrics(
    accelerator: Accelerator,
    step_iter: Iterable,
    step_fn: Callable[[Any], torch.Tensor],   # returns [b, ...] tensor per step
    *,
    cat_dim: int = 0,
    expected_len: Optional[int] = None,
    stream_to: Optional[Callable[[torch.Tensor, int], None]] = None,
) -> Optional[torch.Tensor]

##convenience method
Accelerator.collect_epoch_for_metrics(...)

## Semantics

Calls gather_for_metrics() each step → dedups tail automatically.
On main process: accumulates CPU chunks, torch.cat(..., dim=cat_dim) at the end.
If expected_len is set (or user passes len(dataset)), slice once as a final guard.
If stream_to is provided, call it with each gathered chunk and return a small summary (e.g., {"steps": k, "count": n}) instead of a giant tensor.

## Why useful

Collapses a very common eval/inference idiom into 1 call.
Reduces foot-guns (forgetting dedup/slice; holding everything on GPU).
Keeps low-level primitives (gather, gather_for_metrics) unchanged.

## Backwards compatibility
Purely additive. No behavior changes to existing APIs.
Test plan (CPU & single-process)
Concats multiple step outputs and returns sum(b_i) rows.
Honors expected_len by slicing.
stream_to gets called with correct step indices and shapes.

MengAiDev:
- Add collect_epoch_for_metrics function to accelerate/utils/operations.py
- Add corresponding method to Accelerator class
- Function helps to stream per-step outputs and returns an epoch-level tensor
- Includes options for concatenation dimension, expected length, and streaming to a callback

Fixes #3749 

@SunMarc @SunMarc @zach-huggingface

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -27,7 +27,7 @@ from collections import OrderedDict
 from contextlib import contextmanager
 from functools import partial
 from types import MethodType
-from typing import Any, Callable, Union
+from typing import Any, Callable, Iterable, Optional, Union
 
 import torch
 import torch.utils.hooks as hooks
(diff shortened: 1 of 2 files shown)


S1ro1:
Thank you for the contribution, however, for me his feels a bit high-level for what we aim accelerate to provide, not entirely positive about this. cc @SunMarc for 2nd opinion

SunMarc (commented):
Same concern as @S1ro1, it could be pontentially useful if we integrate it in downstream lib like transformers for ex

MengAiDev:
Thanks for the review! 
The main reason I think it fits in Accelerate is that it's really a **generic distributed data collection tool**, not just for metrics. It handles the tricky parts of gathering and concatenating outputs across devices in a way that's useful for any task needing full-epoch data (like validation, feature collection, or custom stats).

Right now, every downstream library ends up reimplementing this same pattern. Putting it here would provide a standardized, efficient way to handle this common distributed operation and prevent a lot of duplicated effort.

What are your thoughts?

SunMarc:
At this stage, we prefer not to add this in accelerate. Thanks nevertheless for contributing

PR #3760, opened 2025-09-05
fix: `model.set_requires_gradient_sync(False)` should be called to turn off gradient synchronization in FSDP2.

EquationWalker:
In FSDP2, the model(`FSDPModule`) does not have `no_sync()` and instead calls  `model.set_requires_gradient_sync(False)`  to turn off gradient synchronization. See at [torch.distributed.fsdp.FSDPModule.set_requires_gradient_sync](https://docs.pytorch.org/docs/stable/distributed.fsdp.fully_shard.html#torch.distributed.fsdp.FSDPModule.set_requires_gradient_sync)

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1175,7 +1175,7 @@ class Accelerator:
                 yield
             finally:
                 model.set_requires_gradient_sync(True)
-        else:  
+        else:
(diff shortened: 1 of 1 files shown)


S1ro1:
Can you fix quality @EquationWalker ?

PR #3767, opened 2025-09-09
refactor: Use  `with`  in Accelerator.autocast()instead of ` __enter__()` and` __exit__()` for more elegant style.

EquationWalker:
# What does this PR do?
refactor: Use  `with`  in Accelerator.autocast()instead of ` __enter__()` and` __exit__()` for more elegant style.


SunMarc (approved):
Indeed thanks for the clean-up !

SunMarc:
@bot /style

PR #3783, opened 2025-09-18
remove mlflow from testing 

SunMarc:
# What does this PR do?

This PR removes mlflow from the setup so that we don't test it anymore. The tests are failing because we are installing a very old version due to conflicting dependencies.


S1ro1 (approved):
LGTM

Issue #3789, opened 2025-09-20
script: accelerate.py function: clip_grad_norm_  if parameters == [p for p in model.parameters()]: RuntimeError: Boolean value of Tensor with more than one value is ambiguous


Xingyuxingyu:
### System Info

```Shell
accelerate 1.10.1
os: Linux #25-Ubuntu SMP Wed Mar 30 15:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
python: 3.10
numpy: 1.26.0
torch version: 2.5.1+cu118
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

accelerate fsdp model
if accelerator.sync_gradients and train_config.max_grad_norm > 0:
    all_norm = accelerator.clip_grad_norm_(model.parameters(), train_config.max_grad_norm)

### Expected behavior

No RuntimeError

S1ro1:
Hi, can you check the PR #3791 if that fixes it for you?

SunMarc:
Hey @Xingyuxingyu, please share a bettter reproducer !

Xingyuxingyu:
这是来自QQ邮箱的假期自动回复邮件。您好，已收到您的邮件。我会尽快给您回复。

Xingyuxingyu:
这是来自QQ邮箱的假期自动回复邮件。您好，已收到您的邮件。我会尽快给您回复。

PR #3791, opened 2025-09-23
Fix: truth value of tensor with more than 1 variable is unambiguous

S1ro1:
Fixes issue #3789


SunMarc (approved):
This should be safe to merge no ?

PR #3782, opened 2025-09-16
WIP: initial implementation

S1ro1:
cc @stas00 

```python
from accelerate import Accelerator
from accelerate.utils import ParallelismConfig
from deepspeed.runtime.utils import move_to_device
from accelerate.utils.dataclasses import DeepSpeedContextParallelConfig
from torch import tensor
from transformers import AutoModelForCausalLM
import torch

MODEL_ID = "hf-internal-testing/tiny-random-LlamaForCausalLM"
MICRO_BATCH_SIZE = 1

parallelism_config = ParallelismConfig(
    flavour="deepspeed",
    cp_size=2,
    dp_shard_size=4,
    cp_handler=DeepSpeedContextParallelConfig(
        max_length=64, attn_implementation="sdpa"
    ),
)

accelerator = Accelerator(parallelism_config=parallelism_config)

input_ids = tensor(
    [[1, 10, 10, 10, 2, 2], [1, 20, 20, 20, 2, 2]],
)
position_ids = tensor([[0, 1, 2, 3, 4, 5], [0, 1, 2, 3, 4, 5]])
ds = torch.utils.data.TensorDataset(input_ids, position_ids)


def collate_fn(batch):
    input_ids, position_ids = batch[0]
    return dict(
        input_ids=input_ids.unsqueeze(0),
        position_ids=position_ids.unsqueeze(0),
        labels=input_ids.unsqueeze(0),
        shift_labels=input_ids.unsqueeze(0),
    )


model = AutoModelForCausalLM.from_pretrained(MODEL_ID)
optimizer = torch.optim.Adam(model.parameters(), lr=1e-3)
dl = torch.utils.data.DataLoader(ds, batch_size=MICRO_BATCH_SIZE, collate_fn=collate_fn)

model, optimizer, dl = accelerator.prepare(model, optimizer, dl)

sp_group = accelerator.torch_device_mesh["cp"].get_group()
sp_world_size = parallelism_config.cp_size

# Normal training loop
for iter, batch in enumerate(dl):
    batch = move_to_device(batch, model.device)
    outputs = model(**batch)

    shift_labels = batch["shift_labels"]
    loss = model.module.loss_function(
        logits=outputs.logits,
        labels=None,
        shift_labels=shift_labels,
        vocab_size=model.module.config.vocab_size,
    )

    # differentiable weighted per-shard-loss aggregation across ranks
    losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
[...]

PR #3817, opened 2025-10-22
Deepspeed Ulysses/ALST integration

stas00:
This is the completion of the work started by @S1ro1 at https://github.com/huggingface/accelerate/pull/3782 to integrate the ALST/Ulysses long sequence training into HF Accelerate. Paper https://arxiv.org/abs/2506.13996. This is Matej's original code with lots of additional work on top and docs+tests from me.

Here is the corresponding HF Trainer integration PR: https://github.com/huggingface/transformers/pull/41832

If you want to try it out please first install deepspeed from `deepspeed@master` as deepspeed needed some tweaks to make this integration work.

To use this feature a user needs

1. to create `ParallelismConfig`
```
parallelism_config = ParallelismConfig(
    sp_backend="deepspeed",
    sp_size=2,
    sp_handler=DeepSpeedSequenceParallelConfig(attn_implementation="flash_attention_2"),
)

accelerator = Accelerator(parallelism_config=parallelism_config)
```

2. add to their code the use of `shift_labels` and an aggregation of loss across ranks

```
    shift_labels = batch["shift_labels"]
    loss = model.module.loss_function(
        logits=outputs.logits,
        labels=None,
        shift_labels=shift_labels,
        vocab_size=model.module.config.vocab_size,
    )

    # differentiable weighted per-shard-loss aggregation across ranks
    losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
    # special dealing with SFT that has prompt tokens that aren't used in loss computation
    good_tokens = (shift_labels != -100).view(-1).sum()
    good_tokens_per_rank = torch.distributed.nn.functional.all_gather(
        good_tokens, group=sp_group
    )
    total_loss = sum(
        losses_per_rank[rank] * good_tokens_per_rank[rank]
        for rank in range(sp_world_size)
    )
    total_good_tokens = sum(good_tokens_per_rank)
    loss = total_loss / max(total_good_tokens, 1)
```

### Quality validation

I wrote 3 accelerate-based scripts (attached at the end of the OP):
- 1 gpu
- 4 gpus w/ fsdp
- 4 gpus w/ deepspeed ulysses

The loss checks out with very small variations due the precision loss in aggregation.

Screenshot 2025-10-22 at 09 22 50

### TODO

- [x] tests
- [x] docs
- [x] Deepspeed PR got merged https://github.com/deepspeedai/DeepSpeed/pull/7638
- [x] deepspeed dependency version needs to be bumped up to whatever next version will be

These are really needed for the HF Trainer PR https://github.com/huggingface/transformers/pull/41832 but since it anchors on accelerate let's make the dependency here instead.

- [x] need to wait for a merge of https://github.com/deepspeedai/DeepSpeed/pull/7649
- [x] need to wait for a new Deepspeed release after above is merged probably 1.18.2 and update here the requirements.

### Scripts used to perform the quality validation

The scripts and config files are:

- [cp-1gpu.py](https://github.com/user-attachments/files/23059047/cp-1gpu.py)
- [cp-1gpu.sh](https://github.com/user-attachments/files/23059048/cp-1gpu.sh)
- [cp-alst.accelerate-config.yml](https://github.com/user-attachments/files/23059049/cp-alst.accelerate-config.yml)
- [cp-alst.ds-config.json](https://github.com/user-attachments/files/23059050/cp-alst.ds-config.json)
- [cp-alst.py](https://github.com/user-attachments/files/23059377/cp-alst.py)
- [cp-alst.sh](https://github.com/user-attachments/files/23059052/cp-alst.sh)
- [cp-fsdp.py](https://github.com/user-attachments/files/23059053/cp-fsdp.py)
- [cp-fsdp.sh](https://github.com/user-attachments/files/23059054/cp-fsdp.sh)

You run the `.sh` files

cc: @SunMarc

### DIFF
--- a/tests/test_dataclasses.py
+++ b/tests/test_dataclasses.py
@@ -34,8 +34,12 @@ def _should_skip_cp_test(cp_size):
 
 
 def _should_skip_sp_test(sp_size):
-    """Check if CP test should be skipped based on sp_size and deepspeed version."""
-    return sp_size > 1 and not compare_versions("deepspeed", ">=", BETA_SP_AVAILABLE_DEEPSPEED_VERSION)
+    """Check if SP test should be skipped based on sp_size and deepspeed version."""
+    if sp_size <= 1:
+        return False
+    if not is_deepspeed_available():
+        return True
+    return not compare_versions("deepspeed", ">=", BETA_SP_AVAILABLE_DEEPSPEED_VERSION)
 
 
 def _should_skip_tp_test(tp_size):


src/accelerate/accelerator.py
```diff
@@ -1619,6 +1619,10 @@ def _prepare_cp(self, *args):
         from torch.distributed.tensor.experimental import context_parallel
         from torch.distributed.tensor.experimental._attention import set_rotate_method
 
+        if not self.is_fsdp2:
+            # deepspeed handles cp other way
+            return args
```
SunMarc: can we do a deepspeed check instead ? Also, is there nothing we can offload to this function instead of modifying `_prepare_deepspeed` ?

src/accelerate/accelerator.py
```diff
@@ -2162,6 +2170,24 @@ def _prepare_deepspeed(self, *args):
             ):
                 scheduler = obj
 
+        mpu = None
+        if cp_size > 1:
+            if is_dataloader_present and model is None:
+                raise ValueError(
+                    "You cannot pass a dataloader to `accelerate.prepare()` without passing a model when using Context Parallelism."
+                )
```
SunMarc: Note that in Trainer, we prepare the dataloader seperate from the model.

src/accelerate/accelerator.py
```diff
@@ -2323,6 +2349,20 @@ def _prepare_deepspeed(self, *args):
                     type(result[i]).__name__ in deepspeed.runtime.lr_schedules.VALID_LR_SCHEDULES
                 ):
                     result[i] = scheduler
+                elif isinstance(result[i], torch.utils.data.DataLoader):
+                    if cp_size > 1:
+                        cp_group = groups._get_sequence_parallel_group()
+                        cp_world_size = groups._get_sequence_parallel_world_size()
+                        cp_rank = groups._get_sequence_parallel_rank()
+                        print(cp_group)
```
SunMarc: remove print

src/accelerate/utils/dataclasses.py
```diff
@@ -2189,6 +2189,40 @@ def __post_init__(self):
...
+    def __post_init__(self):
+        if self.max_length is None:
+            if "PARALLELISM_CONFIG_CP_MAX_LENGTH" not in os.environ:
+                raise ValueError(
+                    "max_length must be provided either through the constructor or the environment variable PARALLELISM_CONFIG_CP_MAX_LENGTH"
+                )
+            self.max_length = int(os.environ["PARALLELISM_CONFIG_CP_MAX_LENGTH"])
+
+        if self.attn_implementation is None:
+            self.attn_implementation = os.environ.get("PARALLELISM_CONFIG_CP_ATTN_IMPLEMENTATION", None)
```
SunMarc: Minor nit, for the global env var, maybe have a DEEPSPEED prefix to them ?

src/accelerate/utils/dataclasses.py
```diff
@@ -2189,6 +2189,40 @@ def __post_init__(self):
...
+
+        if self.attn_implementation is not None and self.attn_implementation not in [
+            "flash_attention_2",
+            "flash_attention_3",
+            "sdpa",
+        ]:
+            raise ValueError(
+                f"Invalid attn_implementation: {self.attn_implementation}. Must be one of 'flash_attention_2', 'flash_attention_3' or 'sdpa'."
+            )
```
SunMarc: eager don't work ?

src/accelerate/accelerator.py
```diff
@@ -2162,6 +2170,29 @@ def _prepare_deepspeed(self, *args):
...
+                )
+
+            from deepspeed.runtime.sequence_parallel.ulysses_sp import UlyssesSPAttentionHF, UlyssesSPDataLoaderAdapter
+            from deepspeed.utils import groups
+
+            mpu = UlyssesSPAttentionHF.register_with_transformers(
+                model_name_or_path=model,
+                sequence_parallel_size=cp_size,
+                max_length=cp_handler.max_length,
+                core_attn_implementation=cp_handler.attn_implementation or model.config.attn_implementation,
```
SunMarc: minor nit but maybe check that model do have a `config` attribute ? Also, does this integration only works with transformers model ?

src/accelerate/parallelism_config.py
```diff
@@ -56,6 +61,7 @@ class ParallelismConfig:
 
     """
 
+    flavour: Literal["torch", "deepspeed"] = "torch"
```
SunMarc: just a personal nit but maybe we can rename it  to `mode` or `engine`

src/accelerate/utils/dataclasses.py
stas00: eager needs a 4D attention mask, which becomes huge and unsustainable for very long seqlen - so we need to rely on `position_ids` supported by FA perfectly and somewhat by SDPA (it ignores sub-samples).

src/accelerate/parallelism_config.py
stas00: I agree, this is Matej's original - I wanted to propose something similar like you did - which one do you prefer? also `backend` is perhaps an even more suitable one?

SunMarc: let's go for `backend` then !

src/accelerate/accelerator.py
stas00: should we do this in steps, integrating into Accelerate core first and then adapting to HF Trainer if needed or to try and integrate it into HF Trainer right away? Whatever you think is the most efficient strategy.

src/accelerate/utils/dataclasses.py
stas00: My guess is that perhaphs Matej thought it might be reusable to other CP backends?

src/accelerate/accelerator.py
sfc-gh-sbekman: > minor nit but maybe check that model do have a config attribute ? 

Is it always the case? I can surely add the check.

> Also, does this integration only works with transformers model ?

I had no idea Accelerate started to support non HF models. ALST/Ulysses has been designed to integrated into HF Transformers.

src/accelerate/accelerator.py
stas00: > can we do a deepspeed check instead? 

done

> Also, is there nothing we can offload to this function instead of modifying _prepare_deepspeed ?

I can't see of anything, it's closely tied to deepspeed's setup.

src/accelerate/accelerator.py
stas00: check added.

stas00:
@SunMarc, 
- I did some more tweaks to improve/simplify UX
- added docs - did I miss any places where I should mention this backend?
- added tests - I didn't actually see any torch CP e2e tests - do they even exist? in any case I wrote a simple e2e test for this PR - the quality checks are already extensively tested in the deepspeed repo

So we just have a few conversations above to complete and otherwise I'm just waiting for the deepspeed to make a new version so that we could anchor on it here. Otherwise it's ready for your complete review - but don't merge just yet until we get the new ds version here.

And then we can discuss the HF Trainer integration. Should we somehow mark this API as experimental to let users use it for a bit and possibly adjust things? If so please give me an example to follow.

src/accelerate/accelerator.py
SunMarc: thx !

src/accelerate/accelerator.py
SunMarc: I think it will be better to try integrating it into HF Trainer right away to see any potential blockers

SunMarc:
> added tests - I didn't actually see any torch CP e2e tests - do they even exist? in any case I wrote a simple e2e test for this PR - the quality checks are already extensively tested in the deepspeed repo

There are some e2e examples in the example/torch_native_parallelism folder but we are not running them in the CI. 

> And then we can discuss the HF Trainer integration. Should we somehow mark this API as experimental to let users use it for a bit and possibly adjust things? If so please give me an example to follow.

Let's try to integrate it into HF Trainer before merging this PR. Once it is tightly coupled to Trainer, even if the API is marked as experimental, we will most likely try to limit breaking changes. For experimental features, we just put it on the docs, like for big model inference (probably need to remove the warning for this feature)

```
<Tip warning={true}>

This API is quite new and still in its experimental stage. While we strive to provide a stable API, it's possible some small parts of the public API will change in the future.

</Tip>
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,100 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+> accelerate launch --use-fsdp --fsdp-activation-checkpointing=TRUE examples/torch_native_parallelism/nd_parallel.py --cp-size=8 --sequence-length=128000
 > ```
 
+## DeepSpeed/ALST/UlyssesSP backend
+
+ALST/UlyssesSP implements a sequence parallelism using attention head parallelism as explained in [this paper](https://arxiv.org/abs/2506.13996) - for simplicity we re-use the concept and the setup of context parallelism, which from the user's end of view is the same - multiple gpus are used to process a single batch.
+
+To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU and a few other things enabled to get the full power of ALST, for details please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
+
```
SunMarc: what would it take to enable the other features ?

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,100 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+- `seq_length` and `seq_length_is_variable` are used to deal with sequence lengths. If `seq_length_is_variable=True` the backend will work with a sequence length that may change between batches, in which case `seq_length` value can be set to anything divisible by the context parallel degree or not set at all. In this case on every `forward` the sequence variables will be derived from input. If `False` then `seq_length` needs to match the batch's sequence length dimension, which then will have to be padded to be always the same. The default is `True`.
+- `attn_implementation` is one of `sdpa`, `flash_attention_2` or ``flash_attention_3`. This sequence parallel implementation uses `position_ids` instead of `attention_mask` therefore `eager` can't work here until it'd support working with `position_ids`. Also please note that `sdpa` doesn't handle correctly combined into one multiple-samples, it'd attend to the whole sample as one. If the samples aren't combined `sdpa` will work correctly. Therefore Flash Attention should be the ideal choise as it always works.
```
SunMarc:
```suggestion
- `attn_implementation` is one of `sdpa`, `flash_attention_2` or `flash_attention_3`. This sequence parallel implementation uses `position_ids` instead of `attention_mask` therefore `eager` can't work here until it'd support working with `position_ids`. Also please note that `sdpa` doesn't handle correctly multiple samples combined into one, it'd attend to the whole sample as one. If the samples aren't combined `sdpa` will work correctly. Therefore Flash Attention should be the ideal choice as it always works.
```

src/accelerate/parallelism_config.py
```diff
@@ -254,11 +260,13 @@ def __post_init__(self):
...
             if self.tp_handler is None:
-                self.tp_handler = TorchTensorParallelConfig()
+                self.tp_handler = TorchTensorParallelConfig() if self.backend == "torch" else None
 
         if self.cp_size > 1:
             if self.cp_handler is None:
-                self.cp_handler = TorchContextParallelConfig()
+                self.cp_handler = (
+                    TorchContextParallelConfig() if self.backend == "torch" else DeepSpeedContextParallelConfig()
+                )
```
SunMarc: maybe we can perform some checks for `backend` to make sure that the user passed the correct `tp_handler` or `cp_handler` class ?

tests/test_dataclasses.py
```diff
@@ -246,5 +246,23 @@ def test_cp_handler(self):
...
+        """Test CP DeepSpeed/ALST/UlyssesSP handler with various configurations."""
+
+        # Any cp_size > 1 requires torch >= BETA_CP_AVAILABLE_PYTORCH_VERSION, we use placeholder for this check as this test doesn't depend on a specific size
+        if _should_skip_cp_test(2):
+            pytest.skip(f"tests with `cp_size>1` require torch >= {BETA_CP_AVAILABLE_PYTORCH_VERSION}")
+
+        from accelerate.utils import DeepSpeedContextParallelConfig
+
+        cp_handler = DeepSpeedContextParallelConfig()
+        pc = ParallelismConfig(cp_size=2, cp_handler=cp_handler)
```
SunMarc:
```suggestion
        pc = ParallelismConfig(backend="deepspeed", cp_size=2, cp_handler=cp_handler)
```

SunMarc (approved):
Thanks for adding the docs and the tests, this looks really nice. Just some minor nits

stas00:
Taking this out of code comments so that it doesn't disappear with 'Resolve conversation'

>> To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU and a few other things enabled to get the full power of ALST, for details please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).

> @SunMarc: what would it take to enable the other features ?

Those features belong to 
1. HF Transformers:
- `TiledFusedLogitsLoss` https://github.com/huggingface/transformers/issues/41306 - status: discussion
- `TiledMLP` https://github.com/huggingface/transformers/issues/41826 - status: just created
2. HF Transformers and pytorch
- Activation checkpoint offload to cpu/nvme - trying to get this into pytorch core first  https://github.com/pytorch/pytorch/issues/158657 - status: discussion - once pytorch supports it, we can easily integrate into HF Transformers.

docs/source/concept_guides/context_parallelism.md
stas00: https://github.com/huggingface/accelerate/pull/3817#issuecomment-3438216249

src/accelerate/accelerator.py
stas00: sounds good, Marc - I will work on that shortly.

src/accelerate/parallelism_config.py
stas00: good call, Marc. Added and added the corresponding test.

stas00:
@SunMarc, HF Trainer integration: https://github.com/huggingface/transformers/pull/41832

src/accelerate/accelerator.py
stas00: I can see now it's indeed a problem in HF Trainer as you said. Matej wrote it. Let me see if I can find a workaround.

stas00: ok, so I managed to get some HF Trainer tests going and this setup doesn't quite work. if I pass a model in trainer:

```
dataloader = self.accelerator.prepare(DataLoader(dataset, **dataloader_params)), model)
```
everything gets borked - the epoch iterator returns `<accelerate.data_loader.DataLoaderShard object at 0x7f13d1f148c0>` instead of batches of data.

If I temporarily hack around this and recreate the model manually in `_prepare_deepspeed` I overcome this but then it never sends the DL to  `_prepare_deepspeed`, which is absolutely crucial, since the DL needs to be wrapped in a special adapter.

I'm not sure what to do here. the DL never gets passed to `_prepare_deepspeed` in HF Trainer case, but it should be working somehow in the normal DS use case... perhaps it just was never passed to `_prepare_deepspeed` in the past and that needs to change?

Writing my own HF Accelerate script was simple, but HF Trainer is ... Any ideas?

stas00:
As I'm working on HF Trainer integration the other thing came up. In the case of UlyssesSP we re-use the same gpus for DP and SP/CP - so if you have 4 gpus it's either:
- dp=4, cp/sp=4 - all 4 gpus are used both for cp and dp, thanks to DL magic
- dp=4, cp/sp=2 - there are 2 replicas of CP/SP

The special DL takes a unique input from each rank and then uses all gpus to process a single batch from gpu0, then processes a batch from gpu1, etc... So it's like normal DP, except it takes SP/CP times to process all inputs.

So the existing CP API/agreement doesn't fit too well here. Thoughts?

SP = sequence parallelism (as it's not context parallelism even though we try to use the same API).

src/accelerate/accelerator.py
stas00: update: I came up with a workaround where I wrap the DL in an additional accelerate call I added after model gets prepared in HF Trainer. Otherwise it's madness.

I'm now chasing the train batch size problem - will update when I have something working.

stas00:
@SunMarc, it's unclear how to configure the PC plugin, should this work?

```
parallelism_config:
  parallelism_config_dp_replicate_size: 1
  parallelism_config_dp_shard_size: 1
  parallelism_config_tp_size: 1
  parallelism_config_cp_size: 2

  -------------- (just visual separator to point to the next line) -----
  parallelism_config_cp_comm_strategy: alltoall
```

asking here:
https://github.com/huggingface/transformers/pull/41860#issuecomment-3451784095

otherwise users can't configure those things other than by writing code, which seems incomplete/awkward/not-user-friendly.

So they should be able to do this:

```
parallelism_config:
  parallelism_config_dp_replicate_size: 1
  parallelism_config_dp_shard_size: 1
  parallelism_config_tp_size: 1
  parallelism_config_cp_size: 2
    -------------- (just visual separator to point to the next line) -----
  parallelism_config_backend: deepspeed
  parallelism_config_seq_length_is_variable: true
  parallelism_config_attn_implementation: flash_attention_2
```
or should there be another sub-section instead for the PC plugins?

This goes back to your question about whether env vars should have a sub-space or share the common prefix?

---

edit: I made the above work.

stas00:
@SunMarc, this is ready.

src/accelerate/accelerator.py
stas00: I sorted this out.

docs/source/concept_guides/context_parallelism.md
```diff
@@ -142,23 +142,55 @@ accelerator = Accelerator(
...
 
 Instead of setting these values in `DeepSpeedContextParallelConfig` object, you can also use the environment variables to accomplish the same - here they are correspondingly to the end of the list above.
+- `PARALLELISM_CONFIG_CP_BACKEND`
 - `PARALLELISM_CONFIG_CP_SEQ_LENGTH`
 - `PARALLELISM_CONFIG_CP_SEQ_LENGTH_IS_VARIABLE`
 - `PARALLELISM_CONFIG_CP_ATTN_IMPLEMENTATION`
 
-If not passed in the code `cp_size` can be set via `--parallelism_config_cp_size` CLI argument.
+If not passed in the code `cp_size` can be set via `--parallelism_config_cp_size` CLI argument. SAme for other arguments. You can also do the accelerate config file style config, e.f. for 2 gpus:
+
```
SunMarc:
```suggestion
If not passed in the code `cp_size` can be set via `--parallelism_config_cp_size` CLI argument. Same for other arguments. You can also do the accelerate config file style config, e.f. for 2 gpus:

```

src/accelerate/accelerator.py
```diff
@@ -2338,6 +2315,54 @@ def _prepare_deepspeed(self, *args):
                 # It should be done by the launcher but it does not work for multi-node runs
                 os.environ["DEEPSPEED_USE_HPU"] = "true"
 
+            mpu = None
+            if cp_size > 1:
```
SunMarc: let's add somewhere a check that we have the correct backend

src/accelerate/parallelism_config.py
```diff
@@ -257,16 +260,27 @@ def __post_init__(self):
             self.tp_size = int(os.environ.get("PARALLELISM_CONFIG_TP_SIZE", "1"))
         if self.cp_size is None:
             self.cp_size = int(os.environ.get("PARALLELISM_CONFIG_CP_SIZE", "1"))
+        if self.cp_backend is None:
+            self.cp_backend = os.environ.get("PARALLELISM_CONFIG_CP_BACKEND", "torch")
 
         if self.tp_size > 1:
             if self.tp_handler is None:
-                self.tp_handler = TorchTensorParallelConfig() if self.backend == "torch" else None
+                self.tp_handler = TorchTensorParallelConfig() if self.cp_backend == "torch" else None
```
SunMarc:
```suggestion
                self.tp_handler = TorchTensorParallelConfig()
```

For now, let's just use `TorchTensorParallelConfig` in all cases as we don't really need a `tp_backend` for now.

src/accelerate/state.py
```diff
@@ -991,7 +991,7 @@ def __init__(
                 if not os.environ.get("ACCELERATE_ALLOW_CP_STANDALONE", "false").lower() == "true":
                     if self.parallelism_config and self.parallelism_config.cp_enabled and fsdp_plugin is None:
                         raise ValueError(
-                            "`cp_size > 1` specified in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use context parallelism, as we also shard the model across the device mesh to save more memory"
+                            "`cp_size > 1` specified in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use context parallelism with backend==torch, as we also shard the model across the device mesh to save more memory"
```
SunMarc:
```suggestion
                            "`cp_size > 1` specified in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use context parallelism with `cp_backend=torch`, as we also shard the model across the device mesh to save more memory"
```

tests/test_dataclasses.py
```diff
@@ -256,12 +256,17 @@ def test_cp_deepspeed_handler(self):
         from accelerate.utils import DeepSpeedContextParallelConfig
 
         cp_handler = DeepSpeedContextParallelConfig()
-        pc = ParallelismConfig(cp_size=2, cp_handler=cp_handler)
-
+        pc = ParallelismConfig(backend="deepspeed", cp_size=2, cp_handler=cp_handler)
```
SunMarc:
```suggestion
        pc = ParallelismConfig(cp_backend="deepspeed", cp_size=2, cp_handler=cp_handler)
```

tests/test_dataclasses.py
```diff
@@ -256,12 +256,17 @@ def test_cp_deepspeed_handler(self):
...
         assert pc.cp_handler is not None, "CP handler should be set"
         assert pc.cp_handler.seq_length_is_variable is True, "by default we set to expect a variable seqlen"
 
-        with pytest.raises(ValueError, match=f"Invalid attn_implementation"):
+        with pytest.raises(
+            ValueError,
+            match="ParallelismConfig's backend=torch requires <class 'accelerate.utils.dataclasses.TorchContextParallelConfig'>, but cp_handler was set to <class 'accelerate.utils.dataclasses.DeepSpeedContextParallelConfig'",
```
SunMarc:
```suggestion
            match="ParallelismConfig's cp_backend=torch requires <class 'accelerate.utils.dataclasses.TorchContextParallelConfig'>, but cp_handler was set to <class 'accelerate.utils.dataclasses.DeepSpeedContextParallelConfig'",
```

tests/test_dataclasses.py
```diff
@@ -256,12 +256,17 @@ def test_cp_deepspeed_handler(self):
...
+        ):
+            pc = ParallelismConfig(backend="torch", cp_size=2, cp_handler=cp_handler)
```
SunMarc:
```suggestion
            pc = ParallelismConfig(cp_backend="torch", cp_size=2, cp_handler=cp_handler)
```

SunMarc (approved):
Thanks for improving the PR ! Just a few minor nits

src/accelerate/accelerator.py
stas00: done.

stas00:
FYI, the added e2e test will break until deepspeed==0.18.2 is released (or we could switch it to test with `deepspeed@master` until then) as I have just pushed there a more user-friendly API that I have already incorporated here to simplify logic.

src/accelerate/accelerator.py
```diff
@@ -2080,6 +2080,7 @@ def _prepare_deepspeed(self, *args):
         is_dataloader_present = any(isinstance(obj, torch.utils.data.DataLoader) for obj in args)
         tp_size = deepspeed_plugin.deepspeed_config.get("tensor_parallel", {}).get("autotp_size", 0)
 
+        cp_backend = self.parallelism_config.cp_backend if self.parallelism_config else 1
```
SunMarc:
```suggestion
        cp_backend = self.parallelism_config.cp_backend if self.parallelism_config else None
```

src/accelerate/accelerator.py
```diff
@@ -2317,6 +2318,9 @@ def _prepare_deepspeed(self, *args):
 
             mpu = None
             if cp_size > 1:
+                if cp_backend != "deepspeed":
+                    raise ValueError("In order to use the configured {cp_size=} with DeepSpeed, you need to configure cp_backend='deepspeed', yet you configured it to be {cp_backend=}.")
+
```
SunMarc:
```suggestion
                    raise ValueError(f"In order to use the configured {cp_size=} with DeepSpeed, you need to configure cp_backend='deepspeed', yet you configured it to be {cp_backend=}.")

```

stas00: weird, it says outdated - pushed your proposal manually.

stas00:
Thank you for the additional fixes, Marc

stas00:
please don't merge this yes, waiting for a new DS release to anchor on.

examples/alst_ulysses_sequence_parallelism/cp-alst.py
```diff
@@ -0,0 +1,156 @@
...
+micro_batch_size = 1
+
+parallelism_config = ParallelismConfig(
+    backend="deepspeed",
+    cp_size=4,
+    # dp_shard_size=1, # set if dp is wanted as well
+    cp_handler=DeepSpeedContextParallelConfig(
+        seq_length=256,
+        seq_length_is_variable=True,
+        attn_implementation="sdpa",
```
kashif:
```suggestion
    cp_backend="deepspeed",
    cp_size=4,
    # dp_shard_size=1, # set if dp is wanted as well
    cp_handler=DeepSpeedContextParallelConfig(
        cp_seq_length=256,
        cp_seq_length_is_variable=True,
        cp_attn_implementation="sdpa",
```

examples/alst_ulysses_sequence_parallelism/README.md
```diff
@@ -0,0 +1,19 @@
+# Deepspeed's ALST/Ulysses sequence parallelism
+
+This is an example of the use of Ulysses Sequence Parallelism, which uses attention head parallelism and which is part of the Arctic Long Sequence Training project at [ArcticTraining](https://github.com/snowflakedb/ArcticTraining). [This paper](https://arxiv.org/abs/2506.13996) goes into the details of this protocol.
```
kashif:
```suggestion
This is an example of the use of Ulysses Sequence Parallelism, which uses attention head parallelism and is part of the Arctic Long Sequence Training project at [ArcticTraining](https://github.com/snowflakedb/ArcticTraining). [This paper](https://arxiv.org/abs/2506.13996) goes into the details of this protocol.
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -24,6 +24,13 @@ With sequence length of 128k, the memory requirement of the attention matrix is
 
 Context parallelism allows us to shard the inputs to the attention computation along the sequence dimension and compute the attention in parallel on multiple GPUs. With this, we can train models with long sequences, scaling potentially to 1M+ sequence length.
 
+## Supported context parallelism backends
+
+Multiple context parallelism backends are currently supported
+
+1. `torch`: PyTorch/FSDP2,which implements several of Ring Attention context parallel protocols [tutorial](https://docs.pytorch.org/tutorials/unstable/context_parallel.html) and [api](https://docs.pytorch.org/docs/stable/distributed.tensor.html#torch.distributed.tensor.experimental.context_parallel).
```
kashif:
```suggestion
1. `torch`: PyTorch/FSDP2, which implements several of Ring Attention context parallel protocols [tutorial](https://docs.pytorch.org/tutorials/unstable/context_parallel.html) and [api](https://docs.pytorch.org/docs/stable/distributed.tensor.html#torch.distributed.tensor.experimental.context_parallel).
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+> These examples were created with a script you can find [in the examples folder](https://github.com/huggingface/accelerate/blob/main/examples/torch_native_parallelism/nd_parallel.py). To run the example on 8 H100 GPUs (128k sequence length), you can use the following command:
 > ```bash
-> accelerate launch --use-fsdp --fsdp-activation-checkpointing=TRUE examples/fsdp2/nd_parallel.py --cp-size=8 --sequence-length=128000
+> accelerate launch --use-fsdp --fsdp-activation-checkpointing=TRUE examples/torch_native_parallelism/nd_parallel.py --cp-size=8 --sequence-length=128000
 > ```
 
+## DeepSpeed/ALST/UlyssesSP backend
+
+ALST/UlyssesSP implements a sequence parallelism using attention head parallelism as explained in [this paper](https://arxiv.org/abs/2506.13996) - for simplicity we re-use the concept and the setup of context parallelism, which from the user's end of view is the same - multiple gpus are used to process a single batch.
```
kashif:
```suggestion
ALST/UlyssesSP implements sequence parallelism using attention head parallelism, as explained in [this paper](https://arxiv.org/abs/2506.13996). For simplicity, we reuse the concept and setup of context parallelism, which, from the user's perspective, is the same: multiple GPUs are used to process a single batch.
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+
+To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU and a few other things enabled to get the full power of ALST, for details please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
```
kashif:
```suggestion
To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components, so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU, and a few other things enabled to get the full power of ALST. For details, please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+- `cp_seq_length` and `cp_seq_length_is_variable` are used to deal with sequence lengths. If `cp_seq_length_is_variable=True` the backend will work with a sequence length that may change between batches, in which case `cp_seq_length` value can be set to anything divisible by the context parallel degree or not set at all. In this case on every `forward` the sequence variables will be derived from input. If `False` then `seq_length` needs to match the batch's sequence length dimension, which then will have to be padded to be always the same. The default is `True`.
+- `cp_attn_implementation` is one of `sdpa`, `flash_attention_2` or `flash_attention_3`. This sequence parallel implementation uses `position_ids` instead of `attention_mask` therefore `eager` can't work here until it'd support working with `position_ids`. Also please note that `sdpa` doesn't handle correctly multiple samples combined into one, it'd attend to the whole sample as one. If the samples aren't combined `sdpa` will work correctly. Therefore Flash Attention should be the ideal choice as it always works.
```
kashif:
```suggestion
- `cp_attn_implementation` is one of `sdpa`, `flash_attention_2` or `flash_attention_3`. This sequence parallel implementation uses `position_ids` instead of `attention_mask` therefore, `eager` can't work here until it supports working with `position_ids`. Also, please note that `sdpa` doesn't handle multiple samples combined into one correctly; it will attend to the whole sample as one. If the samples aren't combined, `sdpa` will work correctly. Therefore, Flash Attention should be the ideal choice as it always works.
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+
+Instead of setting these values in `DeepSpeedContextParallelConfig` object, you can also use the environment variables to accomplish the same - here they are correspondingly to the end of the list above.
+- `PARALLELISM_CONFIG_CP_BACKEND`
+- `PARALLELISM_CONFIG_CP_SEQ_LENGTH`
+- `PARALLELISM_CONFIG_CP_SEQ_LENGTH_IS_VARIABLE`
+- `PARALLELISM_CONFIG_CP_ATTN_IMPLEMENTATION`
+
+If not passed in the code `cp_size` can be set via `--parallelism_config_cp_size` CLI argument. Same for other arguments. You can also do the accelerate config file style config, e.f. for 2 gpus:
```
kashif:
```suggestion
If not passed in the code, `cp_size` can be set via `--parallelism_config_cp_size` CLI argument. Same for other arguments. You can also do the accelerate config file style config, e.g., for 2 GPUs:
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+    cp_size=2,
+    dp_shard_size=2,
+    cp_handler=DeepSpeedContextParallelConfig(...),
+)
+```
+Here we use 4 gpus, with 2 sequence parallelism replicas. Deepspeed-ZeRO is what drives the data parallelism.
+
+Please note that a lot of magic is hidden inside [UlyssesSPDataLoaderAdapter](https://github.com/deepspeedai/DeepSpeed/blob/64c0052fa08438b4ecf4cae30af15091a92d2108/deepspeed/runtime/sequence_parallel/ulysses_sp.py#L442). It's used behind the scenes, wrapping your original DataLoader object, but you should be aware of it should you run into any problems. It also automatically injects the correct `shift_labels` into the batch dictionary, before the batch gets sharded across the participating ranks.
+
+Now the only remaining piece to start using ALST/UlyssesSP is to aggregate the loss across ranks using a differentiable `all_gather` to get the grads right. The following code does it, while also exlcuding any masked out with `-100` tokens, to get the correct average:
```
kashif:
```suggestion
Now the only remaining piece to start using ALST/UlyssesSP is to aggregate the loss across ranks using a differentiable `all_gather` to get the grads right. The following code does it, while also excluding any masked out with `-100` tokens, to get the correct average:
```

docs/source/concept_guides/context_parallelism.md
```diff
@@ -91,11 +113,136 @@ This can scale your context size to 1M+ sequence length potentially. Below, we s
...
+```
+
+If you use [Liger Kernel](https://github.com/linkedin/Liger-Kernel) it already knows how to handle `shift_labels` so you don't need to go through manual loss calculation, just calling `model(**batch)` will already get the `loss` calculated and done in a very memory-efficient way. If you didn't know about Liger-Kernel - it's highly recommended to be used especially for long sequence length since it liberates a lot of working memory that can be used for handling longer sequences.
+
+If you want to see what HF Accelerate did behind the scenes please read [this full integration tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
+
+For an example of an Accelerate training loop with enabled ALST/UlyssesSP see [examples/alst_ulysses_sequence_parallelism](https://github.com/huggingface/accelerate/blob/main/examples/alst_ulysses_sequence_parallelism).
+
+[!Warning]
+> This API is quite new and still in its experimental stage. While we strive to provide a stable API, it's possible some small parts of the public API will change in the future.
```
kashif:
```suggestion
> This API is quite new and still in its experimental stage. While we strive to provide a stable API, some small parts of the public API may change in the future.
```

stas00:
Thank you for the thorough prose review, Kashif. Now I need to find time to rename `s/cp/sp/` here and in the HF Trainer PR, to pave the way for cp+sp 2d parallelism.

stas00:
@kashif, @SunMarc the `s/cp/sp/` conversion is done. I have also added a dedicated doc to SP, while cross-linking SP and CP docs.

Otherwise to save your time most changes were just a careful `s/cp/sp/` and replicating some `cp_*` methods where needed, over what you have already reviewed.

I left the `(sp|cp)_backend` from the original implementation since it has already been coded and paves way for other backends. There are quite a few papers out there with variations on both CP and SP. And then at some point someone could code the USP paper to allow SP+CP.

Let me know when this is OK and I will adapt the HF Trainer PR to match this one. 

But don't just merge it yet, should I discover the HF Trainer side needs more tweaks.

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+accelerate launch --parallelism-config-sp-size 8  ...
+```
+
+> [!Tip]
+> You can also set the `sp_size` and other configuration in the `accelerate config` command, which will save them in your `accelerate` configuration file, so you don't have to pass them every time you launch your script.
+
+> [!Tip]
+> sequence parallelism combines with data parallelism. It doesn't require additional GPUs.
+> So if you have 8 gpus you can do: `--parallelism-config-dp-size 8 --parallelism-config-sp-size 8`. Or you can use the `ParallelismConfig` class to set them programmatically.
+
```
SunMarc: from what you wrote below , this will use 16 gpus no ?

> Here we use 4 gpus, with 2 sequence parallelism replicas. Deepspeed-ZeRO is what drives the data parallelism here.

src/accelerate/commands/launch.py
```diff
@@ -816,21 +816,36 @@ def launch_command_parser(subparsers=None):
...
+        help="Sequence length for when batches are all of the same length. For variable sequence lengths across batches set `parallelism_config_sp_seq_length_is_variable=True`",
     )
 
     parallelism_config_args.add_argument(
-        "--parallelism_config_cp_seq_length_is_variable",
+        "--parallelism_config_sp_seq_length_is_variable",
         type=bool,
         default=True,
-        help="If `True` will work with a sequence length that may change between batches, in which case `parallelism_config_cp_seq_length` value can be set to anything divisible by cp size or remain unset. If `False` then `parallelism_config_cp_seq_length` needs to match the batch's sequence length dimension. The default is `True`.",
+        help="If `True` will work with a sequence length that may change between batches, in which case `parallelism_config_sp_seq_length` value can be set to anything divisible by cp size or remain unset. If `False` then `parallelism_config_sp_seq_length` needs to match the batch's sequence length dimension. The default is `True`.",
```
SunMarc:
```suggestion
        help="If `True` will work with a sequence length that may change between batches, in which case `parallelism_config_sp_seq_length` value can be set to anything divisible by sp size or remain unset. If `False` then `parallelism_config_sp_seq_length` needs to match the batch's sequence length dimension. The default is `True`.",
```

src/accelerate/parallelism_config.py
```diff
@@ -52,7 +52,12 @@ class ParallelismConfig:
             The size of the context parallel group. Currently not supported, but reserved for future use and enabled
             for downstream libraries.
         cp_backend (`str`, defaults to `torch`):
-            Which CP backend to use. `torch` (FSDP2) or `deepspeed` (ALST/Ulysses)
+            Which CP backend to use: `torch` (FSDP2)
+        sp_size (`int`, defaults to `1`):
+            The size of the context parallel group. Currently not supported, but reserved for future use and enabled
+            for downstream libraries.
+        sp_backend (`str`, defaults to `torch`):
+            Which CP backend to use:`deepspeed` (ALST/Ulysses)
```
SunMarc:
```suggestion
        sp_backend (`str`, defaults to `torch`):
            Which SP backend to use:`deepspeed` (ALST/Ulysses)
```

src/accelerate/parallelism_config.py
```diff
@@ -52,7 +52,12 @@ class ParallelismConfig:
             The size of the context parallel group. Currently not supported, but reserved for future use and enabled
             for downstream libraries.
         cp_backend (`str`, defaults to `torch`):
-            Which CP backend to use. `torch` (FSDP2) or `deepspeed` (ALST/Ulysses)
+            Which CP backend to use: `torch` (FSDP2)
+        sp_size (`int`, defaults to `1`):
+            The size of the context parallel group. Currently not supported, but reserved for future use and enabled
+            for downstream libraries.
```
SunMarc:
```suggestion
        sp_size (`int`, defaults to `1`):
            The size of the sequence parallel group.
```

src/accelerate/parallelism_config.py
```diff
@@ -262,26 +279,41 @@ def __post_init__(self):
...
+                sp_backends_config_map = dict(
+                    deepspeed=DeepSpeedSequenceParallelConfig,
+                )
+                if self.sp_backend not in sp_backends_config_map or not isinstance(
+                    self.sp_handler, sp_backends_config_map[self.sp_backend]
+                ):
+                    raise ValueError(
+                        f"DeepSpeedSequenceParallelConfig requires sp_backends=deepspeed, but sp_handler was set to {type(self.sp_handler)}"
+                    )
+
```
SunMarc: Let's remove this as this is not needed for now
```suggestion
```

src/accelerate/parallelism_config.py
SunMarc:
```suggestion
        sp_backend (`str`, defaults to `deepspeed`):
            Which CP backend to use:`deepspeed` (ALST/Ulysses)
```

src/accelerate/parallelism_config.py
```diff
@@ -262,26 +279,41 @@ def __post_init__(self):
             self.cp_size = int(os.environ.get("PARALLELISM_CONFIG_CP_SIZE", "1"))
         if self.cp_backend is None:
             self.cp_backend = os.environ.get("PARALLELISM_CONFIG_CP_BACKEND", "torch")
+        if self.sp_size is None:
+            self.sp_size = int(os.environ.get("PARALLELISM_CONFIG_SP_SIZE", "1"))
+        if self.sp_backend is None:
+            self.sp_backend = os.environ.get("PARALLELISM_CONFIG_SP_BACKEND", "torch")
```
SunMarc:
```suggestion
        if self.sp_backend is None:
            self.sp_backend = os.environ.get("PARALLELISM_CONFIG_SP_BACKEND", "deepspeed")
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+## How to use sequence parallelism?
+
+```diff
+from accelerate.utils import ParallelismConfig, DeepSpeedSequenceParallelConfig
+
+parallelism_config = ParallelismConfig(
++     sp_backend="deepspeed",
++     sp_size=4,
++     sp_handler=DeepSpeedContextParallelConfig(
```
SunMarc:
```suggestion
+     sp_handler=DeepSpeedSequenceParallelConfig(
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+
+To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components, so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU, and a few other things enabled to get the full power of ALST. For details, please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
+
+To configure the `deepspeed` backend:
+
+```python
+parallelism_config = ParallelismConfig(
+    sp_backend="deepspeed",
+    sp_size=4,
+    sp_handler=DeepSpeedContextParallelConfig(
```
SunMarc:
```suggestion
    sp_handler=DeepSpeedSequenceParallelConfig(
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+- `sp_attn_implementation` is one of `sdpa`, `flash_attention_2` or `flash_attention_3`. This sequence parallel implementation uses `position_ids` instead of `attention_mask` therefore, `eager` can't work here until it supports working with `position_ids`. Also, please note that `sdpa` doesn't handle multiple samples combined into one correctly; it will attend to the whole sample as one. If the samples aren't combined, `sdpa` will work correctly. Therefore, Flash Attention should be the ideal choice as it always works.
+
+Instead of setting these values in `DeepSpeedContextParallelConfig` object, you can also use the environment variables to accomplish the same - here they are correspondingly to the end of the list above.
```
SunMarc:
```suggestion
Instead of setting these values in `DeepSpeedSequenceParallelConfig` object, you can also use the environment variables to accomplish the same - here they are correspondingly to the end of the list above.
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+As mentioned earlier Ulysses sequence parallelism is normally overlayed with data parallelism - same ranks are used for feeding unique data streams and also perform Ulysses Sequence Parallelism. But you could also create replicas like so:
+
+```python
+parallelism_config = ParallelismConfig(
+    dp_shard_size=2,
+    sp_backend="deepspeed",
+    sp_size=2,
+    sp_handler=DeepSpeedContextParallelConfig(...),
```
SunMarc:
```suggestion
    sp_handler=DeepSpeedSequenceParallelConfig(...),
```

tests/test_dataclasses.py
```diff
@@ -246,28 +252,28 @@ def test_cp_torch_handler(self):
...
+        from accelerate.utils import DeepSpeedSequenceParallelConfig
 
-        cp_handler = DeepSpeedContextParallelConfig()
-        pc = ParallelismConfig(cp_backend="deepspeed", cp_size=2, cp_handler=cp_handler)
-        assert pc.cp_handler is not None, "CP handler should be set"
-        assert pc.cp_handler.cp_seq_length_is_variable is True, "by default we set to expect a variable seqlen"
+        sp_handler = DeepSpeedSequenceParallelConfig()
+        pc = ParallelismConfig(sp_backend="deepspeed", sp_size=2, sp_handler=sp_handler)
+        assert pc.sp_handler is not None, "CP handler should be set"
+        assert pc.sp_handler.sp_seq_length_is_variable is True, "by default we set to expect a variable seqlen"
```
SunMarc:
```suggestion
        assert pc.sp_handler is not None, "SP handler should be set"
        assert pc.sp_handler.sp_seq_length_is_variable is True, "by default we set to expect a variable seqlen"
```

tests/test_dataclasses.py
```diff
@@ -246,28 +252,28 @@ def test_cp_torch_handler(self):
...
 
         with pytest.raises(
             ValueError,
-            match="ParallelismConfig's cp_backend=torch requires <class 'accelerate.utils.dataclasses.TorchContextParallelConfig'>, but cp_handler was set to <class 'accelerate.utils.dataclasses.DeepSpeedContextParallelConfig'",
+            match="DeepSpeedSequenceParallelConfig requires sp_backends=deepspeed, but sp_handler was set to <class 'accelerate.utils.dataclasses.DeepSpeedSequenceParallelConfig'",
         ):
-            pc = ParallelismConfig(cp_backend="torch", cp_size=2, cp_handler=cp_handler)
+            pc = ParallelismConfig(sp_backend="torch", sp_size=2, sp_handler=sp_handler)
```
SunMarc: we can remove this

src/accelerate/parallelism_config.py
```diff
@@ -187,6 +199,11 @@ def cp_enabled(self):
         """True if context parallelism is enabled, i.e. `cp_size > 1`."""
         return self.cp_size > 1
 
+    @property
+    def sp_enabled(self):
+        """True if context parallelism is enabled, i.e. `cp_size > 1`."""
```
SunMarc:
```suggestion
        """True if context parallelism is enabled, i.e. `sp_size > 1`."""
```

SunMarc (commented):
Thanks for updating and adding new docs ! Just a few minor nits but overall fine. I think we can move to HF Trainer to adapt this. cc @kashif on TRL side

docs/source/concept_guides/sequence_parallelism.md
stas00: no, DP and SP are using the same GPUs.

You can make smaller SP replicas if you wish. E.g. with 4 gpus you can do SP=2, so there will be 2 replicas. But DP is never affected with SP, SP wants a different data stream on each rank. The DL adapter then reads all the streams from all gpus and then processes them 1 sample at a time using all ranks. then the next one.

stas00: Please let me know how I can explain this better. Basically you just leave DP alone and focus on SP.

stas00:
Thank you for catching all the places I missed, Marc!

OK for me to move to sync HF Trainer?

docs/source/concept_guides/sequence_parallelism.md
SunMarc: Okay understood ! just a minor typo but it should be `parallelism-config-dp-shard-size`

```suggestion
> So if you have 8 gpus you can do: `--parallelism-config-dp-shard-size 8 --parallelism-config-sp-size 8`. Or you can use the `ParallelismConfig` class to set them programmatically.

```

docs/source/concept_guides/sequence_parallelism.md
SunMarc: then in this case `dp_shard_size` should be 4 no if we are using 4 gpus. and sp_size should stay at 2 for the 2 sequence parallelism replicas ?

stas00: I think so, yes.

stas00: pushed the correction.

stas00: I run tests and it appears that `dp_replicate_size` is the right config name for this use case - so fixed the docs.

SunMarc:
> OK for me to move to sync HF Trainer?

yeah let's move to the Trainer !

stas00:
Any idea why this test fails with:
```
FAILED tests/test_dataclasses.py::TestParallelismConfig::test_sp_deepspeed_handler - importlib.metadata.PackageNotFoundError: No package metadata was found for deepspeed
```

It works fine on my setup.

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,204 @@
...
+        losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
+        # special dealing with SFT that has prompt tokens that aren't used in loss computation
+        good_tokens = (shift_labels != -100).view(-1).sum()
+        good_tokens_per_rank = torch.distributed.nn.functional.all_gather(
+            good_tokens, group=sp_group
+        )
+        total_loss = sum(
+            losses_per_rank[rank] * good_tokens_per_rank[rank]
+            for rank in range(sp_world_size)
+        )
```
kashif, line 198:
```suggestion
        # Skip ranks with zero valid tokens to avoid NaN contamination (NaN * 0 = NaN)
        total_loss = sum(
            losses_per_rank[rank] * good_tokens_per_rank[rank]
            for rank in range(sp_world_size)
            if good_tokens_per_rank[rank] > 0
        )
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,206 @@
...
+```bash
+accelerate launch --parallelism-config-sp-size 8  ...
+```
+
+> [!Tip]
+> You can also set the `sp_size` and other configuration in the `accelerate config` command, which will save them in your `accelerate` configuration file, so you don't have to pass them every time you launch your script.
+
+> [!Tip]
+> sequence parallelism combines with data parallelism. It doesn't require additional GPUs.
+> So if you have 8 gpus you can do: `--parallelism-config-dp-shard-size 8 --parallelism-config-sp-size 8`. Or you can use the `ParallelismConfig` class to set them programmatically.
```
kashif, line 82:
```suggestion
> So if you have 8 gpus you can do: `--parallelism-config-dp-shard-size 8 --parallelism-config-sp-size 8`. Or you can use the `ParallelismConfig` class to set them programmatically.
>
> **Important**: You must ensure `dp_replicate_size × dp_shard_size × sp_size = num_processes`. For example, with 8 GPUs and `sp_size=8`, you need `dp_shard_size=1` (since 1 × 1 × 8 = 8). With 4 GPUs and `sp_size=2`, you could use `dp_shard_size=2` (since 1 × 2 × 2 = 4) for 2D parallelism.
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,208 @@
...
+Currently the only sequence parallelism backend is `deepspeed`, which comes from the modernized Ulysses SP which is part of the [Arctic Long Sequence Training technology](https://arxiv.org/abs/2506.13996). There is also a [tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/) should you want to integrate it into your own code directly.
+
+## How to use sequence parallelism?
+
+```diff
+from accelerate.utils import ParallelismConfig, DeepSpeedSequenceParallelConfig
+
+parallelism_config = ParallelismConfig(
++     sp_backend="deepspeed",
++     sp_size=4,
```
kashif, line 57:
```suggestion
+# Example: 4 GPUs with sp_size=4, dp_shard_size=1
+# Ensure: dp_replicate_size × dp_shard_size × sp_size = 1 × 1 × 4 = 4 GPUs
parallelism_config = ParallelismConfig(
+     sp_backend="deepspeed",
+     sp_size=4,
+     dp_shard_size=1,  # Explicit: no data parallelism
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,211 @@
...
+ALST/UlyssesSP implements sequence parallelism using attention head parallelism, as explained in [this paper](https://arxiv.org/abs/2506.13996). For simplicity, we reuse the concept and setup of sequence parallelism, which, from the user's perspective, is the same: multiple GPUs are used to process a single batch.
+
+To give a sense of what ALST made possible - it allowed us to train in bf16 with 500K tokens on a single H100 GPU, 3.7M on a single node, and 15M on Llama-8B using just four nodes. This feature of HF Accelerate enables only 1 of the 3 ALST components, so the achievable sequence length will be smaller. You'd want TiledMLP, Activation checkpoint offload to CPU, and a few other things enabled to get the full power of ALST. For details, please refer to [this tutorial](https://www.deepspeed.ai/tutorials/ulysses-alst-sequence-parallelism/).
+
+To configure the `deepspeed` backend:
+
+```python
+parallelism_config = ParallelismConfig(
+    sp_backend="deepspeed",
+    sp_size=4,
```
kashif, line 100:
```suggestion
# Example: 4 GPUs with sp_size=4, dp_shard_size=1
# Ensure: dp_replicate_size × dp_shard_size × sp_size = 1 × 1 × 4 = 4 GPUs
parallelism_config = ParallelismConfig(
    sp_backend="deepspeed",
    sp_size=4,
    dp_shard_size=1,  # Explicit: no data parallelism
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,214 @@
...
+
+```yaml
+distributed_type: DEEPSPEED
+deepspeed_config:
+  deepspeed_config_file: path/to/ds_config.json
+machine_rank: 0
+num_machines: 1
+num_processes: 2
+parallelism_config:
+  parallelism_config_sp_size: 2
```
kashif, line 137:
```suggestion
parallelism_config:
  parallelism_config_dp_replicate_size: 1
  parallelism_config_dp_shard_size: 1  # Must satisfy: 1 × 1 × 2 = 2 num_processes
  parallelism_config_sp_size: 2
```

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -0,0 +1,216 @@
...
+  parallelism_config_sp_attn_implementation: sdpa
+
+```
+
+As mentioned earlier Ulysses sequence parallelism is normally overlayed with data parallelism - same ranks are used for feeding unique data streams and also perform Ulysses Sequence Parallelism. But you could also create replicas like so:
+
+```python
+parallelism_config = ParallelismConfig(
+    dp_replicate_size=2,
+    sp_size=2,
```
kashif, line 152:
```suggestion
```python
# Example: 4 GPUs with 2D parallelism (SP=2, DP=2)
# Ensure: dp_replicate_size × dp_shard_size × sp_size = 2 × 1 × 2 = 4 GPUs
parallelism_config = ParallelismConfig(
    dp_replicate_size=2,
    dp_shard_size=1,  # Explicit: no sharding within replicas
    sp_size=2,
```

src/accelerate/test_utils/scripts/external_deps/test_ds_alst_ulysses_sp.py
```diff
@@ -0,0 +1,125 @@
...
+        vocab_size=unwrapped_model.config.vocab_size,
+    )
+
+    if sp_size > 1:
+        # differentiable weighted per-shard-loss aggregation across ranks
+        losses_per_rank = torch.distributed.nn.functional.all_gather(loss, group=sp_group)
+        # special dealing with SFT that has prompt tokens that aren't used in loss computation
+        good_tokens = (shift_labels != -100).view(-1).sum()
+        good_tokens_per_rank = torch.distributed.nn.functional.all_gather(good_tokens, group=sp_group)
+        total_loss = sum(losses_per_rank[rank] * good_tokens_per_rank[rank] for rank in range(sp_world_size))
```
kashif:
```suggestion
        total_loss = sum(
            losses_per_rank[rank] * good_tokens_per_rank[rank]
            for rank in range(sp_world_size)
            if good_tokens_per_rank[rank] > 0
        )
```

src/accelerate/parallelism_config.py
```diff
@@ -41,11 +46,17 @@ class ParallelismConfig:
...
         cp_size (`int`, defaults to `1`):
             The size of the context parallel group. Currently not supported, but reserved for future use and enabled
             for downstream libraries.
-        tp_handler (`~utils.TorchTensorParallelConfig`, defaults to `None`):
-            The handler for the tensor parallel group.
+        cp_backend (`str`, defaults to `torch`):
+            Which CP backend to use: `torch` (FSDP2)
+        sp_size (`int`, defaults to `1`):
+            The size of the sequence parallel group.
+        sp_backend (`str`, defaults to `torch`):
```
kashif:
```suggestion
        sp_backend (`str`, defaults to `deepspeed`):
```

tests/test_dataclasses.py
kashif:
```suggestion
    BETA_SP_AVAILABLE_DEEPSPEED_VERSION,
    BETA_TP_AVAILABLE_PYTORCH_VERSION,
    BETA_TP_AVAILABLE_TRANSFORMERS_VERSION,
)
from accelerate.utils.imports import is_deepspeed_available, is_transformers_available
```

tests/test_dataclasses.py
```diff
@@ -32,6 +33,11 @@ def _should_skip_cp_test(cp_size):
     return cp_size > 1 and not is_torch_version(">=", BETA_CP_AVAILABLE_PYTORCH_VERSION)
 
 
+def _should_skip_sp_test(sp_size):
+    """Check if CP test should be skipped based on sp_size and deepspeed version."""
+    return sp_size > 1 and not compare_versions("deepspeed", ">=", BETA_SP_AVAILABLE_DEEPSPEED_VERSION)
```
kashif:
```suggestion
def _should_skip_sp_test(sp_size):
    """Check if SP test should be skipped based on sp_size and deepspeed version."""
    if sp_size <= 1:
        return False
    if not is_deepspeed_available():
        return True
    return not compare_versions("deepspeed", ">=", BETA_SP_AVAILABLE_DEEPSPEED_VERSION)
```

tests/test_dataclasses.py
```diff
@@ -20,10 +20,15 @@
...
     BETA_CP_AVAILABLE_PYTORCH_VERSION,
+    BETA_SP_AVAILABLE_DEEPSPEED_VERSION,
     BETA_TP_AVAILABLE_PYTORCH_VERSION,
     BETA_TP_AVAILABLE_TRANSFORMERS_VERSION,
 )
-from accelerate.utils.imports import is_transformers_available
+    BETA_SP_AVAILABLE_DEEPSPEED_VERSION,
+    BETA_TP_AVAILABLE_PYTORCH_VERSION,
+    BETA_TP_AVAILABLE_TRANSFORMERS_VERSION,
+)
```
kashif:
```suggestion
```

tests/test_dataclasses.py
kashif:
```suggestion
def _should_skip_sp_test(sp_size):
    """Check if SP test should be skipped based on sp_size and deepspeed version."""
    if sp_size <= 1:
        return False
    if not is_deepspeed_available():
        return True
    return not compare_versions("deepspeed", ">=", BETA_SP_AVAILABLE_DEEPSPEED_VERSION)
```

SunMarc:
Can we merge this PR @stas00 ?

stas00:
Thank you for further improvements/fixes, Kashif

Yes, let's merge it. Thank you.

egangu:
Merging this PR seems to have invalidated the original cp implement.
In _prepare_cp, the function automatically passes when parallelism_config.sp_backend is "deepspeed"
https://github.com/huggingface/accelerate/blob/d1c96ba9ebcedcab15cf43a98853107b3352b3fd/src/accelerate/accelerator.py#L1639
However, parallelism_config.sp_backend can only be set to "deepspeed"
https://github.com/huggingface/accelerate/blob/d1c96ba9ebcedcab15cf43a98853107b3352b3fd/src/accelerate/parallelism_config.py#L319

kashif:
@egangu so for `CP` one would need to add the appropriate `cp_size` and `cp_backend` configs

egangu:
> @egangu so for `CP` one would need to add the appropriate `cp_size` and `cp_backend` configs

You're right. But what I mean is that the [original CP](https://huggingface.co/docs/accelerate/concept_guides/context_parallelism) cannot be enabled on the current version, no matter how user set the `cp_size` or `cp_backend` configurations.
This is because in the `_prepare_cp` function (the core code that enables the original CP), as I pointed out above, the current implementation automatically skips it.
Rolling back the accelerate version to 1.11 will enable the original CP.

kashif:
thanks for the report @egangu  let me test and fix

PR #3835, opened 2025-11-09
[Bug] Update torch.optim.Optimizer parameter states after tensor parallelism

naomili0924:
_This PR adds two functions:_

1) update torch.optimizer.parameter reference in the self._prepare_cp() function, similar to fsdp_switch_optimizer_parameters
2) In fsdp stage, if model parameters are DTensor (passing through tensor parallelism previously), convert DTensor to Tensor for dist.broadcast()

This PR aims to fix this issue 
https://github.com/huggingface/accelerate/issues/3820
https://github.com/huggingface/accelerate/issues/3821

=========================================================================
_Below are reproduction process:_

Reproduced [this issue](https://github.com/huggingface/accelerate/issues/3820) based on its original yaml and python file given in the issue description:
`accelerate launch --config_file hessian_toolkit/configs/fsdp2_single_node.yaml     minimal_reproduce_accelerate_bug.py
`
However to fully fix this issue we need two parts:

1. Wrap lm_head as a single and separate module in, please modify the python file as:

```
class LmHeadWrapper(torch.nn.Module):
    def __init__(self, lm_head):
        super().__init__()
        self.lm_head = lm_head
    def forward(self, x):
        return self.lm_head(x)

model = AutoModelForCausalLM.from_pretrained(
        MODEL_NAME,
        use_cache=False,
        **model_kwargs
    )

model.lm_head = LmHeadWrapper(model.lm_head)
optimizer = torch.optim.SGD(model.parameters(), lr=0.01)
```
_Why we need this change:_

In AutoModelForCausalLM, the model’s lm_head (an nn.Linear layer with shape [embedding_size, vocab_size]) is automatically added during from_pretrained.
By design, lm_head shares the same weight reference as embed_tokens due to the symmetric input–output projection setup.

Before applying Tensor Parallelism (TP), model.named_parameters() only includes embed_tokens and the backbone parameters — the lm_head is not listed separately.

However, when TP is applied, it explicitly separates embed_tokens and lm_head by assigning them independent DTensor shards. This effectively modifies the model architecture by introducing a distinct lm_head module.

Without wrapping the model with an LmHeadWrapper, a KeyError occurs because the post-TP model’s architecture no longer matches the original state dictionary structure.

image

2. After TP, model's tensors were changed to DTensor so the torch.optim.Optimizer param_group must be redirected to the new DTensor location, otherwise, fsdp2_switch_optimizer_parameters will throw Key Error for old addresses.

With this two change, I run the script and got:
image

# What does this PR do?

[...]

PR #3845, opened 2025-11-22
fix module and optimizer parameter mismatch before prepare_tp_

naomili0924:
# What does this PR do?

1. Add debug helper function to detect tensor address mismatch in optimizer and module
2. Add DTensor type class in get_named_parameters if DTensor exists before _prepare_tp

**Root Cause of Failing test_tp.py from https://github.com/huggingface/accelerate/pull/3835**

1) We use detect_module_and_optimizer_parameter_mismatch right at the beginning of the accelerator.prepare(...) function and found in the test_tp_working, the module starts with 154 DTensors but with self.is_fsdp2 = False. This is reasonable because users can call module tp_parallism before accelerator.prepare()

image

image

2) However, in the _get_named_parameters function, if self.is_fsdp=False, we will get Dtensor's location (which is an abstract class) not the sharded tensor location. As a result, we have 154 parameter address mismatched before prepare_tp.

image

Here what I believe, for torch.optim.Optimizer the only correct Tensor address is Dtensor._local_tensor.data_ptr() not the Dtensor abstract class reference. So we should use _local_tensor as long as DTensor exists in the module.parameter() even when self.is_fsdp2 = False.

Fixes # (issue) https://github.com/huggingface/accelerate/pull/3835

**Before this pull request, test_working_of_tp failed due to address mismatch:** 
image

**With this pull request, we can run test_working_of_tp successfully.**
image
**We can run test_working_of_tp_and_fsdp successfully.**
image

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -634,12 +634,6 @@ class Accelerator:
         # Set a flag tensor for early stopping and other breakpoints
         self.flag_tensor = None
 
-        # Count how many dTensor in the module before preparation because user might call:
-        # tp_mesh, dp_mesh = init_tensor_parallel(...)
-        # model = distribute_module(model, mesh=tp_mesh)  # <- now model has DTensor
-        # model = accelerator.prepare(model)
-        self.initial_dtensor_count = 0
-
         check_os_kernel()
 
     @property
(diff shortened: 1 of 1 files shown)


naomili0924:
@SunMarc please take a look at this pull request addressing the failed test case from https://github.com/huggingface/accelerate/pull/3835

Based on the root cause section I wrote in the description, I don't think it is not caused by https://github.com/huggingface/accelerate/pull/3835 itself, but feel free to ping me for more discussion and modification.

I'm located in the US, sorry if I respond late. Thank you so much!!!

src/accelerate/accelerator.py
```diff
@@ -3958,7 +4008,7 @@ def _get_named_parameters(self, *args, drop_refs=False):
 
                     accessor_mapping[WeightWithDynamicFloat8CastTensor] = "_tensor"
                 # we know we're in FSDP2 so DTensor is available
-                if self.is_fsdp2:
+                if self.is_fsdp2 or self.initial_dtensor_count > 0:
```
SunMarc: the best fix here would be to just change the condition to one where we check if distributed is available + pytorch version.

SunMarc (commented):
Thanks, left a comment. Also, left a comment here, would you mind fixing it ? https://github.com/huggingface/accelerate/pull/3835#discussion_r2556863426

src/accelerate/accelerator.py
naomili0924: Do you mean change the condition to:

`isinstance(tensor, DTensor)? `

SunMarc: this is why we do in transformers for example to check if dtensor is available:

```python
_torch_distributed_available = torch.distributed.is_available()
_is_dtensor_available = _torch_distributed_available and is_torch_greater_or_equal("2.5")
if _is_dtensor_available:
    from torch.distributed.tensor import DTensor
```

naomili0924: Thanks for your advice, changed in this pull request.

naomili0924:
> Thanks, left a comment. Also, left a comment here, would you mind fixing it ? [#3835 (comment)](https://github.com/huggingface/accelerate/pull/3835#discussion_r2556863426)

Thanks for your help!

I checked the prepare_tp won't modify the key value in the module.parameter. 
As a result, new_named_params and old_named_params share the same key, so we can safely remove fsdp2_canonicalize_names here.

src/accelerate/accelerator.py
```diff
@@ -1410,6 +1417,48 @@ def _prepare_one(self, obj, first_pass=False, device_placement=None):
...
+        from torch.distributed.tensor import DTensor
+
+        param_reference_dict = {"module": [], "optimizer": [], "parameter_func": [], "dtensor": 0}
+
+        def _get_tensor_address(p):
+            if isinstance(p, DTensor):
+                return p._local_tensor.data_ptr()
+            return p.data_ptr()
+
+        for obj in args:
```
SunMarc: we don't need those then, can you can up a bit the pr ?

naomili0924: sure! cleaned it up.

SunMarc (approved):
Thanks

PR #3847, opened 2025-11-24
feat: added fine tuning example focused on TPUs

tengomucho:
The example script is focused on showing off fine-tuning is possible with current version of accelerate and it is written to be run on TPUs. It has been successfully run and tested on a TPU v5 litepod-8, and it shows how it is possible to perform a fine-tuning task on such hardware thanks to accelerate and FSDPv2, using transformers and Torch XLA.

### DIFF
--- a/examples/finetune_lm_tpu.py
+++ b/examples/finetune_lm_tpu.py
@@ -1,4 +1,3 @@
-
 # Copyright 2025 The HuggingFace Inc. team. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


examples/finetune_lm_tpu.py
```diff
@@ -0,0 +1,157 @@
...
+    args = parser.parse_args()
+
+    # NOTE: this section can be adapted to load any dataset you want.
+    dataset_id = args.dataset_id
+    dolly_dataset = load_dataset(dataset_id, split="train")
+
+    train(
+        model_id=args.model_id,
+        dataset=dolly_dataset,
+    )
```
SunMarc, line 170: since there is no reamdme for this example, is there a specific cli command that you recommend using to run this training script, if so maybe we can put it here or at the top of the file?

tengomucho: Done!

SunMarc: thanks

SunMarc:
@bot /style

PR #3864, opened 2025-11-29
fix: mixed_precision param from accelerate config should be used for FSDP mp

kmehant:
# What does this PR do?

For some reason, previous code uses `self._mixed_precision` to set which is not being set before, instead it should be using the `mixed_precision` param that is being passed to that function. This would then allow us to use accelerate config's `mixed_precision` param directly to set FSDPv1/v2 MP.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -1005,6 +1005,7 @@ class AcceleratorState:
                     self.parallelism_config is not None and self.parallelism_config.cp_enabled
                 ):
                     self.distributed_type = DistributedType.FSDP
+                    self._mixed_precision = mixed_precision
                     if self._mixed_precision != "no" and fsdp_plugin is not None:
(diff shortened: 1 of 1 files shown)


src/accelerate/state.py
```diff
@@ -1005,6 +1005,7 @@ def __init__(
                     self.parallelism_config is not None and self.parallelism_config.cp_enabled
                 ):
                     self.distributed_type = DistributedType.FSDP
+                    self._mixed_precision = mixed_precision
```
SunMarc, line 1008: this should already be set above, if this changed, can you check why ?

kmehant: Just checked! right it was being set! Thank you.

Issue #3885, opened 2025-12-13
CpuOffload pre_forward and post_forward should have @torch.compiler.disable(), otherwise torch.compile fails in many situations it ought to succeed


doctorpangloss:
### System Info

```Shell
- `Accelerate` version: 1.12.0
- Platform: Linux-6.14.0-36-generic-x86_64-with-glibc2.39
- `accelerate` bash location: .venv/bin/accelerate
- Python version: 3.12.11
- Numpy version: 2.2.6
- PyTorch version: 2.9.1+cu128
- PyTorch accelerator: CUDA
- System RAM: 125.71 GB
- GPU type: NVIDIA RTX A5000
- `Accelerate` default config:
	Not found
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

This is kind of a nuanced problem.

First, let's start by saying

```
from accelerate.hooks import CpuOffload
CpuOffload.pre_forward = torch.compiler.disable(CpuOffload.pre_forward)
CpuOffload.post_forward = torch.compiler.disable(CpuOffload.post_forward)
```

fixes the issue that I'm going to report to you. Let's dElVe into why.

Here is the breaking code:

```py
import torch

from diffusers import QwenImagePipeline, QwenImageTransformer2DModel
from sdnq.common import use_torch_compile as triton_is_available
from sdnq import sdnq_post_load_quant

torch._logging.set_logs(graph_breaks=True)
def test_sdnq_qwen_image_fast():
    model_id = "wavespeed/Qwen-Image-bf16"

    transformer = QwenImageTransformer2DModel.from_pretrained(
        model_id,
        subfolder="transformer",
        torch_dtype=torch.bfloat16,
        trust_remote_code=True
    )

    assert transformer.device == torch.device("cpu")

    # quantize
    transformer = sdnq_post_load_quant(
        transformer,
        weights_dtype="int8",
        quantized_matmul_dtype="int8",
        group_size=0,
        svd_rank=32,
        svd_steps=8,
        use_svd=False,
        use_quantized_matmul=triton_is_available,
        use_stochastic_rounding=False,
        dequantize_fp32=False,
        non_blocking=True,
        add_skip_keys=True,
[...]

PR #3888, opened 2025-12-16
Disable hook compile

SunMarc:
# What does this PR do?

This PR fixes https://github.com/huggingface/accelerate/issues/3885. We disable compile in our hooks as it shouldn't be needed since we are just moving data around devices.


sayakpaul (approved):
Perfect!

PR #3908, opened 2026-01-14
fix tp only bug

sywangyi:
@SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1588,32 +1588,38 @@ class Accelerator:
 
         old_named_params = self._get_named_parameters(*tuple(result), drop_refs=True)
 
-        for arg in result:
-            if not isinstance(arg, torch.nn.Module):
-                continue
+        from torch.distributed.tensor import DTensor
 
-            from torch.distributed.tensor import DTensor, Replicate
-            from transformers.integrations.tensor_parallel import ReplicateParallel
-
-            model: torch.nn.Module = arg
-            tp_plan = ReplicateParallel
-
-            for name, param in model.named_parameters():
-                if isinstance(param, DTensor):
+        if self.is_fsdp2:
+            for arg in result:
+                if not isinstance(arg, torch.nn.Module):
                     continue
 
-                dp = DTensor.from_local(param, device_mesh=device_mesh["tp"], placements=[Replicate()])
-                param_name, param_type = name.rsplit(".", 1)
-                module_to_tp = model.get_submodule(param_name)
+                from torch.distributed.tensor import Replicate
+                from transformers.integrations.tensor_parallel import ReplicateParallel
 
-                tp_plan().prepare_module_tp(module_to_tp, device_mesh["tp"])
-                if not isinstance(dp, torch.nn.Parameter):
-                    dp = torch.nn.Parameter(dp, requires_grad=param.requires_grad)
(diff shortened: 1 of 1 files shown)


sywangyi:
```
import argparse

import torch
from transformers import AutoModelForCausalLM, AutoTokenizer, Trainer, TrainingArguments

from utils import get_dataset


MODEL_ID = "NousResearch/Hermes-3-Llama-3.1-8B"


def parse_args():
    parser.add_argument("--sequence-length", type=int, default=1024)
    parser.add_argument("--checkpoint-frequency", type=int, default=100)
    parser.add_argument("--model-name", type=str, default=MODEL_ID)
    parser.add_argument("--save-dir", type=str, default=f"./accelerate-nd-parallel-{MODEL_ID.split('/')[-1]}")
    parser.add_argument("--device-type", type=str, default="auto")
    return parser.parse_args()


def main():
    # If ParallelismConfig is not initialized with __init__, it reads from env vars
    # which were set by using config
    args = parse_args()
    #    pc = ParallelismConfig()
    if args.device_type == "auto":
        args.device_type = torch.accelerator.current_accelerator().type

    model_kwargs = {}
    model_kwargs["tp_plan"] = "auto"

    tokenizer = AutoTokenizer.from_pretrained(args.model_name)
    model = AutoModelForCausalLM.from_pretrained(args.model_name, use_cache=False, **model_kwargs)

    if tokenizer.pad_token is None:
        tokenizer.pad_token = tokenizer.eos_token

    packed_dataset = get_dataset(tokenizer, args.sequence_length)

    training_args = TrainingArguments(
        output_dir=args.save_dir,
        num_train_epochs=1,
        #        parallelism_config=pc,
        per_device_train_batch_size=1,
        logging_steps=5,
        save_steps=args.checkpoint_frequency,
        learning_rate=5e-5,
        remove_unused_columns=False,
        max_steps=200,
        bf16=True,
    )

    trainer = Trainer(
        model=model,
        args=training_args,
        processing_class=tokenizer,
        train_dataset=packed_dataset,
    )

    trainer.train()
    trainer.save_model()


if __name__ == "__main__":
    main()
```
torchrun --nproc_per_node=4 trainer_tp.py --sequence-length 1024

src/accelerate/accelerator.py
```diff
@@ -1588,33 +1588,39 @@ def _prepare_tp(self, *args):
...
-            if not isinstance(arg, torch.nn.Module):
-                continue
+        from torch.distributed.tensor import DTensor
 
-            from torch.distributed.tensor import DTensor, Replicate
-            from transformers.integrations.tensor_parallel import ReplicateParallel
+        if self.is_fsdp2:
+            for arg in result:
+                if not isinstance(arg, torch.nn.Module):
+                    continue
```
SunMarc, line 1596: can you add a comment above to explain why this is needed for fsdpv2 ?

SunMarc (approved):
Thanks a lot ! Just a nit

src/accelerate/accelerator.py
sywangyi: what I know is we could not add this for TP only case, or it will cause crash in model.forward.

SunMarc: ok ! in any case we will probably need to rework a bit tp + fsdpv2 as tp impl changed a bit.

PR #3917, opened 2026-01-30
Enable evaluation during deepspeed Sequence Parallel

jp1924:
# What does this PR do?

Hello accelerate team!

First, I opened this PR to enable evaluation when performing Sequence Parallel.

The history of this PR is connected to different PRs like this:

https://github.com/huggingface/transformers/pull/43517 > https://github.com/deepspeedai/DeepSpeed/pull/7821 > accelerate PR
(I didn't know PRs would be chained like this lol)

So, the core maintainer in deepspeed judged that we need to ask for opinions from the accelerate side, so this PR was opened.

First, through this PR, in accelerator.py > _prepare_deepspeed's SP applying part, we add `disable_in_eval` to skip Sequence Parallel during evaluation.
_(Why we need to skip it, check https://github.com/huggingface/transformers/pull/43517 )_

I hardcoded `disable_in_eval` for now, but this needs more discussion.

We might need to make it configurable for users, like adding `disable_in_eval` to `ParallelismConfig` or handling it conditionally, etc.

cc: @tohtana

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -23,7 +23,6 @@ from typing import Any
 
 import torch
 
-from . import parse_flag_from_env
 from ..commands.config.config_args import SageMakerConfig
 from ..utils import (
     DynamoBackend,
@@ -40,6 +39,7 @@ from ..utils import (
 from ..utils.constants import DEEPSPEED_MULTINODE_LAUNCHERS
 from ..utils.other import get_free_port, is_port_in_use, merge_dicts
 from ..utils.versions import compare_versions
+from . import parse_flag_from_env
 from .dataclasses import DistributedType, SageMakerDistributedType
 
 


src/accelerate/accelerator.py
```diff
@@ -2383,6 +2383,7 @@ def _prepare_deepspeed(self, *args):
                     seq_length_is_variable=sp_handler.sp_seq_length_is_variable,
                     core_attn_implementation=sp_handler.sp_attn_implementation,
                     micro_batch_size=batch_size_per_device,
+                    disable_in_eval=True,
```
SunMarc: this could become an arg in `sp_handler` but fine to first hardcode it. Also disable_in_eval is not available in the latest release of deepseed no ?

jp1924: @SunMarc Sorry for the late reply.

- this could become an arg in sp_handler but fine to first hardcode it.
  - Agreed. If issues arise later we can open a separate PR to fix it.
- Also disable_in_eval is not available in the latest release of deepspeed, no?
  - It has been merged in the deepspeed git, but it's not yet available in the latest release. If compatibility across versions is a concern, we could use inspect to check whether the sp_attn_implementation has disable_in_eval; if it does, add it, otherwise omit it.

SunMarc: > It has been merged in the deepspeed git, but it's not yet available in the latest release. If compatibility across versions is a concern, we could use inspect to check whether the sp_attn_implementation has disable_in_eval; if it does, add it, otherwise omit it.

Let's do that then ! Please update the PR and we can merge this

jp1924: @SunMarc 
OK, I added it as you requested. 
Uh... and do we need to add a separate evaluation run to the tests as well?

SunMarc: It's something that should be added on transformers side I think instead of here

SunMarc:
@bot /style

jp1924:
thank you!

PR #3935, opened 2026-02-24
Neuron integration

michaelbenayoun:
# What does this PR do?

Add support for AWS Trainium chips (Neuron Cores).

Missing integrations:

- `src/accelerate/hooks.py`

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -829,7 +829,7 @@ class PartialState:
             self.device = torch.device("cpu") if self._cpu else self.default_device
             return
         device = str(self.distributed_type).split(".")[-1].replace("MULTI_", "").lower()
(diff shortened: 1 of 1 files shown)


michaelbenayoun:
Done

BrownianNotion:
@michaelbenayoun awesome work! I'm not sure I understand the relationship between accelerate's Trainium support and optimum-neuron, would you be able to help me out?

Is the idea that optimum-neuron is a higher-level wrapper, and is the trainium drop-in replacement for TRL, while accelerate only handles parallelism and allows finer-grained control of the training loop? What made trainium sufficiently different such that it received its own library, as opposed to being part of TRL? Are the current plans to keep the two separate or will they be unified in future?

Thank you so much!

michaelbenayoun:
Trainium now has native support in PyTorch, which was not the case before when it was built on top of torch-xla. It is the main motivation behind moving from a standalone library (`optimum-neuron`) to integration inside the various libraries we have at HF.

PR #3936, opened 2026-02-24
Fix RNG state setting for HPU

michaelbenayoun:
# What does this PR do?

The RNG state for HPUs is not properly set because `rng_state` is used instead of `rng_type`. Fixing it.


IlyasMoutawwakil (approved):
lgtm!

PR #3937, opened 2026-02-24
fix: load the HPU RNG state

michaelbenayoun:
# What does this PR do?

Loading the HPU RNG state is missing. Adding it here.


IlyasMoutawwakil (approved):
lgtm!

Issue #3814, opened 2025-10-14
Deadlock when running multi tasks


RuizhePang:
### System Info

```Shell
accelerate version: 1.10.1
OS: ubuntu 20.04.6 LTS
python version: 3.11.13
numpy version: 2.3.3
torch version: 2.8.0
accelerate configuration:

ddp_kwargs = DistributedDataParallelKwargs(find_unused_parameters=True)
self.accelerator = Accelerator(
    log_with="tensorboard",
    project_dir="./tensorboard_logs",
    kwargs_handlers=[ddp_kwargs],
    gradient_accumulation_steps=self.grad_accumulation_steps,
    **accelerate_kwargs,
)
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

I use dynamic batch sampler:
```python
class DynamicBatchSampler(Sampler):
    """Extension of Sampler that will do the following:
    1.  Change the batch size (essentially number of sequences)
        in a batch to ensure that the total number of frames are less
        than a certain threshold.
    2.  Make sure the padding efficiency in the batch is high.
    3.  Shuffle batches each epoch while maintaining reproducibility.
    """

    def __init__(
        self, dataset: Dataset, frames_threshold: int, max_samples=0, random_seed=None, drop_residual: bool = False
    ):
        self.frames_threshold = frames_threshold
        self.max_samples = max_samples
        self.random_seed = random_seed
        self.epoch = 0

        indices, batches = [], []
        logger.info("Sorting dataset by frame lengths... This can be slow if duration was not precomputed")
        # for idx in tqdm(range(len(dataset)), desc="Sorting dataset... "):
        #     indices.append((idx, dataset.get_frame_len(idx)))
        for idx in range(len(dataset)):
            indices.append((idx, dataset.get_frame_len(idx)))
        indices.sort(key=lambda elem: elem[1])

        batch = []
        batch_frames = 0
        # for idx, frame_len in tqdm(
        #     indices, desc=f"Creating dynamic batches with {frames_threshold} audio frames per gpu"
        # ):
        logger.info(f"Creating dynamic batches with {frames_threshold} audio frames per gpu")
        for idx, frame_len in indices:
[...]

PR #3940, opened 2026-02-25
Fix deadlock in BatchSamplerShard with drop_last=True

Kabir08:
## What does this PR do?

Fixes issue #3814 - deadlock in BatchSamplerShard when using drop_last=True with custom batch samplers that dont implement __len__.

## Changes made:

1. Modified _iter_with_no_split() in src/accelerate/data_loader.py to handle custom batch samplers without __len__ by setting last_batch_idx = -1 and adding immediate yield logic when drop_last=True

2. Added test_batch_sampler_with_drop_last_no_len() test in tests/test_data_loader.py to verify the fix works correctly with generator-based batch samplers

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -218,6 +218,13 @@ class BatchSamplerShard(BatchSampler):
     def _iter_with_no_split(self):
         initial_data = []
         batch_to_yield = []
+        # Get the total number of batches, if available
+        try:
+            last_batch_idx = len(self.batch_sampler) - 1 if len(self.batch_sampler) > 0 else -1
+        except TypeError:
+            # If batch_sampler doesn't have __len__, we can't determine the last batch
+            # This can happen with custom batch samplers that are generators
+            last_batch_idx = -1
         for idx, batch in enumerate(self.batch_sampler):
             # We gather the initial indices in case we need to circle back at the end.
             if not self.drop_last and idx < self.num_processes:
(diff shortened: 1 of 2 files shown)


SunMarc (commented):
Does this PR your use case @RuizhePang ?

yuxinyuan:
The test case in the PR seems odd. When drop_last is set to True, the sampler should drop the uneven data. That's what "drop last" mean.

Kabir08:
Thanks for the review, My initial focus was on resolving the multi-process deadlock with generator-based samplers, but I realize the current logic and test case end up bypassing the actual drop_last behavior. I'll revise the logic so that it properly drops the incomplete batch (by breaking instead of yielding), I'll also update test_batch_sampler_with_drop_last_no_len() to properly assert that the uneven data is dropped. I'll work on this and push an updated commit shortly

Kabir08:
Hello, 
Sorry for taking so long, I had a medical emergency, I need a recommendation on which of these approach should i implement

Option A: Warn + fall back to drop_last=False behavior when length is unknown.
`# At the beginning of _iter_with_no_split
if self.drop_last and last_batch_idx < 0:
    logger.warning_once(
        "drop_last=True was requested, but the batch sampler has no __len__ (generator style). "
        "Cannot safely drop the last incomplete batch without length information → "
        "falling back to drop_last=False behavior to prevent deadlock / inconsistency."
    )
    self.drop_last = False   # or just skip the special drop logic`

Option B: Try to drop, but best-effort 
Modify the loop so that:

1. We never yield a batch if we suspect it's the tail and incomplete.
2. But since we can't know globally, we can use a heuristic:
after seeing an incomplete batch, stop yielding entirely on that process (i.e. break instead of yield).

`if should_yield_immediately and self.drop_last:
    # We know this is potentially the last batch and it's incomplete
    # → do NOT yield it → just break / skip
    if last_batch_idx >= 0:
        break
    else:
        # For true generators: once we see an incomplete batch, stop forever
        # (very conservative — may drop more than necessary, but safe)
        break   # instead of continue + yield`

This can drop more than strictly needed (if the generator keeps producing full batches after a partial one), but it never leaks partial batches, and it avoids deadlock.

Option C – Length discovery phase
Consume the iterator once without yielding to count total batches, then rewind/re-shard.


@yuxinyuan @SunMarc

yuxinyuan:
I'm confused. The `_iter_with_no_split` in `main` never touches the length of the wrapped batch sampler. Why is the exact length needed to drop last in your use case.

PR #3944, opened 2026-02-26
Fix mutable default in Megatron init and IndexError on empty ModuleList

jashshah999:
## What does this PR do?

Fixes two bugs:

1. **`utils/megatron_lm.py`** - `initialize()` used `args_defaults={}` as a default parameter. Mutable defaults are shared across calls, so any mutation persists into subsequent invocations.

2. **`utils/other.py`** - `is_repeated_block()` accessed `module[0]` without checking if the `ModuleList` was empty. An empty `ModuleList` would raise `IndexError` before the `all()` check could short-circuit.


SunMarc:
@bot /style

PR #3945, opened 2026-02-26
Prepare TP fix

michaelbenayoun:
# What does this PR do?

This ignores TP preparation when FSDP is enable and no dtensor is detected since we are not relying on them in Transformers anymore.

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1587,6 +1587,20 @@ class Accelerator:
         # Second pass: prepare schedulers
         result = [self._prepare_one(obj) if not isinstance(obj, torch.nn.Module) else obj for obj in result]
 
+        for arg in args:
+            if not isinstance(arg, torch.nn.Module):
+                continue
+            model = arg
+
+            from torch.distributed.tensor import DTensor
+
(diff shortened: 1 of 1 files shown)


SunMarc:
Please rebase this PR ! i've fixed the CI here : https://github.com/huggingface/accelerate/pull/3948

src/accelerate/accelerator.py
```diff
@@ -1601,6 +1603,14 @@ def _prepare_tp(self, *args):
...
 
                 model: torch.nn.Module = arg
+
+                if not any(isinstance(p, DTensor) for p in model.parameters()):
+                    logger.warning(
+                        "The model parameters are not sharded by DTensor, we skip the TP preparation. If you are using "
+                        "a PreTrained model it is exepected and this warning can be ignored."
+                    )
+                    return result
+
```
SunMarc: let's move this right after

```python 
# Second pass: prepare schedulers
        result = [self._prepare_one(obj) if not isinstance(obj, torch.nn.Module) else obj for obj in result]
```

michaelbenayoun: Done

PR #3955, opened 2026-03-04
Fix testing ci

SunMarc:
# What does this PR do?

This PR fixes a bunch of tests in our CI


kylesayrs:
It seems like the `_is_hf_initialized` change is required by transformers, hence requiring a dependency bump?
EDIT: seems like this is just isolated to an incompatibility with `init_empty_weights`. This function is never used by transformers source code, so it's fine.

PR #3962, opened 2026-03-06
Fix: Conditionally import `torch.distributed.algorithms.join` in `accelerator.py`

0xDELUXA:
# What does this PR do?

`torch.distributed` is not available in all PyTorch builds (for example, Windows ROCm). Importing `torch.distributed.algorithms.join.Join` unconditionally at the top level of `accelerator.py` causes an immediate crash with:
```
ModuleNotFoundError: No module named 'torch._C._distributed_c10d'; 'torch._C' is not a package
```
This makes `import accelerate` fail entirely on those builds, even for users who have no intention of using distributed training.

## Changes
- Wrapped `from torch.distributed.algorithms.join import Join` in `if torch.distributed.is_available():`
- `Join` is only used inside `join_uneven_inputs()` which is documented as multi-GPU DDP only, so this guard is safe and consistent with how the rest of the file handles optional distributed functionality

## Prior art
The same fix was applied to `huggingface/diffusers` in:
- [huggingface/diffusers#12420](https://github.com/huggingface/diffusers/pull/12420)
- [huggingface/diffusers#12425](https://github.com/huggingface/diffusers/pull/12425)

A previous fix for the same class of issue in `huggingface/transformers` was made in:
- [huggingface/transformers#40038](https://github.com/huggingface/transformers/pull/40038)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -154,7 +154,8 @@ if is_megatron_lm_available():
         megatron_lm_prepare_model_optimizer_scheduler,
     )
 
-from torch.distributed.algorithms.join import Join
+if torch.distributed.is_available():
+    from torch.distributed.algorithms.join import Join
 
 
 if is_torch_xla_available():


0xDELUXA:
@SunMarc [huggingface/transformers#44507](https://github.com/huggingface/transformers/pull/44507) was recently merged. This one should be good to go as well. Thanks!

Issue #3964, opened 2026-03-09
Security: Unsafe torch.load in FSDP2 scaler path (CWE-502)


spartan8806:
## Security Vulnerability Report

**Severity**: HIGH (CVSS 3.1: 7.8)
**CWE**: CWE-502 — Deserialization of Untrusted Data
**Reporter**: Conner Webber (conner.webber000@gmail.com)
**90-day disclosure deadline**: 2026-06-06

---

### Vulnerability

**File**: `accelerate/accelerator.py`, line ~3823 (FSDP2 scaler loading path)

**Code**:
```python
scaler_state = torch.load(input_scaler_file)
```

This call to `torch.load()` is missing `weights_only=True`, which allows arbitrary Python object deserialization via pickle.

### Context

Every OTHER `torch.load()` call in the accelerate codebase uses either `weights_only=True` or the project's own safe wrapper `accelerate.utils.other.load()`. The safe equivalent already exists at `checkpointing.py:283`:

```python
scaler_state = load(input_scaler_file)
```

This is clearly an oversight in the FSDP2 code path that was not updated when the rest of the codebase was hardened.

### Impact

A malicious `scaler.pt` file placed in a checkpoint directory (e.g., via a shared model on HuggingFace Hub, a compromised checkpoint, or supply chain attack) achieves **arbitrary code execution** when a user resumes FSDP2 training with gradient scaling enabled.

The attack surface is significant because:
1. Users routinely download and resume from shared checkpoints
2. HuggingFace Hub is a primary distribution channel for model checkpoints
3. The scaler file is loaded automatically during `load_state()` without user inspection
4. No sandboxing or validation occurs before deserialization

### CVSS 3.1: 7.8 (HIGH)

`AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H`

- **Attack Vector**: Local (malicious file must be present on disk, e.g., downloaded checkpoint)
- **Attack Complexity**: Low (no special conditions needed)
- **Privileges Required**: None
- **User Interaction**: Required (user must call `load_state` on the malicious checkpoint)
- **Confidentiality/Integrity/Availability**: All High (arbitrary code execution)

### Suggested Fix

Replace:
```python
scaler_state = torch.load(input_scaler_file)
```

With either:
```python
from accelerate.utils.other import load
scaler_state = load(input_scaler_file)
```

Or:
```python
scaler_state = torch.load(input_scaler_file, weights_only=True)
```

[...]

PR #3966, opened 2026-03-09
Security: Fix unsafe torch.load in FSDP2 scaler path (#3964)

davidequarracino:
Security fix for #3964 (CWE-502). Added weights_only=True to torch.load in FSDP2 scaler path (line 3811) to prevent arbitrary code execution. This PR was split to prioritize safety.


src/accelerate/accelerator.py
davidequarracino, line 1: FIX: Prevent unauthorized code execution (CWE-502).
If new models fail to load, it is likely due to non-standard data types.
Instead of disabling security, register the required types as "safe" in PyTorch.
scaler_state = torch.load(input_scaler_file, weights_only=True)

PR #3973, opened 2026-03-13
Torchao offload

SunMarc:
# What does this PR do?

We need this for cpu offload with torchao since they keep changing their param class


SunMarc:
@bot /style

PR #3985, opened 2026-03-20
[FSDP2] Cast model to uniform dtype before fully_shard to fix mixed-dtype AssertionError

roycho96:
# What does this PR do?

When `mixed_precision` is enabled, casts model parameters to uniform dtype before `fully_shard()` to prevent `_init_mp_dtypes()` AssertionError.

## Problem

FSDP2's `_init_mp_dtypes()` requires uniform `orig_dtype` across all trainable parameters in a param group. With mixed dtypes, the first forward call crashes:
```
AssertionError: FSDP expects uniform original parameter dtype but got {torch.bfloat16, torch.float32}
```

FSDP2's `fsdp2_prepare_model()` currently passes the mixed-dtype model directly to `fully_shard()` without normalizing dtypes.

## Fix

Cast all parameters to the mixed precision `param_dtype` before `fully_shard()`, after `model_has_params4bit` detection. Params4bit models are skipped to avoid destroying quantized weights.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -666,20 +666,19 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
             model_has_params4bit = True
             break
 
-    # FSDP2 requires uniform original parameter dtype within each param group.
-    # Norm/embedding weights stored in fp32 while the rest is bf16/fp16, causing _init_mp_dtypes() AssertionError.
+    # FSDP2 requires uniform orig_dtype among trainable params in each group.
+    # Upcast to fp32 master weights; MixedPrecisionPolicy.param_dtype handles compute cast.
     if accelerator.mixed_precision != "no" and not model_has_params4bit:
-        mp_policy = fsdp2_plugin.mixed_precision_policy
-        if mp_policy is not None and getattr(mp_policy, "param_dtype", None) is not None:
-            target_dtype = mp_policy.param_dtype
-        elif accelerator.mixed_precision == "bf16":
-            target_dtype = torch.bfloat16
-        elif accelerator.mixed_precision == "fp16":
-            target_dtype = torch.float16
-        else:
-            target_dtype = None
-        if target_dtype is not None:
-            model.to(target_dtype)
+        upcasted_params = []
+        for name, param in model.named_parameters():
+            if param.requires_grad and param.dtype != torch.float32:
+                upcasted_params.append(name)
+                param.data = param.data.to(torch.float32)
+        if accelerator.is_main_process and upcasted_params:
+            warnings.warn(
+                "FSDP upcast of low precision parameters to fp32 (since mixed_precision != 'no') may affect the precision of model checkpoints. "
+                f"This effects {len(upcasted_params)} parameters: {upcasted_params}..."
+            )
 
     if fsdp2_plugin.cpu_ram_efficient_loading and not model_has_params4bit:
         # Context: `fully_shard` moves the model to GPU if it was on CPU, however it can also be on `meta` and then it stays there even after `fully_shard`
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -666,6 +666,21 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
...
+        if mp_policy is not None and getattr(mp_policy, "param_dtype", None) is not None:
+            target_dtype = mp_policy.param_dtype
+        elif accelerator.mixed_precision == "bf16":
+            target_dtype = torch.bfloat16
+        elif accelerator.mixed_precision == "fp16":
+            target_dtype = torch.float16
+        else:
+            target_dtype = None
+        if target_dtype is not None:
+            model.to(target_dtype)
```
SunMarc: hmmmm for mixed precision, do fsdp need the weights in fp16/bf16 and will keep a trace of the weights in fp32 to actually perform mixed precision ? or it is the other way around ? Can you share a reproducer also ?

SunMarc (commented):
Left a comment !

src/accelerate/utils/fsdp_utils.py
roycho96: > hmmmm for mixed precision, do fsdp need the weights in fp16/bf16 and will keep a trace of the weights in fp32 to actually perform mixed precision ? or it is the other way around ? Can you share a reproducer also ?

you're right. proper mixed precision keeps fp32 master weights and casts to bf16/fp16 only during forward/backward compute. my current fix downcasts master weights, which collapses mixed precision into pure bf16/fp16 training and loses optimizer update precision.

roycho96: here is a reproducer

repro.py
```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer, TrainingArguments, Trainer, DataCollatorForLanguageModeling
from datasets import Dataset

    "HuggingFaceTB/SmolLM2-135M-Instruct", torch_dtype=torch.bfloat16
)
tokenizer = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM2-135M-Instruct")
tokenizer.pad_token = tokenizer.eos_token

# Inspect dtypes before forcing any changes
import os
if int(os.environ.get("RANK", "0")) == 0:
    dtype_counts = {}
    for name, p in model.named_parameters():
        dtype_counts[p.dtype] = dtype_counts.get(p.dtype, 0) + 1
    print(f"[rank0] param dtype distribution (initial): {dtype_counts}")

# Force norm layers to fp32 to reliably trigger the mixed-dtype FSDP2 issue
for name, module in model.named_modules():
    if "norm" in name.lower():
        module.to(torch.float32)

if int(os.environ.get("RANK", "0")) == 0:
    dtype_counts = {}
    for name, p in model.named_parameters():
        dtype_counts[p.dtype] = dtype_counts.get(p.dtype, 0) + 1
    print(f"[rank0] param dtype distribution (after forcing norms fp32): {dtype_counts}")

dataset = Dataset.from_list([{"text": f"Hello {i}"} for i in range(32)])
dataset = dataset.map(
    lambda x: tokenizer(x["text"], truncation=True, max_length=32, padding="max_length"),
    batched=True,
    remove_columns=["text"],
)

trainer = Trainer(
    model=model,
    args=TrainingArguments(
        output_dir="/tmp/repro",
        bf16=True,
        num_train_epochs=1,
        per_device_train_batch_size=2,
        logging_steps=1,
        save_strategy="no",
        report_to="none",
    ),
    train_dataset=dataset,
    data_collator=DataCollatorForLanguageModeling(tokenizer, mlm=False),
)
trainer.train()
```

fsdp2_config.yaml
```yaml
distributed_type: FSDP
fsdp_config:
  fsdp_version: 2
  fsdp_transformer_layer_cls_to_wrap: LlamaDecoderLayer
mixed_precision: bf16
num_processes: 2
```

accelerate launch --config_file fsdp2_config.yaml repro.py

roycho96:
@SunMarc 
I moved the existing (but no-op) fp32 upcasting logic to before fully_shard so it actually takes effect. This keeps fp32 master weights, letting MixedPrecisionPolicy.param_dtype handle the bf16/fp16 cast during compute.

Also, I added model_has_params4bit guard to skip the upcast for QLoRA.

SunMarc:
> I moved the existing (but no-op) fp32 upcasting logic to before fully_shard so it actually takes effect. This keeps fp32 master weights, letting MixedPrecisionPolicy.param_dtype handle the bf16/fp16 cast during compute.

WDYM by no-op (it worked if the whole model was in bf16 no )? Also btw, i found this when looking for why they decided to have the upcast so late in the preparation of the model https://github.com/huggingface/accelerate/pull/2674#discussion_r1630737765 but it shouldn't be an issue for fsdpv2 to move this I think !

SunMarc (approved):
Thanks, this looks better !

roycho96:
> > I moved the existing (but no-op) fp32 upcasting logic to before fully_shard so it actually takes effect. This keeps fp32 master weights, letting MixedPrecisionPolicy.param_dtype handle the bf16/fp16 cast during compute.
> 
> WDYM by no-op (it worked if the whole model was in bf16 no )? Also btw, i found this when looking for why they decided to have the upcast so late in the preparation of the model [#2674 (comment)](https://github.com/huggingface/accelerate/pull/2674#discussion_r1630737765) but it shouldn't be an issue for fsdpv2 to move this I think !

I mean upcast actually works only in the FSDP1 path (accelerator.py:2020 uses param.data = param.data.to(torch.float32)). FSDP2 version was reassigns the local variable, so nothing ever got upcasted regardless of input dtype.

Btw, pushed a ruff fix.

Issue #3983, opened 2026-03-19
[FSDP2] auto-exclude incompatible Params4bit from fully_shard to prevent silent QLoRA corruption


roycho96:
## Summary

When FSDP2 is used with a QLoRA model using default `bnb_4bit_quant_storage` (uint8), `fully_shard()` either crashes or silently produces incorrect training results (`loss=0.0`, `grad_norm=nan`). There is no warning or error guiding users to the correct configuration.

`fsdp2_prepare_model()` already detects `Params4bit` and skips meta device movement. This proposal extends that pattern to also exclude incompatible `Params4bit` from sharding via `ignored_params`.

## Library Info

- `torch`: 2.8.0+cu128
- `bitsandbytes`: 0.48.2
- `transformers`: 4.57.2

## Context

Most QLoRA users don't set `bnb_4bit_quant_storage` explicitly, which defaults to uint8. The [official FSDP-QLoRA path](https://huggingface.co/docs/bitsandbytes/main/en/fsdp_qlora) requires setting it to a floating dtype (bf16/fp16/fp32) to make `Params4bit` compatible with FSDP sharding.

When a user combines FSDP2 + QLoRA with the default uint8 storage, two things break silently:

### 1. `nn.Parameter` creation crash

`_fsdp_param.py:_init_sharded_param()` creates `nn.Parameter(sharded_data)` with default `requires_grad=True`. For uint8 tensors:
```
RuntimeError: only Tensors of floating point dtype can require gradients
```

### 2. quant_state destruction (if crash is bypassed)

`fully_shard()` chunks/pads/converts `Params4bit` into DTensors, destroying the `quant_state` mapping. The model produces all-zero outputs:
```
{'loss': 0.0, 'grad_norm': nan, ...}
```

## Existing Params4bit handling in `fsdp2_prepare_model()`

The function already detects and special-cases `Params4bit`:
```python
model_has_params4bit = False
for name, param in model.named_parameters():
    if param.__class__.__name__ == "Params4bit":
        model_has_params4bit = True
        break

# Used to skip meta device movement (line 669) and buffer re-registration (line 709)
if fsdp2_plugin.cpu_ram_efficient_loading and not model_has_params4bit:
    ...
```

However, the `fully_shard()` call itself has no such guard that incompatible `Params4bit` are still passed to sharding.

## Proposed solution

[...]

PR #3987, opened 2026-03-21
[FSDP2] Auto-exclude non-floating frozen Params4bit from fully_shard to prevent QLoRA crash

roycho96:
# What does this PR do?

Auto-excludes incompatible BnB `Params4bit` from FSDP2 sharding via `ignored_params` to prevent QLoRA crash.

Fixes [#3983](https://github.com/huggingface/accelerate/issues/3983)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc


roycho96:
> Do we get the same results as with `bnb_4bit_quant_storage` being a float ? Did you test on a e2e example to compare both ? Thanks a lot !

@SunMarc Thanks for the review!

I ran an e2e comparison on single GPU (Qwen3-0.6B + QLoRA + SFT, same seed/hyperparams):

| Step | uint8 (this PR) | bf16 quant_storage (official) |
|------|----------------|-------------------------------|
| 1 | 1.2914 | 1.2914 |
| 2 | 1.2571 | 1.2516 |
| 3 | 1.2310 | 1.2281 |
| 4 | 1.1131 | 1.1202 |
| 5 | 1.0541 | 1.0563 |
| 6 | 0.9757 | 0.9793 |
| 7 | 0.9656 | 0.9687 |
| 8 | 0.9125 | 0.9066 |
| **avg** | **1.1001** | **1.1003** |

Loss curves are nearly identical (< 0.6% difference at any step). 
The minor variations are expected from floating-point ordering 
differences in the two quant_storage representations. 
The dequantized values are equivalent.

roycho96:
> Another question I have is that if we put those into ignored param, will fsdp stil manage to shard correctly the model across multiple gpus since most of params are Params4bit ?

You're right. With most parameters being Params4bit and excluded  from sharding, the base weights are replicated on each GPU. FSDP  still shards the trainable LoRA parameters and gradients, but the memory benefit from weight sharding is limited in this path.
The primary goal of this PR is crash prevention. Currently, users with default QLoRA config (`bnb_4bit_quant_storage` unset) get a crash on FSDP2. I think this is a common QLoRA setup.
For users who need full base weight sharding (e.g. 70B+ models), the path with `bnb_4bit_quant_storage=bf16` remains the right choice and is unaffected by this PR.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -657,14 +657,31 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
...
+            if (not p.requires_grad) and (not p.is_floating_point()) and (not p.is_complex())
+        }
+        if incompatible_params4bit:
+            ignored = set(fsdp2_kwargs.get("ignored_params", set()))
+            fsdp2_kwargs["ignored_params"] = ignored | incompatible_params4bit
+            if accelerator.is_main_process:
+                logger.info(
+                    f"Found {len(incompatible_params4bit)} non-floating frozen Params4bit. "
+                    "Excluding from FSDP2 sharding to prevent quant_state corruption."
+                )
```
SunMarc: then maybe we should warn the user to set the type for the storage to get the right memory saving

roycho96: Good point, updated the message to guide users toward the floating quant_storage path.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -657,14 +657,33 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
...
     for name, param in model.named_parameters():
         # this is a temporary fix whereby loading models with bnb params cannot be moved from
         # GPU to a meta device due with FSDP2 because torch operations don't return the original class type
         # bypassing the move to meta will still cause the VRAM spike, but at least it still will load
         if param.__class__.__name__ == "Params4bit":
-            model_has_params4bit = True
-            break
+            params4bit.append(param)
+
+    model_has_params4bit = len(params4bit) > 0
```
SunMarc: let's not append all param in params4bit. let's keep model_has_params4bit as we did before

tests/fsdp/test_fsdp.py
```diff
@@ -471,6 +471,93 @@ def test_param_mapping_error_handling(self):
 
         AcceleratorState._reset_state(True)
 
+    def test_fsdp2_ignored_params_non_floating_params4bit(self):
```
SunMarc: let's remove those tests, it's fine to not add for this as the tests are a bit too much over-engineered

roycho96: removed

src/accelerate/utils/fsdp_utils.py
roycho96: > let's not append all param in params4bit. let's keep model_has_params4bit as we did before

Done. Reverted to model_has_params4bit flag with early detection, and moved the incompatible filtering into the same loop to avoid a second pass over parameters.

SunMarc: thanks !

PR #3989, opened 2026-03-24
Save model-only feature in `save_state`

### LINKED ISSUE #3990: [Feature] Save model-only feature in `save_state`
Take advantage of `save_state` robustness against specialized use-cases (`FSDP`, `DeepSpeed`, `Multi-GPU`, etc.)  and implement save model-only feature. 

Implementation as well as details can be found in  https://github.com/huggingface/accelerate/pull/3989 .

3manifold:
# What does this PR do?

Add   _save-model-only_ feature to `save_state`.

By leveraging the existing `save_state` implementation -which already correctly handles complex sub-cases like `FSDP`, `DeepSpeed`, and` Multi-GPU`- we can extend it to save only the model weights. The current `save_model` function does not appear to handle these cases (and more) efficiently; for instance, using `save_model` ([link](https://github.com/huggingface/accelerate/blob/0d8ca1a44f85a35df62499eb2174608e0f3473f8/src/accelerate/accelerator.py#L3423)) in anything other than _single-process_  vanilla runs can cause bugs.

Additionally, it feels more natural to have a unified function for both state and model use cases, rather than explicitly passing `model: torch.nn.Module` as required by the current `save_model` implementation.

Let me know if you agree with this approach. If so, I will proceed with adding the documentation and can discuss the future of the existing `save_model` function.

resolves #3990

## Benchmark 
I conducted experiments where the `model.safetensors` is `12G` and `optimizer.bin` is `22G` and the `save_state` time was reduced by `70.44%`  (save all: `393.67 sec` against model_only: `116.35 sec`).

 

## Who can review?

 
- Core parts of the library: @BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -3639,6 +3639,8 @@ class Accelerator:
         os.makedirs(output_dir, exist_ok=True)
         logger.info(f"Saving current state to {output_dir}")
 
+        save_model_only = save_model_func_kwargs.pop("save_model_only", False)
+
         if self.distributed_type == DistributedType.XLA:
             # Finish running the previous step before checkpointing
             xm.mark_step()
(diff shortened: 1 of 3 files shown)


3manifold:
Is this feature still relevant?

PR #4008, opened 2026-04-14
test: doc build with bucket push

rtrompier:
Test PR to validate bucket-based doc upload workflow. Will be closed after validation.


rtrompier:
Works as expected

PR #4013, opened 2026-04-19
Add `minimum_batch_size` parameter and OOM retry logging to `find_executable_batch_size`

GitGlimpse895:
# What does this PR do?

Extends `find_executable_batch_size` in `utils/memory.py` with two
practical improvements for production training runs:

1. **`minimum_batch_size` parameter** — the current implementation only
   stops when `batch_size` hits exactly `0`, which means a user who knows
   their model cannot meaningfully train below e.g. batch size 8 has no
   way to express that floor. The loop also risks an infinite cycle if a
   custom `reduce_batch_size_fn` ever stalls above zero. The new parameter
   (default `1`, fully backward-compatible) raises a clear `RuntimeError`
   the moment the batch size would cross the floor, instead of grinding
   silently to zero.

2. **OOM retry logging** — every retry currently happens in total silence,
   making distributed jobs appear frozen and giving operators no signal
   about progress. A `logging.warning` is emitted on each OOM reduction,
   visible via Python's standard logging machinery with no overhead on the
   success path.

Both changes are covered by new unit tests in `tests/test_memory_utils.py`.

Fixes # (issue)

## Before submitting
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr), Pull Request section?

## Who can review?

@BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/utils/memory.py
+++ b/src/accelerate/utils/memory.py
@@ -20,6 +20,7 @@ A collection of utilities for ensuring that training can always occur. Heavily i
 import functools
 import gc
 import inspect
+import logging
 from typing import Optional
 
 import torch
@@ -116,9 +117,13 @@ def should_reduce_batch_size(exception: Exception) -> bool:
     return False
 
 
+logger = logging.getLogger(__name__)
+
+
 def find_executable_batch_size(
     function: Optional[callable] = None,
     starting_batch_size: int = 128,
+    minimum_batch_size: int = 1,
     reduce_batch_size_fn: Optional[callable] = None,
 ):
     """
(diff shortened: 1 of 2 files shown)


SunMarc (commented):
What would be the use case of that ? If you use this, it will just try to find the min value of the batch size that works.

GitGlimpse895:
Hi @SunMarc, thanks for taking the time to review this!

The logic behind the minimum_batch_size was to allow users to set a "floor" for cases where training isn't statistically meaningful below a certain threshold, and the logging was intended to give some visibility during long distributed runs.

However, I completely understand the preference for keeping the API lean. Since the current implementation already handles finding a working batch size effectively, and this is more of a "nice-to-have" improvement rather than a critical fix, I’ll go ahead and close this one out to avoid unnecessary maintenance overhead.

Thanks again for the feedback!

PR #4015, opened 2026-04-22
Fsdp2 fully_shard embedding and norm 

SunMarc:
# What does this PR do?

This PR updates how we fully shard the model. We create two new units:
- one with the embedding (with reshard_after_forward=value passed by the user which is usually True)
- one with the final norm + output_embedding (with reshard_after_forward=False hardcoded because we need to gather again for the backward right after)


qgallouedec:
lgtm! for context, it was flagged in https://github.com/huggingface/trl/pull/5575 where we measured a significant slowdown because of `reshard_after_forward=True` being true by default

PR #4019, opened 2026-04-24
fix(checkpointing): save/restore SeedableRandomSampler for map-style datasets

Anai-Guo:
## Problem

Fixes #3996.

When using `use_seedable_sampler=True` with a **map-style** (non-iterable) dataset, `save_state()` / `load_state()` silently drops the `SeedableRandomSampler.epoch` counter. On resume the counter resets to 0, so the sampler re-derives the same seed sequence (`initial_seed + 0`, `initial_seed + 1`, …) that was already used before the checkpoint — the model trains on the exact same shuffle order it already saw.

## Root cause

`save_state()` and `load_state()` both gate sampler persistence on:

```python
if isinstance(dataloader.dataset, IterableDatasetShard):
    ...
```

`SeedableRandomSampler` is attached to **both** iterable and map-style dataloaders (when `use_seedable_sampler=True`), but the guard prevents it from being saved for the map-style case.

## Fix

Remove the `IterableDatasetShard` guard and persist any `SeedableRandomSampler`, regardless of dataset type. On the load side, add an `input_sampler_file.exists()` guard for backwards-compat with older checkpoints that do not contain a sampler file.

```diff
-        from .data_loader import IterableDatasetShard, SeedableRandomSampler
-
-        if isinstance(dataloader.dataset, IterableDatasetShard):
-            sampler = dataloader.get_sampler()
-            if isinstance(sampler, SeedableRandomSampler):
-                save(sampler, output_sampler_file, ...)
+        from .data_loader import SeedableRandomSampler
+
+        sampler = dataloader.get_sampler()
+        if isinstance(sampler, SeedableRandomSampler):
+            save(sampler, output_sampler_file, ...)
```

The fix is identical in shape for both `save_state` and `load_state`; the load path additionally checks `input_sampler_file.exists()` to stay compatible with checkpoints produced before this fix.

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -131,13 +131,12 @@ def save_accelerator_state(
     for i, dataloader in enumerate(dataloaders):
         sampler_name = f"{SAMPLER_NAME}.bin" if i == 0 else f"{SAMPLER_NAME}_{i}.bin"
         output_sampler_file = output_dir.joinpath(sampler_name)
-        # Only save if we have our custom sampler
-        from .data_loader import IterableDatasetShard, SeedableRandomSampler
+        # Save SeedableRandomSampler state for both IterableDataset and map-style datasets
+        from .data_loader import SeedableRandomSampler
 
-        if isinstance(dataloader.dataset, IterableDatasetShard):
-            sampler = dataloader.get_sampler()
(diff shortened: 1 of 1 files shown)


Anai-Guo:
Still relevant — happy to address any review feedback. Please keep open.

SunMarc:
It doesn't fix the issue you've linked

PR #4022, opened 2026-04-28
Fix region compilation fsdpv2

SunMarc:
# What does this PR do?

This PR fixes region compilation with FSDPv2. `compile_regions` rebuilds each repeated block as `torch.compile(submodule)` → returns an `OptimizedModule` wrapper. That wrapper's __call__ bypasses `nn.Module._call_impl`, so the forward/pre hooks `fully_shard` adds afterwards never fire → no per-layer all-gather/reshard → no overlap. Basically the same thing we did for deepspeed.

Compiling the whole model doesn't lead to speed-up, maybe we should do this as the default path ?

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -129,7 +129,7 @@ from .utils.constants import (
     SCALER_NAME,
 )
 from .utils.modeling import get_state_dict_offloaded_model
-from .utils.other import compile_regions, compile_regions_deepspeed, is_compiled_module
+from .utils.other import compile_regions, compile_regions_deepspeed, compile_regions_fsdp2, is_compiled_module
 
 
 if is_deepspeed_available():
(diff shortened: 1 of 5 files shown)


AmineDiro:
fix `trl.SFTTrainer` slower compile  🎉

**Setup** (same as before): Qwen3-30B-A3B (MoE, 128 experts, 48 layers) · 2×8 H100 SXM 80GB · FSDP2 DP=16 · seq_len=16384 · SFTTrainer w/ grad ckpt, bf16, packing=True.

**Required yaml addition** (the only diff vs my slow-path repro):

```yaml
dynamo_config:
  backend: inductor
  use_fullgraph: true
  use_regional_compilation: true
```

**Results:**

| Setup | MFU | ms/step | vs eager |
|---|---|---|---|
| raw `fully_shard` + eager | 25.0% | 3,888 | — |
| raw `fully_shard` + per-layer compile | 32.1% | 3,031 | 1.28× faster |
| accelerate `fsdp2_prepare_model` + eager | 23.4% | 4,160 | — |
| accelerate `fsdp2_prepare_model` + per-layer compile **(BEFORE this PR)** | 9.8% | 9,900 | 2.4× slower |
| accelerate `fsdp2_prepare_model` + per-layer compile + `use_regional_compilation=true` **(THIS PR)** | ~32% | ~3,000 | 1.28× faster — matches raw fast path |

`mfu_window` samples on this PR over 3 logging steps: **32.55 / 31.27 / 31.81 %**. Regression fully closed.

**Repro script** (gist): https://gist.github.com/AmineDiro/9fd331214626b60e4d421264637b3828 — same as my [original repro](https://gist.github.com/AmineDiro/2457fbee70662d584a116cc3ca80dd07), only the yaml gains the `dynamo_config` block.

PR #4045, opened 2026-05-21
feat: add DispatchDataLoader for custom iterable dataloaders

### LINKED ISSUE #2975: Barebones dataloader to allow for any type of iterable object to be used. Should just handle device placement

KrishVenky:
Closes #2975

Adds `DispatchDataLoader`, a lightweight wrapper that makes any Python iterable usable with `Accelerator.prepare()` without requiring it to be a `torch.utils.data.DataLoader`. Also adds `custom_classes` to `DataLoaderConfiguration` so the Accelerator can auto-detect and wrap user-defined iterable types.

## What does this PR do?

Implements the feature requested in #2975 ("barebones dataloader for any iterable"):

- **`DispatchDataLoader`** (`data_loader.py`): wraps any iterable, calls `__iter__` on each pass, and applies automatic device placement via `send_to_device`. Exported from the top-level `accelerate` package.
- **`DataLoaderConfiguration.custom_classes`** (`utils/dataclasses.py`): a tuple of class types. When set, `Accelerator.prepare()` detects instances of those classes and wraps them in `DispatchDataLoader` automatically.
- **`_prepare_one`** (`accelerator.py`): recognises `DispatchDataLoader` directly, and wraps `custom_classes` instances.

**Usage:**
```python
from accelerate import Accelerator, DispatchDataLoader
from accelerate.utils import DataLoaderConfiguration

class MyDataSource:
    def __iter__(self):
        for i in range(100):
            yield {"input_ids": torch.tensor([i])}

config = DataLoaderConfiguration(custom_classes=(MyDataSource,))
accelerator = Accelerator(dataloader_config=config)
loader = accelerator.prepare(MyDataSource())  # returns DispatchDataLoader

### DIFF
--- a/src/accelerate/__init__.py
+++ b/src/accelerate/__init__.py
@@ -23,7 +23,7 @@ from .big_modeling import (
     init_on_device,
     load_checkpoint_and_dispatch,
 )
-from .data_loader import skip_first_batches
+from .data_loader import DispatchDataLoader, skip_first_batches
 from .inference import prepare_pippy
 from .launchers import debug_launcher, notebook_launcher
 from .parallelism_config import ParallelismConfig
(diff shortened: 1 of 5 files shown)


SunMarc:
Not needed in the end but thanks !

PR #4046, opened 2026-05-22
Fix MLflowTracker.store_init_configuration mutating the caller's config dict

ATOM00blue:
## What does this PR do?

`MLflowTracker.store_init_configuration` drops config values that exceed MLflow's
`MAX_PARAM_VAL_LENGTH`, but it did so by deleting them from the input dict in place:

```python
for name, value in list(values.items()):
    if len(str(value)) > mlflow.utils.validation.MAX_PARAM_VAL_LENGTH:
        logger.warning_once(...)
        del values[name]   # mutates the caller's dict
```

`Accelerator.init_trackers` passes the **same** `config` object to every tracker:

```python
if config is not None:
    for tracker in self.trackers:
        tracker.store_init_configuration(config)
```

So when MLflow drops an over-long value, that key is permanently removed from the
user's `config` dict and is therefore missing from every other tracker (TensorBoard,
WandB, etc.) that runs after MLflow, as well as from any later use of the dict by the
user. This is silent, order-dependent data loss.

### Fix

Build a filtered list of in-range items instead of mutating the input, which is the
same approach already used in `MLflowTracker.log` (it builds a separate `metrics` dict).
The warning and the dropping of over-long values from what gets logged to MLflow are
both preserved; only the in-place mutation of the caller's dict is removed.

### Reproduction (before this PR)

```python
import tempfile, mlflow
from accelerate.tracking import MLflowTracker

tracker = MLflowTracker(experiment_name="repro", logging_dir=tempfile.mkdtemp())
tracker.start()
config = {"learning_rate": 0.001, "long_note": "x" * (mlflow.utils.validation.MAX_PARAM_VAL_LENGTH + 1)}
tracker.store_init_configuration(config)
print(config)  # -> {'learning_rate': 0.001}  ("long_note" was deleted from the caller's dict)
```

After this PR, `config` still contains `long_note`; only the value sent to MLflow is dropped.

## Before submitting

A regression test (`MLflowTrackingTest::test_store_init_configuration_does_not_mutate_input`)
is added; it fails on `main` and passes with this change.

## Who can review?

@BenjaminBossan @SunMarc

---

Developed with AI coding assistance; reviewed and submitted by the author.


SunMarc (approved):
LGTM

SunMarc:
@bot /style

PR #4048, opened 2026-05-25
test: add regression test for no_split_module_classes accepting set type

UFO0506:
## What this PR does

Adds a regression test for `get_balanced_memory` and `infer_auto_device_map`
when `no_split_module_classes` is passed as a `set` instead of a `list` or `tuple`.

## Background

In accelerate <= 1.5.0, passing a `set` caused:
TypeError: unhashable type: 'set'
File "accelerate/utils/modeling.py", in get_balanced_memory
if set(no_split_children.keys()) == set(no_split_module_classes):

This was silently fixed as part of #2345, but no regression test was added.
This PR adds explicit coverage so the fix cannot regress undetected.

## Testing

```bash
pytest tests/test_modeling_utils.py::ModelingUtilsTester::test_get_balanced_memory_no_split_module_classes_set -v
# 1 passed
```

### DIFF
--- a/tests/test_modeling_utils.py
+++ b/tests/test_modeling_utils.py
@@ -926,6 +926,7 @@ class ModelingUtilsTester(unittest.TestCase):
 
         # Also verify infer_auto_device_map handles set input
         from accelerate import infer_auto_device_map
+
         device_map = infer_auto_device_map(
             model,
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
LGTM

SunMarc:
@bot /style

PR #4051, opened 2026-05-27
DOC: document missing parameters in load_accelerator_state, find_executable_batch_size, and send_to_device

kratos0718:
## What this fixes

Three public API functions had parameters present in their signatures but missing from their docstrings.

### `load_accelerator_state()` — missing `dataloaders`
`dataloaders` is a **required** parameter (no default value) but was completely absent from the `Args` section, unlike the equivalent `save_accelerator_state()` which documents it correctly.

### `find_executable_batch_size()` — missing `reduce_batch_size_fn`
Allows users to provide a custom batch size reduction strategy instead of the default multiply-by-0.9 behavior on OOM. Undocumented means users can't discover or use this customization point.

### `send_to_device()` — missing `non_blocking` and `skip_keys`
- `non_blocking`: enables async GPU transfers, useful for overlapping data movement with computation
- `skip_keys`: excludes specific dict keys from device transfer

## Files changed
- `src/accelerate/checkpointing.py`
- `src/accelerate/utils/memory.py`
- `src/accelerate/utils/operations.py`

### DIFF
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -204,6 +204,8 @@ def load_accelerator_state(
             A list of optimizer instances
         schedulers (`List[torch.optim.lr_scheduler._LRScheduler]`):
             A list of learning rate schedulers
+        dataloaders (`List[torch.utils.data.DataLoader]`):
+            A list of dataloader instances used in your program
         process_index (`int`):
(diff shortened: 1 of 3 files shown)


SunMarc (approved):
Thanks !

PR #4052, opened 2026-05-27
Fix implicit padding in `split_between_processes` when `apply_padding=False` and `num_samples < num_processes`

### LINKED ISSUE #4053: [Bug]  Implicit padding when splitting input between processes while padding flag  is disabled
### System Info

```Shell
- `accelerate 1.13.0`
- `Python 3.13.0`
- `torch 2.10.0`
- OS: `Ubuntu 24.04.4 LTS`
```

### Information

- [x] The official example scripts 

### Tasks

- [x] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`) 

### Reproduction

Modified example from `examples/inference/distributed/distributed_image_generation.py` using 3 dataset prompts and 2 GPUs.
 

```bash
root@amachine:/# CUDA_VISIBLE_DEVICES=0,1 accelerate launch infer.py
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to a value of `2`
                More than one GPU was found, enabling multi-GPU training.
                If this was unintended please pass in `--num_processes=1`.
        `--num_machines` was set to a value of `1`
        `--mixed_precision` was set to a value of `'no'`
        `--dynamo_backend` was set to a value of `'no'`
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
  0%|                                                                                                                                                                                                                                                                   | 0/2 [00:00<?, ?it/s]
1716 ['bond']
1716 ['molecule']
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00, 14027.77it/s]
[rank0]:[W527 14:25:00.934755209 ProcessGroupNCCL.cpp:1476] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
  0%|                                                                                                                                                                                                                                                                   | 0/2 [00:00<?, ?it/s]
1717 ['element']
1717 ['molecule']
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00, 14926.35it/s]
```

```python
import os
import time

import fire
import torch
from diffusers import DiffusionPipeline
from tqdm import tqdm

from accelerate import PartialState
from accelerate.utils import gather_object


START_TIME = time.strftime("%Y%m%d_%H%M%S")
DTYPE_MAP = {"fp32": torch.float32, "fp16": torch.float16, "bf16": torch.bfloat16}


def get_batches(items, batch_size):
    num_batches = (len(items) + batch_size - 1) // batch_size
    batches = []

    for i in range(num_batches):
        start_index = i * batch_size
        end_index = min((i + 1) * batch_size, len(items))
        batch = items[start_index:end_index]
        batches.append(batch)

    return batches


def main(
    # ckpt_id = "segmind/tiny-sd",
    # save_dir: str = "./evaluation/examples",
    # seed: int = 1,
    # batch_size: int = 4,
    # num_inference_steps: int = 20,
    # guidance_scale: float = 4.5,
    # dtype: str = "fp16",
    # low_mem: bool = False,
):
    # pipeline = DiffusionPipeline.from_pretrained(ckpt_id, torch_dtype=DTYPE_MAP[dtype])

    # save_dir = save_dir + f"_{START_TIME}"

    distributed_state = PartialState()
    parti_prompts = load_dataset("nateraw/parti-prompts", split="train")
    single_prompt = parti_prompts["Prompt"][:3]
    data_loader = get_batches(items=single_prompt, batch_size=distributed_state.num_processes)


    # if low_mem:
    #     pipeline.enable_model_cpu_offload(gpu_id=distributed_state.device.index)
    # else:
    #     pipeline = pipeline.to(distributed_state.device)

    # if distributed_state.is_main_process:
    #     if not os.path.exists(save_dir):
    #         os.makedirs(save_dir)
    #         print(f"Directory '{save_dir}' created successfully.")
    #     else:
    #         print(f"Directory '{save_dir}' already exists.")

    count = 0
    for _, prompts_raw in tqdm(enumerate(data_loader), total=len(data_loader)):
        input_prompts = []

        with distributed_state.split_between_processes(prompts_raw) as prompts: 
        	print(os.getpid(), prompts)
            # generator = torch.manual_seed(seed)
            # images = pipeline(
            #     prompts, num_inference_steps=num_inference_steps, guidance_scale=guidance_scale, generator=generator
            # ).images
            # input_prompts.extend(prompts)

        # distributed_state.wait_for_everyone()

        # images = gather_object(images)
        # input_prompts = gather_object(input_prompts)

        # if distributed_state.is_main_process:
        #     for image, prompt in zip(images, input_prompts):
        #         count += 1
        #         temp_dir = os.path.join(save_dir, f"example_{count}")

        #         os.makedirs(temp_dir)
        #         prompt = "_".join(prompt.split())
        #         image.save(f"image_{prompt}.png")

    # if distributed_state.is_main_process:
    #     print(f">>> Image Generation Finished. Saved in {save_dir}")


if __name__ == "__main__":
    fire.Fire(main)
```

### Expected behavior

Process 1716 should print `['bond']`, ` ['molecule']` and process 1717 should print  `['element']`.

3manifold:
# What does this PR do?

Function `split_between_processes` applies implicit  padding even if     `apply_padding=False`. This results to redundant calculations during distributed inference since the same inference (i.e. prompt)  is called twice etc. Apart from wasting resources, this can cause conflicts among competing processes, for instance, disk I/O. 

The same bug is duplicated across two code branches  of the same function. Both stem from the identical incorrect assumption — *"a process always receives at least one element"* — and both produce identical wrong behaviour (silent padding when `apply_padding=False`). The only difference is syntax:

- Bug 1 (list/Tensor): guards with `result = inputs[-1:]`
- Bug 2 (Dataset): guards with `start_index = len(inputs) - 1`

###  Example  with 1 sample and 2 GPUs
 

| Process | start_index | end_index | Normal slice |
|---------|-------------|-----------|--------------|
| 0       | 0           | 1         | `inputs[0:1]` → `[sample_0]` ✓ |
| 1       | 1           | 1         | `start_index >= len(inputs)` triggers! → `inputs[-1:]` = `[sample_0]` ✗ |

**Process 1's `start_index` (1) equals `len(inputs)` (1), triggering the guard.** Instead of returning an empty slice, it silently returns the last element — **regardless of `apply_padding`**.

```python 
def _split_values(inputs, start_index, end_index):
    if isinstance(inputs, (list, tuple, torch.Tensor)):
        if start_index >= len(inputs):      
            result = inputs[-1:]              # ← always returns last element
        else:
            result = inputs[start_index:end_index]
        if apply_padding:
            ...  
```

resolves  https://github.com/huggingface/accelerate/issues/4053

 

## Who can review?
 
  @BenjaminBossan @SunMarc

### DIFF
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -483,7 +483,7 @@ class PartialState:
 
                         # The tensor needs to be on the device before we can pad it
                         tensorized_result = send_to_device(result, self.device)
-                        result = pad_across_processes(tensorized_result, pad_index=inputs[-1])
+                        result = pad_across_processes(tensorized_result, pad_index=pad_index=send_to_device(inputs[-1], self.device))
                     else:
                         result += [inputs[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
                 return result
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -483,7 +483,9 @@ class PartialState:
 
                         # The tensor needs to be on the device before we can pad it
                         tensorized_result = send_to_device(result, self.device)
-                        result = pad_across_processes(tensorized_result, pad_index=send_to_device(inputs[-1], self.device))
+                        result = pad_across_processes(
+                            tensorized_result, pad_index=send_to_device(inputs[-1], self.device)
+                        )
                     else:
                         result += [inputs[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
                 return result
@@ -497,7 +499,7 @@ class PartialState:
 
                     if isinstance(inputs, Dataset):
                         clamped_start = min(start_index, len(inputs))
-                        clamped_end   = min(end_index,   len(inputs))
+                        clamped_end = min(end_index, len(inputs))
                         result_idcs = list(range(clamped_start, clamped_end))
                         if apply_padding:
                             result_idcs += [len(inputs) - 1] * (
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -483,7 +483,7 @@ class PartialState:
 
                         # The tensor needs to be on the device before we can pad it
                         tensorized_result = send_to_device(result, self.device)
-                        result = pad_across_processes(tensorized_result, pad_index=pad_index=send_to_device(inputs[-1], self.device))
+                        result = pad_across_processes(tensorized_result, pad_index=send_to_device(inputs[-1], self.device))
                     else:
                         result += [inputs[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
                 return result
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -476,10 +476,7 @@ class PartialState:
 
         def _split_values(inputs, start_index, end_index):
             if isinstance(inputs, (list, tuple, torch.Tensor)):
-                if start_index >= len(inputs):
-                    result = inputs[-1:]
-                else:
-                    result = inputs[start_index:end_index]
+                result = inputs[start_index:end_index]
                 if apply_padding:
                     if isinstance(result, torch.Tensor):
                         from accelerate.utils import pad_across_processes, send_to_device
@@ -488,7 +485,7 @@ class PartialState:
                         tensorized_result = send_to_device(result, self.device)
                         result = pad_across_processes(tensorized_result, pad_index=inputs[-1])
                     else:
-                        result += [result[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
+                        result += [inputs[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
                 return result
             elif isinstance(inputs, dict):
                 for key in inputs.keys():
@@ -499,13 +496,11 @@ class PartialState:
                     from datasets import Dataset
 
                     if isinstance(inputs, Dataset):
-                        if start_index >= len(inputs):
-                            start_index = len(inputs) - 1
-                        if end_index > len(inputs):
-                            end_index = len(inputs)
-                        result_idcs = list(range(start_index, end_index))
+                        clamped_start = min(start_index, len(inputs))
+                        clamped_end   = min(end_index,   len(inputs))
+                        result_idcs = list(range(clamped_start, clamped_end))
                         if apply_padding:
-                            result_idcs += [end_index - 1] * (
+                            result_idcs += [len(inputs) - 1] * (
                                 num_samples_per_process + (1 if num_extras > 0 else 0) - len(result_idcs)
                             )
                         return inputs.select(result_idcs)


src/accelerate/state.py
```diff
@@ -476,10 +476,7 @@ def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor, ap
 
         def _split_values(inputs, start_index, end_index):
             if isinstance(inputs, (list, tuple, torch.Tensor)):
-                if start_index >= len(inputs):
-                    result = inputs[-1:]
-                else:
-                    result = inputs[start_index:end_index]
+                result = inputs[start_index:end_index]
```
3manifold, line 479: Python slicing returns `[]` naturally for out-of-bounds, which is the correct behavior when `apply_padding=False`.

src/accelerate/state.py
```diff
@@ -488,7 +485,7 @@ def _split_values(inputs, start_index, end_index):
                         tensorized_result = send_to_device(result, self.device)
                         result = pad_across_processes(tensorized_result, pad_index=inputs[-1])
                     else:
-                        result += [result[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
+                        result += [inputs[-1]] * (num_samples_per_process + (1 if num_extras > 0 else 0) - len(result))
```
3manifold, line 490: Use `inputs[-1]` instead of `result[-1]`: `result` may be empty when this process received no samples.

src/accelerate/state.py
```diff
@@ -499,13 +496,11 @@ def _split_values(inputs, start_index, end_index):
                     from datasets import Dataset
 
                     if isinstance(inputs, Dataset):
-                        if start_index >= len(inputs):
-                            start_index = len(inputs) - 1
-                        if end_index > len(inputs):
-                            end_index = len(inputs)
-                        result_idcs = list(range(start_index, end_index))
+                        clamped_start = min(start_index, len(inputs))
```
3manifold, line 501: Clamp both indices to valid range without artificially  inflating `start_index` to `len-1` when `apply_padding` is `False`.

src/accelerate/state.py
```diff
@@ -499,13 +496,11 @@ def _split_values(inputs, start_index, end_index):
...
+                        clamped_end   = min(end_index,   len(inputs))
+                        result_idcs = list(range(clamped_start, clamped_end))
                         if apply_padding:
-                            result_idcs += [end_index - 1] * (
+                            result_idcs += [len(inputs) - 1] * (
```
3manifold, line 505: Pad using the last real element (`len(inputs)-1`), not  `end_index-1` which can be wrong when `result_idcs` is empty.

src/accelerate/state.py
```diff
@@ -488,7 +485,7 @@ def _split_values(inputs, start_index, end_index):
                         tensorized_result = send_to_device(result, self.device)
                         result = pad_across_processes(tensorized_result, pad_index=inputs[-1])
```
SunMarc: results might be empty now 
```suggestion
                        result = pad_across_processes(tensorized_result, pad_index=pad_index=send_to_device(inputs[-1], self.device))
```

SunMarc (approved):
LGTM ! just a nit

SunMarc:
@bot style

SunMarc:
@bot /style

src/accelerate/state.py
```diff
@@ -476,19 +476,16 @@ def split_between_processes(self, inputs: list | tuple | dict | torch.Tensor, ap
...
-                    result = inputs[start_index:end_index]
+                result = inputs[start_index:end_index]
                 if apply_padding:
                     if isinstance(result, torch.Tensor):
                         from accelerate.utils import pad_across_processes, send_to_device
 
                         # The tensor needs to be on the device before we can pad it
                         tensorized_result = send_to_device(result, self.device)
-                        result = pad_across_processes(tensorized_result, pad_index=inputs[-1])
+                        result = pad_across_processes(tensorized_result, pad_index=pad_index=send_to_device(inputs[-1], self.device))
```
SunMarc:
```suggestion
                        result = pad_across_processes(tensorized_result, pad_index=send_to_device(inputs[-1], self.device))
```

SunMarc (approved):

SunMarc:
@bot /style

PR #4054, opened 2026-05-28
docs: add aggregate profiler memory example

### LINKED ISSUE #3300: [Feature Request] add a memory logging manager
Logging accelerator and CPU memory during training models is a common scenario all practitioners run into. As discussed with @muellerzr over Slack, `accelerate` should have a utility for that (its location would be [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/memory.py)).

`peft` has a good reference here:
https://github.com/huggingface/peft/blob/ae55fdcc5c4830e0f9fb6e56f16555bafca392de/examples/oft_dreambooth/train_dreambooth.py#L421

Opening it up as a feature request in case anyone's interested in contributing this. This would be a massive help, IMO.

aryanputta:
## Summary
- add a small aggregate-memory example to the profiler guide
- show how to pair `Accelerator.profile()` with `torch.cuda.reset_peak_memory_stats()`
- document a lightweight way to log peak allocated and reserved memory without adding a new manager API

## Why
Issue #3300 asks for a memory logging manager, but the maintainer discussion suggests that a documented minimal example may already cover the main need for many users.

## Testing
- docs-only change
- reviewed rendered markdown locally in diff form

## Related
- Closes #3300

### DIFF
--- a/docs/source/usage_guides/profiler.md
+++ b/docs/source/usage_guides/profiler.md
@@ -161,6 +161,39 @@ The resulting table output (omitting some columns):
 Self CPU time total: 69.332ms
 ```
 
+If you need one or two aggregate numbers for experiment tracking instead of a full operator table, combine the profiler with the runtime memory stats exposed by PyTorch. This is often enough for logging peak memory usage per run:
+
+```python
+import torch
+from accelerate import Accelerator, ProfileKwargs
+
+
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
Thanks !

PR #4056, opened 2026-05-29
Fix gloo backend error in test_load_checkpoint_and_dispatch_with_broa…

kaixuanliu:
In test case: `tests/test_load_checkpoint_and_dispatch_with_broadcast.py::TestLoadCheckpointAndDispatchWithBroadcast::test_load_checkpoint_and_dispatch_ddp`
the DDP test only registered xccl(using default backend), but DDP's logger requires a gloo backend on the PG (which xccl doesn't fall back to, unlike NCCL), so init now uses the hybrid `cpu:gloo,xpu:xccl` — matching the FSDP setup already in state.py: [L243](https://github.com/huggingface/accelerate/blob/507ea7b29903546f7c773dbe36c8ca128f7c78bb/src/accelerate/state.py#L243)


kaixuanliu:
@SunMarc pls help review, thx!

SunMarc (approved):
Thanks !

kaixuanliu:
The failed CI should not be introduced by this PR. Seems dependant transformers is not correctly installed.

PR #4060, opened 2026-05-30
Add register_tracker_class to register custom trackers by name

### LINKED ISSUE #2734: [Feature Request] Allows register custom trackers to internal tracker type registry
Currently `accelerate` maintains an internal known tracker registry [`LOGGER_TYPE_TO_CLASS`](https://github.com/huggingface/accelerate/blob/2c767338f29989e54ce93b3036b41e02013af7a7/src/accelerate/tracking.py#L960), which is used for tracker initializing, filtering, among other things. However, this makes working with custom trackers annoying in user code. One such example is when initializing custom trackers, instead of something like

```python
accelerator = Accelerator(log_with=args.tracker_names)
```

one has to write:

```python
trackers = [name if name != 'custom_tracker' else CustomTracker() for name in args.tracker_names]
accelerator = Accelerator(log_with=trackers)
```

This is annoying in that users have to write many boilerplate code to manage the init args of both custom trackers and built-in trackers in one place. If the users can register their custom trackers into `LOGGER_TYPE_TO_CLASS`, it will be possible to initialize the custom trackers in the same (and uniform) way as the built-in trackers by:

```python
init_kwargs = dict(
    wandb=dict(
        # init args for wandb
    ),
    custom_tracker=dict(
        # init args for custom tracker named `custom_tracker`
    ),
)
self.accelerator.init_trackers(self.args.project, config, init_kwargs)
```

The API for registering custom trackers could be something like:

```python
accelerator.register_custom_tracker(name, CustomTracker)
```

I have examined the source code, and there seems no obstacle in doing so, as long as the `GeneralTracker` protocol is obeyed by the custom trackers. To make sure of it, it may be beneficial to mark `GeneralTracker` as an abstract base class, though.

javierdejesusda:
# What does this PR do?

Adds `register_tracker_class`, a small public API to register a custom `GeneralTracker` subclass into the internal `LOGGER_TYPE_TO_CLASS` registry, so it can be selected by string name in `Accelerator(log_with=...)`, exactly like the built-in trackers.

This removes the boilerplate described in #2734, where users currently have to special-case custom trackers when building `log_with`:

```python
trackers = [name if name != "custom" else CustomTracker() for name in args.tracker_names]
accelerator = Accelerator(log_with=trackers)
```

With this change:

```python
from accelerate.tracking import register_tracker_class

register_tracker_class(MyTracker)  # MyTracker.name == "my_tracker"
accelerator = Accelerator(log_with="my_tracker")
```

The API matches what was agreed in the issue thread (`name` taken from the class attribute, per @muellerzr's suggestion). It is exposed as a module-level `register_tracker_class` in `accelerate.tracking`.

## Details

- Validates that the class subclasses `GeneralTracker` and defines `name` and `requires_logging_directory`.
- Warns (does not raise) when a registration overwrites an existing name, so shadowing a built-in is intentional but visible.
- Extends `filter_trackers` so registered names resolve through the registry, including when combined with `"all"`, with the same logging-directory guard applied to the built-in trackers.
- Adds tests covering registration, use-by-name end to end, the `requires_logging_directory` path, overwrite warnings, validation errors, and the `"all"` combination, plus usage-guide and API-reference docs.

Fixes #2734

## Before submitting

- [x] Was this discussed/approved via a GitHub issue or the forum? Yes, #2734 (the API was approved by @muellerzr and @SunMarc in the issue thread).
- [x] Did you make sure to update the documentation with your changes?

## Who can review?

@SunMarc @BenjaminBossan

### DIFF
--- a/docs/source/usage_guides/tracking.md
+++ b/docs/source/usage_guides/tracking.md
@@ -170,6 +170,25 @@ tracker = MyCustomTracker("some_run_name")
 accelerator = Accelerator(log_with=[tracker, "all"])
 ```
 
+### Registering a custom tracker by name
+
+Instead of passing an instance, you can register a custom tracker class so it can be selected by its `name`, just like the built-in trackers. This is handy when the trackers to use are read from a config file or the command line:
+
+```python
+from accelerate import Accelerator
+from accelerate.tracking import GeneralTracker, register_tracker_class
+
+
+class MyCustomTracker(GeneralTracker):
+    name = "my_tracker"
+    requires_logging_directory = False
+    # ... implement the rest of the `GeneralTracker` interface
+
+
+register_tracker_class(MyCustomTracker)
+accelerator = Accelerator(log_with="my_tracker")
+```
+
 ## Accessing the internal tracker 
 
 If some custom interactions with a tracker might be wanted directly, you can quickly access one using the 
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -18,6 +18,7 @@
 import json
 import os
 import time
+import warnings
 from functools import wraps
 from typing import Any, Optional, Union
 
@@ -1300,7 +1301,7 @@ def register_tracker_class(tracker_class: type[GeneralTracker]):
             "Set it to `True` if the tracker needs a logging directory, or `False` otherwise."
         )
     if tracker_class.name in LOGGER_TYPE_TO_CLASS:
-        logger.warning(
+        warnings.warn(
             f"A tracker with the name '{tracker_class.name}' is already registered and will be overwritten by "
             f"{tracker_class}. This is especially significant when shadowing a built-in tracker."
         )
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -1341,17 +1341,7 @@ def filter_trackers(
         if not isinstance(log_with, (list, tuple)):
             log_with = [log_with]
         if "all" in log_with or LoggerType.ALL in log_with:
-            loggers = [o for o in log_with if issubclass(type(o), GeneralTracker)]
-            # Also keep custom trackers registered via `register_tracker_class` that are named alongside "all".
-            # Apply the same logging-dir guard as the explicit-name branch for consistency.
-            for custom_name in log_with:
-                if custom_name not in LoggerType and custom_name in LOGGER_TYPE_TO_CLASS:
-                    tracker_init = LOGGER_TYPE_TO_CLASS[str(custom_name)]
-                    if tracker_init.requires_logging_directory and logging_dir is None:
-                        raise ValueError(f"Logging with `{custom_name}` requires a `logging_dir` to be passed in.")
-                    if custom_name not in loggers:
-                        loggers.append(custom_name)
-            loggers += get_available_trackers()
+            loggers = [o for o in log_with if issubclass(type(o), GeneralTracker)] + get_available_trackers()
         else:
             for log_type in log_with:
                 if (
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -1257,6 +1257,56 @@ LOGGER_TYPE_TO_CLASS = {
 }
 
 
+def register_tracker_class(tracker_class: type[GeneralTracker]):
+    """
+    Registers a custom [`GeneralTracker`] subclass so it can be referenced by its `name` in the `log_with` argument of
+    [`Accelerator`], the same way as the built-in trackers.
+
+    The tracker must be registered before instantiating the [`Accelerator`] that uses it.
+
+    Args:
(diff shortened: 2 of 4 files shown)


src/accelerate/accelerator.py
```diff
@@ -3268,6 +3268,39 @@ def wait_for_everyone(self):
         """
         wait_for_everyone()
 
+    @staticmethod
+    def register_tracker_class(tracker_class: type[GeneralTracker]):
+        """
+        Thin wrapper around [`~tracking.register_tracker_class`]. Registers a custom [`GeneralTracker`] subclass so it
```
SunMarc: let's not expose it here

docs/source/usage_guides/tracking.md
```diff
@@ -170,6 +170,27 @@ tracker = MyCustomTracker("some_run_name")
...
+    requires_logging_directory = False
+    # ... implement the rest of the `GeneralTracker` interface
+
+
+register_tracker_class(MyCustomTracker)
+accelerator = Accelerator(log_with="my_tracker")
+```
+
+A registered tracker is instantiated with the `project_name` passed to [`Accelerator.init_trackers`] as the first positional argument. When `requires_logging_directory` is `True`, the logging directory is passed as the second positional argument (matching the built-in tracker convention), so the `__init__` should accept `(run_name, logging_dir, **kwargs)` in that case. Tracker-specific keyword arguments from `init_kwargs` are always forwarded as `**kwargs`. Register the class before creating the [`Accelerator`] that uses it.
+
```
SunMarc: remove this

```suggestion
```

src/accelerate/tracking.py
```diff
@@ -1293,14 +1343,28 @@ def filter_trackers(
...
+            loggers = [o for o in log_with if issubclass(type(o), GeneralTracker)]
+            # Also keep custom trackers registered via `register_tracker_class` that are named alongside "all".
+            # Apply the same logging-dir guard as the explicit-name branch for consistency.
+            for custom_name in log_with:
+                if custom_name not in LoggerType and custom_name in LOGGER_TYPE_TO_CLASS:
+                    tracker_init = LOGGER_TYPE_TO_CLASS[str(custom_name)]
+                    if tracker_init.requires_logging_directory and logging_dir is None:
+                        raise ValueError(f"Logging with `{custom_name}` requires a `logging_dir` to be passed in.")
+                    if custom_name not in loggers:
+                        loggers.append(custom_name)
```
SunMarc: well here, log_with is either "all" or "LoggerType.ALL" no ?

tests/test_tracking.py
```diff
@@ -731,6 +734,192 @@ def test_log(self):
                 assert data == truth
 
 
+class MyRegisteredTracker(GeneralTracker):
```
SunMarc, line 812: keep tests simple

SunMarc (commented):
Left a couple of comments thanks !

src/accelerate/accelerator.py
javierdejesusda: Removed, module-level `register_tracker_class` is the only entry point now.

docs/source/usage_guides/tracking.md
javierdejesusda: Done, removed.

src/accelerate/tracking.py
javierdejesusda: `log_with` is the full list here, so it can be e.g. `["my_tracker", "all"]` the loop picks up a tracker registered by name when it's listed alongside `"all"`, since `"all"` on its own only pulls in the built-ins. Happy to drop it and keep just instance + `"all"` if you'd prefer the smaller surface.

tests/test_tracking.py
javierdejesusda: Will trim these to the shared fixture plus the core cases, together with the `filter_trackers` decision above (some depend on it).

src/accelerate/tracking.py
SunMarc: yeah, i think it is better to drop it. it's a bit counter intuitive to pass ["my_tracker", "all"]

javierdejesusda: Done in 155194db: "all" is back to main's behavior, so a registered name passed alongside "all" is no longer picked up. Selecting by name still works on its own via log_with="my_tracker".

javierdejesusda:
Extra commit d22cb634: the overwrite warning used logger.warning, but accelerate's logger raises RuntimeError before PartialState/Accelerator init, so shadowing or re-registering a name in the documented "register first" flow would crash instead of warn. Switched that call to warnings.warn (matches accelerator.py) and the test now asserts via assertWarns.

javierdejesusda:
Not stale, this is still ready for review.  @SunMarc when you have a moment, could you take another look? Happy to adjust anything else if needed.

SunMarc (approved):
Thanks !

PR #4061, opened 2026-06-01
Add `register_tracker_class()` for custom tracker registration by name

### LINKED ISSUE #2734: [Feature Request] Allows register custom trackers to internal tracker type registry

devangpratap:
# What does this PR do?

Adds `register_tracker_class()` so custom tracker classes can be passed by string name to `log_with` and `init_trackers`, matching built-in tracker behavior. Previously users had to pass pre-instantiated tracker objects.

```python
from accelerate import Accelerator, register_tracker_class

register_tracker_class(MyTracker)
accelerator = Accelerator(log_with="my_tracker")
accelerator.init_trackers("project", init_kwargs={"my_tracker": {"key": "value"}})
```

Also exposed as `Accelerator.register_tracker_class()` per the API discussed in the issue.

Fixes #2734

## Before submitting
      Pull Request section?
      to it if that's the case. https://github.com/huggingface/accelerate/issues/2734

## Who can review?

@SunMarc @muellerzr

### DIFF
--- a/src/accelerate/__init__.py
+++ b/src/accelerate/__init__.py
@@ -28,6 +28,7 @@ from .inference import prepare_pippy
 from .launchers import debug_launcher, notebook_launcher
 from .parallelism_config import ParallelismConfig
 from .state import PartialState
+from .tracking import register_tracker_class
 from .utils import (
     AutocastKwargs,
     DataLoaderConfiguration,
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -42,7 +42,7 @@ from .optimizer import AcceleratedOptimizer
 from .parallelism_config import ParallelismConfig
 from .scheduler import AcceleratedScheduler
 from .state import AcceleratorState, GradientState, PartialState
-from .tracking import LOGGER_TYPE_TO_CLASS, GeneralTracker, filter_trackers
+from .tracking import LOGGER_TYPE_TO_CLASS, GeneralTracker, filter_trackers, register_tracker_class
 from .utils import (
     MODEL_NAME,
     SAFE_WEIGHTS_INDEX_NAME,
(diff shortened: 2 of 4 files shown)


SunMarc:
I will focus on this PR as it was opened first: https://github.com/huggingface/accelerate/pull/4060

PR #4071, opened 2026-06-10
fix: preserve dataloader iteration state when resuming from checkpoint

TomQunChao:
# What does this PR do?

## Problem
Even when using raw checkpoints, resuming training does not reproduce the same loss values as training from scratch. For example(with use of DataLoaderShard):

- **Direct training**: steps 0 → 500
- **Resume training**: steps 0 → 200 → save_ckpt → load_ckpt → 500

When comparing loss at step 201 between these two scenarios, the values are inconsistent.

Further investigation revealed this only occurs when resuming from a non-epoch boundary. With 125 steps per epoch, resuming at step 200 (mid-epoch 1) triggers the bug, while resuming at epoch boundaries works correctly.

## Root Cause
The issue lies in the interaction between `set_epoch` and `skip_first_batches` in the training loop:

1. When resuming at epoch 1 with `steps_trained_in_current_epoch=75`, the original dataloader correctly has its sampler epoch set to 1 via `epoch_dataloader.set_epoch(1)`
2. `skip_first_batches` creates a new `DataLoaderShard`, but its `iteration` attribute is initialized to 0
3. When iteration begins, `__iter__` calls `self.set_epoch(self.iteration)`, which calls `set_epoch(0)` 
4. This incorrectly resets the sampler's epoch from 1 back to 0, causing data loading order to diverge

The sampler epoch mismatch leads to different random seeds for data shuffling, resulting in non-reproducible loss values.

## Fix
Preserve the original dataloader's `iteration` value when creating the new dataloader in `skip_first_batches`. This ensures the sampler's epoch remains correct after batch skipping, maintaining data sequence consistency.

## Verification
- Direct training (0→500) and resume training (0→200→save→load→500) now produce identical loss at step 201
- Loss reproducibility is maintained regardless of where the checkpoint was saved (epoch boundaries or mid-epoch)
- No impact on standard (non-resume) training behavior


SunMarc (commented):
Can you add a test ? Thanks !

TomQunChao:
> Can you add a test ? Thanks !

Done! Added unit tests for iteration preservation in skip_first_batches. PTAL, thanks.

tests/test_data_loader.py
```diff
@@ -645,6 +645,40 @@ def test_set_epoch_in_batch_sampler(self):
...
+        dataloader = DataLoaderShard(dataset, batch_sampler=batch_sampler)
+
+        dataloader.set_epoch(1)
+        new_dataloader = skip_first_batches(dataloader, num_batches=2)
+
+        # Consume the iterator; __iter__ calls self.set_epoch(self.iteration) internally.
+        list(new_dataloader)
+
+        # The sampler epoch must remain 1 (propagated by set_epoch(1)), not be reset to 0.
+        assert batch_sampler.epoch == 1
```
SunMarc: this passes without the fix

SunMarc (commented):
should this be propagated to DataLoaderDispatcher also ?

tests/test_data_loader.py
TomQunChao: Done! I have repair the unit test, and also apply the patch to DataloaderDispatcher. PTAL, thanks

BUT there is another issue in DataloaderDispatcher(cannot propagate set_epoch to sampler when using SkipBatchSampler), which will repair at next PR.

SunMarc (approved):
SGTM !

SunMarc:
@bot /style

PR #4073, opened 2026-06-12
Fix flaky behavior in gated-model CLI test by making the test determi…

sywangyi:
…nistic and independent of local Hugging Face cache state.

@SunMarc please help review


sywangyi:
Background
The test in test_cli.py:469 could fail on machines where meta-llama/Llama-2-7b-hf was already cached locally.
That made test outcomes environment-dependent instead of validating the intended gated-access behavior.

SunMarc (approved):
LGTM

PR #4077, opened 2026-06-13
fix(utils): don't crash check_os_kernel on non-standard kernel release strings

eldar702:
## Summary
`check_os_kernel()` aborts `Accelerator` initialization with an opaque `ValueError: not enough values to unpack` on Linux hosts whose kernel release string doesn't contain a full `X.Y.Z` version — e.g. custom kernels, some containers/WSL reporting `"5.15"`, `"6.1-arch1"`, or `"unknown"`. The cause is `_, version, *_ = re.split(r"(\d+\.\d+\.\d+)", info.release)`, which yields too few parts when there's no `X.Y.Z` match.

## Fix
Use `re.search` for the `X.Y.Z` pattern and return early (skip the version warning) when no standard version is present. On standard release strings the extracted version is identical to before — no behavior change.

## Tests
Regression test covering non-standard release strings (RED before / GREEN after). `tests/test_utils.py`: 46 passed; ruff clean.

*AI-assisted contribution.*

### DIFF
--- a/src/accelerate/utils/other.py
+++ b/src/accelerate/utils/other.py
@@ -536,7 +536,13 @@ def check_os_kernel():
     if system != "Linux":
         return
 
-    _, version, *_ = re.split(r"(\d+\.\d+\.\d+)", info.release)
+    # Some kernels report a release string without a full X.Y.Z version (e.g. custom builds,
+    # certain containers/WSL). In that case we cannot meaningfully compare versions, so we skip
+    # the check rather than raising -- this function is only meant to emit a best-effort warning.
+    match = re.search(r"\d+\.\d+\.\d+", info.release)
+    if match is None:
+        return
+    version = match.group()
     min_version = "5.5.0"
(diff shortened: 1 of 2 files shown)


eldar702:
Still ready for review whenever a maintainer has bandwidth — this is a small, self-contained fix (parses `check_os_kernel`'s kernel-release string more defensively so it doesn't crash on non-standard formats). No CI is configured to run automatically on this repo's fork PRs, so there's nothing red blocking it; happy to make any changes requested.

eldar702:
Still ready for review whenever a maintainer has bandwidth. Nothing has changed since my last update — this remains a small, self-contained fix (`check_os_kernel` parses the kernel-release string more defensively so it doesn't crash on non-standard formats), no CI runs automatically on fork PRs here, and there's nothing red blocking it. Happy to address any feedback.

SunMarc:
Not a real issue, closing !

PR #4083, opened 2026-06-18
[Offloading] Support full disk offloading

kylesayrs:
## Purpose ##
* Support full disk offloading of an entire model
  * This is useful for testing and cases like [vllm-project/llm-compressor](https://github.com/vllm-project/llm-compressor) which heavily utilizes disk offloading to compress large models
* Add missing handling for when tied weights are disk offloaded
* This PR can be merged independently from https://github.com/huggingface/transformers/pull/46749

## Changes ##
* Force hook to be added if any modules are dispatched to disk (disk offloading relies on hooks)
* Do not add disk offloaded meta tensors to the `tied_params_map`. Because the data pointer for meta tensors is always `0`, this would cause all weights to share the same onloaded same weight value
  * This is a bug that was likely never encountered during testing, as embeddings are almost never disk offloaded
  * It is possible to share onloads between disk-offloaded params, but this current system doesn't support it, and the cases in which that's actually useful would be very rare

## Testing ##
* Added full disk offloading test
* Tested changes with `test_disk_offloading.py` in conjunction with https://github.com/huggingface/transformers/pull/46749

<details open><summary>test_disk_offloading.py</summary>

```python
import torch
import pytest

# Qwen3-0.6B has tied weights, two copies in checkpoint
# Llama-3.2-1B-Instruct has tied weights, only one copy in checkpoint
# tinysmokeqwen3 has untied weights
@torch.no_grad()
@pytest.mark.parametrize("model_id,cpu_mem", [
    ("Qwen/Qwen3-0.6B", 0),
    ("Qwen/Qwen3-0.6B", 6e8),
    ("meta-llama/Llama-3.2-1B-Instruct", 0),
    ("meta-llama/Llama-3.2-1B-Instruct", 1e9),
    ("nm-testing/tinysmokeqwen3", 0),
    ("nm-testing/tinysmokeqwen3", 1.5e6),
])
def test_tied_in_checkpoint(model_id, cpu_mem):
    true_model = AutoModelForCausalLM.from_pretrained(model_id, device_map="auto")
        model_id,
        device_map="auto",
        max_memory={"cpu": cpu_mem},
        offload_folder="./offload_folder",
    )

    inputs = {key: value.to(true_model.device) for key, value in model.dummy_inputs.items()}

    true_outputs = true_model(**model.dummy_inputs)
    outputs = model(**model.dummy_inputs)
    assert torch.nn.functional.mse_loss(true_outputs.logits, outputs.logits) < 1e-2
```

</details>

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -373,10 +373,10 @@ def dispatch_model(
     # in the unique device and the user can decide where to dispatch the model.
     # If the model is quantized, we always force-dispatch the model
     devices = set(device_map.values())
-    if (len(devices) > 1) or "disk" in devices or force_hooks:
+    if len(devices) > 1 or "disk" in devices or force_hooks:
         if main_device is None:
             non_offloaded_devices = devices - {"cpu", "disk"}
-            main_device = non_offloaded_devices[0] if non_offloaded_devices else "cpu"
+            main_device = list(non_offloaded_devices)[0] if non_offloaded_devices else "cpu"
 
         if main_device != "cpu":
             cpu_modules = [name for name, device in device_map.items() if device == "cpu"]
@@ -420,13 +420,16 @@ def dispatch_model(
         tied_params_map = {}
         for group in tied_params:
             for param_name in group:
-                # data_ptr() is enough here, as `find_tied_parameters` finds tied params simply by comparing `param1 is param2`, so we don't need
-                # to care about views of tensors through storage_offset.
-                data_ptr = recursive_getattr(model, param_name).data_ptr()
-                tied_params_map[data_ptr] = {}
+                offloaded_param = recursive_getattr(model, param_name)
 
                 # Note: To handle the disk offloading case, we can not simply use weights_map[param_name].data_ptr() as the reference pointer,
                 # as we have no guarantee that safetensors' `file.get_tensor()` will always give the same pointer.
+                # Therefore, we must skip and induce a runtime cost in the rare case that two tied disk-offloaded params are onloaded simultaneously
+                if offloaded_param.device.type != "meta":
+                    # data_ptr() is enough here, as `find_tied_parameters` finds tied params simply by comparing `param1 is param2`, so we don't need
+                    # to care about views of tensors through storage_offset.
+                    data_ptr = recursive_getattr(model, param_name).data_ptr()
+                    tied_params_map[data_ptr] = {}
 
         attach_align_device_hook_on_blocks(
             model,
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -428,7 +428,7 @@ def dispatch_model(
                 if offloaded_param.device.type != "meta":
                     # data_ptr() is enough here, as `find_tied_parameters` finds tied params simply by comparing `param1 is param2`, so we don't need
                     # to care about views of tensors through storage_offset.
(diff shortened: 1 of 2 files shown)


src/accelerate/big_modeling.py
```diff
@@ -508,12 +510,9 @@ def wrapper(*args, **kwargs):
             device = f"musa:{device}"
         elif is_neuron_available() and isinstance(device, int):
             device = f"neuron:{device}"
-        if device != "disk":
-            model.to(device)
         else:
-            raise ValueError(
-                "You are trying to offload the whole model to the disk. Please use the `disk_offload` function instead."
-            )
+            assert device != "disk"
```
SunMarc: let's not add this assert
```suggestion
```

tests/test_big_modeling.py
```diff
@@ -777,6 +777,19 @@ def test_dispatch_model_with_unused_submodules_multi_device(self):
             output = model(x)
             torch.testing.assert_close(expected, output.cpu(), atol=ATOL, rtol=RTOL)
 
+    def test_dispatch_model_disk_only(self):
+        model = ModelForTest()
+
+        device_map = {"linear1": "disk", "batchnorm": "disk", "linear2": "disk"}
```
SunMarc, line 784: You can just do that instead {"":"disk"}

```suggestion
        device_map = {"": "disk"}
```

src/accelerate/big_modeling.py
```diff
@@ -421,13 +420,16 @@ def dispatch_model(
...
-                tied_params_map[data_ptr] = {}
+                offloaded_param = recursive_getattr(model, param_name)
 
                 # Note: To handle the disk offloading case, we can not simply use weights_map[param_name].data_ptr() as the reference pointer,
                 # as we have no guarantee that safetensors' `file.get_tensor()` will always give the same pointer.
+                # Therefore, we must skip and induce a runtime cost in the rare case that two tied disk-offloaded params are onloaded simultaneously
+                if offloaded_param.device.type != "meta":
+                    # data_ptr() is enough here, as `find_tied_parameters` finds tied params simply by comparing `param1 is param2`, so we don't need
+                    # to care about views of tensors through storage_offset.
+                    data_ptr = recursive_getattr(model, param_name).data_ptr()
```
SunMarc:
```suggestion
                    data_ptr = offloaded_param.data_ptr()
```

SunMarc (commented):
Thanks a lot, just a few nits !

tests/test_big_modeling.py
kylesayrs: I agree, this is more reflective of the actual device map that is used in practice.

src/accelerate/big_modeling.py
kylesayrs: Thanks

tests/test_big_modeling.py
kylesayrs: Actually, cpu and disk offloading requires an expanded device map. Otherwise `extract_submodules_state_dict` as used by `dispatch_model` doesn't work properly, which means that `offload_state_dict` doesn't work properly.

I can attempt to support non-expanded device maps for `dispatch_model`, although it increases the scope of changes a bit. I've add a note in the test, I would recommend leaving this for a separate PR.

SunMarc: okay sg !

SunMarc (approved):
Thanks a lot !

PR #4084, opened 2026-06-19
SP ulysses native implementation

SunMarc:
# What does this PR do?

This PR adds a native SP implementation in accelerate so that it is compatible with FSDPv2 and deepspeed. 

cc @qgallouedec for viz

### DIFF
--- a/src/accelerate/accelerator.py
+++ b/src/accelerate/accelerator.py
@@ -1539,6 +1539,8 @@ class Accelerator:
 
         if self.parallelism_config and self.parallelism_config.cp_enabled:
             args = self._prepare_cp(*args)
+        if self.parallelism_config and self.parallelism_config.sp_enabled:
+            args = self._prepare_sp(*args)
         # for megatron-lm, we don't need to prepare TE AO at this moment
         if self.distributed_type != DistributedType.MEGATRON_LM:
             if self.fp8_backend == FP8BackendType.TE:
@@ -1558,6 +1560,8 @@ class Accelerator:
                 self._prepare_one(obj, first_pass=True, device_placement=d) for obj, d in zip(args, device_placement)
             )
             result = tuple(self._prepare_one(obj, device_placement=d) for obj, d in zip(result, device_placement))
+        if self.parallelism_config and self.parallelism_config._sequence_sharding_enabled:
+            result = self._wrap_sequence_sharding_dataloaders(result)
         if tpu_should_fix_optimizer:
             # 2. grabbing new model parameters
             new_named_params = self._get_named_parameters(*result)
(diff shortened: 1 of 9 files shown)


src/accelerate/accelerator.py
```diff
@@ -1670,6 +1674,40 @@ def _prepare_cp(self, *args):
...
+        contiguously over the `sp` group. `prepare_data_loader` already hands sp ranks the same
+        sample (dp-aware sharding divides out tp*cp*sp), so the wrapper only does the sequence
+        split; packed cu_seqlens are pushed onto the handler (`set_varlen`)."""
+        from .utils.sequence_parallel import SequenceShardingDataLoader
+
+        shard_group = self.torch_device_mesh["sp"].get_group()
+        wrapped = []
+        for obj in result:
+            if isinstance(obj, torch.utils.data.DataLoader) and not isinstance(obj, SequenceShardingDataLoader):
+                obj = SequenceShardingDataLoader(obj, shard_group, attention=self._sp_attention)
```
SunMarc, line 1707: `_sp_attention` is a bit ugly, this might change. lmk @winglian if you have a better design for this in order to deal with packed sequences.

src/accelerate/parallelism_config.py
```diff
@@ -352,8 +394,35 @@ def _set_size(self, parallelism: str, size: int):
...
+        handler_cls = (
+            DeepSpeedSequenceParallelConfig if self.sp_backend == "deepspeed" else AccelerateSequenceParallelConfig
+        )
+        if not isinstance(self.sp_handler, handler_cls):
+            self.sp_handler = handler_cls()
+
     def _validate_accelerator(self, accelerator: "Accelerator"):
         _warnings = set()
+        # Resolve auto backends + enforce backend<->engine compatibility now that the engine is known.
+        self._resolve_backends(accelerator)
```
winglian, line 425: is the intention to build the mesh before we validate it in https://github.com/huggingface/accelerate/blob/3dea8f4750ef9c3c71e282325d7cc27b7fbd59a5/src/accelerate/accelerator.py#L475-L476

src/accelerate/utils/sequence_parallel.py
```diff
@@ -0,0 +1,243 @@
...
+        k = GatherSeqScatterHeads.apply(key, group, sp).transpose(1, 2)
+        v = GatherSeqScatterHeads.apply(value, group, sp).transpose(1, 2)
+        # Drop the LOCAL-shard kwargs (position_ids + flash varlen indices keyed to the pre-gather
+        # length); re-inject the GLOBAL packed cu_seqlens for the full sequence the all-to-all rebuilt.
+        for k_ in ("position_ids", "cu_seq_lens_q", "cu_seq_lens_k", "max_length_q", "max_length_k"):
+            kwargs.pop(k_, None)
+        if self.cu_seqlens is not None:
+            kwargs["cu_seq_lens_q"] = kwargs["cu_seq_lens_k"] = self.cu_seqlens.to(q.device)
+            kwargs["max_length_q"] = kwargs["max_length_k"] = self.max_seqlen
+        out, _ = self.attn_fn(module, q, k, v, None, **kwargs)
```
winglian, line 126: when using sdpa, the None here breaks packing since sdpa doesn't use cu_seql_lens, we need to pass position_ids iirc for spa to generate the packed/varlen masK

src/accelerate/utils/sequence_parallel.py
```diff
@@ -0,0 +1,243 @@
...
+        raise ValueError(
+            f"Ulysses SP expects a HF Transformers model with a `config` attribute, got {type(model).__name__}."
+        )
+    from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS
+
+    _check_kv_divisible(model, dist.get_world_size(sp_group))
+    # Capture the model's ORIGINAL attention. If SP was already enabled (re-prepare, second model),
+    # the registered entry is our own handler — unwrap it so we don't wrap a wrapper (double all-to-all).
+    attn_impl_str = model.config.get_text_config()._attn_implementation
+    current = ALL_ATTENTION_FUNCTIONS[attn_impl_str]
```
winglian, line 144: I had to tackle something similar in my branch where this is a global rewrite of the attn dict, so could have unintended consequences when using something like DPO, RL, or other pattern that has a second model loaded.

PR #4093, opened 2026-06-29
Fix pad_input_tensors to pad batch to a multiple of num_processes

iamsharduld:
## What

`pad_input_tensors` (`_pad_input_tensors`) computes the padding as:

```python
to_pad = num_processes - (batch_size // num_processes)
```

which is unrelated to the padding needed to make the batch splittable across processes. It:
- pads batches that are **already divisible** (e.g. `batch_size=8, num_processes=4` → **10**), and
- produces sizes that are frequently **not a multiple of `num_processes`** (e.g. `6 → 9`, `12 → 13`,
  `10,np=5 → 13`), so the padded batch still can't be chunked evenly.

`inference.py::pippy_forward` relies on this to split a batch into `num_chunks` for `ScheduleGPipe`,
so a non-divisible result breaks even micro-batching. The existing `test_slice_and_concatenate` cases
pass only because each `(batch_size, num_processes)` pair happens to coincide with the correct value.

```python
from accelerate.utils.operations import pad_input_tensors
import torch
pad_input_tensors(torch.rand(8,4), 8, 4).shape   # [10,4]  -- should be [8,4]
pad_input_tensors(torch.rand(6,4), 6, 4).shape   # [9,4]   -- should be [8,4]
```

## Fix

Pad dim-0 up to the next multiple of `num_processes` (0 when already divisible):

```python
to_pad = (num_processes - batch_size % num_processes) % num_processes
```

This satisfies the docstring example (`[3,4,4], np=4 → [4,4,4]`), all 8 existing assertions, and
always yields a size `>= batch_size` divisible by `num_processes`.

## Tests

`tests/test_utils.py::test_pad_input_tensors_divisibility` — checks `8→8`, `6→8`, and the invariant
`result % num_processes == 0 and result >= batch_size` across `num_processes∈[1,8], batch_size∈[1,29]`
(129/232 cases were non-divisible before). Existing `test_slice_and_concatenate` still passes. CPU-only.

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -814,16 +814,9 @@ def pad_input_tensors(tensor, batch_size, num_processes, dim=0):
     """
 
     def _pad_input_tensors(tensor, batch_size, num_processes, dim=0):
-        remainder = batch_size // num_processes
-        last_inputs = batch_size - (remainder * num_processes)
-        if batch_size // num_processes == 0:
-            to_pad = num_processes - batch_size
-        else:
-            to_pad = num_processes - (batch_size // num_processes)
-        # In the rare case that `to_pad` is negative,
-        # we need to pad the last inputs - the found `to_pad`
-        if last_inputs > to_pad & to_pad < 1:
-            to_pad = last_inputs - to_pad
+        # Pad dim-0 up to the next multiple of num_processes so the batch can be split
+        # evenly across processes (0 padding when batch_size is already divisible).
+        to_pad = (num_processes - batch_size % num_processes) % num_processes
         old_size = tensor.shape
         new_size = list(old_size)
         new_size[0] = batch_size + to_pad
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -392,6 +392,19 @@ class UtilsTester(unittest.TestCase):
         # We should expect there to be 66 items now
         assert result.shape == torch.Size([66, 4, 4])
 
+    def test_pad_input_tensors_divisibility(self):
+        # pad_input_tensors must pad dim-0 up to the next multiple of num_processes
+        # (0 padding when already divisible), so the batch can be split evenly.
+        # Regression: the old code padded 8->10 and 6->9 (not divisible by 4).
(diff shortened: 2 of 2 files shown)


vineethsaivs:
This was closed by the stale bot rather than on review, and the bug it describes is still on `main`. I ran into it independently and can confirm the analysis, so leaving the verification here in case a maintainer wants to reopen.

Exhaustively over `num_processes` 1 to 8 and `batch_size` 1 to 24, **104 of the 192 pairs come back with a padded batch that is not a multiple of `num_processes`**, so `ScheduleGPipe` still cannot chunk it evenly. Some of the shortest cases:

| num_processes | batch_size | padded | padded % num_processes |
| --- | --- | --- | --- |
| 2 | 2 | 3 | 1 |
| 3 | 3 | 5 | 2 |
| 4 | 4 | 7 | 3 |
| 4 | 6 | 9 | 1 |
| 4 | 8 | 10 | 2 |
| 5 | 10 | 13 | 3 |

The `batch_size == num_processes` rows are the clearest tell: nothing needs padding there at all, and `pad_input_tensors(x, 4, 4)` returns 7 rows. `pippy_forward` happens not to hit that one because it guards with `if found_batch_size != num_chunks`, but `pad_input_tensors` is exported from `accelerate.utils`, so a direct caller does.

The reason `test_slice_and_concatenate` passes today is that all eight of its pairs land on a coincidence: seven have `batch_size <= num_processes` or a small remainder, and the eighth (`61`, `6`) goes through the negative-`to_pad` correction. Every one of the eight expectations is also satisfied by the round-up form `(num_processes - batch_size % num_processes) % num_processes`, so the fix here does not change a single existing assertion, it only adds the cases the suite never covered.

Worth noting that #4150 is open against the same function for a different defect (the padded axis is hardcoded to `new_size[0]` instead of `new_size[dim]`), so the two are complementary rather than competing.

@iamsharduld are you still planning to pick this back up? Happy to leave it with you; I only wanted the analysis on the record so it does not go stale a second time.

iamsharduld:
Thanks for confirming! Yes would be be happy if someone from review team picks this up next week:)

PR #4097, opened 2026-07-02
Add `neuron` backend for torch dynamo

michaelbenayoun:
# What does this PR do?

As per title.

Enables to use the `--torch-compile` flag in Transformers with the Trainer.


SunMarc (approved):
LGTM

PR #4104, opened 2026-07-06
Fix --parallelism_config_sp_seq_length_is_variable parsing "False" as "True"

### LINKED ISSUE #4102: `--parallelism_config_sp_seq_length_is_variable False` parses as True
### System Info

```Shell
Branch: main

```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```python
import argparse

parser.add_argument(
    type=bool,
    default=True,
)

args = parser.parse_args([
    "False",
])

```

Actual output:

```text
bool
True
```

```bash
accelerate launch \
  --use_fsdp \
  --fsdp_version 2 \
  --use_parallelism_config \
  train.py
```

### Expected behavior

Passing:

```bash
```

should either:

Suggested tests:

CNTRL-tek91:
Fixes #4102

This fixes a bug in the `accelerate launch` CLI where passing `--parallelism_config_sp_seq_length_is_variable False` was silently parsed as `True`.

The argument used `type=bool`, and `bool()` on a string only checks whether the string is empty, so `bool("False")` is `True`. Any non-empty value was being read as `True`, with no way to actually turn this option off from the command line.

I switched it to `str_to_bool`, which is already used elsewhere in this file for boolean flags. The parsed value later gets written to an environment variable and compared against the literal string "true" in `dataclasses.py`, so I used `partial(str_to_bool, to_bool=True)` to get a real bool back instead of `str_to_bool`'s default int, keeping that comparison correct.

Added a test in `tests/test_cli.py` covering the explicit `False`, explicit `True`, and omitted cases.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -20,6 +20,7 @@ import logging
 import os
 import subprocess
 import sys
+from functools import partial
 from pathlib import Path
 
 import torch
@@ -963,7 +964,7 @@ def launch_command_parser(subparsers=None):
 
     parallelism_config_args.add_argument(
         "--parallelism_config_sp_seq_length_is_variable",
-        type=bool,
+        type=partial(str_to_bool, to_bool=True),
         default=True,
         help="If `True` will work with a sequence length that may change between batches, in which case `parallelism_config_sp_seq_length` value can be set to anything divisible by sp size or remain unset. If `False` then `parallelism_config_sp_seq_length` needs to match the batch's sequence length dimension. The default is `True`.",
     )
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -15,7 +15,8 @@
 import argparse
 import unittest
 
-from accelerate.utils.launch import prepare_multi_gpu_env
+from accelerate.commands.launch import launch_command_parser
+from accelerate.utils.launch import prepare_extend_env_parallelism_config, prepare_multi_gpu_env
 
 
 class TestPrepareMultiGpuEnv(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


src/accelerate/commands/launch.py
```diff
@@ -963,7 +964,7 @@ def launch_command_parser(subparsers=None):
 
     parallelism_config_args.add_argument(
         "--parallelism_config_sp_seq_length_is_variable",
-        type=bool,
+        type=partial(str_to_bool, to_bool=True),
```
SunMarc, line 967:
```suggestion
        type=str_to_bool,
```

src/accelerate/commands/launch.py
```diff
@@ -963,7 +964,7 @@ def launch_command_parser(subparsers=None):
...
         default=True,
```
SunMarc, line 968:
```suggestion
        default="true",
```

tests/test_launch.py
```diff
@@ -15,7 +15,8 @@
 import argparse
```
SunMarc, line 15: let's just have one test ! we can remove this one

SunMarc (commented):
LGTM, just a nit so that it follows the same behavior as the others

src/accelerate/commands/launch.py
CNTRL-tek91: Thanks for taking a look!

With this change, I think it would create a new bug. Without `to_bool=True`, `str_to_bool` returns either a `1` or `0` int instead of a real bool. That int value gets written with `str()` into an env variable, and when it's read back, `ParallelismConfig.__post_init__` compares it against the literal string `"true"`.

If, for example, `str_to_bool` returned `1`, then `str(1)` is `"1"`, not `"true"`, so this would flip the default case and an explicit `True` to evaluate as `False` later on, leaving only the explicit `False` correct.

I confirmed this by running the parser directly:

```
omitted        -> 1 (int), str() -> "1"
explicit True  -> 1 (int), str() -> "1"
explicit False -> 0 (int), str() -> "0"
```

Other flags in this file get away with bare `str_to_bool` since they're read back through `parse_flag_from_env`, which reparses the value and accepts ints. This flag is read back with a direct string comparison instead, so it actually needs a real bool, which is what `to_bool=True` gives us.

src/accelerate/commands/launch.py
CNTRL-tek91: For this one, since `default=True` is a real bool and not a string, argparse doesn't run it through `type` at all, so it already stringifies to `"True"` correctly. I feel like changing it to `default="true"` isn't entirely necessary, but if you'd still like it for consistency with the other flags, I'm glad to make that change as long as `type` stays `partial(str_to_bool, to_bool=True)`.

I also tested that pairing, and it gives the correct value in the omitted, explicit `True`, and explicit `False` cases. Let me know if you'd like me to make that change!

tests/test_launch.py
CNTRL-tek91: If it's alright with you, can we keep both? The issue asked for both a parser test and an env bridge test. This one covers the actual env variable write, which is where the bug actually showed up. Let me know if you still think it's redundant!

PR #4107, opened 2026-07-08
Fix megatron_lm bool launch flags parsing "False" as True

### LINKED ISSUE #4102: `--parallelism_config_sp_seq_length_is_variable False` parses as True

teddytennant:
The seven `--megatron_lm_*` boolean launch flags (`use_custom_fsdp`, `no_load_optim`, `eod_mask_loss`, `overlap_cpu_optimizer_d2h_h2d`, `no_save_optim`, `optimizer_cpu_offload`, `use_precision_aware_optimizer`) are declared with `type=bool`. argparse applies the callable to the raw string, and `bool("False")` is `True`, so passing e.g. `--megatron_lm_no_save_optim False` wrongly *enables* the flag.

This switches each to `type=partial(str_to_bool, to_bool=True)`, matching the converter already used for other boolean CLI options. The value serializes via `str(...)` and is read back with `str_to_bool(...) == 1` downstream, so `"True"`/`"False"` round-trip correctly.

Same class of bug as #4102, which #4104 fixes only for `--parallelism_config_sp_seq_length_is_variable`; this completes it for the megatron flags.

### Test

Added `test_megatron_lm_bool_flags` in `tests/test_cli.py`, asserting `False`/`True`/default parse to the correct booleans.

```
pytest tests/test_cli.py -k megatron
```

Closes #4102.

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -20,6 +20,7 @@ import logging
 import os
 import subprocess
 import sys
+from functools import partial
 from pathlib import Path
 
 import torch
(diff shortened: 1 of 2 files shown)


teddytennant:
Still relevant. This fixes megatron_lm bool launch flags where passing "False" was parsed as True.

PR #4108, opened 2026-07-08
Fix sequence parallel bool parsing

### LINKED ISSUE #4102: `--parallelism_config_sp_seq_length_is_variable False` parses as True

lxingy3:
﻿Summary
- parse `--parallelism_config_sp_seq_length_is_variable` with the existing boolean helper instead of `bool`
- add launch parser and env-bridge coverage for the `False` spelling

Tests
- `PYTHONPATH=src python -m pytest tests/test_launch.py -q`
- `python -m ruff format --check src/accelerate/commands/launch.py tests/test_launch.py`
- `python -m ruff check tests/test_launch.py`

Closes #4102

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -963,7 +963,7 @@ def launch_command_parser(subparsers=None):
 
     parallelism_config_args.add_argument(
         "--parallelism_config_sp_seq_length_is_variable",
-        type=bool,
+        type=lambda value: str_to_bool(value, to_bool=True),
         default=True,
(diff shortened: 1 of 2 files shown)


lxingy3:
Still relevant. I rechecked current `main`, and this argument still uses `type=bool`, so the `False` spelling remains truthy. The PR is still mergeable, and I reran the two focused regression tests on the branch with Python 3.14: both passed. I would like to keep this active.

PR #4110, opened 2026-07-08
Add per-module fp32 precision for norms under FSDP2

winglian:
# What does this PR do?

We had a report that Arcee's AFMoE had poor convergence when fine-tuning using FSDP. This was triaged to the norms getting cast to bf16. 

Expose FullyShardedDataParallelPlugin.fp32_modules (+ env var override and fp32_modules_auto_detect) so callers can pre-shard selected modules (e.g. RMSNorm/LayerNorm) with their own fp32 MixedPrecisionPolicy before fsdp2_prepare_model's bottom-up class-wrap loop, instead of having them folded into the enclosing decoder layer's lower-precision flat-param group. Auto-detects the transformers `_keep_in_fp32_modules_strict` signal by default. Handles the final-norm/lm_head tail carve-out and raises a clear error instead of silently mis-casting meta-device params when cpu_ram_efficient_loading isn't enabled.

Fix 4 pre-existing FSDP2 unit tests broken by the new attribute access on bare Mock() plugin objects, and add CPU-only coverage for the pattern matcher, the auto-detect union, meta-device handling, and the fsdp2_prepare_model wrapping/tail-carve-out behavior.

@SunMarc 

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/src/accelerate/utils/dataclasses.py
+++ b/src/accelerate/utils/dataclasses.py
@@ -1618,6 +1618,18 @@ class FullyShardedDataParallelPlugin:
         ignored_modules (`Optional[Union[Iterable[torch.nn.Module], str]]`, defaults to `None`):
             A list of modules to ignore when wrapping with FSDP. When passing a string, will match the modules by name
             using regex fullmatch. If `fsdp_version` is set to 2, the modules are converted to parameters and used.
+        fp32_modules (`Optional[List[str]]`, defaults to `None`):
+            A list of class-name patterns (e.g. norm layers) to pre-shard with their own fp32 `MixedPrecisionPolicy`
+            before the rest of the model is wrapped, so they don't get folded into a lower-precision `fully_shard`
+            group. A pattern without a `.` matches as a suffix of the class name (`"RMSNorm"` matches
+            `LlamaRMSNorm`); a pattern with a `.` must exactly match `f"{cls.__module__}.{cls.__name__}"`.
+            Blank/whitespace-only patterns are ignored. Unioned with `fp32_modules_auto_detect`. Only supported when
+            `fsdp_version` is set to `2`.
+        fp32_modules_auto_detect (`bool`, defaults to `True`):
+            Whether to also union in the modules implied by the model's `_keep_in_fp32_modules_strict` attribute
+            (the 🤗 Transformers signal for the same purpose on the bf16 load path). These are matched as an
+            unanchored substring against each module's fully-qualified name, not as class-name patterns. Set to
+            `False` to only use `fp32_modules`. Only supported when `fsdp_version` is set to `2`.
         state_dict_type (`Union[str, torch.distributed.fsdp.StateDictType]`, defaults to `'FULL_STATE_DICT'`):
             State dict type to use. If a string, it must be one of `full_state_dict`, `local_state_dict`, or
             `sharded_state_dict`.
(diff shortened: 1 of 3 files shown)


src/accelerate/utils/dataclasses.py
```diff
@@ -1720,6 +1732,22 @@ class FullyShardedDataParallelPlugin:
...
+        },
+    )
+    fp32_modules_auto_detect: Optional[bool] = field(
+        default=None,
+        metadata={
+            "help": "Whether to automatically union modules implied by the model's `_keep_in_fp32_modules_strict` "
+            "attribute into the fp32 module set used by `fp32_modules`. Defaults to `True`. Only supported when "
+            "`fsdp_version` is set to 2."
+        },
+    )
```
SunMarc, line 1750: let's not have two args. maybe we can do that automatically when fp32_modules is None. If the user set it to [], then we don't update anything.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -741,6 +817,24 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
...
+            param_dtype=torch.float32,
+            reduce_dtype=torch.float32,
+            output_dtype=fsdp2_kwargs["mp_policy"].param_dtype,
+        )
+        for name, module in get_module_children_bottom_up(model, return_fqns=True)[:-1]:
+            if isinstance(module, FSDPModule):
+                continue
+            if not _module_selected_for_fp32(name, module, fp32_class_patterns, fp32_name_patterns):
+                continue
+            _upcast_module_to_fp32(module, fsdp2_plugin.cpu_ram_efficient_loading)
```
SunMarc, line 835: just to be sure, this should be a no-op for the modules thatthat were kept in fp32 no ?

src/accelerate/utils/fsdp_utils.py
```diff
@@ -624,6 +624,82 @@ def fsdp2_apply_ac(accelerator, model: torch.nn.Module):
...
+            "their real values. Either enable `FullyShardedDataParallelPlugin(cpu_ram_efficient_loading=True)` "
+            "(accelerate's supported meta-device reload path) or materialize the model before calling "
+            "`accelerator.prepare()`."
+        )
+    for param in params:
+        if param.is_floating_point() and param.dtype != torch.float32:
+            param.data = param.data.to(torch.float32)
+    for buffer in buffers:
+        if buffer.is_floating_point() and buffer.dtype != torch.float32:
+            buffer.data = buffer.data.to(torch.float32)
```
SunMarc, line 700: same strcuture, simplify a bit

src/accelerate/utils/dataclasses.py
```diff
@@ -1618,6 +1618,18 @@ class FullyShardedDataParallelPlugin:
         ignored_modules (`Optional[Union[Iterable[torch.nn.Module], str]]`, defaults to `None`):
             A list of modules to ignore when wrapping with FSDP. When passing a string, will match the modules by name
             using regex fullmatch. If `fsdp_version` is set to 2, the modules are converted to parameters and used.
+        fp32_modules (`Optional[List[str]]`, defaults to `None`):
+            A list of class-name patterns (e.g. norm layers) to pre-shard with their own fp32 `MixedPrecisionPolicy`
+            before the rest of the model is wrapped, so they don't get folded into a lower-precision `fully_shard`
+            group. A pattern without a `.` matches as a suffix of the class name (`"RMSNorm"` matches
+            `LlamaRMSNorm`); a pattern with a `.` must exactly match `f"{cls.__module__}.{cls.__name__}"`.
```
SunMarc, line 1625: I think it is better to just keep name_pattern for now instead of also introducing class pattern. this is more flexible and it matches what we have in transformers. Also, I don't think much ppl will actually set those.

src/accelerate/utils/fsdp_utils.py
```diff
@@ -741,6 +817,24 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
         if hasattr(model, "tie_weights"):
             model.tie_weights()
 
+    # Pre-shard fp32-selected modules (e.g. norms) with their own fp32 `MixedPrecisionPolicy` before the
+    # class-wrap loop below. Later wraps already skip `FSDPModule` instances, so this also carves a
+    # matching final norm out of the tail group further down.
+    fp32_class_patterns, fp32_name_patterns = resolve_fp32_module_patterns(fsdp2_plugin, model)
```
SunMarc, line 823: as i said maybe just need the name_patterns for now

tests/fsdp/test_fsdp.py
```diff
@@ -690,6 +698,340 @@ def forward(self, x):
         assert getattr(model.head, "_compiled_call_impl", None) is not None
 
 
+class _FakeRMSNorm(torch.nn.Module):
```
SunMarc, line 701: instead of those mocked tests, maybe let's just have one nice integration tests or update existing tests to check that some layers are correctly kept in fp32. This will be better than those numerous tests

SunMarc (commented):
Thanks, left some comments

PR #4119, opened 2026-07-16
Fix FSDP sharded checkpoint path resolution

HaomingSong:
# What does this PR do?

This PR fixes how the FSDP sharded checkpoint loaders distinguish between a checkpoint root directory and an already-resolved PyTorch Distributed Checkpoint shard directory.

Current `load_fsdp_model` and `load_fsdp_optimizer` searched for `FSDP_MODEL_NAME` or `OPTIMIZER_NAME` anywhere in the full `input_dir` string. 
https://github.com/huggingface/accelerate/blob/665444ceb62211f2b410d0d0fdb4bc013c5effdf/src/accelerate/utils/fsdp_utils.py#L308-L312
https://github.com/huggingface/accelerate/blob/665444ceb62211f2b410d0d0fdb4bc013c5effdf/src/accelerate/utils/fsdp_utils.py#L214-L218
If an unrelated parent directory contained one of these markers, the loader incorrectly treated the checkpoint root as the shard directory and attempted to read `.metadata` one level too high.

### Failure example

A valid optimizer checkpoint layout:

```text
/tmp/accelerate_optimizer_parent/epoch_0/
└── optimizer_0/
    └── .metadata
```

And current logic evaluated the full path:

```python
input_dir = "/tmp/accelerate_optimizer_parent/epoch_0"

ckpt_dir = (
    os.path.join(input_dir, "optimizer_0")
    if "optimizer" not in input_dir
    else input_dir
)
```

Because an ancestor directory contains `optimizer`, the resolved directory was incorrect:

```text
input_dir:    /tmp/accelerate_optimizer_parent/epoch_0
expected_dir: /tmp/accelerate_optimizer_parent/epoch_0/optimizer_0
resolved_dir: /tmp/accelerate_optimizer_parent/epoch_0
```

PyTorch Distributed Checkpoint consequently searched for metadata in the checkpoint root instead of the optimizer shard directory:

```text
  File "<stdin>", line 25, in <module>
  File ".../torch/distributed/checkpoint/filesystem.py", line 934, in read_metadata
    with self.fs.create_stream(path, "rb") as metadata_file:
  File ".../torch/distributed/checkpoint/filesystem.py", line 521, in create_stream
    with path.open(mode) as stream:
  File ".../pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
FileNotFoundError: [Errno 2] No such file or directory:
'/tmp/accelerate_optimizer_parent/epoch_0/.metadata'
```

The same false-positive behavior could affect model checkpoints when an ancestor directory contained `pytorch_model_fsdp`.

### Fix

The purpose of this fix is to preserve both supported input forms:

- A checkpoint root, such as `epoch_0`, under which the loader should find `pytorch_model_fsdp_0` or `optimizer_0`.
- An exact shard directory, such as `epoch_0/optimizer_0`, which should be used directly.

Directory detection now compares only the final path component against the exact indexed shard directory name. Parent directory names no longer affect checkpoint resolution.

### Tests

The existing FSDP checkpointing integration test now writes checkpoints beneath a parent directory containing both `optimizer` and `pytorch_model_fsdp`. It exercises saving and resuming model and optimizer state through the affected paths.

Local validation:

- `ruff check .`
- `ruff format --check .`
- `python -m compileall -q src/accelerate/utils/fsdp_utils.py tests/fsdp/test_fsdp.py`
- Both affected FSDP integration-test nodes collected successfully. They were skipped locally because the environment does not expose the required multiple GPU devices; GPU CI will exercise them.

No new dependencies or public API changes are introduced.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc


SunMarc (approved):
Thanks for the fix ! Can you fix the merge conflict @HaomingSong  ?

HaomingSong:
Thanks for reviewing @SunMarc! I have resolved the merge conflict, and the previous CI failure seems not caused by this PR. Coulud you please take another look :)

PR #3850, opened 2025-11-25
Allow non-Tensor values in a batch with `dispatch_batches=True`

tomaarsen:
Resolves #3849 

# What does this PR do?
* Allow non-Tensor values in a batch with `dispatch_batches=True`, matching behaviour for when `dispatch_batches=False` or when `accelerate` is not used

## Details
Rerunning the script from #3849 now also gives this for the previously broken case:

### Accelerator, with `IterableDataset`

```
Batch:
query_input_ids: <class 'torch.Tensor'> with shape torch.Size([4, 13])
query_token_type_ids: <class 'torch.Tensor'> with shape torch.Size([4, 13])
query_attention_mask: <class 'torch.Tensor'> with shape torch.Size([4, 13])
query_str_parameter: <class 'str'> parameter_value
query_bool_parameter: <class 'bool'> True
query_str_list: <class 'list'> ['list_item_1', 'list_item_2']
answer_input_ids: <class 'torch.Tensor'> with shape torch.Size([4, 328])
answer_token_type_ids: <class 'torch.Tensor'> with shape torch.Size([4, 328])
answer_attention_mask: <class 'torch.Tensor'> with shape torch.Size([4, 328])
answer_str_parameter: <class 'str'> parameter_value
answer_bool_parameter: <class 'bool'> True
answer_str_list: <class 'list'> ['list_item_1', 'list_item_2']

Batch:
query_input_ids: <class 'torch.Tensor'> with shape torch.Size([1, 11])
query_token_type_ids: <class 'torch.Tensor'> with shape torch.Size([1, 11])
query_attention_mask: <class 'torch.Tensor'> with shape torch.Size([1, 11])
query_str_parameter: <class 'str'> parameter_value
query_bool_parameter: <class 'bool'> True
query_str_list: <class 'list'> ['list_item_1', 'list_item_2']
answer_input_ids: <class 'torch.Tensor'> with shape torch.Size([1, 164])
answer_token_type_ids: <class 'torch.Tensor'> with shape torch.Size([1, 164])
answer_attention_mask: <class 'torch.Tensor'> with shape torch.Size([1, 164])
answer_str_parameter: <class 'str'> parameter_value
answer_bool_parameter: <class 'bool'> True
answer_str_list: <class 'list'> ['list_item_1', 'list_item_2']
```
As expected!

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
[...]

PR #4120, opened 2026-07-16
Fix concatenate() for non-tensor batch fields across multiple batches

### LINKED ISSUE #4111: concatenate() still raises TypeError on non-tensor batch fields when num_processes > 1
### Context

#3849 reported that `DataLoaderDispatcher` (used when `dispatch_batches=True`, the default for an `IterableDataset`) crashed on non-tensor batch values (str/bool/list) returned by a data collator. 

#3850 fixed the single-process case by adding a `len(data) == 1` short-circuit in `concatenate()`:
```python
elif isinstance(data[0], torch.Tensor):
    return torch.cat(data, dim=dim)
elif isinstance(data, (tuple, list)) and len(data) == 1:
    return data[0]
else:
    raise TypeError(f"Can only concatenate tensors but got {type(data[0])}")
```

### The remaining bug

`DataLoaderDispatcher._fetch_batches` gathers exactly `num_processes` raw batches before calling `concatenate`:
```python
for _ in range(self.state.num_processes):
    self._update_state_dict()
    batches.append(next(iterator))
batch = concatenate(batches, dim=0)
```

As soon as `num_processes > 1` and a batch field isn't a tensor, `concatenate()` still raises the exact same `TypeError` as before #3850.

The PR review discussion already covers this exact gap, but it's easy to miss since it's buried in review comments on a merged PR rather than tracked as an open issue: https://github.com/huggingface/accelerate/pull/3850#discussion_r2560098952
>  @SunMarc: "This is a fix that only works when we have one gpu (only one batch is passed in concatenate). The issue happens when we have multi-gpu..."
>
> @tomaarsen: "The suggestion here works fine with single-process, but the len(data) is simply equal to the number of processes. In short: it'll always fail in multi-process settings."
>
> @tomaarsen (final): "Agreed, it's a step in the right direction, even though multi-process support is not possible... Not supporting IterableDataset + MultiGPU is a bit annoying."

So this is a confirmed, intentional trade-off, not an oversight. I'm opening this mainly so it's trackable/discoverable outside the PR thread.

### Minimal repro
```python
from accelerate.utils.operations import concatenate
import torch

data = [
    {"input_ids": torch.tensor([[1, 2]]), "label": True},
    {"input_ids": torch.tensor([[3, 4]]), "label": False},
]
concatenate(data)
# TypeError: Can only concatenate tensors but got <class 'bool'>
```

### Ask
@SunMarc mentioned in review: "We can add in the docstring that if we receive only one batch of data, we will return it as it." The current docstring says "If there is only a single batch of data, it is returned as-is," but doesn't spell out that non-tensor fields remain broken for num_processes > 1 on any Accelerate version. Could that be made explicit, so downstream library authors don't assume #3850 was a general fix? 

We hit exactly that assumption gap in `trl` KTOTrainer, where a collator returned a bool field as a plain list, fine for `num_processes == 1` (where CI happened to pass on latest Accelerate), broken for real multi-process streaming eval on any Accelerate version. We're fixing it on our end by making that field a tensor, but wanted to flag the gap here in case it affects other libraries relying on `dispatch_batches=True` with non-tensor collator outputs.

Kaif10:
Fixes #4111.

## What
`accelerate.utils.operations.concatenate` raised `TypeError: Can only concatenate tensors` whenever a batch field wasn't a tensor (e.g. `bool`, `str`, `list`) and more than one batch was concatenated — the path hit by `DataLoaderDispatcher._fetch_batches` when `num_processes > 1`.

#3850 fixed only the single-batch case (`len(data) == 1` returns `data[0]`); as discussed there by @SunMarc and @tomaarsen, multi-process non-tensor support was left as a known gap ("it'll always fail in multi-process settings"). This PR closes that gap and expands the docstring @SunMarc asked for.

## How
- **List-valued non-tensor fields** (one entry per sample) are chained along the batch dimension into a single list — the natural analogue of `torch.cat` for per-sample metadata (this is what collate_fns emit; cf. huggingface/trl#6325).
- **Scalar non-tensor fields shared by a whole batch** are collected into a list with one value per batch.
- Tensor, nested list/tuple/dict, and single-batch behavior are unchanged.
- Docstring now documents the non-tensor semantics explicitly.

## Repro (before)
```python
from accelerate.utils.operations import concatenate
import torch
concatenate([{"input_ids": torch.tensor([[1,2]]), "label": True},
             {"input_ids": torch.tensor([[3,4]]), "label": False}])
# TypeError: Can only concatenate tensors but got <class 'bool'>
```
After: `{'input_ids': tensor([[1,2],[3,4]]), 'label': [True, False]}`.

## Tests
Extended `tests/test_utils.py::test_concatenate` (where #3850 added its tests) with multi-batch non-tensor cases: scalar bool/str, the #4111 repro, list-of-strings chaining, uneven batch sizes, nested dicts, and tuple-structure preservation. New tests fail on `main` and pass with this change; the rest of `test_utils.py` is unaffected.

---
This change was drafted with the help of an AI coding assistant and verified locally on a CPU-only setup (torch 2.13.0+cpu): the repro and the full `test_utils.py::test_concatenate` suite pass, and `ruff format --check` is clean on the changed files.

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -719,7 +719,15 @@ def slice_tensors(data, tensor_slice, process_index=None, num_processes=None):
 def concatenate(data, dim=0):
     """
     Recursively concatenate the tensors in a nested list/tuple/dictionary of lists of tensors with the same shape.
-    If there is only a single batch of data, it is returned as-is.
+
+    Non-tensor values are only supported along the first dimension (`dim` is ignored for them) and are handled as
+    follows:
+
+    - If there is only a single batch of data, it is returned as-is, even if it contains non-tensor values.
+    - Lists of non-tensor values (e.g. a list of strings, holding one value per sample in the batch) are chained
+      into a single list, mimicking a concatenation along the batch dimension.
+    - Any other non-tensor value (e.g. a string or a boolean shared by a whole batch) is collected into a list
+      holding the value of each batch.
 
     Args:
         data (nested list/tuple/dictionary of lists of tensors `torch.Tensor`):
@@ -731,6 +739,12 @@ def concatenate(data, dim=0):
         The same data structure as `data` with all the tensors concatenated.
     """
     if isinstance(data[0], (tuple, list)):
+        if isinstance(data[0], list) and not any(
+            isinstance(sample, (torch.Tensor, tuple, list, Mapping)) for sample in data[0]
+        ):
+            # Lists of non-tensor values hold one value per sample in the batch, so we chain them into a single
+            # list, mimicking a concatenation along the batch dimension.
+            return [sample for batch in data for sample in batch]
         return honor_type(data[0], (concatenate([d[i] for d in data], dim=dim) for i in range(len(data[0]))))
     elif isinstance(data[0], Mapping):
         return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
@@ -738,6 +752,10 @@ def concatenate(data, dim=0):
         return torch.cat(data, dim=dim)
     elif isinstance(data, (tuple, list)) and len(data) == 1:
         return data[0]
+    elif isinstance(data, (tuple, list)):
+        # Multiple batches of non-tensor values (e.g. a string or a boolean shared by a whole batch) cannot be
+        # concatenated, so we collect the value of each batch into a list instead.
+        return list(data)
     else:
         raise TypeError(f"Can only concatenate tensors but got {type(data[0])}")
 
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -726,8 +726,9 @@ def concatenate(data, dim=0):
     - If there is only a single batch of data, it is returned as-is, even if it contains non-tensor values.
     - Lists of non-tensor values (e.g. a list of strings, holding one value per sample in the batch) are chained
       into a single list, mimicking a concatenation along the batch dimension.
-    - Any other non-tensor value (e.g. a string or a boolean shared by a whole batch) is collected into a list
-      holding the value of each batch.
+    - Any other scalar non-tensor value (e.g. a string or a boolean shared by a whole batch) is collected into a
+      list holding the value of each batch. Non-scalar, non-tensor values (e.g. numpy arrays) are not concatenated
+      and raise, as before.
 
     Args:
         data (nested list/tuple/dictionary of lists of tensors `torch.Tensor`):
@@ -752,12 +753,15 @@ def concatenate(data, dim=0):
         return torch.cat(data, dim=dim)
     elif isinstance(data, (tuple, list)) and len(data) == 1:
         return data[0]
-    elif isinstance(data, (tuple, list)):
-        # Multiple batches of non-tensor values (e.g. a string or a boolean shared by a whole batch) cannot be
-        # concatenated, so we collect the value of each batch into a list instead.
+    elif isinstance(data, (tuple, list)) and all(
+        isinstance(d, (str, int, float, bool, bytes, type(None))) for d in data
+    ):
+        # Multiple batches of a scalar non-tensor value (e.g. a string or a boolean shared by a whole batch)
+        # cannot be concatenated along the batch dim, so we collect the value of each batch into a list instead.
+        # Non-scalar non-tensor values (e.g. numpy arrays) fall through and raise, preserving prior behavior.
         return list(data)
     else:
-        raise TypeError(f"Can only concatenate a list or tuple of batches but got {type(data)}")
+        raise TypeError(f"Can only concatenate tensors but got {type(data[0])}")
 
 
 class CannotPadNestedTensorWarning(UserWarning):
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -757,7 +757,7 @@ def concatenate(data, dim=0):
         # concatenated, so we collect the value of each batch into a list instead.
         return list(data)
     else:
-        raise TypeError(f"Can only concatenate tensors but got {type(data[0])}")
+        raise TypeError(f"Can only concatenate a list or tuple of batches but got {type(data)}")
 
 
 class CannotPadNestedTensorWarning(UserWarning):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -491,13 +491,55 @@ class UtilsTester(unittest.TestCase):
         assert result["a"].shape == torch.Size([4, 3])
         assert result["b"].shape == torch.Size([4, 4])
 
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
Verified the list-vs-tuple distinction is the right call, not just a heuristic that happens to work: the new check only fires for `isinstance(data[0], list)` (not `tuple`), and the added test explicitly locks in that tuples keep the old structural, per-position recursion (`data = [(tensor, "s1"), (tensor, "s2")]` still concatenates position 0 as tensors and collects position 1 as a list), while a list of plain scalars gets chained as per-sample batch data instead. That's a real, useful distinction, since a tuple like `(features, labels)` per batch item genuinely means something different from a list of per-sample string labels.

Traced through the two new branches: for a list where every element is a non-container (the `not any(isinstance(sample, (Tensor, tuple, list, Mapping))...)` guard), chaining `[sample for batch in data for sample in batch]` correctly flattens across batches. For the remaining top-level case (multiple batches of a single scalar each, e.g. one bool per batch), `list(data)` replaces the old unconditional `TypeError`. Test coverage includes the linked issue's exact repro, uneven batch sizes (the drop_last=False remainder case), and nested-dict handling. LGTM.

src/accelerate/utils/operations.py
```diff
@@ -731,13 +739,23 @@ def concatenate(data, dim=0):
...
     elif isinstance(data[0], Mapping):
         return type(data[0])({k: concatenate([d[k] for d in data], dim=dim) for k in data[0].keys()})
     elif isinstance(data[0], torch.Tensor):
         return torch.cat(data, dim=dim)
     elif isinstance(data, (tuple, list)) and len(data) == 1:
         return data[0]
+    elif isinstance(data, (tuple, list)):
+        # Multiple batches of non-tensor values (e.g. a string or a boolean shared by a whole batch) cannot be
+        # concatenated, so we collect the value of each batch into a list instead.
+        return list(data)
```
Sanjays2402, line 762: with this catch-all the TypeError below is unreachable: anything that gets past `data[0]` is already a list/tuple, so unsupported values (a numpy array batch, say) now silently come back as `list(data)` instead of erroring. worth keeping the raise for types you actually do not want to collect per-batch?

ErenAta16: You're right, and it's my miss too. My approval up-thread checked the list-vs-tuple distinction on the first new branch and didn't scrutinize this catch-all, which is the half that actually widens behavior.

Confirmed it against `main` and the branch:

```
main:   concatenate([np.array([1,2]), np.array([3,4])])         -> TypeError: Can only concatenate tensors but got <class 'numpy.ndarray'>
main:   concatenate([{"x": np.array([1,2])}, {"x": np.array([3,4])}]) -> TypeError (same)
branch: concatenate([np.array([1,2]), np.array([3,4])])         -> [array([1, 2]), array([3, 4])]
branch: dict field "x"                                          -> [array([1, 2]), array([3, 4])]
```

So a numpy-array field silently comes back un-concatenated as a Python list of arrays, where it used to raise. That's the worst kind of change for this function, because a numpy array is exactly the case where the caller almost certainly wanted concatenation (or at least a loud error), unlike a bare `str`/`bool` that genuinely can't be concatenated along the batch dim. The gather path would then hand downstream code a list of arrays instead of one array, and nothing signals it.

You're also right that the `raise` is now effectively dead for the normal calling convention. It's still reachable, but only if `data` itself is a bare ndarray rather than a list of batches:

```
branch: concatenate(np.array([1,2,3])) -> TypeError: ... got <class 'numpy.int64'>
```

which isn't how this function is called in practice, so the safety net is gone for real inputs.

The fix that keeps the feature and restores the net is to narrow the catch-all to the scalar-ish types it's actually meant to collect, rather than "any non-tensor":

```python
elif isinstance(data, (tuple, list)) and all(
    isinstance(d, (str, int, float, bool, bytes, type(None))) for d in data
):
    return list(data)
```

Verified that keeps the string/bool feature and brings back the ndarray error:

```
narrowed: concatenate([{"c": "s1"}, {"c": "s2"}])            -> {'c': ['s1', 's2']}      (feature preserved)
narrowed: concatenate([np.array([1,2]), np.array([3,4])])   -> TypeError restored
```

If the intent is genuinely to also collect array-like non-tensors per batch, then it should be deliberate and documented rather than falling out of a broad `isinstance(data, (tuple, list))`, and probably worth a test asserting the numpy case does whatever is chosen. Either way the current version changes numpy handling as a side effect, which I don't think is intended. Thanks for catching it.

Kaif10: Good catch — you're right. Since `concatenate` is always called with a list/tuple of batches (both the top-level callers and its own recursion pass a list), the non-tensor collection branch above now handles every list/tuple case, so the old `raise TypeError(... type(data[0]) ...)` was no longer reachable.

I've repointed that guard to the case it should actually protect: `data` itself not being a list/tuple of batches (e.g. a bare string or array passed by mistake), keyed on `type(data)`. That keeps a meaningful, reachable error instead of dead code (and avoids silently returning `None` on misuse). Pushed in c19fdd0e.

Kaif10: Great catch, and thank you for the thorough verification — you're completely right. Silently turning a batch of numpy arrays into a list of arrays is the wrong behavior; that's exactly the case where the caller wanted concatenation or a loud error, unlike a bare `str`/`bool`.

I've adopted your fix: the collection branch is now narrowed to genuinely scalar types, so non-scalar non-tensor values (numpy arrays, etc.) fall through to the `TypeError` again — restoring the prior behavior for them while keeping the string/bool feature:

```python
elif isinstance(data, (tuple, list)) and all(
    isinstance(d, (str, int, float, bool, bytes, type(None))) for d in data
):
    return list(data)
```

Added a regression test asserting a numpy-array field still raises `TypeError`, and updated the docstring to state that non-scalar non-tensor values are not collected. Verified: `str`/`bool` fields still collect, tensors still concatenate, and both the ndarray-field and bare-ndarray-list cases raise again. Pushed in b8f3c602.

Kaif10:
@SunMarc gentle ping on this one — it closes the multi-process gap you and @tomaarsen identified in #3850, where non-tensor batch fields still raised for `num_processes > 1`.

It's had two rounds of community review since: the collection branch is now narrowed to scalar types, so numpy arrays continue to raise rather than being silently returned un-concatenated, and the docstring documents the non-tensor semantics you asked for. Tests cover the #4111 repro, list-chaining, uneven batch sizes, nested dicts, and the numpy case.

Happy to rebase or adjust anything if you'd like it shaped differently.

Kaif10:
@SunMarc Gentle ping in case this slipped through — small fix for `concatenate()` dropping non-tensor batch fields after the first batch, with a community approval already on it. Happy to rebase or adjust if anything's needed.

PR #4135, opened 2026-07-24
Fix typos and a broken kwarg in the sequence parallelism guide

Savi-Swar:
A few small fixes to `docs/source/concept_guides/sequence_parallelism.md`:

- "has has to be divisible" → "has to be divisible"
- "divisibilty" → "divisibility"
- In the first `ParallelismConfig` example, `sp_seq_length_is_variable: true` uses a YAML-style `:` and a lowercase `true` inside a Python call, so copy-pasting the snippet raises a `SyntaxError`. Changed it to `sp_seq_length_is_variable=True`, which matches the working example further down the same page.

Docs only, no code changes.

### DIFF
--- a/docs/source/concept_guides/sequence_parallelism.md
+++ b/docs/source/concept_guides/sequence_parallelism.md
@@ -35,8 +35,8 @@ The following articles go into a very detailed explanation of the differences be
 - https://huggingface.co/blog/exploding-gradients/ulysses-ring-attention
 
 A quick summary adapting from one of the articles:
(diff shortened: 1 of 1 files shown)


Savi-Swar:
Still valid on main: the doubled word and the typo are at lines 38-39 of sequence_parallelism.md, and the Python example still has the YAML-style sp_seq_length_is_variable: true, which is a SyntaxError if pasted. Happy to rebase if needed.

Issue #3861, opened 2025-11-28
NCCL timeout for validation


WenxiongZhou:
### System Info

```Shell
accelerate 1.12.0
```

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

```def main()
  ......
   # init accelerator config
    accelerator_project_config = ProjectConfiguration(project_dir=args.output_dir, logging_dir=logging_dir)
    timeout_kwargs = InitProcessGroupKwargs(backend="nccl", timeout=timedelta(seconds=1800))

    accelerator = Accelerator(
        gradient_accumulation_steps=args.gradient_accumulation_steps,
        mixed_precision=args.mixed_precision,
        log_with=args.report_to,
        project_config=accelerator_project_config,
        kwargs_handlers=[timeout_kwargs]
    )
  ... ...
    for epoch in range(first_epoch, args.num_train_epochs):
        unet.train()
        train_loss = 0.0
        for step, batch in enumerate(train_dataloader):
            ... ...
        accelerator.wait_for_everyone()
        if accelerator.is_main_process:
            if (
                epoch % args.validation_epochs == 0
            ):
                accelerator.wait_for_everyone()

                # The models need unwrapping because for compatibility in distributed training mode.
                pipeline = DepthPipeline.from_pretrained(
                    args.pretrained_model_name_or_path,
                    unet=unwrap_model(unet),
                    text_encoder=unwrap_model(text_encoder),
                    vae=unwrap_model(vae),
                    torch_dtype=weight_dtype,
                    safety_checker=None
                )

                metrics = log_validation(
                    pipeline,
                    args,
                    accelerator,
                    generator,
                    val_dataloader,
                    accelerator.device,
                    global_step
                )

                if metrics["rmse"] < best_rmse:
                    best_rmse = metrics["rmse"]
                    pipeline.save_pretrained(os.path.join(args.output_dir, "best"))

                del pipeline
                torch.cuda.empty_cache()
        accelerator.wait_for_everyone()
```

### Expected behavior

I want to perform validation after each epoch of training, but due to the long validation time (over 10 minutes), NCCL timeout (after 10 minutes). Although I set the NCCL timeout to 30 minutes, it doesn't take effect.

[...]

Issue #1089, opened 2023-02-16
how to cleanly exit when using `accelerate launch`


vladmandic:
### System Info

```Shell
sorry for the silly question, i've searched and cannot find the answer

i'm launching python script using
`accelerate launch script.py`

script itself runs gradioapp web server - and when gradio queues are enabled, there is no clean way to shutdown web server. which means the only way to exit is `os._exit(0)`.

but if i do that, i always get a long traceback from `accelerate` itself - how to tell accelerate that exit is fine and don't print traceback?


note: accelerate 0.16 added `--quiet` flag, it doesn't change the behavior

environment:
- accelerate 0.16.0
- python 3.10.6
- torch 2.0.0.dev20230211+cu118 CUDA: 11.8 cuDNN: 8.7.0
- os: ubuntu 22.04 running in wsl2 on windows 11


╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/vlado/.local/bin/accelerate:8 in <module>                                                  │
│                                                                                                  │
│   5 from accelerate.commands.accelerate_cli import main                                          │
│   6 if __name__ == '__main__':                                                                   │
│   7 │   sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])                         │
│ ❱ 8 │   sys.exit(main())                                                                         │
│   9                                                                                              │
│                                                                                                  │
│ /home/vlado/.local/lib/python3.10/site-packages/accelerate/commands/accelerate_cli.py:45 in main │
│                                                                                                  │
│   42 │   │   exit(1)                                                                             │
│   43 │                                                                                           │
│   44 │   # Run                                                                                   │
│ ❱ 45 │   args.func(args)                                                                         │
│   46                                                                                             │
│   47                                                                                             │
│   48 if __name__ == "__main__":                                                                  │
│                                                                                                  │
│ /home/vlado/.local/lib/python3.10/site-packages/accelerate/commands/launch.py:1097 in            │
│ launch_command                                                                                   │
│                                                                                                  │
│   1094 │   elif defaults is not None and defaults.compute_environment == ComputeEnvironment.AMA  │
│   1095 │   │   sagemaker_launcher(defaults, args)                                                │
│   1096 │   else:                                                                                 │
│ ❱ 1097 │   │   simple_launcher(args)                                                             │
│   1098                                                                                           │
│   1099                                                                                           │
│   1100 def main():                                                                               │
│                                                                                                  │
[...]

Issue #314, opened 2022-04-12
About Timeout when use Multi-gpu training


macheng6:
When I used the single-node multi-GPU mode to train, a timeout error was reported. The strange thing is that for the first few epochs, the code works fine. This error was reported after the end of a step eval in the middle.

The reported error message is：

[E ProcessGroupNCCL.cpp:587] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(OpType=BROADCAST, Timeout(ms)=1800000) ran for 1808499 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:587] [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(OpType=BROADCAST, Timeout(ms)=1800000) ran for 1808493 milliseconds before timing out.

[E ProcessGroupNCCL.cpp:341] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data. To avoid this inconsistency, we are taking the entire process down.

terminate called after throwing an instance of 'std::runtime_error'
what(): [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(OpType=BROADCAST, Timeout(ms)=1800000) ran for 1808493 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:341] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data. To avoid this inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'

sgugger:
The torch distributed launcher contains a timeout close at 30 minutes that will cancel the processes if met. It seems some instruction in your evaluation loop takes longer than that, which is why you are getting this error. You can increase this timeout by using a `InitProcessGroupKwargs` (see [here](https://github.com/huggingface/accelerate/blob/8595834292b1e7e0543c6d75eb1857da25062b6d/src/accelerate/kwargs_handlers.py#L90) and passing it to the `Accelerator` object under the argument `kwargs_handlers`.

macheng6:
Thanks.&nbsp;




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "huggingface/accelerate"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年4月12日(星期二) 晚上7:52
***@***.***&gt;;
***@***.******@***.***&gt;;
[...]

Issue #223, opened 2021-12-31
Multi-GPU Training+Single-GPU Eval runs into time out


cronoik:
Hi everyone,

we run into a timeout when we evaluate for more than 30 minutes on a single GPU. Is there a way to tell the other GPU to wait until the main GPU completes the evaluation?
```
[E ProcessGroupNCCL.cpp:587] [Rank 2] Watchdog caught collective operation timeout: WorkNCCL(OpType=BROADCAST, Timeout(ms)=1800000) ran for 1802157 milliseconds before timing out. 
Traceback (most recent call last): 
Traceback (most recent call last): 
 File "scripts/training_test.py", line 172, in <module> 
 File "scripts/training_test.py", line 172, in <module> 
   main(args)    
main(args) File "scripts/training_test.py", line 167, in main 
 
 File "scripts/training_test.py", line 167, in main 
   train(args.config) 
 File "scripts/training_test.py", line 148, in train 
   train(args.config) 
 File "scripts/training_test.py", line 148, in train 
   trainer.train_pipeline() 
 File "/home/azureuser/mytrainer.py", line 182, in train_pipeline 
   trainer.train_pipeline() 
 File "/home/azureuser/mytrainer.py", line 182, in train_pipeline 
       for step, batch in enumerate(pbar):for step, batch in enumerate(pbar): 
 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/tqdm/std.py", line 1168, in __iter__ 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/tqdm/std.py", line 1168, in __iter__ 
       for obj in iterable:for obj in iterable: 
 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/accelerate/data_loader.py", line 301, in __iter__ 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/accelerate/data_loader.py", line 301, in __iter__ 
       synchronize_rng_states(self.rng_types, self.generator)synchronize_rng_states(self.rng_types, self.generator) 
 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/accelerate/utils.py", line 110, in synchronize_rng_states 
 File "/anaconda/envs/cronoik_test/lib/python3.8/site-packages/accelerate/utils.py", line 110, in synchronize_rng_states 
[...]

PR #4139, opened 2026-07-28
[launch] Report one actionable diagnostic when distributed ranks fail (and make --quiet work for multi-GPU)

mosafariuk:
## What does this PR do?

When a rank dies during `accelerate launch --multi_gpu` (or FSDP / single-node DeepSpeed), `torch.distributed.elastic` hands accelerate a fully structured `ChildFailedError` — one `ProcessFailure` per rank with exitcode, signal, pid, timestamp and error-file path. Today the launchers catch it with a generic `except Exception` and re-raise unchanged, so that structure reaches the user as raw traceback text underneath the per-rank log flood — and torchelastic's own "Root Cause (first observed failure)" heuristic can name a *victim* rank (one that timed out waiting) while the true culprit is the rank that logged nothing.

This PR reads that object and prints **one bordered diagnostic block**: the first observed failure (rank, pid, exitcode, signal) explicitly labelled as torchelastic's earliest-timestamp *heuristic*, a signal-class breakdown separating watchdog-style aborts from ranks the elastic agent tore down afterwards ("collateral, not causes"), pointers to every preserved per-rank artifact, and the exact re-run command for per-rank capture and Flight Recorder. It also:

* forwards torchrun's existing `--local_ranks_filter` / `--redirects` (accelerate already forwards `--tee` / `--log_dir` through the same `_filter_args` path) — and refuses the lossy configuration: filtering without file capture auto-enables `--tee 3 --log_dir` with a warning, so muted console output is demoted to files, never destroyed;
* makes `--quiet` apply to the multi-GPU path (its help text currently disclaims it, which is part of what #1089 reports); in quiet mode the exit code becomes `128 + signum` (a watchdog SIGABRT yields 134 instead of a lossy 1);
* catches `KeyboardInterrupt` in `simple_launcher`, so Ctrl-C exits 130 instead of dumping accelerate's own traceback — the direct ask in #1089;
* keeps the classifier/formatter in a new pure-function module (`accelerate.utils.error_reporting`) with no `torch` import, unit-testable without GPUs.

The three modes, stated precisely: the default **interprets** the failure (strictly additive — nothing is removed); `--quiet` **silences** the Python-side launcher traceback; `--local_ranks_filter` **mutes the C++-side console flood with preservation** (torchrun's OS-level redirection; per-rank files always written). No mode deletes diagnostic data.

## Verification (real hardware)

* **Hardware validation:** tested end-to-end on a 2x NVIDIA L40 instance (PyTorch 2.12.1+cu130, NCCL 2.27.7) with a genuine NCCL watchdog timeout provoked by freezing one rank while its peer entered an all-reduce.
* **Default mode:** the watchdog killed **rank 0 — the victim waiting on frozen rank 1** — and the diagnostic block classified the SIGABRT as a watchdog abort, displayed the victim caveat over exactly the case it exists for, then re-raised `ChildFailedError` unchanged (no logs masked).
* **`--quiet`:** exit code 134 (128+SIGABRT), zero launcher tracebacks, single diagnostic report.
* **`--local_ranks_filter`:** eliminated the C++ console flood (169 -> 81 lines, zero watchdog lines on console) while the guard auto-enabled `--tee`/`--log_dir`; the full C++ stack traces were verified preserved in per-rank `stderr.log` files.
* **Regression safety:** healthy runs byte-identical (exit 0, zero new output); ordinary Python exceptions keep their full traceback; `tests/test_cli.py` and `tests/test_utils.py` pass.

<details>
<summary><b>Before</b> — raw waterfall on the L40s (excerpt of 169 lines)</summary>

```
[rank0]:[E728 18:15:39.660703688 ProcessGroupNCCL.cpp:2197] [PG ID 0 PG GUID 0(default_pg) Rank 0] Process group watchdog thread terminated with exception: NCCL error in: /pytorch/torch/csrc/distributed/c10d/NCCLUtils.cpp:461, Unknown, NCCL version 2.27.7
ncclInternalError: Internal check failed.
Last error:
Unknown
... (~50 more lines of C++ watchdog / NCCL / teardown output) ...
terminate called after throwing an instance of 'c10::DistBackendError'
  what():  [PG ID 0 PG GUID 0(default_pg) Rank 0] Process group watchdog thread terminated with exception: NCCL error in: /pytorch/torch/csrc/distributed/c10d/NCCLUtils.cpp:461, Unknown, NCCL version 2.27.7
ncclInternalError: Internal check failed.
terminate called recursively
W0728 18:15:41.375000 528 torch/distributed/elastic/multiprocessing/api.py:1014] Sending process 676 closing signal SIGTERM
```
</details>

<details>
<summary><b>After</b> — the diagnostic block this PR adds (default mode, printed before the unchanged re-raise)</summary>

```
accelerate launch: `repro_gpu_crash.py` failed on 2 rank(s)
------------------------------------------------------------------------------
First observed failure (earliest timestamp -- HEURISTIC, not a verdict):
  rank      : 0 (local_rank 0)
  exitcode  : -6 (SIGABRT), pid 675
  error_file: <none written -- see @record note below>

  A rank exiting on SIGABRT is the signature of PyTorch's NCCL watchdog or heartbeat monitor tearing the process down after a collective exceeded its timeout. This is by design and is not catchable in Python: the abort is raised on a C++ watchdog thread, not on the thread that issued the collective.

  The rank that reports a timeout is frequently a VICTIM waiting on a peer. The root-cause rank is often the one that produced no output at all (it hung), so treat 'first observed failure' as a heuristic, not a verdict.

Rank breakdown:
  aborted (watchdog / hard fault)     : 1  -> ranks [0]
  torn down by the launcher           : 1  -> ranks [1]
  raised a Python exception / exited  : 0

  Ranks torn down by SIGTERM/SIGKILL were killed by the elastic agent AFTER
  the first failure. Their tracebacks are collateral, not causes.

Preserved per-rank detail:
  none -- worker entrypoints were not decorated with
  `torch.distributed.elastic.multiprocessing.errors.record`, so no error
  files were written (exit codes and signals only).
  Per-rank stdout/stderr was NOT captured to disk for this run.
  Re-run with:  accelerate launch --tee 3 --log_dir ./accelerate_logs ...

Suggested next steps:
  1. Re-run with per-rank capture and a quiet console:
       accelerate launch --tee 3 --log_dir ./accelerate_logs \
                         --local_ranks_filter 0 ...
  2. Enable Flight Recorder to see which collective each rank was in:
       TORCH_NCCL_TRACE_BUFFER_SIZE=2000 TORCH_NCCL_DUMP_ON_TIMEOUT=1
  3. Look first at the rank that produced NO output.
==============================================================================
  File "/home/ubuntu/.local/bin/accelerate", line 6, in <module>
    sys.exit(main())
```
</details>

## What this deliberately does NOT do

* **No recovery, no retry.** PyTorch core's position (pytorch/pytorch#163546) is that `ProcessGroupNCCL` is not built for recovery. This PR improves *presentation* of a crash the launcher already handles.
* **No claim to catch NCCL watchdog timeouts in Python.** The watchdog aborts from a C++ thread via `std::terminate` (RFC pytorch/pytorch#46874, by design); this PR classifies the resulting SIGABRT at the launcher, where it is observable.
* **No exception swallowing, no data deletion.** The default is strictly additive: print one block, then re-raise the original `ChildFailedError` (same type, same exit semantics). `ACCELERATE_DISTRIBUTED_ERROR_SUMMARY=0` disables the block entirely. If the formatter itself fails for any reason it returns `None` and the original exception propagates exactly as before — a bug in the diagnostics can never hide the real crash.
* **No collectives, barriers or state rescue in the crash path.** (Lightning shipped the opposite once — Lightning-AI/pytorch-lightning#6842 — and removed it after it deadlocked.) The handler runs in the launcher process, formats, and exits or re-raises.

## Scope limitation: DeepSpeed multi-node

DeepSpeed's PDSH/OpenMPI multi-node launchers run through `subprocess.Popen` and expose only the child's return code, so there is no `ChildFailedError` on that path (unless elastic training is enabled); behavior there is unchanged.

## Tests

`tests/test_error_reporting.py` — classifier/formatter units (signal bucketing, minimum-timestamp selection, `-6 -> 134` mapping, empty-input safety, the heuristic caveat, the `None`-fallback guard), plus an in-process harness that reaches the real `except ChildFailedError` clause on zero-GPU CI by monkeypatching `torch.distributed.run.run` — chosen deliberately because a `--cpu` subprocess test dispatches to `simple_launcher` and never reaches the clause. A real watchdog firing is not CI-inducible on hosted runners; the hardware validation above covers it.

## Backwards compatibility

New code executes only on the crash path or when a new flag is passed. Exit codes and exception types are unchanged by default.

Refs #1089. Related reports of this failure mode: #3861, #2183, #314, #223. Precedent for launcher-layer mitigation of a distributed footgun: #2195.

## Who can review?

@muellerzr @SunMarc @BenjaminBossan

### DIFF
--- a/src/accelerate/commands/launch.py
+++ b/src/accelerate/commands/launch.py
@@ -20,6 +20,7 @@ import logging
 import os
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 import torch
@@ -314,6 +315,27 @@ def launch_command_parser(subparsers=None):
         help=(
             "Base directory to use for log files when using torchrun/torch.distributed.run as launcher. "
             "Use with --tee to redirect std streams info log files."
+        ),
+    )
+    distributed_args.add_argument(
+        "-r",
+        "--redirects",
+        type=str,
+        default="0",
+        help=(
+            "Redirect std streams into per-rank log files under --log_dir (e.g. `3` for stdout+stderr on all "
+            "ranks, `0:1,1:2` for per-rank control). Passed through to torch.distributed.run."
+        ),
+    )
+    distributed_args.add_argument(
+        "--local_ranks_filter",
+        "--local-ranks-filter",
+        type=str,
+        default="",
+        help=(
+            "Only echo logs from these local ranks to the console (e.g. `0`). Passed through to "
+            "torch.distributed.run. Per-rank log files are still written in full; if --tee/--log_dir are unset "
+            "they are enabled automatically so that muted output is never lost."
         ),
     )
     distributed_args.add_argument(
@@ -983,11 +1005,68 @@ def launch_command_parser(subparsers=None):
     return parser
 
 
+def _guard_console_filter(args):
+    """Never mute a rank whose output is not being written to disk.
+
+    torchrun routes a filtered rank's console stream to os.devnull when no redirect is
+    configured, which would destroy the muted output instead of demoting it to a file.
+    """
+    if not getattr(args, "local_ranks_filter", None):
+        return args
+    if args.log_dir is None:
+        args.log_dir = os.path.join(tempfile.gettempdir(), f"accelerate_logs_{os.getpid()}")
+        logger.warning(
+            f"--local_ranks_filter was set without --log_dir; per-rank logs will be written to "
+            f"{args.log_dir} so that muted output is preserved."
+        )
+    if args.tee in (None, "", "0") and args.redirects in (None, "", "0"):
+        args.tee = "3"
+        logger.warning("--local_ranks_filter was set without --tee/--redirects; enabling --tee 3.")
+    return args
+
+
+def _report_child_failure(error, entrypoint, quiet=False, log_dir=None, tee=None):
+    """Print one aggregated diagnostic for a multi-rank failure, then exit or re-raise.
+
+    Additive by default: the original ``ChildFailedError`` is re-raised unchanged (same
+    exception type, same exit semantics). With ``--quiet``, exits with the first failure's
+    code mapped to the shell convention (signal ``-N`` -> ``128 + N``). If the formatter
+    returns ``None`` (disabled, no per-rank data, or an internal formatting error), the
+    original exception propagates exactly as before.
+
+    Accelerate-side values (``quiet``/``log_dir``/``tee``) must be captured by the caller
+    BEFORE ``_filter_args`` replaces the namespace with torchrun-recognized args only.
+    """
+    from accelerate.utils.error_reporting import classify_rank_failures, summarize_child_failure
+
+    summary = summarize_child_failure(error, entrypoint=entrypoint, log_dir=log_dir, tee=tee)
+    if summary is None:
+        raise error
+    print(summary, file=sys.stderr, flush=True)
+    if quiet:
+        sys.exit(classify_rank_failures(error.failures)["exitcode"])
+    raise error
+
+
 def simple_launcher(args):
     cmd, current_env = prepare_simple_launcher_cmd_env(args)
 
     process = subprocess.Popen(cmd, env=current_env)
-    process.wait()
+    try:
+        process.wait()
+    except KeyboardInterrupt:
+        # The child already received SIGINT from the terminal's process group. Wait briefly,
+        # escalate once, and exit with the conventional 128 + SIGINT code instead of dumping
+        # accelerate's own traceback (see https://github.com/huggingface/accelerate/issues/1089).
+        try:
+            process.wait(timeout=10)
+        except subprocess.TimeoutExpired:
+            process.terminate()
+            try:
+                process.wait(timeout=10)
+            except subprocess.TimeoutExpired:
+                process.kill()
+        sys.exit(130)
     if process.returncode != 0:
         if not args.quiet:
             raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd)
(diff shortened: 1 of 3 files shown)


mosafariuk:
@muellerzr @SunMarc @BenjaminBossan — gentle ping on this one. It's been open since Jul 28 with no reviewer assigned, so I suspect it just didn't land in anyone's queue rather than anyone deciding against it. The 5 workflows are also still awaiting maintainer approval, so there's no CI signal yet either — if someone could approve the run, `quality` and the unit tests would at least give you something to look at.

Happy to split it if the scope is the blocker: the `error_reporting` module plus the diagnostic block is the substantive half, and the `--quiet` / `--local_ranks_filter` / `KeyboardInterrupt` changes for #1089 could go separately. Also glad to drop anything that looks like scope creep.

mosafariuk:
Rebased onto current main (fd01e35c, 2026-08-20) — the branch had drifted a month behind. The previous workflow runs expired after 30 days awaiting approval without being granted, so this push re-requests them with a fresh clock. Same ask as before: if someone can approve the runs, `quality` and the unit tests give you an actual signal to look at (both pass locally against the rebased tree with the pinned `ruff==0.13.1`). The diff is unchanged in substance: the `error_reporting` module plus launcher hooks, validated on 2× L40 with a real NCCL watchdog timeout (details in the description).

PR #4145, opened 2026-07-31
[FSDP2] Add additional_modules_to_shard for sharding custom modules

yuxinyuan:
# What does this PR do?

Adds an opt-in `additional_modules_to_shard` option to `FullyShardedDataParallelPlugin` (FSDP2) that lets users explicitly shard extra modules into their own FSDP2 units, on top of what `auto_wrap_policy` selects.

## Motivation

FSDP2 preparation shards the modules matched by `auto_wrap_policy`, then uses a transformers-specific heuristic to carve the input/output embeddings and the final norm into their own units. Any module that is **neither** matched by the policy **nor** discovered by that heuristic falls into the root FSDP2 unit, which is not resharded after forward.

This is common for composite / multimodal / omni models that stitch together several transformer stacks. Take a Qwen3-TTS-style model as a concrete example: it contains a **speaker encoder**, a standard **LM decoder**, and a **code predictor** (another transformer). The top-level composite also doesn't expose a single `get_input_embeddings`/`get_output_embeddings`/final-norm through the standard `transformers` interface, so the built-in heuristic can't carve those out.

`additional_modules_to_shard` lets users close that gap explicitly — shard the non-standard modules into their own units, and optionally give specific submodules different sharding behavior (e.g. keep a precision-sensitive module in fp32, or skip resharding a custom tail)

## API

`additional_modules_to_shard` is a `selector -> overrides` mapping:

- **selector** (key): an exact fully-qualified module name or an `fnmatch` glob (e.g. `"model.norm"`, `"model.layers.*.mlp"`), or a `tuple` of them. A `str` shards each matched module as its **own** unit; a `tuple` groups the referenced modules into a **single** unit (like the built-in norm+lm_head tail). Names are canonicalized before matching, so selectors written against the original module names still match after `torch.compile` / activation checkpointing.
- **overrides** (value): a `dict` overriding the per-module sharding kwargs — `reshard_after_forward` and/or `mp_policy`; unset keys inherit the model-wide values (`offload_policy` is intentionally not supported).

```python
import torch
from torch.distributed.fsdp import MixedPrecisionPolicy
from accelerate import Accelerator, FullyShardedDataParallelPlugin

    fsdp_version=2,
    additional_modules_to_shard={
        # keep the tied final norm + lm_head in one unit and skip resharding after forward
        ("model.norm", "lm_head"): {"reshard_after_forward": False},
        # keep a precision-sensitive module in fp32
        "vision_tower": {"mp_policy": MixedPrecisionPolicy(param_dtype=torch.float32)},
        # shard every per-layer custom mlp with the model-wide defaults
        "model.layers.*.custom_mlp": {},
    },
)
accelerator = Accelerator(fsdp_plugin=plugin)
```

## Tests

Adds CPU unit tests mirroring the existing mocked `fsdp2_prepare_model` tests (`fully_shard` spied), covering: single/glob/grouped selection, per-module kwarg overrides, precedence over the auto wrap policy, name canonicalization for `torch.compile`/activation checkpointing, group-vs-traversal ordering, config validation, the FSDP1 no-op, and the setter.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

members/contributors who may be interested in your PR. @SunMarc

### DIFF
--- a/docs/source/usage_guides/fsdp.md
+++ b/docs/source/usage_guides/fsdp.md
@@ -114,6 +114,54 @@ fsdp_plugin = FullyShardedDataParallelPlugin(
 accelerator = Accelerator(fsdp_plugin=fsdp_plugin)
 ```
 
+### Sharding additional modules (FSDP2)
+
+For FSDP2, on top of the modules selected by `auto_wrap_policy`, Accelerate automatically carves the input/output
+embeddings and the final norm into their own FSDP2 units. This relies on the standard 🤗 Transformers interface
+(`get_input_embeddings`/`get_output_embeddings` and a conventional final norm). Any large module that is neither
+matched by the `auto_wrap_policy` nor discovered by this heuristic ends up in the root FSDP2 unit, which is not
+resharded after the forward pass, so its full parameters stay resident and increase peak memory.
+
+`additional_modules_to_shard` lets you explicitly shard such modules — useful for non-standard/custom architectures,
+and for modules that need different sharding behavior than the rest of the model. It is **FSDP2 only** and **Python
+only** (it cannot be set through the CLI/YAML config, because its values may hold objects such as a
+`MixedPrecisionPolicy`).
+
+It is a mapping of `selector -> overrides`:
+
+- The **selector** (key) is either a single module identifier or a `tuple` of them. Each identifier is either an
+  exact fully-qualified module name matched against `model.named_modules()` (e.g. `"model.norm"`), or an `fnmatch`-style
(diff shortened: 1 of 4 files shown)


yuxinyuan:
Any feedback maybe ? @SunMarc

SunMarc:
Hey @yuxinyuan , for now, i prefer to keep fsdpv2 api simple. Maybe if there is more interest from the community, we might add it but for now I think it is best to keep it this way ! Feel free to open an issue to track for this ! Closing for now

PR #4152, opened 2026-08-09
Add missing type hints in utils/memory.py

RudrenduPaul:
## What this PR does

Adds missing type hints to two functions in `src/accelerate/utils/memory.py`:

- `clear_device_cache(garbage_collection: bool = False) -> None`
- `release_memory(*objects) -> list`

This follows the same self-identified typing-consistency pattern as my prior
PRs in this repo (#4123, #4124, #4125) — no linked issue, just filling in
type hints that were missing while the rest of the module (and the file's
other two functions, `should_reduce_batch_size` and
`find_executable_batch_size`) already had them.

## Functions updated

- `clear_device_cache`: `garbage_collection` is only ever called with `bool`
  values (`True`/`False`) at every call site in the codebase, and the
  function has no `return` statement, so `-> None` is accurate.
- `release_memory`: builds a `list` from `*objects`, sets every element to
  `None`, and returns that list — verified by reading the full function
  body, so `-> list` is accurate. The variadic `*objects` parameter itself
  is intentionally left unannotated (it accepts arbitrary object types to
  release from memory), consistent with how `*args`/`**kwargs` were left
  unannotated in my earlier `hooks.py` (#4124) and `logging.py` (#4125) PRs.

## Functions intentionally left unchanged

- `should_reduce_batch_size` and `find_executable_batch_size` already have
  complete type hints on every parameter in this file — no changes needed.

## No behaviour change

Type-hint-only change. No logic was modified.

## Testing

- `ruff check src/accelerate/utils/memory.py` — all checks pass.
- `ruff format --check src/accelerate/utils/memory.py` — passes, file
  already formatted.
- `python3 -c "import ast; ast.parse(...)"` on the modified file — passes.

> **Note:** Claude Code was used to assist in drafting this change. All
> changes were reviewed by the submitter.

### DIFF
--- a/src/accelerate/utils/memory.py
+++ b/src/accelerate/utils/memory.py
@@ -37,7 +37,7 @@ from .imports import (
 )
 
 
-def clear_device_cache(garbage_collection=False):
+def clear_device_cache(garbage_collection: bool = False) -> None:
     """
     Clears the device cache by calling `torch.{backend}.empty_cache`. Can also run `gc.collect()`, but do note that
     this is a *considerable* slowdown and should be used sparingly.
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
Thanks, merging this one but in the future, try to fix more of these before opening a PR. There is no use of merging 100's of PR for things like this

PR #4155, opened 2026-08-09
Preserve defaultdict batches during device placement

### LINKED ISSUE #4154: send_to_device crashes on defaultdict batches
## System Info

- `Accelerate` version: 1.14.0; also reproduced on 1.15.0.dev0 at `16cb6eb8`
- Platform: Windows-10-10.0.26200-SP0
- Python version: 3.11.9
- NumPy version: 2.4.6
- PyTorch version: 2.13.0+cpu
- PyTorch accelerator: N/A
- Accelerate default config: Not found

## Information

- [x] My own modified scripts

## Tasks

- [ ] One of the official example tasks
- [x] My own task or dataset

## Reproduction

`Accelerator.prepare()` fails during iteration when a PyTorch `DataLoader` collator returns a standard-library `collections.defaultdict`:

```python
from collections import defaultdict

import torch

def collate(examples):
    return defaultdict(list, input_ids=torch.tensor(examples))


accelerator = Accelerator(cpu=True)
dataloader = accelerator.prepare(
    DataLoader([1, 2], batch_size=2, collate_fn=collate)
)
print(next(iter(dataloader)))
```

Actual result:

```text
TypeError: first argument must be callable or None
```

This also reproduces directly with the public utility:

```python
from accelerate.utils import send_to_device

send_to_device(
    defaultdict(list, input_ids=torch.tensor([1, 2])),
    "cpu",
)
```

`send_to_device` reconstructs every `Mapping` as `type(tensor)(new_mapping)`. For `defaultdict`, the first positional constructor argument is the default factory rather than initial data, which causes the exception.

Relevant source: https://github.com/huggingface/accelerate/blob/16cb6eb80dd9aa8b4df1a63ef57863e455d53b83/src/accelerate/utils/operations.py#L181-L190

I have a focused fix and regression test prepared.

## Expected behavior

Prepared dataloaders should preserve the mapping type, tensor values, and `default_factory`, and yield the batch without error. For the reproduction above, the expected batch is a `defaultdict` containing `tensor([1, 2])`, with missing keys still producing `[]`.

AI disclosure: I used Codex to help identify and independently reproduce this behavior on the current release and main, search for duplicates, test a focused fix, and draft this report. I reviewed the reproduction, patch, and tests and will personally handle follow-up.

aswanth-07:
## What does this PR do?

Preserves standard-library `defaultdict` batches while `send_to_device` recursively moves their values.

`send_to_device` currently reconstructs every mapping as `type(mapping)(items)`. For `defaultdict`, that positional argument is interpreted as the default factory, so prepared dataloaders crash before yielding a batch. This change rebuilds a `defaultdict` with both its original `default_factory` and the moved values, while leaving the existing path unchanged for other mapping types.

Fixes #4154

## Validation

- `pytest tests/test_utils.py::UtilsTester::test_send_to_device -q` - 1 passed
- `pytest tests/test_utils.py -q -k 'not test_convert_to_fp32'` - 43 passed, 5 skipped
- Full repository `ruff check .` - passed
- Full repository `ruff format --check .` - 184 files already formatted
- Public `Accelerator.prepare(DataLoader(...))` reproduction - fixed on CPU

The one excluded test invokes `torch.compile` and fails on this Windows host because the MSVC `cl` compiler is unavailable; it is unrelated to this mapping change.

## Before submitting

- [x] I read the contributor guideline.
- [x] The bug is documented in #4154.
- [x] No documentation update is needed; this restores nested mapping behavior.
- [x] I added a focused regression test.

AI disclosure: I used Codex to help identify and reproduce the bug, search for duplicates, implement the focused patch, run validation, and draft this PR. I reviewed and understand the change and will personally handle review feedback.

### DIFF
--- a/src/accelerate/utils/operations.py
+++ b/src/accelerate/utils/operations.py
@@ -17,6 +17,7 @@ A set of basic tensor ops compatible with tpu, gpu, and multigpu
 
 import pickle
 import warnings
+from collections import defaultdict
 from collections.abc import Mapping
 from contextlib import contextmanager, nullcontext
 from functools import update_wrapper, wraps
@@ -181,12 +182,13 @@ def send_to_device(tensor, device, non_blocking=False, skip_keys=None):
             skip_keys = [skip_keys]
         elif skip_keys is None:
             skip_keys = []
-        return type(tensor)(
-            {
-                k: t if k in skip_keys else send_to_device(t, device, non_blocking=non_blocking, skip_keys=skip_keys)
-                for k, t in tensor.items()
-            }
-        )
+        new_mapping = {
+            k: t if k in skip_keys else send_to_device(t, device, non_blocking=non_blocking, skip_keys=skip_keys)
+            for k, t in tensor.items()
+        }
+        if isinstance(tensor, defaultdict):
+            return type(tensor)(tensor.default_factory, new_mapping)
+        return type(tensor)(new_mapping)
     else:
         return tensor
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -16,7 +16,7 @@ import pickle
 import tempfile
 import unittest
 import warnings
-from collections import UserDict, namedtuple
+from collections import UserDict, defaultdict, namedtuple
 from typing import NamedTuple, Optional
 from unittest.mock import Mock, patch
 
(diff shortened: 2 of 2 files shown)


aswanth-07:
Closing this after a stricter scope review. The implementation is technically correct, but it adds a permanent special case to a shared utility for an uncommon container without demonstrated user demand. I do not think the extra branch is justified. Thanks for the space to investigate it.

PR #4156, opened 2026-08-10
docs: fix documented parameter names that do not match the signature

massimiliano1991:
## What

Three docstrings document a parameter name that does not exist in the function signature:

| location | documented | actual |
|---|---|---|
| `utils/operations.py` `recursively_apply` | `main_type` | `test_type` |
| `tracking.py` `SwanLabTracker.log` | `data` | `values` |
| `tracking.py` `MLflowTracker.log_figure` | `**kwargs` | `**save_kwargs` |

## Why it's worth fixing

For the first two the mismatch is not only cosmetic. Both functions accept `**kwargs`, so a
keyword copied from the documentation is not rejected at the boundary — it is swallowed and
forwarded to the user's own callback:

```python
>>> recursively_apply(my_fn, data, main_type=torch.Tensor)   # as documented
TypeError: my_fn() got an unexpected keyword argument 'main_type'
```

The traceback names `my_fn`, not `recursively_apply`, so nothing points back to the docstring
that suggested the argument.

`recursively_apply`'s docstring also describes `main_type` as a *type* defaulting to
`torch.Tensor`, while `test_type` is a *predicate* defaulting to `is_torch_tensor`. The three
other references to `main_type` in the same docstring (including the `Returns:` line) are
updated to match.

For `MLflowTracker.log_figure` the `**` form absorbs any name, so there is no error — but the
docstring also pointed at `mlflow.log_image` rather than `mlflow.log_figure`. The surrounding
`Args:` block had its description lines un-indented, which breaks rendering, so those lines are
re-indented to the style used by the rest of the file. `SwanLabTracker.log` had the same issue
and now matches the other trackers' `log` methods in that module.

## How these were found

Mechanically: parsing each `Args:` block and diffing it against the `ast` signature, then
checking each hit by hand. Only these three came out as documented-but-absent across
`src/accelerate`.

## Checks

- docs only — no behaviour change, no public API touched
- `ruff format --check` clean on both files; `ruff check` reports only pre-existing findings in
  untouched lines

### DIFF
--- a/src/accelerate/tracking.py
+++ b/src/accelerate/tracking.py
@@ -842,13 +842,13 @@ class MLflowTracker(GeneralTracker):
         Logs an figure to the current run.
 
         Args:
-            figure (Any):
-            The figure to be logged.
+            figure (`Any`):
+                The figure to be logged.
             artifact_file (`str`, *optional*):
-            The run-relative artifact file path in posixpath format to which the image is saved.
-            If not provided, the image is saved to a default location.
-            **kwargs:
-            Additional keyword arguments passed to the underlying mlflow.log_image function.
+                The run-relative artifact file path in posixpath format to which the image is saved. If not
+                provided, the image is saved to a default location.
+            **save_kwargs:
+                Additional keyword arguments passed to the underlying `mlflow.log_figure` function.
         """
         import mlflow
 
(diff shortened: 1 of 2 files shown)


SunMarc (approved):
Thanks

PR #3071, opened 2024-09-03
add support for custom function for reducing the batch size

winglian:
# What does this PR do?

when using auto_find_batch_size in transformers, there is no way to compensate with gradient_accumulation_steps or learning rate when a new batch size is tried. This adds the ability to use a callable function that transformers can set to manage updating the TrainingArgs.

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?
@muellerzr 
Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4160, opened 2026-08-11
Forward reduce_batch_size_fn through the find_executable_batch_size decorator form

vineethsaivs:
## Problem

`find_executable_batch_size` supports being used as a decorator with arguments, and that is the form its own docstring example uses. That path goes through:

```python
if function is None:
    return functools.partial(find_executable_batch_size, starting_batch_size=starting_batch_size)
```

`reduce_batch_size_fn` is not forwarded. It is bound on the first call, dropped by the partial, and on the second call it is `None` again, so the default multiply-by-0.9 back-off runs instead.

```python
current = 256
def halve():
    global current
    current //= 2
    return current

@find_executable_batch_size(starting_batch_size=256, reduce_batch_size_fn=halve)
def train(batch_size): ...

# batch sizes tried:
#   decorator form  [256, 230, 207, 186, 167, 150, 135, 121, ...]   26 attempts
#   direct call     [256, 128, 64, 32, 16]                           5 attempts
```

Passing the function directly, `find_executable_batch_size(train, starting_batch_size=256, reduce_batch_size_fn=halve)`, already works, so the two documented forms disagree.

Every one of those extra attempts is a real out-of-memory-and-retry cycle: allocate, fail, `clear_device_cache(garbage_collection=True)`, try again. A halving strategy from 256 should cost 5 of them and costs 26. Nothing errors and nothing warns, so the only symptom is a slow start-up and a batch size that walks down in 10% steps rather than the steps the caller asked for.

`reduce_batch_size_fn` was added in #3071 and documented in #4051; only the forwarding through the partial was missed.

## Fix

Forward the third argument.

## Test

Three cases added to `tests/test_memory_utils.py`, next to the existing `find_executable_batch_size` tests, which all use the decorator form with `starting_batch_size` only:

- the decorator form honours a custom back-off
- the decorator form and the direct-call form agree on the same input
- the default 0.9 back-off is unchanged when no function is given

```
                                                 before   after
test_reduce_batch_size_fn_is_used_by_the_decorator  FAIL    pass
test_reduce_batch_size_fn_matches_the_direct_call   FAIL    pass
test_default_back_off_is_unchanged                  pass    pass
the 7 pre-existing tests in the file                pass    pass
```

`pytest tests/test_memory_utils.py` gives 2 failed / 8 passed against `main` and 10 passed with this change, on CPU. Both runs use `src/` from a tree checked out at `16cb6eb`, verified by printing `accelerate.utils.memory.__file__`, because an editable install elsewhere on the machine will otherwise shadow the checkout and make the before and after runs test the same code.

`ruff check` and `ruff format --check` are clean on both changed files. Note that `ruff --fix` also wants to remove a pre-existing `pass` at `tests/test_memory_utils.py:143`; I put it back so this diff is only the change described here.


tests/test_memory_utils.py
```diff
@@ -146,6 +146,94 @@ def mock_training_loop_function(batch_size):
...
+            22,
+            19,
+            17,
+            15,
+            13,
+            11,
+            9,
+            8,
+        ]
+
```
SunMarc:
```suggestion
```

SunMarc (commented):
Indeed

Issue #3080, opened 2024-09-04
Using StatefulDataLoader when num_workers>1 doesn't work


kmchiti:
### System Info

```Shell
Accelerate version: 0.34.0
Platform: Linux 5.15.0-101-generic #111-Ubuntu SMP x86_64 GNU/Linux
Python version: 3.10.14
PyTorch version (GPU?): 2.1.2+cu118 True
System RAM: 494 GB
GPU type: 4x NVIDIA A100
```

### Information

- [ ] The official example scripts
- [X] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [X] My own task or dataset (give details below)

### Reproduction

1. Create an IterableDataset.
2. Prepare a toy model (e.g., GPT-2  with a limited vocabulary size and sequence length)
3. Use a DataLoader with use_stateful_dataloader=True in Accelerate.
4. Run the training script on a single GPU (works fine).
5. Run the training script on multiple GPUs using torchrun --nproc_per_node=4 (error occurs).

**Code Example:**

```
from transformers import GPT2Config, GPT2LMHeadModel, GPT2Tokenizer, DataCollatorForLanguageModeling
from datasets import Dataset
from accelerate import Accelerator
from torch.utils.data import DataLoader
from accelerate.utils import DataLoaderConfiguration
import torch

if __name__ == "__main__":
    
    tokenizer = GPT2Tokenizer.from_pretrained("gpt2")
    tokenizer.pad_token = tokenizer.eos_token
    
    ds = Dataset.from_dict({"input_ids": torch.randint(0,10, (4096,64))}).to_iterable_dataset(num_shards=5)
    dataloader_params = {
            "batch_size": 16,
            "collate_fn": DataCollatorForLanguageModeling(tokenizer=tokenizer, mlm=False),
            "num_workers": 4,
            "pin_memory": True,
        }
    data_loader = DataLoader(ds, **dataloader_params)
    
    model = GPT2LMHeadModel(GPT2Config(vocab_size=10, n_positions=64))
    optimizer = torch.optim.AdamW(model.parameters(), lr=5e-5)
    
    
    accelerator = Accelerator(dataloader_config=DataLoaderConfiguration(use_stateful_dataloader=True))
    model, data_loader, optimizer = accelerator.prepare(model, data_loader, optimizer)
    
    
[...]

PR #4165, opened 2026-08-14
Fix stateful dataloader checkpointing across processes

jrsmartin:
## What does this PR do?

This PR fixes stateful dataloader checkpointing during distributed training by saving each process's state separately. A stateful dataloader can hold process-specific cursor, dataset, worker, and RNG state. Currently, every process writes its state to the same `dl_state_dict.bin` filename in a shared checkpoint directory, so the final file contains whichever process wrote last, and other processes can restore the wrong data position.

Distributed checkpoints now use rank-qualified filenames such as `dl_state_dict_rank0.bin`. Loading prefers the current process's file and falls back to the legacy unqualified filename, which preserves compatibility with existing checkpoints. Single-process checkpoint filenames remain unchanged.

## Related context

Discussed in [#3080](https://github.com/huggingface/accelerate/issues/3080), where manually saving and loading the dataloader state on every rank is [described](https://github.com/huggingface/accelerate/issues/3080#issuecomment-2589574065) as the current workaround. This PR addresses that per-rank checkpoint persistence gap.

## Tests

- Added a two-process regression using distinct per-rank dataset streams and a shared checkpoint directory. It verifies exact batch replay after restore and asserts that both rank-qualified state files exist.
- Added a CPU/Gloo entry point so the distributed regression runs in the normal pull-request test suite as well as the existing multi-device suite. It skips when the optional `torchdata` dependency is unavailable.
- Added unit coverage for single-process naming, multiple dataloaders, process-specific load precedence, and legacy fallback.
- Verified the regression failed before the implementation and passes afterward.
- `make quality`
- `pytest -q tests/test_state_checkpointing.py -k "not map_location"` — 18 passed
- `pytest -q tests/test_accelerator.py -k stateful_dataloader` — 17 passed
- CPU/Gloo two-process regression — passed
- Two-GPU distributed data-loop suite — passed

## Before submitting

- [X] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Discussed in [#3080](https://github.com/huggingface/accelerate/issues/3080), where per-rank dataloader state saving/loading is [described](https://github.com/huggingface/accelerate/issues/3080#issuecomment-2589574065) as the required workaround. This PR addresses that per-rank checkpoint persistence gap
- [x] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).

## Who can review?

Anyone in the community is free to review the PR once the tests have passed.

### DIFF
--- a/docs/source/basic_tutorials/migration.md
+++ b/docs/source/basic_tutorials/migration.md
@@ -222,3 +222,4 @@ Any other stateful items to be stored should be registered with the [`~Accelerat
 
 > [!TIP]
 > If you have [`torchdata>=0.8.0`](https://github.com/pytorch/data/tree/main) installed, you can additionally pass `use_stateful_dataloader=True` into your [`~utils.DataLoaderConfiguration`]. This extends Accelerate's DataLoader classes with a `load_state_dict` and `state_dict` function, and makes it so `Accelerator.save_state` and `Accelerator.load_state` also track how far into the training dataset it has read when persisting the model.
+> During distributed training, each process's data loader state is saved and restored independently.
--- a/src/accelerate/checkpointing.py
+++ b/src/accelerate/checkpointing.py
@@ -60,6 +60,12 @@ from .state import PartialState
 logger = get_logger(__name__)
 
 
+def _get_dataloader_state_dict_filename(index: int, process_index: Optional[int] = None) -> str:
+    dataloader_suffix = "" if index == 0 else f"_{index}"
+    process_suffix = "" if process_index is None else f"_rank{process_index}"
+    return f"dl_state_dict{dataloader_suffix}{process_suffix}.bin"
+
+
 def save_accelerator_state(
     output_dir: str,
     model_states: list[dict],
@@ -139,7 +145,9 @@ def save_accelerator_state(
             if isinstance(sampler, SeedableRandomSampler):
                 save(sampler, output_sampler_file, save_on_each_node=save_on_each_node, safe_serialization=False)
         if getattr(dataloader, "use_stateful_dataloader", False):
-            dataloader_state_dict_name = "dl_state_dict.bin" if i == 0 else f"dl_state_dict_{i}.bin"
+            # Stateful dataloader state is process-specific during distributed training.
+            dataloader_process_index = process_index if PartialState().num_processes > 1 else None
+            dataloader_state_dict_name = _get_dataloader_state_dict_filename(i, dataloader_process_index)
             output_dataloader_state_dict_file = output_dir.joinpath(dataloader_state_dict_name)
             state_dict = dataloader.state_dict()
             torch.save(state_dict, output_dataloader_state_dict_file)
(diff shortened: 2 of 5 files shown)


jrsmartin:
@SunMarc would you mind reviewing this PR when you have a chance - thank you!

jrsmartin:
> @SunMarc would you mind reviewing this PR when you have a chance - thank you!

Friendly bump @SunMarc, and tagging @muellerzr since this addresses the per-rank StatefulDataLoader checkpoint workaround discussed in #3080.  Would either of you be able to review? Cheers

GoldenStain:
Thanks for this PR — the per-rank filenames are a real fix for the shared-file collision.

One gap worth flagging during review: the new regression
(`test_stateful_dataloader_save_state_per_process`) builds
`DataLoader(dataset, batch_size=1, num_workers=0)`, which uses the default
`SequentialSampler` — so its "exact batch replay" assertion passes even though
only cursor counters carry values (sequential order is derivable from the
cursor alone). With `shuffle=True` the saved state's sampler slots
(`_index_sampler_state`, `_sampler_iter_state`, `_shared_seed`) are all `None`:
`BatchSamplerShard` doesn't implement `Stateful`, so torchdata's collection
skips the wrapped chain entirely, and resume restores the position but not
the order.

Full analysis + reproducer: #4195. A `shuffle=True` variant of this test
would fail today and would pin the remaining gap in the same suite — happy to
help with that case if useful.

GoldenStain:
Thanks for this PR — the per-rank filenames are a real fix for the shared-file collision.

One gap worth flagging during review: the new regression
(`test_stateful_dataloader_save_state_per_process`) builds
`DataLoader(dataset, batch_size=1, num_workers=0)`, which uses the default
`SequentialSampler` — so its "exact batch replay" assertion passes even though
only cursor counters carry values (sequential order is derivable from the
cursor alone). With `shuffle=True` the saved state's sampler slots
(`_index_sampler_state`, `_sampler_iter_state`, `_shared_seed`) are all `None`:
`BatchSamplerShard` doesn't implement `Stateful`, so torchdata's collection
skips the wrapped chain entirely, and resume restores the position but not
the order.

Full analysis + reproducer: #XXXXXX. A `shuffle=True` variant of this test
would fail today and would pin the remaining gap in the same suite — happy to
help with that case if useful.

Two more variants worth considering for the same suite while you're in there:
(1) a checkpoint taken mid-epoch-0 fails even with `use_seedable_sampler=True`
+ `set_epoch` — the sampler swap lands after `DataLoaderShard` is built, so
epoch 0 still uses the un-swapped vanilla sampler; (2) an epoch-boundary
checkpoint fails for samplers that auto-increment their epoch —
`SeedableRandomSampler` bumps `self.epoch += 1` after each full iteration, and
torchdata's restore runs one extra full sampler cycle past the finished epoch,
shifting the order by exactly one permutation. A correct order-restoring
sampler chain has to clear all three bars (post-epoch-0 mid-epoch, mid-epoch-0,
epoch boundary).

GoldenStain:
Thanks for this PR — the per-rank filenames are a real fix for the shared-file collision.

One gap worth flagging during review: the new regression
(`test_stateful_dataloader_save_state_per_process`) builds
`DataLoader(dataset, batch_size=1, num_workers=0)`, which uses the default
`SequentialSampler` — so its "exact batch replay" assertion passes because sequential
order is derivable from the cursor alone. With `shuffle=True`, `BatchSamplerShard`
historically severed torchdata's sampler-state protocol chain, so resume restored
the cursor position but not the shuffle permutation.

This complementary shuffle order gap is tracked in #4195 and is actively being
addressed in PR #4203. Having both #4165 (per-rank isolation) and #4203 (shuffle state serialization) will complete the shuffle-exact resume story for use_stateful_dataloader under DDP.

jrsmartin:
Thanks for the comment, looks like a valid separate issue. This regression intentionally isolates the per-rank checkpoint-file collision using rank-specific dataset state. Let's wait to see what the reviewers say when they have a chance to take a look.

SunMarc:
will have a look soon !

jrsmartin:
> will have a look soon !

Sounds good, thanks @SunMarc !

PR #4168, opened 2026-08-18
Handle missing accelerate executable in env command

tandede:
## Summary

- replace the platform-specific `which`/`where` subprocess with `shutil.which`
- preserve the existing `Not found` diagnostic when the console script is not on `PATH`
- add regression coverage for invoking the environment command without an `accelerate` executable on `PATH`

## Why

`accelerate env` initializes the reported executable location to `Not found`, but then unconditionally runs an external lookup command. When Accelerate is invoked through an explicit path without activating its environment, that lookup exits non-zero and `subprocess.check_output()` raises `CalledProcessError`. As a result, the diagnostic command crashes instead of reporting the already-intended fallback.

Using `shutil.which()` provides the same lookup on POSIX and Windows while allowing a missing executable to be handled normally.

## Validation

- `PATH=/usr/bin:/bin accelerate env` (reports `` `accelerate` bash location: Not found ``)
- `pytest -q tests/test_cli.py -k 'EnvCommandTester or test_hyphen or test_underscore'` (3 passed)
- `pytest -q tests/test_memory_utils.py` (7 passed)
- `pytest -q tests/test_cli.py -k 'not test_invalid_model_name_transformers'` (38 passed, 1 skipped, 5 subtests passed)
- `make style`
- `make quality`

### DIFF
--- a/src/accelerate/commands/env.py
+++ b/src/accelerate/commands/env.py
@@ -17,7 +17,7 @@
 import argparse
 import os
 import platform
-import subprocess
+from shutil import which
 
 import numpy as np
 import psutil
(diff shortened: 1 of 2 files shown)


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. The change handles the missing `accelerate` executable path and includes regression coverage. It is ready for review, and I’m happy to address any feedback. Thank you!

SunMarc (approved):
Thanks for the change ! This looks much better

PR #4180, opened 2026-08-23
Raise a clear error when FSDP is enabled on a mesh with no shard dimension

qgallouedec:
When FSDP2 is enabled together with a `ParallelismConfig` whose ranks are fully consumed by other dimensions (for example `tp_size == world_size`, which is also what transformers' expert parallelism produces), `fsdp2_prepare_model` slices the mesh with `fsdp_dim_names` and fails with a confusing error:

```
KeyError: "Invalid mesh_dim_names ('dp_shard_cp',) specified. Valid mesh_dim_names are ['tp']."
```

This PR raises an explicit error instead:

```
ValueError: FSDP is enabled but the device mesh is ('tp',), which has no dimension for FSDP to shard across (both `dp_shard_size` and `cp_size` are 1). This usually means a model that is already parallelized another way -- e.g. loaded with `DistributedConfig(tp_size=N)` or `enable_expert_parallel=True`, which makes the whole world size tensor/expert parallel -- was launched under an FSDP config. Either launch it without the FSDP config, or leave ranks for FSDP to use by lowering `tp_size`.
```

Reproduction (2 GPUs):

```python
# torchrun --nproc_per_node 2 repro.py
import torch
import torch.nn as nn
from accelerate.parallelism_config import ParallelismConfig
from accelerate.utils import FullyShardedDataParallelPlugin

accelerator = Accelerator(
    parallelism_config=ParallelismConfig(tp_size=2),
    fsdp_plugin=FullyShardedDataParallelPlugin(fsdp_version=2),
)
model = nn.Linear(8, 8)
```

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -760,19 +760,6 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
     fsdp2_plugin.set_auto_wrap_policy(model)
 
     mesh = getattr(accelerator, "torch_device_mesh", None)
-    pc = accelerator.parallelism_config
-
-    # `fsdp_dim_names` always asks for `dp_shard_cp`, but that joint dimension is only flattened into the mesh when
-    # `dp_shard` or `cp` is enabled. Without either, slicing the mesh below raises a `KeyError` from `device_mesh`.
-    if mesh is not None and not pc.dp_shard_enabled and not pc.cp_enabled:
-        raise ValueError(
-            f"FSDP is enabled but the device mesh is {tuple(mesh.mesh_dim_names)}, which has no dimension for FSDP "
-            "to shard across (both `dp_shard_size` and `cp_size` are 1). This usually means a model that is already "
-            "parallelized another way -- e.g. loaded with `DistributedConfig(tp_size=N)` or "
-            "`enable_expert_parallel=True`, which makes the whole world size tensor/expert parallel -- was launched "
-            "under an FSDP config. Either launch it without the FSDP config, or leave ranks for FSDP to use by "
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -753,6 +753,19 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
...
+    # `dp_shard` or `cp` is enabled. Without either, slicing the mesh below raises a `KeyError` from `device_mesh`.
+    if mesh is not None and not pc.dp_shard_enabled and not pc.cp_enabled:
+        raise ValueError(
+            f"FSDP is enabled but the device mesh is {tuple(mesh.mesh_dim_names)}, which has no dimension for FSDP "
+            "to shard across (both `dp_shard_size` and `cp_size` are 1). This usually means a model that is already "
+            "parallelized another way -- e.g. loaded with `DistributedConfig(tp_size=N)` or "
+            "`enable_expert_parallel=True`, which makes the whole world size tensor/expert parallel -- was launched "
+            "under an FSDP config. Either launch it without the FSDP config, or leave ranks for FSDP to use by "
+            "lowering `tp_size`."
+        )
```
SunMarc: Can't we put that in `_validate_accelerator` ? i feel like should detect this way earlier

SunMarc (commented):
Thanks ! Left a comment

src/accelerate/utils/fsdp_utils.py
qgallouedec: Good call, moved to _validate_accelerator.
The condition only needs the config (is_fsdp2 + neither dp_shard nor cp enabled), so it now raises at Accelerator() init instead of inside prepare_model.

SunMarc (approved):
Thanks

PR #4186, opened 2026-08-27
Avoid double-counting CPU RAM for integrated CUDA

### LINKED ISSUE #4183: get_max_memory() adds a separate cpu entry on integrated CUDA GPUs, double-counting shared RAM (GB10 / DGX Spark)
## Summary

`get_max_memory()` already treats shared memory correctly for MPS, but the same
reasoning is not applied to integrated CUDA devices, so on those the returned
budget counts the same physical RAM twice.

Current code (`src/accelerate/utils/modeling.py`, around L822):

```python
# allocate everything in the mps device as the RAM is shared
if is_mps_available():
    max_memory["mps"] = psutil.virtual_memory().available
else:
    max_memory["cpu"] = psutil.virtual_memory().available
```

The comment states the principle exactly: when RAM is shared, do not hand out a
second pool. An NVIDIA GB10 (DGX Spark) shares its RAM the same way, but it is
a CUDA device, so it takes the `else` branch and gets a separate `cpu` entry
alongside its device entry.

Measured on a GB10:

```python
>>> import torch
>>> from accelerate.utils import get_max_memory
>>> torch.cuda.get_device_properties(0).is_integrated
1
>>> get_max_memory()
{0: 73446760448, 'cpu': 87132405760}
```

73.4 GB + 87.1 GB = **160.5 GB advertised**, on a machine with:

```
$ grep MemTotal /proc/meminfo
MemTotal:       127535336 kB        # 127.5 GB
```

There is no separate device VRAM on this part. `torch.cuda.mem_get_info(0)`
reports `free=74.1 GB, total=130.6 GB`, and those are the same bytes
`/proc/meminfo` describes, reached over NVLink-C2C rather than PCIe. The two
entries alias one pool: allocating against either shrinks the other, which the
returned dict cannot express.

## Why it matters

Everything downstream inherits the number. `infer_auto_device_map` and
`device_map="auto"` treat the entries as independent capacities, so a model
sized between the device entry and the sum can be planned to "fit" across a
split whose halves are the same memory.

The hardware is not niche. GB10 / DGX Spark, GH200, GB200, Jetson Thor and Orin
are all integrated-memory parts, and they are specifically the machines people
buy to run large models locally.

## Prior art in this file

- **#2929** (`get_max_memory()` returns allocated instead of total device
  memory for XPU) is the same class of bug: a device type measured with the
  wrong quantity.
- The MPS branch above is the fix already written, for a different vendor.

## Suggested fix

Happy to send the PR. Two shapes, and I would rather write the one you want
than guess:

**A. Extend the existing shared-RAM rule to integrated CUDA devices.** Where
the MPS check happens, also treat an integrated CUDA device as sharing host
RAM, so no independent `cpu` entry is advertised against it. Smallest change,
and it reuses the reasoning already in the file.

**B. Leave `get_max_memory()` alone and fix the consumer.** Teach
`infer_auto_device_map` that an integrated device and the host are one pool, so
it will not plan a split between two aliases of the same memory.

I lean towards A, since it corrects the value everything else derives from.

Two implementation notes either way:

- `is_integrated` should be read defensively, e.g.
  `getattr(torch.cuda.get_device_properties(i), "is_integrated", False)`.
  setup.py allows `torch>=2.0.0` and I have only confirmed the attribute on
  torch 2.10, so a bare access risks breaking older installs.
- This cannot be exercised on your CI, since it needs integrated-memory
  hardware. A unit test would have to fake the device property (monkeypatch
  `get_device_properties`) and assert the shape of the returned dict. Happy to
  write it that way.

## What I am not claiming

I found this while debugging a 4-bit load that failed under `device_map="auto"`
with "Some modules are dispatched on the CPU or the disk", which
`device_map={"": 0}` resolved. **I could not reproduce that failure with plain
transformers** on the same machine and model, so I am not asserting the double
count caused it. The arithmetic above stands on its own.

## Environment

```
accelerate    1.12.0
transformers  5.5.0
torch         2.10.0a0+b558c986e8.nv25.11  (torch.version.cuda 13.0)
device        NVIDIA GB10, capability (12, 1), is_integrated=1
driver        580.173.02, CUDA runtime 13.0
platform      Linux-6.17.0-1029-nvidia-aarch64-with-glibc2.39, aarch64
python        3.12.3
```

arpansahu:
﻿### Problem
`get_max_memory()` adds a separate `"cpu"` memory entry even when CUDA reports an integrated GPU. Integrated CUDA devices share host RAM, so downstream device-map planning can treat one physical memory pool as independent CUDA and CPU capacities. Closes #4183.

### Reproducer
```python
from types import SimpleNamespace
from unittest.mock import patch
import torch
from accelerate.utils import modeling
from accelerate.utils.modeling import get_max_memory

with patch.object(modeling, "is_mps_available", return_value=False), \
     patch.object(torch.cuda, "device_count", return_value=1), \
     patch.object(torch.cuda, "mem_get_info", return_value=(1234, 5678)), \
     patch.object(torch.cuda, "get_device_properties", return_value=SimpleNamespace(is_integrated=True)), \
     patch.object(torch, "tensor"):
    print(get_max_memory())
# before: {0: 1234, "cpu": ...}
# after:  {0: 1234}
```

### Fix
Track whether any available CUDA device reports `is_integrated` and skip adding a separate CPU memory budget in that case. The attribute is read with `getattr(..., False)` so older PyTorch/device properties keep the existing behavior.

### Testing
Added `tests/test_modeling_utils.py::ModelingUtilsTester::test_get_max_memory_integrated_cuda_does_not_add_cpu`.

Before the fix:
```text
FAILED tests/test_modeling_utils.py::ModelingUtilsTester::test_get_max_memory_integrated_cuda_does_not_add_cpu
1 failed in 7.37s
```

After the fix:
```text
2 passed, 1 skipped in 6.56s
7 passed, 6 skipped in 69.63s (0:01:09)
```

Quality:
```text
ruff check ... --fix: All checks passed!
ruff format ...: 2 files left unchanged
ruff check ...: All checks passed!
ruff format --check ...: 2 files already formatted
```

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -762,6 +762,7 @@ def get_max_memory(max_memory: Optional[dict[Union[int, str], Union[int, str]]] 
 
     if max_memory is None:
         max_memory = {}
+        shared_ram_cuda = False
         # Make sure device is initialized on each device to have the right memory info.
         if is_npu_available():
             for i in range(torch.npu.device_count()):
@@ -816,13 +817,16 @@ def get_max_memory(max_memory: Optional[dict[Union[int, str], Union[int, str]]] 
                 try:
                     _ = torch.tensor([0], device=i)
                     max_memory[i] = torch.cuda.mem_get_info(i)[0]
+                    shared_ram_cuda = shared_ram_cuda or getattr(
+                        torch.cuda.get_device_properties(i), "is_integrated", False
+                    )
                 except Exception:
                     logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
                     continue
-        # allocate everything in the mps device as the RAM is shared
+        # allocate everything in the mps device as the RAM is shared; integrated CUDA devices do the same
         if is_mps_available():
             max_memory["mps"] = psutil.virtual_memory().available
-        else:
+        elif not shared_ram_cuda:
             max_memory["cpu"] = psutil.virtual_memory().available
         return max_memory
 
--- a/tests/test_modeling_utils.py
+++ b/tests/test_modeling_utils.py
@@ -18,7 +18,9 @@ import tempfile
 import unittest
 import warnings
 from collections import OrderedDict
+from types import SimpleNamespace
 from typing import Optional
+from unittest.mock import patch
 
 import torch
 import torch.nn as nn
@@ -34,6 +36,7 @@ from accelerate.test_utils import (
     require_non_hpu,
     torch_device,
 )
+from accelerate.utils import modeling
 from accelerate.utils.modeling import (
     align_module_device,
     check_device_map,
@@ -44,6 +47,7 @@ from accelerate.utils.modeling import (
     dtype_byte_size,
     find_tied_parameters,
     get_balanced_memory,
+    get_max_memory,
     get_module_size_with_ties,
     get_non_persistent_buffers,
     get_state_dict_offloaded_model,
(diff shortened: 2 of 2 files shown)


arpansahu:
Disclosure: this contribution was prepared with AI assistance (GitHub Copilot CLI), and
I am flagging that explicitly rather than leaving you to guess.

The defect, the fix and the regression test were verified locally: the new test fails on
unmodified upstream and passes with the change applied, and the surrounding suite is green.

If your project would prefer not to take AI-assisted contributions, or you would rather this
were reworked and resubmitted by hand, please just close it -- no objection at all, and
apologies for the noise.

SunMarc:
already fixed !

PR #4189, opened 2026-08-28
Address non-persistent buffers by fully-qualified name when recursing

pucedoteth:
# What does this PR do?

`get_non_persistent_buffers` gained an `fqns` argument and `fsdp_utils.py` uses it, but the two older recursive call sites still compare **bare** buffer names against **fully-qualified** ones. Three related defects fall out of that, all fixed here (per the guidance to fix one small issue everywhere rather than in separate PRs).

**1. `remove_non_persistent` is a no-op under recursion.** `named_module_tensors` filters `module.named_buffers(recurse=True)` — whose names are fully qualified — against a set of bare names, so every non-persistent buffer living in a submodule survives the filter:

```python
names = [n for n, _ in named_module_tensors(model, recurse=True, remove_non_persistent=True)]
# ['sub.np_buf']  -- expected []
```

**2. `AlignDevicesHook.init_hook` raises.** It passes those same bare names to `set_module_tensor_to_device`, which resolves names relative to the module. `offload` together with `place_submodules` is reachable from `attach_align_device_hook`, which sets `place_submodules=full_offload` while passing `offload=offload` — so a model offloaded with `offload_buffers=True` and a `preload_module_classes` entry whose submodule holds a non-persistent buffer fails outright:

```
ValueError: PreloadedBlock(
  (sub): SubWithNonPersistentBuffer(...)
) does not have a parameter or a buffer named np_buf.
```

**3. Root buffers were qualified to a leading dot.** Passing `fqns=True` at those sites surfaced this: `named_modules()` yields the root as `""`, so `n + "." + b` produced `".buf"` — not a name any lookup accepts. Now prefixed only when the module name is non-empty.

Existing tests miss all three because `ModelForTest` and `LinearWithNonPersistentBuffers` have no non-persistent buffer below the root, and the one `remove_non_persistent=True` assertion runs without `recurse=True`, where a bare name and an FQN happen to coincide.

Fixes # (no issue filed; found while reading the recent `get_non_persistent_buffers` fix in #4116)

## Tests

Three tests added, each failing before the change and passing after:

- `tests/test_modeling_utils.py::test_get_non_persistent_buffers_fqns_are_fully_qualified` — before: `{'root_np_buf', '.root_np_buf', 'sub.np_buf'}`
- `tests/test_modeling_utils.py::test_named_module_tensors_removes_non_persistent_buffers_when_recursing` — before: `['sub.np_buf']`
- `tests/test_hooks.py::test_attach_align_device_hook_offload_buffers_with_preloaded_submodules` — before: the `ValueError` above

Run locally on CPU (torch 2.10, Python 3.12):

```
pytest tests/test_modeling_utils.py tests/test_hooks.py   ->  56 passed, 6 skipped
pytest tests/test_offload.py                              ->   4 passed
ruff check / ruff format --check (0.13.1, as pinned in setup.py)  ->  clean
```

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @muellerzr

### DIFF
--- a/src/accelerate/hooks.py
+++ b/src/accelerate/hooks.py
@@ -335,7 +335,7 @@ class AlignDevicesHook(ModelHook):
                         module, name, self.execution_device, tied_params_map=self.tied_params_map
                     )
             elif self.offload_buffers and self.execution_device is not None:
-                for name in get_non_persistent_buffers(module, recurse=self.place_submodules):
+                for name in get_non_persistent_buffers(module, recurse=self.place_submodules, fqns=True):
                     set_module_tensor_to_device(
                         module, name, self.execution_device, tied_params_map=self.tied_params_map
                     )
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -460,7 +460,7 @@ def named_module_tensors(
     if include_buffers:
         non_persistent_buffers = set()
         if remove_non_persistent:
-            non_persistent_buffers = get_non_persistent_buffers(module, recurse=recurse)
+            non_persistent_buffers = get_non_persistent_buffers(module, recurse=recurse, fqns=True)
         for named_buffer in module.named_buffers(recurse=recurse):
             name, _ = named_buffer
             if name not in non_persistent_buffers:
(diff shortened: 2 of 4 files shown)


SunMarc:
not a real issue, closing

PR #4194, opened 2026-08-30
Fix split batch sampler lengths

Marchematics:
Match BatchSamplerShard.__len__ with the batches yielded to each process when split_batches is enabled without even_batches.

Test: python3 -m py_compile src/accelerate/data_loader.py tests/test_data_loader.py

AI assistance was used.


SunMarc:
not a real issue, closing

PR #4198, opened 2026-08-31
Ask transformers whether a model supports context parallelism

qgallouedec:
Follow-up to #4177, and depends on huggingface/transformers#48442. Draft until that merges.

#4177 pattern-matches `layer_types` and `sliding_window` from inside accelerate. @SunMarc noted in review that this belongs in transformers, and huggingface/transformers#48442 adds `supports_context_parallel` for it. This swaps the inline check for that property.

It also picks up something the config check could not express: a model can set `_supports_context_parallel = False` to rule context parallelism out whatever its config, which is the case for gpt-oss (attention sinks need a custom kernel that SDPA cannot provide).

Defaults to allowed when the property is absent, so an older transformers behaves exactly as it does today.

Leaves `_refuse_recurrent_layers_under_sequence_parallelism` alone on purpose. `supports_context_parallel` is also `False` for sliding-window models, and I have not verified that those are broken under Ulysses, so reusing it there would refuse models that may well work.

### DIFF
--- a/src/accelerate/big_modeling.py
+++ b/src/accelerate/big_modeling.py
@@ -802,11 +802,8 @@ def _attach_context_parallel_hooks(
 
     """
 
-    # The hook below discards the attention mask and forces `is_causal=True`, which only matches the
-    # model's own masking for plain causal attention. `supports_context_parallel` is transformers'
-    # answer for whether that substitution is safe; it also lets a model rule context parallelism out
-    # whatever its config, which inspecting `layer_types` here could not express. Defaults to allowed,
-    # so a transformers without the property behaves as before.
+    # The hook below drops the attention mask and forces `is_causal=True`; transformers says whether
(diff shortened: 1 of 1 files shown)


SunMarc (approved):
That would be the better solution but let's see if this is worth adding it in transformers itself.

qgallouedec:
Holding this until #4209 lands, then rebuilding it on top, two changes:

- The `getattr(model, "supports_context_parallel", True)` default admits sliding-window / chunked models on every transformers without the property, i.e. every release through 5.16.1, undoing #4177. use `..._TRANSFORMERS_VERSION = "5.17.0"` constant and `compare_versions("transformers", ">=", ...)`, the property on new transformers, the current inspection otherwise.
- #4209 factors that inspection into `_non_full_attention_layer_types` and reuses it for Ulysses, where the CP property doesn't apply (SP allows sliding-window with flash attention). So this PR only swaps the CP side, and the helper stays.

PR #4203, opened 2026-09-02
Restore shuffle order when resuming a stateful DataLoader

### LINKED ISSUE #4195: `use_stateful_dataloader` under multi-process training restores the cursor, not the shuffle order — sampler permutation is never serialized
### System Info

```Shell
- accelerate `1.13.0`, torch `2.13.0`, torchdata `0.11.0`, Linux
- Reproduced on multi-GPU DDP and on 2-process CPU/Gloo (reproducer below needs no GPU)
```

### Information

- [x] My own modified scripts

### Tasks

### Reproduction

```bash
python -m torch.distributed.run --standalone --nproc_per_node=2 issue_repro.py save   /tmp/sfd_ckpt
python -m torch.distributed.run --standalone --nproc_per_node=2 issue_repro.py resume /tmp/sfd_ckpt
# on a GPU machine, prefix with: CUDA_VISIBLE_DEVICES=""
```

```python
# issue_repro.py — DDP + use_stateful_dataloader + shuffle=True:
# resume restores the right POSITION but the wrong ORDER.
import os
import sys

import torch

MODE, OUT = sys.argv[1], sys.argv[2]
RANK = int(os.environ["RANK"])
SEED, N, BS, SKIP = 1234, 64, 4, 3  # 16 batches/epoch; checkpoint after 3 batches of epoch 1

torch.manual_seed(SEED)  # identical in both phases -> reconstruction is deterministic
acc = Accelerator(cpu=True, dataloader_config=DataLoaderConfiguration(use_stateful_dataloader=True))
loader = acc.prepare(DataLoader(TensorDataset(torch.arange(N)), batch_size=BS, shuffle=True, num_workers=0))


def as_list(batch):
    return batch[0].tolist()  # TensorDataset yields a 1-tuple


if MODE == "save":
    os.makedirs(OUT, exist_ok=True)
    epoch0 = [as_list(b) for b in loader]          # consume full epoch 0
    it = iter(loader)                              # epoch 1
    consumed = [as_list(next(it)) for _ in range(SKIP)]
    state = loader.state_dict()                    # per-rank save, as in #3080/#4165
    torch.save(state, f"{OUT}/dl_rank{RANK}.bin")
    torch.save(
        {"epoch0": epoch0, "consumed": consumed, "cont": [as_list(b) for b in it]},
        f"{OUT}/ref_rank{RANK}.pt",
    )
    print(f"[rank {RANK}] saved dl state keys: {list(state.keys())}")
else:
    ref = torch.load(f"{OUT}/ref_rank{RANK}.pt", weights_only=False)
    loader.load_state_dict(torch.load(f"{OUT}/dl_rank{RANK}.bin", weights_only=False))
    resumed = [as_list(b) for b in loader]
    print(f"[rank {RANK}] batches consumed before checkpoint: {ref['consumed']}")
    print(f"[rank {RANK}] true epoch-1 remainder            : {ref['cont']}")
    print(f"[rank {RANK}] resumed stream                     : {resumed}")
    print(f"[rank {RANK}] epoch-0 remainder (for reference)  : {ref['epoch0'][SKIP:]}")
    assert resumed == ref["epoch0"][SKIP:], "not even the deterministic epoch-0 replay"
    assert resumed != ref["cont"], "resumed matched the true continuation -- bug not reproduced"
    print(f"[rank {RANK}] POSITION restored, ORDER lost: epoch-0's order at epoch-1's position")
```

### Expected behavior

Either:

- exact resume under multi-process as well (preferred — matching torchdata's
  native semantics: the restored stream equals the recorded continuation), or
- an explicit warning at save/load time that only the cursor is being restored
  when `num_processes > 1` and the sampler chain is shuffled.

Observed output (rank 0; rank 1 analogous):

```
# save
[rank 0] saved dl state keys: ['_index_sampler_state', '_sampler_iter_state', '_sampler_iter_yielded', '_num_yielded', '_IterableDataset_len_called', '_shared_seed', 'fetcher_state', 'dataset_state', '_iterator_finished']

# resume
[rank 0] batches consumed before checkpoint: [[39, 18, 10, 13], [63, 27, 2, 62], [25, 6, 37, 50]]
[rank 0] true epoch-1 remainder            : [[55, 16, 59, 0], [22, 46, 17, 57], [23, 45, 19, 51], [21, 26, 40, 36], [20, 5, 41, 32]]
[rank 0] resumed stream                     : [[15, 41, 52, 61], [5, 50, 18, 14], [31, 53, 36, 47], [35, 9, 23, 58], [22, 7, 51, 55]]
[rank 0] epoch-0 remainder (for reference)  : [[15, 41, 52, 61], [5, 50, 18, 14], [31, 53, 36, 47], [35, 9, 23, 58], [22, 7, 51, 55]]
[rank 0] POSITION restored, ORDER lost: epoch-0's order at epoch-1's position
```

YeonwooSung:
## What does this PR do?

`use_stateful_dataloader=True` restored how far the loader had walked, but not the shuffle permutation. After a completed epoch, a fresh loader rebuilt `RandomSampler` at epoch 0 and then skipped N batches of the *wrong* order. Mid-epoch resume on a new process had the same failure when the sampler's generator was not serialized.

This persists the state needed to rebuild the current epoch's permutation:

- `BatchSamplerShard` now yields a stateful iterator. `StatefulDataLoader` stores that as `_sampler_iter_state` (generator snapshot taken *before* the inner sampler consumes it, plus how many batches were yielded).
- `RandomSampler(generator=None)` gets a dedicated generator so the permutation is not tied to the caller's global RNG.
- `SeedableRandomSampler` epoch / `initial_seed` are restored.
- `DataLoaderAdapter.state_dict` also records `_accelerate_iteration` after a completed pass so `set_epoch` is applied on load.

A new loader + `load_state_dict` (or `Accelerator.save_state` / `load_state`) then sees the same remaining batches.

Fixes #4195. Related: #4196 takes a similar iterator snapshot but does not cover the epoch-boundary case or `Accelerator.save_state`.

## Tests

All under `@require_torchdata_stateful_dataloader`, using `prepare_data_loader(..., num_processes=2)` (no GPU):

- Full epoch 0, checkpoint after 3 batches of epoch 1, resume into a **new** loader — ranks 0 and 1, with and without `use_seedable_sampler`
- Mid-first-epoch resume, same matrix
- Rank 0 and rank 1 see different orders
- Resume does not depend on re-seeding torch
- `RandomSampler.generator is None`
- `save_accelerator_state` / `load_accelerator_state`

Existing stateful DataLoader tests still pass.

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -333,6 +333,27 @@ def _restore_shuffle_state(batch_sampler, state):
         sampler.generator.set_state(state["generator_state"])
 
 
+def _replace_sampler_iter_state(state_dict, new_iter_state):
+    """Point a torchdata state_dict at the start of the next epoch's permutation."""
+    if not isinstance(state_dict, dict):
+        return
+    if "_sampler_iter_state" in state_dict:
+        state_dict["_sampler_iter_state"] = new_iter_state
+    if "_num_yielded" in state_dict:
+        state_dict["_num_yielded"] = 0
+    if "_sampler_iter_yielded" in state_dict:
+        state_dict["_sampler_iter_yielded"] = 0
+    index_state = state_dict.get("_index_sampler_state")
+    if isinstance(index_state, dict) and "samples_yielded" in index_state:
+        index_state["samples_yielded"] = 0
+    snapshot = state_dict.get("_snapshot")
+    if isinstance(snapshot, dict):
+        _replace_sampler_iter_state(snapshot, new_iter_state)
+        main = snapshot.get("_main_snapshot")
+        if isinstance(main, dict):
+            _replace_sampler_iter_state(main, new_iter_state)
+
+
 class _BatchSamplerShardIterator:
     """Stateful iterator over a `BatchSamplerShard`.
 
@@ -608,12 +629,22 @@ class DataLoaderAdapter:
             self.dl_state_dict = self.base_dataloader.state_dict()
             # Potentially modify the state_dict to adjust for prefetching
             self.adjust_state_dict_for_prefetch()
-            # Then tag if we are at the end of the dataloader
-            self.dl_state_dict["_iterator_finished"] = self.end_of_dataloader
-            # Persist the sampler epoch so a fresh loader does not restart at epoch 0.
             iteration = getattr(self, "iteration", 0)
-            if iteration:
-                self.dl_state_dict[_ACCELERATE_ITERATION] = iteration
+            if getattr(self, "end_of_dataloader", False):
+                # Last batch of this epoch: persist the *next* epoch. The finished
+                # iterator still holds the pre-epoch shuffle snapshot, which would
+                # replay the epoch that just ended.
+                self.dl_state_dict["_iterator_finished"] = False
+                self.dl_state_dict[_ACCELERATE_ITERATION] = iteration + 1
+                batch_sampler = getattr(self, "batch_sampler", None)
+                if batch_sampler is None:
+                    batch_sampler = getattr(self, "sampler", None)
+                post = _capture_shuffle_state(batch_sampler) if batch_sampler is not None else {}
+                _replace_sampler_iter_state(self.dl_state_dict, {"yielded": 0, "shuffle_state": post})
+            else:
+                self.dl_state_dict["_iterator_finished"] = False
+                if iteration:
+                    self.dl_state_dict[_ACCELERATE_ITERATION] = iteration
 
 
 class DataLoaderShard(DataLoaderAdapter, DataLoaderStateMixin):
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -82,8 +82,10 @@ class SeedableRandomSampler(RandomSampler):
     Needed specifically in distributed cases, when the random generator for each GPU needs to start from the same seed
     and be fully reproducible on multiple iterations.
 
-    If a custom `generator` is passed, it will rely on its initial seed as well as the current iteration it is on
-    (stored in `self.epoch`).
+    The permutation for an epoch is `manual_seed(self.epoch + self.initial_seed)`. `set_epoch` only assigns
+    `self.epoch`; `__iter__` never increments it. Callers (including [`DataLoaderShard`]) must call `set_epoch`
+    between passes. An implicit increment during iteration would shift the next permutation when a completed-epoch
+    checkpoint is restored, because the restore path may replay one extra full sampler cycle.
     """
 
     def __init__(self, *args, **kwargs):
@@ -105,10 +107,9 @@ class SeedableRandomSampler(RandomSampler):
         # print("Setting seed at epoch", self.epoch, seed)
         self.generator.manual_seed(seed)
         yield from super().__iter__()
-        self.set_epoch(self.epoch + 1)
 
     def set_epoch(self, epoch: int):
-        "Sets the current iteration of the sampler."
+        "Sets the current iteration of the sampler. Does not increment; assignment only."
         self.epoch = epoch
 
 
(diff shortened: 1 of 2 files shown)


GoldenStain:
### Additional Findings & Verification on Stateful Dataloader Shuffle Order

Thanks @YeonwooSung for opening PR #4203! We have done extensive empirical probing across single-process, 2-process Gloo, and multi-GPU setups around exact shuffle continuation. Here are a few important edge cases and observations from our test matrix that may be helpful for review and test coverage in #4203:

#### 1. How restore proceeds without sampler state (and why single-process minimal repros can pass by coincidence)
In vanilla Accelerate + torchdata without sampler state persistence, torchdata falls back to `itertools.islice(iter(index_sampler), consumed)`. This draws a *fresh* permutation from the global RNG at restore time:
- In a minimal single-process test (same seed, no intermediate RNG consumers), the RNG draw sequence coincidentally matches the original run's epoch-start draw.
- In real multi-epoch training (model initialization, a second evaluation dataloader, transform/augmentation RNG draws), the RNG consumption diverged wildly, and the restored permutation was completely scrambled.

#### 2. Epoch-boundary checkpoints and auto-incrementing samplers
When a checkpoint is saved on the final microbatch / after a completed epoch (where torchdata records `_iterator_finished=True`), torchdata's restore logic first fast-forwards through the *finished* epoch (one extra full sampler cycle) before beginning the new epoch.
- Any sampler whose `__iter__` or iterator auto-increments its epoch (e.g. `SeedableRandomSampler` doing `self.epoch += 1`) gets shifted by an extra epoch permutation.
- A robust order-restoring sampler must have a `set_epoch(epoch)` that **only assigns** and never auto-increments implicitly during iteration/restore.

#### 3. `SeedableRandomSampler` epoch-0 swap timing
In Accelerate 1.13.0, `DataLoaderShard` constructor / adapter initialization materialized an initial iterator before `dataloader.set_sampler(sampler)` replaced the inner sampler.
- A checkpoint taken mid-epoch-0 was sampled with the vanilla `RandomSampler`, so `use_seedable_sampler=True` alone only worked for checkpoints at or after epoch 1.
- PR #4203's patch resetting `base._iterator = None` on `use_seedable_sampler` fixes this exact gap.

#### 4. Summary of Required Test Bars
To ensure complete robustness, any shuffle restoration fix should verify all four scenarios under DDP:
1. Mid-epoch-0 checkpoint resume (clean process)
2. Mid-epoch-N ($N \ge 1$) checkpoint resume
3. Epoch-boundary / completed-epoch checkpoint resume ($N \to N+1$)
4. Resumed stream matches the uninterrupted reference stream **bit-for-bit** across all ranks without requiring the user to manually re-seed torch before recreating the loader.

---

*(Note: We also observed a separate data partition defect when `num_workers > 0` under DDP due to prepare-time iterator prefetch before cross-rank RNG sync, which is tracked separately in #4204).*

YeonwooSung:
Thanks for the detailed restore matrix — this is exactly the extra cycle we were missing.

**Sampler contract.** `SeedableRandomSampler.__iter__` used to call `set_epoch(self.epoch + 1)` after a full pass. Combined with torchdata replaying a finished epoch when `_iterator_finished=True`, that shifted the next permutation by one extra epoch. `set_epoch` is now assignment-only; `DataLoaderShard` already calls it at the start of each pass. Completed-epoch snapshots persist `iteration + 1` explicitly so restore does not depend on that implicit increment.

**Tests.** Resume is now checked against a never-interrupted reference stream after extra RNG consumption (a second shuffled eval loader, `torch.randn`, `torch.randperm`) and without the caller re-seeding torch. The four bars run for both ranks under `prepare_data_loader(..., num_processes=2)`, with and without `use_seedable_sampler`:

1. Mid-epoch-0
2. Mid-epoch-N (N = 1)
3. Epoch-boundary / completed-epoch (N → N+1)
4. Bit-for-bit match of the concatenated resumed stream against the reference

The epoch-0 `base._iterator = None` swap after `set_sampler` is unchanged. The `num_workers > 0` partition issue remains out of scope here (tracked in #4204).

PR #4206, opened 2026-09-02
Fix FSDP2/ PEFT/`FULL_STATE_DICT` dropping every rank's adapter shard except rank 0

AmineDiro:
On FSDP2 with `fsdp_state_dict_type=FULL_STATE_DICT` and a PEFT model, `save_fsdp_model(..., adapter_only=True)` writes only rank 0's slice of each adapter tensor, and `load_fsdp_model(..., adapter_only=True)` restores nothing on the other ranks. 

On N GPUs, resuming reverts `(N-1)/N` of the adapter to its init — `lora_B` to zero, `lora_A` to random.

`transformers.Trainer` always hits this: `get_fsdp_ckpt_kwargs()` returns `{"adapter_only": True}` unconditionally whenever `save_fsdp_model` accepts it.

## Repro

LoRA `r=32`, `target_modules="all-linear"` on `trl-internal-testing/small-Qwen2ForCausalLM-2.5`, FSDP2, 2×H100. Instead of training, each rank fills its **own local slice** of every adapter tensor with a rank-specific value — rank 0 gets `lora_A=0.03`/`lora_B=0.01`, rank 1 gets `0.04`/`0.02` — so a slice restored from the wrong rank is unmistakable. 

Then: save → fresh model → zero the adapter → load → measure. The zeroing matters; without it a value that was simply never overwritten looks like a successful load.

| config | rank 0 after load | rank 1 after load | artifact | size | `lora_B` values in file |
|---|---|---|---|---|---|
| **FULL + `adapter_only`** ← what Trainer uses | 0.03 / 0.01 ✅ | **0.00 / 0.00** ❌ lost | `pytorch_model_fsdp.bin` | 191,949 B | **`[0.01]`** — rank 1's `0.02` absent |
| FULL, no flag | 0.03 / 0.01 ✅ | 0.04 / 0.02 ✅ | `pytorch_model_fsdp.bin` | 156,173,213 B | `[0.01, 0.02]` |
| SHARDED + `adapter_only` | 0.03 / 0.01 ✅ | 0.04 / 0.02 ✅ | `pytorch_model_fsdp_0/` (2 × `.distcp`) | 459,168 B | `[0.01, 0.02]` |
| SHARDED, no flag | 0.03 / 0.01 ✅ | 0.04 / 0.02 ✅ | `pytorch_model_fsdp_0/` | 156,363,611 B | `[0.01, 0.02]` |

The saved entries are ungathered `DTensor`s holding just the local half:

```
FULL + adapter_only:  ...layers.0.self_attn.q_proj.lora_B.weight
    type=DTensor  shape=(128, 32)  placements=(Shard(dim=0),)  to_local().shape=(64, 32)  distinct=[0.01]
FULL + no flag:       ...layers.0.self_attn.q_proj.lora_B.default.weight
    type=Tensor   shape=(128, 32)                                                         distinct=[0.01, 0.02]
```

`load_fsdp_model` even reports the failure in-band and it gets thrown away: rank 1 lists every
`lora_A`/`lora_B` key as missing while rank 0 lists only base-model keys. No caller looks at the
returned `_IncompatibleKeys`.

## The fix
- **Save** — `ignore_frozen_params=True` drops the frozen weights and gathers the rest. In a LoRA run the trainable tensors *are* the adapter.
- **Load** — `strict=False` because the file holds the adapter only. Torch broadcasts from rank 0 and re-slices on every rank.
- `ignore_frozen_params` filters *before* the gather, so the frozen base model is never materialized on rank 0.

All five runs (the four above plus the no-zeroing resume) now come back correct on both ranks, the file holds `[0.01, 0.02]`, and `load_fsdp_model` reports no missing LoRA keys on either rank.

## Tests

Two layers: one cheap enough to run on every PR, one that needs 2 GPUs.

- **`FSDP2PeftStateDictTest`** (`tests/fsdp/test_fsdp.py`) — single process, CPU, runs on every PR. Can't check sharding, but asserts the adapter state dict is keyed by the model's own FQNs, which is only true if the `sd_options` path was taken. A second test round-trips the adapter, which guards `strict=False`. Reverting only `fsdp_utils.py` turns the pair from 2 passed into 1 failed.
- **`tests/fsdp/peft_checkpointing.py`** — the only test that proves each rank's shard survives, since the bug can't appear on one rank. Every rank fills its own slice with a distinct value, saves, rebuilds, zeroes, loads, and checks it got its own value back. `FSDP2IntegrationTest.test_peft_checkpointing` runs it across both state-dict types × `adapter_only` on/off. `@slow` + `@require_multi_device` like the FSDP tests around it, so it runs in the nightly `multi_gpu` job. Against the unpatched code: `rank 1 did not get its shard of ...lora_A.default.weight back: expected 2.0, got values in [0.0, 0.0]`.
- `peft` is the only added dependency, test-only, in `test_dev`. Unavoidable — the branch is gated on `is_peft_model`, so a test has to build a real `PeftModel` to reach it.
- Heads up: the nightly job installs with `--no-deps` and gets its dependencies from the prebuilt GPU image, so both tests skip until that image is rebuilt with the new extra.

### DIFF
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -20,6 +20,7 @@ import warnings
 from collections import defaultdict
 from collections.abc import Iterable
 from contextlib import nullcontext
+from dataclasses import replace
 from pathlib import Path
 from typing import Callable, Union
 
@@ -55,6 +56,13 @@ def disable_fsdp_ram_efficient_loading():
 
 def _get_model_state_dict(model, adapter_only=False, sd_options=None):
     if adapter_only and is_peft_model(model):
+        if sd_options is not None and sd_options.full_state_dict:
+            from torch.distributed.checkpoint.state_dict import get_model_state_dict
+
+            # NOTE: Under FSDP2 `get_peft_model_state_dict` returns sharded `DTensor`s
+            #  that it has no handling for.
+            return get_model_state_dict(model, options=replace(sd_options, ignore_frozen_params=True))
+
         from peft import get_peft_model_state_dict
 
         return get_peft_model_state_dict(model, adapter_name=model.active_adapter)
@@ -70,6 +78,12 @@ def _get_model_state_dict(model, adapter_only=False, sd_options=None):
 
 def _set_model_state_dict(model, state_dict, adapter_only=False, sd_options=None):
     if adapter_only and is_peft_model(model):
+        # Invariant: `sd_options` is not None only for FSDP2
+        if sd_options is not None and sd_options.full_state_dict:
+            from torch.distributed.checkpoint.state_dict import set_model_state_dict
+
+            return set_model_state_dict(model, state_dict, options=replace(sd_options, strict=False))
+
         from peft import set_peft_model_state_dict
 
         return set_peft_model_state_dict(model, state_dict, adapter_name=model.active_adapter)
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -59,8 +59,6 @@ def _get_model_state_dict(model, adapter_only=False, sd_options=None):
         if sd_options is not None and sd_options.full_state_dict:
             from torch.distributed.checkpoint.state_dict import get_model_state_dict
 
-            # NOTE: Under FSDP2 `get_peft_model_state_dict` returns sharded `DTensor`s
-            #  that it has no handling for.
             return get_model_state_dict(model, options=replace(sd_options, ignore_frozen_params=True))
 
         from peft import get_peft_model_state_dict
--- a/src/accelerate/utils/fsdp_utils.py
+++ b/src/accelerate/utils/fsdp_utils.py
@@ -76,7 +76,6 @@ def _get_model_state_dict(model, adapter_only=False, sd_options=None):
 
 def _set_model_state_dict(model, state_dict, adapter_only=False, sd_options=None):
     if adapter_only and is_peft_model(model):
-        # Invariant: `sd_options` is not None only for FSDP2
         if sd_options is not None and sd_options.full_state_dict:
             from torch.distributed.checkpoint.state_dict import set_model_state_dict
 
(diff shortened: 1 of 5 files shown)


src/accelerate/utils/fsdp_utils.py
```diff
@@ -55,6 +56,13 @@ def disable_fsdp_ram_efficient_loading():
 
 def _get_model_state_dict(model, adapter_only=False, sd_options=None):
     if adapter_only and is_peft_model(model):
+        if sd_options is not None and sd_options.full_state_dict:
+            from torch.distributed.checkpoint.state_dict import get_model_state_dict
+
+            # NOTE: Under FSDP2 `get_peft_model_state_dict` returns sharded `DTensor`s
+            #  that it has no handling for.
```
```suggestion
```

src/accelerate/utils/fsdp_utils.py
```diff
@@ -70,6 +78,12 @@ def _get_model_state_dict(model, adapter_only=False, sd_options=None):
 
 def _set_model_state_dict(model, state_dict, adapter_only=False, sd_options=None):
     if adapter_only and is_peft_model(model):
+        # Invariant: `sd_options` is not None only for FSDP2
+        if sd_options is not None and sd_options.full_state_dict:
+            from torch.distributed.checkpoint.state_dict import set_model_state_dict
+
+            return set_model_state_dict(model, state_dict, options=replace(sd_options, strict=False))
+
```
```suggestion
        if sd_options is not None and sd_options.full_state_dict:
            from torch.distributed.checkpoint.state_dict import set_model_state_dict

            return set_model_state_dict(model, state_dict, options=replace(sd_options, strict=False))

```

tests/fsdp/peft_checkpointing.py
```diff
@@ -0,0 +1,122 @@
+# Copyright 2025 The HuggingFace Team. All rights reserved.
```
SunMarc:
```suggestion
# Copyright 2026 The HuggingFace Team. All rights reserved.
```

SunMarc (approved):
Thanks a lot for this ! It would have been better that the fix happens in peft but happy to deal with this edge case here is it is simpler !

PR #4208, opened 2026-09-02
Fix `save_model` dropping root-level parameters and buffers for offloaded models

Kayvan-Zahiri:
# What does this PR do?

`get_state_dict_offloaded_model` skips the root module (`if name == "": continue`), so parameters and buffers registered directly on it are collected by nobody and vanish. `Accelerator.save_model` routes to this function for any offloaded model, so it writes an incomplete checkpoint with no warning. On a module with `cls_token`/`pos_embed` on the root, after `disk_offload`:

```
expected: ['blocks.0.bias', 'blocks.0.weight', 'cls_token', 'head.bias', 'head.weight', 'pos_embed']
saved   : ['blocks.0.bias', 'blocks.0.weight', 'head.bias', 'head.weight']
```

Including the root exposes a second bug. Offloading appends an `io_same_device` hook, so the root carries a `SequentialHook`; `has_offloaded_params` returns `False` for it and `align_module_device` then raises `ValueError: cls_token is on the meta device`. Both now resolve the offloading `AlignDevicesHook` through one helper that unwraps `SequentialHook`.

Keys and values now round-trip exactly for `cpu_offload`, `disk_offload` and `dispatch_model`. `ModelForTest` and `NestedModelForTest` register nothing on the root, which is why the existing test could not catch this, so I added a model that does.

## Before submitting

The three new/extended tests fail on `main` and pass here. `tests/test_modeling_utils.py`, `test_utils.py`, `test_offload.py`, `test_big_modeling.py`, `test_hooks.py` and `test_accelerator.py` give the same 7 failures before and after (unrelated local toolchain issues), 151 passed before and 152 after.

## Who can review?

@SunMarc @BenjaminBossan

### DIFF
--- a/src/accelerate/utils/modeling.py
+++ b/src/accelerate/utils/modeling.py
@@ -1750,8 +1750,8 @@ def get_state_dict_offloaded_model(model: nn.Module):
     state_dict = {}
     placeholders = set()
     for name, module in model.named_modules():
-        if name == "":
-            continue
+        # the root module is included so that parameters and buffers registered directly on it are saved
+        prefix = f"{name}." if name else ""
 
         try:
             with align_module_device(module, "cpu"):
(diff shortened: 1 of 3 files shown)


SunMarc:
Not a real issue, did this happened while saving a real model you worked on ?  if so, update the PR description. No need to create an example just for the sake of the PR.

PR #4209, opened 2026-09-02
Add an FSDP2-native Ulysses sequence parallelism backend

qgallouedec:
# What does this PR do?

Adds `sp_backend="torch"`: Ulysses sequence parallelism on FSDP2, no DeepSpeed. Redo of #4084. DeepSpeed backend unchanged.

Why: torch CP only takes SDPA with a causal mask. Ulysses keeps the model's own attention call, so flash attention, packing and sliding windows work.

How:
- `sp` is a device mesh dim, folded into the FSDP sharding mesh like `cp`.
- Same hook as DeepSpeed ALST: the model's key in `ALL_ATTENTION_FUNCTIONS` is wrapped with all-to-all (scatter heads, gather sequence) around the original attention. Only prepared modules are affected, a second model in the process keeps plain attention.
- Batch sharding reuses `maybe_context_parallel`, so a CP loop works unchanged.
- Unset `sp_backend` resolves to `deepspeed` under DeepSpeed, `torch` otherwise.
- Refused with a clear error: eager/flex, sliding window under sdpa, linear attention / Mamba / conv layers, `sp_size` not dividing the heads.

Tested against an unsharded forward/backward (logits, loss, grads), 4xH100:
- sdpa and hub flash-attn2, packed and unpacked, sp=2/4, sp=2 x dp_shard=2
- bit-identical to the DeepSpeed backend on the same batch
- Llama 3.2, Qwen3, Phi-3, Qwen3-MoE, GLM-4-MoE, DeepSeek-V3, Cohere2, Gemma3, gpt-oss, Qwen2.5-VL, Qwen3-VL, Idefics3 (text)
- Qwen3-0.6B @16k per GPU: 70.9 GiB / 595 ms on 1 GPU vs 33.9 GiB / 328 ms with sp=2 (deepspeed: 37.0 GiB / 351 ms), same loss curve

Found on the way: huggingface/transformers#48491 (Qwen3-VL and Gemma3 VLM ignore `shift_labels`).

Differences with #4084, and why:
- Wrapper only acts on prepared modules instead of rewriting the attention dict for the whole process: a second model (DPO/RL reference) keeps plain attention.
- `position_ids` are gathered rather than dropped in favour of `cu_seqlens`: sdpa cannot use `cu_seqlens`, so packing broke there. Packed sdpa and FA2 are now verified exact.
- No dataloader wrapper, the CP context manager is reused: keeps one loop for CP and SP. A shared adapter comes in a follow-up.
- FSDP2 only, like CP: `_prepare_deepspeed` is untouched, so the DeepSpeed path is provably unchanged.
- An explicit backend that does not match the engine raises instead of silently switching.
- Refuses what would train silently wrong (eager/flex, sliding window under sdpa, linear attention, Mamba, conv layers, non-divisible heads) and handles gpt-oss sinks and VLMs.

Follow-ups: shared CP/SP batch adapter, kv-head replication.

## Who can review?

@SunMarc


docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -213,3 +218,46 @@ For an example of an Accelerate training loop with enabled ALST/UlyssesSP see [e
...
+    # Pre-shifted labels: shifting inside a shard would drop the token that sits on the next rank.
+    batch["shift_labels"] = torch.nn.functional.pad(batch["labels"][:, 1:], (0, 1), value=-100)
+    buffers = [batch["input_ids"], batch["labels"], batch["shift_labels"], batch["position_ids"]]
+    with accelerator.maybe_context_parallel(buffers=buffers, buffer_seq_dims=[1, 1, 1, 1], no_restore_buffers=set(buffers)):
+        outputs = model(**batch)
+        loss = outputs.loss
+        ...
+```
+
+The sequence length must be a multiple of `sp_size`. `position_ids` tell attention where each token sits in the gathered sequence and where packed documents start; if the batch has none, `prepare` fills in the global positions of the shard. Padding masks are not applied: right-pad, or pack with `position_ids` (with flash attention, since `sdpa` materializes a full mask for packed batches). Aggregate the loss over the `sp` group with the token-weighted differentiable `all_gather` shown above for the DeepSpeed backend, using `accelerator.torch_device_mesh["sp"].get_group()`.
```
SunMarc, line 252: > The sequence length must be a multiple of `sp_size`.
Can't we automatically pad if this is not the case ? like i feel like this will create some issues no ?

docs/source/concept_guides/sequence_parallelism.md
```diff
@@ -213,3 +218,46 @@ For an example of an Accelerate training loop with enabled ALST/UlyssesSP see [e
...
+
+### Limitations
+
+- `sp_size` must divide the number of attention heads and of key/value heads.
+- Sliding-window layers need a flash attention implementation; with `sdpa` they are refused. Layers that mix tokens along the sequence outside of attention (chunked attention, linear attention, Mamba, LFM2 convolutions) are refused, as are `eager`, `flex_attention` and paged attention.
+- Context parallelism and sequence parallelism are mutually exclusive.
```
SunMarc, line 262: This can be fixed but not worth for now

src/accelerate/utils/dataclasses.py
```diff
@@ -2309,6 +2310,22 @@ def __post_init__(self):
...
+    runs on FSDP2 and wraps whatever attention implementation the model was loaded with (`sdpa`, `flash_attention_2`,
+    `flash_attention_3` or a hub flash kernel), so there is nothing to configure yet.
+    """
+
+    def __post_init__(self):
+        if not is_torch_version(">=", BETA_SP_AVAILABLE_PYTORCH_VERSION):
+            raise ValueError(
+                f"FSDP2-based Ulysses sequence parallelism is only available in PyTorch {BETA_SP_AVAILABLE_PYTORCH_VERSION} and later versions. "
+                "Please upgrade your PyTorch version."
+            )
```
SunMarc, line 2326: hmmm fine to block but i remember that the torch impl could be used for all, like ddp, fsdp or deepspeed.

src/accelerate/utils/sequence_parallel.py
```diff
@@ -0,0 +1,236 @@
...
+            raise ValueError(
+                f"Attention implementation `{attn_implementation}` is not registered in transformers' "
+                "`ALL_ATTENTION_FUNCTIONS`, so Ulysses sequence parallelism cannot wrap it. "
+                f"Valid implementations are {ALL_ATTENTION_FUNCTIONS.valid_keys()}."
+            )
+        ulysses_attention = UlyssesAttention(attn_implementation, ALL_ATTENTION_FUNCTIONS[attn_implementation])
+        ALL_ATTENTION_FUNCTIONS.register(attn_implementation, ulysses_attention)
+        _ulysses_attention_functions[attn_implementation] = ulysses_attention
+    for module in model.modules():
+        ulysses_attention.register_module(module, group, model)
```
SunMarc, line 200: nice trick to make sure that the original attn is not corrupted if someone uses another model for example

src/accelerate/utils/sequence_parallel.py
```diff
@@ -0,0 +1,236 @@
...
+            attention_mask = _packed_causal_mask(kwargs["position_ids"])
+
+        attn_output, attn_weights = self.attention_function(module, query, key, value, attention_mask, **kwargs)
+
+        # [batch, seq, heads / sp_size, head_dim] -> [batch, local_seq, heads, head_dim]
+        attn_output = _SeqAllToAll.apply(group, attn_output, 1, 2)
+        return attn_output, attn_weights
+
+
+_ulysses_attention_functions: dict[str, UlyssesAttention] = {}
```
SunMarc, line 174: nice trick to make sure that calling `register_ulysses_attention` twice won't modify again the ulysses attention.

src/accelerate/big_modeling.py
```diff
@@ -846,3 +853,89 @@ def _self_attn_pre_forward_hook(_module, module_args, module_kwargs):
...
+                tokens = module_kwargs.get("inputs_embeds")
+            if tokens is None and module_args:
+                tokens = module_args[0]
+            if tokens is not None:
+                local_seq_len = tokens.shape[1]
+                module_kwargs["position_ids"] = torch.arange(
+                    sp_rank * local_seq_len, (sp_rank + 1) * local_seq_len, device=tokens.device
+                ).unsqueeze(0)
+        ulysses_attention.set_position_ids(model, module_kwargs.get("position_ids"))
+        return module_args, module_kwargs
```
SunMarc, line 955: explain what happens here

src/accelerate/accelerator.py
```diff
@@ -1670,6 +1681,20 @@ def _prepare_cp(self, *args):
...
+        # Ulysses on FSDP2: `sp` is a mesh dimension folded into the FSDP sharding group (see
+        # `ParallelismConfig.dp_shard_cp_dim_names`), so FSDP's reduce-scatter sums the partial gradients of the
+        # sequence parallel ranks and nothing else has to be done outside of attention.
+        sp_mesh = self.torch_device_mesh["sp"]
+        self._sp_context = functools.partial(sequence_parallel, mesh=sp_mesh)
+
+        for arg in args:
+            if isinstance(arg, torch.nn.Module):
+                _refuse_recurrent_layers_under_sequence_parallelism(arg)
+                _attach_sequence_parallel_hooks(arg, sp_mesh)
```
SunMarc: maybe you can have a _prepare_sp_model function here. And split a bit what we do in _refuse_recurrent_layers_under_sequence_parallelism and _attach_sequence_parallel_hooks into functions. Asking this because there is a lot of things mixed like validation, attn update and hooks

src/accelerate/parallelism_config.py
```diff
@@ -55,8 +56,10 @@ class ParallelismConfig:
             Which CP backend to use: `torch` (FSDP2)
         sp_size (`int`, defaults to `1`):
             The size of the sequence parallel group.
-        sp_backend (`str`, defaults to `deepspeed`):
-            Which SP backend to use:`deepspeed` (ALST/Ulysses)
+        sp_backend (`str`, *optional*):
+            Which SP backend to use: `deepspeed` (ALST/Ulysses, requires DeepSpeed) or `torch` (Ulysses on FSDP2,
+            with `sp` as a device mesh dimension). If unset, resolves to `deepspeed` under DeepSpeed and `torch`
+            otherwise; an explicit value that does not match the distributed type raises.
```
SunMarc, line 62: no particular preference but i named it sp_backend="accelerate" in my PR because "torch" was a bit broad. Like the deepspeed impl also uses torch. But happy to use "torch".

src/accelerate/parallelism_config.py
```diff
@@ -134,12 +138,18 @@ def non_dp_dim_names(self):
 
     @property
     def dp_shard_cp_dim_names(self):
-        """Names of enabled dimensions which will be flattened into a joint mesh across which is model sharded in FSDP."""
+        """
+        Names of enabled dimensions which will be flattened into a joint mesh across which is model sharded in FSDP.
+        Context and sequence parallel ranks hold different parts of the same batch, so their gradients are partial
+        and FSDP's reduce-scatter across this joint mesh is what sums them.
+        """
```
SunMarc, line 145: i thought about renaming it at some point since it is a bit confusing that sp is not in the name with the sp PR. But that's fine xD

src/accelerate/state.py
```diff
@@ -971,6 +971,8 @@ def __init__(
                         self.downcast_bfloat = False
             elif os.environ.get("ACCELERATE_USE_DEEPSPEED", "false").lower() == "true" and not cpu:
                 self.distributed_type = DistributedType.DEEPSPEED
+                if self.parallelism_config is not None:
+                    self.parallelism_config._resolve_sp_backend(self.distributed_type)
```
SunMarc: I prefer to not do that in the state if possible. Can you check my PR https://github.com/huggingface/accelerate/pull/4084/changes ? I create a `_resolve_backends` that we put in `_validate_accelerator` to swith to the right config if the user didn't set the correct one.

src/accelerate/state.py
```diff
@@ -991,6 +993,8 @@ def __init__(
                 DistributedType.MULTI_HPU,
                 DistributedType.MULTI_NEURON,
             ]:
+                if self.parallelism_config is not None:
+                    self.parallelism_config._resolve_sp_backend(self.distributed_type)
```
SunMarc: same here

src/accelerate/state.py
```diff
@@ -1005,6 +1009,19 @@ def __init__(
...
+                    and self.parallelism_config.sp_backend == "torch"
+                ):
+                    if fsdp_plugin is None:
+                        raise ValueError(
+                            "`sp_size > 1` with `sp_backend='torch'` specified in the `parallelism_config`, but no `fsdp_plugin` was provided. We need a `fsdp_plugin` to use Ulysses sequence parallelism with `sp_backend=torch`, as we also shard the model across the device mesh to save more memory"
+                        )
+                    if fsdp_plugin.fsdp_version == 1:
+                        raise ValueError(
+                            "Using `sp_size>1` with `sp_backend='torch'` requires FSDP2, but the provided `fsdp_plugin` is using FSDP1. "
+                        )
```
SunMarc: tbh, i feel like all these check should live in `_validate_accelerator`. it is bloating the state too much with things that are only about `parallel_config` validation

src/accelerate/utils/sequence_parallel.py
```diff
@@ -0,0 +1,236 @@
+# Copyright 2026 The HuggingFace Inc. team. All rights reserved.
```
SunMarc, line 1: can we rename this to sequence_parallel.py ?

src/accelerate/parallelism_config.py
```diff
@@ -304,9 +318,8 @@ def __post_init__(self):
                         f"ParallelismConfig's cp_backend={self.cp_backend} requires {cp_backends_config_map[self.cp_backend]}, but cp_handler was set to {type(self.cp_handler)}"
                     )
 
-        if self.sp_size > 1:
-            if self.sp_handler is None:
-                self.sp_handler = DeepSpeedSequenceParallelConfig()
+        if self.sp_size > 1 and self.sp_backend is not None:
+            self._init_sp_handler()
```
SunMarc, line 327: we should init a default config at least i feel like, either the torch of the deepspeed. This config can be changed later on depending on the distributed type easily.

src/accelerate/utils/ulysses.py
```diff
@@ -0,0 +1,236 @@
...
+        s_aux = kwargs.get("s_aux")
+        if isinstance(s_aux, torch.Tensor) and s_aux.ndim == 1:
+            kwargs["s_aux"] = s_aux.chunk(dist.get_world_size(group))[dist.get_rank(group)]
+
+        # The model built `attention_mask` for its local shard, it does not apply to the gathered sequence. Flash
+        # attention derives document boundaries from `position_ids`; SDPA gets a rebuilt mask when sequences are
+        # packed and otherwise relies on `is_causal`.
+        attention_mask = None
+        if self.attn_implementation == "sdpa":
+            attention_mask = _packed_causal_mask(kwargs["position_ids"])
```
SunMarc: just wondering, i see that you are relying on `positions_ids` for all attention (which is nice as it works for `sdpa`) but i think that we are passing wrong  "cu_seq_lens_q", "cu_seq_lens_k" no ? won't it impact flashattn attention ? if we are not using them at all, i suggest dropping them

src/accelerate/utils/ulysses.py
```diff
@@ -0,0 +1,236 @@
...
+                "Pass `position_ids` explicitly, or disable sequence parallelism for this model."
+            )
+        if position_ids.ndim == 3:
+            position_ids = position_ids[0]
+
+        # [batch, heads, local_seq, head_dim] -> [batch, heads / sp_size, seq, head_dim]
+        query = _SeqAllToAll.apply(group, query, 1, 2)
+        key = _SeqAllToAll.apply(group, key, 1, 2)
+        value = _SeqAllToAll.apply(group, value, 1, 2)
+        kwargs["position_ids"] = _gather_along_dim(position_ids, 1, group)
```
SunMarc: is it possible to not have to do an all gather here ? i didn't check what it would cost. I guess it would require to directly fetch its value from `sequence_parallel` context manager, not sure if this can be done cleanly as the user would have to pass somehow the attn object. With our own dataloader, it's easier but still a bit ugly + requires to create the dataloader.

src/accelerate/utils/ulysses.py
```diff
@@ -0,0 +1,236 @@
...
+                f"{module.__class__.__name__} was called without them and none were seen at the model's forward. "
+                "Pass `position_ids` explicitly, or disable sequence parallelism for this model."
+            )
+        if position_ids.ndim == 3:
+            position_ids = position_ids[0]
+
+        # [batch, heads, local_seq, head_dim] -> [batch, heads / sp_size, seq, head_dim]
+        query = _SeqAllToAll.apply(group, query, 1, 2)
+        key = _SeqAllToAll.apply(group, key, 1, 2)
+        value = _SeqAllToAll.apply(group, value, 1, 2)
```
SunMarc: wdyt about the impl i did. I feel like it was a bit more easy to read `GatherSeqScatterHeads` and `GatherHeadsScatterSeq` in https://github.com/huggingface/accelerate/pull/4084

SunMarc (commented):
Thanks a lot for this implementation ! Left a bunch of small comments but it should be rather easy to fix ! One question I had is does this integrate cleanly with GRPO or not. With my integration, it wasn't that practical due to the dataloader but maybe there were other blockers.

src/accelerate/utils/ulysses.py
qgallouedec: Right, good catch! FA takes passed `cu_seq_lens` over `position_ids`, so boundaries computed for the local shard would be applied to the gathered sequence silently.

Our packed FA2 test builds inputs from `position_ids` only, which is why it never caught it. Dropping the four FA kwargs after the gather, and adding a test arm that passes stale local ones.

src/accelerate/state.py
qgallouedec: Done in 0c8185a1.
moved out of state.py into a `_resolve_backends(accelerator)` at the top of `_validate_accelerator`, same shape as #4084.

src/accelerate/state.py
qgallouedec: https://github.com/huggingface/accelerate/pull/4209#discussion_r3938916188

src/accelerate/state.py
qgallouedec: https://github.com/huggingface/accelerate/pull/4209#discussion_r3938916188

src/accelerate/parallelism_config.py
qgallouedec: https://github.com/huggingface/accelerate/pull/4209#discussion_r3938916188

src/accelerate/accelerator.py
qgallouedec: Done in a004827a. `_prepare_sp_model` now does
1. refuse recurrent layers,
2. refuse unsupported attention, 
3. register the Ulysses attention, 
4. attach the hooks.

The validation that was inside `_attach_sequence_parallel_hooks` is its own function, `_refuse_unsupported_attention_under_sequence_parallelism`, and the hooks function only attaches the position_ids hook.

src/accelerate/big_modeling.py
qgallouedec: Commented in a004827a

src/accelerate/utils/ulysses.py
qgallouedec: Measured it: 120-270 us per gather, latency-bound (flat from 4k to 512k local tokens), so 9-20 ms
per step for 36 layers with recompute at 2-8 GPUs. Nothing at 1M (0.005% of the step), a few % at 8k.

| GPUs | per gather | per step (72 gathers) |
|---|---|---|
| 2 | 120-210 us | 9-15 ms |
| 8 | 160-270 us | 12-20 ms |

It's avoidable without touching the context manager: the model pre-hook already sees `position_ids`
once per forward, so gather there, stash the full tensor, and have attention read the stash instead
of gathering each layer's kwarg. 72 collectives -> 1 per step.


```python
# torchrun --nproc_per_node N bench.py
import time, torch, torch.distributed as dist
dist.init_process_group("nccl"); rank = dist.get_rank(); world = dist.get_world_size()
torch.cuda.set_device(rank)
LAYERS, PASSES = 36, 2  # forward + gradient-checkpoint recompute
for local_seq in (4096, 65536, 524288):
    t = torch.arange(local_seq, device="cuda", dtype=torch.int64).unsqueeze(0)
    out = [torch.empty_like(t) for _ in range(world)]
    for _ in range(10): dist.all_gather(out, t)  # warmup
    torch.cuda.synchronize(); dist.barrier(); t0 = time.perf_counter()
    for _ in range(LAYERS * PASSES): dist.all_gather(out, t); torch.cat(out, dim=1)
    torch.cuda.synchronize(); dt = (time.perf_counter() - t0) * 1e3
    if rank == 0:
        print(f"world={world} local_seq={local_seq:>7,} gathers/step={LAYERS*PASSES} total={dt:6.2f} ms/step "
              f"({dt/(LAYERS*PASSES)*1e3:5.0f} us each)", flush=True)
dist.destroy_process_group()
```

qgallouedec: Done in 196abb08: the pre-hook gathers once, attention reads the stash.
Measured on the same node, Qwen3-1.7B (28 layers), sp=2, 2xH100:

| seq | all_gather per step | step time |
|---|---|---|
| 8k  | 56 -> 1 | 1.538 s -> 1.539 s |
| 32k | 56 -> 1 | 13.296 s -> 13.338 s |

So it's not "a few % at 8k", they overlap compute and cost nothing measurable. The change stays for the collective count, not for speed.

src/accelerate/utils/ulysses.py
qgallouedec: Done in 07971586, with the test arm that passes stale local boundaries.

src/accelerate/utils/sequence_parallel.py
qgallouedec: Renamed to sequence_parallel.py in b89bbc1f

src/accelerate/utils/ulysses.py
qgallouedec: Fair, the four call sites read identically and mean opposite exchanges.

Took your names in f224a730: `_gather_seq_scatter_heads` / `_gather_heads_scatter_seq`, thin wrappers over the one generic all-to-all whose backward is just "swap scatter and gather".

Kept that primitive rather than two hand-permuted `Functions`: the `.transpose(1, 2)` in your backwards is the part I'd rather not carry.

docs/source/concept_guides/sequence_parallelism.md
qgallouedec: Padding has to know what each buffer is (`labels` pad with -100, `position_ids` by continuing the
count), and `sequence_parallel` takes anonymous buffers by design, same contract as torch's `context_parallel`. So I think it belongs in the collator: `pad_to_multiple_of` in TRL.

I think we might create more issue by trying to support auto-padding. Keeping the raise.

src/accelerate/utils/dataclasses.py
qgallouedec: The reason is gradients, not attention: `sp` is folded into `dp_shard_cp`, so FSDP2's reduce-scatter sums the SP ranks' partial gradients.
DDP or ZeRO would average them as independent batches, so supporting them needs a custom reduction hook for a setup that saves no memory.
Same reason `cp_size` requires FSDP2 today.

Happy to revisit if someone needs SP on DDP, but I think in practice, user maximize seq len and play with GAS.

src/accelerate/parallelism_config.py
qgallouedec: Right, that makes sense. Though I think we should keep "torch", for consistency with `cp_backend="torch"` next to it: same meaning there, the torch-native path on FSDP2. Your point applies to that one too, so I'd rather the two knobs share a
vocabulary than fix it on one side.

src/accelerate/parallelism_config.py
qgallouedec: Agreed it undersells what it holds now!
I'll leave it: it's the FSDP mesh dim consumers read by name, and it matches torchtitan's name for the same joint dim. The docstring on `dp_shard_cp_dim_names` says sp is folded in and why.

qgallouedec:
thks!


> does this integrate cleanly with GRPO or not. With my integration, it wasn't that practical due to the dataloader but maybe there were other blockers.

Not today, and I don't think it needs to be: TRL's GRPOConfig raises for `sp_size > 1` because GRPO builds its inputs after generation, so nothing at the dataloader level can shard them.

But RL doesn't reach the lengths where SP matters. Without CP or SP, chunked loss plus activation offload already gets one GPU past 100k tokens, and RL tops out in the tens of thousands usually. 1M-token training is SFT / mid-training territory, and that's where the support is.

If long-rollout RL ever gets there, the torch backend makes it tractable though!: it never touches the dataloader, it shards whatever the trainer wraps in the `sequence_parallel` context, so it's TRL-side work in GRPO's loss, not an accelerate change.

PR #4213, opened 2026-09-04
Fix `estimate-memory` for timm>=1.0.29 by adding the `hf-hub:` prefix

### LINKED ISSUE #4212: `accelerate estimate-memory` fails with timm 1.0.29: model name has no source prefix
### System Info

```Shell
- `Accelerate` version: 1.15.0.dev0
- Platform: Linux-6.17.0-1032-oem-x86_64-with-glibc2.39
- `accelerate` bash location: .venv/bin/accelerate
- Python version: 3.12.3
- Numpy version: 2.5.2
- PyTorch version: 2.14.0+cpu
- PyTorch accelerator: N/A
- System RAM: 90.00 GB
	Not found

timm 1.0.29, huggingface_hub 1.30.0
```

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

```shell
pip install "timm==1.0.29"
accelerate estimate-memory timm/resnet50.a1_in1k --library_name timm
```

This is the example from the model size estimator docs. On current main it fails with:

```
Loading pretrained config for `timm/resnet50.a1_in1k` from `timm`...
  ...
  File ".../timm/models/_factory.py", line 50, in parse_model_name
    raise ValueError(
ValueError: Model name 'timm/resnet50.a1_in1k' has no source prefix but looks like a Hugging Face Hub repo id or a local path. Use 'hf-hub:timm/resnet50.a1_in1k' to load from the Hub or 'local-dir:timm/resnet50.a1_in1k' to load from a local folder.
```

The same command works with timm 1.0.28.

As far as I can tell this comes from huggingface/pytorch-image-models#2727, which shipped in timm 1.0.29. A name with a slash now needs a source prefix such as hf-hub:. Older versions dropped the repo owner and looked the rest up in the registry. accelerate passes the bare repo id to timm.create_model, so any Hub repo id fails now.

CI on main shows the same thing, for example https://github.com/huggingface/accelerate/actions/runs/33763859075/job/100676635381 (step "Run CLI tests"):

```
FAILED tests/test_cli.py::ModelEstimatorTester::test_invalid_model_name_timm - ValueError: Model name 'muellerzr/dummy' has no source prefix but looks like a Hugging Face Hub repo id or a local path. ...
FAILED tests/test_cli.py::ModelEstimatorTester::test_timm_model - ValueError: Model name 'timm/resnet50.a1_in1k' has no source prefix but looks like a Hugging Face Hub repo id or a local path. ...
```

### Expected behavior

The command should print the memory table like it does with timm 1.0.28. For a repo that is not a timm model, I would expect the usual "Tried to load ... with timm but a possible model to load was not found inside the repo." message rather than the raw timm ValueError.

I have a small fix for this and would be happy to open a PR.

iamsharduld:
# What does this PR do?

`accelerate estimate-memory timm/resnet50.a1_in1k --library_name timm` fails with timm 1.0.29:

    ValueError: Model name 'timm/resnet50.a1_in1k' has no source prefix but looks like a Hugging Face Hub repo id or a local path. Use 'hf-hub:timm/resnet50.a1_in1k' to load from the Hub or 'local-dir:timm/resnet50.a1_in1k' to load from a local folder.

Since huggingface/pytorch-image-models#2727 (shipped in timm 1.0.29), timm needs a source prefix for anything that looks like a repo id. Older versions dropped the repo owner and looked the rest up in the registry, which only worked because timm names its Hub repos after the architecture. The two timm tests in tests/test_cli.py fail on main because of this, see https://github.com/huggingface/accelerate/actions/runs/33763859075/job/100676635381.

I now add the hf-hub: prefix before calling timm.create_model when the name looks like a repo id, and leave bare architecture names like resnet50 alone. With the prefix, timm reads config.json from the Hub and raises its own errors when the file is missing or belongs to another library. I map those to the existing "Tried to load ... with timm but ..." message, so the user sees the same error as with timm 1.0.28.

Note that timm names now resolve through the Hub config.json rather than the registry entry with the same name. That means user-owned timm repos work as well.

I tested with timm 1.0.28 and 1.0.29. I added unit tests for the prefix rule and the error mapping, and one end-to-end test for a repo whose config.json belongs to another library. No version pin is needed, timm has supported the hf-hub: prefix for a long time.

Fixes #4212

## Before submitting
      Pull Request section?
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Nothing user-facing changed, the documented `timm/resnet50.a1_in1k` example works as before.

## Who can review?

@SunMarc (Command Line Interface)

### DIFF
--- a/src/accelerate/commands/estimate.py
+++ b/src/accelerate/commands/estimate.py
@@ -17,7 +17,7 @@ from typing import Optional
 
 import torch
 from huggingface_hub import model_info
-from huggingface_hub.utils import GatedRepoError, RepositoryNotFoundError
+from huggingface_hub.utils import EntryNotFoundError, GatedRepoError, RepositoryNotFoundError
 
 from accelerate import init_empty_weights
 from accelerate.commands.utils import CustomArgumentParser
@@ -52,7 +52,13 @@ def check_has_model(error):
     Checks what library spawned `error` when a model is not found
     """
     message = str(error)
-    if is_timm_available() and isinstance(error, RuntimeError) and "Unknown model" in message:
+    # `timm.create_model("hf-hub:...")` reads `config.json` from the Hub without wrapping the errors: a missing file
+    # raises `EntryNotFoundError` and a `config.json` from another library has no `architecture` key (`KeyError`)
+    if is_timm_available() and (
+        (isinstance(error, RuntimeError) and "Unknown model" in message)
+        or isinstance(error, EntryNotFoundError)
+        or (isinstance(error, KeyError) and "architecture" in message)
+    ):
         return "timm"
     elif (
         is_transformers_available()
@@ -62,6 +68,21 @@ def check_has_model(error):
         return "transformers"
     else:
         return "unknown"
+
+
+def add_timm_hub_prefix(model_name: str) -> str:
+    """
+    Adds the `hf-hub:` source prefix that `timm.create_model` needs to load `model_name` from the Hub. Bare
+    architecture names from the `timm` registry (such as `resnet50`) are returned unchanged. So are names that already
+    carry a source prefix (such as `hf-hub:timm/resnet50.a1_in1k`), which only direct callers can pass: `verify_on_hub`
+    in `create_empty_model` rejects them before this function runs.
+
+    `timm>=1.0.29` refuses a Hub repo id without the prefix. Earlier versions dropped the repo owner and resolved the
+    rest of the name through the registry.
+    """
+    if ":" in model_name or "/" not in model_name:
+        return model_name
+    return f"hf-hub:{model_name}"
 
 
 def create_empty_model(
@@ -136,7 +157,7 @@ def create_empty_model(
             )
         print(f"Loading pretrained config for `{model_name}` from `timm`...")
         with init_empty_weights():
-            model = timm.create_model(model_name, pretrained=False)
+            model = timm.create_model(add_timm_hub_prefix(model_name), pretrained=False)
     else:
         raise ValueError(
             f"Library `{library_name}` is not supported yet, please open an issue on GitHub for us to add support."
@@ -263,7 +284,7 @@ def gather_data(args):
         model = create_empty_model(
             args.model_name, library_name=args.library_name, trust_remote_code=args.trust_remote_code
         )
-    except (RuntimeError, OSError, ValueError) as e:
+    except (RuntimeError, OSError, ValueError, KeyError) as e:
         library = check_has_model(e)
         if library != "unknown":
             raise RuntimeError(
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -550,6 +565,25 @@ def test_timm_model(self):
...
+        assert add_timm_hub_prefix("local-dir:/path/to/model") == "local-dir:/path/to/model"
+
+    @require_timm
+    def test_check_has_model_timm(self):
+        # Unknown architecture in the `timm` registry
+        assert check_has_model(RuntimeError("Unknown model (dummy)")) == "timm"
+        # Hub repo without a `timm` `config.json`, raised by `timm.create_model("hf-hub:...")`
+        assert check_has_model(EntryNotFoundError("Entry Not Found for url: .../config.json.")) == "timm"
+        # Hub repo whose `config.json` belongs to another library, raised by `timm.create_model("hf-hub:...")`
+        assert check_has_model(KeyError("architecture")) == "timm"
```
SunMarc:
```suggestion
```

src/accelerate/commands/estimate.py
```diff
@@ -52,7 +52,13 @@ def check_has_model(error):
     Checks what library spawned `error` when a model is not found
     """
     message = str(error)
-    if is_timm_available() and isinstance(error, RuntimeError) and "Unknown model" in message:
+    # `timm.create_model("hf-hub:...")` reads `config.json` from the Hub without wrapping the errors: a missing file
+    # raises `EntryNotFoundError` and a `config.json` from another library has no `architecture` key (`KeyError`)
```
SunMarc:
```suggestion
```

src/accelerate/commands/estimate.py
```diff
@@ -64,6 +70,21 @@ def check_has_model(error):
...
 
+def add_timm_hub_prefix(model_name: str) -> str:
+    """
+    Adds the `hf-hub:` source prefix that `timm.create_model` needs to load `model_name` from the Hub. Bare
+    architecture names from the `timm` registry (such as `resnet50`) are returned unchanged. So are names that already
+    carry a source prefix (such as `hf-hub:timm/resnet50.a1_in1k`), which only direct callers can pass: `verify_on_hub`
+    in `create_empty_model` rejects them before this function runs.
+
+    `timm>=1.0.29` refuses a Hub repo id without the prefix. Earlier versions dropped the repo owner and resolved the
+    rest of the name through the registry.
```
SunMarc: keep it short

SunMarc (commented):
Thanks, just a couple of nits. Also, does this still works with older version of timm ?

iamsharduld:
Applied all three suggestions: dropped the unit test and the comment, and shortened the docstring.

On older timm versions: yes, it still works. The hf-hub: prefix has been accepted since timm 0.6 (the hf_hub: spelling even earlier), and I ran the estimator tests with timm 1.0.28 as well as 1.0.29 after this change.

Issue #4173, opened 2026-08-20
Circular import between accelerate.utils and accelerate.big_modeling breaks concurrent imports


adrihans:
### System Info

```Shell
accelerate: 1.14.0 (also verified present on `main`)
torch:      2.10.0
python:     3.11.11
platform:   macOS-26.5.1-arm64 (also hit on linux/arm64, python 3.11.11, in Docker)
bitsandbytes: NOT installed
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [ ] One of the scripts in the examples/ folder of Accelerate or an officially supported `no_trainer` script in the `examples` folder of the `transformers` repo (such as `run_no_trainer_glue.py`)
- [x] My own task or dataset (give details below)

### Reproduction

`accelerate` contains a circular import between `accelerate.utils` and `accelerate.big_modeling`:

```
accelerate/utils/__init__.py:226   from .bnb import has_4bit_bnb_layers, load_and_quantize_model
  accelerate/utils/bnb.py:29         from ..big_modeling import dispatch_model, init_empty_weights
    accelerate/big_modeling.py:25      from .hooks import (...)
      accelerate/hooks.py:23             from .utils import (...)     <-- back to accelerate.utils
```

Single-threaded this happens to resolve, but only by an ordering accident:
`accelerator.py:35` does `from accelerate.utils.dataclasses import FP8BackendType`
*before* `accelerator.py:37` imports `big_modeling`, so `accelerate.utils` is always
fully initialised first and the cycle never has to be broken.

That accident stops protecting you as soon as two threads import `accelerate`
submodules concurrently: they enter the cycle at different nodes, CPython's import
lock deadlock avoidance lets one of them proceed with a partially initialised
module, and the import fails.

This script fails reliably (10/10 runs here) on a clean install of accelerate 1.14.0.
No GPU, no bitsandbytes, and no other library required:

```python
"""Minimal reproducer: concurrent imports of accelerate submodules fail."""
import importlib
import threading

barrier = threading.Barrier(2)
errors = []


def do_import(name):
    barrier.wait()
    try:
        importlib.import_module(name)
    except BaseException as exc:
        errors.append(f"{name} -> {type(exc).__name__}: {exc}")


threads = [
    threading.Thread(target=do_import, args=(name,))
    for name in ("accelerate.utils", "accelerate.big_modeling")
]
for t in threads:
    t.start()
for t in threads:
    t.join()

[...]

PR #4216, opened 2026-09-04
Fix circular import between accelerate.state and accelerate.utils

### LINKED ISSUE #4215: Second circular import: accelerate.state <-> accelerate.utils (same pattern as #4173)
While looking into #4173, I found a second circular import that follows the same pattern but doesn't go through `accelerate.utils.bnb`.

`accelerate/state.py` and `accelerate/optimizer.py` both import from the aggregating `accelerate.utils` package (`from .utils import ...`). Several submodules that `accelerate/utils/__init__.py` eagerly loads import back from `accelerate.state`:

```
src/accelerate/utils/operations.py:27:from ..state import AcceleratorState, PartialState
src/accelerate/utils/modeling.py:32:from ..state import AcceleratorState
src/accelerate/utils/other.py:32:from ..state import PartialState
src/accelerate/utils/tqdm.py:22:from ..state import PartialState
src/accelerate/utils/random.py:21:from ..state import AcceleratorState
```

So `accelerate.state` depends on `accelerate.utils`, and `accelerate.utils` depends back on `accelerate.state`, the same shape as the `utils <-> big_modeling` cycle in #4173.

Single threaded imports work fine today because `accelerate/__init__.py` imports `.accelerator` first, and that chain happens to finish loading the full `utils` package before `state.py` gets imported, so the cycle never actually gets exercised. That's the same "ordering accident" described in #4173, just at a different pair of entry points. The reproducer from that issue triggers it here too, with `accelerate.state` and `accelerate.utils.operations` as the two names raced from separate threads.

I've opened a fix that has `state.py` and `optimizer.py` import from the specific leaf modules that define what they need (`utils/dataclasses.py`, `utils/environment.py`, `utils/imports.py`, `utils/operations.py`) instead of the aggregating package, the same remediation already used for `bnb.py`.

One thing worth flagging separately: `accelerate/__init__.py` imports `.accelerator` before anything else, and `accelerator.py` itself pulls in most of the rest of the package. Because of that, concurrent first imports of two different `accelerate` submodules will contend on that same shared bootstrap regardless of which specific internal cycle gets fixed, and CPython's import lock can hit its deadlock detector even without a real cycle involved once initialization takes any noticeable time. I don't think that's something a small PR can reasonably fix, since it would mean rethinking how much `accelerate/__init__.py` and `accelerator.py` load eagerly, but it seems worth a maintainer discussion since it means this bug class is likely to keep resurfacing at new entry points as more submodules get added or reordered.

Rakshit-gen:
Fixes #4215

`accelerate/state.py` and `accelerate/optimizer.py` imported from the aggregating `accelerate.utils` package. A handful of submodules that `accelerate/utils/__init__.py` eagerly loads (`operations.py`, `modeling.py`, `other.py`, `tqdm.py`, `random.py`) import `AcceleratorState` or `PartialState` back from `accelerate.state`, so `state.py` depending on the whole `utils` package created a real circular import, the same shape as the `utils <-> big_modeling` cycle from #4173.

It resolves fine in a normal single threaded import because `accelerate/__init__.py` always finishes loading `accelerator.py`, and with it the full `utils` package, before it reaches `state.py`. That ordering stops protecting you once two threads import different `accelerate` submodules at the same time.

## Changes

- `state.py` now imports `DistributedType`, `DynamoBackend`, `GradientAccumulationPlugin` and `SageMakerDistributedType` from `utils.dataclasses`, the environment helpers from `utils.environment`, and the availability checks from `utils.imports`, instead of the aggregating `utils` package.
- `optimizer.py` now imports `DistributedType` from `utils.dataclasses`, `is_lomo_available`/`is_torch_xla_available` from `utils.imports`, and `honor_type` from `utils.operations`, instead of the aggregating `utils` package.

Both are self-contained leaf modules with no import back to `accelerate.state` or `accelerate.optimizer`, so this removes the cycle without changing what either file actually uses.

## Tests

Added `CircularImportTester` to `tests/test_imports.py`. It statically parses `state.py` and `optimizer.py` and asserts neither one imports the `utils` package directly, so a future PR reintroducing `from .utils import ...` in either file will fail this test. I checked it against the pre-fix source and it fails there, and passes after this change.

Also ran the existing `tests/test_optimizer.py` and `tests/test_scheduler.py` suites and a CPU-only smoke test that prepares a model, optimizer and scheduler through `Accelerator` and runs a training step, both pass.

### DIFF
--- a/src/accelerate/optimizer.py
+++ b/src/accelerate/optimizer.py
@@ -17,7 +17,9 @@ import inspect
 import torch
 
 from .state import AcceleratorState, GradientState
-from .utils import DistributedType, honor_type, is_lomo_available, is_torch_xla_available
+from .utils.dataclasses import DistributedType
+from .utils.imports import is_lomo_available, is_torch_xla_available
+from .utils.operations import honor_type
 
 
 if is_torch_xla_available():
--- a/src/accelerate/state.py
+++ b/src/accelerate/state.py
@@ -25,15 +25,18 @@ from typing import Any, Callable
 
 import torch
 
-from .utils import (
-    DistributedType,
-    DynamoBackend,
-    GradientAccumulationPlugin,
+from .utils.dataclasses import DistributedType, DynamoBackend, GradientAccumulationPlugin, SageMakerDistributedType
+from .utils.environment import (
     check_cuda_fp8_capability,
     check_cuda_p2p_ib_support,
-    deepspeed_required,
     get_cpu_distributed_information,
     get_int_from_env,
+    parse_choice_from_env,
+    parse_flag_from_env,
+    set_numa_affinity,
+)
+from .utils.imports import (
+    deepspeed_required,
     is_datasets_available,
     is_deepspeed_available,
     is_fp8_available,
@@ -48,11 +51,7 @@ from .utils import (
     is_torch_xla_available,
     is_xccl_available,
     is_xpu_available,
-    parse_choice_from_env,
-    parse_flag_from_env,
-    set_numa_affinity,
 )
-from .utils.dataclasses import SageMakerDistributedType
 
 
 if is_torch_xla_available():
--- a/tests/test_imports.py
+++ b/tests/test_imports.py
@@ -11,9 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import ast
 import subprocess
 import sys
+import unittest
+from pathlib import Path
 
+import accelerate
 from accelerate.test_utils import require_transformer_engine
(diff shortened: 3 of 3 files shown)


Rakshit-gen:
@SunMarc could you take a look when you get a chance?

SunMarc:
not a real issue, so closing.

Rakshit-gen:
@SunMarc can you explain why?
According to me the issue is real because state.py imports from utils, while some utils modules import back from state.py, creating a circular dependency.

It works in normal imports because the current import order happens to hide the problem. With concurrent imports, that ordering isn't guaranteed and it can fail with a partially initialized module.

The fix removes this dependency entirely by importing directly from the specific utils modules. I can also add a minimal reproducer if needed.

PR #4217, opened 2026-09-04
Fix load_accelerator_state only restoring one RNG backend

### LINKED ISSUE #3960: [Bug] RNG states from multiple backends (e.g. CUDA + HPU) are saved but only one is restored on load_state
### System Info

```Shell
- `Accelerate` version: 1.13.0
- Platform: macOS-26.3-arm64-arm-64bit-Mach-O
- `accelerate` bash location: /Users/xxxxx/miniconda3/envs/accelerate/bin/accelerate
- Python version: 3.14.2
- Numpy version: 2.4.1
- PyTorch version: 2.10.0
- PyTorch accelerator: N/A
- System RAM: 16.00 GB
        Not found
```

### Information

- [ ] My own modified scripts

### Tasks

### Reproduction

The bug is visible by comparing the source code of `save_accelerator_state` and `load_accelerator_state` in `src/accelerate/checkpointing.py`.

- `save_accelerator_state` uses independent `if` statements for each backend → it saves RNG states for **all** backends that report as available.  
  → https://github.com/huggingface/accelerate/blob/main/src/accelerate/checkpointing.py#L161-L176

- `load_accelerator_state` uses a chain of `elif` (with `else` for CUDA) → only **one** branch is executed, so only **one** RNG state is restored. Others are silently ignored.  
  → https://github.com/huggingface/accelerate/blob/main/src/accelerate/checkpointing.py#L298-L313

**Current behavior**  
`save_accelerator_state` saves RNG for every detected backend.  
`load_accelerator_state` restores only one due to the `elif` chain.  
This breaks reproducibility on resume whenever more than one backend is available (e.g. both CUDA and HPU report as available), because dropout, data shuffling, augmentation randomness, etc. will diverge from the saved state.

### Expected behavior

When `accelerator.save_state()` is called and multiple accelerator backends report as available (e.g. both `is_cuda_available()` and `is_hpu_available()` return `True`), RNG states for **all** such backends should be saved into the checkpoint (which they currently are — independent `if` statements in `save_accelerator_state`).

On `accelerator.load_state()`, **all** of those saved RNG states should be restored, so that resuming training continues with exactly the same random number generator states as when the checkpoint was created.

This ensures full reproducibility across checkpoint/resume cycles, even in environments where multiple hardware backends are detectable (mixed-driver containers, cloud images with several vendor packages pre-installed, research setups, etc.).

**Minimal fix**  
Replace the `elif` chain in `load_accelerator_state` with independent `if` statements (matching the structure used in `save_accelerator_state`), e.g.:

```diff
-if is_mlu_available():
+if is_mlu_available():
    ...
-elif is_sdaa_available():
+elif is_sdaa_available():   # keep inner chain for mutually exclusive MLU/SDAA/MUSA
    ...
-elif is_musa_available():
+elif is_musa_available():
    ...
-elif is_hpu_available():
+if is_hpu_available():
    torch.hpu.set_rng_state_all(states["torch_hpu_manual_seed"])
-elif is_neuron_available():
+if is_neuron_available():
    torch.neuron.set_rng_state_all(states["torch_neuron_manual_seed"])
-else:
+if is_cuda_available():
    torch.cuda.set_rng_state_all(states["torch_cuda_manual_seed"])

### PR DESCRIPTION
**Pull request by Rakshit-gen**
Fixes #3960

`save_accelerator_state` uses independent `if` statements for each RNG backend, so it saves a seed for every backend that reports as available. On the load side, HPU, Neuron and CUDA were chained together with `elif`, ending in a bare `else` for CUDA. That causes two separate problems.

If more than one backend reports as available, only the first one in the `elif` chain actually gets its RNG state restored on `load_state`, so training resumes with a stale RNG state on the others. That's the reproducibility issue reported in #3960.

There's a second problem the `elif`/`else` structure causes that wasn't in the original report: if none of XPU, MLU, SDAA, MUSA, HPU, Neuron or CUDA are available, for example a plain CPU machine or a TPU/XLA setup, the `else` branch still runs unconditionally and tries to read a `torch_cuda_manual_seed` key that `save_accelerator_state` never wrote (since it only writes that key when `is_cuda_available()` is true). That raises a `KeyError`, which gets silently swallowed by the surrounding `try/except`, and the `xm.set_rng_state` call that comes right after in the same block never runs, so on XLA the saved XLA RNG state never actually gets restored on resume.

## Change

Switched HPU, Neuron and CUDA in `load_accelerator_state` from an `elif` chain to independent `if` statements, matching the structure `save_accelerator_state` already uses. Each backend that was actually saved gets restored now, regardless of what else is or isn't available. MLU/SDAA/MUSA stay as an `elif` chain since `save_accelerator_state` treats them the same way.

#3960 already reported the first half of this and #3988 proposed close to the same fix, but it went stale without a maintainer review and got closed along with the issue. The underlying bug is still on main, so picking that back up here with the same approach, plus a couple of regression tests that were missing before.

## Tests

Added two tests to `tests/test_state_checkpointing.py`:

- `test_rng_state_loads_without_accelerator_specific_backend`: patches every specific backend to unavailable and asserts loading a saved checkpoint logs "All random states loaded successfully" instead of silently failing.
- `test_rng_state_loads_for_every_saved_backend`: patches two backends (HPU and CUDA) as available at once and asserts both get their RNG state restored, not just the first one in the chain.

Checked both against the code before this change, they fail there and pass after. Also ran the full `tests/test_state_checkpointing.py` suite (CPU only, 22 tests), all pass.

### DIFF
--- a/tests/test_state_checkpointing.py
+++ b/tests/test_state_checkpointing.py
@@ -20,6 +20,7 @@ import shutil
 import tempfile
 import uuid
 from contextlib import contextmanager
+from unittest.mock import MagicMock, patch
 
 import pytest
 import torch
@@ -28,6 +29,7 @@ from torch import nn
 from torch.utils.data import DataLoader, TensorDataset
 
 from accelerate import Accelerator
+from accelerate.checkpointing import load_accelerator_state, save_accelerator_state
 from accelerate.test_utils import (
     DEFAULT_LAUNCH_COMMAND,
     execute_subprocess_async,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Rakshit-gen** [comment]
@SunMarc could you take a look when you get a chance?

File: tests/test_state_checkpointing.py [hunk 1]
```diff
@@ -309,6 +311,61 @@ def test_invalid_registration(self):
...
+            patch.object(torch, "hpu", hpu, create=True),
+            patch.object(torch.cuda, "get_rng_state_all", cuda_get),
+            patch.object(torch.cuda, "set_rng_state_all", cuda_set),
+        ):
+            Accelerator()
+            save_accelerator_state(tmpdir, [], [], [], [], process_index=0, step=0)
+            load_accelerator_state(tmpdir, [], [], [], [], process_index=0)
+        hpu.set_rng_state_all.assert_called_once_with("hpu-rng-state")
+        cuda_set.assert_called_once_with("cuda-rng-state")
+
```
SunMarc:
```suggestion
```

tests/test_state_checkpointing.py
```diff
@@ -28,6 +29,7 @@
 from torch.utils.data import DataLoader, TensorDataset
 
 from accelerate import Accelerator
+from accelerate.checkpointing import load_accelerator_state, save_accelerator_state
```
SunMarc:
```suggestion
```

tests/test_state_checkpointing.py
```diff
@@ -20,6 +20,7 @@
 import tempfile
 import uuid
 from contextlib import contextmanager
+from unittest.mock import MagicMock, patch
```
SunMarc:
```suggestion
```

SunMarc (approved):
SG

Issue #3911, opened 2026-01-23
Severe GPU training performance degradation due to incorrectly set `KMP_AFFINITY`


hexfaker:
GPU training performance severely degrades due to incorrectly set `KMP_AFFINITY`.

# Symptoms

- GPU idles ~80% of the time
- No bottlenecks in IO or CPU utilization - 7 out of 8 CPU cores are idle
- Training is inexplicably slow despite hardware being underutilized

# Environment
<details>
<summary>accelerate env output </summary>

```
- `Accelerate` version: 1.12.0
- Platform: Linux-6.14.0-1010-aws-x86_64-with-glibc2.39
- `accelerate` bash location: /home/ubuntu/prj/s2g-lmdm/.pixi/envs/default/bin/accelerate
- Python version: 3.10.0
- Numpy version: 1.26.4
- PyTorch version: 2.7.1
- PyTorch accelerator: CUDA
- System RAM: 61.94 GB
- GPU type: NVIDIA L40S
- `Accelerate` default config:
        - compute_environment: LOCAL_MACHINE
        - distributed_type: NO
        - mixed_precision: no
        - use_cpu: False
        - debug: False
        - num_processes: 1
        - machine_rank: 0
        - num_machines: 1
        - gpu_ids: all
        - rdzv_backend: static
        - same_network: True
        - main_training_function: main
        - enable_cpu_affinity: True
        - downcast_bf16: no
        - tpu_use_cluster: False
        - tpu_use_sudo: False
        - tpu_env: []
  ```
The data is on local fast nvme disk.
</details>

# Reason

Main training process and all DataLoader workers are pinned to CPU core 0 due to `KMP_AFFINITY` being incorrectly set.

# The Bug

In [`src/accelerate/utils/launch.py`](https://github.com/huggingface/accelerate/blob/v1.12.0/src/accelerate/utils/launch.py#L133), `ACCELERATE_USE_CPU` is set as a string:

```python
current_env["ACCELERATE_USE_CPU"] = str(args.cpu or args.use_cpu)
```

Then [on line 164](https://github.com/huggingface/accelerate/blob/v1.12.0/src/accelerate/utils/launch.py#L164), it's checked directly:

```python
if current_env["ACCELERATE_USE_CPU"]:
    current_env["KMP_AFFINITY"] = "granularity=fine,compact,1,0"
```

**The problem**: `str(False)` returns the string `"False"`, and in Python, all non-empty strings are truthy:

```python
>>> bool("False")
True
[...]

PR #4222, opened 2026-09-07
Fix accelerate launch --cpu never setting the KMP variables

yupengtang:
# What does this PR do?

`accelerate launch --cpu` no longer sets `KMP_AFFINITY` and `KMP_BLOCKTIME`.

#3912 gated those two on CPU mode with:

```python
if parse_flag_from_env(current_env["ACCELERATE_USE_CPU"], False):
```

`parse_flag_from_env` takes an environment variable *name* and does `os.environ.get(key, str(default))`. Here it is handed the *value*, the string `"True"` or `"False"`, so it looks for an environment variable literally called `True`, does not find one, and returns the default. The guard is `False` on every launch.

Before #3912 the line was `if current_env["ACCELERATE_USE_CPU"]:`, which was always `True` because `"False"` is a non-empty string, so the variables were set on GPU launches too. That was the bug reported in #3911. The fix replaced one constant with the other.

On `main`, with `prepare_simple_launcher_cmd_env`:

```
accelerate launch --cpu      ACCELERATE_USE_CPU='True'   KMP_AFFINITY=None  KMP_BLOCKTIME=None
accelerate launch (no flag)  ACCELERATE_USE_CPU='False'  KMP_AFFINITY=None  KMP_BLOCKTIME=None
```

and the mechanism, which is easy to confirm directly:

```
>>> parse_flag_from_env("True", False)
False
>>> os.environ["True"] = "yes"       # an env var actually named "True"
>>> parse_flag_from_env("True", False)
True
```

With this PR:

```
accelerate launch --cpu      ACCELERATE_USE_CPU='True'   KMP_AFFINITY='granularity=fine,compact,1,0'  KMP_BLOCKTIME='1'
accelerate launch (no flag)  ACCELERATE_USE_CPU='False'  KMP_AFFINITY=None                            KMP_BLOCKTIME=None
```

The flag is now tested the same way the line two above it already tests it (`current_env["ACCELERATE_USE_CPU"] = str(args.cpu or args.use_cpu)`), and the rest of the launch code does the same thing (`commands/launch.py` uses `args.use_cpu` and `args.cpu` directly throughout). The import that #3912 added for the broken call has no other user in the file, so it goes with it.

## Testing

`tests/test_cli.py::AccelerateLauncherTester::test_cpu_launch_sets_kmp_env` covers both directions, so neither the original #3911 behaviour nor this one can come back. It fails on `main` with `KeyError: 'KMP_AFFINITY'` and passes here.

- `pytest tests/test_cli.py`: 37 passed, 3 skipped, 1 failed. The failure is `test_accelerate_test`, which fails identically on `main` on this machine with `FileNotFoundError: 'accelerate-launch'`, since I run from source rather than an installed console script.
- `ruff check .` and `ruff format --check .` with the 0.13.1 pinned in `setup.py`: clean for these files. `ruff check` also reports one unrelated F821 in `commands/env.py` on `main`, which #4221 fixes.

Ran on Python 3.13, torch 2.11.0, Linux, CPU.

## Before submitting

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -40,7 +40,6 @@ from ..utils import (
 from ..utils.constants import DEEPSPEED_MULTINODE_LAUNCHERS
 from ..utils.other import get_free_port, is_port_in_use, merge_dicts
 from ..utils.versions import compare_versions
-from . import parse_flag_from_env
 from .dataclasses import DistributedType, SageMakerDistributedType
 
 
@@ -161,7 +160,7 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
     if (num_processes is not None and num_processes > 1) or num_machines > 1:
         current_env["MASTER_ADDR"] = args.main_process_ip if args.main_process_ip is not None else "127.0.0.1"
         current_env["MASTER_PORT"] = str(args.main_process_port) if args.main_process_port is not None else "29500"
-    if parse_flag_from_env(current_env["ACCELERATE_USE_CPU"], False):
+    if args.cpu or args.use_cpu:
         current_env["KMP_AFFINITY"] = "granularity=fine,compact,1,0"
         current_env["KMP_BLOCKTIME"] = str(1)
 
--- a/src/accelerate/utils/launch.py
+++ b/src/accelerate/utils/launch.py
@@ -160,7 +160,7 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
     if (num_processes is not None and num_processes > 1) or num_machines > 1:
         current_env["MASTER_ADDR"] = args.main_process_ip if args.main_process_ip is not None else "127.0.0.1"
         current_env["MASTER_PORT"] = str(args.main_process_port) if args.main_process_port is not None else "29500"
-    if args.cpu or args.use_cpu:
+    if str_to_bool(current_env["ACCELERATE_USE_CPU"]):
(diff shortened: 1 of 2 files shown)


src/accelerate/utils/launch.py
```diff
@@ -161,7 +160,7 @@ def prepare_simple_launcher_cmd_env(args: argparse.Namespace) -> tuple[list[str]
     if (num_processes is not None and num_processes > 1) or num_machines > 1:
         current_env["MASTER_ADDR"] = args.main_process_ip if args.main_process_ip is not None else "127.0.0.1"
         current_env["MASTER_PORT"] = str(args.main_process_port) if args.main_process_port is not None else "29500"
-    if parse_flag_from_env(current_env["ACCELERATE_USE_CPU"], False):
+    if args.cpu or args.use_cpu:
```
SunMarc:
```suggestion
    if str_to_bool(current_env["ACCELERATE_USE_CPU"]):
```

SunMarc (approved):
Added the correct fix

src/accelerate/utils/launch.py
```diff
@@ -40,7 +40,6 @@
 from ..utils.constants import DEEPSPEED_MULTINODE_LAUNCHERS
 from ..utils.other import get_free_port, is_port_in_use, merge_dicts
 from ..utils.versions import compare_versions
```
SunMarc, line 42:
```suggestion
from ..utils.versions import compare_versions
from . import str_to_bool
```

PR #3405, opened 2025-02-21
Fix AMD GPU support with should_reduce_batch_size()

cameronshinn:
Even though torch has NVIDIA and AMD GPUs operate under the cuda namespace, the out of memory error for AMD GPUs is different. When trying to determine if a model can fit on an AMD GPU, this function will evaluate to false for a `torch.OutOfMemoryError`. This PR adds another check for the error string.

Example error messge:
```
'HIP out of memory. Tried to allocate 64.00 GiB. GPU 0 has a total capacity of 63.98 GiB of which 48.63 GiB is free. Of the allocated memory 15.02 GiB is allocated by PyTorch, and 129.49 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_HIP_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)'
```

# What does this PR do?

Fixes # (issue)

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/accelerate/blob/main/CONTRIBUTING.md#submitting-a-pull-request-pr),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/accelerate/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/accelerate/tree/main/docs#writing-documentation---specification).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #4227, opened 2026-09-08
Treat MPS out-of-memory errors as OOM in find_executable_batch_size

### LINKED ISSUE #4226: find_executable_batch_size does not catch MPS out-of-memory errors
`find_executable_batch_size` re-raises on the first MPS out-of-memory error instead of lowering the batch size.

`should_reduce_batch_size` (`src/accelerate/utils/memory.py`, lines 108 to 116) matches on `" out of memory."`, with a trailing period. The MPS allocator raises:

```
RuntimeError: MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 448.00 KiB, max allowed: 242.48 MiB). Tried to allocate 256.00 MiB on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

"out of memory" is followed by " (" rather than ".", so no statement matches and the decorator re-raises. The message has been the same since at least torch v2.1.0 (`aten/src/ATen/mps/MPSAllocator.mm`), and it is a plain `RuntimeError`, not `torch.OutOfMemoryError`.

### Reproduction

accelerate 1.15.0.dev0 at 9e5d1de, torch 2.14.0, macOS arm64, Python 3.13. The watermark ratio makes the allocator refuse anything above about 240 MiB, so the OOM fires without filling the GPU.

```python
# repro.py, following the pattern in docs/source/basic_tutorials/troubleshooting.md
import sys, torch
from accelerate.utils import find_executable_batch_size

use_cpu = "--cpu" in sys.argv
accelerator = Accelerator(cpu=use_cpu)
unit = 2**25 if not use_cpu else 2**44  # 32 MiB per sample on MPS; 16 TiB per sample on CPU
tried = []

@find_executable_batch_size(starting_batch_size=32)
def inner(batch_size):
    tried.append(batch_size)
    accelerator.free_memory()
    x = torch.empty(batch_size * unit, dtype=torch.uint8, device=accelerator.device)
    return x.numel() // unit

print("device:", accelerator.device)
try:
    print("succeeded with batch_size =", inner())
except RuntimeError as e:
    print("raised", type(e).__name__ + ":", str(e)[:120])
print("batch sizes tried:", tried)
```

```
$ PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.02 PYTORCH_MPS_LOW_WATERMARK_RATIO=0.01 python repro.py
device: mps
raised RuntimeError: MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 448.00 KiB, max allowed: 242.48 MiB). Tried to all
batch sizes tried: [32]

$ python repro.py --cpu
device: cpu
succeeded with batch_size = 3
batch sizes tried: [32, 28, 25, 22, 19, 17, 15, 13, 11, 9, 8, 7, 6, 5, 4, 3]
```

The CPU allocator's message (`DefaultCPUAllocator: can't allocate memory`) is matched and the batch size steps down. On MPS the first failure propagates.

`tests/test_memory_utils.py` does not catch this. `raise_fake_out_of_memory` raises `f"{torch_device.upper()} out of memory."`, which on an MPS host is `"MPS out of memory."` and matches the `" out of memory."` statement. The real allocator message does not.

### Expected behavior

The batch size steps down on MPS the way it does on CPU and CUDA.

`transformers.Trainer` with `auto_find_batch_size=True` delegates to this function through `trainer_utils.find_executable_batch_size`, so that option cannot recover from an OOM on Apple silicon.

Adding `"MPS backend out of memory"` to `_statements` fixes it, the same way #3405 added the HIP message. Opening a PR with that change and a test.

caiotheodoro:
Fixes #4226

`should_reduce_batch_size` matches OOM errors on `" out of memory."` with a trailing period. The MPS allocator raises `RuntimeError("MPS backend out of memory (MPS allocated: ...). Tried to allocate ...")`, so nothing matched and `find_executable_batch_size` re-raised on the first attempt on Apple silicon. This adds the MPS message to `_statements`, as #3405 did for HIP, plus a test that uses the allocator's real message.

Checked on macOS arm64 with torch 2.14.0 and `PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.02` to force the OOM early. Before: the decorated function raised at batch size 32. After: it stepped from 32 down to 7 and succeeded. On CPU the sequence is unchanged. Over 12 sample exception messages (CUDA, HIP, XPU, CPU, HPU, cuDNN, two MPS variants, the test helper's string, an unrelated `RuntimeError`, a `ValueError`, and a two-arg `RuntimeError`) only the two MPS messages change, from `False` to `True`.

```
pytest tests/test_memory_utils.py -q   # 8 passed; the new test fails on main
make quality                            # clean
```

Drafted with Claude Opus / Fable 5.1. Reviewed by Muse Spark 1.3 and GLM 5.3 as judges before submission.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@SunMarc @BenjaminBossan

### DIFF
--- a/tests/test_memory_utils.py
+++ b/tests/test_memory_utils.py
@@ -126,24 +126,6 @@ class MemoryTest(unittest.TestCase):
         ]
         assert [bs, arg1] == [8, "hello"]
 
-    def test_mps_out_of_memory(self):
-        batch_sizes = []
-
-        @find_executable_batch_size(starting_batch_size=16)
-        def mock_training_loop_function(batch_size):
-            nonlocal batch_sizes
-            batch_sizes.append(batch_size)
-            if batch_size != 8:
-                # Message raised by the MPS allocator, which has no trailing period after "out of memory".
-                raise RuntimeError(
-                    "MPS backend out of memory (MPS allocated: 8.00 GB, other allocations: 384.00 KB, max allowed: "
-                    "9.07 GB). Tried to allocate 2.00 GB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 "
-                    "to disable upper limit for memory allocations (may cause system failure)."
-                )
-
-        mock_training_loop_function()
-        assert batch_sizes == [16, 14, 12, 10, 9, 8]
-
     def test_start_zero(self):
         @find_executable_batch_size(starting_batch_size=0)
         def mock_training_loop_function(batch_size):
--- a/tests/test_memory_utils.py
+++ b/tests/test_memory_utils.py
@@ -126,6 +126,24 @@ class MemoryTest(unittest.TestCase):
         ]
         assert [bs, arg1] == [8, "hello"]
 
+    def test_mps_out_of_memory(self):
+        batch_sizes = []
+
+        @find_executable_batch_size(starting_batch_size=16)
+        def mock_training_loop_function(batch_size):
+            nonlocal batch_sizes
+            batch_sizes.append(batch_size)
+            if batch_size != 8:
+                # Message raised by the MPS allocator, which has no trailing period after "out of memory".
+                raise RuntimeError(
+                    "MPS backend out of memory (MPS allocated: 8.00 GB, other allocations: 384.00 KB, max allowed: "
+                    "9.07 GB). Tried to allocate 2.00 GB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 "
+                    "to disable upper limit for memory allocations (may cause system failure)."
+                )
+
+        mock_training_loop_function()
+        assert batch_sizes == [16, 14, 12, 10, 9, 8]
+
     def test_start_zero(self):
         @find_executable_batch_size(starting_batch_size=0)
         def mock_training_loop_function(batch_size):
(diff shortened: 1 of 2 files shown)


tests/test_memory_utils.py
```diff
@@ -126,6 +126,24 @@ def mock_training_loop_function(batch_size, arg1):
...
+                # Message raised by the MPS allocator, which has no trailing period after "out of memory".
+                raise RuntimeError(
+                    "MPS backend out of memory (MPS allocated: 8.00 GB, other allocations: 384.00 KB, max allowed: "
+                    "9.07 GB). Tried to allocate 2.00 GB on private pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 "
+                    "to disable upper limit for memory allocations (may cause system failure)."
+                )
+
+        mock_training_loop_function()
+        assert batch_sizes == [16, 14, 12, 10, 9, 8]
+
```
SunMarc: we can't test on mps on the CI in any case but i've tested locally
```suggestion
```

SunMarc (approved):
Thanks a lot !

PR #4234, opened 2026-09-09
Capture stateful dataloader state lazily so prepare() does not draw the shuffle permutation

### LINKED ISSUE #4204: `use_stateful_dataloader` + `num_workers>0` draws epoch-0 permutation during `prepare()` before cross-rank RNG sync — corrupts DDP data partition
### System Info

```Shell
- `accelerate` version: 1.13.0 (and main branch)
- `torch` version: 2.13.0
- `torchdata` version: 0.11.0
- Platform: Linux
- Python version: 3.11
- Distributed environment: 2-process CPU/Gloo (reproducible without GPU)
```

### Information

- [x] My own modified scripts

### Tasks

#### Summary & Root Cause

With `DataLoaderConfiguration(use_stateful_dataloader=True)`, `accelerator.prepare()` wraps the dataloader as a torchdata `StatefulDataLoader` and immediately invokes `base_dataloader.state_dict()` inside `DataLoaderAdapter.__init__` to record the initial dataloader state.

When `num_workers > 0`:
1. `StatefulDataLoader.state_dict()` calls `self._get_iterator()` when `self._iterator is None` to snapshot iterator state.
2. The multiprocessing iterator's `_reset()` spawns worker processes and primes the prefetch queue (`prefetch_factor * num_workers` items), which iterates `_sampler_iter`.
3. For a vanilla `RandomSampler(generator=None)`, this immediately draws the epoch-0 permutation seed from the process's **local global RNG at `prepare()` time**.
4. Cross-rank dataloader RNG synchronization (`DataLoaderShard.__iter__` -> `synchronize_rng_states`) only happens when iteration actually starts, so any divergence in per-rank RNG before `prepare()` (e.g. no global `torch.manual_seed` set, deliberate per-rank seeding, or unequal RNG consumption across ranks during model/dataset setup) causes ranks to derive **different** epoch-0 permutations.
5. `BatchSamplerShard` then rank-interleaves divergent permutations, which **corrupts the data partition during epoch 0** (samples are duplicated across ranks and others are dropped).

*(Note: This affects uninterrupted training runs during epoch 0, independent of any checkpoint/resume operation).*

#### Minimal Reproducible Example

Run with 2 processes on CPU (no GPU needed):

```bash
# Case A (diverged/unseeded RNG across ranks): partition is corrupted
python -m torch.distributed.run --standalone --nproc_per_node=2 repro_prepare_prefetch.py noseed

# Case B (identical seed explicitly set before prepare): partition is intact
python -m torch.distributed.run --standalone --nproc_per_node=2 repro_prepare_prefetch.py seed
```

`repro_prepare_prefetch.py`:

```python
import sys
import torch

MODE = sys.argv[1]  # "seed" or "noseed"
N, BS = 96, 4

if MODE == "seed":
    torch.manual_seed(1234)  # identical on both ranks before prepare()

acc = Accelerator(cpu=True, dataloader_config=DataLoaderConfiguration(use_stateful_dataloader=True))
loader = acc.prepare(
    DataLoader(TensorDataset(torch.arange(N)), batch_size=BS, shuffle=True, num_workers=6)
)

# Iterate through epoch 0
flat = [i for b in loader for i in b[0].tolist()]

gathered = [None] * acc.num_processes
dist.all_gather_object(gathered, flat)
acc.wait_for_everyone()

if acc.is_main_process:
    union = [i for rank_items in gathered for i in rank_items]
    union_set = set(union)
    missing = len(set(range(N)) - union_set)
    duplicates = len(union) - len(union_set)
    print(
        f"mode={MODE}: epoch-0 union={len(union_set)}/{N}, missing={missing}, duplicates={duplicates}"
    )
```

#### Observed Output

```
mode=noseed: epoch-0 union=72/96, missing=24, duplicates=24
mode=seed:   epoch-0 union=96/96, missing=0, duplicates=0
```

With `num_workers=6` and no prior shared seed synchronization, 24 out of 96 samples are duplicated across ranks and 24 samples are completely omitted from epoch 0.

### Expected behavior

`accelerator.prepare(dataloader)` should not draw from the sampler or consume generator RNG before iteration begins:

1. **Lazy state capture (Preferred)**: `DataLoaderAdapter.__init__` should avoid materializing `_get_iterator()` during initialization, capturing state lazily on the first step/checkpoint instead.
2. **Post-init cleanup / reset**: If `state_dict()` must be called during `prepare_data_loader`, the early iterator should be cleared (`dataloader.base_dataloader._iterator = None` or equivalent) to reap worker pools and defer the sampler draw to iteration time.
3. At minimum, `synchronize_rng_states` should run prior to adapter initial state capture if `use_stateful_dataloader=True` and `num_workers > 0`.

jayzuccarelli:
# What does this PR do?

Fixes #4204

This is the lazy capture I proposed on the issue on Sep 2.

`DataLoaderAdapter.__init__` calls `base_dataloader.state_dict()` right after building the `StatefulDataLoader`. With `num_workers > 0` that builds the iterator and draws the sampler permutation at `prepare()` time, before `synchronize_rng_states` runs in `__iter__`, so each rank shards a different epoch-0 permutation.

Capture it lazily instead: `dl_state_dict` starts as `None` and `_update_state_dict` fills it once iteration starts. `state_dict()` falls back to the base dataloader if nothing was captured yet, so checkpointing before the first batch works as before.

Added a single-process test (shuffle order follows the seed set after init) and a 2-process `num_workers > 0` case in `test_distributed_data_loop.py`. Both fail on main.

## Before submitting
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? #4204

## Who can review?

@SunMarc

### DIFF
--- a/src/accelerate/data_loader.py
+++ b/src/accelerate/data_loader.py
@@ -441,7 +441,9 @@ class DataLoaderAdapter:
             self.base_dataloader = DataLoader(dataset, batch_sampler=batch_sampler, **kwargs)
 
         if hasattr(self.base_dataloader, "state_dict"):
-            self.dl_state_dict = self.base_dataloader.state_dict()
+            # Captured lazily in `_update_state_dict`: with `num_workers > 0`, calling `state_dict()` here would
+            # build the iterator and draw the sampler permutation before the RNG is synchronized across processes.
+            self.dl_state_dict = None
 
     def __getattr__(self, name):
         # Avoid infinite recursion if we try to access a nonexistent base_dataloader attribute.
@@ -451,6 +453,8 @@ class DataLoaderAdapter:
         return getattr(self.base_dataloader, name)
 
     def state_dict(self):
+        if self.dl_state_dict is None:
+            return self.base_dataloader.state_dict()
         return self.dl_state_dict
 
     def load_state_dict(self, state_dict):
--- a/src/accelerate/test_utils/scripts/test_distributed_data_loop.py
+++ b/src/accelerate/test_utils/scripts/test_distributed_data_loop.py
@@ -312,6 +312,21 @@ def _test_stateful_dataloader_resume(accelerator, iterable):
         accelerator.dataloader_config = old_dataloader_config
 
 
+def test_stateful_dataloader_with_workers(accelerator):
+    """
+    With `num_workers > 0`, capturing the state dict at prepare time used to build the iterator and draw the epoch-0
+    permutation before the RNG was synchronized across processes, so each rank sharded a different permutation.
+    """
+    old_dataloader_config = accelerator.dataloader_config
+    try:
+        accelerator.dataloader_config = DataLoaderConfiguration(use_stateful_dataloader=True)
+        torch.manual_seed(accelerator.process_index)
+        loader = DataLoader(DummyDataset(), shuffle=True, batch_size=BATCH_SIZE, num_workers=NUM_WORKERS)
+        test_data_loader(loader, accelerator)
+    finally:
+        accelerator.dataloader_config = old_dataloader_config
+
+
 def test_stateful_dataloader(accelerator):
     """
     Tests that a stateful dataloader can be iterated over, saved after a few batches using `load_state_dict`, and then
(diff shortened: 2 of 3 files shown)


jayzuccarelli:
@SunMarc gentle ping when you have a moment. This is the only open PR for #4204: stateful dataloader state is captured lazily, so `prepare()` no longer draws from the shuffle generator. Happy to adjust anything.

jayzuccarelli:
Added a regression test for the resume case @SeverinVisionary reported in #4204: save mid-epoch on a shuffled loader, resume into a fresh one, check the remainder matches (`num_workers` 0 and 2). It fails on main and passes with this change.

PR #4267, opened 2026-09-15
Filter nonpersistent buffers by their qualified names

vineethsaivs:
Recursive tensor enumeration retains nonpersistent child buffers and can drop a persistent root buffer with the same leaf name.

**Cause:** named_buffers returns qualified names, while the exclusion set contains unqualified names. Qualified collection also adds a spurious leading dot for root buffers.

**Fix:** Collect qualified exclusion names and preserve root names without a leading dot.

**Test:** `PYTHONPATH=src python -m pytest tests/test_modeling_utils.py -k 'named_module_tensors or non_persistent' -q`. The new nested-buffer regression fails before; all three selected tests pass after. The expected saved names are checked against torch.nn.Module.state_dict(). Changed-file pre-commit passes. Native CPU tests; no distributed or GPU run.


vineethsaivs:
Could a maintainer approve the CI workflows? The focused buffer tests and changed-file pre-commit checks pass locally.

PR #4273, opened 2026-09-16
fix: bind model_index by value in MegatronEngine's param_sync_func callbacks

kratos0718:
## What does this PR do?

`MegatronEngine.get_module_config` builds one callback per model chunk:

```python
config.param_sync_func = [
    lambda x: self.optimizer.finish_param_sync(model_index, x)
    for model_index in range(len(self.module))
]
```

The lambda captures `model_index` by reference, not by value - a closure binds the *variable*, not whatever it happened to hold when the lambda was created. By the time any of these callbacks is actually invoked (later, once `overlap_param_gather` and `delay_param_gather` training kicks in), the comprehension has already finished and `model_index` holds its final value - `len(self.module) - 1` - for *every* callback in the list. Whichever model chunk's callback fires, it reports the *last* chunk's index to `finish_param_sync`, not its own.

Fixed with the standard default-argument capture (`model_index=model_index`), which evaluates the current value at lambda-creation time instead of call time, so each callback gets its own bound copy - this is the same list used one line above for `no_sync_func`/`grad_sync_func`, but those bind plain method references (`model_chunk.no_sync`) rather than a lambda closing over the loop variable, so they aren't affected by this.

Found via [codehound](https://github.com/kratos0718/codehound), a static analyzer I built that flags exactly this pattern (a lambda inside a loop that gets stored rather than called immediately).

Fixes # (no issue filed - a self-contained, narrowly-scoped correctness fix)

## Before submitting
- [ ] Was this discussed/approved via a Github issue or the forum? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? (not applicable - internal training-loop wiring, no public API change)

`get_module_config()` itself needs the real `megatron-core` package to exercise directly (`get_args`/`get_model_config` import straight from `megatron.core`, which isn't available in this dev/test environment), so the new test mirrors the exact list-comprehension shape in isolation with a lightweight optimizer stand-in, and asserts each callback reports its own index. Verified it fails against the pre-fix code (all three callbacks report index `2`) and passes against the fix.

## Who can review?

Tagging @SunMarc per the "Core parts of the library" note above.


SunMarc (commented):
We don't accept fixes for megatron anymore

PR #4274, opened 2026-09-16
Return the wrapped optimizer step result

vineethsaivs:
## What does this PR do?

Preserve the result of `optimizer.step()`. After `Accelerator.prepare()`, LBFGS currently returns `None` instead of the closure loss because the wrapper discards the result. Return it from both step paths; skipped steps still return `None`.

Test: `PYTHONPATH=src python -m pytest tests/test_optimizer.py -k OptimizerReturn -q`. The CPU regression fails before with `None` instead of `4.0`, passes after, and checks that LBFGS lowers the loss. Pre-commit passes.

## Before submitting

- [x] Read the contribution guidelines.
- [x] Added a regression test.
- Issue discussion: not previously discussed.
- Documentation: existing optimizer return behavior is restored.


vineethsaivs:
Could you approve the CI run? The focused CPU tests and changed-file pre-commit checks pass.

PR #4275, opened 2026-09-17
Clamp SkipBatchSampler and SkipDataLoader __len__ to zero when skip_batches exceeds the dataloader

yoarajota:
`skip_first_batches(dataloader, num_batches)` returns an object whose `__len__` is negative when `num_batches` is larger than the number of batches in `dataloader`. Calling `len()` on it raises:

```
ValueError: __len__() should return >= 0
```

Repro on `main` (commit f13f7c13):

```python
from torch.utils.data import BatchSampler, DataLoader
from accelerate.data_loader import SkipBatchSampler, SkipDataLoader, skip_first_batches

dl = DataLoader(list(range(8)), batch_size=4)  # 2 batches
skipped = skip_first_batches(dl, num_batches=10)
len(skipped)  # ValueError: __len__() should return >= 0

len(SkipBatchSampler(BatchSampler(range(8), batch_size=4, drop_last=False), skip_batches=10))  # same
len(SkipDataLoader(list(range(8)), batch_size=4, skip_batches=10))  # same
```

`__iter__` on all three already handles this: it yields nothing. Only `__len__` is out of step.

`SkipBatchSampler.__len__` (data_loader.py:1352) and `SkipDataLoader.__len__` (data_loader.py:1383) subtract without a floor:

```python
return len(self.batch_sampler) - self.skip_batches
return len(self.base_dataloader) - self.skip_batches
```

CPython rejects a negative `__len__` return, so `len(dataloader)` raises. So does `list(dataloader)`, which calls `operator.length_hint`, and so does any training loop that reads `len(dataloader)` for a progress bar or a scheduler step total.

### Fix

```python
return max(len(self.batch_sampler) - self.skip_batches, 0)
return max(len(self.base_dataloader) - self.skip_batches, 0)
```

The `skip_batches <= len` path is unchanged.

### When this happens

Resume-from-checkpoint calls `skip_first_batches(dataloader, resume_step)`. If the resuming run builds a shorter dataloader than the run that wrote the checkpoint (smaller world size, larger batch size, `drop_last` turned on), `resume_step` can land past the end. Today that is a `ValueError` from `len()`; with the floor the loop sees an empty dataloader and moves on to the next epoch.

### Test

`DataLoaderTester::test_skip_more_batches_than_available` covers the three constructors.

```
# this branch
$ python -m pytest tests/test_data_loader.py -q
25 passed, 13 skipped

# new test against unmodified main
$ python -m pytest tests/test_data_loader.py::DataLoaderTester::test_skip_more_batches_than_available -q
1 failed
E   ValueError: __len__() should return >= 0
```

The 13 skips are the `StatefulDataLoaderTester` cases with `torchdata` absent; the count is the same on `main`. `ruff==0.13.1 check` and `ruff format --check` pass on both files.


yoarajota:
Closing as a duplicate — #4264 already covers this exact fix (the same clamp on `SkipBatchSampler.__len__`/`SkipDataLoader.__len__`), opened a couple of days before this one. Apologies for the noise.

PR #4278, opened 2026-09-18
[Fix] Support custom accelerators via torch.accelerator and fix model.to(device) bug introduced in #4083

### LINKED ISSUE #3813: Supporting new hardware
Hi, this is Haifeng from Google.
We are working closely with the PyTorch community to support Google's accelerators,
which is currently not supported by `accelerate` (for example, `device_map='auto'` would not work since `accelerate` does not discover any device not in its supported devices.) since it is not released to the public nor built into the PyTorch core library yet.
However, we do expect it to be an officially supported hardware by PyTorch soon.

We do not want our customers to wait too long to use `accelerate` with it after we release it to the public.
I wonder how we can better support it in the `accelerate` library,
and if there is anything we should do in advance before its release.

Thanks!

haifeng-jin:
# What does this PR do?

Fixes #3813

### Context & Problem
In Accelerate, auto-sharding (`device_map="auto"`) and model dispatch rely on discovery and resolution of accelerator devices. Historically, device discovery (`get_max_memory`, `get_balanced_memory`) and integer-device resolution (`dispatch_model`, `set_module_tensor_to_device`) have maintained an explicit hardcoded list of hardware backends (`npu`, `mlu`, `sdaa`, `musa`, `neuron`, `xpu`, `hpu`, etc.).

This PR addresses two issues:

1. **Bug introduced in PR #4083 (`model.to(device)` is not called for accelerator devices)**:
   In PR #4083 (commit `2f87576d`), the previously unindented call `if device != "disk": model.to(device)` in `dispatch_model` was mistakenly changed to `else: model.to(device)` attached directly to the `if is_npu_available(): ... elif is_neuron_available():` chain.
   Because of this bug, `model.to(device)` is **never called** for any of the matched accelerator devices (`npu`, `mlu`, `sdaa`, `musa`, `neuron`). Instead, execution only enters the accelerator-specific branch (which sets the device index) and skips `model.to(device)` entirely, failing to move the model to the target device.

2. **Lack of support for custom / out-of-tree accelerators**:
   Any hardware accelerator adhering to PyTorch's official [Accelerator Integration standard](https://pytorch.org/docs/stable/accelerator/index.html) but not hardcoded in Accelerate (such as Google TPU via PyTorch's native `torch_tpu`, or other custom out-of-tree accelerators) currently falls through to CUDA:
   - `get_max_memory()`: Calls `torch.cuda.device_count()`, discovering 0 devices and returning only `cpu`. As a result, `device_map="auto"` assigns all layers to CPU.
   - `dispatch_model()`: When given `device_map={"": 0}`, calling `model.to(0)` defaults to `cuda:0` in PyTorch C++ rather than the active accelerator.

### Solution
1. **Fix the PR #4083 bug**: Un-indents `if device != "disk": model.to(device)` out of the `elif` ladder so `model.to(device)` is properly called for all accelerator devices as originally intended prior to PR #4083.
2. **Generic `torch.accelerator` integration**: Adds `torch.accelerator` support as the **final `elif`** before the fallback `else:` across the modeling pipeline:
   - **`get_max_memory`**: Discovers available device count and memory info via `torch.accelerator.device_count()` and `torch.accelerator.get_memory_info()`.
   - **`get_balanced_memory`**: Resolves `expected_device_type` via `torch.accelerator.current_accelerator().type`.
   - **`set_module_tensor_to_device` & `dispatch_model`**: Resolves integer device index to `f"{acc_type}:{device}"` before calling `model.to(device)`.
3. **Unit test**: Added `test_dispatch_model_with_accelerator` in `tests/test_big_modeling.py`.

### Backwards Compatibility
- **Zero regression risk for existing backends**: By placing `torch.accelerator` as the final `elif`, all existing vendor checks (`npu`, `mlu`, `sdaa`, `musa`, `neuron`, `hpu`, `cuda`) remain 100% untouched and follow their existing paths.
- **Graceful fallback**: If `torch.accelerator` is unavailable (e.g. older PyTorch versions), execution safely falls through to the existing `else:` branch.

## Before submitting
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case. (Fixes #3813)

## Who can review?
- Big modeling: @SunMarc


src/accelerate/utils/modeling.py
```diff
@@ -812,6 +814,19 @@ def get_max_memory(max_memory: Optional[dict[Union[int, str], Union[int, str]]]
...
+                try:
+                    _ = torch.tensor(0, device=torch.device(acc_type, i))
+                    if hasattr(torch.accelerator, "get_memory_info"):
+                        max_memory[i] = torch.accelerator.get_memory_info(i)[0]
+                    else:
+                        dev_mod = torch.get_device_module(acc_type)
+                        max_memory[i] = dev_mod.mem_get_info(i)[0]
+                except Exception:
+                    logger.info(f"Device {i} seems unavailable, Proceeding to check subsequent devices.")
+                    continue
```
SunMarc: maybe we should first deal with cuda then do that as a last resort ? otherwise, cuda will also take this path no ?

src/accelerate/utils/modeling.py
```diff
@@ -850,6 +865,8 @@ def get_max_memory(max_memory: Optional[dict[Union[int, str], Union[int, str]]]
         num_devices = torch.xpu.device_count()
     elif is_hpu_available():
         num_devices = torch.hpu.device_count()
+    elif hasattr(torch, "accelerator") and torch.accelerator.is_available():
+        num_devices = torch.accelerator.device_count()
```
SunMarc: same

src/accelerate/utils/modeling.py
```diff
@@ -985,6 +1002,8 @@ def get_balanced_memory(
         expected_device_type = "hpu"
     elif is_mps_available():
         expected_device_type = "mps"
+    elif hasattr(torch, "accelerator") and torch.accelerator.is_available():
+        expected_device_type = torch.accelerator.current_accelerator().type
```
SunMarc: same

src/accelerate/big_modeling.py
```diff
@@ -510,8 +510,9 @@ def wrapper(*args, **kwargs):
             device = f"musa:{device}"
         elif is_neuron_available() and isinstance(device, int):
             device = f"neuron:{device}"
-        else:
-            model.to(device)
+        elif isinstance(device, int) and hasattr(torch, "accelerator") and torch.accelerator.is_available():
+            device = f"{torch.accelerator.current_accelerator().type}:{device}"
+        model.to(device)
```
SunMarc: since we didn't do it for cuda, there must be a reason so don't update for cuda

src/accelerate/utils/modeling.py
```diff
@@ -339,6 +339,8 @@ def set_module_tensor_to_device(
                 device = f"musa:{device}"
             elif is_hpu_available():
                 device = "hpu"
+            elif hasattr(torch, "accelerator") and torch.accelerator.is_available():
+                device = f"{torch.accelerator.current_accelerator().type}:{device}"
```
SunMarc: just guard for cuda as we don't do it

SunMarc (commented):
Thanks, just a few nits

src/accelerate/utils/modeling.py
haifeng-jin: You are right! instead of "if cuda", I excluded cuda from this if.
This would keep the cuda path 100% intact.

src/accelerate/utils/modeling.py
haifeng-jin: done

src/accelerate/utils/modeling.py
haifeng-jin: done

src/accelerate/utils/modeling.py
haifeng-jin: done

src/accelerate/big_modeling.py
haifeng-jin: done. excluded cuda from this if.

haifeng-jin:
Thanks, @SunMarc !
I explicitly checked and excluded cuda from the newly added if, so the cuda path is 100% intact.

PR #4291, opened 2026-09-20
Fix FSDP2 fp16 checkpointing advancing the GradScaler it is restoring

### LINKED ISSUE #4289: [Bug] FSDP2 + fp16: load_state() does not round-trip the GradScaler state

HaomingSong:
# What does this PR do?

Fixes #4289

## The fix

Hand torch's optimizer state-dict APIs the unwrapped `torch.optim.Optimizer` in both helpers. The dummy step then runs the plain optimizer's `step()` and cannot reach the scaler.

With the scaler off that path, the pre-initialization is no longer needed, so the scaler goes back through `load_accelerator_state` like every other backend. That also removes the "the scaler must be loaded before the optimizer" ordering constraint.

FSDP1 is untouched: it goes through `FSDP.optim_state_dict` / `FSDP.optim_state_dict_to_load`, neither of which calls `_init_optim_state`. Only the `fsdp_version == 2` branches changed.

## Validation

**New tests** (`tests/fsdp/test_fsdp.py::FSDP2OptimizerScalerStateTest`) drive the real `save_fsdp_optimizer` / `load_fsdp_optimizer`, single-process on CPU ( torch's DCP falls back to a single-rank checkpoint there) . This follows `FSDP2PeftStateDictTest` in the same file: the helper level is testable without devices, and multi-rank behaviour stays with the launcher-based tests.

| Test | w/ fix | Call sites reverted |
| --- | --- | --- |
| `test_load_preserves_a_lazy_scaler` | pass | fail — ` _scale is None` |
| `test_load_preserves_an_already_used_scaler` | pass | fail — scale 1024.0 → 2048.0 |
| `test_save_preserves_the_scaler_when_the_optimizer_state_is_empty` | pass | fail — scale 1024.0 → 2048.0 |
| `test_unwrap_is_idempotent_and_leaves_plain_optimizers_alone` | pass | pass (the helper is unchanged) |

**End to end**, 2x H200, real FSDP2 + fp16 through `Accelerator.save_state()` / `load_state()` (the reproduction in #4289), on this PR:

```text
[load] scaler.pt on disk      : {'scale': 1024.0, ..., '_growth_tracker': 1}
[load] scaler after load_state: {'scale': 1024.0, ..., '_growth_tracker': 1}
[load] ROUND-TRIPS            : True
[same-process] no error
```

**Environment:** accelerate at `b795b483`, torch 2.10.0+cu128, Python 3.12.12, NumPy 2.2.6, Linux x86_64, 2x NVIDIA H200 (driver 595.71.05, CUDA 13.2).

**Quality:** `ruff check .` and `ruff format --check .` pass with the pinned ruff 0.13.1 (`make quality`), and `git diff --check` is clean.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

- Fully-Sharded Data Parallism: @SunMarc


src/accelerate/utils/fsdp_utils.py
```diff
@@ -288,6 +288,17 @@ def load_fsdp_model(fsdp_plugin, accelerator, model, input_dir, model_index=0, a
...
+def _unwrap_accelerated_optimizer(optimizer):
+    """Return the plain `torch.optim.Optimizer` underneath accelerate's wrapper."""
+    # Imported here rather than at module scope: `accelerate.optimizer` imports `accelerate.utils`,
+    # so a top-level import would be circular.
+    from ..optimizer import AcceleratedOptimizer
+
+    while isinstance(optimizer, AcceleratedOptimizer):
+        optimizer = optimizer.optimizer
+    return optimizer
+
```
SunMarc: Just rename it as _unwrap_optimizer(optimizer)

src/accelerate/utils/fsdp_utils.py
```diff
@@ -288,6 +288,17 @@ def load_fsdp_model(fsdp_plugin, accelerator, model, input_dir, model_index=0, a
     return load_result
 
 
+def _unwrap_accelerated_optimizer(optimizer):
+    """Return the plain `torch.optim.Optimizer` underneath accelerate's wrapper."""
+    # Imported here rather than at module scope: `accelerate.optimizer` imports `accelerate.utils`,
+    # so a top-level import would be circular.
```
```suggestion
```

src/accelerate/utils/fsdp_utils.py
```diff
@@ -323,7 +334,7 @@ def save_fsdp_optimizer(fsdp_plugin, accelerator, optimizer, model, output_dir,
         if fsdp_plugin.fsdp_version == 2:
             from torch.distributed.checkpoint.state_dict import get_optimizer_state_dict
 
-            optim_state = get_optimizer_state_dict(model, optimizer, options=sd_options)
+            optim_state = get_optimizer_state_dict(model, _unwrap_accelerated_optimizer(optimizer), options=sd_options)
```
SunMarc: add a small comment on why we unwrap the optimizer only for fsdpv2. keep it short. or we could add the explanation in the unwrap function

src/accelerate/accelerator.py
```diff
@@ -3860,20 +3859,10 @@ def _inner(folder):
...
-            self.scaler.load_state_dict(scaler_state)
-            # We also need to call the `_lazy_init_scale_growth_tracker` to initialize the scaler, as it would else be called
-            # on the first call to scale
-            self.scaler._lazy_init_scale_growth_tracker(self.scaler._device)
-            logger.info("GradScaler state loaded successfully")
-        else:
-            scaler = self.scaler
+        # Restored below by `load_accelerator_state`, on the same path as every other backend:
+        # `load_fsdp_optimizer` hands torch's optimizer state-dict API the unwrapped optimizer, so
+        # nothing between here and the first `backward()` can reach the scaler.
```
SunMarc: Thanks for the clean up ! 
```suggestion
```

tests/fsdp/test_fsdp.py
```diff
@@ -128,6 +134,98 @@ def test_adapter_state_dict_round_trip(self):
         assert restored == {0.5}
 
 
+@require_fsdp2
+class FSDP2OptimizerScalerStateTest(AccelerateTestCase):
+    """Saving and loading FSDP2 optimizer state must leave an fp16 `GradScaler` untouched.
+
```
SunMarc: Can we keep it simple ? just 1 or 2 tests should be enough for this. Try to reduce the comments also

SunMarc (commented):
Thanks a lot for fixing this behavior ! Left a few comments that should be easy to fix

src/accelerate/utils/fsdp_utils.py
HaomingSong: thanks! renamed in #[0ff96b3](https://github.com/huggingface/accelerate/pull/4291/commits/0ff96b3a79839dfc27fa4960a12bd4ba15a54ac5)

src/accelerate/utils/fsdp_utils.py
HaomingSong: thanks! add explation in the unwrap function in [8a8d4a8](https://github.com/huggingface/accelerate/pull/4291/commits/8a8d4a84106c396f5a47f9d672afb237f14a36e8)

tests/fsdp/test_fsdp.py
HaomingSong: yep! [4ff90cb](https://github.com/huggingface/accelerate/pull/4291/commits/4ff90cbf5398f76449baeb6cbeaf3cb27f31bfd1) reduced the test to 2 (1 save + 1 load) and also reduced the comments

HaomingSong:
Thanks for the comments @SunMarc ! Fixed them in the latest commits. Happy to make further changes if needed :)

HaomingSong:
Explanation for [`1cd48f4`](https://github.com/huggingface/accelerate/pull/4291/commits/1cd48f4232b3859029fc2a8bd217a045951d8e46): 

When CI runs on torch 2.5.1 ([`test.yml#L58`](https://github.com/huggingface/accelerate/blob/e3188ec078c39e5eccb84a2bbf7567a461eaccdb/.github/workflows/test.yml#L58)), which is below [`FSDP2_PYTORCH_VERSION`](https://github.com/huggingface/accelerate/blob/e3188ec078c39e5eccb84a2bbf7567a461eaccdb/src/accelerate/utils/constants.py#L46) (2.6.0), the [`FullyShardedDataParallelPlugin`](https://github.com/huggingface/accelerate/blob/e3188ec078c39e5eccb84a2bbf7567a461eaccdb/src/accelerate/utils/dataclasses.py#L1834-L1835) will raise `ImportError: FSDP2 requires PyTorch >= 2.6.0`. Because the new `FSDP2OptimizerScalerStateTest` instantiate [`FullyShardedDataParallelPlugin(fsdp_version=2)`](https://github.com/huggingface/accelerate/blob/1cd48f4232b3859029fc2a8bd217a045951d8e46/tests/fsdp/test_fsdp.py#L142)) in unit test. 

Therefore [`1cd48f4`](https://github.com/huggingface/accelerate/pull/4291/commits/1cd48f4232b3859029fc2a8bd217a045951d8e46) makes `require_fsdp2` use `FSDP2_PYTORCH_VERSION` instead of the hard-coded [`2.5.0`](https://github.com/huggingface/accelerate/blob/e3188ec078c39e5eccb84a2bbf7567a461eaccdb/src/accelerate/test_utils/testing.py#L260), to make test skip on torch < 2.6.

SunMarc (approved):
Thanks a lot, much cleaner

PR #4300, opened 2026-09-21
Test causal-LM training with DDP, gradient accumulation and mixed precision

Titus-von-Koeller:
Distributed training can finish successfully while producing incorrectly scaled updates or never exercising the requested precision. This adds an offline regression test comparing Accelerate DDP with a separate plain-PyTorch training loop, using the same precision and effective batch.

Four ordinary cases cover FP32, FP32 with gradient accumulation, BF16 and FP16. They compare losses and parameter updates, check agreement between ranks, and verify that the requested precision is exercised. A separate FP16 recovery test explicitly injects an infinite gradient at the second update attempt on every rank. It checks that this update is skipped, weights stay unchanged, the loss scale falls, and subsequent updates succeed. A tiny randomly initialized Gemma 4 model and synthetic data keep the tests offline.

## Validation

All five training cases and 14 CPU configuration checks pass with Torch 2.5.1 / Transformers 5.17.0 and Torch 2.13.0 / Transformers 5.14.1, using two RTX 4090s for training. Deliberately halving gradients, disabling autocast or perturbing one rank's parameters fails the corresponding checks. Missing or mistimed fault injection also fails the recovery test. The relative update check catches scaling errors that small absolute parameter differences can hide: halved gradients produce about 51% error against the unchanged 15% bound.

The reference bypasses Accelerate's training wrappers but shares PyTorch and Transformers. These cases use equal valid-target counts and complete accumulation windows. Hosted GPU validation remains outstanding.

Related: [#4303](https://github.com/huggingface/accelerate/pull/4303) fixes labels and loss normalization in the accumulation example. The PRs can be merged independently.


SunMarc:
Please make the description easier to read, okay to have llm help you but it should still feel like you wrote them or do a shorter description. see pablo [advice](https://huggingface.slack.com/archives/C032L2251NG/p1789974601183279?thread_ts=1789541536.271359&cid=C032L2251NG)

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+
+def flatten_parameters(model):
+    return torch.cat([p.detach().flatten() for p in model.parameters()])
+
+
+def create_model():
+    """Create a tiny, randomly initialized Gemma 4 text model for offline tests."""
+    return Gemma4ForCausalLM(
+        Gemma4TextConfig(
+            vocab_size=32,
```
SunMarc, line 47: don't create a model from config but use an existing checkpoint. check what trl or what transformers trainer do

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+    gradient_accumulation_steps,
+):
+    losses, step_was_skipped, loss_scales, parameters_unchanged_during_accumulation, parameters_changed = (
+        [],
+        [],
+        [],
+        [],
+        [],
+    )
+
```
SunMarc: keep the training script as simple as possible

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+            if mixed_precision == "fp16" and len(losses) == 1 and accelerator.sync_gradients:
+                next(model.parameters()).grad.fill_(float("inf"))
+            optimizer.step()
+            optimizer.zero_grad()
+
+            window_loss += loss.detach() / gradient_accumulation_steps
+            if accelerator.sync_gradients:
+                did_skip = accelerator.optimizer_step_was_skipped
+                if did_skip:
+                    assert torch.equal(parameters_before_step, flatten_parameters(model))
```
SunMarc: i don't see the point of checking this

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+            parameters_before_step = flatten_parameters(model).clone()
+            loss = model(input_ids=batch, labels=batch).loss
+            accelerator.backward(loss)
+
+            if mixed_precision == "fp16" and len(losses) == 1 and accelerator.sync_gradients:
+                next(model.parameters()).grad.fill_(float("inf"))
+            optimizer.step()
+            optimizer.zero_grad()
+
+            window_loss += loss.detach() / gradient_accumulation_steps
```
SunMarc, line 172: We shouldn't have to scale the loss ourselves no ? accelerate should handle it

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+            if accelerator.sync_gradients:
+                did_skip = accelerator.optimizer_step_was_skipped
+                if did_skip:
+                    assert torch.equal(parameters_before_step, flatten_parameters(model))
+                # Every microbatch has equal valid-target counts, so this mean
+                # reports the same effective-batch objective as the reference.
+                losses.append(accelerator.reduce(window_loss, reduction="mean").item())
+                parameters_changed.append(not torch.equal(parameters_before_step, flatten_parameters(model)))
```
SunMarc: i still don't understand why we care about this

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+    window_loss = torch.zeros((), device=accelerator.device)
+
+    for (batch,) in dataloader:
+        with accelerator.accumulate(model):
+            parameters_before_step = flatten_parameters(model).clone()
+            loss = model(input_ids=batch, labels=batch).loss
+            accelerator.backward(loss)
+
+            if mixed_precision == "fp16" and len(losses) == 1 and accelerator.sync_gradients:
+                next(model.parameters()).grad.fill_(float("inf"))
```
SunMarc: why this ?

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+
+def main():
+    args = parse_args()
+
+    accelerator = None
+    if not args.reference:
+        accelerator = Accelerator(
+            mixed_precision=args.mixed_precision,
+            gradient_accumulation_steps=args.gradient_accumulation_steps,
+            kwargs_handlers=[GradScalerKwargs(init_scale=128.0)],
```
SunMarc: why pass this ? is the default value not good enough ?

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # Limit numerical variation when comparing different batch layouts and process counts.
+    torch.set_float32_matmul_precision("highest")
+    set_seed(1337, deterministic=True)
+
+    num_updates = 10
+    effective_batch_size = 8
+    input_ids = torch.randint(
+        3, 32, (num_updates * effective_batch_size, 12), generator=torch.Generator().manual_seed(1337)
+    )
```
SunMarc, line 223: can't we use a real dataset to get those input_ids. The goal is to have an integration test that does real training, not just mock the training

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+            gradient_accumulation_steps=args.gradient_accumulation_steps,
+        )
+
+    hook.remove()
+    observations["linear_output_dtypes"] = sorted(linear_output_dtypes)
+    observations["parameter_dtypes"] = sorted({str(p.dtype) for p in model.parameters()})
+    observations = [observations]
+    if accelerator:
+        # DDP replicas must agree exactly, independently of reference tolerances.
+        replicas = accelerator.gather(flatten_parameters(model).unsqueeze(0))
```
SunMarc, line 263: we don't need this, the ddp replicas being the same is tested by torch. we don't need to bother with this

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+                {
+                    "losses": losses,
+                    "parameters": flatten_parameters(model).cpu().tolist(),
+                    "parameter_delta_norm": (flatten_parameters(model).cpu() - initial_parameters).norm().item(),
+                    "world_size": 1 if args.reference else accelerator.num_processes,
+                    "ranks": observations,
+                },
+                allow_nan=False,
+            )
+        )
```
SunMarc, line 279: checking the loss should be enough. if you want , you can also keep the params to do comparison but we shouldn't have to keep `parameter_delta_norm`

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
...
+def parse_args():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--output", type=Path, required=True)
+    parser.add_argument("--reference", action="store_true")
+    parser.add_argument("--mixed-precision", choices=("no", "bf16", "fp16"), default="no")
+    parser.add_argument("--gradient-accumulation-steps", type=int, default=1)
+    return parser.parse_args()
+
+
+def train_reference(model, optimizer, input_ids, effective_batch_size, mixed_precision, device):
```
SunMarc: i don't mind keeping this if you want to do comparison with the base pytorch impl. But it will requires a lot of scripts + modifications. Right now this script only handles mixed precision and not grad acc or ddp. It could a bit complex to deal with all these cases or a mix of them as it would just mean reimplementing accelerate here. For some simple use cases why not

src/accelerate/test_utils/scripts/external_deps/train_causal_lm.py
```diff
@@ -0,0 +1,266 @@
+# Copyright 2026 The HuggingFace Team. All rights reserved.
```
SunMarc, line 1: in general, please simplify the code here. check what is done in trainer or in trl

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+)
+def test_ddp_training_matches_reference(
+    tmp_path,
+    mixed_precision,
+    gradient_accumulation_steps,
+    loss_atol,
+    parameter_atol,
+):
+    """DDP, accumulation and AMP match a same-precision plain-PyTorch reference."""
+    pytest.importorskip("transformers.models.gemma4", reason="Requires Transformers with Gemma 4 support")
```
SunMarc: we don't need this, just checking if transformers is installed should satisfy gemma4 support

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+    if mixed_precision == "bf16":
+        for device in range(2):
+            with torch.cuda.device(device):
+                if not torch.cuda.is_bf16_supported():
+                    pytest.skip("Both CUDA devices must support BF16")
```
SunMarc: we don't need this

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+    config_path = tmp_path / "ddp_config.json"
+    # Do not inherit a developer's default launcher configuration.
+    ClusterConfig(
+        compute_environment=ComputeEnvironment.LOCAL_MACHINE,
+        distributed_type=DistributedType.MULTI_GPU,
+        mixed_precision=mixed_precision,
+        debug=False,
+        use_cpu=False,
+        num_processes=2,
+    ).to_json_file(config_path)
```
SunMarc, line 182: instead of doing that, we should create an accelerate config file and use that. For values that we want to overwrite, we can pass it as a cli arg eg --mixed_precision as you did below

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+    ]
+
+    # Seed Python hashing before launch too: older Gemma 4 implementations register
+    # RoPE buffers from a set, while DDP broadcasts buffers in registration order.
+    with patch_environment(
+        omp_num_threads=1,
+        cublas_workspace_config=":4096:8",
+        hf_hub_offline="1",
+        pythonhashseed="0",
+    ):
```
SunMarc, line 203: why patch ?

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+    launch_command = get_launch_command(config_file=str(config_path), main_process_port=get_torch_dist_unique_port())
+    training_args = [
+        "--mixed-precision",
+        mixed_precision,
+        "--gradient-accumulation-steps",
+        str(gradient_accumulation_steps),
+    ]
+
+    # Seed Python hashing before launch too: older Gemma 4 implementations register
+    # RoPE buffers from a set, while DDP broadcasts buffers in registration order.
```
SunMarc: don't use gemma4 then

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+    for rank_results in ddp_results["ranks"]:
+        assert len(rank_results["parameters_unchanged_during_accumulation"]) == 10 * (gradient_accumulation_steps - 1)
+        assert all(rank_results["parameters_unchanged_during_accumulation"])
+
+    reference_parameters = torch.tensor(reference_results["parameters"], dtype=torch.float64)
+    ddp_parameters = torch.tensor(ddp_results["parameters"], dtype=torch.float64)
+    # Absolute bounds can hide incorrectly scaled small updates. Normalize the
+    # final-parameter error by the reference's net movement from initialization.
+    assert reference_results["parameter_delta_norm"] > 0
+    relative_update_error = (reference_parameters - ddp_parameters).norm().item() / reference_results[
```
SunMarc: too many checks, some of them are not needed at all. keep it simple. checking the loss is already good enough

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+        pytest.param("no", 1, 1e-4, 1e-5, id="fp32_ddp"),
+        # Does it still work when we add gradient accumulation?
+        pytest.param("no", 2, 1e-4, 1e-5, id="fp32_accumulation"),
+        # Does accumulation work with BF16 mixed precision?
+        pytest.param("bf16", 2, 1e-2, 1e-3, id="bf16_accumulation"),
+        # Does it work with FP16 gradient scaling and skipped updates?
+        pytest.param("fp16", 2, 1e-3, 1e-4, id="fp16_accumulation"),
+    ],
+)
+def test_ddp_training_matches_reference(
```
SunMarc, line 57: I don't like a single test that deal with all cases. Check what is done in Trainer or TRL.

tests/test_causal_lm_training.py
```diff
@@ -0,0 +1,153 @@
...
+        omp_num_threads=1,
+        cublas_workspace_config=":4096:8",
+        hf_hub_offline="1",
+        pythonhashseed="0",
+    ):
+        reference_process = execute_subprocess_async(
+            [sys.executable, script_path, "--reference", "--output", str(reference_results_path)] + training_args,
+            timeout=90,
+        )
+        assert reference_process.returncode == 0, f"Reference process failed: {reference_process.stderr}"
```
SunMarc, line 208: the reference is not really a real reference. it's the same as doing accelerate without grad acc on how rank without ddp right ?

SunMarc (commented):
Thanks, left a couple of comments. Make sure you understand all the lines that are written there and question if whether we should keep them or not. The philosophy stated [here](https://huggingface.slack.com/archives/C032L2251NG/p1789541536271359) is also applicable to accelerate.

PR #4302, opened 2026-09-21
Remove duplicate debug key from SageMaker config fixture

Titus-von-Koeller:
Removes a repeated `debug: false` entry that makes the SageMaker fixture fail `check-yaml`. The loaded configuration is unchanged. The YAML hook and all three `ClusterConfigTester` tests pass after the fix.


SunMarc (approved):
Thanks

PR #4304, opened 2026-09-21
Enforce subprocess execution timeouts and clean up interrupted launches

Titus-von-Koeller:
`execute_subprocess_async` waits for output with a timeout, then waits for the process to exit without one. A child that sleeps for five seconds can therefore outlive `timeout=1` and still return successfully. This makes the timeout cover both output and process exit, and raises `TimeoutError` with the output captured so far.

Cleanup also needs to stop workers left by the launcher: they can keep its output pipes open even after it exits. On POSIX, the helper now launches a separate process group and kills that group on timeout, cancellation or reader failure. Detached workers are outside that guarantee, and startup and cleanup can add time beyond the timeout.

The default changes from 180 seconds to `None`: enforcing the old default would kill an existing passing test that [runs for over 330 seconds](https://github.com/huggingface/accelerate/actions/runs/35553590568/job/106192770954). Explicit timeouts are enforced; callers that omit one still need a job-level limit.

The regression tests check that a timed-out child raises an error and that launched processes stop, including when the child closes its output early or its workers outlive it. All 14 cases pass on Python 3.10 and 3.14; timeout regressions fail with the old helper. Native Windows behavior remains unverified.

Related: [#4301](https://github.com/huggingface/accelerate/pull/4301) fixes signal-exit reporting independently. Existing explicit timeouts, including those in [#4300's DDP tests](https://github.com/huggingface/accelerate/pull/4300) and [#4303's example test](https://github.com/huggingface/accelerate/pull/4303), need checking on their actual runners when combined with this fix.


src/accelerate/test_utils/testing.py
```diff
@@ -759,16 +760,9 @@ async def _stream_subprocess(cmd, env=None, stdin=None, timeout=None, quiet=Fals
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
         env=env,
+        start_new_session=os.name == "posix",
```
SunMarc, line 763: why check posix ?

src/accelerate/test_utils/testing.py
```diff
@@ -778,18 +772,56 @@ def tee(line, sink, pipe, label=""):
...
+            # Continue draining after cancellation: waiting for exit alone can
+            # deadlock when a pipe's reader transport is paused on buffered data.
+            stdout, stderr = await p.communicate()
+            out.extend(stdout.decode("utf-8", errors="replace").splitlines())
+            err.extend(stderr.decode("utf-8", errors="replace").splitlines())
+
+        if isinstance(error, asyncio.TimeoutError):
+            stdout, stderr = "\n".join(out), "\n".join(err)
+            raise TimeoutError(
+                f"'{' '.join(map(str, cmd))}' timed out after {timeout} seconds\n\n"
```
SunMarc, line 807: can you explain a bit better why we need to do this ? The description of the PR is a bit too dense. Also is there any tests that were failing due to not having that ? can you show me any minimal tests that fails when using execute_subprocess_async so that i can better understand why we need to do this change ?

SunMarc (commented):
Feels like we are overcomplicating something that works or require minimal fixes. Also you added too many tests .

PR #4305, opened 2026-09-21
Surface the child process error in `simple_launcher` without withholding its output

### LINKED ISSUE #4277: `simple_launcher` discards the child's error, unlike the elastic launch path

albertvillanova:
`simple_launcher` runs the training script with `subprocess.Popen` and, on a non-zero exit, raises a `CalledProcessError` built from the return code and the command alone. The child's stderr is inherited rather than read, so a caller that catches the exception learns only that something exited non-zero.

Fixes #4277.

### Motivation

The multi-process path does not have this problem. `multi_gpu_launcher` delegates to `torch.distributed.run`, and torch elastic propagates the child's root exception to the parent through error files, so the parent raises a `ChildFailedError` carrying the child's traceback. Its docstring gives the rationale: the parent is "a simple nanny process", so "child errors should be propagated to the scheduler for accurate root cause diagnostics". `simple_launcher` is the same nanny.

This matters to callers that handle the failure programmatically rather than by reading a terminal. A test suite that retries known transient infrastructure failures by matching on the exception, for instance, sees `CalledProcessError: Command '[...]' returned non-zero exit status 1.` whether the child hit a genuine bug or ran out of GPU memory, so it cannot tell the two apart.

### Solution

Tee the child's stderr rather than capturing it. Only stderr is piped, so stdout keeps its terminal and draining the pipe on the main thread cannot deadlock. Each chunk is written through to `sys.stderr` as it arrives, so `accelerate launch` still prints live, and a bounded tail, the last 512 KiB, is kept for the error.

On failure that tail is attached two ways:

- as `stderr=` on the `CalledProcessError`, for a caller that wants to read or log it,
- as the chained cause, so the child's error prints above the traceback and is reachable by anything that walks `__cause__`.

The second is what makes the failure legible to a caller that only inspects the exception, since `str(CalledProcessError)` includes neither `output` nor `stderr`.

`--quiet` keeps its current behaviour.

### Tests

Four tests in `tests/test_launch.py`, driven through `launch_command_parser`:

- the child's stderr reaches the terminal and is attached to the raised error, both as the attribute and as the chained cause,
- a successful child's output is not withheld,
- `--quiet` still exits 1 without raising,
- a child that floods stderr keeps the tail bounded while its full output still reaches the terminal. That is the case which deadlocks a `wait()`-then-read launcher and buffers without limit in one that drains with `communicate()`.

Against the current implementation the first and last of those fail, as they should.


albertvillanova:
I think the CI errors are unrelated.

tests/test_launch.py
```diff
@@ -71,3 +81,70 @@ def test_auto_port_selection(self):
...
+        "no",
+        "--num_cpu_threads_per_process",
+        "1",
+    ]
+    if quiet:
+        argv.append("--quiet")
+    return launch_command_parser().parse_args([*argv, script])
+
+
+class TestSimpleLauncher:
```
SunMarc, line 105: If possible, let's just have one test. Also did you check if those test where failing before your fix ?

src/accelerate/commands/launch.py
```diff
@@ -989,14 +994,27 @@ def launch_command_parser(subparsers=None):
     return parser
 
 
+class ChildProcessFailure(RuntimeError):
+    """Tail of a launched process's stderr, raised as the cause so the failure survives the process boundary."""
+
+
 def simple_launcher(args):
```
SunMarc: we don't need this specific class no, just use RuntimeError(stderr)

SunMarc (commented):
Thanks, just a nit

src/accelerate/commands/launch.py
albertvillanova: Done, now `RuntimeError(stderr)`.

tests/test_launch.py
albertvillanova: Consolidated into a single test, and yes.

The child now floods stderr before failing, which covers the write-through, the absence of a deadlock, the bounded tail and the attached cause in one case. I dropped the `--quiet` test, since this PR does not change that path.

Against `main` today the consolidated test fails, at `e.stderr` being `None`. The two write-through assertions do pass there, correctly, because stderr is inherited today. What is missing is the attachment, which is where it fails.

SunMarc (approved):
Thanks !

